PR #2484, opened 2020-07-21
feat(traffictile.h): Adds versioning checks

PureW:
Allocates one of the spare fields in the TrafficTileHeader
for versioning data, with the version being made up of
the first four bytes of the MD5 checksum of baldr/traffictile.h


src/baldr/CMakeLists.txt
```diff
@@ -69,8 +84,9 @@ set(sources
     verbal_text_formatter_us_tx.cc
     verbal_text_formatter_factory.cc)
 
-#basic timezone stuff
 list(APPEND sources
+    ${CMAKE_CURRENT_BINARY_DIR}/traffic_tile_version.h
```
mandeepsandhu, line 88: should we use `GENERATED_VERSION_HEADER` here as well?

PureW: Yup, good catch!

PureW:
The `echo -n` statement failed on Osx so I rewrote it with `printf` and a sub-shell which required a separate script. Lets see if this passes all CI.

valhalla/baldr/traffictile.h
```diff
@@ -149,8 +151,11 @@ class TrafficTile {
   }
 
   const volatile TrafficSpeed& trafficspeed(const uint32_t directed_edge_offset) const {
-    if (header == nullptr)
+    if (header == nullptr) {
+      return INVALID_SPEED;
+    } else if (header->traffic_tile_version != TRAFFIC_TILE_VERSION) {
```
kevinkreiser, line 156: Why not have one if block and use an ||

PureW: Logged different messages while debugging, I just pushed a commit that collaps them :+1:

PureW:
Ok, I finally got all the platforms in CI passing after fixing the nvm setup in Osx and removing the pipes which upset Windows. cc @mandeepsandhu

.circleci/config.yml
```diff
@@ -19,8 +19,8 @@ commands:
             curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.5/install.sh | bash
             echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV
             echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
-            echo 'nvm install v8.11.2' >> $BASH_ENV
-            echo 'nvm alias default v8.11.2' >> $BASH_ENV
+            source $BASH_ENV && nvm install v8.11.2
+            source $BASH_ENV && nvm alias default v8.11.2
```
PureW, line 23: CircleCi is supposed to source this itself, but on Osx it doesn't.

src/baldr/CMakeLists.txt
```diff
@@ -25,12 +25,25 @@ add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/date_time_windows_zones.h
 
 file(GLOB headers ${VALHALLA_SOURCE_DIR}/valhalla/baldr/*.h)
 
+# Builds a checksum of the traffictile header and store it in `traffic_tile_version.h`
+set(GENERATED_VERSION_HEADER "${CMAKE_CURRENT_BINARY_DIR}/traffic_tile_version.h")
+message("Generating traffictile version header")
+configure_file(${VALHALLA_SOURCE_DIR}/valhalla/baldr/traffictile.h
```
PureW, line 31: @danpat suggested a way to evaluate this entirely in cmake, with the hack here to implicitly add a dependency on the file to re-run checksum on changes. My first attempt without this hack only generated the checksum on first run.

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ executors:
 commands:
   install_macos_dependencies:
     steps:
-      - run: brew install protobuf cmake ccache libtool boost-python libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
+      - run: brew install protobuf cmake ccache coreutils libtool boost-python libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
```
danpat, line 13: This change isn't needed now that the checksum is done in pure cmake

PureW: I left it in since it was listed as a build requirement on osx in readme: https://github.com/valhalla/valhalla#building-from-source

But I can remove it, fix pushed.

PR #2481, opened 2020-07-21
feat(gbench): Adds gbench of GetSpeed with traffic data

PureW:
Also breaks out test-utility functions into `test/util/traffic_utils.h

PR #2485, opened 2020-07-22
Refactor mapmatching configuration to use a struct

genadz:
Add configuration struct for mapmatching. It can be initialized with default values or with `boost::property_tree` (if some parameters are missing -> use default values, don't throw any exception).

## Tasklist

 - [ ] Add tests

### DIFF
--- a/src/meili/config.cc
+++ b/src/meili/config.cc
@@ -1,0 +1,112 @@
+#include "meili/config.h"
+
+#include "macro.h"
+#include "midgard/logging.h"
+
+#define NONNEGATIVE_VALUE_MSG(value, name)                                                           \
+  std::string("Expect '") + name + "' to be nonnegative (got: " + std::to_string(value) + ")"
+
+#define POSITIVE_VALUE_MSG(value, name)                                                              \
+  std::string("Expect '") + name + "' to be positive (got: " + std::to_string(value) + ")"
+
+namespace {
+template <typename T>
+inline void
+ReadParamOptional(T& value, const boost::property_tree::ptree& ptree, const std::string& name) {
+  if (auto item = ptree.get_optional<T>(name))
+    value = *item;
+}
+
+inline bool FindValue(const boost::property_tree::ptree& node, const std::string& value) {
+  return std::find_if(node.begin(), node.end(),
+                      [&value](const boost::property_tree::ptree::value_type& item) {
+                        return item.second.get_value<std::string>() == value;
+                      }) != node.end();
+}
+} // namespace
+
+namespace valhalla {
+namespace meili {
+
+Config::Config(const boost::property_tree::ptree& params) {
+  Read(params);
+}
+
+void Config::Read(const boost::property_tree::ptree& params) {
+  candidate_search.Read(params);
+  transition_cost.Read(params);
+  emission_cost.Read(params);
+  routing.Read(params);
+}
+
+void Config::CandidateSearch::Read(const boost::property_tree::ptree& params) {
+  ReadParamOptional(search_radius_meters, params, "default.search_radius");
+  CHECK_THROWS(search_radius_meters >= 0.f,
+               NONNEGATIVE_VALUE_MSG(search_radius_meters, "search_radius"));
+
+  if (const auto node = params.get_child_optional("customizable")) {
+    is_search_radius_customizable = FindValue(*node, "search_radius");
+  }
+
+  ReadParamOptional(max_search_radius_meters, params, "default.max_search_radius");
+  CHECK_THROWS(max_search_radius_meters > 0.f,
+               POSITIVE_VALUE_MSG(max_search_radius_meters, "max_search_radius"));
+
+  ReadParamOptional(cache_size, params, "grid.cache_size");
+  ReadParamOptional(grid_size, params, "grid.size");
+}
+
+void Config::TransitionCost::Read(const boost::property_tree::ptree& params) {
+  ReadParamOptional(beta, params, "default.beta");
+  CHECK_THROWS(beta > 0.f, POSITIVE_VALUE_MSG(beta, "beta"));
+
+  ReadParamOptional(breakage_distance_meters, params, "default.breakage_distance");
+  CHECK_THROWS(breakage_distance_meters > 0.f,
+               POSITIVE_VALUE_MSG(breakage_distance_meters, "breakage_distance"));
+
+  if (const auto node = params.get_child_optional("customizable")) {
+    is_breakage_distance_customizable = FindValue(*node, "breakage_distance");
+  }
+
+  ReadParamOptional(max_route_distance_factor, params, "default.max_route_distance_factor");
+  CHECK_THROWS(max_route_distance_factor > 0.f,
+               POSITIVE_VALUE_MSG(max_route_distance_factor, "max_route_distance_factor"));
+
+  ReadParamOptional(max_route_time_factor, params, "default.max_route_time_factor");
+  CHECK_THROWS(max_route_time_factor > 0.f,
+               POSITIVE_VALUE_MSG(max_route_time_factor, "max_route_time_factor"));
+
+  ReadParamOptional(turn_penalty_factor, params, "default.turn_penalty_factor");
+  CHECK_THROWS(turn_penalty_factor >= 0.f,
+               NONNEGATIVE_VALUE_MSG(turn_penalty_factor, "turn_penalty_factor"));
+
+  if (const auto node = params.get_child_optional("customizable")) {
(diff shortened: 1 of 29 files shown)


src/meili/config.cc
```diff
@@ -0,0 +1,113 @@
...
+void Config::Read(const boost::property_tree::ptree& params) {
+  candidate_search.Read(params);
+  transition.Read(params);
+  emission.Read(params);
+  routing.Read(params);
+  customizable.Read(params);
+}
+
+void Config::CandidateSearch::Read(const boost::property_tree::ptree& params) {
+  ReadParamOptional(search_radius, params, "default.search_radius");
```
genadz, line 43: I'm not sure about this: should we treat parameters from json as optional or required ?

test/mapmatch_config.cc
```diff
@@ -0,0 +1,146 @@
+#include "baldr/rapidjson_utils.h"
+#include <boost/property_tree/ptree.hpp>
+
+#include "meili/config.h"
+
+#include "test.h"
+
+namespace {
+
+boost::property_tree::ptree json_to_pt(const std::string& json) {
```
mookerji, line 10: ACK! We have [12 places](https://github.com/valhalla/valhalla/search?q=%22ptree+json_to_pt%22&unscoped_q=%22ptree+json_to_pt%22) where this function is copied. We should add it to a header somewhere and not have 13 places. 

See also: https://github.com/valhalla/valhalla/pull/2481#discussion_r460284761

/cc @PureW

valhalla/meili/config.h
```diff
@@ -0,0 +1,81 @@
...
+    float max_route_distance_factor = 5.f;
+    // maximum allowed ratio of route time distance to the corresponding measurements time distance
+    float max_route_time_factor = 5.f;
+    // it's used to determine 'turn cost' depending on turn degree
+    float turn_penalty_factor = 200.f;
+
+    void Read(const boost::property_tree::ptree& params);
+  };
+
+  struct Emission {
```
mookerji, line 43: This struct specifically characterizes a likelihood function, so I'd call it `EmissionLikelihood`. Same for `TransitionLikelihood`.

valhalla/meili/config.h
```diff
@@ -0,0 +1,81 @@
...
+  // create config with default parameters
+  Config() = default;
+  // read parameters from boost::ptree (use default values for missing params)
+  explicit Config(const boost::property_tree::ptree& params);
+  // override default config parameters
+  void Read(const boost::property_tree::ptree& params);
+
+  struct CandidateSearch {
+    // default search radius (meters) for measurements; it's used to determine maximum search radius
+    float search_radius = 50.f;
```
mookerji, line 18: For these variables that have SI units, we should add a units suffix, like `search_radius_meters`.

src/meili/config.cc
```diff
@@ -0,0 +1,113 @@
+#include "meili/config.h"
+#include "midgard/logging.h"
+
+#define CHECK_POSITIVE_PARAM(value, name)                                                            \
```
mookerji, line 4: Macro is a great idea, but I think we can generalize this. My thoughts: 
- Add a `macro.h` header ... somewhere. I'm not sure what folder to put this in. 
- Generalize `CHECK_POSITIVE_PARAM` to a more general conditional check, like a `CHECK_THROW(<condition>, <message>)` macro. You could probably define something like:

```
#define CHECK_ARGUMENT_THROW(condition, exc_info)  \
  do {                                                                                           \
    if (!(condition)) {                                                                    \
      throw std::invalid_argument(exc_info);                             \
    }                                                                                              \
  } while (0)
```

I probably butchered the syntax here, but you get the idea.

valhalla/meili/config.h
```diff
@@ -0,0 +1,81 @@
...
+  struct Routing {
+    // interpolation distance (meters); threshold for merging close points together
+    float interpolation_distance = 10.f;
+
+    void Read(const boost::property_tree::ptree& params);
+  };
+
+  // define options that can be dynamically reassigned with user request
+  struct Customizable {
+    bool turn_penalty_factor = true;
```
mookerji, line 61: Would it make more sense for each struct here to have a `is_customizable=true|false` flag?

src/meili/config.cc
mookerji: I _think_ this it should be optional, with a sensible default configuration. My understanding though is that this is different from the current behavior, but should be fine because we are relaxing (vs strengthening) an API constraint (i.e., what was previously optional is now required).

src/meili/config.cc
```diff
@@ -0,0 +1,113 @@
...
+
+void Config::Customizable::Reset() {
+  turn_penalty_factor = false;
+  search_radius = false;
+  gps_accuracy = false;
+  breakage_distance = false;
+  interpolation_distance = false;
+}
+
+void Config::Customizable::Read(const boost::property_tree::ptree& params) {
```
mookerji, line 91: See comment below about possibly moving the flag into each struct. This property of each configuration should be colocated with each value its describing.

src/meili/map_matcher.cc
```diff
@@ -492,21 +492,21 @@ struct path_t {
 namespace valhalla {
 namespace meili {
 
-MapMatcher::MapMatcher(const boost::property_tree::ptree& config,
+MapMatcher::MapMatcher(const Config& config,
```
mookerji, line 495: I don't think the API compatibility should be an issue, I've also cc'd some downstream users of the mapmatching library in case they have opinions.

valhalla/meili/config.h
mookerji: It should probably be `TransitionCost`, etc. since that's existing terminology used in meili.

src/thor/worker.cc
```diff
@@ -293,8 +293,8 @@ void thor_worker_t::parse_measurements(const Api& request) {
 
   // we require locations
   try {
-    auto default_accuracy = matcher->config().get<float>("gps_accuracy");
-    auto default_radius = matcher->config().get<float>("search_radius");
+    auto default_accuracy = matcher->config().emission.gps_accuracy;
```
mookerji, line 296: Let's get rid of these `auto`'s when we see them. I think in this case, it makes the code to understand.

test/mapmatch_config.cc
genadz: good catch! will fix

valhalla/meili/config.h
genadz: I thought about it. We can't have one flag for the whole struct because the struct may have customizable and non-customizable fields the same time. I think, the best solution here to remove flags from `Customizable` struct and put to others that have customizable params (looks like below you suggested the same).

src/meili/config.cc
genadz: done. the main idea here was to avoid copying an exception message

src/thor/worker.cc
genadz: changed for some cases considering readability.

test/utils.h
```diff
@@ -0,0 +1,28 @@
...
+  rapidjson::read_json(ss, pt);
+  return pt;
+}
+
+inline rapidjson::Document to_document(const std::string& request) {
+  rapidjson::Document d;
+  auto& allocator = d.GetAllocator();
+  d.Parse(request.c_str());
+  if (d.HasParseError())
+    throw valhalla::valhalla_exception_t{100};
```
kevinkreiser, line 24: this seems a bit strange, can you mention why you would want to use this exception rather than just a generic one? these exceptions are to be used in request processing. i suspect this here is only used for test cases?

valhalla/macro.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <exception>
+
+#define CHECK_THROWS(condition, message)                                                             \
+  {                                                                                                  \
+    if (!(condition))                                                                                \
+      throw std::invalid_argument(message);                                                          \
```
kevinkreiser, line 8: does something catch this exception further up and convert it into a valhalla_exception_t ? because those are the only ones that get turned into proper messages that the user of the service will see.

test/utils.h
genadz: I've just moved this function from several cpps into header file in order to avoid copying

valhalla/macro.h
genadz: good question. I think it's out of responsibility of this code. `std::invalid_argument` is inherited from `std::exception`, so if we handle `std::exception` everything should be ok

mookerji:
This is looking really good, @genadz ! I'm out, but hopefully @kevinkreiser can help you get this merged.

test/utils.h
genadz: generally, this function it's not used anywhere. thinks, it's better to remove it at all

PR #2296, opened 2020-04-02
Fix Some ASAN Bugs

kevinkreiser:
So we were running with address sanitizer turned on and found that some unit tests barf (@mookerji pointed this out before of course). Anyway I've quickly fixed a few of them here and also found out that when you want to use gtest with libasan you need to link pthread or else you get this:

```
/usr/bin/ld: ../lib/libgtestd.a(gtest-all.cc.o): undefined reference to symbol 'pthread_key_delete@@GLIBC_2.2.5'
/lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO aus der Kommandozeile fehlt
collect2: error: ld returned 1 exit status
```

There are still some tests failing with ASAN:

- turnlanes
- narrative_dictionary
- python_valhalla
- node_valhalla

I'm not worried about the last two as both of them also spit out a warning like: 

    ==5781==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.

Which makes me think I dont care to fix that at this moment. I'll see if i can fix the other two although their ASAN output is very scant:

```
[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (285 ms total)
[  PASSED  ] 4 tests.

=================================================================
==3061==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 784 byte(s) in 14 object(s) allocated from:
    #0 0x7f853763af30 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdef30)
    #1 0x7f8535960d97 in argz_add_sep (/lib/x86_64-linux-gnu/libc.so.6+0xa0d97)

SUMMARY: AddressSanitizer: 784 byte(s) leaked in 14 allocation(s).
```

PR #2487, opened 2020-07-22
Sanitizer checks

SvetlanaBayarovich:
# Issue

1. Sanitizer options handling is refactored
2. UB sanitizer support is added
3. Address sanitizer is enabled in CI

## Tasklist

 - [ ] Add tests

## Requirements / Relations


CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ option(ENABLE_CCACHE "Speed up incremental rebuilds via ccache" ON)
 option(ENABLE_COVERAGE "Build with coverage instrumentalisation" OFF)
 option(ENABLE_COMPILER_WARNINGS "Build with compiler warnings" OFF)
 option(ENABLE_SANITIZER "Use memory sanitizer for Debug build" OFF)
```
SvetlanaBayarovich, line 29: I would rather rename this option if we choose to have separate flags for each sanitizer. Sth like 
```suggestion
option(ENABLE_ADDRESS_SANITIZER "Use memory sanitizer for Debug build" OFF)
```

mandeepsandhu: Thats a good suggestion. We can also have a top-level sanitizer flag that automatically enables all the sub-flags. So users can either switch on all flags sanitizer flags at once by enabling the top-level flag, or select specific ones they want.

.circleci/config.yml
```diff
@@ -164,6 +164,53 @@ jobs:
...
+      - run: node ./scripts/validate_changelog.js
+      - run: git submodule sync && git submodule update --init
+      - restore_cache:
+          keys:
+            - ccache-debug-linux-x86_64-{{ .Branch }}
+            - ccache-debug-linux-x86_64
+      - run: npm install --ignore-scripts
+      - run: mkdir build
+      - run: ls -la /usr/bin/
+      - run: clang-5.0 --version
```
mandeepsandhu, line 206: Is this needed in this workflow? I guess we're building with GCC here.

.circleci/config.yml
```diff
@@ -306,6 +353,8 @@ workflows:
     jobs:
       - build-base
       - lint-build-debug
+      - sanitizer-build-debug
+      - clang-build-debug
```
mandeepsandhu, line 357: Do we need to enable _both_ workflows here? We're also trying to reduce our CI build times (see https://github.com/mapbox/valhalla-internal/issues/763) so wanted to understand if both of these are mandatory.

CMakeLists.txt
mookerji: We also need to update the CMake build flags table in the top-level README.

cmake/SanitizerOptions.cmake
```diff
@@ -0,0 +1,13 @@
+if(ENABLE_SANITIZER)
```
mookerji, line 1: This should be `ENABLE_ADDRESS_SANITIZER`, per the suggestion earlier.

cmake/SanitizerOptions.cmake
```diff
@@ -0,0 +1,13 @@
...
+  message(STATUS "Enabling address sanitizer (ASan).")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fsanitize=address")
+endif()
+
+if(ENABLE_UB_SANITIZER)
```
mookerji, line 8: For consistent naming, this flag should be `ENABLE_UNDEFINED_SANITIZER`.

cmake/SanitizerOptions.cmake
```diff
@@ -0,0 +1,13 @@
+if(ENABLE_SANITIZER)
+  message(STATUS "Enabling address sanitizer (ASan).")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fsanitize=address")
```
mookerji, line 5: On setting the flag: we could use the other cmake module that's in this folder, which will check to see if the flag is supported by the compiler being used: https://github.com/valhalla/valhalla/blob/d89b886222ee5757655f5da62a337a589b51c0db/cmake/HandleLibcxxFlags.cmake#L231. This will make it easier to support the build across platforms.

.circleci/config.yml
```diff
@@ -164,6 +164,53 @@ jobs:
...
+      - run: node ./scripts/validate_changelog.js
+      - run: git submodule sync && git submodule update --init
+      - restore_cache:
+          keys:
+            - ccache-debug-linux-x86_64-{{ .Branch }}
+            - ccache-debug-linux-x86_64
+      - run: npm install --ignore-scripts
+      - run: mkdir build
+      - run: ls -la /usr/bin/
+      - run: clang-5.0 --version
```
mookerji, line 182: @SvetlanaBayarovich what version of Clang are you running? Why [Clang 5](https://releases.llvm.org/5.0.0/tools/clang/docs/ReleaseNotes.html) here vs something more recent? Is it the version of Clang supported by the CircleCI image for this workflow? I would be in-favor of turning the `build-lint-debug` CI task into one using [clang 10](https://releases.llvm.org/10.0.0/tools/clang/docs/index.html).

What I'm currently running: 

```
$ clang --version
clang version 10.0.0
Target: x86_64-apple-darwin19.5.0
Thread model: posix
InstalledDir: /usr/local/opt/llvm/bin
```

.circleci/config.yml
```diff
@@ -164,6 +164,53 @@ jobs:
       - run: make -C build coverage
       - run: /bin/bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"
 
+  clang-build-debug:
```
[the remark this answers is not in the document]
mookerji: Made a related comment below: since we run clang tools as part of the `build-lint-debug` job, I would be in favor of turning that task into one using clang 10 on Linux and not adding two additional jobs. The current compiler used there is gcc 5.4.0, which is from ~5 years ago.

What do you think @SvetlanaBayarovich ?

cmake/SanitizerOptions.cmake
```diff
@@ -0,0 +1,13 @@
...
+endif()
```
mookerji, line 6: Per the [ASan](https://clang.llvm.org/docs/AddressSanitizer.html), to make the ASAN stacktraces easier to interpret, we should also add two additional flags: `-fno-omit-frame-pointer -fno-optimize-sibling-calls`.

README.md
```diff
@@ -206,6 +206,8 @@ Also note that we run some `clang-tidy` linting over the code as well (see `.cla
 
 `scripts/clang-tidy-only-diff.sh` is run in CI and will the build if it detects any issues.
 
+Additionally a check with address sanitizer is run in CI. It is recommended to perform testing with sanitizer locally prior to commiting. In order to compile with sanitizer use flag `-DENABLE_SANITIZER=ON` during `cmake` confiuration step. As long as leak sanitizer (which is a part of address sanitizer) is not currently supported across different platforms it is disabled in the CI. You can disable it loclally with the environment variable `ASAN_OPTIONS=detect_leaks=0`.
```
mookerji, line 209: spelling: 
- confiuration => configuration
- loclally.=> locally

mookerji: We should also note that ASAN should be run with debug symbols (`-DCMAKE_BUILD_TYPE=Debug`)?

.circleci/config.yml
SvetlanaBayarovich: No, we certainly do not. Sorry, this PR became kind of mess while I was trying to figure out how to get CI working

.circleci/config.yml
SvetlanaBayarovich: The story behind this is the following. I was trying to understand how to make CI with address sanitizer. The problem was that it failed with the error 
```
[ 97%] Generating node_valhalla.log
==23592==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
npm ERR! Test failed.  See above for more details.
test/CMakeFiles/run-node_valhalla.dir/build.make:70: recipe for target 'test/node_valhalla.log' failed
```
My initial idea was that it was happening due to the lack of support on the side of `gcc`, that's why I was trying `clang`. As I found out by searching `build-base` logs, the clang version that is being installed is 5.0. However it looks like the compiler is not the reason of the issue.

.circleci/config.yml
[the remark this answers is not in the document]
SvetlanaBayarovich: I had troubles with CI. My initial idea was that it was happening due to the lack of support on the side of `gcc`, that's why I was trying `clang`. ([details](https://github.com/valhalla/valhalla/pull/2487#discussion_r460789889))

.circleci/config.yml
SvetlanaBayarovich: No, removed

.circleci/config.yml
```diff
@@ -164,6 +164,53 @@ jobs:
...
+      - run: npm install --ignore-scripts
+      - run: mkdir build
+      - run: ls -la /usr/bin/
+      - run: clang-5.0 --version
+      - run: |
+          # NOTE: -Werror disabled in CI, as we currently have >4k warnings.
+          cd build \
+            && cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_SANITIZER=ON -DBUILD_SHARED_LIBS=Off -DENABLE_NODE_BINDINGS=Off
+      - run: export ASAN_OPTIONS=detect_leaks=0 && make -C build -j2 tests check
+      # Note: we save the cache here before doing linting so that if linting fails, we can rebuild quickly
```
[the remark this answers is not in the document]
SvetlanaBayarovich: Yep, removed

.circleci/config.yml
```diff
@@ -164,6 +164,53 @@ jobs:
...
+      - run: npm install --ignore-scripts
+      - run: mkdir build
+      - run: ls -la /usr/bin/
+      - run: clang-5.0 --version
+      - run: |
+          # NOTE: -Werror disabled in CI, as we currently have >4k warnings.
+          cd build \
+            && cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_SANITIZER=ON -DBUILD_SHARED_LIBS=Off -DENABLE_NODE_BINDINGS=Off -DCMAKE_C_COMPILER=clang-5.0 -DCMAKE_CXX_COMPILER=clang++-5.0
+      - run: export ASAN_OPTIONS=detect_leaks=0 && make -C build -j2 tests check
+      # Note: we save the cache here before doing linting so that if linting fails, we can rebuild quickly
```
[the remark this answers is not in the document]
SvetlanaBayarovich: Yep, removed

.circleci/config.yml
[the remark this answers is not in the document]
PureW: > I would be in favor of turning that task into one using clang 10 on Linux and not adding two additional jobs. The current compiler used there is gcc 5.4.0, which is from ~5 years ago.

We still want to build on both gcc and clang as I understand it. Why not enable more checks and sanitizers on the Osx build where we use clang already?

As for the old gcc, I think the better solution would be to migrate away from Ubuntu 16 as the base image.

cmake/SanitizerOptions.cmake
SvetlanaBayarovich: Done

README.md
SvetlanaBayarovich: Done

cmake/SanitizerOptions.cmake
SvetlanaBayarovich: Done

cmake/SanitizerOptions.cmake
SvetlanaBayarovich: Done

cmake/SanitizerOptions.cmake
SvetlanaBayarovich: Do you suggest sth like 
```suggestion
  append_flags_if_supported(CMAKE_SHARED_LINKER_FLAGS -fsanitize=address)
```
? (I understand this is the most proper one here)
I generally like the approach but what disturbs me is that this function (and actually all the rest) produces no indication if the flag is not supported. I'd really rather have some warning in such case.
WDYT @mookerji? cc @PureW

mookerji: Oh, I agree that there should definitely be a warning produced by the functions/macros in that CMake file. I just tried adding one and got a warning: 

```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 0d7bb6d4a..8ade1b5c0 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -63,6 +63,7 @@ endif()

 function (cxx_add_warning_flags target)
   target_add_compile_flags_if_supported(${target} PRIVATE -Wall -Wextra)
+  target_add_compile_flags_if_supported(${target} PRIVATE -Wtotallyfakeflag)
   if (ENABLE_WERROR)
    target_add_compile_flags_if_supported(${target} PRIVATE -Werror)
   endif()
```

```
$ cmake -DENABLE_PYTHON_BINDINGS=OFF -DENABLE_NODE_BINDINGS=OFF -DCMAKE_BUILD_TYPE=Debug -DENABLE_SANITIZER=OFF -DENABLE_COMPILER_WARNINGS=ON -DENABLE_WERROR=OFF -DENABLE_COVERAGE=OFF ../
-- Configuring in debug mode
-- Using ccache to speed up incremental builds
-- Using pbf headers from /usr/local/include
-- Using pbf libs from /usr/local/lib/libprotobuf.dylib
-- Using pbf release libs from
-- Using pbf debug libs from /usr/local/lib/libprotobuf.dylib
-- Using pbf-lite
-- Found SQLite3: /usr/local/opt/sqlite3/lib/libsqlite3.dylib
-- Enabling compiler warning flags
-- Performing Test LIBCXX_SUPPORTS_WALL_FLAG
-- Performing Test LIBCXX_SUPPORTS_WALL_FLAG - Success
-- Performing Test LIBCXX_SUPPORTS_WEXTRA_FLAG
-- Performing Test LIBCXX_SUPPORTS_WEXTRA_FLAG - Success
-- Performing Test LIBCXX_SUPPORTS_WTOTALLYFAKEFLAG_FLAG
-- Performing Test LIBCXX_SUPPORTS_WTOTALLYFAKEFLAG_FLAG - Failed
```

.circleci/config.yml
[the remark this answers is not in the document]
mookerji: Well, I suggested using that particular task (`build-lint-debug`) since it was already a central point of using Clang tools for linting/debugging/etc. The other tasks are still using gcc. I'm also fine with enabling some of these checks in OSX as well. Upgrading Ubuntu is fine, but we should use a fixed, explicit, recent compiler version.

Regarding sanitizer support, we should prioritize getting sanitizer support functional on LLVM (vs their ports to GCC).

CMakeLists.txt
SvetlanaBayarovich: Hmmm, one flag to rule them all... Done 🙂

.circleci/config.yml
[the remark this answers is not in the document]
SvetlanaBayarovich: It turned out that the issues I was having were related only to ASan+python/node bindings interaction (order of library loading). The current solution I suggest is to disable the bindings in the lint job and add sanitizers to it.

The bindings are still build e.g. in `build-release` job. This job has `-DENABLE_PYTHON_BINDINGS=On` and node bindings are `On` by default according to our CMakeLists. I can also add `-DENABLE_NODE_BINDINGS=On` to make it more explicit. WDYT?

cmake/SanitizerOptions.cmake
SvetlanaBayarovich: Wow, didn't know the check function also prints something.
Implemented this part according to the suggested approach, hope I got it correctly. PTAL

cmake/SanitizerOptions.cmake
```diff
@@ -0,0 +1,34 @@
...
+# Inlclude build macros for updating configuration variables
+include(HandleLibcxxFlags)
+
+set (SANITIZER_FLAGS_LIST "")
+set (SANITIZER_EXE_LINKER_FLAGS_LIST "")
+set (SANITIZER_SHARED_LINKER_FLAGS_LIST "")
+
+if(ENABLE_ADDRESS_SANITIZER)
+  message(STATUS "Enabling address sanitizer (ASan).")
+  set(CMAKE_REQUIRED_FLAGS "-fsanitize=address")
```
SvetlanaBayarovich, line 15: Need this trick for `-fsanitize=address`. Without it the flag was not found for some reason. The idea is from [here](https://github.com/arsenm/sanitizers-cmake/blob/master/cmake/sanitize-helpers.cmake#L120-L121)

Interesting that `-fsanitize=undefined` is ok without this trick

.circleci/config.yml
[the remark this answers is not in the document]
SvetlanaBayarovich: UPD ^^ as long as node bindings are removed the point relates only to python bindings

SvetlanaBayarovich:
I think everything is mostly clarified here, looking forward to review

cmake/SanitizerOptions.cmake
PureW: Doesn't this overwrite any previous value of `CMAKE_REQUIRED_FLAGS`?

SvetlanaBayarovich: That's a good point. I can't find any sensible doc regarding this flag, only came across a note saying "CMAKE_REQUIRED_FLAGS is explicitly designed for temporary flags used in CMake's various check_<something> macros". **But** in order to be on the safe side I'll better save old values and restore it afterwards

SvetlanaBayarovich: Done

.circleci/config.yml
```diff
@@ -98,11 +98,12 @@ jobs:
           # NOTE: -Werror disabled in CI, as we currently have >4k warnings.
           cd build \
             && cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=On -DCPACK_GENERATOR=DEB \
-                -DENABLE_COMPILER_WARNINGS=On -DENABLE_WERROR=Off -DCMAKE_EXPORT_COMPILE_COMMANDS=On
+                -DENABLE_COMPILER_WARNINGS=On -DENABLE_WERROR=Off -DCMAKE_EXPORT_COMPILE_COMMANDS=On \
+                -DENABLE_ADDRESS_SANITIZER=ON -DENABLE_PYTHON_BINDINGS=Off
```
kevinkreiser, line 102: @SvetlanaBayarovich is it true that the build fails with `ENABLE_UNDEFINED_SANITIZER=On`? because of https://github.com/valhalla/valhalla/pull/2296#issuecomment-608207370

kevinkreiser: and i guess address sanitizer finds something in the python bindings as well?

SvetlanaBayarovich: @kevinkreiser The build fails with `leak sanitizer` (which is a part of ASan) enabled with what looks very similar to the comment you shared. And that is quite a curious finding of yours 🤔 I planned to investigate it separately and enable "pure" ASan on CI for now. This leak snaitizer is not well supported on all the platforms and is actually unavailable for my Apple's `clang` which is really irritating

SvetlanaBayarovich: As for python bindings, the problem is entirely different - it is about link order of libraries:
```
==22382==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
test/CMakeFiles/run-python_valhalla.dir/build.make:62: recipe for target 'test/python_valhalla.log' failed
```
Does it actually make sense to test python bindings?

.circleci/config.yml
```diff
@@ -98,11 +98,12 @@ jobs:
...
       - run: make -C build -j4
       - run: make -C build utrecht_tiles
       - run: make -C build -j2 tests
-      - run: make -C build -j2 check
+      - run: export ASAN_OPTIONS=detect_leaks=0 && make -C build -j2 check
```
mandeepsandhu, line 106: Can we have a short comment here on why the leak sanitizer is disabled? I know you have a comment in the README, but sometimes one might forget to look there, so might be easier to readers of the code to see an inline comment. Nothing elaborate, just a short one saying the leak sanitizer is disabled since its not supported on all platforms.

BTW, does enabling leak sanitizer cause the python bindings to fail to buid?

cmake/SanitizerOptions.cmake
```diff
@@ -0,0 +1,36 @@
...
+include(HandleLibcxxFlags)
+
+set (SANITIZER_FLAGS_LIST "")
+set (SANITIZER_EXE_LINKER_FLAGS_LIST "")
+set (SANITIZER_SHARED_LINKER_FLAGS_LIST "")
+
+if(ENABLE_ADDRESS_SANITIZER)
+  message(STATUS "Enabling address sanitizer (ASan).")
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "-fsanitize=address")
```
mandeepsandhu, line 16: You can also "append" to the flags variable with:
```
set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -fsanitize=address")
```
I haven't tried it, but I _think_ it should work.

.circleci/config.yml
SvetlanaBayarovich: Enabling leak sanitizer leads to errors that look very similar to what @kevinkreiser described in  [#2296 (comment)](https://github.com/valhalla/valhalla/pull/2296#issuecomment-608207370).

Python bindings to fail to build because of ASan itself. ASan requires some specific library load order which causes difficulties in case of python bindings.

cmake/SanitizerOptions.cmake
SvetlanaBayarovich: It does work. It also looks safer to me so I'll use it, thanks

.circleci/config.yml
kevinkreiser: it does make sense to test the python bindings. many collaborators are making use of them. i suppose that its ok to remove from debug so long as we dont remove from the other builds

PR #2488, opened 2020-07-23
Fix dereferencing of end for std::lower_bound and possible UB

LaGrunge:
Fixes UB and segfault when std::lower_bound returns end.


gknisely:
@LaGrunge Can you update the changelog please?

LaGrunge:
Added mention in Changelog

gknisely:
@LaGrunge Thank you!

PR #2492, opened 2020-07-24
Add build-tiles benchmark

merkispavel:
# Issue

Benchmarks are supposed to help to track the tile building performance

## Tasklist

 - [ ] Add tests

## Requirements / Relations


merkispavel:
Here's how benchmark report looks at the moment
```
------------------------------------------------------------
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
sydney.osm.pbf          18.6 ms         18.4 ms           38
stockholm.osm.pbf       16.6 ms         16.3 ms           43
moscow.osm.pbf          74.8 ms         73.3 ms            9
monaco.osm.pbf          98.6 ms         98.0 ms            7
london.osm.pbf           210 ms          209 ms            3
munchen.osm.pbf          186 ms          185 ms            4
porto.osm.pbf            233 ms          232 ms            3
andorra.osm.pbf          260 ms          258 ms            3
```

bench/mjolnir/config.json
```diff
@@ -0,0 +1,256 @@
+{
```
merkispavel, line 1: have doubts about committing the whole config file

danpaz: Just an idea: could the config file be generated at runtime using `scripts/valhalla_build_config`?

bench/mjolnir/buildtiles.cc
```diff
@@ -0,0 +1,55 @@
...
+
+#include "baldr/rapidjson_utils.h"
+#include "filesystem.h"
+#include "mjolnir/util.h"
+
+using namespace valhalla::mjolnir;
+
+namespace {
+
+void BM_SomeFunction(benchmark::State& state, const std::string& pbf_path) {
```
mandeepsandhu, line 11: This can be named something more appropriate :) Like `BM_BuildTiles` or some such.

bench/mjolnir/buildtiles.cc
```diff
@@ -0,0 +1,55 @@
...
+  uintmax_t size;
+
+  bool operator<(const FileWithSize& file) const {
+    return size < file.size;
+  }
+};
+
+std::vector<FileWithSize> CollectInputProtobufs(const std::string& path) {
+  std::vector<FileWithSize> files;
+  for (filesystem::recursive_directory_iterator i(path), end; i != end; ++i)
```
mandeepsandhu, line 39: Style nit: Can we add braces around single-line `for` loops (like you've done above)? IMO its less error prone when adding more lines to the loop body.

bench/mjolnir/buildtiles.cc
```diff
@@ -0,0 +1,55 @@
...
+void BM_SomeFunction(benchmark::State& state, const std::string& pbf_path) {
+  valhalla::midgard::logging::Configure({{"type", ""}});
+
+  std::string const kConfigFilePath = VALHALLA_SOURCE_DIR "bench/mjolnir/config.json";
+  boost::property_tree::ptree config;
+  rapidjson::read_json(kConfigFilePath, config);
+  config.get_child("mjolnir").erase("tile_extract");
+  config.get_child("mjolnir").erase("tile_url");
+
+  BuildStage const start_stage_str = string_to_buildstage("initialize");
```
mandeepsandhu, line 20: Style nit: some place we have the `const` qualifier before the type and at other places its after (eg in function args & for loop). Maybe have the form "`const` <Type> var" everywhere? (I think thats how Valhalla codebase uses it in most places).

bench/mjolnir/buildtiles.cc
merkispavel: Ooops, was opypasted from the tutorials I guess. Thx :)

bench/mjolnir/config.json
mookerji: I think it's not a big deal to commit this file. We could generate it at build/CMake time. In the past, I've also tried leaving out parts of the configuration that I thought were unused, but the configuration reader in those cases will throw an exception or segfault.

One alternative, if we're worried about configuration sprawl (not a historical concern), would be to have a shared default, configuration that could be used by some of the benchmarks. The `meili` benchmark config file actually has nothing special about it.

bench/mjolnir/buildtiles.cc
```diff
@@ -0,0 +1,55 @@
...
+  bool operator<(const FileWithSize& file) const {
+    return size < file.size;
+  }
+};
+
+std::vector<FileWithSize> CollectInputProtobufs(const std::string& path) {
+  std::vector<FileWithSize> files;
+  for (filesystem::recursive_directory_iterator i(path), end; i != end; ++i)
+    files.push_back({i->path(), i->file_size()});
+  std::sort(files.begin(), files.end());
```
mookerji, line 41: Alternatively, we could use an explicit manifest of the PBF files. There's a similar example here for benchmark fixtures: https://github.com/valhalla/valhalla/blob/master/bench/meili/mapmatch.cc#L97

bench/mjolnir/config.json
merkispavel: cmake-idea sounds good to me

bench/mjolnir/buildtiles.cc
```diff
@@ -0,0 +1,41 @@
...
+
+#include "baldr/rapidjson_utils.h"
+#include "mjolnir/util.h"
+
+using namespace valhalla::mjolnir;
+
+int main(int argc, char** argv) {
+  valhalla::midgard::logging::Configure({{"type", ""}});
+
+  const auto build_tiles = [](benchmark::State& state, const std::string& pbf_path,
```
merkispavel, line 11: `benchmark::RegisterBenchmark` accepts Lambda type as an argument, so it's not possible to make it work with  function

bench/mjolnir/buildtiles.cc
```diff
@@ -0,0 +1,41 @@
...
+    for (auto _ : state) {
+      state.PauseTiming();
+      build_tile_set(config, {pbf_path}, BuildStage::kInitialize,
+                     static_cast<BuildStage>(static_cast<uint8_t>(stage) - 1));
+      state.ResumeTiming();
+      benchmark::DoNotOptimize(build_tile_set(config, {pbf_path}, stage, stage));
+    }
+  };
+
+  const std::string kProtobufs[] = {"stockholm.osm.pbf"};
```
merkispavel, line 28: Turned out that end-to-end building takes quite a lot time so bench runs only 1 iteration even with small pbfs(like this one)

merkispavel: I don't mind merging such benchmarks to the master. However, I have doubts they'll be useful enough. A few thoughts of mine
1) Building tiles from small pbfs & large pbfs give 2 different times. E.g a stage that takes quite a lot time with small input data might work pretty fast on large amount of data. So we still need to test our changes/feature branch on larger input
2) My intuition tells me that benchmarks should measure fast functions that may occur in "hot spots"(e.g there're a lot of benchmarks of `insert`/`erase` stl methods). On the other hand, our stage-functions are called once during tile-building. What might work here is to add hot-functions-benchmark later: need to collect perf/gperftools statistics with large pbf and choose functions/methods with the most spent time(%)
3) The same code measurements might deviate by +- 5-10% sometimes(somewhy) so you might not even notice speed up

merkispavel:
https://github.com/valhalla/valhalla/pull/2492#discussion_r462759295

bench/mjolnir/buildtiles.cc
PureW: It's possible to tell gbench to run more iterations:
```
benchmark [--benchmark_list_tests={true|false}]
          [--benchmark_filter=<regex>]
          [--benchmark_min_time=<min_time>]
          [--benchmark_repetitions=<num_repetitions>]
          [--benchmark_report_aggregates_only={true|false}]
          [--benchmark_display_aggregates_only={true|false}]
          [--benchmark_format=<console|json|csv>]
          [--benchmark_out=<filename>]
          [--benchmark_out_format=<json|console|csv>]
          [--benchmark_color={auto|true|false}]
          [--benchmark_counters_tabular={true|false}]
          [--v=<verbosity>]
```

PureW: Is it possible to choose a smaller pbf to build tiles from in order to speed it up as well?

merkispavel:
Getting weird error in `build-release-x86`, can't reproduce it locally(tried os x and ubuntu 16.04). Will continue to debug it via ssh, I have no clue 😕 
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```

PureW:
> Getting weird error in `build-release-x86`, can't reproduce it locally(tried os x and ubuntu 16.04).

We just upgraded x86 to ubuntu 18.04, you may want to use `valhalla/valhalla:build-x86-latest` image in case you're not.

bench/mjolnir/buildtiles.cc
```diff
@@ -0,0 +1,100 @@
...
+                   static_cast<BuildStage>(static_cast<uint8_t>(stage) - 1), false);
+
+    for (auto _ : state) {
+      state.PauseTiming();
+      // copy the files needed for the stage
+      CopyDir(kStageDir, kTileDir);
+      const boost::property_tree::ptree config = GetConfig(kTileDir);
+      state.ResumeTiming();
+
+      benchmark::DoNotOptimize(build_tile_set(config, {pbf_path}, stage, stage, false));
```
merkispavel, line 64: tldr: need to disable `pbf_release_memory` parameter to make it work
Oh god, spent a few days.. The only job(release-x86) used to fail before I disabled `pbf_memory_release`(last argument). This parameter is responsible for calling(or not calling) `protobuf::ShutdownProtobufLibrary()`. However, the docs say that it just delete all allocated objects... Weird :/

bench/mjolnir/buildtiles.cc
```diff
@@ -0,0 +1,100 @@
...
+
+void RegisterByStageBenchmark() {
+  const auto build_tiles = [](benchmark::State& state, const std::string& pbf_path,
+                              BuildStage stage) {
+    const std::string kStageDir = "bench/mjolnir/stage_files";
+    const std::string kTileDir = "bench/mjolnir/tiles";
+    // prepare files for the stage once and copy them every time before running the stage
+    const boost::property_tree::ptree stage_config = GetConfig(kStageDir);
+    build_tile_set(stage_config, {pbf_path}, BuildStage::kInitialize,
+                   static_cast<BuildStage>(static_cast<uint8_t>(stage) - 1), false);
```
merkispavel, line 55: There are 2 options that I see:
- run this "preparation" command before  every "build-stage" call
- run it once, save all files and copy them before every "build-stage" call

the second one turned out to be faster

bench/mjolnir/buildtiles.cc
merkispavel: > It's possible to tell gbench to run more iterations, using `--benchmark_min_time` I think
> 
> ```
> benchmark [--benchmark_list_tests={true|false}]
>           [--benchmark_filter=<regex>]
>           [--benchmark_min_time=<min_time>]
>           [--benchmark_repetitions=<num_repetitions>]
>           [--benchmark_report_aggregates_only={true|false}]
>           [--benchmark_display_aggregates_only={true|false}]
>           [--benchmark_format=<console|json|csv>]
>           [--benchmark_out=<filename>]
>           [--benchmark_out_format=<json|console|csv>]
>           [--benchmark_color={auto|true|false}]
>           [--benchmark_counters_tabular={true|false}]
>           [--v=<verbosity>]
> ```

I found `Iteration` method to set fixed number of iterations. Looks ok from the first sight

merkispavel: > Is it possible to choose a smaller pbf to build tiles from in order to speed it up as well?

Turned out `shortcuts` stage takes significant amount of time even with small pbf(300ms -> 30ms when shortcuts-opiton is disabled). Though with enabled shortucts-option it takes 300ms for the end-to-end build. Looks acceptable for the benchmark

PR #2493, opened 2020-07-24
fix: clang-analyzer-core.CallAndMessage warnings

keygang:
It's very dangerous to forget to do checks pointers for nullptr, so this PR fix clang-analyzer-core.CallAndMessage warnings.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/loki/reach.cc
+++ b/src/loki/reach.cc
@@ -172,6 +172,9 @@ directed_reach Reach::exact(const valhalla::baldr::DirectedEdge* edge,
   // fake up the input location
   const baldr::GraphTile* tile = nullptr;
   const auto* node = reader.GetEndNode(edge, tile);
+  if (node == nullptr) {
+    return reach;
+  }
   auto ll = node->latlng(tile->header()->base_ll());
   locations_.Mutable(0)->mutable_ll()->set_lng(ll.first);
   locations_.Mutable(0)->mutable_ll()->set_lat(ll.second);
--- a/src/thor/astar.cc
+++ b/src/thor/astar.cc
@@ -228,7 +228,7 @@ void AStarPathAlgorithm::ExpandForward(GraphReader& graphreader,
 
   // Handle transitions - expand from the end node of each transition
   if (!from_transition && nodeinfo->transition_count() > 0) {
-    const NodeTransition* trans = tile->transition(nodeinfo->transition_index());
+    const NodeTransition* trans = tile->transition(nodeinfo->transition_index()); // NOLINT
     for (uint32_t i = 0; i < nodeinfo->transition_count(); ++i, ++trans) {
       if (trans->up()) {
         hierarchy_limits_[node.level()].up_transition_count++;
--- a/src/thor/costmatrix.cc
+++ b/src/thor/costmatrix.cc
@@ -357,6 +357,9 @@ void CostMatrix::ForwardSearch(const uint32_t index, const uint32_t n, GraphRead
 
     // Handle transitions - expand from the end node of the transition
     if (!from_transition && nodeinfo->transition_count() > 0) {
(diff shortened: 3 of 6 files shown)


src/loki/reach.cc
```diff
@@ -172,6 +172,9 @@ directed_reach Reach::exact(const valhalla::baldr::DirectedEdge* edge,
   // fake up the input location
   const baldr::GraphTile* tile = nullptr;
   const auto* node = reader.GetEndNode(edge, tile);
+  if (node == nullptr) {
```
kevinkreiser, line 175: good catch

src/thor/astar.cc
```diff
@@ -228,7 +228,7 @@ void AStarPathAlgorithm::ExpandForward(GraphReader& graphreader,
 
   // Handle transitions - expand from the end node of each transition
   if (!from_transition && nodeinfo->transition_count() > 0) {
-    const NodeTransition* trans = tile->transition(nodeinfo->transition_index());
+    const NodeTransition* trans = tile->transition(nodeinfo->transition_index()); // NOLINT
```
kevinkreiser, line 231: this seems superfluous? can you please revert

src/thor/costmatrix.cc
```diff
@@ -357,6 +357,9 @@ void CostMatrix::ForwardSearch(const uint32_t index, const uint32_t n, GraphRead
 
     // Handle transitions - expand from the end node of the transition
     if (!from_transition && nodeinfo->transition_count() > 0) {
+      if (tile == nullptr) {
```
kevinkreiser, line 360: this is not needed. you'll notice that before each call to expand, there is an if that protects from calling it with a null tile

src/thor/multimodal.cc
```diff
@@ -526,6 +526,9 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
 
   // Handle transitions - expand from the end node each transition
   if (!from_transition && nodeinfo->transition_count() > 0) {
+    if (tile == nullptr) {
```
kevinkreiser, line 529: this is also not needed, the top of the function checks this condition already to make sure tile is not null

src/thor/route_action.cc
```diff
@@ -416,6 +416,9 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
         // Form output information based on path edges
         if (api.trip().routes_size() == 0 || api.options().alternates() > 0)
           route = api.mutable_trip()->mutable_routes()->Add();
+        if (route == nullptr) {
```
kevinkreiser, line 419: this is pedantic. the only way this can happen is if you run your system out of ram. if that happens you have way worse problems on your  hands. id prefer we not add this check

src/thor/route_action.cc
```diff
@@ -424,7 +427,9 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
       }
     }
   }
-
+  if (route == nullptr) {
```
kevinkreiser, line 430: same

src/thor/route_action.cc
```diff
@@ -504,6 +509,9 @@ void thor_worker_t::path_depart_at(Api& api, const std::string& costing) {
         // Form output information based on path edges. vias are a route discontinuity map
         if (api.trip().routes_size() == 0 || api.options().alternates() > 0)
           route = api.mutable_trip()->mutable_routes()->Add();
+        if (route == nullptr) {
```
kevinkreiser, line 512: same

mskurydin:
Moved a minimalistic change to the PR here: https://github.com/valhalla/valhalla/pull/2504

PR #2497, opened 2020-07-28
nit: Use 4 bytes and header guard for TRAFFIC_TILE_VERSION

PureW:


PureW:
Closing due to https://github.com/valhalla/valhalla/pull/2503

Issue #2494, opened 2020-07-24
valhalla::baldr::NodeInfo::heading - undefined behavior


jaegs:
When running Mjolnir with a sanitizer it crashes: 
```
runtime error: shift exponent 1304 is too large for 64-bit type 'uint64_t' (aka 'unsigned long')
valhalla/src/baldr/nodeinfo.cc:222:49: runtime error: shift exponent 1304 is too large for 64-bit type 'uint64_t' (aka 'unsigned long')
    #0 0x7eff35f30d5f in valhalla::baldr::NodeInfo::heading(unsigned int) const third-party/valhalla/src/baldr/nodeinfo.cc:222
    #1 0x7eff361a4c2c in (anonymous namespace)::GetTurnTypes(valhalla::baldr::DirectedEdge, unsigned int, std::set<valhalla::baldr::Turn::Type, std::less<valhalla::baldr::Turn::Type>, std::allocator<valhalla::baldr::Turn::Type> >&, valhalla::baldr::NodeInfo&, valhalla::mjolnir::GraphTileBuilder&, valhalla::baldr::GraphReader&, std::mutex&) third-party/valhalla/src/mjolnir/graphenhancer.cc:210
    #2 0x7eff3616f81d in (anonymous namespace)::UpdateTurnLanes(valhalla::mjolnir::OSMData const&, unsigned int, bool, valhalla::baldr::DirectedEdge&, valhalla::baldr::NodeInfo&, valhalla::mjolnir::GraphTileBuilder&, valhalla::baldr::GraphReader&, std::mutex&, std::vector<valhalla::baldr::TurnLanes, std::allocator<valhalla::baldr::TurnLanes> >&) third-party/valhalla/src/mjolnir/graphenhancer.cc:415
[...]

PR #2498, opened 2020-07-28
Ubsan findings fixes

SvetlanaBayarovich:
# Issue

Resolves findings of the Undefined behavior sanitizer.

This PR is to changes from PR in my fork (https://github.com/SvetlanaBayarovich/valhalla/pull/1)
Changes are the same only the one regarding directededge.cc file is dropped as it is already merged to master

## Tasklist

 - [ ] Add tests

## Requirements / Relations


SvetlanaBayarovich:
An update regarding performance. On CI benchmark results on `build-release` job are
for `master` ([link](https://app.circleci.com/pipelines/github/valhalla/valhalla/3183/workflows/5ccd2513-c24a-47f2-8772-824afa1c6514/jobs/26879))
```
-----------------------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------
BM_UtrechtCostMatrix/1        0.114 ms        0.114 ms         6026 Routes=8.76286k/s
BM_UtrechtCostMatrix/2         3.44 ms         3.44 ms          216 Routes=582.236/s
BM_UtrechtCostMatrix/4         32.5 ms         32.4 ms           22 Routes=123.34/s
BM_UtrechtCostMatrix/8         75.0 ms         74.9 ms            9 Routes=106.757/s
BM_UtrechtCostMatrix/16         182 ms          182 ms            4 Routes=87.9684/s
BM_UtrechtCostMatrix/32         395 ms          395 ms            2 Routes=81.1087/s
BM_UtrechtCostMatrix/64         992 ms          991 ms            1 Routes=64.5778/s
BM_UtrechtCostMatrix/128       2627 ms         2620 ms            1 Routes=48.8553/s
BM_UtrechtCostMatrix/256       8499 ms         8481 ms            1 Routes=30.1863/s

-----------------------------------------------------------------------------------
Benchmark                                         Time             CPU   Iterations
-----------------------------------------------------------------------------------
OfflineMapmatchFixture/BasicOfflineMatch     520011 ns       517812 ns         1204
BM_ManyCases/0                              2533241 ns      2525737 ns          301
BM_ManyCases/1                              3206183 ns      3206149 ns          216
BM_ManyCases/2                              3386477 ns      3380603 ns          203
BM_ManyCases/3                              7969016 ns      7957341 ns           79
```
for `ub-fixes` ([link](https://app.circleci.com/pipelines/github/valhalla/valhalla/3191/workflows/2a1cbc85-beb2-4da9-abf6-b16ae5a261a4/jobs/26921))
```
-----------------------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------
BM_UtrechtCostMatrix/1        0.115 ms        0.114 ms         6404 Routes=8.75265k/s
BM_UtrechtCostMatrix/2         2.69 ms         2.69 ms          256 Routes=743.667/s
BM_UtrechtCostMatrix/4         23.1 ms         23.1 ms           30 Routes=173.103/s
BM_UtrechtCostMatrix/8         58.8 ms         58.8 ms           12 Routes=136.057/s
BM_UtrechtCostMatrix/16         135 ms          135 ms            5 Routes=118.086/s
BM_UtrechtCostMatrix/32         292 ms          292 ms            2 Routes=109.459/s
BM_UtrechtCostMatrix/64         978 ms          978 ms            1 Routes=65.4333/s
BM_UtrechtCostMatrix/128       2721 ms         2721 ms            1 Routes=47.049/s
BM_UtrechtCostMatrix/256       7988 ms         7988 ms            1 Routes=32.0499/s

-----------------------------------------------------------------------------------
Benchmark                                         Time             CPU   Iterations
-----------------------------------------------------------------------------------
OfflineMapmatchFixture/BasicOfflineMatch     587367 ns       587355 ns         1255
BM_ManyCases/0                              1921540 ns      1921491 ns          294
BM_ManyCases/1                              2217112 ns      2217072 ns          335
BM_ManyCases/2                              2549283 ns      2549260 ns          259
BM_ManyCases/3                              7325829 ns      7325599 ns           92
```

SvetlanaBayarovich:
While testing locally I see the following numbers
for `master`
```
-----------------------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------
BM_UtrechtCostMatrix/1        0.036 ms        0.036 ms        19888 Routes=27.9032k/s
BM_UtrechtCostMatrix/2         1.59 ms         1.59 ms          422 Routes=1.25584k/s
BM_UtrechtCostMatrix/4         19.8 ms         19.7 ms           31 Routes=202.588/s
BM_UtrechtCostMatrix/8         51.5 ms         51.5 ms           10 Routes=155.326/s
BM_UtrechtCostMatrix/16         127 ms          127 ms            5 Routes=125.528/s
BM_UtrechtCostMatrix/32         294 ms          294 ms            2 Routes=108.929/s
BM_UtrechtCostMatrix/64         758 ms          758 ms            1 Routes=84.415/s
BM_UtrechtCostMatrix/128       1813 ms         1813 ms            1 Routes=70.6176/s
BM_UtrechtCostMatrix/256       5632 ms         5631 ms            1 Routes=45.4632/s

-----------------------------------------------------------------------------------
Benchmark                                         Time             CPU   Iterations
-----------------------------------------------------------------------------------
OfflineMapmatchFixture/BasicOfflineMatch     597513 ns       597417 ns          866
BM_ManyCases/0                              1830187 ns      1829279 ns          319
BM_ManyCases/1                              2251921 ns      2251327 ns          281
BM_ManyCases/2                              2637502 ns      2636133 ns          263
BM_ManyCases/3                              9742034 ns      9737754 ns           57
```
for `ub-fixes`
```
-----------------------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------
BM_UtrechtCostMatrix/1        0.035 ms        0.035 ms        20040 Routes=28.6054k/s
BM_UtrechtCostMatrix/2         1.62 ms         1.62 ms          424 Routes=1.23668k/s
BM_UtrechtCostMatrix/4         20.0 ms         20.0 ms           32 Routes=200.248/s
BM_UtrechtCostMatrix/8         55.9 ms         55.9 ms           12 Routes=143.227/s
BM_UtrechtCostMatrix/16         127 ms          127 ms            5 Routes=125.924/s
BM_UtrechtCostMatrix/32         305 ms          305 ms            2 Routes=105.028/s
BM_UtrechtCostMatrix/64         747 ms          747 ms            1 Routes=85.7047/s
BM_UtrechtCostMatrix/128       1805 ms         1805 ms            1 Routes=70.9263/s
BM_UtrechtCostMatrix/256       5664 ms         5662 ms            1 Routes=45.211/s

-----------------------------------------------------------------------------------
Benchmark                                         Time             CPU   Iterations
-----------------------------------------------------------------------------------
OfflineMapmatchFixture/BasicOfflineMatch     605216 ns       605045 ns          893
BM_ManyCases/0                              1816144 ns      1815517 ns          329
BM_ManyCases/1                              2269957 ns      2268949 ns          295
BM_ManyCases/2                              2560568 ns      2559744 ns          262
BM_ManyCases/3                              9740998 ns      9736263 ns           57
```

src/baldr/nodeinfo.cc
```diff
@@ -221,10 +221,15 @@ void NodeInfo::set_stop_index(const uint32_t stop_index) {
 // Get the heading of the local edge given its local index. Supports
 // up to 8 local edges. Headings are expanded from 8 bits.
 uint32_t NodeInfo::heading(const uint32_t localidx) const {
-  // Make sure everything is 64 bit!
-  uint64_t shift = localidx * 8; // 8 bits per index
-  return static_cast<uint32_t>(std::round(
-      ((headings_ & (static_cast<uint64_t>(255) << shift)) >> shift) * kHeadingExpandFactor));
+  if (localidx > kMaxLocalEdgeIndex) {
```
kevinkreiser, line 224: can this actually happen? this should be limited by `nodeinfo::local_edge_count_` which is 3 bits

SvetlanaBayarovich: It does happen in the unit tests in my experience and others also ran into this issue (I don't know how exactly). See the discussion from the [initial PR](https://github.com/SvetlanaBayarovich/valhalla/pull/1#discussion_r459616364)

gknisely: @kevinkreiser https://github.com/valhalla/valhalla/issues/2494

PureW:
I measured no difference between 0f41d17 and master when running 23886 routes.

_originally posted at https://github.com/SvetlanaBayarovich/valhalla/pull/1#issuecomment-665803447_

src/baldr/nodeinfo.cc
kevinkreiser: The point is it shouldn't happen and indicates another bug upstream, right? Instead of papering over it we should follow the call stack and figure our why bogus values are being sent here and fix that

SvetlanaBayarovich: Oh, I assumed it might be an expected behavior as long as there was such check in `set_heading`.  I'll try to look into it considering your remark regarding `local_edge_count_`

gknisely: @kevinkreiser yes.  I was questioning the same on the initial PR.  @dnesbitt61 any idea on why we would send in high values here or is it something we need to dig into

dnesbitt61: May need to dig into this more. As I have mentioned, NodeInfo::heading is a field that could/should be deprecated in the future. Its only use outside of data creation is within map matching to find a "turn cost" based on heading changes. This should really use some form of the costing object or use the directed edge turn_type information. In meili the heading method is protected from using a local_index > 7. 
I think the other uses of heading (GraphEnhancer and ShortcutBuilder) could use local storage of headings rather than actually storing them in the tile. Here is where some high values perhaps could be sent in - I haven't tracked this down recently.

SvetlanaBayarovich: The first time I encounter this issue is during ` "Building Utrecht Tiles..."` custom command, namely on 
```
  COMMAND ${CMAKE_BINARY_DIR}/valhalla_build_tiles
      --inline-config '{"mjolnir":{"id_table_size":1000,"tile_dir":"test/data/utrecht_tiles","timezone":"test/data/tz.sqlite","admin":"${VALHALLA_SOURCE_DIR}/test/data/netherlands_admin.sqlite","hierarchy":true,"shortcuts":true,"concurrency":1,"logging":{"type":""}}}'
      -s build -e cleanup
      ${VALHALLA_SOURCE_DIR}/test/data/utrecht_netherlands.osm.pbf
```
The callstack at that moment is
```
...
10 valhalla::baldr::NodeInfo::heading(unsigned int) const      nodeinfo.cc   227
11 (anonymous namespace)::GetTurnTypes()                       graphenhancer.cc   214    
12 (anonymous namespace)::UpdateTurnLanes()                    graphenhancer.cc    418  
13 (anonymous namespace)::enhance(...)                         graphenhancer.cc    1790    
...
```
When during second pass node with index `1116` is processed, it has `edge_index` equal to `3310` (which is a valid value for edge?) and without any further checks it goes down to `GetTurnTypes` function which requests heading

SvetlanaBayarovich: I see that the index to get heading is calculated as `nodeinfo.edge_index() + j` where `j` is `j = 0; j < nodeinfo.edge_count(); j++`. So this does not look like a correct calculation for `localidx` to be used in this function. The `heading` value is needed for `GetTurnTypes` function. If enhancement should be performed for each edge (not only for those 8 local edges) need to get heading from somewhere else

SvetlanaBayarovich: Regarding this PR, should I create a separate issue for this problem and drop this change from in order to deal with it separately?
What's your opinion @kevinkreiser @dnesbitt61 ?

kevinkreiser: i think that would be great yes. it think its clear that something is wrong with the `UpdateTurnLanes` code. It assumes that certain objects are gauranteed to be in vectors but i think sometimes the vectors are empty and so it uses junk data. splitting that out and working on that separately is a good idea

src/mjolnir/valhalla_add_predicted_traffic.cc
```diff
@@ -61,7 +61,7 @@ struct TrafficSpeeds {
 };
 
 // Convert big endian bytes to little endian
-int16_t to_little_endian(const int16_t val) {
+int16_t to_little_endian(const uint16_t val) {
   return (val << 8) | ((val >> 8) & 0x00ff);
```
kevinkreiser, line 65: yeah so bitshifting signed numbers is implementation dependent so im not sure what was happening here. i'm not sure if we have any negative numbers in the input but if we do this change will definitely have an impact on what we do with the data. i'll follow up with you offline with some sample data to see if we need to worry about it

valhalla/midgard/encoded.h
```diff
@@ -88,7 +88,7 @@ std::string encode(const container_t& points, const int precision = 1e6) {
   // handy lambda to turn an integer into an encoded string
   auto serialize = [&output](int number) {
     // move the bits left 1 position and flip all the bits if it was a negative number
-    number = number < 0 ? ~(number << 1) : (number << 1);
+    number = number < 0 ? ~(static_cast<unsigned int>(number) << 1) : (number << 1);
```
kevinkreiser, line 91: this is again bit shifting a signed number and in this case we know its negative. im surprised this doesnt change lots of stuff.. one good test for this would be to run valhalla_export_edges on a larger tileset since it dumps out the whole tilesets shape. then we can look at the diffs in there to see what this does. maybe we were just lucky on the platforms that we run that the implementation dependent handling was equivalent to the above? we need to test this more rigorously

SvetlanaBayarovich: Even more lucky, because `implementation dependent` is for right shift of negative value, for left it's UB.
What exactly do you suggest to test? As for correctness, there is a unit-test eg [this one](https://github.com/valhalla/valhalla/blob/master/test/encode.cc#L393) which does encoding-decoding and definitely creates the case with negative values.

kevinkreiser: well my suggestion was as i stated:

> one good test for this would be to run valhalla_export_edges on a larger tileset ...

yep i am aware of the unit test (i wrote the original), but its not extensive. the reason i want to do a broader test is because this touches literally everything. every single route, mapmatch, how the data is stored. this is a fundamental change so we better be sure its ok. same with the other one above

kevinkreiser: i wrote a small program just to check `int << 1` vs `static_cast<unisigned int>(int) << 1` and it seems that at least on my computer the two operations have identical results.

src/baldr/nodeinfo.cc
SvetlanaBayarovich: Dropping this change. Separate issue was created

valhalla/midgard/encoded.h
SvetlanaBayarovich: @kevinkreiser I did the testing we discussed
1. Build tiles
2. export edges -> compare
3. Add live traffic -> compare tiles
for Utrecht tiles. Comparison showed no difference. Is there any testing to be done, eg regarding [this comment](https://github.com/valhalla/valhalla/pull/2498#discussion_r463638921) ?

Issue #2500, opened 2020-07-28
Traffic Tile Versioning Too Strict


kevinkreiser:
for now we should disable this checksum check:
https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/traffictile.h#L156

At the moment we compute checksum based off of the contents of the header and verify that the same header was used by other systems that may be creating traffic tiles. This means that even changes to comments or adding new helper functions to the header will make the software no longer compatible with previous versions of the data. This is not desirable as it breaks the common contract of our tile format which can be understood as, if the structure of the data hasnt changed, then the software is still compatible with it.

We have a couple of options we can do here:

1. use manual versioning system whereby we hard code the version in the header. this puts the burden on us to know when we should change this (this is how regular graph tiles work)
2. find a way to reduce the checksum to just the fields of the various traffic structures. we could do this by wrapping them in comments and grepping them out in the cmake command for example. this has the downside that we cannot rename any of the fields.

PR #2503, opened 2020-07-29
fix(traffic-tile-version): Locks down traffic tile format

PureW:
and formalises process of changing format. Approach is same as the
regular data tiles, i.e. forward compatible by only using spare bits.

There's a 'break-the-glass' solution for when the format can move no
more, and a new TRAFFIC_TILE_VERSION must be created

Fixes https://github.com/valhalla/valhalla/issues/2500

### DIFF
--- a/valhalla/baldr/traffictile.h
+++ b/valhalla/baldr/traffictile.h
@@ -17,6 +17,7 @@
 #else
 #include <stdint.h>
 #endif
+#include "valhalla.h"
 
 #ifndef C_ONLY_INTERFACE
 namespace valhalla {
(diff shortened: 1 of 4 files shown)


valhalla/baldr/traffictile.h
```diff
@@ -28,6 +27,14 @@ using std::uint32_t;
 using std::uint64_t;
 #endif
 
+// The version of traffic tile format
+// This is not intended to allow for easy migration between in-compatible data-formats
+// Changes are still expected to be applied in a backwards compatible way by using
+// the spare bits available.
+// This is more of a break-the-glass escape hatch for when the current format has
+// reached its limits
+const uint8_t TRAFFIC_TILE_VERSION = 3;
```
kevinkreiser, line 36: can we use valhalla_version_major here

PureW: Yup, fix pushed.

PR #2505, opened 2020-07-29
fixes #2501

### LINKED ISSUE #2501: Windows Builds are Broken
At the moment we cant build windows because we cant install the zlib dependency. This seems to be happening because of: https://github.com/microsoft/vcpkg/issues/12606

It seems that we can tell vcpkg to lock to a specific version to avoid the currently broken dependency. cc @mloskot @mattrjackson @gknisely

kevinkreiser:
fixes #2501 

an attempt to fix windows CI by cloning vcpkg at an earlier commit hash as was done here: https://github.com/expertisesolutions/efl/pull/346/files#diff-222da3d5159ccbea64c56d21b445f986R77


mandeepsandhu:
Seems like the build is failing still. I can't see what the exact error is. I see a whole bunch of warnings and then:
```
...
##[error]Cmd.exe exited with code '1'
```

Do you know what command is failing?

kevinkreiser:
@mandeepsandhu yep seems like other changes since the build has been failing are now also causing a failure:

> D:\a\1\s\valhalla/baldr/traffictile.h(137): error C2580: 'valhalla::baldr::TrafficSpeed::TrafficSpeed(const valhalla::baldr::TrafficSpeed &)': multiple versions of a defaulted special member functions are not allowed [D:\a\build\src\baldr\valhalla-baldr.vcxproj]

mloskot:
Thanks @kevinkreiser !

PR #2508, opened 2020-07-30
Ubuntu 20.04 build-env

PureW:
Moves the build-env to ubuntu 20.04.

It seems to build locally. ~It wont build on CI due to CI still using the 16.04 image for now.~


docker/install-shared-deps.sh
```diff
@@ -1,19 +1,63 @@
...
-apt-get install -y \
+# Lets start with setting us up for installing the non-trusted ppa of primeserver
+apt-get update --assume-yes
+apt-get install --assume-yes \
+    gnupg2
+
+# We're manually adding the sources since the ppa has no 20.04 package yet
+cat << EOF > /etc/apt/sources.list.d/foo.list
+deb http://ppa.launchpad.net/valhalla-core/valhalla/ubuntu bionic main
+deb-src http://ppa.launchpad.net/valhalla-core/valhalla/ubuntu bionic main
```
kevinkreiser, line 15: yeah im blocking on this. i have prime_server ready to go on 20.04 the problem is i just need to push it up when i find the time and let the ppa build it. ill do that when i get a free moment and then we can simplify this

kevinkreiser:
ok i think we can change it back to latest and merge!

PR #2510, opened 2020-07-30
double bucket precision fix

PureW:
While doing the property based tests of DoubleBucketQueue in https://github.com/valhalla/valhalla/compare/anders-double-bucket-prop-test

the following test-case emerged failing. It seems the `float` precision for the internal calculations of the next bucket in `empty_overflow` is inadequate and results in the wrong boundaries being calculated. Specifically, `maxcost_` ends up being equal to `min` when `min` should be less than `maxcost_`.

The last commit fixes this by changing the internal state to use `double`s.


dnesbitt61:
If the issue was in empty_overflow would it have been possible to just use double precision within this method to fix the issue? Did you do a performance comparison to see if float-double casting might slow things down since almost all costing math is floating point.

PureW:
> If the issue was in empty_overflow would it have been possible to just use double precision within this method to fix the issue? Did you do a performance comparison to see if float-double casting might slow things down since almost all costing math is floating point.

I divided up the changes some more. Turns out turning `mincost_` to a double is all that is required. Which makes sense, as that's the accumulator that sees a lot of additions. I couldn't get the test passing by simply casting the operations on the rhs of `mincost_ += ` to doubles.

Unfortunately, the change of `mincost_` to double seems to increase a-star routing times with ~3% on my test set :(

test/double_bucket_queue.cc
```diff
@@ -18,18 +19,28 @@ namespace {
...
   for (auto expected : expectedorder) {
+    // Do the same transform that's done in `edgecost()`
+    expected = (uint32_t)(float)expected;
     uint32_t labelindex = adjlist.pop();
-    EXPECT_EQ(edgelabels[labelindex], expected) << "TryAddRemove: expected order test failed";
+    auto edgelabel = 0;
+    if (labelindex != kInvalidLabel) {
+      edgelabel = edgelabels[labelindex];
+    } else {
+      fprintf(stderr, "#####################################  INVALID LABEL INDEX\n");
```
mandeepsandhu, line 41: Curious, why we're not using `cerr`?

test/double_bucket_queue.cc
```diff
@@ -72,6 +83,16 @@ TEST(DoubleBucketQueue, TestClear) {
   TryClear(costs);
 }
 
+TEST(DoubleBucketQueue, RC4FloatPrecisionErrors) {
+  // Tests what happens when the internal floats in DoubleBucketQueue loses
+  // precision
+
+  std::vector<uint32_t> costs = {1320209856};
```
mandeepsandhu, line 90: Whats this value? Also, why do run sort on the single element array (does it trigger a bug)?

test/double_bucket_queue.cc
PureW: I found the format syntax so much simpler to use than the piping using `<<` operator, especially when trying to print in different representations. For this line in particular it didn't matter much I guess.

test/double_bucket_queue.cc
PureW: Yes, the testcase is one generated from the property test RNG in https://github.com/valhalla/valhalla/compare/anders-double-bucket-prop-test

The sort is not important, it was simply the same pattern that the rest of the tests were following.

test/double_bucket_queue.cc
mandeepsandhu: It looks like the rest of the files in `test` use cout/cerr for console logs. I would prefer to avoid and 3rd style in the mix. If you feel strongly about adding `fprintf`, then thats fine too.

src/CMakeLists.txt
```diff
@@ -99,6 +99,12 @@ function(valhalla_module)
     target_compile_options(${library} PUBLIC -ftest-coverage -fprofile-arcs)
   endif()
 
+  # On more modern x86 processors that support SSE2, specifying the compiler options -mfpmath=sse -msse2
+  # ensures all float and double operations are performed in SSE registers and correctly rounded.
+  # These options do not affect the ABI and should therefore be used whenever possible for
+  # predictable numerical results
+  target_add_compile_flags_if_supported(${library} PRIVATE -mfpmath=sse -msse2)
```
mandeepsandhu, line 106: It seems these options are required only for x86-32, and are enabled by default on x86-64 (see section on `-mfpmath=unit` in https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html). Maybe worthwhile mentioning it here. Better still, can we conditionally enable them only for x86-32? (if not, this is fine too)

kevinkreiser: i agree with this, can we limit this intervention to 32bit x86 builds? it would be nice if we could specify this only for the 32bit build in CI tbh. does cmake allow us to generically add compiler flags with an of the `-D` options? i suppose we need to add our own option for that. that would be my preference. @PureW what do you think?

kevinkreiser: it seems that these flags are not completel redundant with automatically generated ones on 64bit bit x86 platforms however, measuring the performance difference shows its always faster than defaults.

test/double_bucket_queue.cc
```diff
@@ -5,6 +5,7 @@
 #include <cmath>
 #include <cstdint>
 #include <iostream>
+#include <string>
```
kevinkreiser, line 8: not needed i guess

PR #2515, opened 2020-08-05
Fixed seed for shuffle at validate stage

merkispavel:
# Issue

This PR
- forces using `mjolnir.concurrency` config field rather than `concurrency`
- sets fixed seed for random_shuffle

In reproducible tile build we trust

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,6 +3,5 @@ coverage:
   ignore:
     - third_party/.*
     - bench/.*
-    - test/.*
 
 comment: off
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,5 +3,6 @@ coverage:
   ignore:
     - third_party/.*
     - bench/.*
+    - test/.*
 
 comment: off
--- a/src/mjolnir/graphtilebuilder.cc
+++ b/src/mjolnir/graphtilebuilder.cc
@@ -930,7 +930,6 @@ void GraphTileBuilder::AddBins(const std::string& tile_dir,
     auto bin = tile->GetBin(i % kBinsDim, i / kBinsDim);
     bins[i].assign(bin.begin(), bin.end());
     bins[i].insert(bins[i].end(), more_bins[i].cbegin(), more_bins[i].cend());
-    std::sort(bins[i].begin(), bins[i].end());
     shift += more_bins[i].size();
   }
   shift *= sizeof(GraphId);
--- a/src/mjolnir/graphtilebuilder.cc
+++ b/src/mjolnir/graphtilebuilder.cc
@@ -930,6 +930,7 @@ void GraphTileBuilder::AddBins(const std::string& tile_dir,
     auto bin = tile->GetBin(i % kBinsDim, i / kBinsDim);
     bins[i].assign(bin.begin(), bin.end());
     bins[i].insert(bins[i].end(), more_bins[i].cbegin(), more_bins[i].cend());
+    std::sort(bins[i].begin(), bins[i].end());
     shift += more_bins[i].size();
   }
   shift *= sizeof(GraphId);
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -556,7 +556,7 @@ void GraphValidator::Validate(const boost::property_tree::ptree& pt) {
   for (const auto& id : tileset) {
     tilequeue.emplace_back(id);
   }
-  // fixate seed for reproducible tile build
+  // fixed seed for reproducible tile build
   std::shuffle(tilequeue.begin(), tilequeue.end(), std::mt19937(3));
 
   // Remember what the dataset id is in case we have to make some tiles
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -568,8 +568,7 @@ void GraphValidator::Validate(const boost::property_tree::ptree& pt) {
   // Setup threads
   std::vector<std::shared_ptr<std::thread>> threads(
       std::max(static_cast<unsigned int>(1),
-               pt.get_child("mjolnir").get<unsigned int>("concurrency",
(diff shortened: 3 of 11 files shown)


src/mjolnir/graphvalidator.cc
```diff
@@ -556,7 +556,8 @@ void GraphValidator::Validate(const boost::property_tree::ptree& pt) {
   for (const auto& id : tileset) {
     tilequeue.emplace_back(id);
   }
-  std::random_shuffle(tilequeue.begin(), tilequeue.end());
+  // fixate seed for reproducible tile build
+  std::shuffle(tilequeue.begin(), tilequeue.end(), std::mt19937(3));
```
merkispavel, line 560: utrecht tiles build use concurrency=1 so it wasn't the main cause. However, the fixating seed still makes sense imho

src/mjolnir/graphtilebuilder.cc
```diff
@@ -930,6 +930,7 @@ void GraphTileBuilder::AddBins(const std::string& tile_dir,
     auto bin = tile->GetBin(i % kBinsDim, i / kBinsDim);
     bins[i].assign(bin.begin(), bin.end());
     bins[i].insert(bins[i].end(), more_bins[i].cbegin(), more_bins[i].cend());
+    std::sort(bins[i].begin(), bins[i].end());
```
merkispavel, line 933: the main fix

merkispavel:
Going to add a test to track such issue in the future. Will commit soon

src/mjolnir/graphvalidator.cc
kevinkreiser: i cant understand how fixing the seed doesnt resolve the issue. how is the order of the bin indeterminant if only 1 thread is doing the work?

src/mjolnir/graphtilebuilder.cc
kevinkreiser: this has a very unfortunate side effect which is that edges in the bin that are in this tile wont be the first edges in the bin. i would argue that this is not a "sensible" sorting of the data for that reason. we can work around it by providing our own sort lambda here though.

src/mjolnir/graphvalidator.cc
merkispavel: yea, turned out `Validate` section looked at the wrong `concurrency` field. Now it takes `mjolnir.concurrency` => fixated seed solves the issue

merkispavel: btw, seed also helps with single-thread `reproduce_tile_build`-test because `random_shuffle` uses `rand()` under the hood(I guess) so different builds of the test use different shuffles in that case
So we might think about setting seed for rand function in the very begging of the tile building
```cpp
// e.g

bool build_tile_set(...) {
  srand(3); // fixated magic seed for rand() function
  ...
```

merkispavel:
omg, codecov fails because of the uncovered code in `test/gurka/test_reproduce_tile_build.cc` 😕

src/mjolnir/graphvalidator.cc
kevinkreiser: @merkispavel that makes sense but... that has the effect of being global which isnt good for unit tests, i think what you have here is good where you pass the generator to shuffle

kevinkreiser:
@merkispavel dont worry about codecov we frequently ignore it, sometimes it also produces nonsense results (or at least seemingly so)

codecov.yml
```diff
@@ -3,5 +3,6 @@ coverage:
   ignore:
     - third_party/.*
     - bench/.*
+    - test/.*
```
kevinkreiser, line 6: doing this brings our codecoverage from 85% down to 79% :(

src/mjolnir/graphvalidator.cc
```diff
@@ -556,7 +556,8 @@ void GraphValidator::Validate(const boost::property_tree::ptree& pt) {
   for (const auto& id : tileset) {
     tilequeue.emplace_back(id);
   }
-  std::random_shuffle(tilequeue.begin(), tilequeue.end());
+  // fixate seed for reproducible tile build
```
kevinkreiser, line 559: `s/fixate/fixed`

codecov.yml
merkispavel: this is probably the kind of change that should be discussed with the whole team. **I reverted this one** as codecov check is not marked as "required"(didn't notice it before)
Though it makes total sense to me: we should mix test code coverage with business-logic code coverage

src/mjolnir/graphtilebuilder.cc
kevinkreiser: the sort lambda could look like this:

```c++
[&tile_id](const GraphId& edge_a, const GraphId& edge_b) {
  // this tile always wins
  if(edge_a.tile_base() == tile_id && edge_b.tile_base() != tile_id)
    return true;
  // this tile always wins
  if(edge_b.tile_base() == tile_id && edge_a.tile_base != tile_id)
    return false;
  
  // either they are both in the same tile or they are both not in this tile
  // in either case we want to sort by level then by tile then by id but the
  // bits are arranged backwards for that so we flip them
  return edge_a.flip() < edge_b.flip();
}
```

and then we make a flip function in graphid.cc because this is useful:

```c++
  /**
   * Flips the bit order from id/tile/level to level/tile/id, so that sorting makes more sense
  *  The ordering of the bits in the id could be thought of as a design oversight
   * @return   Returns the graphid but with the fields in flipped order
   */
uint64_t flip() const {
  return static_cast<uint64_t>(id()) | (static_cast<uint64_t>(tileid()) << 21) | (static_cast<uint64_t>(level()) << 25);
}
```

we could then find where we do sorting like this (i think it must be all over the code) and use the flip function instead: https://github.com/valhalla/valhalla/blob/master/src/loki/node_search.cc#L209

PR #2442, opened 2020-06-25
Invalid match indexes cause out of vector bounds lookup

dnesbitt61:
Protect against invalid match indexes in Uturns when setting edge_trimming lnglat.

Fixes #2439 

# Issue

What issue is this PR targeting? If there is no issue that addresses the problem, please open a corresponding issue and link it here.

## Tasklist

 - [ ] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [ ] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #2517, opened 2020-08-06
fix garbage coordinates when map matched route has uturn

kevinkreiser:
fixes #2442

### DIFF
--- a/src/thor/trace_route_action.cc
+++ b/src/thor/trace_route_action.cc
@@ -392,16 +392,19 @@ void thor_worker_t::build_route(
       const auto* prev_segment = i > 0 ? path.second[i - 1] : nullptr;
       const auto* segment = path.second[i];
       const auto* next_segment = i < path.second.size() - 1 ? path.second[i + 1] : nullptr;
-      // if we uturn onto this edge we must trim the beginning and if we uturn off we trim the end
-      bool uturn_onto =
-          prev_segment && prev_segment->edgeid != segment->edgeid && prev_segment->target < 1.f;
-      bool uturn_off_of =
-          next_segment && segment->edgeid != next_segment->edgeid && segment->target < 1.f;
-      if (uturn_onto || uturn_off_of) {
-        edge_trimming[i] = {{uturn_onto, match_results[segment->first_match_idx].lnglat,
-                             segment->source},
-                            {uturn_off_of, match_results[segment->last_match_idx].lnglat,
-                             segment->target}};
+
+      // Note: below we use the operator[] to default initialize the "trim" boolean to false
+      // and then we overwrite it. Since its a pair this handles the case when just one is set
+
+      // if we uturn onto this edge we must trim the beginning
+      if (prev_segment && prev_segment->edgeid != segment->edgeid && prev_segment->target < 1.f) {
+        edge_trimming[i].first = {true, match_results[segment->first_match_idx].lnglat,
+                                  segment->source};
+      }
+      // if we uturn off of this edge we must trim the end
+      if (next_segment && segment->edgeid != next_segment->edgeid && segment->target < 1.f) {
+        edge_trimming[i].second = {true, match_results[segment->last_match_idx].lnglat,
+                                   segment->target};
       }
     }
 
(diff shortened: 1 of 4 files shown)


src/thor/trace_route_action.cc
```diff
@@ -392,16 +392,19 @@ void thor_worker_t::build_route(
       const auto* prev_segment = i > 0 ? path.second[i - 1] : nullptr;
       const auto* segment = path.second[i];
       const auto* next_segment = i < path.second.size() - 1 ? path.second[i + 1] : nullptr;
-      // if we uturn onto this edge we must trim the beginning and if we uturn off we trim the end
-      bool uturn_onto =
-          prev_segment && prev_segment->edgeid != segment->edgeid && prev_segment->target < 1.f;
-      bool uturn_off_of =
-          next_segment && segment->edgeid != next_segment->edgeid && segment->target < 1.f;
-      if (uturn_onto || uturn_off_of) {
-        edge_trimming[i] = {{uturn_onto, match_results[segment->first_match_idx].lnglat,
```
kevinkreiser, line 401: as @dnesbitt61 mentioned the match index can be `-1` but only in the case that we dont need to trim. sadly the trimming datastructure is a pair and so you have to specify both first and second. if though one of them doesnt need to be set in this case. to get around this i assign one at a time and rely on the default initializer to intialize the pair with false and invalid lat lon etc

valhalla/meili/match_result.h
```diff
@@ -94,7 +94,7 @@ struct EdgeSegment {
 
 struct MatchResults {
   MatchResults(std::vector<MatchResult>&& results, std::vector<EdgeSegment>&& segments, float score)
-      : results(results), segments(segments), score(score) {
+      : results(std::move(results)), segments(std::move(segments)), score(score) {
```
yuzheyan, line 97: maybe using std::forward?

yuzheyan, line 97: maybe using std::forward to handle rv, and std::move to handle lv ref? Just some nitpicking

kevinkreiser:
so the request that exposes this ends up showing this at the important part of the code:

```
(gdb) p *prev_segment
$9 = {edgeid = {value = 397962009178}, source = 0, target = 1, first_match_idx = -1, 
  last_match_idx = -1, discontinuity = false}
(gdb) p *segment
$10 = {edgeid = {value = 985399450202}, source = 0, target = 0.974345565, 
  first_match_idx = -1, last_match_idx = 25, discontinuity = false}
(gdb) p *next_segment
$11 = {edgeid = {value = 1357316774490}, source = 0.0256544352, target = 1, 
  first_match_idx = 25, last_match_idx = -1, discontinuity = false}
```

basically it turns onto an edge and then just before reaching the end of the edge it turns around and leaves the edge.

so i made a test case for this:

```c++
TEST(Standalone, UturnTrimmingAsan) {
A--1--2--B
         |
         |
D--3--4--C--5--6--E
                                )";

  const gurka::ways ways = {
      {"AB", {{"highway", "primary"}}},
      {"BC", {{"highway", "primary"}}},
      {"DCE", {{"highway", "primary"}}},
  };

  const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/uturn_asan");

  auto result =
      gurka::match(map, {"2", "6", "3"}, "through", "auto", R"({"penalize_immediate_uturn":false})");
}
```

But it doesnt expose the same problem..

test/gurka/test_match.cc
```diff
@@ -86,3 +86,23 @@ TEST(Standalone, UturnMatch) {
...
+  const gurka::ways ways = {
+      {"AB", {{"highway", "primary"}}},
+      {"BC", {{"highway", "primary"}}},
+      {"DCE", {{"highway", "primary"}}},
+  };
+
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/uturn_asan");
+
+  auto result =
```
PureW, line 106: Shouldn't the result be checked?

PR #2518, opened 2020-08-07
Parseways speedup

merkispavel:
# Issue

This is POC that string comparisons take significant amount of time during parseways stage.
Both tilebuilding(parseways stage) on my machine and benchmarks showed 40% speedup
North America extracted region: 49min -> 29min

---

The main idea is to replace multiple string comparisons
```cpp
if (tag.first == "a")
   ...
else if (tag.first == "b")
 ...
else if (tag.first == "c")
...
```
with
```cpp
map<std::string, std::function<void()>> handlers;
handlers["a"] = []() { ... };
handlers["b"] = []() { ... };

// and then just call
handlers[tag.first]();
```

## Tasklist

 - [ ] Add tests

## Requirements / Relations


mandeepsandhu:
Neat trick! 👍

LaGrunge:
Just like an Idea, have a look at 
https://github.com/smilingthax/cttrie
for your PR

merkispavel:
> Just like an Idea, have a look at
> https://github.com/smilingthax/cttrie
> for your PR

Sure, I also thought about a trie: used unordered_map just for the POC. However, didn't know about **compile-time trie**. Thanks for the hint!

It should be noticed that my previous measurements(profiling) showed that string comparisons take 42% of the time while this PR already gives -40%.
However, there are also things(e.g cache hits/misses) that are not shown in the call stacks. In measurements we trust!

LaGrunge:
I am pretty sure, map is not bottleneck here, and you did pretty good job.
Maybe using this lib, you will have smaller diff
I have used it in same task, but I do not insist of course

LaGrunge:
Also will be glad to help you with profiling and thankful for your advice about measuring tools

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1785,6 +1831,22 @@ struct graph_callback : public OSMPBF::Callback {
     bss_nodes_.reset(bss_nodes);
   }
 
+  // merkispavel
```
gknisely, line 1834: clean up

merkispavel: oops, used this one as a landmark..

gknisely:
> This looks fantastic. Did you compare the stats(number of nodes, edges, exits, restrictions...all the numbers) from the output from valhalla_build_tiles.

You could run this on a smaller pbf to do a quick check.

LaGrunge:
It seems to me, you have committed some changes to have reproduced build tiles. So like some kind of late prove we can see md5 of planet build before and without your PR. I can test it, if you want

mandeepsandhu:
> > Just like an Idea, have a look at
> > https://github.com/smilingthax/cttrie
> > for your PR
> 
> Sure, I also thought about a trie: used unordered_map just for the POC. However, didn't know about **compile-time trie**. Thanks for the hint @LaGrunge !
> 

Trie's are helpful if you have strings where we do a lot of prefix matches. I'm not if thats the pattern with the strings being compared here.

kevinkreiser:
i swore that @kdiluca tried this trick some months ago...

merkispavel:
> > This looks fantastic. Did you compare the stats(number of nodes, edges, exits, restrictions...all the numbers) from the output from valhalla_build_tiles.
> 
> You could run this on a smaller pbf to do a quick check.

Makes total sense. 
This is my first iteration: seems like a working POC.  The next step is to make sure that nothing is broken. I thought about comparing raw bin files produced by `parseways` stage. Or I can check something from your ideas(number of nodes..). So yea, to be continued soon

merkispavel:
> It seems to me, you have committed some changes to have reproduced build tiles. So like some kind of late prove we can see md5 of planet build before and without your PR. I can test it, if you want

End-to-end planet build might be an overkill here in my view. As the PR "touches" only parseways stage it might be enough to check the consistency of the files produced by the parseways stage only. 

About reproducibility, I guess a single-thread build will be reproducible after #2515 . However, there're still some unresolved issues with multithreaded build(it's still flaky a little bit). We should be careful here
> I can test it if you want

If you have time - sure, why not 👍

merkispavel:
I hope I found and fixed all hidden bugs: a harmless (from the first sight) change took a day of my life :(
I tested parseways stage output of my branch against master: all output files' checksums match!

notes:
- Parseways isn't completely "reproducible" at the moment. Tested the PR with https://github.com/valhalla/valhalla/pull/2530
- Tested with North America pbf

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1785,6 +1880,24 @@ struct graph_callback : public OSMPBF::Callback {
     bss_nodes_.reset(bss_nodes);
   }
 
+  // WayCallback tag handlers
+  using TagHandler = std::function<void()>;
+  std::unordered_map<std::string, TagHandler> tag_handlers_;
+  // Tag handlers capture these fields
```
kevinkreiser, line 1886: instead of capturing `this` you should have captured `&` and not had to make all of these temporaries into member variables

kevinkreiser:
other than the moving of temporaries to member variables i think this looks ready to merge

src/mjolnir/pbfgraphparser.cc
merkispavel: > instead of capturing `this` you should have captured `&` and not had to make all of these temporaries into member variables

The idea of making these variables be fields is to construct `tag_handlers` map once(i.e in constructor). I didn't measure but I guess constructing map on every `way_callback` call may affect performance quite a lot

kevinkreiser: ah i see that is a good reason to do this. its unfortunate though since they are temporaries but the messiness is worth the speed up

PR #2519, opened 2020-08-07
Update timezone database in correct location

### LINKED ISSUE #2118: valhalla_build_timezones API has changed
Looks like with commit https://github.com/valhalla/valhalla/commit/76ab18c71e5ee0d3af53fd190e4873e82d4ad996, valhalla_build_timezones prints out constructed database into stdout. I'd say its somewhat unexpected on the basis of docs, where it is stated (or suggested) that timezone file will be in 'mjolnir.timezone' after running that script.

irees:
# Issue

I noticed that `valhalla_build_timezone` broke at some point; it dumps the sqlite db to stdout instead of updating it in the location specified in the config file.

This PR restores the previous behavior.

Fixes #2118 

## Tasklist

 - [ ] Add tests

## Requirements / Relations

None

### DIFF
--- a/scripts/valhalla_build_timezones
+++ b/scripts/valhalla_build_timezones
@@ -15,6 +15,18 @@ if  ! which unzip >/dev/null; then
     exit 1
 fi
 
+config=$1
+if [ ! -f $config ]; then
+    echo "Config file not found $config"
+    exit 1
+fi
+tz_file=$(jq -r '.mjolnir.timezone' $config)
+if [ ! -d $(dirname $tz_file) ]; then
+    echo "Timezone directory not found $(dirname $tz_file)"
+    exit 1
+fi
+rm -f $tz_file
+
 rm -rf dist
 rm -f ./timezones-with-oceans.shapefile.zip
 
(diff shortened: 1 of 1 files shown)


irees:
Converting back to a draft, since it looks like this breaks other things and/or timezones are handled by some other process now. Looking into it.

kevinkreiser:
@irees yeah its not broken. i changed to be this way on purpose. it seems rediculous to me that you need a config file to run the script at all. any normal bash program would either take the output location as an argument or, as i did, write the output to stdout and have the caller forward it to file. if this is causing you trouble i think you can solve it pretty readily with something like this:

```bash
tz_location=$(cat valhalla.json | jq -r .mjolnir.timezone)
valhalla_build_timezones > ${tz_location}
```

kevinkreiser:
@irees i updated the one spot in the docs where it had the old syntax for the script. if you want to change the script to take an argument (the location where the output should go) that might be reasonable?

irees:
@kevinkreiser The updated docs are sufficient; I was just confused when I got a large sqlite database binary dumped into my log files :) I'll close this PR.

PR #2352, opened 2020-05-04
Add support for LuaJIT

danpat:
# Issue

Migrates to LuaJIT for lua scripting.  Speeds up the `parse` phase of `valhalla_build_tiles` by about 12%.

```
# Before
$ /usr/bin/time ./valhalla_build_tiles -c config.json --start parse --end parse bulgaria-latest.osm.pbf
       38.91 real        35.94 user         1.91 sys
# After
$ /usr/bin/time ./valhalla_build_tiles -c config.json --start parse --end parse bulgaria-latest.osm.pbf
       30.35 real        28.23 user         1.61 sys
```

## Tasklist

 - [x] Add LuaJIT to test environment on CI to make sure it's not totally borked.
 - [ ] Add tests
 - [x] Add #fixes with the issue number that this PR addresses
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [x] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #2521, opened 2020-08-10
Towards better Windows compatibility: TOOLS, documentation & Python

nilsnolde:
#2300 

This is the first bit to improve Windows compatibility for Valhalla with big thanks to @mattrjackson who did most of the changes:
- update Lua syntax to 5.3 (include new bit operator instead of separate library from 5.2: http://www.lua.org/manual/5.3/manual.html#3.4.2)
- more OS specific path issues (which probably could be handles more gracefully than what I did)

With these modifications, it's possible to run the valhalla_build_tiles.exe and the valhalla_run_* executables.

I'll also try to build TOOLS and DATA_TOOLS on Azure, which should work now. Although there's no testing enabled yet for Windows.

## Tasklist

Let me know what you think of the tasks below...

 - [ ] dev documentation for Windows in `/docs`, short and crispy
 - [ ] build Python bindings on Windows (hopefully only a documentation item as well)
 - [ ] **maybe**: as long as no proper testing is possible on Azure, manually test some of the executables in the Win pipelines?


PureW:
> update Lua syntax to 5.3 

I believe we're using LuaJIT for performance reasons which uses the 5.1 syntax?

src/mjolnir/graphtilebuilder.cc
```diff
@@ -191,9 +191,12 @@ GraphTileBuilder::GraphTileBuilder(const std::string& tile_dir,
 
 // Output the tile to file. Stores as binary data.
 void GraphTileBuilder::StoreTileData() {
+  std::string name{tile_dir_ + filesystem::path::preferred_separator +
+                            GraphTile::FileSuffix(header_builder_.graphid())};
```
kevinkreiser, line 195: maybe we should modify this function to take the separator. the problem here is sometimes we use it for an actual file path and sometimes we use it for an identifier for the tile. so allowing to pass the separator here would allow both use-cases in the code

danpat:
@nilsnolde @PureW An early version of my LuaJIT PR did support multiple syntaxes (by re-writing the `.lua` file when it was converted to a string - it's a simple search/replace operation currently), but we scrapped that for simplicity and decided to just stick with the LuaJIT 5.1 syntax + extensions.

@nilsnolde From https://github.com/valhalla/valhalla/pull/2352, LuaJIT outperformed Lua5.3 by ~12% for the `parse` phase of `valhalla_build_tiles` - do you think the gains are significant enough to justify regressing performance there?

src/mjolnir/graphtilebuilder.cc
nilsnolde: I'll look into it, makes sense

nilsnolde:
Thanks for the insight @danpat ! I found it a little confusing, but maybe you can shed some light on it: using LuaJIT means one can entirely get rid of Lua5.x libraries? I have both installed on Win, might explain why I have to manually tell cmake how to link the proper paths to include and lib (and likely doing it wrong on top of that). Then it might use Lua5.3 which is why it can't make sense of the `bit.bor` calls. I'll try that again on Wednesday, family day tmrw:)

PureW:
> using LuaJIT means one can entirely get rid of Lua5.x libraries?

You need luajit libraries. It's a completely different implementation of lua 5.1.

nilsnolde:
Right, it probably got confused since I had both installed, luajit and 5.3.

BTW, no Windows CI builds on PRs? Or are subsequent webhooks disabled when the first CI fails? (and I'll run format.sh, forgot about it)

src/mjolnir/graphtilebuilder.cc
nilsnolde: I changed the signature to take an optional argument and all callers where applicable. However, you're sure that's actually needed @kevinkreiser ? Would it matter if ID contain backslashes on Windows? I tried it without the conditional and at least I could build tiles and use the run_* apps. I didn't try any of this on Linux though. Would also be less changes to the code:)

nilsnolde:
@danpat @PureW removing Lua5.3 from my packages helped! Lua back to previous state.

I also followed Kevin's advice and allowed an optional arg for GraphTile::Suffix, so it's quite a few changed files. Didn't touch the tests, they'd fail currently (if they'd run for Windows). For some other time:)

Not sure if the README addition is really needed. I'm rather considering writing a blog about it, being more detailed. In the current state a Windows build is pretty deducible from the Linux instructions (+ the CI). What do you think?

Once you're happy I'll properly prepare the PR. Was only working on Windows so far and don't wanna deal with why `format.sh` isn't working in the WSL.

lua/graph.lua
```diff
@@ -3,964 +3,887 @@
 --postgres
 
 highway = {
-["motorway"] =          {["auto_forward"] = "true",  ["truck_forward"] = "true",  ["bus_forward"] = "true",  ["taxi_forward"] = "true",  ["moped_forward"] = "false", ["motorcycle_forward"] = "true",  ["pedestrian"] = "false", ["bike_forward"] = "false"},
```
PureW, line 6: What's going on in this file? If it's only formatting, could you move this to a separate PR with an added check in CI to keep it formatted?

nilsnolde: oh crap, that's terrible, somehow skipped my attention.. sorry, will revert..

nilsnolde: done

src/mjolnir/graphtilebuilder.cc
kevinkreiser: sometimes its used as a string for accessing the file system, sometimes its used as a url. at least in the latter we need to be able to always use `/` even if we are running on windows. i suppose the function has become misnamed over the years in that case!

one option would be to make change you suggested where it just always uses the preferred separator by default but then add another helper function called `GraphTile::URLSuffix` along side it which calls it with just `/`. then in `MakeSingleTileUrl` we'd call `GraphTile::URLSuffix`

kevinkreiser: oh also id be happy to help with this!

nilsnolde: Ah right, somehow I forgot about the URL access, I actually did accommodate for that. Did you have a chance to look at the changes? 

`URLSuffix` would be more elegant and less implicit than `isFilePath` kwarg to `FileSuffix`. Your decision, I can integrate the changes, no worries. Thanks for the offer though:)

kevinkreiser: @nilsnolde `isFilePath` should be fine (though id suggest using `_` rather than camal case). there is a bit of a problem here which is that we use the `facet` with a hardcoded separator of `/` to separate the numbers in the path. i think what we'd want to do is instantiate two facets above one with prefered separtor and one with `/` and then based on your `is_file_path` boolean imbue with one or the other.

kevinkreiser: i think if you do taht you'll hardly need to change the `FileSuffix` function

nilsnolde: cool, thanks for the advice. the way you generate the id was a bit mysterious to me, makes more sense now though.

nilsnolde: should be neat enough now:)

nilsnolde:
sorry for all the bs commits! guess that'll need a squash..

nilsnolde:
Finally the Win CI builds work. Will do the same for appveyor then.

Before I finish this off:
- quick approval of the code changes pls
- should we switch to latest VS 2019 in CI? Rather than extending the matrix with 2019?
- how about "manually" running e.g. `valhalla_build_tiles.exe` & `valhalla_run_route.exe` after the build in CI with one of the demo OSM files? As long as test suite isn't running (would like to look into that as well).
- took longer than I expected, so I'll save Python support for another day

PureW:
> * quick approval of the code changes pls

Looks ok to me.
> 
> * should we switch to latest VS 2019 in CI? Rather than extending the matrix with 2019?

Yes please!

> * how about "manually" running e.g. `valhalla_build_tiles.exe` & `valhalla_run_route.exe` after the build in CI with one of the demo OSM files? As long as test suite isn't running (would like to look into that as well).

Not sure what this means, almost none of us build valhalla on Windows and have no ability to do this?

nilsnolde:
> Not sure what this means, almost none of us build valhalla on Windows and have no ability to do this?

I meant running a few TOOLS/DATA_TOOLS executables on CI after it built. Cheap testing, but wouldn't be much effort and prevent at least _some_ runtime regressions for Win?

PureW:
> > Not sure what this means, almost none of us build valhalla on Windows and have no ability to do this?
> 
> I meant running a few TOOLS/DATA_TOOLS executables on CI after it built. Cheap testing, but wouldn't be much effort and prevent at least _some_ runtime regressions for Win?

Ohh yea definitely, that's a great idea.

nilsnolde:
Ok, should be good for review now.

Few more things:
- I wanted to also "test" `valhalla_build_admins.exe` (never tried before on Win), but had trouble with spatialite: sqlite couldn't load the module. Tried for an hour (adding "spatialite.dll" as module name in the code; copying latest official spatialite DLLs which almost worked but errored on another spatialite dependency), didn't work. Smth wrong with the `vcpkg` installation of spatialite, but couldn't find any Github issues.
- When I enabled the address sanitizer, the tests don't pass currently:
```
[FAIL] truckcost
[==========] Running 1 test from 1 test suite.
[----------] 1 test from TruckCost
[ RUN      ] TruckCost.testTruckCostParams
[       OK ] TruckCost.testTruckCostParams (888 ms)
[----------] 1 test from TruckCost (888 ms total)

[==========] 1 test from 1 test suite ran. (889 ms total)
[  PASSED  ] 1 test.

=================================================================
==11043==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 57 byte(s) in 1 object(s) allocated from:
    #0 0x7f9222b9483a in __interceptor_realloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:164
    #1 0x7f9221d67ffb in __argz_add_sep (/usr/lib/libc.so.6+0x91ffb)
    #2 0x7f92fffffffe  (<unknown module>)

SUMMARY: AddressSanitizer: 57 byte(s) leaked in 1 allocation(s).
make[3]: *** [test/CMakeFiles/run-truckcost.dir/build.make:81: test/truckcost.log] Error 1
make[3]: *** Deleting file 'test/truckcost.log'
make[2]: *** [CMakeFiles/Makefile2:9069: test/CMakeFiles/run-truckcost.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:4897: test/CMakeFiles/check.dir/rule] Error 2
make: *** [Makefile:2052: check] Error 2
```

nilsnolde:
x86 build for win is still failing.. don't really understand why.. CMake is set to x86 & Release build, yet it's complaining that there's no config for Win32 & Debug build. Found a similar issue over at [CMake](https://gitlab.kitware.com/cmake/cmake/-/issues/19951).

Anyone got an idea (cc @mloskot @mattrjackson)? If not, I'll try a cross compile myself later this week and see if I can resolve. My Win disk is completely full already, 88 GB to work on a single project :boom: 

Will also add some caching on Azure once x86 works (almost 1 hour CI builds right now!): https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#get-started

nilsnolde:
>  sqlite couldn't load the [spatialite] module.

This is where it exits:
https://github.com/valhalla/valhalla/blob/b2fed9f5c9ff5320ddae99df70fc5614c7243867/src/mjolnir/util.cc#L153

But this worked:
https://github.com/valhalla/valhalla/blob/b2fed9f5c9ff5320ddae99df70fc5614c7243867/src/mjolnir/valhalla_build_admins.cc#L308

So, the code knows where to find `spatialite` but `sqlite` doesn't. I tried sorts of `mod_name`s (DLL is named spatialite.dll on Win), even specifying the whole path, which works on SQLite command line (though a different SQLite installation than the one linked here). I have no idea why it doesn't find spatialite.. Someone got a tip what I could try?

SvetlanaBayarovich:
@nilsnolde As for the sanitizer question, this looks like a known issue with Leak Sanitizer (which is a part of Address Sanitizer): see the [#2296(comment)]( https://github.com/valhalla/valhalla/pull/2296#issuecomment-608304654). Due to this Leak Sanitizer is disabled in CI. You can also disable it while testing locally, e.g. `ASAN_OPTIONS=detect_leaks=0 make check`

mloskot:
> x86 build for win is still failing.. don't really understand why.. CMake is set to x86 & Release build, yet it's complaining that there's no config for Win32 & Debug build.
> Anyone got an idea (cc @mloskot @mattrjackson)? 

The problem is in the CMake command line(s) in this PR: `cmake -A %PLATFORM%` translates to `cmake -A x86` which is wrong.
This should read `cmake -A Win32` as per [allowed values for -A](https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2016%202019.html) and the VS generator.

.azure-pipelines.yml
```diff
@@ -57,9 +56,16 @@ jobs:
 
       - script: |
           cmake --version
-          cmake -H$(Build.SourcesDirectory) -B%BUILD_DIR% -G "%GENERATOR%" -DCMAKE_TOOLCHAIN_FILE=%VCPKG_DIR%\scripts\buildsystems\vcpkg.cmake -DVCPKG_APPLOCAL_DEPS=ON -DENABLE_CCACHE=OFF -DENABLE_DATA_TOOLS=OFF -DENABLE_HTTP=OFF -DENABLE_PYTHON_BINDINGS=OFF -DENABLE_SERVICES=OFF -DENABLE_TOOLS=OFF -DENABLE_BENCHMARKS=OFF
+          cmake -H$(Build.SourcesDirectory) -B%BUILD_DIR% -G "Visual Studio 16 2019" -A %PLATFORM% -T host=%PLATFORM% -DCMAKE_TOOLCHAIN_FILE=%VCPKG_DIR%\scripts\buildsystems\vcpkg.cmake -DVCPKG_APPLOCAL_DEPS=ON -DENABLE_CCACHE=OFF -DENABLE_DATA_TOOLS=ON -DENABLE_HTTP=OFF -DENABLE_PYTHON_BINDINGS=OFF -DENABLE_SERVICES=OFF -DENABLE_TOOLS=ON -DENABLE_BENCHMARKS=OFF -DLUA_INCLUDE_DIR=%VCPKG_DIR%\installed\%PLATFORM%-windows\include\luajit -DLUA_LIBRARIES=%VCPKG_DIR%\installed\%PLATFORM%-windows\lib\lua51.lib
```
mloskot, line 59: Since `PLATFORM=x86`, the argument `-A %PLATFORM%` is not correct for CMake 32-bit build, see https://github.com/valhalla/valhalla/pull/2521#issuecomment-675605376

nilsnolde: Ugh, stupid mistake then.. Thanks @mloskot

nilsnolde:
Good for merge from my side @kevinkreiser @PureW 

Caching on Azure with `vcpkg` is still painful.. which is mostly the `vcpkg` part as there's no preconfigured task for it, we'd have to [bake our own](https://docs.microsoft.com/en-us/azure/devops/extend/develop/add-build-task?view=azure-devops) (which needs privileges on Azure) and follow this one: https://github.com/lukka/CppBuildTasks#developers-information

mloskot:
Thanks @nilsnolde for the great set of improvements.

I used to get excited and configured all free CI services (AzP, AV, Travis, CircleCI) for projects I maintain, but that eventually becomes pain to maintain. Then, the disrupting GitHub Actions arrived and IMHO we should consider complete switch to GA as single CI with jobs for the big three: Linux, Windows and macOS.

nilsnolde:
Yeah, I also loathe all the external CI stuff by now.. For internal stuff we keep using Gitlab with its amazing infrastructure, but the lack of SEO prevents public repos unfortunately. Didn't try Github Actions yet, but I'd be all in favor of ditching external CI.

Now that I got you here @mloskot, did you ever encounter this on Win with spatialite? https://github.com/valhalla/valhalla/pull/2521#issuecomment-675134204

mloskot:
> Didn't try Github Actions yet, but I'd be all in favor of ditching external CI.

I am moving home during next weeks, and GSoC is finalising, so I won't be able to do anything soon, but I will keep this task in mind for the autumn.

> Now that I got you here @mloskot, did you ever encounter this on Win with spatialite?

I have seen this issue, but it did not ring any bells. I may be wrong, can't verify it now, but I think did not build Valhalla with SQLite support.

PR #2522, opened 2020-08-10
[Do Not Merge] New Urban Area Flag

kdiluca:
Added new urban flag that can be set if edge is within city boundaries to data processing and new config option

# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


dnesbitt61:
So this uses the urban flag to influence speeds. But it doesn't actually set a flag in the Valhalla graph tiles?

src/mjolnir/graphbuilder.cc
```diff
@@ -1028,6 +1029,10 @@ void BuildTileSet(const std::string& ways_file,
 
         graphtile.nodes().back().set_timezone(tz_index);
 
+        // set the density as needed.
+        if (use_urban_tag && node.urban())
+          graphtile.nodes().back().set_density(kMaxDensity);
```
kevinkreiser, line 1034: its unfortunate that we have to do this here and have density set in two places but there is no way around it

PR #2528, opened 2020-08-13
Assert Costing Name is Present

kevinkreiser:
In the recosting PR we added the concept of naming a costing, this was needed so we could specify multiple of the same type with different options. We also changed the osrm serializer to use this name as the "weight_name". The problem was, we didnt make sure that the name was ever set. If it is provided in the request it is used but if its not provided the default is not set. This could cause undefined behavior (in older versions of protobuf, recent versions of protobuf seem to return empty string for unset optional strings). At any rate I added an assert which made a bunch of tests fail. I then fixed the costing options parsing to set a default and let it be overrided if provided by the user. Once that was done almost everything was passing except for the god forsaken pin point tests :smile: 

This is because they use stored versions of the request in pbf format. So i needed to update those serialized pbfs. To do that I added a small snippet of code to the pinpoints to properly set the costing name:

```c++
  if (!request.options().costing_options(request.options().costing()).has_name()) {
    request.mutable_options()
        ->mutable_costing_options(request.options().costing())
        ->set_name(Costing_Enum_Name(request.options().costing()));
    std::ofstream out(filename, std::ios::out | std::ios::binary | std::ios::trunc);
    out << request.SerializeAsString();
    out.close();
  }
```

After that all tests passed. I also noticed that some of the includes were messed up between <> and "" so i fixed those as well. General rule of thumb is public headers reference valhalla headers with <> where as source and private headers can use "".


src/baldr/laneconnectivity.cc
```diff
@@ -1,8 +1,8 @@
 #include "baldr/laneconnectivity.h"
+#include "midgard/logging.h"
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/split.hpp>
 #include <string.h>
-#include <valhalla/midgard/logging.h>
```
kevinkreiser, line 5: the bulk of the changes in this pr are just fixing includes like this. ill comment only on the substantive parts of the rest of the pr

src/proto_conversions.cc
```diff
@@ -0,0 +1,215 @@
+#include "proto_conversions.h"
+
+namespace valhalla {
+
+bool Options_Action_Enum_Parse(const std::string& action, Options::Action* a) {
```
kevinkreiser, line 5: since these are indeed proto conversions, i moved them out of worker.h/cc and into the proto_conversions.h/cc

src/sif/autocost.cc
```diff
@@ -556,6 +556,7 @@ void ParseAutoCostOptions(const rapidjson::Document& doc,
                           const std::string& costing_options_key,
                           CostingOptions* pbf_costing_options) {
   pbf_costing_options->set_costing(Costing::auto_);
+  pbf_costing_options->set_name(Costing_Enum_Name(pbf_costing_options->costing()));
```
kevinkreiser, line 559: every costing does this now, uses its enum to get a string name as a default costing name. since `ParseSharedCostOptions` is called after this, the default here will be overriden if the user has set a name themselves. all the other changes in costing files are the same as this one

src/worker.cc
```diff
@@ -942,217 +943,6 @@ void from_json(rapidjson::Document& doc, Options& options) {
 
 namespace valhalla {
 
-bool Options_Action_Enum_Parse(const std::string& action, Options::Action* a) {
```
kevinkreiser, line 945: these seemed more suited to proto_conversions.h/cc so i moved them

valhalla/proto_conversions.h
```diff
@@ -195,4 +195,24 @@ inline TripLeg_Use GetTripLegUse(const baldr::Use use) {
...
+bool ShapeMatch_Enum_Parse(const std::string& match, ShapeMatch* s);
+const std::string& ShapeMatch_Enum_Name(const ShapeMatch match);
+bool Options_Format_Enum_Parse(const std::string& format, Options::Format* f);
+const std::string& Options_Format_Enum_Name(const Options::Format match);
+const std::string& Options_Units_Enum_Name(const Options::Units unit);
+bool FilterAction_Enum_Parse(const std::string& action, FilterAction* a);
+const std::string& FilterAction_Enum_Name(const FilterAction action);
+bool DirectionsType_Enum_Parse(const std::string& dtype, DirectionsType* t);
+bool PreferredSide_Enum_Parse(const std::string& pside, valhalla::Location::PreferredSide* p);
+bool RoadClass_Enum_Parse(const std::string& rc_name, valhalla::RoadClass* rc);
```
kevinkreiser, line 216: moved these here from worker.h

valhalla/worker.h
```diff
@@ -14,23 +14,6 @@
...
-bool ShapeMatch_Enum_Parse(const std::string& match, ShapeMatch* s);
-const std::string& ShapeMatch_Enum_Name(const ShapeMatch match);
-bool Options_Format_Enum_Parse(const std::string& format, Options::Format* f);
-const std::string& Options_Format_Enum_Name(const Options::Format match);
-const std::string& Options_Units_Enum_Name(const Options::Units unit);
-bool FilterAction_Enum_Parse(const std::string& action, FilterAction* a);
-const std::string& FilterAction_Enum_Name(const FilterAction action);
-bool DirectionsType_Enum_Parse(const std::string& dtype, DirectionsType* t);
-bool PreferredSide_Enum_Parse(const std::string& pside, valhalla::Location::PreferredSide* p);
-bool RoadClass_Enum_Parse(const std::string& rc_name, valhalla::RoadClass* rc);
```
kevinkreiser, line 32: moved these to proto_conversions.h

src/tyr/route_serializer_osrm.cc
```diff
@@ -189,6 +185,7 @@ void route_summary(json::MapPtr& route, const valhalla::Api& api, bool imperial,
   route->emplace("duration", json::fp_t{duration, 3});
 
   route->emplace("weight", json::fp_t{weight, 3});
+  assert(api.options().costing_options(api.options().costing()).has_name());
```
kevinkreiser, line 188: adding this failed a bunch of tests in debug mode

valhalla/proto_conversions.h
```diff
@@ -195,4 +195,24 @@ inline TripLeg_Use GetTripLegUse(const baldr::Use use) {
       return TripLeg_Use_kRoadUse;
   }
 }
+
+// to use protobuflite we cant use descriptors which means we cant translate enums to strings
+// and so we reimplement the ones we use here. newer versions of protobuf provide these even
+// for the lite bindings so we avoid collisions by picking somewhat goofy names. an alternative
+// which would allow us to delete this completely would be to target a newer protobuf version
```
PureW, line 202: A newer protobuf  sounds worthwhile

PR #2530, opened 2020-08-14
Make parseways stage reproducible

merkispavel:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/valhalla/mjolnir/osmnode.h
+++ b/valhalla/mjolnir/osmnode.h
@@ -43,6 +43,7 @@ struct OSMNode {
   // Lat,lng of the node
   float lng_;
   float lat_;
+  uint32_t spare2_;
 
   OSMNode() {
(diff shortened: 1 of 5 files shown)


valhalla/mjolnir/osmnode.h
```diff
@@ -43,6 +43,7 @@ struct OSMNode {
   // Lat,lng of the node
   float lng_;
   float lat_;
+  uint32_t spare2_;
```
merkispavel, line 46: The only explanation is compiler ompimizes out `memset` for implicit padding. Files doesn't match without this fix

kevinkreiser: this is the right thing to do anyway, without it on a different system you might get different behavior. basically this is a bug

PR #2532, opened 2020-08-14
New Toll Gantry NodeType

kdiluca:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


valhalla/sif/dynamiccost.h
```diff
@@ -874,7 +875,8 @@ class DynamicCost {
     if (node->type() == baldr::NodeType::kBikeShare) {
       c += bike_share_cost_;
     }
-    if (node->type() == baldr::NodeType::kTollBooth || (edge->toll() && !pred->toll())) {
+    if (node->type() == baldr::NodeType::kTollBooth || node->type() == baldr::NodeType::kTollGantry ||
```
danpaz, line 878: Unlike toll booths, gantries are suspended over the road, so should we actually skip adding a cost when crossing a toll gantry? From the [osm wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtoll_gantry):

> The gantry must have no physical barrier and allow cars to drive through it unimpeded, though some might require the driver to slow down

kdiluca: @danpaz, good catch.  I agree with you.  @gknisely, what are your thoughts?

gknisely: agree should not add to the cost

gknisely (approved):

PR #2540, opened 2020-08-17
Renables gurka-tests

PureW:
that were disabled in https://github.com/valhalla/valhalla/commit/0dc4de5a8fc5dfc7fc22920c9c8a05dd4187e977


kevinkreiser:
looks like we need to move asan to a different ci task. combining that with debug/coverage checks takes so long CI times out

.circleci/config.yml
```diff
@@ -99,14 +99,14 @@ jobs:
           cd build \
             && cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=On -DCPACK_GENERATOR=DEB \
                 -DENABLE_COMPILER_WARNINGS=On -DENABLE_WERROR=Off -DCMAKE_EXPORT_COMPILE_COMMANDS=On \
-                -DENABLE_ADDRESS_SANITIZER=ON -DENABLE_PYTHON_BINDINGS=Off
+                -DENABLE_PYTHON_BINDINGS=On
       - run: make -C build -j4
       - run: make -C build utrecht_tiles
       - run: make -C build -j2 tests
       # Leak Sanitizer is disabled as it is not supported on all platforms
```
SvetlanaBayarovich, line 106: Could move this comment as well

kevinkreiser: haha good eyes!

.circleci/config.yml
```diff
@@ -133,11 +132,12 @@ jobs:
       - run: mkdir build
       - run: |
           cd build && cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=On -DENABLE_PYTHON_BINDINGS=On \
-          -DCPACK_GENERATOR=DEB -DCPACK_PACKAGE_VERSION_SUFFIX="-0ubuntu1-$(lsb_release -sc)"
-      - run: make -C build -j4
+          -DCPACK_GENERATOR=DEB -DCPACK_PACKAGE_VERSION_SUFFIX="-0ubuntu1-$(lsb_release -sc)" -DENABLE_ADDRESS_SANITIZER=ON
+      - run: make -C build -j2
```
genadz, line 136: why `-j2` ? does `-DENABLE_ADDRESS_SANITIZER=ON` have impact on memory usage while building ?

kevinkreiser: its hard to tell. CI would get OOM killed when it was j4 but previously the debug build was j4 and didnt get killed. i think potentially CIs reserve capacity ebbs and flows and this causes us to go through hard times here and there. what i do know is that we are allocated 2 cpus in CI so j4 is abusive.

merkispavel: > does -DENABLE_ADDRESS_SANITIZER=ON have impact on memory usage while building ?
yeah, it does as it additionally stores for each allocated byte whether it's valid or not(roughly)

PR #2541, opened 2020-08-18
Apply unity build to valhalla modules

SvetlanaBayarovich:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/loki/height_action.cc
```diff
@@ -10,9 +11,6 @@ using namespace valhalla::baldr;
 using namespace valhalla::skadi;
 
 namespace {
-PointLL to_ll(const valhalla::Location& l) {
```
kevinkreiser, line 13: thanks for doing this cleanup!

src/loki/isochrone_action.cc
```diff
@@ -3,15 +3,12 @@
...
 
 using namespace valhalla;
 using namespace valhalla::baldr;
 
-namespace {
-midgard::PointLL to_ll(const valhalla::Location& l) {
-  return midgard::PointLL{l.ll().lng(), l.ll().lat()};
-}
-
+namespace isochrone_action {
```
kevinkreiser, line 11: i dont think we need to name this namespace

kevinkreiser: oh i see it is exactly that to avoid collisions

src/thor/isochrone.cc
```diff
@@ -10,38 +10,9 @@ using namespace valhalla::midgard;
 using namespace valhalla::baldr;
 using namespace valhalla::sif;
 
-namespace {
-
-// Method to get an operator Id from a map of operator strings vs. Id.
-uint32_t GetOperatorId(const GraphTile* tile,
```
SvetlanaBayarovich, line 16: Not used here

src/baldr/nodeinfo.cc
```diff
@@ -10,7 +10,7 @@ using namespace valhalla::baldr;
 
 namespace {
 
-json::MapPtr access_json(uint16_t access) {
+json::MapPtr access_jsonNI(uint16_t access) {
```
SvetlanaBayarovich, line 13: Very similar to implementation in https://github.com/valhalla/valhalla/blob/bea80dc49c7199472122dd275adf65de1b784232/src/baldr/directededge.cc#L9
Probably makes sense to extract somewhere for common use (but I don' know the right place)

src/baldr/nodeinfo.cc
```diff
@@ -55,7 +55,7 @@ json::MapPtr admin_json(const AdminInfo& admin, uint16_t tz_index) {
  * @return  Returns an updated value for the bit field.
  */
 uint32_t
-OverwriteBits(const uint32_t dst, const uint32_t src, const uint32_t pos, const uint32_t len) {
+OverwriteBitsNI(const uint32_t dst, const uint32_t src, const uint32_t pos, const uint32_t len) {
```
SvetlanaBayarovich, line 58: Same https://github.com/valhalla/valhalla/blob/bea80dc49c7199472122dd275adf65de1b784232/src/baldr/directededge.cc#L32

src/thor/dijkstras.cc
```diff
@@ -12,9 +12,9 @@ using namespace valhalla::sif;
 namespace {
 
 // Method to get an operator Id from a map of operator strings vs. Id.
-uint32_t GetOperatorId(const GraphTile* tile,
-                       uint32_t routeid,
-                       std::unordered_map<std::string, uint32_t>& operators) {
+uint32_t GetOperatorIdDS(const GraphTile* tile,
```
SvetlanaBayarovich, line 15: Also same implementation as in 
https://github.com/valhalla/valhalla/blob/de546d76250402d0f37d06eebe665fe443788a71/src/thor/multimodal.cc#L14

Believe should extract somewhere (also not related to proto stuff, so not to `proto_conversions`)

valhalla/baldr/graphtile.h
```diff
@@ -613,6 +613,15 @@ class GraphTile {
     return traffic_tile;
   }
 
+  /**
+   * Convenience method to get an operator Id from a map of operator strings vs. Id
+   * @param  routeid Id of route
+   * @param  operators  Map of operator strings vs. Id
```
SvetlanaBayarovich, line 619: Not sure the description is very helpful 😞

src/baldr/nodeinfo.cc
SvetlanaBayarovich: Done

src/thor/dijkstras.cc
SvetlanaBayarovich: Done

nilsnolde:
BTW, should we activate this again? Was there anything blocking this back in the day?

Speeding up build times is always appreciated:)

kevinkreiser:
the way that it speeds up builds is by re-organizing the code so that the compiler has to do less work. basically you try to move code into less and therefor larger files (you can break them up and use compiler directives to recombine them). but the real pain is that you have to organize yourself and your project around how to make the compiler do less work. to me, just writing correct and legible code is already hard enough. reorganizing (or even keeping in mind the organization) so as to make the compilers job easier is yet another axis to optimize on and definitely secondary to performance/correctness/legibility.

i wouldnt be opposed to doing some investigation of the biggest offenders for multiple compilations or whatever and refactoring those one at a time. but juggling that against the rest of the stuff we want to do functionality-wise is an easy second place. i do however think, that maintaining organization that really adheres to this will be hard without a very very large change in the project structure which def doesnt make sense. so yeah id be interested in tools to measure this maybe even running a report when master merges or something? before we'd do that though we need to get performance measurements in place so that we can see how moving stuff (eg from inline) affects that.

TLDR; its not free, we should tread carefully

nilsnolde:
I see. Didn’t look like the most impactful changes to me, but I also have no idea what consequences it’d have to keep unity builds for the future of the project. Completely trust your assessment there, 20% time saving doesn’t justify major headaches.

kevinkreiser:
i dont mean my comments to completely discredit the idea. if we have some smaller header that can (semantically) be part of a larger one, no big deal to me to combine them (we do this from time to time). but if we really wanted to go all in, the code would be very hard to maintain. some middle ground is probably the way to go. i think just doing general clean up like less small headers where it makes sense, reducing the APIs exposed in our headers to just the stuff that is needed (we've talked about this so much), and sticking to IWYU will help the compiler and us without having to do unity explicitly or implicitly via build system functionality.

however the main gain for unity builds is when you can basically group all the stuff into one header and one massive compilation unit. if i understand the idea correctly, we could, for example, rewire midgards build to be:

midgard.hpp
```c++
// just include everything from midgard in here
#include "pointll.h"
#include "aabb.h"
#include "util.h"
...
```

and midgard.cpp
```c++
// include all the sources
#include "pointll.cc"
#include "aabb.cc"
#include "util.cc"
...
```

and over in cmake land we only tell it about those top level sources, everythings are "private" headers/sources. the maintenance burden is then to avoid the include order causing problems. things need to be declared and defined in the right order and it can easily become very annoying to have to bother with this. it would also limit the effect parallelism would have on the build when dependencies are involved, ie when every other part of the project depends on midgard, all that stuff has to wait until at least the massive single header is parsed before its compilation can begin.

however that is not what this pr is trying to do. instead its using cmake features that analyse and turn your code into a unity build behind the scenes. that can also be fine but does require constant hoop jumping (the bulk of this pr) to not break it. the same hoop jumping about order as mentioned before but now less understandable because its happening automatically by the  build system.

its also unclear what the actual output of compilation is, does it somehow mangle your headers into a jumbo header, or does that happen internally just for compilation leaving the external structure intact? if the former the end result would be close to useless as a library but it would be fine for a binary-only use-case. if the latter then the only drawback would be the hoop jumping maintenance.

TLDR; worth looking at but i wouldn't think of it as a top priority. maybe it would be useful on the less tumultuous parts of the library like midgard because we are iterating on that code as much every day.

PR #2542, opened 2020-08-19
Fix Broken C++ to PBF Conversion of Search Candidate Filtered Edges

kevinkreiser:
There is a bit of code that changes c++ objects from loki search into pbf objects to be used in the rest of the routing computation. Its a wart that should be refactored away but its a fairly decent chunk of work at the moment. So in the mean time we have a conversion function that we call. That function had a copy paste fail. See the diff


src/thor/astar.cc
```diff
@@ -432,6 +432,7 @@ void AStarPathAlgorithm::SetOrigin(GraphReader& graphreader,
             cost.cost += dest_path_edge.distance();
             cost.cost = std::max(0.0f, cost.cost);
             dist = 0.0;
+            break;
```
kevinkreiser, line 435: this loop expected (correctly) that it would never see any duplicate edge candidates so it didnt need to break once it found the matching one, but its was wasteful to keep looping over ids that it didnt need to. however because of the bug in filtering pbf marshalling, the same edge is seen twice when we relax the astarheuristic and run the route a second time. this means that that subtraction above `cost -= remainder_cost;` happens twice which makes the cost negative :womp-womp:

valhalla/baldr/pathlocation.h
```diff
@@ -161,7 +161,7 @@ struct PathLocation : public Location {
     }
 
     auto* filtered_edges = l->mutable_filtered_edges();
-    for (const auto& e : pl.edges) {
+    for (const auto& e : pl.filtered_edges) {
```
kevinkreiser, line 164: this was the copy paste fail that makes it so the same edge appears twice in the list that the astar algorithm gets. this happens when we decide to run the route again with relaxed constraints: https://github.com/valhalla/valhalla/blob/86b947c744484c6694ae81d019a322b578d224c7/src/thor/route_action.cc#L321-L322

you can see it says well somethign went wrong with the routing so we're going to allow all the edge candidates to find a better route (either because it failed or because we didnt like the ferry on the pedestrian route)

test/gurka/test_pathlocation_serialization.cc
```diff
@@ -0,0 +1,115 @@
...
+  TryLocationSerializeAndDeserialize(loc);
+}
+
+TEST_F(PathlocationSerialization, TestRadius) {
+  PointLL point;
+  PathLocation loc(point, Location::StopType::BREAK, 42, 42, 3);
+  TryLocationSerializeAndDeserialize(loc);
+}
+
+TEST_F(PathlocationSerialization, TestEdges) {
```
kevinkreiser, line 91: all of the tests but this test were moved into the gurka directory so i could test the edges part of the marshalling

PR #2545, opened 2020-08-20
add back whitelisting and unit tests for it

kevinkreiser:
long ago we accidentally removed the action whitelisting functionality from the code. this puts it back and adds unit tests for it.

reference: https://github.com/valhalla/valhalla/commit/1175a5a7a1eec98ad4b8b9b0cfb93c3a6e2416c3#diff-571fae4697ac934f38db80af7b71d54dL69

### DIFF
--- a/test/loki_service.cc
+++ b/test/loki_service.cc
@@ -3,6 +3,9 @@
 
 #include "baldr/rapidjson_utils.h"
 #include "midgard/logging.h"
+#include "proto/api.pb.h"
+#include "proto_conversions.h"
+
 #include <boost/property_tree/ptree.hpp>
 #include <prime_server/http_protocol.hpp>
 #include <prime_server/prime_server.hpp>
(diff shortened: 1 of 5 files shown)


test/loki_service.cc
```diff
@@ -327,27 +330,17 @@ const std::vector<std::pair<uint16_t, std::string>> osrm_responses{
     {400,
      R"({"code":"InvalidValue","message":"The successfully parsed query parameters are invalid."})"}};
 
-zmq::context_t context;
```
kevinkreiser, line 330: the diff looks odd here because i pulled the configuration making out into its own function. this part of the test and the tests that rely on it didnt change

test/loki_service.cc
```diff
@@ -428,7 +440,37 @@ TEST(LokiService, test_osrm_failure_requests) {
   run_requests(osrm_requests, osrm_responses);
 }
 
-// todo: test_success_requests??
+TEST(LokiService, test_actions_whitelist) {
```
kevinkreiser, line 443: this is the only real change to the test, a new test that checks getting rejected and checks getting accepted for all actions

PR #2547, opened 2020-08-20
The events/incidents in intersections

PureW:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/thor/triplegbuilder.cc
```diff
@@ -864,6 +867,8 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
     }
   }
 
+  valhalla::thor::addIncidents(*trip_edge, graphreader, edge);
```
danpaz, line 870: Should we put this behind a request parameter?

PureW: Yup, what about `incidents` which is similar to the existing `guidance_views`?

danpaz: Sounds good 👍

src/tyr/route_serializer_osrm.cc
```diff
@@ -325,6 +334,47 @@ struct IntersectionEdges {
...
+          serialized_incidents = *ptr;
+        } else {
+          throw std::logic_error("This should not be possible");
+        }
+      }
+    }
+    bool only_id = incidents_on_previous_edge.count(incident.id()) > 0;
+    auto json_incident = serialize_incident(incident, only_id);
+    serialized_incidents->emplace_back(json_incident);
+    doc.emplace("incidents", serialized_incidents);
```
danpaz, line 374: Should this line not be outside the for loop?

src/tyr/route_serializer_osrm.cc
```diff
@@ -367,6 +417,35 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
...
+      intersection->emplace("toll_collection", toll_collection_point_json);
+    }
+
+    // Mark border crossings
+    auto curr_country_code = etp->GetCountryCode(i);
+    auto prev_country_code = (i > 0) ? etp->GetCountryCode(i - 1) : "";
+    if (prev_country_code != "" && prev_country_code != curr_country_code) {
+      auto border_crossing_json = json::map({});
+      border_crossing_json->emplace("country_code", curr_country_code);
+      intersection->emplace("border_crossing", border_crossing_json);
```
danpaz, line 433: 🤔 I'm wondering if we want to simply add the country_code and state_code wherever we have it instead of a `border_crossing` attribute. It would then be up to the client to infer a border crossing when one of the codes change.

valhalla/proto_conversions.h
```diff
@@ -1,9 +1,10 @@
 #pragma once
-
 #include <valhalla/baldr/graphconstants.h>
 #include <valhalla/proto/api.pb.h>
 #include <valhalla/sif/costconstants.h>
 
+#include <random>
```
danpaz, line 6: Unused

src/thor/triplegbuilder_utils.cc
```diff
@@ -0,0 +1,17 @@
+#include "midgard/encoded.h"
+#include "proto_conversions.h"
+#include "thor/triplegbuilder.h"
+
+#include <ctime>
+#include <random>
```
danpaz, line 6: Unused

src/thor/triplegbuilder.cc
kevinkreiser: i think this, like all the other things should be an attribute filter. check https://github.com/valhalla/valhalla/blob/master/valhalla/thor/attributes_controller.h for more information :smile:

PureW: Ok, pushed a commit that makes use of the attributes-controller instead of a new proto option

src/tyr/route_serializer_osrm.cc
PureW: Yup, good catch, although it works anyway (although slightly ineffecient) due to fetching existing element from the json, as evidenced by the unittest. Pushing update.

src/tyr/route_serializer_osrm.cc
danpaz: I pushed a commit with this change https://github.com/valhalla/valhalla/pull/2547/commits/f7c93f0f5813e26c922b25388449930da0e18573

proto/trip.proto
```diff
@@ -226,6 +226,30 @@ message TripLeg {
     optional float default_speed = 44;       // km/h
     optional Restriction restriction = 45;
     optional bool destination_only = 46;
+
+    message Incident {
```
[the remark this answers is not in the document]
PureW: Yep, node makes more sense, moving it there!

valhalla/thor/attributes_controller.h
```diff
@@ -69,6 +69,7 @@ const std::string kEdgeTruckSpeed = "edge.truck_speed";
 const std::string kEdgeTruckRoute = "edge.truck_route";
 const std::string kEdgeDefaultSpeed = "edge.default_speed";
 const std::string kEdgeDestinationOnly = "edge.destination_only";
+const std::string kEdgeShowIncidents = "edge.show_incidents";
```
danpaz, line 72: I guess this will become `node.incidents`?

src/odin/enhancedtrippath.cc
```diff
@@ -1760,6 +1760,16 @@ std::string EnhancedTripLeg_Admin::ToString() const {
 
   return str;
 }
+google::protobuf::RepeatedPtrField<valhalla::TripLeg_Node_Incident> empty;
```
mandeepsandhu, line 1763: Should this made `const` since we're using it as one?

src/proto_conversions.cc
```diff
@@ -1,7 +1,52 @@
...
 
+using namespace valhalla;
+
 namespace valhalla {
 
+std::string incidentTypeToString(const TripLeg_Node_Incident& incident) {
+  std::string incident_type;
+  switch (incident.type()) {
+    case TripLeg_Node_Incident_Type_ACCIDENT:
+      incident_type = "accident";
```
mandeepsandhu, line 11: Maybe we can avoid a copy by returning directly from the `case` statements? (not sure if the compiler can do away with the string copy being created here, maybe it can?)

src/tyr/route_serializer_osrm.cc
```diff
@@ -325,6 +334,37 @@ struct IntersectionEdges {
...
+  for (const auto& incident : incidents) {
+    {
+      // Bring up any already existing array
+      auto existing = doc.find("incidents");
+      if (existing != doc.end()) {
+        if (auto* ptr =
+                boost::get<std::shared_ptr<valhalla::baldr::json::Jarray>>(&existing->second)) {
+          serialized_incidents = *ptr;
+        } else {
+          throw std::logic_error("This should not be possible");
```
mandeepsandhu, line 358: Can we put a bit more information into the error message? Like "could not get existing incidents" or some such.

src/tyr/route_serializer_osrm.cc
```diff
@@ -325,6 +334,37 @@ struct IntersectionEdges {
...
+void addsIncidents(
+    const google::protobuf::RepeatedPtrField<valhalla::TripLeg_Node_Incident>& incidents,
+    json::Jmap& doc) {
+  if (incidents.size() == 0) {
+    // No incidents, nothing to do
+    return;
+  }
+  json::ArrayPtr serialized_incidents = std::shared_ptr<json::Jarray>(new json::Jarray());
+  for (const auto& incident : incidents) {
+    {
```
mandeepsandhu, line 350: Whats the purpose of this additional scope?

src/tyr/route_serializer_osrm.cc
```diff
@@ -325,6 +334,37 @@ struct IntersectionEdges {
...
+      // Bring up any already existing array
+      auto existing = doc.find("incidents");
```
mandeepsandhu, line 352: We don't need this inside the loop, right?

src/tyr/route_serializer_osrm.cc
PureW: Keeping the local context small on the temporary variables used for bringing out any existing array in the json doc.

src/tyr/route_serializer_osrm.cc
PureW: Correct, I thought I moved this out in a previous version of the code. Anyway, moving out of the loop.

src/proto_conversions.cc
PureW: Not sure either. I'd assume the compiler could optimize it, but I can make it a direct return.

src/odin/enhancedtrippath.cc
```diff
@@ -1761,5 +1761,16 @@ std::string EnhancedTripLeg_Admin::ToString() const {
...
 }
 
+const google::protobuf::RepeatedPtrField<valhalla::TripLeg_Node_Incident> empty;
+
+const google::protobuf::RepeatedPtrField<valhalla::TripLeg_Node_Incident>&
+EnhancedTripLeg_Node::incidents() const {
+  if (mutable_node_) {
+    return mutable_node_->incidents();
+  } else {
+    return empty;
```
kevinkreiser, line 1771: im not sure if it works for repeated fields but protobuf has t his thing called default_instance. you might be able to do something like: `return google::protobuf::RepeatedPtrField<valhalla::TripLeg_Node_Incident>::default_instance;`

src/proto_conversions.cc
```diff
@@ -1,7 +1,51 @@
...
+      return "road_closure";
+      break;
+    case TripLeg_Node_Incident_Type_ROAD_HAZARD:
+      return "road_hazard";
+      break;
+    case TripLeg_Node_Incident_Type_WEATHER:
+      return "weather";
+      break;
+  };
+  return "Unhandled case";
```
kevinkreiser, line 46: throw?

src/tyr/route_serializer_osrm.cc
```diff
@@ -535,6 +602,30 @@ std::string exits(const valhalla::DirectionsLeg_Maneuver_Sign& sign) {
   return exits;
 }
 
+valhalla::baldr::json::MapPtr serialize_incident(const TripLeg_Node_Incident& incident) {
+  auto metadata_json = json::map({});
+
+  {
+    uint64_t id = static_cast<uint32_t>(incident.id());
+    metadata_json->emplace("id", id);
```
kevinkreiser, line 610: why not cast inline like the others below?

PureW: I had a really long, opaque compiler error which didn't resolve until I did this. In hindsight, it seemed like the json document wouldn't allow both uint64 and uint32 inside.

With the increase of `id` to 64bits, this entire cast goes away.

src/odin/enhancedtrippath.cc
PureW: Doesn't seem to exist for repeatedPtrField.

proto/trip.proto
```diff
@@ -277,6 +277,30 @@ message TripLeg {
...
+        ROAD_CLOSURE = 9;
+        ROAD_HAZARD = 10;
+        WEATHER = 11;
+      };
+      optional Type type = 1;
+      optional string description = 2;
+      optional uint64 start_time = 3;     // START_TIME as seconds since epoch UTC
+      optional uint64 end_time = 4;       // END_TIME seconds since epoch UTC
+      optional uint64 creation_time = 5;  // time the incident was created/updated in seconds since epoch UTC
+      optional uint32 id = 6;
```
kevinkreiser, line 301: we might want to make the id some sort of a unique information carrier for the important bits of the incident:

```
incident_id {
  uint64_t start_lon: 26; //int(lon * 100000)
  uint64_t start_lat : 26; //int(lat * 100000)
  uint64_t type: 4;
  uint64_t time: 8; //epoch second modulo 255
}
```

PureW: I like this idea. I'll implement this in the follow-on PR where the real incident data is implemented.

kevinkreiser (dismissed):

src/tyr/route_serializer_osrm.cc
danpaz: Looks like this change got reverted in a later commit 😞

danpaz: Removed in https://github.com/valhalla/valhalla/pull/2560

PR #2555, opened 2020-08-24
Shortest-ish

kevinkreiser:
This PR adds a new costing option "shortest" which tells the costing to prefer the shortest distance paths. Previously we had only the `auto_shorter` costing which kind of allowed this just for auto. This new costing option makes that costing obsolete.

Setting `"shortest": true` in your request means that all other costing related penalties are disabled and only distance is considered. This has one caveat which is that we havent used this option to disable hierarchy pruning which means that for some modes of transport, those which cull lower road classes as the algirthm progresses, the route is not gauranteed to be the absolute shortest (when ignoring shorter paths on low class roads for example).


nilsnolde:
@kevinkreiser I tried it now with a few examples on a Vienna dataset and it looks good for directions! All cases I tried were providing (visually) the same result as `auto_shorter`.

Couple of things, maybe you have some explanation for this:
- IMO the option of shortest/fastest could be first-class option in the request, not "hidden" in costing_options, as this is not weight/additional cost, but a very different way to calculate the cheapest path. I think smth like `mode: shortest/fastest` would be better for UX. But maybe there's some implementation concerns?
- matrix I can't really qualify too much.. though it _seems_ fine, all distances are maximum the same as for fastest. However it's often quite significantly shorter than `auto_shorter`. Seems like in terms of distance: auto > auto_shorter > auto:shortest
- isochrones I'm not so sure.. Here, `auto_shorter` behaves way different than `shortest: true`. While `auto_shorter` looks a lot like `auto` fastest, `shortest: true` is really off. If I'm not mistaken, there's cost metric and a stop criteria for isochrones. Cost is distance (more or less) but what metric (and/or unit) is the stop criteria (since the `contour` parameter is in minutes..)? See below for Vienna and 20 minutes(/km?):
image

yellow: ORS with shortest 20 km (and seriously generalized..)
green: auto
blue: auto_shorter
red: auto/shortest: true

parameters (though this is custom TomTom data..):
```{ "costing": "auto", "show_locations": true, "contours": [ { "time": 20 } ], "polygons": true, "id": 1, "costing_options": { "auto": { "shortest": true } }, "locations": [ { "lon": 16.387214, "lat": 48.126554, "radius": 200 } ] }```

**Comment by kevinkreiser** [comment]
@nilsnolde would you be able to find some time to help write `gurka` tests for this functionality? There is now a routing tests file in gurka where we can test some basic things like shortest

**Comment by kevinkreiser** [comment]
The isochrones make a lot of sense to me. In `auto` we favor high speed low time routes at the cost of distance, in `auto_shorter` we do that a little bit less, in `shortest` we say we dont care about time at all we just want a really short path, so we end up eating up all of our time just traveling very little distance. shortest should make isochrones that are basically circles (in dense networks)

**Comment by kevinkreiser** [comment]
with respect to making it a first class option, there is a bit of a problem. we have costing options precisely because we support multiple modes per request (only in the case of multimodal). this has kind of really screwed up our request format because it introduces a lot of complexity. anyway, if you make shortest a first class option, then we need to apply it to all costing modes which might not make sense. if you were doing multi modal where you were driving a car and then riding a bike or walking, i could see doing shortest for the biking and walking but not for the car driving. having the option be nested in the costing options for each costing type allows us to only apply it where we want it to be applied. i know we only support limited multimodal at the moment (walk/transit and walk/bike) but still if we were to make it a top level thing now we would remove the ability to control its application to whatever costing models were selected

**Comment by nilsnolde** [comment]
> would you be able to find some time to help write gurka tests for this functionality?

Yeah sure! Likely end of next week, some vacation coming up tmrw.

> The isochrones make a lot of sense to me.

Maybe I was just way too surprised by the difference in area. And confusing the cut-off criteria with the costing metrics.

A distance-based cutoff for reachability (with either shortest or fastest) could also make sense of top of `time`, e.g. for EV reachability. But that's for a different issue/PR..

**Comment by nilsnolde** [comment]
Gurka test added, other profiles should follow shortly ;)

**Comment by nilsnolde** [comment]
One thing I noticed though @kevinkreiser : if I use less extreme differences in speed for the gurka test, e.g. 40 & 80 kph for the ascii map (instead of 25 & 120 kph), also fastest takes the shortcut "ABCD" instead of the (allegedly) faster detour "ABEFCD". I guess bcs of turn costs? Shortest disregards **any** costs & penalties other than distance right?

File: test/gurka/test_shortest.cc [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+          {{"/costing_options/" + profile + "/shortest", "1"}});
+}
+
+TEST_F(ShortestTest, ScooterShortest) {
+  std::string profile = "motor_scooter";
+  doTests(profile, {"AB", "BE", "EF", "FC", "CD"},
+          {{"/costing_options/" + profile + "/shortest", "1"},
+           {"/costing_options/" + profile + "/top_speed", "120"},
+           {"/costing_options/" + profile + "/use_primary", "1.0"}});
+}
```
nilsnolde, line 91: This one still fails.. `top_speed` is not handed over to the `EdgeCost()`, it's still 45 kph.. Am I doing smth wrong?

test/gurka/test_shortest.cc
```diff
@@ -0,0 +1,98 @@
...
+
+TEST_F(ShortestTest, BikeShortest) {
+  std::string profile = "bicycle";
+  doTests(profile, {"AB", "BG", "GH", "HC", "CD"},
+          {{"/costing_options/" + profile + "/shortest", "1"},
+           {"/costing_options/" + profile + "/use_roads", "1.0"}});
+}
```
nilsnolde, line 98: Also fails.. Not entirely sure how to properly test it either.. Any suggestion?

test/gurka/test_shortest.cc
nilsnolde: argh.. had some stuff stashed.. just realized when half the profiles were missing.. one sec

nilsnolde: yeah, no that didn't help, still failing bcs `top_speed` is not 120 but 45

test/gurka/test_shortest.cc
```diff
@@ -0,0 +1,98 @@
...
```
nilsnolde, line 98: Would appreciate input on testing bike & pedestrian. Not sure if we need those profiles actually, but for completeness sake maybe (?)

src/sif/pedestriancost.cc
```diff
@@ -633,11 +633,16 @@ bool PedestrianCost::AllowedReverse(const baldr::DirectedEdge* edge,
...
+  if (shortest_) {
+    float sec = edge->length() * (kSecPerHour * 0.001f) / static_cast<float>(speed);
+    return Cost(edge->length(), sec);
+  }
+  
   // Ferries are a special case - they use the ferry speed (stored on the edge)
   if (edge->use() == Use::kFerry) {
-    auto speed = tile->GetSpeed(edge, flow_mask_, seconds);
-    float sec = edge->length() * (kSecPerHour * 0.001f) / static_cast<float>(speed);
+    float sec = edge->length() * (kSecPerHour * 0.001f) / static_cast<float>(edge->speed());
```
nilsnolde, line 645: seemed wrong here

src/sif/motorscootercost.cc
```diff
@@ -430,14 +430,22 @@ Cost MotorScooterCost::EdgeCost(const baldr::DirectedEdge* edge,
 
   if (edge->use() == Use::kFerry) {
     assert(speed < speedfactor_.size());
-    float sec = (edge->length() * speedfactor_[speed]);
+    // Use the edge speed (should be the speed of the ferry)
+    float sec = (edge->length() * speedfactor_[edge->speed()]);
```
nilsnolde, line 434: same as in pedestrian

nilsnolde:
Added the other profiles and updated the tests @kevinkreiser . A few suggestions on motorscooter/bike/ped tests would be appreciated, maybe you have a smart idea. You don't show up in the reviewer list for some reason..

test/gurka/test_shortest.cc
nilsnolde: dup

nilsnolde:
Oh crap.. accidentally pulled in all the MinGW changes.. I'll revert.

nilsnolde:
Resolved conflicts and commented out non-working test code. If you have a grand idea what's a edge config for testing scooter/bike/ped @kevinkreiser I'm all ears. Or if you think we can survive without testing those profiles, this would be ready for final review.

kevinkreiser:
@nilsnolde yep ill get around to adding more testing for this shortly for those other costings we need to think more about what makes a lower cost for them. for ped its walkways vs alleyways, for bike its dedicated bike paths vs those with cars on them, and for scooters i need to review the code haha

nilsnolde:
good to go now as well @kevinkreiser was a little challenging to fit all these profiles on the small test map, but a wild combo of osm tags somehow worked:)

nilsnolde:
let me fix the debug lint build

docs/api/turn-by-turn/api-reference.md
```diff
@@ -140,6 +143,7 @@ These additional options are available for bicycle costing methods.
...
 |`bss_return_cost`| This value is useful when `bikeshare` is chosen as travel mode. It is meant to give the time will be used to return a rental bike. This value will be displayed in the final directions and used to calculate the whole duation. The default value is 120 seconds.|
 |`bss_return_penalty`| This value is useful when `bikeshare` is chosen as travel mode. It is meant to describe the potential effort to return a rental bike. This value won't be displayed and used only inside of the algorithm.|
+| `shortest` | Changes the metric to quasi-shortest, i.e. purely distance-based costing. Note, this will disable all other costings & penalties. Also note, `shortest` will not disable hierarchy pruning, leading to potentially sub-optimal routes for some profiles. The default is `false`. |
```
kevinkreiser, line 146: `profiles` is an osrm term we should use the word `costing models` here

src/sif/bicyclecost.cc
```diff
@@ -566,18 +566,22 @@ Cost BicycleCost::EdgeCost(const baldr::DirectedEdge* edge,
                            const baldr::GraphTile* tile,
                            const uint32_t seconds) const {
   auto speed = tile->GetSpeed(edge, flow_mask_, seconds);
+  assert(speed < speedfactor_.size());
+  float sec = (edge->length() * speedfactor_[speed]);
```
kevinkreiser, line 570: unfortunately these changes seem to be a bit problematic in that they return the wrong number of seconds along the edge if its one of the cases below (steps or ferry). seems we need to maybe just put a ternary in each return for the cost

src/sif/motorscootercost.cc
```diff
@@ -426,15 +426,23 @@ Cost MotorScooterCost::EdgeCost(const baldr::DirectedEdge* edge,
   auto speed = tile->GetSpeed(edge, flow_mask_, seconds);
 
   if (edge->use() == Use::kFerry) {
-    assert(speed < speedfactor_.size());
-    float sec = (edge->length() * speedfactor_[speed]);
+    float ferry_speed = static_cast<float>(edge->speed());
+    assert(ferry_speed < speedfactor_.size());
+    float sec = (edge->length() * speedfactor_[ferry_speed]);
     return {sec * ferry_factor_, sec};
```
kevinkreiser, line 432: missed a spot here

src/sif/pedestriancost.cc
kevinkreiser: i dont think its wrong actually. the get speed function will eventually return edge->speed, it just makes sure to check if there isnt any higher resolution speed information on the edge. i think it should probably stay as it was in case in the future we get better information about ferries and their speeds at different times of day (unlikely but would remove a gotcha)

kevinkreiser:
I think to get this merge we want to do 2 more things:

1. We want to actually remove `auto_shorter` from the docs and write a tiny bit of code that when we see auto shorter we turn it into an `auto` with `shortest` turned on
2. We need to benchmark the algorithm with shortest turned on to see that its not different in terms of performance characteristics. I dont suspect it to be because we dont turn off heirarchy culling but its just something we should check to be sure we dont need to add another service limit conifugration for these types of routes

src/sif/bicyclecost.cc
nilsnolde: thx, I'll have a look tmrw morning at this logic again

nilsnolde: Oh hell yeah, there was actually a lot wrong in ped/bike costings.. Must've been a serious case of copy/paste dumbness. I went over all costings again and fixed mostly ped/bike

src/sif/motorscootercost.cc
nilsnolde: nonsense

src/sif/pedestriancost.cc
nilsnolde: gotcha, reviewed the other ferry costs as well and reverted what I did there

docs/api/turn-by-turn/api-reference.md
nilsnolde: oops, too ORS biased..

src/sif/bicyclecost.cc
```diff
@@ -592,6 +586,20 @@ Cost BicycleCost::EdgeCost(const baldr::DirectedEdge* edge,
...
+    float sec = (edge->length() * speedfactor_[speed]);
+    return Cost(edge->length(), sec);
+  }
+
+  // Stairs/steps - high cost (travel speed = 1kph) so they are generally avoided.
+  if (edge->use() == Use::kSteps) {
+    float sec = (edge->length() * speedfactor_[1]);
+    return {sec * kBicycleStepsFactor, sec};
+  }
+
```
nilsnolde, line 602: rearranged a little so `bike_speed` won't be uselessly computed if there's a ferry. Also kept all duration-relevant conditions, was a serious oversight before on my part..

src/sif/pedestriancost.cc
```diff
@@ -435,6 +435,9 @@ class PedestrianCost : public DynamicCost {
                                  const uint32_t idx) const override {
     // Cases with both time and penalty: country crossing, ferry, gate, toll booth
     sif::Cost c;
+    if (shortest_) {
+      return c; // shortest ignores any penalties in favor of path length
+    }
```
nilsnolde, line 440: not sure about this one @kevinkreiser.. from the description it sounds like it should be added here as well for ped part of multimodal costing?

src/sif/pedestriancost.cc
```diff
@@ -630,14 +633,21 @@ bool PedestrianCost::AllowedReverse(const baldr::DirectedEdge* edge,
 Cost PedestrianCost::EdgeCost(const baldr::DirectedEdge* edge,
                               const baldr::GraphTile* tile,
                               const uint32_t seconds) const {
+  auto speed = tile->GetSpeed(edge, flow_mask_, seconds);
```
nilsnolde, line 636: I'll revert this one again, doesn't make sense..

src/worker.cc
```diff
@@ -769,6 +769,17 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+  // shortest=true costing option. maybe remove auto_shortest in v4?
+  if (costing_str == "auto_shorter") {
+    costing_str = "auto";
+    auto autoshorter_options = rapidjson::GetValueByPointer(doc, "/costing_options/auto_shorter");
+    if (autoshorter_options) {
+      rapidjson::SetValueByPointer(doc, "/costing_options/auto", *autoshorter_options);
+    }
+    rapidjson::Pointer("/costing_options/auto/shortest").Set(doc, true);
+  }
+
```
nilsnolde, line 782: changes costing to `auto` if `auto_shorter` was specified in the request.. should also make sure that all costing options are transferred appropriately

test/auto_shorter_deprecated.cc
```diff
@@ -0,0 +1,67 @@
...
+  Api request;
+  std::string request_str =
+      R"({"locations":[{"lat":52.114622,"lon":5.131816},{"lat":52.048267,"lon":5.074825}],"costing":"auto_shorter"})";
+  ParseApi(request_str, Options::route, request);
+
+  EXPECT_EQ(Costing_Enum_Name(request.options().costing()), "auto");
+  EXPECT_EQ(request.options().costing_options(static_cast<int>(auto_)).shortest(), true);
+}
+
+TEST(AutoShorterDeprecated, test_auto_shorter_shortest_false) {
```
nilsnolde, line 36: test name is stupid. same below. will change before it'll be merged

nilsnolde:
> We want to actually remove auto_shorter from the docs and write a tiny bit of code that when we see auto shorter we turn it into an auto with shortest turned on

should be mostly done now @kevinkreiser .

> We need to benchmark the algorithm with shortest turned on

I can do that, but I can't really imagine how this could possibly be less performant as it's mostly skipping cost updates. If anything I'd imagine this to be tiny weeny little bit faster, no?

`gurka.test_traffic.cc:BasicUpdates` is failing for some reason.. can't see how this PR changes anything related to that test tbh..

nilsnolde:
Also, we either have to keep most of the auto_shorter code (in `auto_cost.cc` etc) or remove it also from `proto_conversions.cc:Costing_Enum_Parse` to get rid of the `switch` error in the CI builds. Not sure what other surprises the latter option yields.. Just keep the code?

src/worker.cc
nilsnolde: all critical cases are tested BTW

kevinkreiser:
@nilsnolde ill remove auto_shorter i have a decent idea how we can make this pretty straight forward to deprecate but not lose support for old clients that are makign use of it.

src/sif/bicyclecost.cc
kevinkreiser: unfortunately there are still some missed spots. if its a ferry its going to bail early and ignore shortest. also at the very end of the function (ie the normal case doesnt consider shortest). remember that for shortest we want to retain the original time estimate (`Cost::sec`) but we just want to set the cost part (`Cost::cost`) to edge length. i think the easiest way to understand that change would be a ternary in every return. i'll re-organize it and you can have a look to see what i mean

src/sif/bicyclecost.cc
```diff
@@ -750,7 +758,7 @@ Cost BicycleCost::TransitionCost(const baldr::DirectedEdge* edge,
   penalty *= (bike_accom * avoid_roads) + use_roads_;
 
   // Return cost (time and penalty)
-  c.cost += (seconds * (turn_stress + 1.0f)) + penalty;
+  c.cost += shortest_ ? 0.f : (seconds * (turn_stress + 1.0f)) + penalty;
```
kevinkreiser, line 761: this is also not quite right. above the `base_transition_cost` function will have set the cost to something other than 0. but if we are doing shortest we need it to always be 0. actually thats probably a good thing to check in the funciton and set to 0. let me do that

kevinkreiser: haha me from 2 months ago tried to do that but made a mistake... let me fix my mistake :smile:

test/auto_shorter_deprecated.cc
kevinkreiser: i suggest we remove this whole test suite and instead just add the `ParseApi` part as individual test cases in your other new test file that is specific for testing shortest. that way we can test that the parsing override works and gives us an `auto` cost with `shortest` set to true. let me see if i can do that quickly

src/sif/bicyclecost.cc
nilsnolde: :facepalm:

src/worker.cc
```diff
@@ -769,6 +769,34 @@ void from_json(rapidjson::Document& doc, Options& options) {
 
   // costing defaults to none which is only valid for locate
   auto costing_str = rapidjson::get<std::string>(doc, "/costing", "none");
+
+  // auto_shorter is deprecated and will be turned into
+  // shortest=true costing option. maybe remove in v4?
+  if (costing_str == "auto_shorter") {
+    costing_str = "auto";
+    rapidjson::Pointer("/costing").Set(doc, "auto");
```
nilsnolde, line 777: nit: consistent use of `rapidjson` shortcut methods, i.e. `SetValueByPointer`. I'll change that quickly

nilsnolde (approved):
looks great thanks!!

nilsnolde (dismissed):
...

nilsnolde (approved):
LGTM

PR #2556, opened 2020-08-24
Multicue enter roundabout

dgearhart:
# Issue

Originally thought that the having the enter roundabout maneuver part of a multi-cue would be too much info - seems like it is needed.
For example: Turn right onto Main Street. Then, Enter the roundabout and take the 3rd exit onto 1st Avenue.

## Tasklist

 - [x] Add tests


danpaz:
Will you add a test? Would a gurka test would work for this?

test/gurka/test_instructions_roundabout_multicue.cc
```diff
@@ -0,0 +1,198 @@
...
+  gurka::assert::raw::expect_maneuvers(result, {DirectionsLeg_Maneuver_Type_kStart,
+                                                DirectionsLeg_Maneuver_Type_kRoundaboutEnter,
+                                                DirectionsLeg_Maneuver_Type_kRoundaboutExit,
+                                                DirectionsLeg_Maneuver_Type_kDestination});
+  int maneuver_index = 0;
+
+  // Verify the start then enter roundabout multicue
+  gurka::assert::raw::expect_instructions_at_maneuver_index(
+      result, maneuver_index, "Drive east on Main Street/A 1.", "",
+      "Drive east on Main Street, A 1. Then Enter the roundabout and take the 3rd exit.",
```
mandeepsandhu, line 83: Should `Enter` be capitalized? like `Drive east on Main Street, A 1. Then enter the roundabout ...` ?

test/gurka/test_instructions_roundabout_multicue.cc
```diff
@@ -0,0 +1,198 @@
...
+  gurka::assert::raw::
+      expect_instructions_at_maneuver_index(result, ++maneuver_index,
+                                            "Enter the roundabout and take the 3rd exit.",
+                                            "Enter the roundabout and take the 3rd exit.",
+                                            "Enter the roundabout and take the 3rd exit.", "");
+
+  // Verify the exit_roundabout instructions
+  gurka::assert::raw::expect_instructions_at_maneuver_index(
+      result, ++maneuver_index, "Exit the roundabout.", "",
+      "Exit the roundabout. Then You will arrive at your destination.", "Continue for 200 meters.");
```
mandeepsandhu, line 96: Looks like we do a simple concatenation of the text instruction with the `Then` phrase. We should probably look into make the first letter lower-case. Need not be done now though.

dgearhart: since it is for voice and not display - did not think it was worth it. I do not hear any issues with using it this way

mandeepsandhu: Aren't the text instructions shown somewhere? I guess we can revisit it later if someone reports an issue.

PR #2557, opened 2020-08-24
Merge valhalla updates

austinhess:
Includes upstream fixes since May which might solve some routing issues.


austinhess:
Clicked wrong base branch; delete.

PR #2558, opened 2020-08-25
map matching uturn trimming fix

### LINKED ISSUE #2552: Major erroneous doubling back of route produced by trace_route
While trying to use `/trace_route` for a recorded pedestrian GPS track along a section of the John Muir Trail in CA, the output included a large erroneous segment of doubling back that did not reflect any component of the track. The GPS data itself is fairly clean and doesn't reverse its course, and simply heads continuously south along the trail. But in the routed geometry,   at a point near an intersection at 119.478699 W, 37.757385 N, there is a U-turn instruction with directions to head back north for about 7 km, then another instruction to go back 7 km south approximately back to that same point, and then the route continues south in accordance with the actual track geometry.

I attempted to use `turn_penalty_factor` and other `trace_options` parameters as described in the [map matching docs](https://valhalla.readthedocs.io/en/latest/api/map-matching/api-reference/), but not only did they not fix the issue, they didn't change the route output at all. The only thing which seemed to resolve the doubling back was to truncate the track before it reached the intersection mentioned above. While I understand there are issues with doubling back when there is significant jitter due to GPS, this example seems particularly confusing because there is virtually no such variation in the recorded GPS data, the routing output is insensitive to the turn penalty factor, and the doubled-back segments of the route are very long.

Here is a sample curl command to reproduce the issue:
```
curl --location --request POST '<ROUTING SERVER URL>' \
--data-raw '{"encoded_polyline": "wssdgAx_lwbF?NfC{@bGgEbQ}TxG{KdE}TbBcFxBgDhMsF??rAm@dTjAfm@wC|EvCfCtEfEdE~HzAhHfDtD|J|J|IpB~HlJ`I~HtNx\\zLvNpG|JhCjFl@rFjAjK~HzL|TjAbGrFxLjP~RzG\\bVvObe@zUpQ`GnOvD`Bz@tE^|OtD|TlKvIrF`G_@bQ`HlPjB~WzKnXlJbQxBbeAr[dTfCfJxBv]vDv_BhChHyB`b@LxLm@fYuNjF?~h@yBrKiCfr@\\r[rFfh@jAhWfOtItD??`NbG~HzApLpGbLrFz[lU`LbQfOdOxL~IbLnH|OdE~W|JtO|JtUhMdEtDjF|J`XbQb[dOpRhMnNrF~H~I??xQ|TlKxLxGdPhHhLdEnTnHdEd@tEe@~G~CxCtO?jBfCkBtFoNnS_D`Rd@zKfEdO~HbG~R|JbLdD`b@~h@l_@n]nM\\rLtOzJfE`DrE`LbGfEfNj@`HbBrFd@|J|@hC?`G`BfEdj@j~@dEjBdEl@|OrEhMrGjdAp\\pWvCbLm@nS~Hx[bGnSOz[_I~HNjKhCbjAOlPtEdwAxB??`Xl@dJ`GbV^dOxA`N?yC?zF?pv@fD??bG^zExAfJ^`LjA|JtDvN^hRgD|Y{ApLyB~^}@~Mm@vX{JbG}@nNsFhRiC~HeErPsP|JqGtEyBxQgEb[^nSiC~HO~CgCbGwD`SiC~\\~I`SpH`G~HfOvb@nCbQtE~RpHde@LrE~ClK|JjVxBjUbBfOdEzJvDhC~CNfC_@pHk@pBz@|E|J????xBpGpGhCvXbF|^eO~H}JfEeOhRgNjZ`G`I{@xLhBxLvNjFiC`InS`L`ItPn]`GrPb[hb@zf@lh@rFxMfCvDtP}@dOl@bLnH|TsE??hWuEtm@yBf^bF`SgDrKhCrj@uEvIvCli@xC~RpGhMgDrFfDnNzAxLjVha@tOd[vNxVb[nSvl@hWlrApH~h@be@b`Dl_@j`@~CrQ~M~R??f@|@rKvCnl@dFvw@n]zQsFxL`H|TnIpQOhMhC|JpGbG|J|c@xLdEdEvD{AzE}TbB]bQ|^hH{UfM~]`Nuc@nHtc@rGeOrFpQdJ{KvIz`@pLeP|E~ShHcGz@|@oDhv@jAzJfEpHzt@vc@nShVb[d[~RzK|EdE??pg@pf@tSxa@hCxLhf@p{@zVdE`b@b[ptApfApl@n]x[vCnSaGp\\xApg@tE\\?yGm@zFl@lUm@dh@yBnOoHhu@dd@`SnSbVpf@bQngAkBtlAqGp{@oCzu@e@tc@tInh@fO`]xGtYt^vv@pCvXjd@bdB`SdOjUboA??rApGyBxb@_Inq@lKxv@sGlgBaGfrD`Bfx@hCzj@|TfiC~MnIxp@bxBpH`]????nDdErKxBnIrFlJxMtEOlOyL????yBnRfCrQzGrPzOxMpXz@tIjLv^b[zj@bz@nSz`@zOvNhM`G??vDxBha@fOvDm@~MsQjFcFfc@{AdJjK`IjB~HwDxLzA`SpH~MvX`g@fw@pGha@`SfYrP~{@tJfm@e@fY??Wvb@~Cli@vDp\\|OvYbLvXjPvaA|OhWxLlTbBxLWbQ~ClKvNzKfTfN|Y|TtEnI~Hj_@`GfNbVpS`Xxu@~RlTvNlUrL|Tpa@fm@pBjK}@hMgCvDcBzJ|@tPW~g@gCpQiMjVsFlT?jWaItOgHnRdEjWxB~RpBxL~HlKb[zUlE`H`D`HWbQcLxLcQpGuNvDoDzANdDrFdF\\jA", "costing": "pedestrian", "shape_match": "map_snap", "use_timestamps": false, "trace_options": {"turn_penalty_factor": 500}}'

### PR DESCRIPTION
**Pull request by kevinkreiser**
when a uturn is detected in a map match we need to trim the edges shape where the uturn occurs. we recently fixed code that always set trimming information even when it wasnt needed. that fix assumed that the downstream code actually cared to whether or not that boolean telling it to trim was set. this was not the case..

this may fix #2552

### DIFF
--- a/src/thor/route_action.cc
+++ b/src/thor/route_action.cc
@@ -75,7 +75,7 @@ void via_discontinuity(
     // Insert a second discontinuity so the next (opposing) edge is trimmed at the end from
     // 1-dist along to 1
     vias.insert({path_index + (flip_index ? 0 : 1),
-                 {{false, snap_ll, 1.0f - dist_along}, {true, PointLL(), 1.0f}}});
+                 {{true, snap_ll, 1.0f - dist_along}, {false, PointLL(), 1.0f}}});
   }
 }
 
--- a/src/thor/triplegbuilder.cc
+++ b/src/thor/triplegbuilder.cc
@@ -135,6 +135,7 @@ void SetShapeAttributes(const AttributesController& controller,
                                   [distance_total_pct](const decltype(speeds)::value_type& s) {
                                     return distance_total_pct <= std::get<0>(s);
                                   });
+    assert(speed_itr != speeds.cend());
     for (auto i = shape_begin + 1; i < shape.size(); ++i) {
       // If there is a change in speed here we need to make a new shape point and continue from
       // there
@@ -1543,6 +1544,10 @@ void TripLegBuilder::Build(
         edge_begin_info.trim = true;
         edge_begin_info.distance_along = start_pct;
         edge_begin_info.vertex = start_vrt;
+      } // No trimming needed
+      else if (!edge_begin_info.trim) {
+        edge_begin_info.distance_along = 0;
+        edge_begin_info.vertex = edge_shape.front();
       }
 
       // Handle partial shape for last edge
@@ -1550,6 +1555,10 @@ void TripLegBuilder::Build(
         edge_end_info.trim = true;
         edge_end_info.distance_along = end_pct;
         edge_end_info.vertex = end_vrt;
+      } // No trimming needed
+      else if (!edge_end_info.trim) {
+        edge_end_info.distance_along = 1;
+        edge_end_info.vertex = edge_shape.back();
       }
 
       // Overwrite the trimming information for the edge length now that we know what it is
@@ -1560,12 +1569,13 @@ void TripLegBuilder::Build(
       auto edge_length = static_cast<float>(directededge->length());
       trim_shape(edge_begin_info.distance_along * edge_length, edge_begin_info.vertex,
                  edge_end_info.distance_along * edge_length, edge_end_info.vertex, edge_shape);
-      // Add edge shape to trip
-      trip_shape.insert(trip_shape.end(),
-                        (edge_shape.begin() + ((edge_begin_info.trim || is_first_edge) ? 0 : 1)),
+      // Add edge shape to the trip and skip the first point when its redundant with the previous edge
+      // TODO: uncommment correct removal of redundant shape after odin can handle uturns
+      // trip_shape.insert(trip_shape.end(), edge_shape.begin() + !is_first_edge, edge_shape.end());
+      trip_shape.insert(trip_shape.end(), edge_shape.begin() + !edge_begin_info.trim,
                         edge_shape.end());
 
-      // If edge_begin_info.exists and is not the first edge then increment begin_index since
+      // If edge_begin_info.trim and is not the first edge then increment begin_index since
       // the previous end shape index should not equal the current begin shape index because
       // of discontinuity
       if (edge_begin_info.trim && !is_first_edge) {
--- a/test/gurka/test_match.cc
+++ b/test/gurka/test_match.cc
@@ -1,4 +1,6 @@
 #include "gurka.h"
+#include "midgard/encoded.h"
+#include "midgard/util.h"
 #include <gtest/gtest.h>
 
 using namespace valhalla;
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: src/thor/triplegbuilder.cc [hunk 1]
```diff
@@ -1543,13 +1544,21 @@ void TripLegBuilder::Build(
         edge_begin_info.trim = true;
         edge_begin_info.distance_along = start_pct;
         edge_begin_info.vertex = start_vrt;
+      } // No trimming needed
+      else if (!edge_begin_info.trim) {
+        edge_begin_info.distance_along = 0;
+        edge_begin_info.vertex = edge_shape.front();
```
kevinkreiser, line 1550: when we are not trimming either the front or the end we need to make sure we still set these things because they are used regardless below in the shape trimming function

src/thor/triplegbuilder.cc
```diff
@@ -1560,12 +1569,13 @@ void TripLegBuilder::Build(
       auto edge_length = static_cast<float>(directededge->length());
       trim_shape(edge_begin_info.distance_along * edge_length, edge_begin_info.vertex,
                  edge_end_info.distance_along * edge_length, edge_end_info.vertex, edge_shape);
-      // Add edge shape to trip
-      trip_shape.insert(trip_shape.end(),
-                        (edge_shape.begin() + ((edge_begin_info.trim || is_first_edge) ? 0 : 1)),
+      // Add edge shape to the trip and skip the first point when its redundant with the previous edge
+      // TODO: uncommment correct removal of redundant shape after odin can handle uturns
+      // trip_shape.insert(trip_shape.end(), edge_shape.begin() + !is_first_edge, edge_shape.end());
+      trip_shape.insert(trip_shape.end(), edge_shape.begin() + !edge_begin_info.trim,
```
kevinkreiser, line 1575: the commented out line above is the correct code and causes no duplicates to be had in the shape. however at uturns, it seems that odin does not detect the uturn has occured unless the shape point is doubled where the uturn occurs. cc @dgearhart

test/gurka/test_match.cc
```diff
@@ -105,4 +107,17 @@ D--3--4--C--5--6--E)";
...
+  // TODO: Remove the duplicate 6 when we fix odin to handle uturn maneuver generation with only one
+  // turn around point
+  auto expected_shape = decltype(shape){
+      map.nodes["2"], map.nodes["B"], map.nodes["C"], map.nodes["6"],
+      map.nodes["6"], map.nodes["C"], map.nodes["3"],
+  };
+  EXPECT_EQ(shape.size(), expected_shape.size());
+  for (int i = 0; i < shape.size(); ++i) {
+    EXPECT_TRUE(shape[i].ApproximatelyEqual(expected_shape[i]));
+  }
```
kevinkreiser, line 122: this test case showed the above behavior so we simply expanded it to check the actual shape was what we thought it should be (this is how we found out that the odin bug with uturn maneuvers exists). we decided not to fix the odin uturn bug at this time and instead opted to leave the extra shape point in (at the uturn point) so that we get the correct maneuver

src/thor/route_action.cc
```diff
@@ -75,7 +75,7 @@ void via_discontinuity(
     // Insert a second discontinuity so the next (opposing) edge is trimmed at the end from
     // 1-dist along to 1
     vias.insert({path_index + (flip_index ? 0 : 1),
-                 {{false, snap_ll, 1.0f - dist_along}, {true, PointLL(), 1.0f}}});
+                 {{true, snap_ll, 1.0f - dist_along}, {false, PointLL(), 1.0f}}});
```
kevinkreiser, line 78: this was a long time copy paste fail..

PR #2564, opened 2020-08-27
nit: Adds missing incident-string

PureW:
How I miss that we don't use `-Wall` and `-Werror`.


PureW:
I was able to use `set_source_files_properties` to enable `-Wall` and `-Werr` on `proto_conversions.cc` which catches this missing switch branch.

danpaz (dismissed):
LGTM, although windows build failed with:

```
cl : command line error D8021: invalid numeric argument '/Werror' [D:\a\build\src\valhalla.vcxproj]
##[error]Cmd.exe exited with code '1'.
```

PR #2568, opened 2020-08-28
Ensure edge.is_urban is set for serializing is_urban

danpaz:
# Issue

#2522 introduced a new is_urban flag to the osrm response, which uses the edge's `density` field to infer whether a road is considered to be urban or not. However, as it's an optional field, in the serializer we should be checking `edge.has_density` to ensure it has been populated before adding it to the response.

This also allows excluding is_urban flag from the response if you don't want it, by setting the exclude filter attribute for edge.density.
 

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/test/urban.cc
+++ b/test/urban.cc
@@ -96,7 +96,6 @@ TEST(Urban2, test_urban) {
   auto json_str = serializeDirections(response);
   rapidjson::Document json;
   json.Parse(json_str);
-  std::cout << "json :: " << json_str << std::endl;
 
   ASSERT_FALSE(json.HasParseError());
 
@@ -126,7 +125,6 @@ TEST(Urban2, test_urban_excluded_by_default) {
   auto json_str = serializeDirections(response);
   rapidjson::Document json;
   json.Parse(json_str);
-  std::cout << "json :: " << json_str << std::endl;
 
   ASSERT_FALSE(json.HasParseError());
 
--- a/test/urban.cc
+++ b/test/urban.cc
@@ -115,6 +115,34 @@ TEST(Urban2, test_density) {
   }
 }
 
+TEST(Urban2, test_density_excluded) {
+  bool is_urban;
+  route_tester tester;
+  std::string request =
+      R"({"locations":[{"lat":52.10160225589803,"lon":5.116925239562988},{"lat":52.09403591712907,"lon":5.113234519958496}],"costing":"auto","units":"miles","format":"osrm","filters":{"action":"exclude","attributes":["edge.density"]}})";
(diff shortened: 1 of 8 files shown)


test/urban.cc
```diff
@@ -115,6 +115,34 @@ TEST(Urban2, test_density) {
...
+  route_tester tester;
+  std::string request =
+      R"({"locations":[{"lat":52.10160225589803,"lon":5.116925239562988},{"lat":52.09403591712907,"lon":5.113234519958496}],"costing":"auto","units":"miles","format":"osrm"})";
+  auto response = tester.test(request);
+
+  // get the osrm json
+  auto json_str = serializeDirections(response);
+  rapidjson::Document json;
+  json.Parse(json_str);
+  std::cout << "json :: " << json_str << std::endl;
```
mandeepsandhu, line 129: Leftover from debugging?

danpaz:
After discussing with @kevinkreiser we decided to add a new filter attribute edge.is_urban rather than relying on the existing edge.density filter attribute. This allows us to make the information is_urban flag opt-in and won't have the unintended consequence of adding multiple properties to the tripleg with a single filter attribute.

cc @kdiluca @gknisely

test/urban.cc
danpaz: Yep good catch

test/urban.cc
```diff
@@ -115,6 +115,34 @@ TEST(Urban2, test_density) {
...
+  ASSERT_FALSE(json.HasParseError());
+
+  // loop over all routes all legs
+  for (const auto& route : json["routes"].GetArray()) {
+    for (const auto& leg : route["legs"].GetArray()) {
+      for (const auto& step : leg["steps"].GetArray()) {
+        for (const auto& intersection : step["intersections"].GetArray()) {
+          EXPECT_EQ(intersection.HasMember("is_urban"), false);
+        }
+        EXPECT_EQ(is_urban, true);
```
mandeepsandhu, line 140: What sets `is_urban` in this test? It seems to be left uninitialized.

danpaz: Leftover from rearranging the tests, thanks for catching 😄

test/urban.cc
```diff
@@ -107,9 +106,35 @@ TEST(Urban2, test_density) {
...
           }
         }
-        EXPECT_EQ(is_urban, true);
+      }
+    }
+  }
+}
+
+TEST(Urban2, test_urban_excluded_by_default) {
+  bool is_urban;
```
mandeepsandhu, line 118: I guess we don't need this anymore?

PR #2573, opened 2020-09-02
Incident Loading Singleton

kevinkreiser:
What is this
----------------

This pr adds the ability for valhalla to access incident information from protobuf tiles that are located in a configured directory. Much like graph tiles. The idea here is that we have incident tiles which will need to be continually refreshed from disk. There are a couple of assumptions that we are making:

1. its unlikely that incidents change frequently
2. its unlikely that there are large numbers of incidents
3. incidents will not affect routing algorithms computations

Then we have a singleton whose job it is to give access to the available incident tiles from the directory. Access is used in triplegbuilder to associate incidents to the pbf route leg. The way these incidents are accessed is via binary search. The speed tiles will take one of the reserved bits to  say that a given record has an incident. Upon seeing this bit we will go look up the incident from the tile.

How are incidents refreshed
---------------------------------------

This deserves an explanation because its the trickier part. The model we use here is a class with a private constructor and one public static function. that function statically initializes a singleton instance of this class. the class's constructor spawns a thread that runs in the background. that thread is responsible for monitoring the filesystem for new incidents. when the thread is first started, the singleton waits for it to initialize (load all the incidents that are available).

the thread and the singleton instance communicate over the thread barrier by sharing a state object. The bulk of communications use atomic booleans (which are hopefully lock free on most implementations). you'll notice a mutex in there as well. the mutex is only used for synchronization of the tile cache which is an unordered_map. synchronization is needed because the unordered_map may reallocate when new tiles are added to it. thus this synchronization is also conditional. when the tileset is static, ie when its a mmapped tar, we do not need to use the mutex to synchronize and instead preallocate all the slots needed for the cache. the shared state object is wrapped in a shared_ptr, from the code comment:

>   // we use a shared_ptr to wrap the state between the watcher thread and the main threads singleton
>   // instance. this gives the responsibility to the last living thread to deallocate the state object.
>   // if we didn't do this, and the watcher thread were still running when the singleton instance got
>   // destructed, then the watcher would be making use of a deallocated state object. this way, if the
>   // watcher is last to die it own the lifetime of the state and if the singleton is the last to die
>   // it owns the lifetime of the state. note that we still need to use atomics inside the state as
>   // only the shared_ptr itself is thread safe, not the thing it points to
 
There are two modes for the incident loading singleton, one which does directory scans (`mjolnir.incident_dir` in the config), which on a modern ssd where changes are happening to the incident directory, takes 15 seconds for a planets worth of incident tiles. The second mode is a memory mapped log file which tells the timestamp when an incident tile was last changed (rather than using mtime of the files on the filesystem). This can be configured with the `mjolnir.incident_log` config option and takes generally subsecond on modern ssds to complete for updates.

There are also two other configuration options to do with error conditions. The first controls the desired loading latency for incidents and sets the maximum time before we should consider an incident update latent. The second controls how many of those latent events can occur before throwing an exception which will bubble up through a runtime request.


src/baldr/graphreader.cc
```diff
@@ -12,9 +12,220 @@
...
+  }
+  return tile;
+}
+
+/**
+ * Thread work function that continually checks for updates to incident tiles. The updates come in
+ * the form of timestamps in a memory mapped file of 64bit integers (one per tile). The first 25 bits
+ * are the tile id and level (as a normal graphid) the remaining 39 bits are the timestamp. When
+ * the timestamp for the last check is older than a timestamp for a given file that file is replaced
+ * with whatever its contents are on disk.
```
PureW, line 65: Perhaps an additional bit or two to describe which buffer in a double buffer contains the newest update?

kevinkreiser: the temporary implementation here didnt actually end up using a double buffer approach. instead it does asynchronous loading and uses shared ptrs to transfer ownership to the reader when the writer invalidates a tile in its store

proto/incidents.proto
```diff
@@ -7,39 +7,38 @@
...
 option optimize_for = LITE_RUNTIME;
-package valhalla.incidents;
+package valhalla;
 
 message IncidentsTile {
   // Sorted list of edge_ids describing what incidents are attached to an edge_id
   repeated Location locations = 1;
   // Look at `incident_locations` to find how to index this array
   repeated Metadata metadata = 2;
-}
```
kevinkreiser, line 17: i removed the namespace (because the rest of the valhalla protos dont have sub namespaces) and i encapsulated the location and metadata defintions in the tile because location already exists elsewhere in the project and wuld be ambiguous. also there was inconsistent whitespace so i fixed that

src/baldr/graphreader.cc
```diff
@@ -357,9 +358,20 @@ GraphReader::GraphReader(const boost::property_tree::ptree& pt,
...
+
   // Reserve cache (based on whether using individual tile files or shared,
   // mmap'd file
   cache_->Reserve(tile_extract_->tiles.empty() ? AVERAGE_TILE_SIZE : AVERAGE_MM_TILE_SIZE);
+
+  // Initialize the incident cache singleton if we have any kind of configuration to do so
+  // if the configuration is wrong or any kind of problem occurs this throws
+  enable_incidents_ = !pt.get<std::string>("incident_log", "").empty() ||
+                      !pt.get<std::string>("incident_dir", "").empty();
+  if (enable_incidents_) {
```
kevinkreiser, line 370: we enable incidents and have the singleton load whatever incidents are there based on config. if its not in the config its completey disabled. the call below will spawn a single background thread which is responsible for loading incidents continually

src/proto_conversions.cc
```diff
@@ -5,46 +5,48 @@ using namespace valhalla;
 
 namespace valhalla {
 
-std::string incidentTypeToString(const valhalla::incidents::Metadata::Type& incident_type) {
```
kevinkreiser, line 8: renames

src/baldr/graphreader.cc
```diff
@@ -880,27 +892,31 @@ int GraphReader::GetTimezone(const baldr::GraphId& node, const GraphTile*& tile)
   return (tile == nullptr) ? 0 : tile->node(node)->timezone();
 }
 
+std::shared_ptr<valhalla::IncidentsTile> GraphReader::GetIncidentTile(const GraphId& tile_id) const {
+  return enable_incidents_ ? incident_singleton_t::get(tile_id.Tile_Base())
```
kevinkreiser, line 896: this had to move from the header to the implementation file because of accessing the singleton

src/baldr/graphreader.cc
```diff
@@ -880,27 +892,31 @@ int GraphReader::GetTimezone(const baldr::GraphId& node, const GraphTile*& tile)
...
+                           : std::shared_ptr<valhalla::IncidentsTile>{};
+}
+
 IncidentResult GraphReader::GetIncidents(const GraphId& edge_id, const GraphTile*& tile) {
   // if we are not doing this for any reason then bail
-  std::shared_ptr<valhalla::incidents::IncidentsTile> itile;
-  if (!simulate_incidents_ || !GetGraphTile(edge_id, tile) ||
+  std::shared_ptr<valhalla::IncidentsTile> itile;
```
kevinkreiser, line 902: rest of thsi file is renames

src/thor/triplegbuilder.cc
```diff
@@ -98,12 +98,12 @@ void AssignAdmins(const AttributesController& controller,
  * @param incident   the incident that applies
  * @param index      what shape index of the leg the index apples to
  */
```
kevinkreiser, line 100: renames

src/tyr/route_serializer_osrm.cc
```diff
@@ -1558,14 +1558,14 @@ TEST(RouteSerializerOsrm, testserializeIncidents) {
     incident->set_begin_shape_index(42);
     incident->set_end_shape_index(42);
 
-    valhalla::incidents::Metadata meta;
+    valhalla::IncidentsTile::Metadata meta;
```
kevinkreiser, line 1561: renames

src/tyr/serializers.cc
```diff
@@ -143,7 +143,7 @@ void serializeIncidentProperties(rapidjson::Writer<rapidjson::StringBuffer>& wri
                                  const TripLeg::Incident& incident,
                                  const std::string& road_class,
                                  const std::string& key_prefix) {
-  const valhalla::incidents::Metadata& meta = incident.metadata();
```
kevinkreiser, line 146: renames

test/gurka/test_route_incidents.cc
```diff
@@ -147,26 +146,26 @@ void check_incident_locations(const valhalla::Api& api,
 struct test_reader : public baldr::GraphReader {
   test_reader(const boost::property_tree::ptree& pt) : baldr::GraphReader(pt) {
     tile_extract_.reset(new baldr::GraphReader::tile_extract_t(pt));
+    enable_incidents_ = true;
```
kevinkreiser, line 149: now that we have actual incident loading, we need to turn on the incident loading singleton in config. for this test we override the functionality of the incident loader by letting the test inject incidents directly so here we just turnon the flag in the graph reader so that it will sitll call the incident tile function

test/gurka/test_route_incidents.cc
```diff
@@ -147,26 +146,26 @@ void check_incident_locations(const valhalla::Api& api,
...
   }
-  virtual std::shared_ptr<valhalla::incidents::IncidentsTile>
+  virtual std::shared_ptr<valhalla::IncidentsTile>
   GetIncidentTile(const baldr::GraphId& tile_id) const {
     auto i = incidents.find(tile_id.Tile_Base());
     if (i == incidents.cend())
       return {};
-    return std::shared_ptr<
```
kevinkreiser, line 156: rest is renames

valhalla/baldr/graphreader.h
```diff
@@ -26,7 +26,7 @@ namespace valhalla {
 namespace baldr {
 
 struct IncidentResult {
-  std::shared_ptr<incidents::IncidentsTile> tile;
+  std::shared_ptr<IncidentsTile> tile;
```
kevinkreiser, line 29: renames and the move the helper function to the source file. maybe we can delete this helper function as well since its never used directly cc @PureW @danpat

valhalla/proto_conversions.h
```diff
@@ -203,9 +203,9 @@ inline TripLeg_Use GetTripLegUse(const baldr::Use use) {
 }
 
 // Get the string representing the incident-type
-std::string incidentTypeToString(const valhalla::incidents::Metadata::Type& incident_type);
```
kevinkreiser, line 206: renames

kevinkreiser:
code coverage is failing, i still need to write unit tests for all the stuff in the singleton

src/baldr/incident_singleton.h
```diff
@@ -0,0 +1,301 @@
...
+  google::protobuf::io::CodedInputStream cs(
+      static_cast<google::protobuf::io::ZeroCopyInputStream*>(&as));
+
+  // try to parse the stream
+  std::shared_ptr<valhalla::IncidentsTile> tile(new valhalla::IncidentsTile);
+  if (!tile->ParseFromCodedStream(&cs)) {
+    LOG_WARN("Incident Watcher failed to parse " + filename);
+    return {};
+  }
+
```
kevinkreiser, line 54: TODO: if the tile we just loaded has no incidents in it we should also `return {}`

src/baldr/incident_singleton.h
```diff
@@ -0,0 +1,301 @@
...
+  if (!file.is_open()) {
+    LOG_WARN("Incident Watcher failed to open " + filename);
+    return {};
+  }
+
+  // prepare a stream for parsing
+  std::string buffer((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
+  google::protobuf::io::ArrayInputStream as(static_cast<const void*>(buffer.c_str()), buffer.size());
+  google::protobuf::io::CodedInputStream cs(
+      static_cast<google::protobuf::io::ZeroCopyInputStream*>(&as));
```
PureW, line 46: This could potentially be collapsed to
```
std::ifstream fin(filename);
ASSERT_TRUE(fin.is_open());

valhalla::incidents::IncidentsTile tile;
tile.ParseFromIstream(&fin);
```

kevinkreiser: sadly not available in our version of protobuf

PureW: Ignore above, our version of protobuf does not support this.

src/baldr/graphreader.cc
PureW: This would make a pretty helpful code comment?

src/baldr/incident_singleton.h
```diff
@@ -0,0 +1,306 @@
...
+
+#ifdef _MSC_VER
+#define MTIME(st_stat) st_stat.st_mtime
+#elif __APPLE__
+#define MTIME(st_stat) st_stat.st_mtime
+#else
+#define MTIME(st_stat) st_stat.st_mtim.tv_sec
+#endif
+
+constexpr time_t DEFAULT_MAX_LOADING_LATENCY = 60;
```
PureW, line 26: Nit: setting the `time_t` value may not be portable

> Portable programs should not use values of this type directly, but always rely on calls to elements of the standard library to translate them to portable types.

src/baldr/incident_singleton.h
```diff
@@ -0,0 +1,306 @@
...
+    std::unique_ptr<valhalla::midgard::sequence<uint64_t>> changelog;
+    filesystem::path inc_log_path(config.get<std::string>("incident_log", ""));
+    filesystem::path inc_dir;
+    try {
+      changelog.reset(new decltype(changelog)::element_type(inc_log_path.string()));
+    } // check for a directory scan mode config
+    catch (...) {
+      inc_dir = filesystem::path(config.get<std::string>("incident_dir", ""));
+      if (!filesystem::is_directory(inc_dir)) {
+        inc_dir = {};
```
PureW, line 168: Maybe log a message that it uses memmap or directory scan? Could be useful in debugging.

src/baldr/incident_singleton.h
kevinkreiser: its not portable in the sense that we dont know how many bits wide that it is. for this use we dont care, we just care that its the arithmetic type that comes out of the `time` function

src/baldr/incident_singleton.h
```diff
@@ -0,0 +1,306 @@
...
+    auto found = state->cache.find(tile_id);
+    // if we dont have a slot make one, this must be synchronized
+    if (found == state->cache.cend()) {
+      // this shouldnt happen in lock free mode but can if you put unexpected tiles in the log/dir
+      if (state->lock_free.load()) {
+        LOG_WARN("Incident watcher skipped " + std::to_string(tile_id) +
+                 " because it was not found in the configured tile extract");
+        return false;
+      }
+      state->mutex.lock();
```
PureW, line 127: Perhaps it would be safer to use a `unique_lock` that gets automatically destructured to prevent a deadlock if the hashtable `insert` below throws?

src/baldr/graphreader.cc
kevinkreiser: most of it is in the comment above, do you specifically mean the part about the background thread?

src/baldr/incident_singleton.h
kevinkreiser: great catch! ill use a scoped lock

PureW (approved):
:+1: Tests are following in next PR

PureW (approved):

PureW (approved):
:shipit:

PR #2575, opened 2020-09-03
Enhance BSS import

xlqian:
Hi,

I hope you're all going well!!

# Issue
This PR is targeting two problems concerning the BSS import.

**1. When the BSS node is projected on an edge whose start node and end node are not in the same tile.** 

 > https://github.com/valhalla/valhalla/blob/master/src/mjolnir/bssbuilder.cc#L159

The solution to this problem may be a little tricky, we couldn't find a simpler way to solve it.

- We first locate to which belong the BSS nodes.
- For every BSS node, we **assume** that we can always find an edge in the same tile of the BSS node to project it and we test only the `forwardaccess()` with the wanted travel mode. (From now on, the BSS can be projected on different edges according to the travel mode, the motivation is explained in the below section). Since we test only the `forwardaccess()`, the projected edge's start node **must** be in the same tile as the BSS node. Then we store the necessary pieces of information in a temporary struct.
- At this point, there are two kinds of edges to be added into the tile: BSS node's outbound edges(whose star tnode are bss node) and way node's outbound edges(whose start node are projected edges star tnode/end node, they are actually the opposite edges of BSS node's outbound edges, their evil brothers, like you said in the codes.)

image

- Once the nearest edge is found, we first add the BSS node and BSS node's outbound edges into the tile. Since they are totally new staff...
- For those way node outbound edges, we do a kind of sort to group way node outbound edges whose start nodes are in the same tile, then we can assign a thread for each tile to add these edges. 

The biggest problem of this method of projection is that, if the real nearest edge's start node and end node are not in the same tile as BSS node... the projection result will be wrong. But in practice, that shouldn't happen very often...

**2. When the BSS node is located between a pedestrian-only way and a bicycle-only way.**
During our internal test, we found that some BSS node may be found in this kind of situation:
image

The BSS node's nearest edge is a cycle lane and it is projected on this edge which **can't** be accessed by a pedestrian. As a result, this BSS is totally ignored during the A* search. 

The solution is to project the BSS node twice, once on a pedestrian-accessible way, another on a bicycle-accessible way. As you may notice, in the majority of cases, BSS nodes are projected twice on the same way, but it doesn't matter, the algo can hanle this naturally.

Note that for each BSS node, 8 edges will be added into the local tile.

Here is a final result: (the pointed line is the pedestrian route, and the solid blue line is the bicycle route):
image

In our opinion, it looks more likely what people will do in real life


src/mjolnir/bssbuilder.cc
```diff
@@ -500,24 +607,70 @@ void BssBuilder::Build(const boost::property_tree::ptree& pt, const std::string&
...
+
+    for (auto& thread : threads) {
+      thread->join();
+    }
   }
 
-  for (auto& thread : threads) {
-    thread->join();
+  // the collection is sorted so that the search will be much faster later.
+  boost::sort(all);
```
kevinkreiser, line 634: why not `std::sort(all.begin(), all.end());`?

xlqian: because it's shorter with boost :)

kevinkreiser:
@xlqian sorry fo rthe huge delay on this. it already looks good to me. i'm happy to merge this once the build is passing and up to date!

kevinkreiser (approved):

xlqian:
@kevinkreiser hthank you for the review and Happy New Year! (or a Better New Year! because the last time when we said HNY, it didn't work :p )   I apologize to protract the PR so long since we have been working on projects...

kevinkreiser:
@xlqian im sorry we havent shipped it yet! it was ready already a couple of times. when the build is green i will push the merge button

kevinkreiser:
doing a lot of merging at the moment. merging an optimizations branch and then merging this one. should be done shortly, just ran out of gas last night :smile:

xlqian:
:tada:

PR #2576, opened 2020-09-03
simplify triplegbuilder::build

kevinkreiser:
exactly what the title says. no change in functionality, just trying to trim down the build function to something a normal human being can make sense of to prepare for some more shape chopping fun

since the beginning of this pr, i also removed the trivial route (single edge) special case and handled the shape trimming accoringly. im going to want to add some test cases for this in gurka just to be on the safe side.


dgearhart (commented):
Do we have tests to verify multimodal changes

kevinkreiser:
@dgearhart nope. who ever added this marshalling code originally didnt add them. honestly i dont feel obligated to do that now since i literally moved the functions into a struct and changed nothing else.

to do this testing right, someone should write gurka like functionality to translate ascii into a transit pbf so that when we build tiles in gurka its picked up. i am not going to be doing that for this.

src/thor/triplegbuilder.cc
```diff
@@ -1258,21 +1484,16 @@ void TripLegBuilder::Build(
...
   std::vector<PointLL> trip_shape;
-  std::string arrival_time;
-  bool assumed_schedule = false;
   uint64_t osmchangeset = 0;
   size_t edge_index = 0;
   const DirectedEdge* prev_de = nullptr;
   const GraphTile* graphtile = nullptr;
-  // TODO: this is temp until we use transit stop type from transitland
-  TransitPlatformInfo_Type prev_transit_node_type = TransitPlatformInfo_Type_kStop;
+  MultimodalBuilder multimodal_builder(origin, time_info);
```
kevinkreiser, line 1493: all of the variables removed above are now member variables of the struct, intialized in the same way as they are here

src/thor/triplegbuilder.cc
```diff
@@ -271,102 +271,6 @@ void SetHeadings(TripLeg_Edge* trip_edge,
...
-/**
- * @param trip_node   Trip node to add transit nodes.
- * @param node        Start nodeinfo of the current edge.
- * @param startnode   Start node of the current edge.
- * @param start_tile  Tile of the start node.
- * @param graphtile   Graph tile of the current edge.
- * @param controller  Controller specifying attributes to add to trip edge.
- *
- */
-void AddTransitNodes(TripLeg_Node* trip_node,
```
kevinkreiser, line 306: moved into the struct, copied verbatim

src/thor/triplegbuilder.cc
```diff
@@ -1258,21 +1484,16 @@ void TripLegBuilder::Build(
...
   size_t edge_index = 0;
   const DirectedEdge* prev_de = nullptr;
   const GraphTile* graphtile = nullptr;
-  // TODO: this is temp until we use transit stop type from transitland
-  TransitPlatformInfo_Type prev_transit_node_type = TransitPlatformInfo_Type_kStop;
+  MultimodalBuilder multimodal_builder(origin, time_info);
 
   for (auto edge_itr = path_begin; edge_itr != path_end; ++edge_itr, ++edge_index) {
     const GraphId& edge = edge_itr->edgeid;
-    const uint32_t trip_id = edge_itr->trip_id;
```
kevinkreiser, line 1275: this was only used two places, so i just removed it and got it from `edge_itr` directly

src/thor/triplegbuilder.cc
```diff
@@ -1340,187 +1561,16 @@ void TripLegBuilder::Build(
       trip_node->mutable_cost()->mutable_transition_cost()->set_cost(edge_itr->transition_cost.cost);
     }
 
-    AddBssNode(trip_node, node, startnode, start_tile, graphtile, mode_costing, controller);
-    AddTransitNodes(trip_node, node, startnode, start_tile, graphtile, controller);
-
-    ///////////////////////////////////////////////////////////////////////////
-    // Add transit information if this is a transit stop. TODO - can we move
-    // this to another method?
-    if (node->is_transit()) {
```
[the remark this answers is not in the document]
merkispavel: As I see the "major" change here is extracting this piece of code to the standalone method. So it shouldn't introduce any bugs(:praying:)

src/thor/triplegbuilder.cc
```diff
@@ -1042,6 +946,328 @@ void AccumulateRecostingInfoForward(const valhalla::Options& options,
...
+        if (controller.attributes.at(kNodeTransitEgressInfoLatLon)) {
+          PointLL ll = node->latlng(start_tile->header()->base_ll());
+          stop_ll->set_lat(ll.lat());
+          stop_ll->set_lng(ll.lng());
+        }
+      }
+    }
+  }
+
+  void AddTransitInfo(TripLeg_Node* trip_node,
```
merkispavel, line 1089: nit: wonder if we need struct here? Maybe adding `AddTransitInfo` method next to the other existed ones is enough
Or is it used to group all three

merkispavel (dismissed):
2 functions are turned into the methods, and the other code piece is extracted to a new method. Looks like a pretty simple harmless refactoring 👍

src/thor/triplegbuilder.cc
```diff
@@ -1042,6 +946,328 @@ void AccumulateRecostingInfoForward(const valhalla::Options& options,
   }
 }
 
+/**
+ * A simple way to encapsulate the marshalling of native data types specific to multi modal into
+ * the protobuf ones that we pass downstream to odin and other clients
+ */
+struct MultimodalBuilder {
```
danpaz, line 953: Optional: my only suggestion would be to move this struct to a new file, it would be nice to trim down the number of lines in this file while you're doing some cleanup. Not a blocker though.

src/thor/triplegbuilder.cc
kevinkreiser: yep that was my first thought but that method specifically keeps state in the main loop of triplegbuilder::build. so i took that state and moved it into the struct to encapsulate it

src/thor/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ set(sources
   trace_attributes_action.cc
   trace_route_action.cc
   triplegbuilder.cc
-  triplegbuilder_utils.cc
+  triplegbuilder_utils.h
```
kevinkreiser, line 25: turned this into a private header

src/thor/triplegbuilder.cc
```diff
@@ -1148,131 +1053,18 @@ void TripLegBuilder::Build(
     trip_path.mutable_shape_attributes();
   }
 
-  // If the path was only one edge we have a special case
```
kevinkreiser, line 1151: handled the one edge special case below in shape trimming code, cc @danpaz

danpaz (approved):

danpaz (approved):

Issue #2592, opened 2020-09-11
Fix Russian localization


zotttttttt:
Three months ago there was an addition of several English phrases that need to be translated. Also, there are a lot of mistakes in the current translation that need to be fixed (`в строну` -> `в сторону`, `Едьте` -> `Двигайтесь` or `Поезжайте` and so on)

I'll make a PR soon.

PR #2593, opened 2020-09-11
Complete rework of the Russian localization

zotttttttt:
# Issue
#2592
The JSON in ru-RU.json is still valid, the structure is the same as before, only the content has been changed (for the better).

## Tasklist

 - [x] Add tests

## Requirements / Relations


danpaz (approved):
Thank you very much @zotttttttt ! 

/cc @yuryleb who was also looking at the Russian localization and may have some input here.

locales/ru-RU.json
```diff
@@ -36,49 +36,49 @@
       "phrases": {
         "0": "Держитесь <RELATIVE_DIRECTION>.",
         "1": "Держитесь <RELATIVE_DIRECTION> по <STREET_NAMES>.",
-        "2": "Держитесь <RELATIVE_DIRECTION> по <BEGIN_STREET_NAMES>. Продолжайте по <STREET_NAMES>.",
-        "3": "Держитесь <RELATIVE_DIRECTION> чтобы остаться на <STREET_NAMES>.",
-        "4": "Bear <RELATIVE_DIRECTION> at <JUNCTION_NAME>.",
-        "5": "Bear <RELATIVE_DIRECTION> toward <TOWARD_SIGN>."
+        "2": "Держитесь <RELATIVE_DIRECTION> по <BEGIN_STREET_NAMES>. Продолжайте движение по <STREET_NAMES>.",
```
yuryleb, line 39: "Двигайтесь <RELATIVE_DIRECTION> по <BEGIN_STREET_NAMES> на <STREET_NAMES>." ?

locales/ru-RU.json
```diff
@@ -36,49 +36,49 @@
...
+        "3": "Держитесь <RELATIVE_DIRECTION>, чтобы остаться на <STREET_NAMES>.",
```
yuryleb, line 40: "Двигайтесь <RELATIVE_DIRECTION> по <STREET_NAMES>." ?

locales/ru-RU.json
```diff
@@ -36,49 +36,49 @@
...
+        "4": "Держитесь <RELATIVE_DIRECTION> по ответвлению <JUNCTION_NAME>.",
```
yuryleb, line 41: "Двигайтесь <RELATIVE_DIRECTION> на <JUNCTION_NAME>." ?

locales/ru-RU.json
```diff
@@ -36,49 +36,49 @@
...
+        "5": "Держитесь <RELATIVE_DIRECTION> в сторону <TOWARD_SIGN>."
```
yuryleb, line 42: "Двигайтесь <RELATIVE_DIRECTION> в сторону <TOWARD_SIGN>." ?

locales/ru-RU.json
```diff
@@ -132,10 +132,10 @@
     },
     "continue": {
       "phrases": {
-        "0": "Продолжайте.",
-        "1": "Продолжайте по <STREET_NAMES>.",
-        "2": "Continue at <JUNCTION_NAME>.",
-        "3": "Continue toward <TOWARD_SIGN>."
+        "0": "Продолжайте дижение.",
```
yuryleb, line 135: Misprint in "движение". And perhaps better "Продолжите" instead of "Продолжайте"?

locales/ru-RU.json
```diff
@@ -354,363 +354,363 @@
...
-          "Заедьте на кольцо и съезжайте на втором съезде.",
-          "Заедьте на кольцо и съезжайте на третьем съезде.",
-          "Заедьте на кольцо и съезжайте на четвертом съезде.",
-          "Заедьте на кольцо и съезжайте на пятом съезде.",
-          "Заедьте на кольцо и съезжайте на шестом съезде.",
-          "Заедьте на кольцо и съезжайте на седьмом съезде.",
-          "Заедьте на кольцо и съезжайте на восьмом съезде.",
-          "Заедьте на кольцо и съезжайте на девятом съезде.",
-          "Заедьте на кольцо и съезжайте на десятом съезде."
+          "Круговое движение. Далее проезжайте по первому съезду.",
```
yuryleb, line 450: "Покиньте круговое движение на первом съезде" ?

yuryleb (commented):
Newer Russian localization looks better than before but unfortunately it has no any relationship with "original" https://github.com/Project-OSRM/osrm-text-instructions 😞 I didn't review the whole project yet but suppose that it will be impossible to embed some pre-processing before collation of final localized string like https://github.com/Project-OSRM/osrm-text-instructions/blob/master/languages/grammar/README.md

locales/ru-RU.json
zotttttttt: That means you need to rewrite "relative_directions" as well. And "Держитесь правее" is a standard for any Russian localization of GPS voiceover,  not "Двигайтесь справа".

locales/ru-RU.json
zotttttttt: Answered in https://github.com/valhalla/valhalla/pull/2593#discussion_r488044670

I really want it to be better, but you can't just rewrite every particular Russian text with just 2 versions of "relative_directions".

locales/ru-RU.json
zotttttttt: Again, there is a particular set of ordinal directions:

`"ordinal_values": [
        "первому",
        "второму",
        "третьему",
        "четвертому",
        "пятому",
        "шестому",
        "седьмому",
        "восьмому",
        "девятому",
        "десятому"
      ],`

zotttttttt: Russian is too hard to rewrite it well when you have just one version of a word in one particular Case (`Падеж`) :)

locales/ru-RU.json
yuryleb: What GPS do you treat as 'standard'? 😉 
But maybe 'Держитесь' will be better, 'Двигайтесь' is just more generic.

locales/ru-RU.json
zotttttttt: Thank you, I'll fix it. And "Продолжите" really feels better

locales/ru-RU.json
zotttttttt: Yandex.Navigator for me :)

locales/ru-RU.json
zotttttttt: Oh no, I misread it. "Продолжите" can be used when you're changing something, there you're just continuing the previous process to some point. Therefore, "Продолжайте" is a better choice

kevinkreiser:
@zotttttttt @yuryleb thank you very much for your work here. i am not well informed about slavic languages however, i would like to mention that if a language needs special functionality in which the replacement tag needs to change based on the grammar of the phrase it is being used in, then we do plan on supporting that in the future. i am well aware of such cases in other languages such as germannic and romance langugaes and i think we can make functionality similar to what is available in OSRM Text Instructions (though likely a different tactic or technique). since we are lacking this ability currently i don't think this is something that should block the current localization efforts here, its just a "bug" or "downside" that we can fix with future fuctionality in my opinion.

please also note that we are not attempting to replace the localization in valhalla with those which are used in OSRM. valhalla has different types of meta data available to its localization engine which we would like to expose in the text strings and this would simply not be possible if we tried that approach.

zotttttttt:
Yes, for sure. I think that every Slavic language would benefit from this new special functionality since they all have so many word forms (e.g., `друг, друга, другу, другом, друге` are the forms of one word in one tense). 

In my take on this localization, I focused more on Valhalla and didn't explore OSRM documentation since these are 2 different projects in my opinion. So, here we have the same thoughts :)

Thank you very much! I'm enjoying this library and would be happy to contribute at least something.

genadz (approved):

kevinkreiser (approved):
@genadz who is also a russian speaker shipped this but there was a conflict, so ive resolved that and am just passing his "ship" forward :smile:

danpaz (approved):
Fixed conflicts and merging.

PR #2579, opened 2020-09-04
Add Flexibility to Remove Use of AdminDB

kdiluca:
# Issue

We want to allow for the flexibility to disable the use of the admin db and to process admin information on the node.

## Tasklist

 - [ ] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [ ] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #2533, opened 2020-08-14
New Rest/Service Areas Use

kdiluca:
# Issue

What issue is this PR targeting? If there is no issue that addresses the problem, please open a corresponding issue and link it here.

## Tasklist

 - [ ] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [ ] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #2594, opened 2020-09-14
Update serialization for rest_stop and admins

danpaz:
# Issue

Follow up to https://github.com/valhalla/valhalla/pull/2579 and https://github.com/valhalla/valhalla/pull/2533, updating the osrm serializer to make use of the new data.

## Tasklist

 - [x] Add tests

## Requirements / Relations


kdiluca (approved):
This LGTM :+1:

src/tyr/route_serializer_osrm.cc
```diff
@@ -1386,6 +1448,22 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
...
+    for (const auto& admin : path_leg.admin()) {
+      auto admin_map = json::map({});
+      if (admin.has_country_code()) {
+        admin_map->emplace("iso_3166_1", admin.country_code());
+        auto country_iso3 = iso2_to_iso3.find(admin.country_code());
+        if (country_iso3 != iso2_to_iso3.end()) {
+          admin_map->emplace("iso_3166_1_alpha3", country_iso3->second);
+        }
+      }
+      // TODO: iso_3166_2 state code
```
dgearhart, line 1462: So does this mean we will have different admin index pointing to the same country code?

danpaz: That's right yes, we'll have duplicate-looking entries in the admins list. I noticed this when crossing state boundaries in the US like this:
```
"admins: [{
  "iso_3166_1: "US",
  "iso_3166_1_alpha3: "USA"
}, {
  "iso_3166_1: "US",
  "iso_3166_1_alpha3: "USA"
}]
```

Other than looking strange in the JSON response, do you think this'll be a problem? I couldn't think of a straightforward workaround for this without changing how admins are listed and indexed.

dgearhart: just thought it might look odd to some users - since we are not outputting the state level admin at this time; yeah, not worth changing if we output state level admin in the future

dgearhart (approved):
:shipit:

kdiluca (approved):

PR #2595, opened 2020-09-14
Enhance Maneuver information on "trails"

dnesbitt61:
Currently, a maneuver only has information on whether the maneuver is an unnamed cycleway, unnamed walkway, or unnamed mtb trail. This PR changes this to include a TrailType within the Maneuver class. This will allow distinguishing named cycleways/walkways/MTB trails from unnamed in addition to just identifying unnamed trails. This can allow more complex logic when transitioning between trails of different types and with different naming (e.g. named vs. unnamed).

RAD testing showed only one small difference in one bicycle route where previously a service road was combined with a named cycleway within the Obvious maneuver comparison. With new code, this maneuver is NOT combined since this is a transition between 2 different trail types (none, and a named cycleway).

# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


dgearhart (approved):
looks good - thanks

PR #2596, opened 2020-09-14
fixes time tracking test

kevinkreiser:
the time tracking test had a small logic error that, when testing the "seconds from now" functionality, it allowed you to use the 60th minute of the hour. of course that is nonsense, so this fixes that test by making sure it only uses 0-59, ie valid minutes of the hour

### DIFF
--- a/test/gurka/test_time_tracking.cc
+++ b/test/gurka/test_time_tracking.cc
@@ -75,10 +75,8 @@ TEST(TimeTracking, make) {
 
     // offset the time from now a bit
     now_str = dt::iso_date_time(tz);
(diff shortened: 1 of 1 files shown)


test/gurka/test_time_tracking.cc
```diff
@@ -75,10 +75,8 @@ TEST(TimeTracking, make) {
 
     // offset the time from now a bit
     now_str = dt::iso_date_time(tz);
-    auto minutes = std::atoi(now_str.substr(now_str.size() - 2, 2).c_str());
-    int offset = 7;
-    if (minutes + offset > 60)
```
kevinkreiser, line 80: if this were `>=` that would also have worked, derp

danpaz (approved):

PR #2597, opened 2020-09-14
Use live traffic in edge filter

genadz:
## Tasklist

 - [x] Add tests

## Requirements / Relations


src/loki/search.cc
```diff
@@ -476,11 +474,14 @@ struct bin_handler_t {
...
     // filtered then the reaches of both edges are the same
-    const DirectedEdge* opp_edge = nullptr;
-    if (reach.outbound > 0 && reach.inbound > 0 && (opp_edge = reader.GetOpposingEdge(edge, tile)) &&
-        edge_filter(opp_edge) > 0.f)
-      directed_reaches[opp_edge] = reach;
-
+    if (reach.outbound > 0 && reach.inbound > 0) {
+      const DirectedEdge* opp_edge = nullptr;
+      const auto opp_edgeid = reader.GetOpposingEdgeId(edge_id, tile);
+      if (opp_edgeid && (opp_edge = tile->directededge(opp_edgeid)) &&
```
kevinkreiser, line 480: i think id prefer not changing this to be a doubly nested `if`. it may have no performance impact or it may, we'll have to do extra testing to find out and i am not sure it really improves readibility. i guess in this case i would have modified the helper functions in graph reader to get both the edge and the edge id? maybe that would be more useful?

src/loki/reach.cc
```diff
@@ -203,6 +181,28 @@ directed_reach Reach::exact(const valhalla::baldr::DirectedEdge* edge,
...
+                    const std::shared_ptr<sif::DynamicCost>& costing,
+                    const baldr::GraphTile*& tile) {
+  // skip nodes which are done or invalid
+  if (!node_id.Is_Valid() || done_.find(node_id) != done_.cend())
+    return;
+  // if the node isnt accessable bail
+  if (!reader.GetGraphTile(node_id, tile))
+    return;
+  const auto* node = tile->node(node_id);
+  if (costing->GetNodeFilter()(node))
```
kevinkreiser, line 195: creating an std function on every call is probably not a good idea

valhalla/sif/dynamiccost.h
```diff
@@ -202,7 +192,7 @@ class DynamicCost {
    * @param   edge  Pointer to edge information.
    * @return  Returns true if access is allowed, false if not.
    */
-  inline virtual bool IsAccessable(const baldr::DirectedEdge* edge) const {
+  inline virtual bool IsAccessible(const baldr::DirectedEdge* edge) const {
```
kevinkreiser, line 195: haha good catch

valhalla/sif/dynamiccost.h
```diff
@@ -212,6 +202,19 @@ class DynamicCost {
...
+   * Determine whether an edge is currently closed due to traffic.
+   * @param  edgeid         GraphId of the opposing edge.
+   * @return  Returns true if the edge is closed due to live traffic constraints, false if not.
+   */
+  inline virtual bool IsClosedDueToTraffic(const baldr::GraphId& edgeid,
+                                           const baldr::GraphTile* tile) const {
+    if (flow_mask_ & baldr::kCurrentFlowMask) {
+      return tile->IsClosedDueToTraffic(edgeid);
+    }
+    return false;
```
kevinkreiser, line 215: this can be a one liner without an `if`

valhalla/sif/dynamiccost.h
```diff
@@ -582,12 +585,21 @@ class DynamicCost {
...
+   * used to rank edges such that desirable edges that might be slightly
+   * farther from the location than a less desirable edge can be chosen.
+   *
+   * Function to be used in location searching which will
    * exclude and allow ranking results from the search by looking at each
    * edges attribution and suitability for use as a location by the travel
    * mode used by the costing method.
    */
-  virtual const EdgeFilter GetEdgeFilter() const = 0;
+  virtual float Filter(const baldr::DirectedEdge* edge,
```
kevinkreiser, line 600: we should change the node filter into a function on the costing object rather than a functor as well. the reason that they werent changed to this in the past was just that the interface was different long ago. this is much better!

valhalla/sif/dynamiccost.h
kevinkreiser:
```c++
return (flow_mask_ & baldr::kCurrentFlowMask) && tile->IsClosedDueToTraffic(edgeid);
```

valhalla/sif/dynamiccost.h
```diff
@@ -212,6 +202,19 @@ class DynamicCost {
            (ignore_oneways_ && (edge->reverseaccess() & access_mask_));
   }
 
+  /*
+   * Determine whether an edge is currently closed due to traffic.
+   * @param  edgeid         GraphId of the opposing edge.
+   * @return  Returns true if the edge is closed due to live traffic constraints, false if not.
+   */
+  inline virtual bool IsClosedDueToTraffic(const baldr::GraphId& edgeid,
```
kevinkreiser, line 210: this should be a private/protected function, there is no need to have this in the public interface

src/loki/reach.cc
```diff
@@ -78,41 +55,36 @@ directed_reach Reach::operator()(const DirectedEdge* edge,
...
+
+    auto directed_edges = tile->GetDirectedEdges(node_id);
+    for (const auto* edge = directed_edges.begin(); edge != directed_edges.end(); ++edge) {
       // TODO: we'd rather say !edge.end_simple_restriction() and not !edge.restrictions()
       // TODO: but we'd need the predecessor information to do that so we punt 1 edge earlier
       // NOTE: we can go through the start of the restriction because only the end would mark a
       // potential stopping point (maybe a path followed the restriction)
 
+      GraphId edge_id = tile->id();
+      edge_id.set_id(node->edge_index() + (edge - directed_edges.begin()));
```
kevinkreiser, line 68: instead of this, you can make the edge_id outside o the loop, and then `++edge_id` up there where you `++edge`. we have a prefix `++` operator for graph ids that works as you would expect

src/loki/search.cc
```diff
@@ -580,17 +581,22 @@ struct bin_handler_t {
...
-          if (opp_reach.outbound >= p_itr->location.min_outbound_reach_ &&
-              opp_reach.inbound >= p_itr->location.min_inbound_reach_) {
-            tile = opp_tile;
-            edge = opp_edge;
-            reach = opp_reach;
-            reachable = true;
+        if (!reachable) {
+          const GraphTile* opp_tile = tile;
+          const GraphId opp_edgeid = reader.GetOpposingEdgeId(edge_id, opp_tile);
+          if (!opp_edgeid)
```
kevinkreiser, line 587: similar nesting addition as the comment i made above. maybe there is a better way to get both the opposing edge and id in one call so we can avoid the nesting

valhalla/sif/dynamiccost.h
genadz: yeah, make sense

src/loki/reach.cc
```diff
@@ -26,46 +26,23 @@ directed_reach Reach::operator()(const DirectedEdge* edge,
...
   max_reach_ = max_reach;
 
-  auto node_filter = costing->GetNodeFilter();
-  auto edge_filter = costing->GetEdgeFilter();
-
   // we keep a queue of nodes to expand from, to prevent duplicate expansion we use a set
   // each node we pop from the set will increase the reach and be added to the done set
   // the done set is used to avoid duplicate expansion of already dequeued nodes
   // we also track how many nodes were added as transitions from other levels
   // this allows us to have "duplicate" nodes but not do any trickery with the expansion
```
kevinkreiser, line 33: move this comment with the method you created for enqueue

valhalla/loki/reach.h
```diff
@@ -60,13 +60,18 @@ class Reach : public thor::Dijkstras {
...
-  directed_reach exact(const valhalla::baldr::DirectedEdge* edge,
+  directed_reach exact(const baldr::DirectedEdge* edge,
                        const baldr::GraphId edge_id,
                        uint32_t max_reach,
-                       valhalla::baldr::GraphReader& reader,
+                       baldr::GraphReader& reader,
                        const std::shared_ptr<sif::DynamicCost>& costing,
                        uint8_t direction = kInbound | kOutbound);
 
+  void enqueue(const baldr::GraphId& node_id,
```
kevinkreiser, line 70: maybe also copy the comment here to describe what this method does. probably this and the clear method and the member variables should be marked `protected` as well

valhalla/baldr/graphreader.h
```diff
@@ -455,6 +455,24 @@ class GraphReader {
...
+   * @param  edgeid      Graph Id of the directed edge.
+   * @param  opp_edge    Reference to a pointer to a const directed edge.
+   * @param  tile        Reference to a pointer to a const tile.
+   * @return  Returns the graph Id of the opposing directed edge. An
+   *          invalid graph Id is returned if the opposing edge does not
+   *          exist (can occur with a regional extract where adjacent tile
+   *          is missing).
+   */
+  GraphId
+  GetOpposingEdgeId(const GraphId& edgeid, const DirectedEdge*& opp_edge, const GraphTile*& tile) {
```
kevinkreiser, line 469: :+1:

valhalla/baldr/graphreader.h
```diff
@@ -506,6 +524,14 @@ class GraphReader {
     return GetGraphTile(edge->endnode(), tile) ? tile->node(edge->endnode()) : nullptr;
   }
 
+  /**
+   * Method to get the begin node of an edge by using its opposing edges end node
+   * @param edge    the edge whose begin node you want
+   * @param tile    reference to a pointer to a const tile
+   * @return        returns GraphId of begin node of the edge (empty if couldn't find)
+   */
+  GraphId GetBeginNodeId(const DirectedEdge* edge, const GraphTile*& tile);
```
kevinkreiser, line 533: maybe put the body of this method in the header as well. as other things start to use it, the importance of the compiler inlining it will increase

test/util/traffic_utils.h
```diff
@@ -125,11 +125,11 @@ void build_live_traffic_data(const boost::property_tree::ptree& config,
...
 /*************************************************************/
-void customize_live_traffic_data(
-    const boost::property_tree::ptree& config,
-    std::function<void(baldr::GraphReader&, baldr::TrafficTile&, int, baldr::TrafficSpeed*)>&
-        setter_cb) {
+using LiveTrafficCustomize = std::function<
+    void(baldr::GraphReader&, baldr::TrafficTile&, int, valhalla::baldr::TrafficSpeed*)>;
 
+void customize_live_traffic_data(const boost::property_tree::ptree& config,
+                                 LiveTrafficCustomize setter_cb) {
```
kevinkreiser, line 132: it probably doesnt matter because its a test but you are copying the function every time you call this

src/loki/search.cc
```diff
@@ -580,17 +577,22 @@ struct bin_handler_t {
...
-            tile = opp_tile;
-            edge = opp_edge;
-            reach = opp_reach;
-            reachable = true;
+        if (!reachable) {
+          const DirectedEdge* opp_edge = nullptr;
+          const GraphTile* opp_tile = tile;
+          const GraphId opp_edgeid = reader.GetOpposingEdgeId(edge_id, opp_edge, opp_tile);
+          if (!opp_edge)
+            continue;
```
kevinkreiser, line 585: this should be combined with the previous `if` like it was before

src/loki/search.cc
```diff
@@ -476,11 +473,14 @@ struct bin_handler_t {
 
     // if the inbound reach is not 0 and the outbound reach is not 0 and the opposing edge is not
     // filtered then the reaches of both edges are the same
+
     const DirectedEdge* opp_edge = nullptr;
-    if (reach.outbound > 0 && reach.inbound > 0 && (opp_edge = reader.GetOpposingEdge(edge, tile)) &&
-        edge_filter(opp_edge) > 0.f)
-      directed_reaches[opp_edge] = reach;
+    const auto opp_edgeid = reader.GetOpposingEdgeId(edge_id, opp_edge, tile);
```
kevinkreiser, line 478: we want this to be further down in the terms for the following `if` statement because its heavier than the other checks that happen before it. but always doing  it regardless you make sure its the worst performance possible. you should put it back inside the `if` statement below

src/loki/search.cc
```diff
@@ -410,11 +407,11 @@ struct bin_handler_t {
         correlated.edges.push_back(std::move(path_edge));
       }
       // correlate its evil twin
+      const DirectedEdge* other_edge = nullptr;
       const GraphTile* other_tile;
-      auto opposing_edge_id = reader.GetOpposingEdgeId(candidate.edge_id, other_tile);
-      const DirectedEdge* other_edge;
-      if (opposing_edge_id.Is_Valid() && (other_edge = other_tile->directededge(opposing_edge_id)) &&
-          edge_filter(other_edge) != 0.0f) {
+      auto opposing_edge_id = reader.GetOpposingEdgeId(candidate.edge_id, other_edge, other_tile);
```
kevinkreiser, line 412: should go back inside the `if` statement so that other shortcircuites happen first

src/sif/autocost.cc
```diff
@@ -268,27 +268,29 @@ class AutoCost : public DynamicCost {
...
+   * mode used by the costing method. It's also used to filter
    * edges not usable / inaccessible by automobile.
    */
-  virtual const EdgeFilter GetEdgeFilter() const {
-    // Throw back a lambda that checks the access for this type of costing
+  float Filter(const baldr::DirectedEdge* edge,
+               const baldr::GraphId& edgeid,
+               const baldr::GraphTile* tile) const override {
+    // take into account live traffic constraints
+    if (IsClosedDueToTraffic(edgeid, tile))
```
kevinkreiser, line 281: since this is unliklely and its a heavier check, we should move it into the `if` below and put it as the last conditions. this allows other cheaper checks to fail first and avoid ever going to look at the traffic data. we should do this for all the costings not just this one

genadz: nice catch

src/sif/autocost.cc
```diff
@@ -367,16 +369,13 @@ bool AutoCost::Allowed(const baldr::DirectedEdge* edge,
                        const uint64_t current_time,
                        const uint32_t tz_index,
                        int& restriction_idx) const {
-
-  if (flow_mask_ & kCurrentFlowMask) {
-    if (tile->IsClosedDueToTraffic(edgeid))
-      return false;
-  }
+  if (IsClosedDueToTraffic(edgeid, tile))
```
kevinkreiser, line 372: same for this one, it should be moved into the `if` below and added as the last condition. this should also be done for all costings that care about traffic not just auto

src/loki/search.cc
genadz: which of them ? there is only `costing->Filter` here

kevinkreiser: ah sorry i mixed this up with another line when i looked at the diff. sorry for that!

src/loki/reach.cc
```diff
@@ -125,20 +91,25 @@ directed_reach Reach::operator()(const DirectedEdge* edge,
...
       if (!reader.GetGraphTile(edge.endnode(), tile))
         continue;
-      const auto* node = tile->node(edge.endnode());
-      const auto* opp_edge = tile->directededge(node->edge_index() + edge.opp_index());
 
+      // get opposing edge id
+      GraphId opp_edgeid = edge.endnode();
+      opp_edgeid.set_id(tile->node(edge.endnode())->edge_index() + edge.opp_index());
+      // get opposing edge
+      const auto* opp_edge = tile->directededge(opp_edgeid);
```
kevinkreiser, line 104: probably call the new helper function you made?

src/loki/search.cc
```diff
@@ -498,14 +498,11 @@ struct bin_handler_t {
...
+      if (costing->Filter(edge, edge_id, tile) == 0.0f) {
         // then we try its opposing edge
-        edge_id = reader.GetOpposingEdgeId(edge_id, tile);
+        edge_id = reader.GetOpposingEdgeId(edge_id, edge, tile);
         // but if we couldnt get it or its filtered too then we move on
-        if (!edge_id.Is_Valid())
-          continue;
-        edge = tile->directededge(edge_id);
-        if (edge_filter(edge) == 0.0f)
+        if (!edge_id || costing->Filter(edge, edge_id, tile) == 0.0f)
```
kevinkreiser, line 505: i think this is a bug. doing `!edge_id` will implicit cast edge_id to uint64_t. the boolean cast needs to be explicit for it to automatically call `Is_Valid` internally. i think just in case you should call `Is_Valid` maybe in the future we should just delete the explicit boolean cast operator in graphid just to avoid having to think about this

src/loki/search.cc
```diff
@@ -579,12 +576,13 @@ struct bin_handler_t {
         // is this edge reachable in the right way
         bool reachable = reach.outbound >= p_itr->location.min_outbound_reach_ &&
                          reach.inbound >= p_itr->location.min_inbound_reach_;
-        // it's possible that it isnt reachable but the opposing is, switch to that if so
-        const GraphTile* opp_tile = tile;
         const DirectedEdge* opp_edge = nullptr;
-        if (!reachable && (opp_edge = reader.GetOpposingEdge(edge, opp_tile)) &&
-            edge_filter(opp_edge) > 0.f) {
-          auto opp_reach = check_reachability(begin, end, opp_tile, opp_edge, edge_id);
```
kevinkreiser, line 587: the last argument here was `edge_id` but in the new code its been changed to `opp_edgeid`. without looking deeper this looks like you either fixed a bug or introduced one, seems like the former to me but i didnt check the code.

src/sif/autocost.cc
```diff
@@ -397,16 +392,13 @@ bool AutoCost::AllowedReverse(const baldr::DirectedEdge* edge,
...
-  if (flow_mask_ & kCurrentFlowMask) {
-    if (tile->IsClosedDueToTraffic(opp_edgeid))
-      return false;
-  }
   // Check access, U-turn, and simple turn restriction.
   // Allow U-turns at dead-end nodes.
-  if (!IsAccessable(opp_edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
+  if (!IsAccessible(opp_edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((opp_edge->restrictions() & (1 << pred.opp_local_idx())) && !ignore_restrictions_) ||
       opp_edge->surface() == Surface::kImpassable || IsUserAvoidEdge(opp_edgeid) ||
```
kevinkreiser, line 399: no need to change anything here, but for future reference we should probably move `IsUserAvoidEdge(opp_edgeid)` to the second last or last condition in this `if` block. if we ever need to try to squeeze some performance we might try that. again, no need to change anything just pointing it out to you

src/loki/reach.cc
genadz: we need to get `edge_id` at first to call new function. so, it will not simplify the code

kevinkreiser: oh i see its because of the loop of edges doest give you the edge_id which the new function needs. well that is a shame. i suppose we could make a helper on `GraphTile::GetEdgeId(const DirectedEdge* e)` that would return the edge id if the edge belonged to the tile and throw otherwise but that would technically be more complex than this. :man_shrugging: nevermind

kevinkreiser (approved):
you love to see that kind of add/deleted line ratio where the additions were all in the unit tests and the rest was just code motion. thanks!

PR #2599, opened 2020-09-15
nit: Fixes some compiler warnings

PureW:


valhalla/meili/routing.h
```diff
@@ -31,8 +31,6 @@ constexpr uint16_t kInvalidDestination = std::numeric_limits<uint16_t>::max();
 class Label : public sif::EdgeLabel {
 public:
   Label() {
-    // zero out the data but set the node Id and edge Id to invalid
-    memset(this, 0, sizeof(Label));
```
PureW, line 35: This seems dangerous as Label is a child of EdgeLabel, meaning that there would be a vtable which this memset would overwrite. I'm not sure what the memset does since this being the constructor, the fields should already be zero'd.

cc @dnesbitt61

kevinkreiser: it looks like the default constructor for EdgeLabel doesnt zero them out though..

valhalla/midgard/sequence.h
```diff
@@ -570,9 +567,10 @@ struct tar {
       // make a copy and blank the checksum
       header_t temp = *this;
       memset(temp.chksum, ' ', 8);
-      int64_t usum = 0, sum = 0;
+      int64_t sum = 0;
+      uint64_t usum = 0;
       // compute the checksum
-      for (int i = 0; i < sizeof(header_t); i++) {
+      for (int i = 0; (size_t)i < sizeof(header_t); i++) {
```
kevinkreiser, line 573: missed a spot

valhalla/midgard/sequence.h
```diff
@@ -570,9 +567,10 @@ struct tar {
       // make a copy and blank the checksum
       header_t temp = *this;
       memset(temp.chksum, ' ', 8);
-      int64_t usum = 0, sum = 0;
```
kevinkreiser, line 573: why did we need to do this?

kevinkreiser: for the comparison with rsum below i guess

valhalla/thor/dijkstras.h
```diff
@@ -87,11 +87,11 @@ class Dijkstras {
 
 protected:
   // A child-class must implement this to learn about what nodes were expanded
-  virtual void ExpandingNode(baldr::GraphReader& graphreader,
-                             const baldr::GraphTile* tile,
-                             const baldr::NodeInfo* node,
-                             const sif::EdgeLabel& current,
-                             const sif::EdgeLabel* previous){};
+  virtual void ExpandingNode(baldr::GraphReader&,
```
kevinkreiser, line 90: hahah good catch... write the comment but dont adhere to it.. classic kevin

valhalla/meili/routing.h
PureW: Discussed more with @kevinkreiser and the safest solution is probably to initialize the values in the `EdgeLabel` constructor

valhalla/sif/edgelabel.h
```diff
@@ -26,7 +26,11 @@ class EdgeLabel {
   /**
    * Default constructor.
    */
-  EdgeLabel() {
+  EdgeLabel()
+      : predecessor_(0), path_distance_(0), restrictions_(0), edgeid_(0), opp_index_(0),
```
kevinkreiser, line 30: i think predecessor should be kInvalidLabel and the graphids should be invalid ones

kevinkreiser (approved):

PR #2602, opened 2020-09-15
nit: Tiny compiler warning nit

PureW:


valhalla/midgard/sequence.h
```diff
@@ -576,7 +576,7 @@ struct tar {
       }
       // check if its right
       uint64_t rsum = octal_to_int(chksum);
```
kevinkreiser, line 578: should have just made thsi an int64 and been done haha

kevinkreiser (approved):

PR #2610, opened 2020-09-21
Add Incidents to TripLeg

kevinkreiser:
This code:

1. mocks out the interface in graphreader for getting incidents (per tile/edge)
2. it adds code to triplegbuilder to cut the edge shape where incidents occur
3. add the incident to the tripleg with the leg shape indices where it occurs

### DIFF
--- a/proto/trip.proto
+++ b/proto/trip.proto
@@ -287,31 +287,6 @@ message TripLeg {
     optional PathCost cost = 12;             // how much cost did it take at this node in the path
     repeated PathCost recosts = 13;          // how much cost did it take at this node in the path for optional recostings
     optional BikeShareStationInfo bss_info = 14;
-
-    message Incident {
-      enum Type {
-        NOT_SET = 0;
-        ACCIDENT = 1;
-        CONGESTION = 2;
-        CONSTRUCTION = 3;
-        DISABLED_VEHICLE = 4;
-        LANE_RESTRICTION = 5;
-        MASS_TRANSIT = 6;
-        MISCELLANEOUS = 7;
-        OTHER_NEWS = 8;
-        PLANNED_EVENT = 9;
-        ROAD_CLOSURE = 10;
-        ROAD_HAZARD = 11;
-        WEATHER = 12;
-      };
-      optional Type type = 1;
-      optional string description = 2;
-      optional uint64 start_time = 3;     // START_TIME as seconds since epoch UTC
-      optional uint64 end_time = 4;       // END_TIME seconds since epoch UTC
-      optional uint64 creation_time = 5;  // time the incident was created/updated in seconds since epoch UTC
-      optional uint64 id = 6;             // An id that can be used for joining incidents in route with map
-    };
-    repeated Incident incidents = 15;
  }
 
   message Admin {
@@ -327,6 +302,32 @@ message TripLeg {
     repeated uint32 speed = 3 [packed=true]; // decimeters per sec
   }
 
+  message Incident {
+    enum Type {
+      NOT_SET = 0;
+      ACCIDENT = 1;
+      CONGESTION = 2;
+      CONSTRUCTION = 3;
+      DISABLED_VEHICLE = 4;
+      LANE_RESTRICTION = 5;
+      MASS_TRANSIT = 6;
+      MISCELLANEOUS = 7;
+      OTHER_NEWS = 8;
+      PLANNED_EVENT = 9;
+      ROAD_CLOSURE = 10;
+      ROAD_HAZARD = 11;
+      WEATHER = 12;
+    };
+    optional Type type = 1;
+    optional string description = 2;
+    optional uint64 start_time = 3;     // START_TIME as seconds since epoch UTC
+    optional uint64 end_time = 4;       // END_TIME seconds since epoch UTC
+    optional uint64 creation_time = 5;  // time the incident was created/updated in seconds since epoch UTC
+    optional uint64 id = 6;             // An id that can be used for joining incidents in route with map
+    optional uint32 begin_shape_index = 7;
+    optional uint32 end_shape_index = 8;
+  };
+
   optional uint64 osm_changeset = 1;
   optional uint64 trip_id = 2;
   optional uint32 leg_id = 3;
@@ -337,6 +338,7 @@ message TripLeg {
   optional string shape = 8;
   optional BoundingBox bbox = 9;
   optional ShapeAttributes shape_attributes = 10;
+  repeated Incident incidents = 11;
 }
 
 message TripRoute {
--- a/proto/trip.proto
+++ b/proto/trip.proto
@@ -231,6 +231,10 @@ message TripLeg {
     optional Restriction restriction = 45;
     optional bool destination_only = 46;
     optional bool is_urban = 47; // uses edge density to decide if edge is in an urban area
+    // for the part of the edge that is used in the path we must know where
+    // it starts and ends along the length of the edge as a percentage
+    optional float source_along_edge = 48;
+    optional float target_along_edge = 49;
   }
 
   message IntersectingEdge {
--- a/src/baldr/graphreader.cc
+++ b/src/baldr/graphreader.cc
@@ -346,7 +346,8 @@ GraphReader::GraphReader(const boost::property_tree::ptree& pt,
     : tile_extract_(get_extract_instance(pt)), tile_dir_(pt.get<std::string>("tile_dir", "")),
       tile_getter_(std::move(tile_getter)),
       max_concurrent_users_(pt.get<size_t>("max_concurrent_reader_users", 1)),
-      tile_url_(pt.get<std::string>("tile_url", "")), cache_(TileCacheFactory::createTileCache(pt)) {
+      tile_url_(pt.get<std::string>("tile_url", "")), cache_(TileCacheFactory::createTileCache(pt)),
+      simulate_incidents_(pt.get<bool>("simulate_incidents", 0)) {
   if (!tile_getter_ && !tile_url_.empty()) {
     tile_getter_ = std::make_unique<curl_tile_getter_t>(max_concurrent_users_,
                                                         pt.get<std::string>("user_agent", ""),
--- a/src/odin/enhancedtrippath.cc
+++ b/src/odin/enhancedtrippath.cc
@@ -1776,16 +1776,5 @@ std::string EnhancedTripLeg_Admin::ToString() const {
   return str;
 }
 
-const google::protobuf::RepeatedPtrField<valhalla::TripLeg_Node_Incident> empty;
-
-const google::protobuf::RepeatedPtrField<valhalla::TripLeg_Node_Incident>&
-EnhancedTripLeg_Node::incidents() const {
-  if (mutable_node_) {
-    return mutable_node_->incidents();
-  } else {
-    return empty;
-  }
-}
-
 } // namespace odin
 } // namespace valhalla
--- a/src/thor/attributes_controller.cc
+++ b/src/thor/attributes_controller.cc
@@ -75,7 +75,7 @@ const std::unordered_map<std::string, bool> AttributesController::kDefaultAttrib
     {kEdgeIsUrban, false},
 
     // Node keys
-    {kNodeIncidents, false},
+    {kIncidents, false},
     {kNodeIntersectingEdgeBeginHeading, true},
     {kNodeIntersectingEdgeFromEdgeNameConsistency, true},
     {kNodeIntersectingEdgeToEdgeNameConsistency, true},
--- a/src/thor/triplegbuilder.cc
+++ b/src/thor/triplegbuilder.cc
@@ -786,90 +786,6 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
 }
 
 /**
- * Add trip intersecting edge.
- * @param  controller   Controller to determine which attributes to set.
- * @param  directededge Directed edge on the path.
- * @param  prev_de  Previous directed edge on the path.
- * @param  local_edge_index  Index of the local intersecting path edge at intersection.
- * @param  nodeinfo  Node information of the intersection.
- * @param  trip_node  Trip node that will store the intersecting edge information.
- * @param  intersecting_de Intersecting directed edge. Will be nullptr except when
- *                         on the local hierarchy.
- */
-void AddTripIntersectingEdge(const AttributesController& controller,
-                             const DirectedEdge* directededge,
-                             const DirectedEdge* prev_de,
-                             uint32_t local_edge_index,
-                             const NodeInfo* nodeinfo,
-                             TripLeg_Node* trip_node,
-                             const DirectedEdge* intersecting_de) {
-  TripLeg_IntersectingEdge* itersecting_edge = trip_node->add_intersecting_edge();
-
-  // Set the heading for the intersecting edge if requested
-  if (controller.attributes.at(kNodeIntersectingEdgeBeginHeading)) {
-    itersecting_edge->set_begin_heading(nodeinfo->heading(local_edge_index));
-  }
-
-  Traversability traversability = Traversability::kNone;
-  // Determine walkability
-  if (intersecting_de->forwardaccess() & kPedestrianAccess) {
-    traversability = (intersecting_de->reverseaccess() & kPedestrianAccess)
-                         ? Traversability::kBoth
-                         : Traversability::kForward;
-  } else {
-    traversability = (intersecting_de->reverseaccess() & kPedestrianAccess)
-                         ? Traversability::kBackward
-                         : Traversability::kNone;
-  }
-  // Set the walkability flag for the intersecting edge if requested
-  if (controller.attributes.at(kNodeIntersectingEdgeWalkability)) {
-    itersecting_edge->set_walkability(GetTripLegTraversability(traversability));
-  }
-
-  traversability = Traversability::kNone;
-  // Determine cyclability
-  if (intersecting_de->forwardaccess() & kBicycleAccess) {
-    traversability = (intersecting_de->reverseaccess() & kBicycleAccess) ? Traversability::kBoth
-                                                                         : Traversability::kForward;
-  } else {
-    traversability = (intersecting_de->reverseaccess() & kBicycleAccess) ? Traversability::kBackward
-                                                                         : Traversability::kNone;
-  }
-  // Set the cyclability flag for the intersecting edge if requested
-  if (controller.attributes.at(kNodeIntersectingEdgeCyclability)) {
-    itersecting_edge->set_cyclability(GetTripLegTraversability(traversability));
-  }
-
-  // Set the driveability flag for the intersecting edge if requested
-  if (controller.attributes.at(kNodeIntersectingEdgeDriveability)) {
-    itersecting_edge->set_driveability(
-        GetTripLegTraversability(nodeinfo->local_driveability(local_edge_index)));
-  }
-
-  // Set the previous/intersecting edge name consistency if requested
-  if (controller.attributes.at(kNodeIntersectingEdgeFromEdgeNameConsistency)) {
-    bool name_consistency =
-        (prev_de == nullptr) ? false : prev_de->name_consistency(local_edge_index);
-    itersecting_edge->set_prev_name_consistency(name_consistency);
-  }
-
-  // Set the current/intersecting edge name consistency if requested
-  if (controller.attributes.at(kNodeIntersectingEdgeToEdgeNameConsistency)) {
-    itersecting_edge->set_curr_name_consistency(directededge->name_consistency(local_edge_index));
-  }
-
-  // Set the use for the intersecting edge if requested
-  if (controller.attributes.at(kNodeIntersectingEdgeUse)) {
-    itersecting_edge->set_use(GetTripLegUse(intersecting_de->use()));
-  }
-
-  // Set the road class for the intersecting edge if requested
-  if (controller.attributes.at(kNodeIntersectingEdgeRoadClass)) {
-    itersecting_edge->set_road_class(GetRoadClass(intersecting_de->classification()));
-  }
-}
-
-/**
  * This adds cost information at every node using supplementary costings provided at request time
  * There are some limitations here:
  * For multipoint routes the date_time used will not reflect the time offset that would have been if
@@ -1244,10 +1160,12 @@ void TripLegBuilder::Build(
       trip_edge->set_length(km);
     }
 
-    // Set shape attributes
+    // How long on this edge?
     auto edge_seconds = edge_itr->elapsed_cost.secs - edge_itr->transition_cost.secs;
     if (edge_itr != path_begin)
       edge_seconds -= std::prev(edge_itr)->elapsed_cost.secs;
+
+    // Set shape attributes
     SetShapeAttributes(controller, graphtile, directededge, trip_shape, begin_index, trip_path,
                        trim_start_pct, trim_end_pct, edge_seconds,
                        costing->flow_mask() & kCurrentFlowMask);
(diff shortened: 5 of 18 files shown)


proto/trip.proto
```diff
@@ -232,6 +232,11 @@ message TripLeg {
     optional bool destination_only = 46;
     optional bool is_urban = 47; // uses edge density to decide if edge is in an urban area
     repeated TaggedName tagged_name = 48;
+
+    // for the part of the edge that is used in the path we must know where
+    // it starts and ends along the length of the edge as a percentage
+    optional float source_along_edge = 49;
+    optional float target_along_edge = 50;
```
kevinkreiser, line 239: this is foreshadowing... there is currently a problem with recosting a path in that it only cares about partial edges at the beginning and end of a leg. sadly this isnt always the case. specifically  you can do uturns in the middle of an edge in the middle of a leg or you can have discontinuities on edges in the middle of a leg for map matching. in this case the recosting will not be correct for such edges (it will use the full cost of the edge). i'll make a separate issue for this.

proto/trip.proto
```diff
@@ -284,31 +289,6 @@ message TripLeg {
     optional PathCost cost = 12;             // how much cost did it take at this node in the path
     repeated PathCost recosts = 13;          // how much cost did it take at this node in the path for optional recostings
     optional BikeShareStationInfo bss_info = 14;
-
-    message Incident {
```
kevinkreiser, line 288: these are on the leg now

proto/trip.proto
```diff
@@ -324,6 +304,32 @@ message TripLeg {
...
+      WEATHER = 12;
+    };
+    optional Type type = 1;
+    optional string description = 2;
+    optional uint64 start_time = 3;     // START_TIME as seconds since epoch UTC
+    optional uint64 end_time = 4;       // END_TIME seconds since epoch UTC
+    optional uint64 creation_time = 5;  // time the incident was created/updated in seconds since epoch UTC
+    optional uint64 id = 6;             // An id that can be used for joining incidents in route with map
+    optional uint32 begin_shape_index = 7;
+    optional uint32 end_shape_index = 8;
```
kevinkreiser, line 330: these two shape indices were added to locate the incident within the leg

src/baldr/graphreader.cc
```diff
@@ -346,7 +346,8 @@ GraphReader::GraphReader(const boost::property_tree::ptree& pt,
     : tile_extract_(get_extract_instance(pt)), tile_dir_(pt.get<std::string>("tile_dir", "")),
       tile_getter_(std::move(tile_getter)),
       max_concurrent_users_(pt.get<size_t>("max_concurrent_reader_users", 1)),
-      tile_url_(pt.get<std::string>("tile_url", "")), cache_(TileCacheFactory::createTileCache(pt)) {
+      tile_url_(pt.get<std::string>("tile_url", "")), cache_(TileCacheFactory::createTileCache(pt)),
+      simulate_incidents_(pt.get<bool>("simulate_incidents", 0)) {
```
kevinkreiser, line 350: this is temporary but makes it possible to completely disable this code path from config. we'll delete this once the presence of incident data begins to dictate whether or not incidents are returned

src/odin/enhancedtrippath.cc
```diff
@@ -1776,16 +1776,5 @@ std::string EnhancedTripLeg_Admin::ToString() const {
...
-
-const google::protobuf::RepeatedPtrField<valhalla::TripLeg_Node_Incident>&
-EnhancedTripLeg_Node::incidents() const {
-  if (mutable_node_) {
-    return mutable_node_->incidents();
-  } else {
-    return empty;
-  }
-}
-
```
kevinkreiser, line 1789: there was no point in this. we use the incidents directly from the tripleg itself. odin doesnt care about them (yet)

src/thor/attributes_controller.cc
```diff
@@ -76,7 +76,7 @@ const std::unordered_map<std::string, bool> AttributesController::kDefaultAttrib
     {kEdgeTaggedNames, true},
 
     // Node keys
-    {kNodeIncidents, false},
+    {kIncidents, false},
```
kevinkreiser, line 79: no longer on the node

src/thor/triplegbuilder_utils.h
```diff
@@ -339,4 +339,172 @@ struct MultimodalBuilder {
...
+ * @param  controller   Controller to determine which attributes to set.
+ * @param  directededge Directed edge on the path.
+ * @param  prev_de  Previous directed edge on the path.
+ * @param  local_edge_index  Index of the local intersecting path edge at intersection.
+ * @param  nodeinfo  Node information of the intersection.
+ * @param  trip_node  Trip node that will store the intersecting edge information.
+ * @param  intersecting_de Intersecting directed edge. Will be nullptr except when
+ *                         on the local hierarchy.
+ */
+void AddTripIntersectingEdge(const AttributesController& controller,
```
kevinkreiser, line 353: i removed this from `triplegbuilder::build` because its huge and distracting. the hope is that one day that function will be lean and mean and a normal human will be able to understand it

src/thor/triplegbuilder_utils.h
```diff
@@ -339,4 +339,172 @@ struct MultimodalBuilder {
...
+ * @param controller               tells us what info we should add about the intersecting edges
+ * @param start_tile               the tile which contains the node
+ * @param node                     the node at which we are copying intersecting edges
+ * @param directededge             the current edge leaving the current node in the path
+ * @param prev_de                  the previous edge in the path
+ * @param prior_opp_local_index    opposing edge local index of previous edge in the path
+ * @param graphreader              graph reader for graph access
+ * @param trip_node                pbf node in the pbf structure we ar ebuilding
+ */
+void AddIntersectingEdges(const AttributesController& controller,
```
kevinkreiser, line 438: same, this method was also extracted

src/tyr/route_serializer_osrm.cc
```diff
@@ -682,13 +648,35 @@ valhalla::baldr::json::MapPtr serializeIncident(const TripLeg_Node_Incident& inc
...
+  if (incident.has_description() && !incident.description().empty()) {
     metadata_json->emplace("description", incident.description());
   }
+  if (incident.has_begin_shape_index()) {
+    metadata_json->emplace("geometry_index_start",
+                           static_cast<uint64_t>(incident.begin_shape_index()));
+  }
+  if (incident.has_end_shape_index()) {
+    metadata_json->emplace("geometry_index_end", static_cast<uint64_t>(incident.end_shape_index()));
+  }
```
kevinkreiser, line 660: here we add the two new properties

src/tyr/route_serializer_osrm.cc
```diff
@@ -682,13 +648,35 @@ valhalla::baldr::json::MapPtr serializeIncident(const TripLeg_Node_Incident& inc
...
+  if (incidents.empty()) {
+    return;
+  }
+  auto serialized_incidents = json::array({});
+  for (const auto& incident : incidents) {
+    auto json_incident = serializeIncident(incident);
+    serialized_incidents->emplace_back(json_incident);
+  }
+  json_leg->emplace("incidents", serialized_incidents);
+}
```
kevinkreiser, line 678: here we add all the incidents to the leg. deleted the other function that did it for the node above

src/tyr/route_serializer_osrm.cc
```diff
@@ -1567,24 +1560,26 @@ TEST(RouteSerializerOsrm, testAddsIncidents) {
 
   rapidjson::Document serialized_to_json;
   {
-    auto intersection_doc = json::Jmap();
-    auto inner_node = TripLeg_Node();
+    auto intersection_doc = json::map({});
```
kevinkreiser, line 1563: from here down is all just unit tests updated to test the serialization on the leg

test/gurka/gurka.h
```diff
@@ -166,12 +166,23 @@ build_config(const std::string& tiledir,
...
+midgard::PointLL to_ll(const nodelayout& nodes, const std::string& node_name) {
+  return nodes.at(node_name);
+}
+std::vector<midgard::PointLL> to_lls(const nodelayout& nodes,
+                                     const std::vector<std::string>& node_names) {
+  std::vector<midgard::PointLL> lls;
+  for (const auto& node_name : node_names)
+    lls.emplace_back(to_ll(nodes, node_name));
+  return lls;
+}
```
kevinkreiser, line 178: small refactor to make it possible to build a valhalla request with lls instead of OSM node names. didnt actually expose this in the match or route method but we'll see what the future holds

test/gurka/gurka.h
```diff
@@ -594,6 +605,57 @@ findEdge(valhalla::baldr::GraphReader& reader,
...
+/**
+ * Finds an edge in the graph based on its begin and end node names
+ *
+ * @param reader           graph reader to look up tiles and edges
+ * @param begin_node_name  name of the begin node
+ * @param end_node_name    name of the end node
+ * @return the edge_id and its edge
+ */
+std::tuple<const baldr::GraphId, const baldr::DirectedEdge*>
+findEdgeByNodes(valhalla::baldr::GraphReader& reader,
```
kevinkreiser, line 617: maybe we can delete the other method that does this, but i find it convenient to just say the begin and end node of the edge i want and just get that edge. this was very useful for the unit tests and is a less verbose and more intuitive way of specifying an edge via its OSM node names

test/gurka/gurka.h
```diff
@@ -640,13 +698,6 @@ valhalla::Api route(const map& map,
   return route(map, {from, to}, costing, options, reader);
 }
 
-valhalla::Api route(const map& map, const std::string& request_json) {
```
kevinkreiser, line 643: i moved this one above and  had the other route method call it. i also wanted to delete the other route method but it was used by a lot of tests. there is a bit of redundancy here. i was trying to clean up a bit of it

test/gurka/test_traffic.cc
```diff
@@ -50,9 +50,8 @@ TEST(Traffic, BasicUpdates) {
     printf("Make some updates to the traffic .tar file. "
            "Mostly just updates every edge in the file to 24km/h, except for one "
            "specific edge (B->D) where we simulate a closure (speed=0, congestion high)");
-    LiveTrafficCustomize cb_setter_24kmh = [&map](baldr::GraphReader& reader,
-                                                  baldr::TrafficTile& tile, int index,
-                                                  baldr::TrafficSpeed* current) -> void {
+    auto cb_setter_24kmh = [&map](baldr::GraphReader& reader, baldr::TrafficTile& tile, int index,
```
kevinkreiser, line 53: these were conflicts. i saw someone made a typedef for the functor but i had already made it a const reference in my implementation and was able to just use `auto` for the lambda. i kept the change but i can revert it if someone has strong feelings

test/gurka/test_traffic.cc
```diff
@@ -187,23 +187,25 @@ TEST(Traffic, CutGeoms) {
...
+        valhalla::baldr::UNKNOWN_TRAFFIC_SPEED_RAW,
+        valhalla::baldr::UNKNOWN_TRAFFIC_SPEED_RAW,
+        valhalla::baldr::UNKNOWN_TRAFFIC_SPEED_RAW,
+        valhalla::baldr::UNKNOWN_TRAFFIC_SPEED_RAW,
+        0u,
+        0u,
+        0u,
+        0u,
+        0u,
+        0u,
```
kevinkreiser, line 200: used one of the bits that was spare to say an edge has an incident on it (or is expected to, not gauranteed)

valhalla/baldr/graphreader.h
```diff
@@ -793,6 +794,52 @@ class GraphReader {
    */
   int GetTimezone(const baldr::GraphId& node, const GraphTile*& tile);
 
+  // TODO: temporary until we have real incidents loaded from incident loading singleton
```
kevinkreiser, line 797: this is the minimal structure of an incident that i needed to be able to structure the tripleg to have incidents on it

valhalla/baldr/graphreader.h
```diff
@@ -793,6 +794,52 @@ class GraphReader {
...
+    uint64_t id;
+  };
+  using incident_tile_t = std::vector<Incident>;
+
+  /**
+   * Returns an incident tile for the given tile id
+   * @param tile_id  the tile id for which incidents should be returned
+   * @return the incident tile for the tile id
+   */
+  virtual std::shared_ptr<incident_tile_t> GetIncidentTile(const GraphId& tile_id) const {
```
kevinkreiser, line 811: this method gets overriden in a test graph reader in my unit test so taht i can inject incidents directly into the reader on the exact edges exactly where i want them to be. made the testing very very nice.

valhalla/baldr/graphreader.h
```diff
@@ -793,6 +794,52 @@ class GraphReader {
...
+    return {};
+  }
+
+  /**
+   * Returns a vector of incidents for the given edge
+   * @param edge_id   which edge you need incidents for
+   * @param tile      which tile the edge lives in, is updated if not correct
+   * @return vector of incidents
+   */
+  std::vector<Incident> GetIncidents(const GraphId& edge_id, const GraphTile*& tile) {
```
kevinkreiser, line 824: other than the return type, this function should be pretty much final

valhalla/baldr/graphreader.h
```diff
@@ -819,6 +866,9 @@ class GraphReader {
   std::unordered_set<GraphId> _404s;
 
   std::unique_ptr<TileCache> cache_;
+
+  // TODO: delete me when incident singleton is ready
+  bool simulate_incidents_;
```
kevinkreiser, line 871: in place of real incidents for now just to enable or disable the logic for fetching incidents

valhalla/baldr/traffictile.h
```diff
@@ -52,12 +52,13 @@ struct TrafficSpeed {
...
-  uint64_t congestion1 : 6; // Stores 0 (unknown), or 1->63 (no congestion->max congestion)
-  uint64_t congestion2 : 6; //
-  uint64_t congestion3 : 6; //
-  uint64_t spare : 2;
+  uint64_t breakpoint1 : 8;   // position = length * (breakpoint1 / 255)
+  uint64_t breakpoint2 : 8;   // position = length * (breakpoint2 / 255)
+  uint64_t congestion1 : 6;   // Stores 0 (unknown), or 1->63 (no congestion->max congestion)
+  uint64_t congestion2 : 6;   //
+  uint64_t congestion3 : 6;   //
+  uint64_t has_incidents : 1; // Are there incidents on this edge in the corresponding incident tile
```
kevinkreiser, line 60: tells the graph reader that it should consider fetching incidents for this edge. due to synchronization it still may be possible that this disagrees with what will be in the incident tiles but it shouldnt matter. if it says they are there and they are not the worst case is we looked and found none in the graph reader function above. no big deal

src/thor/triplegbuilder.cc
```diff
@@ -91,93 +91,190 @@ void AssignAdmins(const AttributesController& controller,
   }
 }
 
+/**
```
kevinkreiser, line 94: this function is used to modify the tripleg when we need to add a new incident or update an existing one. let me know if i should move this to the triplegbuilder private header.

src/thor/triplegbuilder.cc
```diff
@@ -91,93 +91,190 @@ void AssignAdmins(const AttributesController& controller,
...
+  // bail if nothing to do
+  if (!cut_for_traffic && incidents.empty() &&
+      !controller.category_attribute_enabled(kShapeAttributesCategory))
+    return;
+
+  // initialize shape_attributes once
+  if (!leg.has_shape_attributes() &&
+      controller.category_attribute_enabled(kShapeAttributesCategory)) {
+    leg.mutable_shape_attributes();
+  }
```
kevinkreiser, line 157: this got refactored a bit. previously inside of triplegbuilder::build it was initializing shape attributes if that was enabled on the controller, now we do it once here. actually, i guess we dont even need to check if it has it anymore first since protobuf does that check internally.. derp

src/thor/triplegbuilder.cc
```diff
@@ -91,93 +91,190 @@ void AssignAdmins(const AttributesController& controller,
...
+
+  // convenient for bundling info about the spots where we cut the shape
+  struct cut_t {
+    double percent_along;
+    double speed; // meters per second
+    uint8_t congestion;
+    std::vector<const baldr::GraphReader::Incident*> incidents;
+  };
```
kevinkreiser, line 165: the tuple could no longer contain the insanity. this method is doing many things. it is storing per shape information. it is cutting where the speed/congestion changes along the edge and it is also cutting where incidents start and end along the edge.

src/thor/triplegbuilder.cc
```diff
@@ -91,93 +91,190 @@ void AssignAdmins(const AttributesController& controller,
...
+    // cutting for now
+    const auto& traffic_speed = tile->trafficspeed(edge);
+    if (traffic_speed.breakpoint1 > 0) {
+      cuts.emplace_back(cut_t{traffic_speed.breakpoint1 / 255.0, speed,
+                              static_cast<std::uint8_t>(traffic_speed.congestion1)});
+      if (traffic_speed.breakpoint2 > 0) {
+        cuts.emplace_back(cut_t{traffic_speed.breakpoint2 / 255.0, speed,
+                                static_cast<std::uint8_t>(traffic_speed.congestion2)});
+        if (traffic_speed.speed3 != UNKNOWN_TRAFFIC_SPEED_RAW) {
+          cuts.emplace_back(cut_t{1, speed, static_cast<std::uint8_t>(traffic_speed.congestion3)});
```
kevinkreiser, line 183: this bit here hasnt changed, other than `tuple->cut_t`, it just got pulled out of an `if` so it shows as a diff

src/thor/triplegbuilder.cc
```diff
@@ -91,93 +91,190 @@ void AssignAdmins(const AttributesController& controller,
...
+  if (cuts.empty() || cuts.back().percent_along < tgt_pct) {
+    cuts.emplace_back(cut_t{tgt_pct, speed, UNKNOWN_CONGESTION_VAL, {}});
+  }
 
-      // Set shape attributes time per shape point if requested
-      if (controller.attributes.at(kShapeAttributesTime)) {
-        // convert time to milliseconds and then round to an integer
-        trip_path.mutable_shape_attributes()->add_time((time * kMillisecondPerSec) + 0.5);
+  // sort the start and ends of the incidents along this edge
+  for (const auto& i : incidents) {
```
kevinkreiser, line 195: this bit is new. here we add more "cut locations" based on where incident occur on the edge, if any. have a read over it, its well commented

src/thor/triplegbuilder.cc
```diff
@@ -91,93 +91,190 @@ void AssignAdmins(const AttributesController& controller,
...
 
-      // If we just cut the shape we need to go on to the next marker only after setting the attribs
-      std::advance(speed_itr, shift);
+    // Set shape attributes speed per shape point if requested
+    if (controller.attributes.at(kShapeAttributesSpeed)) {
+      // convert speed to decimeters per sec and then round to an integer
+      leg.mutable_shape_attributes()->add_speed((distance * kDecimeterPerMeter / time) + 0.5);
     }
+
+    // Set the incidents if we just cut or we are at the end
```
kevinkreiser, line 269: this loop in the AddShapeAttributes function is what does the cutting and setting of attributes. The only real difference is here where we add incidents to the legs.

src/thor/triplegbuilder.cc
```diff
@@ -788,97 +885,9 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
...
   }
 
-  if (controller.attributes.at(kNodeIncidents)) {
-    addIncidents(*trip_edge, graphreader, edge);
-  }
-
   return trip_edge;
 }
 
-/**
```
kevinkreiser, line 798: moved to private header

src/thor/triplegbuilder.cc
```diff
@@ -1058,13 +1068,7 @@ void TripLegBuilder::Build(
   std::unordered_map<AdminInfo, uint32_t, AdminInfo::AdminInfoHasher> admin_info_map;
   std::vector<AdminInfo> admin_info_list;
 
-  // initialize shape_attributes
```
kevinkreiser, line 1061: moved inside function

src/thor/triplegbuilder.cc
```diff
@@ -1247,20 +1248,30 @@ void TripLegBuilder::Build(
       }
     }
 
+    // Set the portion of the edge we used
+    // TODO: attributes controller and then use this in recosting
+    trip_edge->set_source_along_edge(trim_start_pct);
+    trip_edge->set_target_along_edge(trim_end_pct);
```
kevinkreiser, line 1254: foreshadowing for that issue with recosting and uturns et al

src/thor/triplegbuilder.cc
```diff
@@ -1247,20 +1248,30 @@ void TripLegBuilder::Build(
...
     }
 
-    // Set shape attributes
+    // How long on this edge?
     auto edge_seconds = edge_itr->elapsed_cost.secs - edge_itr->transition_cost.secs;
     if (edge_itr != path_begin)
       edge_seconds -= std::prev(edge_itr)->elapsed_cost.secs;
+
+    // Set shape attributes, sending incidents enables them in the pbf
+    auto incidents = controller.attributes.at(kIncidents)
```
kevinkreiser, line 1269: this is where we call out to graphreader to get incidents if requested

src/thor/triplegbuilder.cc
```diff
@@ -1276,75 +1287,19 @@ void TripLegBuilder::Build(
     // must be done after the edge's shape has been added.
     SetHeadings(trip_edge, controller, directededge, trip_shape, begin_index);
 
-    // Add connected edges from the start node. Do this after the first trip
```
kevinkreiser, line 1279: all of this got moved to the private header

test/gurka/test_route_incidents.cc
```diff
@@ -0,0 +1,685 @@
...
+
+    // generate the lls for the nodes in the map
+    const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10, {.2f, .2f});
+
+    // make the tiles
+    std::string tile_dir = "test/data/route_incidents";
+    map = gurka::buildtiles(layout, ways, {}, {}, tile_dir,
+                            {
+                                {"mjolnir.traffic_extract", tile_dir + "/traffic.tar"},
+                                {"mjolnir.simulate_incidents", "true"},
```
kevinkreiser, line 48: enable traffic and simulation of incidents, so that graphreader functions "work"

test/gurka/test_route_incidents.cc
```diff
@@ -0,0 +1,685 @@
...
+
+// for asserting where the incident starts and ends on the route and how long it is
+struct incident_location {
+  uint64_t id;
+  uint32_t leg_index;
+  uint32_t begin_edge_index;
+  double begin_pct;
+  uint32_t end_edge_index;
+  double end_pct;
+};
```
kevinkreiser, line 67: here we store our test cases. this allows us to say we should expect an incident with such an id, on such leg index, starting on this edge index of the leg at this percent along the edge and ending at this edge index of the leg at this percent along

kevinkreiser: partial edges makes this hard to reason about but if you concentrate its doable

test/gurka/test_route_incidents.cc
```diff
@@ -0,0 +1,685 @@
...
+
+void check_incident_locations(const valhalla::Api& api,
```
kevinkreiser, line 69: this is the work horse of the tests. it checks a bunch of stuff to make sure the incidents output is correct. please read over it

test/gurka/test_route_incidents.cc
```diff
@@ -0,0 +1,685 @@
...
+        << "Incident " << i.id << " on leg " << i.leg_index << " on edge " << i.end_edge_index
+        << " ends at wrong location along the edge";
+  }
+
+  ASSERT_EQ(incident_count, locations.size()) << "Expected number of incidents does not match actual";
+}
+
+// provides us an easy way to mock having incident tiles, each test can override the tile in question
+// a bit more work is needed if we want to do it for more than one tile at a time
+struct test_reader : public baldr::GraphReader {
```
kevinkreiser, line 144: this is what makes it possible to actually test this without incident data. we make a testable graph reader that allows  us to directly modify the incident tile that it uses to answer the GetIncidents function when triplegbuilder calls it. at the moment it only has one tile. which makes it not quite good enough, somehow though i got lucky and the edge indices within the tiles didnt collide. i can expand this to support multiple tiles at a time though to remove that issue if reviewers feel that it is worth it

test/gurka/test_route_incidents.cc
```diff
@@ -0,0 +1,685 @@
...
+                  return a.start_offset < b.start_offset;
+                }
+                return a.edge_index < b.edge_index;
+              });
+  }
+  std::shared_ptr<baldr::GraphReader::incident_tile_t> incident_tile_;
+};
+
+// used to make a graphreader and mark some edges as having incidents
+std::shared_ptr<test_reader> setup_test(const gurka::map& map,
```
kevinkreiser, line 171: this is boilerplate for each test case. it sets up the testable reader and makes the edges that you requested have traffic that says they have incidents so that the graphreader will try to look up their incidents. it returns the reader and the ids of the edges so that the test can use the ids to modify the readers incident data accordingly

test/gurka/test_route_incidents.cc
```diff
@@ -0,0 +1,685 @@
...
+  // do the route
+  auto result = gurka::route(map, {"C", "B"}, "auto",
+                             {{"/filters/action", "include"}, {"/filters/attributes/0", "incidents"}},
+                             graphreader);
+
+  // check its right
+  check_incident_locations(result, {
+                                       {1234, 0, 0, .25, 0, .75},
+                                   });
+}
```
kevinkreiser, line 216: this is the basic layout of each test. turn on incidents for your edgeids, make the incidents in the incident tile, get a route with incidents turned on. then reason about what the incident associated to the leg will look like and check that it looks that way

test/gurka/test_route_incidents.cc
kevinkreiser: (i did expand it to support multiple tiles at a time in the latest commit so its not an issue anymore)

src/thor/triplegbuilder.cc
PureW: Switching to "ignore whitespace changes" in diff makes this much easier to follow

src/thor/triplegbuilder.cc
PureW: Could `i` be named `incident`? Makes it so much easier to read.

kevinkreiser: sure thing!

src/tyr/route_serializer_osrm.cc
```diff
@@ -682,13 +648,35 @@ valhalla::baldr::json::MapPtr serializeIncident(const TripLeg_Node_Incident& inc
   if (incident.type()) {
     metadata_json->emplace("incident_type", valhalla::incidentTypeToString(incident.type()));
```
danpat, line 649: Can we just rename this to `type` - it's already in the `incidents[]` array, this field doesn't need to be prefixed with `incident_`

danpat (changes requested):
Should change `incident_type` JSON field to just `type`

danpat (approved):

PR #2613, opened 2020-09-24
Activate directional turn lanes within step

danpaz:
# Issue

Activates turn lanes on all edges along a maneuver using the existing lane activation heuristics.

This is useful when turn lanes are marked at intersections prior to the transition point, where getting into the correct turn lane earlier on in the maneuver helps to make the correct transition. 

turn_lane_routes.txt:40
**BEFORE**

image

**AFTER**
image

turn_lane_routes.txt:105

**BEFORE**
image

**AFTER
image

## Tasklist

 - [X] Add tests

## Requirements / Relations


src/odin/maneuver.cc
```diff
@@ -166,6 +166,27 @@ bool Maneuver::IsMergeType() const {
           (type_ == DirectionsLeg_Maneuver_Type_kMergeRight));
 }
 
+bool Maneuver::IsLeftType() const {
+  return ((type_ == DirectionsLeg_Maneuver_Type_kUturnLeft ||
```
mandeepsandhu, line 170: I feel there's some scope of code reuse with whats being done [here](https://github.com/valhalla/valhalla/blob/master/src/odin/maneuversbuilder.cc#L2983). Maybe have the lambda test for `maneuver.IsLeftType() || maneuver. IsRightType()`?

mandeepsandhu: Hmmm, or maybe not. I see that some of the types used here are not considered a "turn" in the `is_turn_type` lambda.

src/odin/maneuversbuilder.cc
```diff
@@ -48,6 +48,8 @@ constexpr float kShortContinueThreshold = 0.6f;
 
 constexpr uint32_t kOverlayEdgeMax = 5; // Maximum number of edges to look for matching overlay
 
+constexpr float kUpcomingLanesThreshold = 3.f; // Kilometers
```
dgearhart, line 51: I wonder if we will need to use duration based on speed with a minimum distance

src/odin/maneuversbuilder.cc
```diff
@@ -2812,55 +2815,96 @@ uint16_t ManeuversBuilder::GetExpectedTurnLaneDirection(Maneuver& maneuver) cons
 void ManeuversBuilder::ProcessTurnLanes(std::list<Maneuver>& maneuvers) {
   auto curr_man = maneuvers.begin();
   auto next_man = maneuvers.begin();
+  auto next_next_man = maneuvers.begin();
```
dgearhart, line 2818: I need to step through this in more detail. I was thinking after we talked - to use the previous, current, and next maneuver pattern instead of the `next next maneuver`

danpaz: Made that change. Can you take another look?

dgearhart:
@danpaz thanks - looks much better with the prev, curr, next approach. 
I think we should instrument to RAD test

src/odin/maneuversbuilder.cc
```diff
@@ -2810,55 +2813,97 @@ uint16_t ManeuversBuilder::GetExpectedTurnLaneDirection(Maneuver& maneuver) cons
...
+
+      // Keep track of the remaining step distance in kilometers
+      float remaining_step_distance = 0.f;
+      // Add the prev_edge length as we skip it in the loop below
+      if (prev_edge) {
+        remaining_step_distance += prev_edge->length();
+      }
+
+      // Assign turn lanes within step, walking backwards from end to begin node
+      for (auto index = (prev_man->end_node_index() - 1); index > prev_man->begin_node_index();
```
mandeepsandhu, line 2863: We're not considering the first & end node indexes here. Why is that? I guess my confusion stems for not being able to visualize how these indexes appear on a road segment with turn lanes.

mandeepsandhu, line 2863: nit: instead of `auto` specify the type, in this case, `uint32_t`. I don't know if its possible due to protections up the call-chain, but if `prev_man->end_node_index()` is ever 0, bad juju could happen. Having the type explicit here helps in highlighting the fact that we _could_ have a very large value if the end index was somehow 0.

(2️⃣¢)

test/gurka/test_turn_lanes.cc
```diff
@@ -43,3 +59,240 @@ TEST(Standalone, TurnLanes) {
...
+  // A -> E - takes the leftmost through lane
+  result = gurka::route(map, "A", "E", "auto");
+  validate_turn_lanes(result, {
+                                  {3, "[ left | left;through ACTIVE | through;right ]"},
+                                  {0, ""}, // TODO lanes are tossed when all are through
+                              });
+
+  // A -> G - takes the rightmost through lane
+  result = gurka::route(map, "A", "G", "auto");
+  gurka::assert::osrm::expect_steps(result, {"Bronx and Pelham Parkway", "Pelham Parkway South"});
```
mandeepsandhu, line 117: Why does only this route have extra validations for steps & maneuvers but others don't?

test/gurka/test_turn_lanes.cc
```diff
@@ -43,3 +59,240 @@ TEST(Standalone, TurnLanes) {
...
+                            {"EG",
+                             {{"highway", "primary"},
+                              {"lanes:forward", "2"},
+                              {"turn:lanes:forward", "through|through"}}}};
+
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_turn_lanes_3");
+
+  // A -> G  - only through lanes should be active throughout
+  auto result = gurka::route(map, "A", "C", "auto");
```
mandeepsandhu, line 166: Looks like this route should be `A->G`? `A->C` is repeated otherwise.

danpaz: Good catch! Fixed in https://github.com/valhalla/valhalla/pull/2613/commits/d83764a10d2bade40134078f6a364e17d0875a7d

test/gurka/test_turn_lanes.cc
danpaz: This test map had street names to help distinguish the two parallel roads, so I used those to verify that the correct parallel road was taken.

src/odin/maneuversbuilder.cc
danpaz: Fixed :+1:

mandeepsandhu (approved):

src/odin/maneuversbuilder.cc
dgearhart: @danpaz are we updating this later?

danpaz: @dgearhart yeah, this is just a starting point. We can tweak this threshold later or even make it configurable in the future.

dgearhart (approved):
nice job @danpaz :shipit:

PR #2616, opened 2020-09-24
blend current speed with other speed sources when there is only partial coverage

kevinkreiser:


valhalla/baldr/graphtile.h
```diff
@@ -521,14 +521,32 @@ class GraphTile {
...
     if ((flow_mask & kCurrentFlowMask) && traffic_tile()) {
       auto directed_edge_index = std::distance(const_cast<const DirectedEdge*>(directededges_), de);
       auto volatile& live_speed = traffic_tile.trafficspeed(directed_edge_index);
       // only use current speed if its valid and non zero, a speed of 0 makes costing values crazy
-      uint8_t overall_speed;
-      if (live_speed.valid() && (overall_speed = live_speed.get_overall_speed()) > 0) {
+      if (live_speed.valid() && (partial_live_speed = live_speed.get_overall_speed()) > 0) {
         *flow_sources |= kCurrentFlowMask;
-        return overall_speed;
+        // Live speed covers entire edge, can return early here
```
kevinkreiser, line 532: comment isnt exactly correct, but its the most common case

PureW (approved):

PR #2617, opened 2020-09-25
More reach fixes

kevinkreiser:
well there was one more spot where we accidentally modified the tile out from under the filter method. this pr fixes that and also adds more error handling in graphtile

### DIFF
--- a/src/baldr/graphtile.cc
+++ b/src/baldr/graphtile.cc
@@ -571,33 +571,38 @@ AABB2<PointLL> GraphTile::BoundingBox() const {
 }
 
 iterable_t<const DirectedEdge> GraphTile::GetDirectedEdges(const NodeInfo* node) const {
+  if (node < nodes_ || node >= nodes_ + header_->nodecount()) {
+    throw std::logic_error(
+        std::string(__FILE__) + ":" + std::to_string(__LINE__) +
+        " GraphTile NodeInfo out of bounds: " + std::to_string(header_->graphid()));
+  }
   const auto* edge = directededges_ + node->edge_index();
   return iterable_t<const DirectedEdge>{edge, node->edge_count()};
 }
 
 iterable_t<const DirectedEdge> GraphTile::GetDirectedEdges(const GraphId& node) const {
-  if (node.id() < header_->nodecount()) {
-    const auto* nodeinfo = nodes_ + node.id();
-    return GetDirectedEdges(nodeinfo);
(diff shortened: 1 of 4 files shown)


src/baldr/graphtile.cc
```diff
@@ -571,33 +571,38 @@ AABB2<PointLL> GraphTile::BoundingBox() const {
 }
 
 iterable_t<const DirectedEdge> GraphTile::GetDirectedEdges(const NodeInfo* node) const {
```
kevinkreiser, line 573: changed all these to:

1. logic_error because its definitly that vs a runtime_error
2. put the error check at the top of the method in the if block so the default is not branch inward

src/loki/reach.cc
```diff
@@ -13,6 +13,28 @@ Reach::Reach() : Dijkstras() {
   path_edge->set_end_node(false);
 }
 
+void Reach::enqueue(const baldr::GraphId& node_id,
+                    baldr::GraphReader& reader,
+                    const std::shared_ptr<sif::DynamicCost>& costing,
+                    const baldr::GraphTile* tile) {
```
kevinkreiser, line 19: i moved this method up for visibility since its called right below here. the only change was to not pass the pointer by reference so that when the end node that we are enqueing is in another tile, we wont modify it. modifying it screws over the costing->filter call that happens inside the node expansion loop in the operator() below

valhalla/baldr/graphtile.h
```diff
@@ -260,6 +260,11 @@ class GraphTile {
    * @return returns an iterable collection of node transitions
    */
   midgard::iterable_t<const NodeTransition> GetNodeTransitions(const NodeInfo* node) const {
```
kevinkreiser, line 262: just some more of these but this time in the header

valhalla/loki/reach.h
```diff
@@ -75,7 +75,7 @@ class Reach : public thor::Dijkstras {
   void enqueue(const baldr::GraphId& node_id,
                baldr::GraphReader& reader,
                const std::shared_ptr<sif::DynamicCost>& costing,
-               const baldr::GraphTile*& tile);
+               const baldr::GraphTile* tile);
```
kevinkreiser, line 78: this is the fix, we dont let enqueue modify the pointer outside of its scope. if we do the filter function can get the wrong tile in the next loop iteration: https://github.com/valhalla/valhalla/pull/2597/files#diff-71585ca21c04f5620b9c09883c71b910R65

PureW (approved):

genadz (approved):

PR #2620, opened 2020-09-28
Defines the incident tile schema and incident metadata

PureW:
from protobuf schema.

Adopts the triplegbuilder.cc to use this new format.

An incident tile looks like the following due to the sparseness of
incidents. There is an array `edge_to_incidents` which relates
graph-edges to a respective fat `Incident` struct in the `incidents`
array.

```
 message IncidentsTile {
   // Sorted list of edge_ids describing what incidents are attached to an edge_id
   repeated EdgeToIncident edge_to_incidents = 1;
   // Look at `edge_to_incidents` to find how to index this array
   repeated Incident incidents = 2;
 }
```


proto/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ protobuf_generate_cpp(protobuff_srcs protobuff_hdrs
   trip.proto
   transit.proto
   transit_fetch.proto
+  sideloaded.proto
```
kevinkreiser, line 10: nit: `incident_tile.proto`

proto/sideloaded.proto
```diff
@@ -0,0 +1,72 @@
...
+
+syntax = "proto3";
+option optimize_for = LITE_RUNTIME;
+package valhalla_sideloaded;
+
+message IncidentsTile {
+  // Sorted list of edge_ids describing what incidents are attached to an edge_id
+  repeated EdgeToIncident edge_to_incidents = 1;
+  // Look at `edge_to_incidents` to find how to index this array
+  repeated Incident incidents = 2;
```
kevinkreiser, line 16: i think this would be clearer with some naming changes
```
repeated Incident incidents = 1
repeated MetaData metadata = 2;
```

proto/sideloaded.proto
```diff
@@ -0,0 +1,72 @@
+// File contains definitions for side-loaded data-structures
+//
+// Because of the more distributed nature of this data
+// compared to the previous protobuf structures used in Valhalla
+// it is important to follow the protobuf schema evolution rules
+// when modifying this file
+
+syntax = "proto3";
+option optimize_for = LITE_RUNTIME;
+package valhalla_sideloaded;
```
kevinkreiser, line 10: should be the valhalla namespace like all the rest of our protobuf

proto/trip.proto
```diff
@@ -304,30 +305,11 @@ message TripLeg {
...
-    optional Type type = 1;
-    optional string description = 2;
-    optional uint64 start_time = 3;     // START_TIME as seconds since epoch UTC
-    optional uint64 end_time = 4;       // END_TIME seconds since epoch UTC
-    optional uint64 creation_time = 5;  // time the incident was created/updated in seconds since epoch UTC
-    optional uint64 id = 6;             // An id that can be used for joining incidents in route with map
-    optional uint32 begin_shape_index = 7;
-    optional uint32 end_shape_index = 8;
+  message ValhallaIncident {
+    optional valhalla_sideloaded.Incident metadata = 1;
```
kevinkreiser, line 309: how does this work!? mixing proto2 and proto3? crazy stuff. also instead of doing this...

could we add a metadata field to the edgetoincident structure, but only fill it out when the incident is fetched in the graphreader. then we could just copy the incident verbatim. i mean maybe the secondary look up for the metadata is also overkill im not sure. just some ideas to simplify this

proto/sideloaded.proto
PureW: `Incident` -> `IncidentLocation`?

src/baldr/graphreader.cc
```diff
@@ -880,5 +880,53 @@ int GraphReader::GetTimezone(const baldr::GraphId& node, const GraphTile*& tile)
...
+
+  int begin_index = begin - itile->incident_locations().begin();
+  int end_index = end - itile->incident_locations().begin();
+
+  if (begin_index >= itile->incident_locations_size()) {
+    // No incidents
+    return {nullptr, 0, 0};
+  } else {
+    return {itile, begin_index, end_index};
+  }
```
kevinkreiser, line 916: no need to add an `if` here, both indices will be `itile->incident_locations_size()` when the edge index is not found

src/thor/triplegbuilder.cc
```diff
@@ -98,21 +98,38 @@ void AssignAdmins(const AttributesController& controller,
...
  */
-void UpdateIncident(TripLeg& leg, const baldr::GraphReader::Incident* incident, uint32_t index) {
+void UpdateIncident(const std::shared_ptr<valhalla::incidents::IncidentsTile>& incidents_tile,
+                    TripLeg& leg,
+                    const valhalla::incidents::IncidentLocation* incident_location,
+                    uint32_t index) {
+  if (!incidents_tile) {
+    // No incident tile
+    return;
+  }
```
kevinkreiser, line 108: why would this ever happen? surely it cannot

src/thor/triplegbuilder.cc
```diff
@@ -98,21 +98,38 @@ void AssignAdmins(const AttributesController& controller,
...
+  auto candidate_index = 0;
```
PureW, line 109: Remove.

proto/incidents.proto
```diff
@@ -0,0 +1,72 @@
...
+    }
+    Congestion congestion = 11;
+    repeated string lanes_blocked = 12;
+    uint64 creation_time = 13;
+    string long_description = 14;
+    string clear_lanes = 15;
+    uint64 num_lanes_blocked = 16;
+
+    // IncidentMetadata id
+    uint64 id = 128;
```
kevinkreiser, line 71: can we move this out of metadata and into the incident itself? that should further solidify it as something that is unique to each incident. i know it means duplicating it for incidents that span more than one edge but its nice to be a top level feature

kevinkreiser: ah right we schlepp this along into the tripleg and need to look it up again later

valhalla/baldr/graphreader.h
```diff
@@ -821,24 +823,7 @@ class GraphReader {
...
-    // get the range of incidents we care about and hand it back, equal_range has no lambda option
-    auto begin = std::partition_point(itile->begin(), itile->end(), [&edge_id](const Incident& i) {
-      return i.edge_index < edge_id.id(); // first one that is >= the id we want
-    });
-    auto end = std::partition_point(begin, itile->end(), [&edge_id](const Incident& i) {
-      return i.edge_index <= edge_id.id(); // first one that is > the id we want
-    });
-    return std::vector<Incident>(begin, end);
-  }
+  IncidentResult GetIncidents(const GraphId& edge_id, const GraphTile*& tile);
```
kevinkreiser, line 826: docstring above needs to be updated, since it doesnt return a vector but rather a range and a tile

src/thor/triplegbuilder.cc
```diff
@@ -142,13 +151,14 @@ void SetShapeAttributes(const AttributesController& controller,
...
                         double edge_seconds,
                         bool cut_for_traffic,
-                        const std::vector<baldr::GraphReader::Incident>& incidents) {
+                        const valhalla::baldr::IncidentResult& incidents) {
   // TODO: if this is a transit edge then the costing will throw
 
   // bail if nothing to do
-  if (!cut_for_traffic && incidents.empty() &&
-      !controller.category_attribute_enabled(kShapeAttributesCategory))
+  if (!cut_for_traffic && incidents.tile == nullptr &&
```
kevinkreiser, line 158: should be `incidents.start_index == incidents.end_index`

src/thor/triplegbuilder.cc
```diff
@@ -191,35 +201,45 @@ void SetShapeAttributes(const AttributesController& controller,
...
-    for (auto offset : {
-             std::max(incident.start_offset, src_pct),
-             std::min(incident.end_offset, tgt_pct),
-         }) {
-      // if this is clipped at the beginning of the edge then its not a new cut but we still need to
-      // attach the incidents information to the leg
-      if (offset == src_pct) {
-        UpdateIncident(leg, &incident, shape_begin);
-        continue;
+  if (incidents.tile) {
```
kevinkreiser, line 204: not needed

kevinkreiser: if you really want it though you could put it in the while part of the for loop below

test/gurka/test_route_incidents.cc
```diff
@@ -83,57 +84,61 @@ void check_incident_locations(const valhalla::Api& api,
...
   }
 
   // check all the incidents are in the right place in the route
-  for (const auto& i : locations) {
-    const auto& leg = api.trip().routes(0).legs(i.leg_index);
+  for (const auto& location : locations) {
+    const auto& leg = api.trip().routes(0).legs(location.leg_index);
     // find the incident
-    const valhalla::TripLeg::Incident* incident = nullptr;
+    const valhalla::TripLeg::ValhallaIncident* incident = nullptr;
```
kevinkreiser, line 98: should have named this encapsulation just `Incident` since its namespaced to `TripLeg`

test/gurka/test_route_incidents.cc
```diff
@@ -145,37 +150,82 @@ struct test_reader : public baldr::GraphReader {
...
+    valhalla::incidents::Metadata* metadata = nullptr;
+    // Check if this incident is already tracked
+    for (auto& meta : *incidents_tile.mutable_metadata()) {
+      if (incident_id == meta.id()) {
+        // We've already added the incident, use it instead of adding again
+        metadata = &meta;
+        break;
+      }
+      metadata_index += 1;
+    }
```
kevinkreiser, line 180: why was this necessary? just in case we had a typo in the tests and used the same id twice?

kevinkreiser (dismissed):

kevinkreiser (approved):

PR #2621, opened 2020-09-30
Moves incident serializer logic into a generic serializer

PureW:


src/loki/serializer_common.cc
```diff
@@ -0,0 +1,104 @@
+#include "loki/serializer_common.h"
```
kevinkreiser, line 1: move to tyr with the rest of the serializer methods

src/proto_conversions.cc
```diff
@@ -51,6 +54,43 @@ std::string incidentTypeToString(const valhalla::incidents::Metadata::Type& inci
...
+    case valhalla::incidents::Metadata_Impact_Metadata_Impact_INT_MIN_SENTINEL_DO_NOT_USE_:
+      // Like the name says, do not use. Simply for ensuring full coverage of switch statement
+      break;
+  }
+  // TODO Throw or warn here? Assert maybe to only crash debug build
+  LOG_WARN("Unhandled case in incidentCriticalityToString: " + std::to_string(impact));
+  return "UNHANDLED_CASE";
+}
+
+std::string time_to_string(std::uint64_t seconds_since_epoch) {
```
kevinkreiser, line 85: i would be surprised if we didnt already have this in `baldr::datetime` or some other util somewhere.. in any case this problem belongs there

danpat (dismissed):

kevinkreiser (approved):

PR #2626, opened 2020-10-03
Alternate Route Support

### LINKED ISSUE #1305: Add alternative routes to the interface
## Context
Currently, there is noway to request and receive alternative routes from Valhalla.

## Task
Stub ticket - will add more information as we discuss

### LINKED ISSUE #2580: Need Alternate Route Support in Valhalla
Valhalla is excellent routing engine and i just started using it is quite good it has detailed features and parameters.
I would like to know is there any option for alternate route as OSRM has, In which it creates separate legs for alternative paths.
[As you can see here](http://map.project-osrm.org/?z=13&center=37.718726%2C-122.432585&loc=37.760537%2C-122.455673&loc=37.703109%2C-122.419453&hl=en&alt=0&srv=1)

kevinkreiser:
The brings alternate route support, sort of, to valhalla. At the moment the valhalla reponse format does not support more than one trip object so we'll need to figure out how to move forward on that design mistake (maybe nesting alternates inside the trip? yuck..) In the mean time the OSRM response format does support alternates natively (excellent foresight in that design).

The changes required here were very minor ones to bidirectional a*, basically just to track more than one connection point from the two search trees. The bulk of the work here was that of judging the viability of a path compared to the already found paths. This work is basically the light version of that which is already found in OSRM which was originally described in a paper by Moritz Kobitzsch. The code comments describe the techniques. But we only implemented 2 out of 3. The third would be quite complex for us to track given our algorithm/datastructures and so we left it out.

In addition to that there are unit tests to show that alternates are at least coming out of the engine.

fixes #2580
fixes #1305


src/thor/alternates.cc
```diff
@@ -0,0 +1,128 @@
...
+ * Viability tests for alternate paths based on M. Kobitzsch's Alternative Route
+ * Techniques (2015). Tests verify limited sharing between segments, bounded
+ * stretch, and local optimiality. Any candidate path that meets all the criteria
+ * may be considered a valid alternate to the shortest path.
+ */
+namespace {
+// Defaults thresholds
+float kAtMostLonger = 1.25f; // stretch threshold
+float kAtMostShared = 0.75f; // sharing threshold
+// float kAtLeastOptimal = 0.2f; // local optimality threshold
```
kevinkreiser, line 20: we didnt implement local optimality and the paper actual mentions that sometimes it removes some good results. at the end of the day an implementation that supports that was too painful to stomache at the moment.

src/thor/bidirectional_astar.cc
```diff
@@ -95,8 +96,8 @@ void BidirectionalAStar::Init(const PointLL& origll, const PointLL& destll) {
   // searches.
   cost_diff_ = mincostf - mincostr;
 
-  // Initialize best connection with max cost
-  best_connection_ = {GraphId(), GraphId(), std::numeric_limits<float>::max()};
+  // Initialize best connections as having none
+  best_connections_ = {};
```
kevinkreiser, line 100: this is the main change to the algorithm. instead of keeping track of one connection, we keep track of the a set of the best connections.

src/thor/bidirectional_astar.cc
```diff
@@ -917,129 +918,188 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader, const valhalla
...
-    if (edgelabel.use() == Use::kFerry) {
-      has_ferry_ = true;
-    }
+  LOG_DEBUG("Found connections before stretch filter: " + std::to_string(best_connections_.size()));
+
+  if (allow_alternates) {
+    desired_paths += options.alternates();
+    // Cull alternate paths longer than maximum stretch
+    // TODO: we should skip adding the connection at all if it's greater than stretch
+    filter_alternates_by_stretch(best_connections_);
```
kevinkreiser, line 936: this is where we check the first viability critereon, removing any paths that are so much less optimal that we dont consider them

src/thor/bidirectional_astar.cc
```diff
@@ -917,129 +918,188 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader, const valhalla
...
 
-    // Update edgelabel_index and transition cost to apply at next iteration
-    edgelabel_index = predidx;
-    // We apply the turn cost at the beginning of the edge, as is done in the forward path
-    // Semantically this can be thought of is, how much time did it take to turn onto this edge
-    // To do this we need to carry the cost forward to the next edge in the path so we cache it here
-    previous_transition_cost = edgelabel.transition_cost();
+    // For the first path just add it for subsequent paths only add if it passes viability tests
+    if (paths.empty() ||
+        (validate_alternate_by_sharing(graphreader, shared_edgeids, paths, path, max_sharing) &&
```
kevinkreiser, line 1097: this is where we validate the second criterion by ignoring paths that are too similar to already found ones

src/thor/bidirectional_astar.cc
```diff
@@ -917,129 +918,188 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader, const valhalla
...
-      path.back().elapsed_cost.secs = path[path.size() - 2].elapsed_cost.secs +
-                                      edgelabels_reverse_[idx2].cost().secs +
-                                      edgelabels_forward_[idx1].transition_cost().secs;
-      path.back().elapsed_cost.cost = path[path.size() - 2].elapsed_cost.cost +
-                                      edgelabels_reverse_[idx2].cost().cost +
-                                      edgelabels_forward_[idx1].transition_cost().cost;
-      return paths;
+  // we quit making paths as soon as we've reached the number of paths
+  // that were requested or we run out of paths that we can actually make
+  std::vector<std::vector<PathInfo>> paths;
```
kevinkreiser, line 964: the rest of this method looks like a big diff, but only because its in a for loop for each alternative. the guts of the for loop is pretty much the same as the old code

danpaz (approved):
LGTM

src/thor/alternates.cc
```diff
@@ -0,0 +1,128 @@
...
+/*
+ * Viability tests for alternate paths based on M. Kobitzsch's Alternative Route
+ * Techniques (2015). Tests verify limited sharing between segments, bounded
+ * stretch, and local optimiality. Any candidate path that meets all the criteria
+ * may be considered a valid alternate to the shortest path.
+ */
+namespace {
+// Defaults thresholds
+float kAtMostLonger = 1.25f; // stretch threshold
+float kAtMostShared = 0.75f; // sharing threshold
```
danpaz, line 19: Noting that we chose these thresholds based on OSRM. It may be worth tweaking them in the future, or even exposing a request option to customize them: https://github.com/Project-OSRM/osrm-backend/blob/master/src/engine/routing_algorithms/alternative_path_mld.cpp#L50

PR #2627, opened 2020-10-03
github actions

kevinkreiser:
messing around with github actions


nilsnolde:
I can recommend [ci skip] for commit messages, won’t trigger circleci then. Azure uses smth else, on mobile right now or I’d check for you.

mloskot:
AzP native is NO_CI and Microsoft implemented [ci skip] too but there were issues with that, full story https://github.com/Microsoft/azure-pipelines-agent/issues/1270

nilsnolde:
In Germany that mostly means you're lucky if Google returns a search result within 5 mins.

mloskot:
It's worth to mention, that `[ci skip]` (or equivalents) are processed differently by various CI services:
- Travis: skip command *anywhere* in the commit message, https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build
- AppVeyor: skip command in the *title* (subject) of the commit message, https://www.appveyor.com/docs/how-to/filtering-commits/
- CircleCI: in the *first line of the body* of the commit or the commit’s *title*, https://circleci.com/docs/2.0/skip-build/

It's a mess :-)

kevinkreiser:
this was a nice experiment but the free tier of this is just too slow. its nice to have as an example but at the end of the day if we want to have faster builds we should probably host our own. cloud CI is always too much of a premium.. i would be happy to host our CI myself via: https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

once the new am5 platform comes out i can convert my am4 system to be a valhalla code builder :smile:

PR #2631, opened 2020-10-05
Incident loading tests

kevinkreiser:
This PR adds:

1. documentation about how incidents work
2. unit tests of the incident loading functionality


src/baldr/incident_singleton.h
```diff
@@ -26,41 +26,6 @@ namespace {
 constexpr time_t DEFAULT_MAX_LOADING_LATENCY = 60;
 constexpr size_t DEFAULT_MAX_LATENT_COUNT = 5;
 
-/**
```
kevinkreiser, line 29: moved these to static members so i could unit test them

src/baldr/incident_singleton.h
```diff
@@ -88,7 +53,10 @@ struct incident_singleton_t {
    */
   incident_singleton_t(const boost::property_tree::ptree& config,
                        const std::unordered_set<valhalla::baldr::GraphId>& tileset)
-      : state{new state_t{}}, watcher(watch, config, std::cref(tileset), state) {
+      : state{new state_t{}}, watcher(watch, config, tileset, state, interrupt()) {
+    // let the thread control its own lifetime
+    watcher.detach();
```
kevinkreiser, line 58: we need to detach right away so that if we throw below the thread is already disowned

src/baldr/incident_singleton.h
```diff
@@ -97,24 +65,66 @@ struct incident_singleton_t {
...
+  }
+
+  /**
+   * The default interrupt for the watcher threads main loop. Since its not set, the watcher loop
+   * will run forever. This method is provided so the unit tests can exercise the constructor
+   * @return
+   */
+  virtual std::function<bool(size_t)> interrupt() {
+    return {};
+  }
```
kevinkreiser, line 77: this was just for unit testing purposes

src/baldr/incident_singleton.h
```diff
@@ -97,24 +65,66 @@ struct incident_singleton_t {
...
+
+  /**
+   * Read the contents of a file into an incident tile
+   * @param filename   name of the file on the file system to read into memory
+   * @return a shared pointer with the data of the tile or an empty pointer if it could not be read
+   */
+  static std::shared_ptr<const valhalla::IncidentsTile> read_tile(const std::string& filename) {
```
kevinkreiser, line 84: moved this to here. nothing fancy

src/baldr/incident_singleton.h
```diff
@@ -97,24 +65,66 @@ struct incident_singleton_t {
...
+    // hand back something that isnt modifyable
+    return std::const_pointer_cast<const valhalla::IncidentsTile>(tile);
   }
 
   /**
    * Updates the tile in the states cache
    * @param state     the state to update
    * @param tile_id   the tile id we are loading
    * @param path      the path to the tile file
+   * @param hint      a pointer to an existing iterator into the cache
```
kevinkreiser, line 119: this was added for performance in the detecting of removed tiles, explained below

src/baldr/incident_singleton.h
```diff
@@ -149,20 +159,22 @@ struct incident_singleton_t {
    * The thread begins by deciding whether its just scaning the directory (works for a small number of
    * incidents) or using a memory mapped changelog to communicate about which incidents chnaged last.
    * If the latter is used we preallocate
-   * @param config
-   * @param tileset
-   * @param state
+   * @param config     lets the function know where to look for incidents and desired update frequency
+   * @param tileset    if the tileset is static we can use lockfree mode
+   * @param state      used for interthread communication
+   * @param interrupt  functor that, if set and returns true, stops the main loop of this function
```
kevinkreiser, line 165: in order to unit test this i needed a way to control when the function exists. in practice we never want the function to exit but for testing its quite annoying to have an infinite loop. so we allow an interrupt function here to stop the loop when desired

src/baldr/incident_singleton.h
```diff
@@ -190,11 +202,12 @@ struct incident_singleton_t {
...
 
     // some setup for continuous operation
-    bool first_run = true;
+    size_t run_count = 0;
     time_t last_scan = 0;
-    size_t latent_count = 0;
     time_t max_loading_latency =
         config.get<time_t>("incident_max_loading_latency", DEFAULT_MAX_LOADING_LATENCY);
+    std::unordered_set<uint64_t> seen;
+    seen.reserve(tileset.size());
```
kevinkreiser, line 210: when we do a scan, either of the changelog or the directory, if we track what tiles we see in the tileset. if any of them disappeared (from the changelog or from the directory) then we want to also remove them from the cache

src/baldr/incident_singleton.h
```diff
@@ -203,43 +216,64 @@ struct incident_singleton_t {
...
+      seen.clear();
 
       // we are in memory map mode
       if (changelog) {
         // reload the log if the tileset isnt static
-        if (!state->lock_free.load())
-          changelog.reset(new decltype(changelog)::element_type(inc_log_path.string()));
+        try {
+          if (!state->lock_free.load())
+            changelog.reset(new decltype(changelog)::element_type(inc_log_path.string(), false, 0));
```
kevinkreiser, line 226: this needed a try/catch

src/baldr/incident_singleton.h
```diff
@@ -203,43 +216,64 @@ struct incident_singleton_t {
...
             }
-          } catch (...) { LOG_WARN("Incident watcher ignoring " + i->path().string()); }
+          } // happens when there is a file in the directory that doesnt have a tile-looking name
+          catch (...) {}
+        }
+      }
+
+      // for all the ones we didnt see, they have been removed from the filesystem or changelog
+      // no locking is needed because we don't realloc here we just null out some values
+      for (auto entry = state->cache.begin(); entry != state->cache.end(); ++entry) {
```
kevinkreiser, line 273: here we go through and remove any tiles that went missing in the last round

src/baldr/incident_singleton.h
```diff
@@ -249,25 +283,24 @@ struct incident_singleton_t {
       auto wait = 0;
       if (latency > max_loading_latency) {
         LOG_WARN("Incident watcher is not meeting max loading latency requirement");
-      } // this round finished fast enough, changelog mode is cheap so wait only a second
+      } // this round finished fast enough wait the remainder of the time specified
       else {
-        wait = changelog ? 1 : max_loading_latency - latency;
```
kevinkreiser, line 254: this seemed like a hack. we just listen to the configuration exactly, regardless of what mode its in

src/baldr/incident_singleton.h
```diff
@@ -249,25 +283,24 @@ struct incident_singleton_t {
...
+      if (run_count++ == 0) {
         LOG_INFO("Incident watcher initialized");
-        first_run = false;
         state->signal.notify_one();
       }
 
       // wait just a little before we check again
       std::this_thread::sleep_for(std::chrono::seconds(wait));
-    } while (true);
+    } while (!interrupt || !interrupt(run_count));
```
kevinkreiser, line 301: if you set an interrupt and if the interrupt returns true on this interation, then we bail. needed for unit testing only

src/baldr/incident_singleton.h
PureW: Which in the healthy case should not happen, since tiles that had incidents before but now have 0 incidents, still update the file on disk with an empty tile.

PureW (changes requested):
Could you spell out the contract that the watcher thread expects any incident producing entity should adhere to?

src/baldr/incident_singleton.h
```diff
@@ -97,24 +65,66 @@ struct incident_singleton_t {
...
+  /**
+   * Read the contents of a file into an incident tile
+   * @param filename   name of the file on the file system to read into memory
+   * @return a shared pointer with the data of the tile or an empty pointer if it could not be read
+   */
+  static std::shared_ptr<const valhalla::IncidentsTile> read_tile(const std::string& filename) {
+    // open the file for reading. its normal for this to fail when the file has been removed
+    std::ifstream file(filename, std::ios::in | std::ios::binary);
+    if (!file.is_open()) {
+      return {};
```
PureW, line 88: Would it be good to log a warning here? I don't think this should occur during normal operations?

kevinkreiser: the comment is maybe not expressive enough. if your incident implementation removes files from the file system it can happen that you go to load the file and its been yanked before you can. or if you are using the memory mapped log, you can reset timestamp and remove the file from the filesystem and this code will be triggered. in either case its normal operation to expect the open to fail. previously i had a warning here but then i realized removing files from the filesystem is a pretty normal thing to do to decomission an incident tile

PureW (approved):

PR #2632, opened 2020-10-06
round integer coords for polyline encoding

kevinkreiser:
Once upon a time when we implemented polyline encoding and when we did, we truncated precision... I looked at the history of the file and I found that even the first implementtion I wrote was truncating precision instead of rounding. This fixes that.

I suspect this will break a bunch of tests which will have to be manually updated. Lets hope for the best :smile:


dgearhart (approved):
:shipit:

dnesbitt61 (dismissed):

kevinkreiser:
turns out, the original implementation of google polyline encoding also used `floor` which is probably why i wrote it this way in the first place... https://gist.github.com/ajturner/709233#file-encodedpolyline-js-L300-L301

i cant understand why that would be desirable... but it makes it so that pretty much all other standard encoding libraries dont round, like i the one that i linked in the unit tests back when i wrote this code: https://github.com/rcoup/py-gpolyencode/blob/master/python/gpolyencode.py#L69-L70

oh well no matter, we'll round and squeeze that error bar down by half

src/baldr/edgeinfo.cc
```diff
@@ -165,6 +165,7 @@ uint16_t EdgeInfo::GetTypes() const {
 }
 
 // Returns shape as a vector of PointLL
+// TODO: use shared ptr here so that we dont have to worry about lifetime
```
kevinkreiser, line 168: this is a comment to a future pull requestor to fix this landmind

test/encode.cc
```diff
@@ -78,23 +78,36 @@ TEST(Encode, Polyline5) {
...
 
+  // clang-format off
+/*
+
+python2 -c "
+import random;
+import gpolyencode;
+x = [ [round(random.random() * 180 - 90, 5), round(random.random() * 360 - 180, 5)] for a in range(0, random.randint(1,100)) ];
+p = gpolyencode.GPolyEncoder().encode(x)['points']; print;
+print 'do_polyline_pair(' + str(x).replace('[','{').replace(']','}') + ', \"' + repr(p)[1:-1] + '\");'; print"
```
kevinkreiser, line 101: so this actually doesnt work because their encoder floors instead of rounds but its still useful for posterity (showing how we got these tests in the first place)

test/gurka/test_search_filter.cc
```diff
@@ -17,6 +17,7 @@ class SearchFilter : public ::testing::Test {
     |   2     |
     |         ↑
     |1    4  3|
+    |         |
```
kevinkreiser, line 20: previously the 1 node was pretty much equidistant from either edge, so it was a toss up depending on precision. once precision was modified the toss up went the other way. this change makes it not a toss up

test/minbb.cc
```diff
@@ -26,16 +26,26 @@ struct bb_tester {
...
       for (const auto& node : t->GetNodes()) {
         auto node_ll = node.latlng(t->header()->base_ll());
-        if (!bb.minpt().IsValid())
+        if (!bb.minpt().IsValid()) {
           bb = AABB2<PointLL>{node_ll, node_ll};
-        else
-          bb.Expand(node_ll);
+        }
+
+        // when we see a node that pushes the envelop out we check its edge shape
```
kevinkreiser, line 39: same as before but also uses the encoded shape value for the node since there are differing precisions between them (scaled int vs float)

valhalla/midgard/aabb2.h
```diff
@@ -265,20 +265,27 @@ template <class coord_t> class AABB2 {
   /**
    * Expands (if necessary) the bounding box to include the specified point.
    * @param  point  Point to "add" to this bounding box.
+   * @return returns true if the bbox was expanded
    */
-  void Expand(const coord_t& point) {
+  bool Expand(const coord_t& point) {
```
kevinkreiser, line 270: this was useful for the test so i added it

valhalla/midgard/encoded.h
```diff
@@ -151,8 +151,8 @@ template <class container_t> std::string encode7(const container_t& points) {
   // for each point
   for (const auto& p : points) {
     // shift the decimal point x places to the right and truncate
-    int lon = static_cast<int>(floor(static_cast<double>(p.first) * 1e6));
-    int lat = static_cast<int>(floor(static_cast<double>(p.second) * 1e6));
+    int lon = static_cast<int>(round(static_cast<double>(p.first) * 1e6));
+    int lat = static_cast<int>(round(static_cast<double>(p.second) * 1e6));
```
kevinkreiser, line 155: this and the round above are the real improvement here

danpat (dismissed):

danpaz (approved):

PR #2635, opened 2020-10-08
Fix missing turn lanes on split edge

danpaz:
# Issue

We _seem_ to have a bug in data building where turn lanes are not getting assigned to all the edges that make up an OSM way. 

The only commit here is adding a failing test case. The way `ABC` has `turn:lanes=left|||right`, and is intersected in the middle by way `DB`. The edge on the right of the intersection has turn lanes correctly assigned, but the edge on the left has 0 turn lanes.

/cc @gknisely 

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/test/gurka/test_turn_lanes.cc
+++ b/test/gurka/test_turn_lanes.cc
@@ -29,12 +29,13 @@ TEST(Standalone, TurnLanes) {
          |
          D)";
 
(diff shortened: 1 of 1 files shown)


danpaz:
This issue was fixed in https://github.com/valhalla/valhalla/pull/2679.

Issue #1148, opened 2018-03-15
Support OSM Pronunciation Tag


dnesbitt61:
https://wiki.openstreetmap.org/wiki/Key:name:pronunciation

Should add this to the name list within a tile and add a special type (language?) to the NameInfo structure within EdgeInfo. Could potentially use the additional_fields_  member to differentiate this name from the regular name.

Additionally, should add this information to TripPath for potential use in guidance preparation (Odin) and output serialization. This also includes populating the information in TripPathBuilder.

kevinkreiser:
I think we should support arbitrary key/value pairs, per way, and solve both this problem and any other non critical path data with it. Specifically I'm thinking about all the different languages and permutations of keys that have to do with names. But if we allow for any key/value pairs we could allow, at tile build time, retaining any other attributions we see fit. The idea is that we dont need to structure this data wrt what languages etc with an enum. We can simply just parse the keys on-the-fly when we want to use it. The data will be so small and only used at the narrative generation part (or even just returned with trace_attributes and not even parsed) that it wont matter in terms of performance. This means that flexibility is high but also performance remains unimpacted. It also means we dont have to make ridiculous hard choices about what tags to drop on the floor because they dont fit whatever rigid scheme we cook up.

Concretely what I'm proposing is the following:

* First change the name of this issue to `Support Storing Arbitrary Key/Value Pairs per Way/Edge`
* Add place in the config to store any desirable regexs that mjolnir can use at tile building time to grab auxiliary key/value pairs when parsing ways
* Keep a single string for all key/value pairs. The string should be of the form: `key0\0value0\0key1\0value1...keyN\0valueN\0\0`. Discard any key/value pairs that are empty or contain `\0` 
* Use a bit from `EdgeInfo::PackedItem::spare_` to say whether or not this `EdgeInfo` record has auxiliary tags
* If auxiliary tags exist for this edge, add another `NameInfo` to the end of `EdgeInfo::name_info_list_` with the correct offest into the text list
* Add a convenience method `EdgeInfo::AuxiliaryTags` which caches/returns an `unordered_map<std::string, std::string>` with all the key/value pairs broken out

[...]

PR #2608, opened 2020-09-18
Adding Tunnel Names to Intersection Object

kdiluca:
# Issue

What issue is this PR targeting? If there is no issue that addresses the problem, please open a corresponding issue and link it here.

## Tasklist

 - [ ] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [ ] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

Issue #2254, opened 2020-02-26
English translations of road names are unconditionally included in guidance instructions


1ec5:
Guidance instructions should only include the English translation of a road name if the instructions themselves are in English.

## Problem

Currently, as long as a way is tagged with an English name (`name:en`), the English name is appended to the road’s name in guidance instructions:

https://github.com/valhalla/valhalla/blob/cd5119203ff0b77f37656d68bfe38acb5362f597/src/mjolnir/pbfgraphparser.cc#L608-L610 https://github.com/valhalla/valhalla/blob/534b400a4cba4b1b4b643b69e2e3a549075dccfd/src/mjolnir/osmway.cc#L159-L165

So for example, an instruction for turning right onto [this street](https://www.openstreetmap.org/way/642075015) would label the street as “五權西路二段;Wuquan West Road Section 2” rather than “五權西路二段”. In an application that displays street names in a banner at the top of the screen, the banner inconsistently displays English names depending on whether an English name is provided.

## Impact

It’s very common for roads in OpenStreetMap to be tagged with English names even in non-English-speaking countries, especially in Asia. But typically only the major roads are translated; there isn’t universal coverage. For example:

* Egypt: [90%](https://taginfo.geofabrik.de/africa/egypt/keys/name#combinations)
* Gulf states (Bahrain, Kuwait, Oman, Qatar, Saudi Arabia, UAE): [71%](https://taginfo.geofabrik.de/asia/gcc-states/keys/name#combinations) of `name`-tagged features are also tagged with `name:en`
* South Korea: [65%](https://taginfo.geofabrik.de/asia/south-korea/keys/name#combinations)
* China (PRC, Hong Kong, Macao): [40%](https://taginfo.geofabrik.de/asia/china/keys/name#combinations)
* Taiwan: [26%](https://taginfo.geofabrik.de/asia/taiwan/keys/name#combinations)
* Greece: [20%](https://taginfo.geofabrik.de/europe/greece/keys/name#combinations)

(These statistics are about all features, not just roads, but they give a sense of the prevalence of the issue.)

## Proposed solution

The expedient solution would be to remove the code in `valhalla::mjolnir::OSMWay::GetNames` that processes `name_en_`. The downside is that, if a client requests instructions in English, the instructions will embed only foreign text and not the English translation. On the other hand, at least it won’t be so verbose; verbosity is problematic for both display and speech synthesis.

[...]

PR #2639, opened 2020-10-09
Add support for name:pronunciation tag

danpaz:
# Issue

Adds support for the [`name:pronunciation`](https://wiki.openstreetmap.org/wiki/Key:name:pronunciation) OSM tag using the TaggedName structure introduced in #2608 . According to taginfo approximately [2,770](https://taginfo.openstreetmap.org/keys/?key=name%3Apronunciation) features include this tag. 

Closes https://github.com/valhalla/valhalla/issues/1148

/cc @1ec5

## Tasklist

 - [X] Add tests

## Requirements / Relations


src/mjolnir/osmway.cc
```diff
@@ -182,6 +182,13 @@ std::vector<std::string> OSMWay::GetTaggedNames(const UniqueNames& name_offset_m
       names.emplace_back(std::to_string(static_cast<uint8_t>(TaggedName::kTunnel)) + t);
     }
   }
+  if (name_pronunciation_index_ != 0) {
+    // name:pronunciation names
+    tokens = GetTagTokens(name_offset_map.name(name_pronunciation_index_));
+    for (const auto& t : tokens) {
+      names.emplace_back(std::to_string(static_cast<uint8_t>(TaggedName::kPronunciation)) + t);
```
1ec5, line 189: Does Valhalla still use tags like `alt_name`, `official_name`, and `name:en` in guidance instructions, as claimed in #2254? If so, [other `:pronunciation` tags](https://taginfo.openstreetmap.org/search?q=pronunciation) may be relevant, though `name:pronunciation` is the most prominent of them.

1ec5 (commented):
Glad to see pronunciation overrides make it in here!

kevinkreiser (approved):

src/mjolnir/osmway.cc
kevinkreiser: `alt_name` is a conditional config parameter when building tiles now so you can turn it off. `official_name` is still added as a name regardless of configuration in addition to `name` and also `name:en`. you can check this function for more details: https://github.com/valhalla/valhalla/blob/6add48fa3e283c159ab03f1bc5ed5b4722acc7a2/src/mjolnir/osmway.cc#L99

in the future we'll be able to store language specific tagging (now that we have tagged values support) so we should be able to "undo" the `name:en` fallback to be a bit more graceful

gknisely (dismissed):

proto/tripcommon.proto
```diff
@@ -150,6 +150,7 @@ message TaggedName {
   enum Type {
     kTunnel = 1;
     kBridge = 2;
+    kPronunciation = 3;
```
dgearhart, line 153: if this is for street name pronunciation - then should it be stored with the street name it is associated with?

dgearhart (changes requested):
We should discuss - the pronunciation needs to be associated with its base meaning - street name, destination, etc

dgearhart:
This PR has been superseded with https://github.com/valhalla/valhalla/pull/3132

PR #2641, opened 2020-10-09
remove overly optimistic plan

matkoniecz:
# Issue

No issue reported, just tiny issues that I spotted while trimming part of README (and it seemed unrelated so posted as a separate PR rather than posting in already approved one)

## Tasklist

 - [ ] Add tests

## Requirements / Relations

None


matkoniecz:
Note: maybe also other places should be changed, maybe 2020 is also overly optimistic and it should be changed to 202?

kevinkreiser:
we'll get this one when we do the release, thanks!

PR #2619, opened 2020-09-28
add support to cross-compile with mingw64

nilsnolde:
fixes #2618 

Fixes a couple of cmake & compiler definitions to compile valhalla with MinGW64, tested with fedora:rawhide.

The maintainability surplus should be minimal. I'll gladly help to keep supporting this ability.

Tested:
- some TOOLS executables on Win 10 64-bit
- build and run same branch with MSVC 14.02

Todo:
- [x] clean comments
- [x] build on github actions
- [x] Update the [changelog](CHANGELOG.md)

PR #2644, opened 2020-10-12
Switch python generator to pybind11

### LINKED ISSUE #2628: Switch Python bindings to pybind11
I had serious trouble compiling boost.python for fedora's mingw64 toolchain, most is documented [on SO](https://stackoverflow.com/questions/64118672/compile-boost-python-1-73-0-for-windows-with-mingw64-on-fedora). In the end, it compiled, but then I couldn't get valhalla's bindings to work on Windows in a cross-compiled python, still complains about missing DLLs even though I pulled all relevant ones next to the python lib.

If you're ok with it I'd like to switch to the **header-only** pybind11 to generate the bindings instead of boost.python. I'd include it in `third-party` as a submodule. Already tried a minimum example in cross-compilation and it worked easily on Windows.

Of course I'll keep the API the same (for now anyways, eventually I'd like to make it more pythonic, more a v4 plan I guess).

nilsnolde:
fixes #2628 

The Python bindings can now be generated with header-only [pybind11](https://pybind11.readthedocs.io/en/stable/) instead of the binary boost.python.

I had to overload the `tyr::actor_t::<action>` functions, pybind11 is not as elegant as boost.python with overloading it seems. I basically followed [the docs](https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods).

## Tasklist

 - [ ] better overloading pattern?
 - [ ] check CMakeLists.txt for windows "solution"


nilsnolde:
damn, one branch left with kevin's GH actions commits.. I'll revert.

nilsnolde:
Would've been a huge mess, those commits were in the middle of a ton of other commits.. I just removed the `linux.yml` from the GH actions folder, that was the only affected file..

CMakeLists.txt
```diff
@@ -125,6 +125,13 @@ endif()
 find_package(Threads REQUIRED)
 find_package(ZLIB REQUIRED)
 
+# For some reason Win and Python messes with libcurl
+if(ENABLE_PYTHON_BINDINGS AND WIN32)
+  find_package(CURL REQUIRED)
+  link_libraries(${CURL_LIBRARIES})
+  include_directories(${CURL_INCLUDE_DIRS})
+endif()
+
```
nilsnolde, line 134: NO idea what I'm doing there.. But somehow that's the only way I could make it work when compiling the Python bindings for _and_ on (also one of the very few things I tried..). If that's not set MSVC (and MinGW) errors out with missing links to the `curl` library, similar to https://github.com/valhalla/valhalla/pull/2619#discussion_r498091384 when MinGW didn't find curl..

@mloskot if you have a clue how to that better feel free to comment.. feels very suboptimal.. though I realize you're likely not that interested in the bindings

src/bindings/python/python.cc
```diff
@@ -52,39 +51,41 @@ configure(const boost::optional<std::string>& config = boost::none) {
...
-      .def("TransitAvailable", &valhalla::tyr::actor_t::route, transit_available_overloads())
-      .def("Expansion", &valhalla::tyr::actor_t::route, expansion_overloads())
+PYBIND11_MODULE(valhalla_python, m) {
+  m.def("Configure", py_configure);
 
-      ;
+  py::class_<valhalla::tyr::actor_t, std::shared_ptr<valhalla::tyr::actor_t>>(m, "Actor")
+      .def(py::init<>([]() { return std::make_shared<valhalla::tyr::actor_t>(configure(), true); }))
+      .def("Route", overload_cast_<const std::string&>()(&valhalla::tyr::actor_t::route),
+           "Calculates a route.")
```
nilsnolde, line 66: following the pattern from https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods

src/tyr/actor.cc
```diff
@@ -77,6 +77,10 @@ actor_t::route(const std::string& request_str, const std::function<void()>* inte
   return bytes;
 }
 
+std::string actor_t::route(const std::string& request_str) {
+  return actor_t::route(request_str, NULL, NULL);
+}
+
```
nilsnolde, line 83: I realize this pattern is ugly and repetitive.. hints or a little code sample would be appreciated

src/valhalla_run_route.cc
```diff
@@ -581,6 +581,21 @@ int main(int argc, char* argv[]) {
...
+  boost::optional<boost::property_tree::ptree&> logging_subtree =
+      pt.get_child_optional("thor.logging");
+  if (logging_subtree) {
+    auto logging_config =
+        valhalla::midgard::ToMap<const boost::property_tree::ptree&,
+                                 std::unordered_map<std::string, std::string>>(logging_subtree.get());
+
+    valhalla::midgard::logging::Configure(logging_config);
+  }
+
```
nilsnolde, line 598: weird.. will revert

bench/meili/mapmatch.cc
```diff
@@ -111,7 +111,7 @@ static void BM_ManyCases(benchmark::State& state) {
   valhalla::tyr::actor_t actor(config, true);
   const std::string test_case(LoadFile(kBenchmarkCases[state.range(0)]));
   for (auto _ : state) {
-    benchmark::DoNotOptimize(actor.trace_route(test_case));
+    benchmark::DoNotOptimize(actor.trace_route(test_case, NULL, NULL));
```
nilsnolde, line 114: was an ambiguous call

nilsnolde:
No idea what GH Actions wants to tell me.. `CMake Error: The source directory "/__w/valhalla/valhalla/build/ " does not exist.`

I'll go over to win and fix the Azure x86 build.

bench/meili/mapmatch.cc
mloskot: I'd make it `nullptr`

CMakeLists.txt
mloskot: @nilsnolde I don't know why cURL is needed for the bindings.

I have no experience with pybind11, but is this intentional to remove uses of [FindPythonLibs.cmake](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html)?

nilsnolde: Not asking why it's needed for python (though it shouldn't be needed, maybe some distant dependency from some file..). I'm more asking why MSVC would have problems finding curl as two lines down it again tries to find it. Note, I only need the additional curl lines I added when 1) on Windows & Fedora/MinGW and 2) PYTHON_BINDINGS=ON.

> I have no experience with pybind11, but is this intentional to remove uses of FindPythonLibs.cmake?

Not sure what you mean, but python/pybind has its own CMake where somewhere there's FindPython.cmake by pybind.

nilsnolde:
Only Azure currently works with the Valhalla bindings:
- GH actions: ominous `CMake Error: The source directory "/__w/valhalla/valhalla/build/ " does not exist` error, where I can't see how my changes alter the build directory... Locally this chain of commands works!
- CircleCI: there's a whole bunch of ambiguous calls due to the new `tyr::actor_t::<action>` overloads.. I'll wait with fixing them in case there's a less invasive approach you can suggest.

CMakeLists.txt
mloskot: > I'm more asking why MSVC would have problems finding curl as two lines down it again tries to find it.

Well, it does try to find it, but only `if(ENABLE_HTTP OR ENABLE_DATA_TOOLS)`. 
If that is all false, cURL is not being searched, obviously, but something is still requiring it. 🤷 

All these lines, however, seem redundant indeed
```
# For some reason Win and Python messes with libcurl
if(ENABLE_PYTHON_BINDINGS AND WIN32)
  find_package(CURL REQUIRED)
  link_libraries(${CURL_LIBRARIES})
  include_directories(${CURL_INCLUDE_DIRS})
endif()
```

nilsnolde: Forgot to mention: I had both variables ON and it still didn't find the curl library when bindings were ON as well.

I changed it slightly to 
```
if(NOT TARGET CURL::CURL)
  add_library(CURL::CURL INTERFACE IMPORTED)
  if(ENABLE_HTTP OR ENABLE_DATA_TOOLS OR ENABLE_PYTHON_BINDINGS)
    if(NOT CURL_FOUND)
      find_package(CURL REQUIRED)
      if (WIN32 AND ENABLE_PYTHON_BINDINGS)
        link_libraries(${CURL_LIBRARIES})
      endif()
    endif()
```

Is that better you think? Guess it's the lazy version instead of fully investigating why that happens and fix the source of the problem. I've seen worse patches though;)

mloskot: it seems better w.r.t. to having single place of `find_package(CURL)` call, but the root of the linking issue is still unknown indeed. I have no idea.

valhalla/tyr/actor.h
```diff
@@ -21,33 +21,63 @@ class actor_t {
   std::string route(const std::string& request_str,
                     const std::function<void()>* interrupt = nullptr,
                     Api* api = nullptr);
+  std::string route(const std::string& request_str) {
```
kevinkreiser, line 24: there has to be a way to do this in the bindings instead of polluting the public c++ interface. i think i might checkout this branch and give it a quick hack to see if i can decypher the correct syntax sugar to make it happen :smile:

CMakeLists.txt
```diff
@@ -127,9 +127,12 @@ find_package(ZLIB REQUIRED)
 
 if(NOT TARGET CURL::CURL)
   add_library(CURL::CURL INTERFACE IMPORTED)
-  if(ENABLE_HTTP OR ENABLE_DATA_TOOLS)
+  if(ENABLE_HTTP OR ENABLE_DATA_TOOLS OR ENABLE_PYTHON_BINDINGS)
     if(NOT CURL_FOUND)
       find_package(CURL REQUIRED)
+      if (WIN32 AND ENABLE_PYTHON_BINDINGS)
+        link_libraries(${CURL_LIBRARIES})
```
kevinkreiser, line 134: im guess on windows you are getting a linker error without this? im wondering why that would be.. surely pybind11 doesnt need curl linking?

kevinkreiser: ahh i see the discussion above, can you post the link error you get when you comment out these lines. maybe somethign will pop out at someone

valhalla/tyr/actor.h
nilsnolde: Yeah, totally agreed, thx!

CMakeLists.txt
nilsnolde: jup, linker error. pybind doesn't need curl, must be some include in `python.cc`, couldn't really decipher which one though. it was the same with boost.python.

kevinkreiser:
i cant believe how simple this is. good riddance boost::python. you were way heavier than you were ever worth but thanks for the good times while they lasted :wave:

nilsnolde:
> thanks for the good times while they lasted
He gets a kick in the butt on the way out for all I care :wink: 

Thanks to your inspiration, I pulled most of the changes to `python.cc`, so the rest of the code is left alone. Still could be done with more style, but couldn't figure out how to apply a template class here, then again, didn't spend too much time on it either.

CircleCI should be ok now too, python test works as well.

Nit: I'll give some instructions for a `pre-commit` hook for `./scripts/format.sh` to the dev section of the README.. ALWAYS trips me.

nilsnolde:
And seriously, how can this change in the gh action workflow cause the failure right now.. No idea what to do about it.. Where the hell does it take the path __w/valhalla/**valhalla**/build/ from?!

```diff
diff --git a/.github/workflows/mingw-build.yml b/.github/workflows/mingw-build.yml
index f2e76dc88..692de6e56 100644
--- a/.github/workflows/mingw-build.yml
+++ b/.github/workflows/mingw-build.yml
@@ -38,10 +38,13 @@ jobs:
             -DENABLE_TOOLS=ON \
             -DENABLE_DATA_TOOLS=OFF \
             -DENABLE_SERVICES=OFF \
-            -DENABLE_PYTHON_BINDINGS=OFF \
+            -DENABLE_PYTHON_BINDINGS=ON \
             -DENABLE_CCACHE=OFF \
             -DENABLE_TESTS=OFF \
             -DENABLE_BENCHMARKS=OFF \
             -DLOGGING_LEVEL=DEBUG \
-            -DBoost_PROGRAM_OPTIONS_LIBRARY=/usr/x86_64-w64-mingw32/sys-root/mingw/lib/libboost_program_options-mt-x64.dll.a && \
+            -DBoost_PROGRAM_OPTIONS_LIBRARY=/usr/x86_64-w64-mingw32/sys-root/mingw/lib/libboost_program_options-mt-x64.dll.a \ 
+            -DPYTHON_EXECUTABLE=/usr/x86_64-w64-mingw32/bin/python3 \
+            -DPYTHON_LIBRARIES=/usr/x86_64-w64-mingw32/sys-root/mingw/lib/libpython3.9.dll.a \
+            -DPYTHON_INCLUDE_DIRS=/usr/x86_64-w64-mingw32/sys-root/mingw/include/python3.9/ && \
           mingw64-make -j$(nproc)```

File: CMakeLists.txt (hunk 7 above)
Comment (mloskot) [reply CMakeLists.txt:134]: @nilsnolde If you run CMake with Visual Studio generator and open the solution in VS IDE, you can edit properties of the bindings project, and toggle the [/showIncludes](https://docs.microsoft.com/en-us/cpp/build/reference/showincludes-list-include-files), then build just the bindings (or just compile the `python.cc`) and it will dump all the headers included from the `python.cc`. That may give some clues.

Comment (nilsnolde) [reply CMakeLists.txt:134]: I tried, but couldn't make it work.. Always worked in VS Code, I couldn't stand VS IDE from the first moment on.. I tried a `new project from existing sources` and (with a minute-old fresh install), it configured alright, but then the properties are grayed out:
image

And under Project menu item there's no Properties entry either. 

For @kevinkreiser, here's the error log, unable to link to curl:
```
[build]      Creating library C:/Users/nilsn/Documents/dev/cpp/valhalla/build/src/bindings/python/Release/python_valhalla.lib and object C:/Users/nilsn/Documents/dev/cpp/valhalla/build/src/bindings/python/Release/python_valhalla.exp
[build] valhalla.lib(curler.obj) : error LNK2001: unresolved external symbol __imp_curl_global_init [C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\python_valhalla.vcxproj]
[build] valhalla.lib(curler.obj) : error LNK2001: unresolved external symbol __imp_curl_global_cleanup [C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\python_valhalla.vcxproj]
[build] valhalla.lib(curler.obj) : error LNK2001: unresolved external symbol __imp_curl_easy_init [C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\python_valhalla.vcxproj]
[build] valhalla.lib(curler.obj) : error LNK2001: unresolved external symbol __imp_curl_easy_setopt [C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\python_valhalla.vcxproj]
[build] valhalla.lib(curler.obj) : error LNK2001: unresolved external symbol __imp_curl_easy_perform [C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\python_valhalla.vcxproj]
[build] valhalla.lib(curler.obj) : error LNK2001: unresolved external symbol __imp_curl_easy_cleanup [C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\python_valhalla.vcxproj]
[build] valhalla.lib(curler.obj) : error LNK2001: unresolved external symbol __imp_curl_easy_getinfo [C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\python_valhalla.vcxproj]
[build] C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\Release\python_valhalla.cp39-win_amd64.pyd : fatal error LNK1120: 7 unresolved externals [C:\Users\nilsn\Documents\dev\cpp\valhalla\build\src\bindings\python\python_valhalla.vcxproj]
```

Comment (mloskot) [reply CMakeLists.txt:134]: @nilsnolde That is because you opened CMake directly, so VS runs in CMake mode. I suggested to run CMake from CLI to generate `.sln` and `.vcxproj` files, what then allows you to compile single source files.
Anyway, it does not have to be VS IDE, it can be any UI you prefer, CLI included, as long as you can add `/showIncludes` to `CL.EXE` command line, what actually was the gist of my suggestion.

File: src/bindings/python/python.cc [hunk 8]
```diff
@@ -52,39 +51,70 @@ configure(const boost::optional<std::string>& config = boost::none) {
...
-BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(height_overloads, height, 1, 1);
-BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(transit_available_overloads, transit_available, 1, 1);
-BOOST_PYTHON_MEMBER_FUNCTION_OVERLOADS(expansion_overloads, expansion, 1, 1);
 } // namespace
 
-BOOST_PYTHON_MODULE(valhalla) {
+namespace py = pybind11;
+
+struct simplified_actor_t : public valhalla::tyr::actor_t {
+  simplified_actor_t(const boost::property_tree::ptree& config, bool auto_cleanup = false)
```
kevinkreiser, line 59: may as well remove the `auto_cleanup` argument and pass true to the parent class here instead of setting to true below in the bindings

CMakeLists.txt
kevinkreiser: when you are configuring are you configuring with `ENABLE_HTTP=On`? if not, ie if you have it set to `Off` (the default is `On`), then the code inside curler that makes use of curl should be disabled unless for some reason we have python bindings turning ENABLE_HTTP back on? we should probably have a look through cmake to see if there is a bug there somewhere...

nilsnolde: Pretty sure it's not exactly what you were after, anyways here's the output of compiling python.cc until it can't find the proto headers (which need to be compiled first): https://gist.github.com/nilsnolde/3d7d81d762e1dc21a9f97e70eba96848

I couldn't follow really what you were saying.. I have no idea how to use `.vcxproj` files with `cl.exe`.. The above is the output simply linking a few headers before compiling python.cc in the `valhalla/src/bindings/python/` folder

nilsnolde: Apparently it doesn't matter for the MVSC build whether ENABLE_HTTP is ON or OFF @kevinkreiser. Both configs fail in the exact same way when PYTHON_BINDINGS is ON (without the extra lines above).

kevinkreiser: very interesting we must have a long standing bug there then? ill have a look later on today!

nilsnolde: thanks @mloskot :)

.github/workflows/mingw-build.yml
```diff
@@ -38,10 +38,13 @@ jobs:
...
+            -DENABLE_PYTHON_BINDINGS=ON \
             -DENABLE_CCACHE=OFF \
             -DENABLE_TESTS=OFF \
             -DENABLE_BENCHMARKS=OFF \
             -DLOGGING_LEVEL=DEBUG \
-            -DBoost_PROGRAM_OPTIONS_LIBRARY=/usr/x86_64-w64-mingw32/sys-root/mingw/lib/libboost_program_options-mt-x64.dll.a && \
+            -DBoost_PROGRAM_OPTIONS_LIBRARY=/usr/x86_64-w64-mingw32/sys-root/mingw/lib/libboost_program_options-mt-x64.dll.a \ 
+            -DPYTHON_EXECUTABLE=/usr/x86_64-w64-mingw32/bin/python3 \
+            -DPYTHON_LIBRARIES=/usr/x86_64-w64-mingw32/sys-root/mingw/lib/libpython3.9.dll.a \
+            -DPYTHON_INCLUDE_DIRS=/usr/x86_64-w64-mingw32/sys-root/mingw/include/python3.9 && \
```
mloskot, line 49: @nilsnolde no, but close, 😄 , it's not the trailing slash, but the syntactic incorrect ` && ` before the `\`

nilsnolde: huh? haha seriously I don't see the problem @mloskot .. There's another command following the `&&`, really am I blind? not only that I don't see it, but my local setup is the exact same (in .sh form) and no syntax errors..

nilsnolde: I give up.. in this commit it "worked": https://github.com/valhalla/valhalla/blob/3b3ef50c3cef5fb12545a21cc28ef43509ff1d8d/.github/workflows/mingw-build.yml

And that's no different what I've been doing with the very last commit: https://github.com/valhalla/valhalla/blob/5b43e6c1e3d3a9cf2fee87542db8aa3c4d2a4fef/.github/workflows/mingw-build.yml

First goes through, last one doesn't! wtf..

mloskot: Sorry, I just looked at the diff part, drawing conclusions based on  incomplete picture and a minute before calling it a day. I should have shut up :)

No idea what's going wrong there but I wouldn't be surprised if it was a ln issue with parsing YAML with long complex shell command, perhaps hitting some sort of bug or limit. I've seen YAML issues on Azure Pipelines.

nilsnolde: Well, no kidding, whitespace was hidden there behind the last `\`!!! AAAARGGH! Cost me a good 2 hours of my life.. Thanks @timmccauley for spotting that!

CMakeLists.txt
kevinkreiser: is it possible that the python dll in windows links curl but somehow doesnt advertise that fact? on my version of linux libpython does not link curl and ive had a look at the python soruce and its configure.ac also doesnt seem to reference curl as a dependency. that makes me think that indeed we have some other long standing bug in our cmake files that is causing windows to require curl all the time! still didnt get a chance to look yet but i'll do so shortly. imho this little oddity is the only thing blockign this PR

.github/workflows/mingw-build.yml
mloskot: Ha, good one!

CMakeLists.txt
nilsnolde: Oh boy, now it gets weird.. Python is not dependent on curl: When I compile with `ENABLE_DATA_TOOLS=OFF` (and ENABLE_HTTP=OFF), I have no issues with curl and the python target (and all others which are enabled in that combo) build smoothly. 

It seems smth is not working right with `curler.cc` `#ifdef CURL_STATICLIB`.. If I switch that with smth I know I don't have (e.g. `#ifdef HAVE_HTTP=ON`), then all targets compile **also with `ENABLE_DATA_TOOLS=ON`**. There's even targets in `DATA_TOOLS` which use curl, e.g. `valhalla_fetch_transit` and running that doesn't seem to be an issue:
```
2020/10/15 08:20:04.872656 ←[32;1m[INFO]←[0m Fetching transit feeds
2020/10/15 08:20:07.601253 ←[33;1m[WARN]←[0m 404'd retrying http://transit.land//api/v1/feeds.geojson?per_page=false&api_key=transitland-YOUR_KEY_SUFFIX&bbox=-122.469,37.502,-121.78,38.018&active_feed_version_import_level=4
```

So smth weird seems to happen with the combo Python/ENABLE_DATA_TOOLS and libcurl. Does that ring a bell with anyone?

kevinkreiser:
Ok regarding the cURL issue... There are a couple of things at play here:

1. valhalla disables the code that requires (expects curl symbols in the linked libraries) right here: https://github.com/valhalla/valhalla/blob/15a61f761ad5b37953e917e4a32bc1017db55ca4/src/baldr/curler.cc#L9
2. we turn that preprocessor directive on right here: https://github.com/valhalla/valhalla/blob/15a61f761ad5b37953e917e4a32bc1017db55ca4/CMakeLists.txt#L128
3. notice on the next line though we exclude linking curl for windows....

so yeah this is the problem, for some reason we turn on curl with the preprocessor but disable linking it for windows. @nilsnolde do your builds work if you:

1. delete this: https://github.com/valhalla/valhalla/blob/15a61f761ad5b37953e917e4a32bc1017db55ca4/CMakeLists.txt#L129
2. delete this: https://github.com/valhalla/valhalla/blob/dc82b224b575a9993225f862638c424ec9f622ff/CMakeLists.txt#L133-L135

it seems to me we had an oversight here :smile:

CMakeLists.txt
```diff
@@ -127,9 +127,12 @@ find_package(ZLIB REQUIRED)
...
+      endif()
     endif()
     set_target_properties(CURL::CURL PROPERTIES
       #INTERFACE_LINK_LIBRARIES "${CURL_LIBRARIES}"
```
kevinkreiser, line 138: maybe this will work work:


```suggestion
  if(ENABLE_HTTP OR ENABLE_DATA_TOOLS)
    if(NOT CURL_FOUND)
      find_package(CURL REQUIRED)
    endif()
    set_target_properties(CURL::CURL PROPERTIES
            INTERFACE_INCLUDE_DIRECTORIES "${CURL_INCLUDE_DIR}"
      INTERFACE_COMPILE_DEFINITIONS CURL_STATICLIB)
    set_property(TARGET CURL::CURL APPEND PROPERTY INTERFACE_LINK_LIBRARIES "${CURL_LIBRARIES}")
  else()
    message(STATUS "Using curl from the outside")
  endif()
```

and failing that, maybe could works:

```suggestion
  if(ENABLE_HTTP OR ENABLE_DATA_TOOLS)
    if(NOT CURL_FOUND)
      find_package(CURL REQUIRED)
    endif()
    set_target_properties(CURL::CURL PROPERTIES
            INTERFACE_INCLUDE_DIRECTORIES "${CURL_INCLUDE_DIR}"
      INTERFACE_COMPILE_DEFINITIONS CURL_STATICLIB)
    if(NOT MSVC)
      set_property(TARGET CURL::CURL APPEND PROPERTY INTERFACE_LINK_LIBRARIES "${CURL_LIBRARIES}")
      else()
        link_libraries(${CURL_LIBRARIES})
      endif()
    endif()
  else()
    message(STATUS "Using curl from the outside")
  endif()
```

nilsnolde: Wonderful @kevinkreiser ! and easy.. should've tried to understand the logic here better.. 

Note, the first example errors on
```
[cmake] CMake Error at CMakeLists.txt:137 (set_property):
[cmake]   Property INTERFACE_LINK_LIBRARIES may not contain link-type keyword
[cmake]   "optimized".  The INTERFACE_LINK_LIBRARIES property may contain
[cmake]   configuration-sensitive generator-expressions which may be used to specify
[cmake]   per-configuration rules.
```
Which does sound like a libcurl (or vcpkg?) issue.. didn't find much but same issues for other libs suggest it's the lib's CMake file being the issue. 

Anyways, second option works, not as nice option1, but still a lot better than mine! Fine for you too @kevinkreiser ?

nilsnolde:
Should be good to merge now, thanks @kevinkreiser & @mloskot !

kevinkreiser (approved):

PR #2645, opened 2020-10-13
more locate output

kevinkreiser:
we've been adding some stuff to the tiles specifically real time speeds and incidents this is the beginning of adding that to locate (which is useful for debugging stuff). ~we might take this opportunity to only show the historical traffic speed with the time that is associated with the locate request... haven't decided yet.~

ive also added linear references to the locate output and while i was at it i found and fixed a bunch of issues with the existing linear reference support in the route and trace_route apis


src/baldr/openlr.cc
```diff
@@ -1,97 +0,0 @@
...
-#include "proto/tripcommon.pb.h"
-
-namespace valhalla {
-namespace midgard {
-
-using FormOfWay = OpenLR::LocationReferencePoint::FormOfWay;
-
-FormOfWay road_class_to_fow(const TripLeg::Node& node) {
-  const bool is_missing_rc = !node.has_edge() || !node.edge().has_road_class();
-  if (is_missing_rc || !node.edge().has_use()) {
```
kevinkreiser, line 20: this shouldnt happen. we should never make an openlr without an edge to make it from and every edge will always have a road class.

src/baldr/openlr.cc
```diff
@@ -1,97 +0,0 @@
...
-  }
-  const TripLeg::Use use = node.edge().use();
-  const RoadClass rc = node.edge().road_class();
-  const bool one_way =
-      node.edge().has_traversability() && node.edge().traversability() != TripLeg::kBoth;
-  if (node.edge().roundabout()) {
-    return FormOfWay::ROUNDABOUT;
-  } else if (use == TripLeg::kRampUse || use == TripLeg::kTurnChannelUse) {
-    return FormOfWay::SLIPROAD;
-  } else if (rc == RoadClass::kMotorway && one_way) {
```
kevinkreiser, line 31: while this is probably usually true, the FOW doesnt say that motorway has to be one way

src/baldr/openlr.cc
```diff
@@ -1,97 +0,0 @@
...
-    return FormOfWay::MOTORWAY;
-  } else if (rc <= RoadClass::kTertiary && one_way) {
```
kevinkreiser, line 33: i dont think the road class plays a role in whether a road is single or dual+ carriageway. the best way we can differentiate this is to say can you drive in both directions or can you drive in at least one direction

src/baldr/openlr.cc
```diff
@@ -1,97 +0,0 @@
...
-  openlrs.reserve(leg.node_size());
-  for (const TripLeg::Node& node : leg.node()) {
-    // OpenLR for an edge is composed of two points with the following semantics:
-    // - FRC = relative importance of the road, so use the rank order of the FormOfWay enum.
-    // - Bearing of last point in the segment must point inward (bearing rotated by 180deg),
-    //   in-order to follow OpenLR conventions.
-    const FormOfWay fow = road_class_to_fow(node);
-    const unsigned char frc = static_cast<unsigned char>(fow);
-    unsigned char lfrcnp = frc;
-    const PointLL& start = points.at(node.edge().begin_shape_index());
```
kevinkreiser, line 54: here we were making use of the edge without checking whether or not we have one. the last node in the leg will not have an edge as its the end of the route. so this was undefined behavior but also, semantically wrong. we should only have one openlr between every pair of nodes in the route. therefor the number of linear references in the output should be `leg.node_size() - 1`

src/baldr/openlr.cc
```diff
@@ -1,97 +0,0 @@
...
-    const PointLL& end = points.at(node.edge().end_shape_index());
-    float bearing_deg = start.Heading(end);
```
kevinkreiser, line 56: the bearing should be computed along the shape of the path, not between LRPs. if the edge is really curvy this will make the bearing useless

src/baldr/openlr.cc
```diff
@@ -1,97 +0,0 @@
...
-  const std::vector<PointLL>& points = decode<std::vector<PointLL>>(leg.shape());
-  std::vector<std::string> openlrs;
-  openlrs.reserve(leg.node_size());
-  for (const TripLeg::Node& node : leg.node()) {
-    // OpenLR for an edge is composed of two points with the following semantics:
-    // - FRC = relative importance of the road, so use the rank order of the FormOfWay enum.
-    // - Bearing of last point in the segment must point inward (bearing rotated by 180deg),
-    //   in-order to follow OpenLR conventions.
-    const FormOfWay fow = road_class_to_fow(node);
-    const unsigned char frc = static_cast<unsigned char>(fow);
```
kevinkreiser, line 52: frc should be equivalent to `edge.road_class`. casting the fow probably gives seemingly random values for frc

src/tyr/locate_serializer.cc
```diff
@@ -1,11 +1,57 @@
...
 #include "tyr/serializers.h"
 #include <cstdint>
 
 using namespace valhalla;
+using namespace valhalla::midgard;
 using namespace valhalla::baldr;
 
 namespace {
+
+OpenLR::LocationReferencePoint::FormOfWay get_fow(const baldr::DirectedEdge* de) {
```
kevinkreiser, line 12: i moved the patched versions of these functions over to the serializer where they are used. ive also converted the tests to check them after serialization to get the full round trip whilest also avoiding polluting the public interface with what are implementation details.

src/tyr/serializers.cc
```diff
@@ -34,6 +34,55 @@ namespace {
 midgard::PointLL to_ll(const LatLng& ll) {
   return midgard::PointLL{ll.lng(), ll.lat()};
 }
+using FormOfWay = valhalla::midgard::OpenLR::LocationReferencePoint::FormOfWay;
+
+FormOfWay road_class_to_fow(const valhalla::TripLeg::Edge& edge) {
```
kevinkreiser, line 39: i moved the patched versions of these functions over to the serializer where they are used. ive also converted the tests to check them after serialization to get the full round trip whilest also avoiding polluting the public interface with what are implementation details.

valhalla/baldr/openlr.h
```diff
@@ -323,10 +323,13 @@ struct OpenLr {
...
-         uint8_t negative_offset_bucket)
+         uint8_t negative_offset_bucket,
+         bool point_along_line = false,
+         const Orientation& orientation = Orientation::NoOrientation,
+         const SideOfTheRoad& side_of_the_road = SideOfTheRoad::DirectlyOnRoadOrNotApplicable)
       : lrps(lrps), poff(positive_offset_bucket), noff(negative_offset_bucket),
-        isPointAlongLine(false), orientation(Orientation::NoOrientation),
-        sideOfTheRoad(SideOfTheRoad::DirectlyOnRoadOrNotApplicable) {
+        isPointAlongLine(point_along_line), orientation(orientation),
+        sideOfTheRoad(side_of_the_road) {
```
kevinkreiser, line 332: for locate, we need to do a point along line type of openlr. for that reason i needed to open up these extra options in the constructor

src/tyr/locate_serializer.cc
PureW, line 12: Curious what you think about taking a reference instead of a pointer as argument? It doesn't seem like it should be possible to send a nullpointer so why not rule out that possibility entirely?

src/tyr/locate_serializer.cc
```diff
@@ -1,11 +1,57 @@
...
+  else if ((de->forwardaccess() & kVehicularAccess) && (de->reverseaccess() & kVehicularAccess))
+    return OpenLR::LocationReferencePoint::MULTIPLE_CARRIAGEWAY;
+  else if ((de->forwardaccess() & kVehicularAccess) || (de->reverseaccess() & kVehicularAccess))
+    return OpenLR::LocationReferencePoint::SINGLE_CARRIAGEWAY;
+
+  return OpenLR::LocationReferencePoint::OTHER;
+}
+
+std::string
+linear_reference(const baldr::DirectedEdge* de, float percent_along, const EdgeInfo& edgeinfo) {
```
PureW, line 29: Same comment here.

src/tyr/locate_serializer.cc
```diff
@@ -16,6 +62,50 @@ json::ArrayPtr serialize_edges(const PathLocation& location, GraphReader& reader
...
+        auto live_speed = json::map({});
+        const volatile auto& traffic = tile->trafficspeed(directed_edge);
+        if (traffic.valid()) {
+          live_speed->emplace("overall_speed", static_cast<uint64_t>(traffic.get_overall_speed()));
+          auto speed = static_cast<uint64_t>(traffic.get_speed(0));
+          if (speed == UNKNOWN_TRAFFIC_SPEED_KPH)
+            live_speed->emplace("speed_0", nullptr);
+          else
+            live_speed->emplace("speed_0", speed);
+          auto congestion = (traffic.congestion2 - 1.0) / 62.0;
```
PureW, line 74: Hmm I don't like that there is a transformation here that depends on implementation details from traffictiles.h. Could it be move to that file?

PureW: Also, isn't it wrong with `traffic.congestion2`?

src/tyr/locate_serializer.cc
kevinkreiser: the are essentially the same thing. the only reason i didnt do it is because its not a reference when we get it from graphreader. to turn it into a reference we have to derefrence it (ie an extra step with no benefit). basically pretty much all the other code just uses const pointers for the native data and so this fits the pattern. now that im t hinking of it though, maybe it would be cool to make this a member function of directededge? though that kind of scatters the code around a bit..

test/mapmatch.cc
```diff
@@ -1426,16 +1426,17 @@ TEST(Mapmatch, openlr_parameter_true_osrm_api) {
   const auto& matches = response.get_child("matchings");
   EXPECT_EQ(matches.size(), 1);
   const std::vector<std::string>& expected = {
-      "CwOduyULYhtpAAAV//0bGQ==", "CwOdxCULYBtqAAAM//4bGg==", "CwOdySULXxtrAAAf//EbGw==",
-      "CwOd1yULVxtrAQBV/9AbGw==", "CwOduyULYj/gAAACAAA/EA==",
+      "CwOduyULYiKJAAAV//0iGw==",
+      "CwOdxCULYCKJAAAM//4iGw==",
+      "CwOdySULXyKJAAAf//EiGw==",
+      "CwOd1yULVyKLAABV/9AiGw==",
```
PureW, line 1432: Why the changed testdata?

PureW: Oh I see. You fixed the openlr serialization.

kevinkreiser: look at the diff for the file i removed, i explained that there were many bugs

PureW (approved):

PR #2646, opened 2020-10-13
Fixing some includes 

jfamiglietti:
# Issue

When using Valhalla as a library in an iOS or Android app, these includes don't resolve correctly.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


mandeepsandhu:
Can you fix the linting errors causing the CI failures?

jfamiglietti:
Updated. It took a while for one of them to complete, but it seems ok now.

src/mjolnir/statistics.h
```diff
@@ -10,8 +10,8 @@
 
 #include <sqlite3.h>
 
-#include "baldr/graphconstants.h"
 #include "midgard/aabb2.h"
+#include "valhalla/baldr/graphconstants.h"
```
kevinkreiser, line 14: this one doesnt make sense to me. are you sure this diff is required?

kevinkreiser (changes requested):
i think you can revert the change to statistics.h, or if that is required for you can you possibly show some compiler output so we can get a clue as to why its needed?

src/mjolnir/statistics.h
jfamiglietti: Yeah, you're right. I just tested it and it's fine without the change. It doesn't end up in the include directory in the installed location anyway, so I wonder how I tripped over it. I pushed a change.

kevinkreiser:
@jfamiglietti you'll want to actually revert that change, the formatter cares abotu what line that include lands on (it does some alphabetical sorting)

jfamiglietti:
Yup, I noticed that. It's effectively reverted. Sorry for the noise.

kevinkreiser (approved):

PR #2651, opened 2020-10-15
use OSRM-like turn durations in auto costing

kevinkreiser:
@SvetlanaBayarovich i couldnt commit directly to your fork without having to do it from the github webui, so i made a branch in the valhalla org with all your changes and applied some changes on top.

(supercedes) fixes #2643

We noticed in your original work that the performance hit was something like 5%, so what I added here were a few tweaks to pretty much remove that overhead. Other than that your original port is maintained.

The basic changes I added were: 
 * remove a few branches
 * remove a few divisions
 * inline a few functions
 * swapp out `std::exp` with an sse optimzed approximation

Before I swapped out `std::exp`, i wrote a small program to verify that all values we send through the approximation function would have very small differences to `std::exp`. This was easy since we only compute this function for the range -179 to 180 inclusive. The maximum delta in this computation was `0.06` seconds in the final turn duration. The good news is that using the `fasterexp` function is about 2.5x faster than `std::exp` and the library where its implemented is header only. Here's the test program I wrote to both to check the speed and accuracy of the replacement function:

```c++
#include "fastexp.h"
#include <cmath>
#include <iostream>

constexpr double kTurnPenalty = 7.5;
constexpr double kUTurnPenalty = 20;
constexpr double kTurnBias = 1.075;
constexpr double kTurnBiasInv = 1.0 / kTurnBias;
constexpr double kTrafficLightPenalty = 2;

int main(){
  double diff = 0;

for(int i= 0; i < 1000000; ++i){
  for(int osrm_angle = -179; osrm_angle < 181; osrm_angle++) {
    double turn_bias =  kTurnBias;
    double turn_bias_inv = kTurnBiasInv;

    if (osrm_angle >= 0) {
      double pen0 = kTurnPenalty / (1 + std::exp(-((13 * turn_bias_inv) * osrm_angle / 180 - 6.5 * turn_bias)));
      double pen1 = kTurnPenalty / (1 + fasterexp(-((13 * turn_bias_inv) * osrm_angle / 180 - 6.5 * turn_bias)));
      diff = std::max(diff, std::abs(pen1 - pen0));
    }
    else {
      double pen0 = kTurnPenalty / (1 + std::exp(-((13 * turn_bias) * -osrm_angle / 180 - 6.5 * turn_bias_inv)));
      double pen1 = kTurnPenalty / (1 + fasterexp(-((13 * turn_bias) * -osrm_angle / 180 - 6.5 * turn_bias_inv)));
      diff = std::max(diff, std::abs(pen1-pen0));
    }
  }
}

  std::cout << diff << std::endl;
  return diff;
}
```

After all of that I ran my 14k route benchmark program. With master I was seeing:
```
real	0m41.664s
real	0m40.827s
real	0m40.351s
```

And with this branch I was seeing:
```
real	0m40.844s
real	0m40.843s
real	0m41.809s
```

Which to me seems close enough to the same to not cause room for concern.

UPDATE:

after @danpat mentioned to me that we probably dont need to use `std::exp` exactly to get the desired effect it dawned on me that, because of my test above it makes sense to just build a lookup table statically once and just index into it for the desired value of the function (since its only a finite range for which we compute this). so i've updated to do that. that should get the runtime cost of this down to basically nothing as its really just, which lookup table should we use based on your driving side.


.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "third_party/fastcov"]
 	path = third_party/fastcov
 	url = https://github.com/RPGillespie6/fastcov.git
+[submodule "third_party/fastapprox"]
```
kevinkreiser, line 34: for  (private) header only `fasterexp` implementation

src/baldr/nodeinfo.cc
```diff
@@ -218,15 +218,6 @@ void NodeInfo::set_stop_index(const uint32_t stop_index) {
...
 }
 
-// Get the heading of the local edge given its local index. Supports
-// up to 8 local edges. Headings are expanded from 8 bits.
-uint32_t NodeInfo::heading(const uint32_t localidx) const {
-  // Make sure everything is 64 bit!
-  uint64_t shift = localidx * 8; // 8 bits per index
-  return static_cast<uint32_t>(std::round(
-      ((headings_ & (static_cast<uint64_t>(255) << shift)) >> shift) * kHeadingExpandFactor));
-}
```
kevinkreiser, line 228: inlined this and moved to header file

test/mapmatch.cc
```diff
@@ -1131,8 +1131,8 @@ TEST(Mapmatch, test_discontinuity_duration_trimming) {
   std::vector<int> test_ans_num_routes{2, 2, 2};
   std::vector<std::vector<int>> test_ans_num_legs{{1, 2}, {1, 2}, {1, 2}};
   std::vector<std::vector<float>> test_ans_leg_duration{{26.459, 0.97, 0.454},
-                                                        {48.6, 0.64, 0.961},
-                                                        {220.778, 2.431, 1.899}};
+                                                        {48.159, 0.64, 0.961},
+                                                        {93.388, 2.431, 1.899}};
```
SvetlanaBayarovich, line 1135: By the way, I found out why the difference is so large here. The route has a number of turns. "Utrecht tiles" data does not have traffic information so duration was multiplied by density factor, as a result there were durations like 15, 20 seconds, sometimes even bigger. With several turns it summed up to >100 seconds difference

SvetlanaBayarovich (approved):

SvetlanaBayarovich (dismissed):

SvetlanaBayarovich (dismissed):

SvetlanaBayarovich (approved):

PR #2653, opened 2020-10-16
Enhanced turn lane information

mandeepsandhu:
The turn lane object (intersection.lanes) within a step is enhanced with
new attributes:
* "active": Set to true for lanes which are most appropriate to be in,
considering the current and next maneuver. false otherwise.
* "active_indication": An optional attribute containing the active lane's
arrow indication in the direction of the maneuver. Eg: if active lane has
"indications: [ left, straight ]" and we're turning left from this lane,
then "active_indication" will be "left".

The existing attribute "valid", will be set to true for lanes which can
be taken for completing the maneuver, but might not be the best lane to
be in, considering the next maneuver.

### DIFF
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -131,8 +131,14 @@ message StreetName {
 }
 
 message TurnLane {
+  enum State {
+    kInvalid = 0;
+    kValid = 1;
+    kActive = 2;
+  }
   optional uint32 directions_mask = 1;
-  optional bool is_active = 2;
+  optional State state = 2 [default = kInvalid];
+  optional uint32 active_direction = 3;
 }
 
 enum RoadClass {
--- a/src/odin/enhancedtrippath.cc
+++ b/src/odin/enhancedtrippath.cc
@@ -477,7 +477,7 @@ float EnhancedTripLeg_Edge::GetLength(const Options::Units& units) {
 
 bool EnhancedTripLeg_Edge::HasActiveTurnLane() const {
   for (const auto& turn_lane : turn_lanes()) {
-    if (turn_lane.is_active()) {
+    if (turn_lane.state() == TurnLane::kActive) {
       return true;
     }
   }
@@ -508,20 +508,28 @@ bool EnhancedTripLeg_Edge::HasTurnLane(uint16_t turn_lane_direction) const {
 uint16_t EnhancedTripLeg_Edge::ActivateTurnLanesFromLeft(uint16_t turn_lane_direction,
                                                          uint16_t activated_max) {
   uint16_t activated_count = 0;
-  // Make sure turn lane has a direction
-  if (!HasNonDirectionalTurnLane()) {
-    for (auto& turn_lane : *(mutable_turn_lanes())) {
-      // Stop processing the lanes if the activated maximum has been reached
-      if (activated_count >= activated_max) {
-        break;
-      }
 
-      // If the turn lane is in the specified direction then activate the lane
-      // and increment the activated count
-      if (turn_lane.directions_mask() & turn_lane_direction) {
-        turn_lane.set_is_active(true);
+  // Make sure turn lane has a direction
+  // TODO: Consider activating lanes even if some lanes are non-directional
+  // as long as there are directional lanes in the direction of the maneuver
+  if (HasNonDirectionalTurnLane()) {
+    return activated_count;
+  }
+
+  for (auto& turn_lane : *(mutable_turn_lanes())) {
+    // Process lanes matching the turn_lane_direction
+    if (turn_lane.directions_mask() & turn_lane_direction) {
+      // TODO: Use lane connectivity to skip impossible lanes
+      // activate upto activated_max lanes
+      if (activated_count < activated_max) {
+        turn_lane.set_state(TurnLane::kActive);
         ++activated_count;
+      } else {
+        // Mark non-active lane in the direction of maneuver as valid
+        turn_lane.set_state(TurnLane::kValid);
       }
+      // Set the active direction for active & valid lanes
+      turn_lane.set_active_direction(turn_lane_direction);
     }
   }
   return activated_count;
@@ -530,22 +538,28 @@ uint16_t EnhancedTripLeg_Edge::ActivateTurnLanesFromLeft(uint16_t turn_lane_dire
 uint16_t EnhancedTripLeg_Edge::ActivateTurnLanesFromRight(uint16_t turn_lane_direction,
                                                           uint16_t activated_max) {
   uint16_t activated_count = 0;
-  // Make sure turn lane has a direction
-  if (!HasNonDirectionalTurnLane()) {
-    for (auto turn_lane_iter = mutable_turn_lanes()->rbegin();
-         turn_lane_iter != mutable_turn_lanes()->rend(); ++turn_lane_iter) {
-      //    for (auto& turn_lane : *(mutable_turn_lanes())) {
-      // Stop processing the lanes if the activated maximum has been reached
-      if (activated_count >= activated_max) {
-        break;
-      }
 
-      // If the turn lane is in the specified direction then activate the lane
-      // and increment the activated count
-      if (turn_lane_iter->directions_mask() & turn_lane_direction) {
-        turn_lane_iter->set_is_active(true);
+  // Make sure turn lane has a direction
+  // TODO: Consider activating lanes even if some lanes are non-directional
+  // as long as there are directional lanes in the direction of the maneuver
+  if (HasNonDirectionalTurnLane()) {
+    return activated_count;
+  }
+
+  for (auto turn_lane_iter = mutable_turn_lanes()->rbegin();
+       turn_lane_iter != mutable_turn_lanes()->rend(); ++turn_lane_iter) {
+    if (turn_lane_iter->directions_mask() & turn_lane_direction) {
+      // TODO: Use lane connectivity to skip impossible lanes
+      // activate upto activated_max lanes
+      if (activated_count < activated_max) {
+        turn_lane_iter->set_state(TurnLane::kActive);
         ++activated_count;
+      } else {
+        // Mark non-active in the same direction as the maneuver, as valid
+        turn_lane_iter->set_state(TurnLane::kValid);
       }
+      // Set the active direction for active & valid lanes
+      turn_lane_iter->set_active_direction(turn_lane_direction);
     }
   }
   return activated_count;
@@ -915,8 +929,16 @@ std::string EnhancedTripLeg_Edge::TurnLanesToString() const {
     }
 
     // Output if marked as active
-    if (turn_lane.is_active()) {
+    if (turn_lane.state() == TurnLane::kActive) {
       str += " ACTIVE";
+    } else if (turn_lane.state() == TurnLane::kValid) {
+      str += " VALID";
+    }
+
+    // Append the active direction if the lane is active or valid
+    if (turn_lane.state() != TurnLane::kInvalid) {
+      str += ", ACTIVE_DIR ";
+      str += kTurnLaneNames.at(turn_lane.active_direction());
     }
   }
   str += " ]";
--- a/src/odin/enhancedtrippath.cc
+++ b/src/odin/enhancedtrippath.cc
@@ -838,6 +838,15 @@ std::string EnhancedTripLeg_Edge::TurnLanesToString() const {
     }
 
     uint16_t mask = turn_lane.directions_mask();
+    auto indication_to_str = [&turn_lane](uint16_t ind) -> std::string {
+      if (turn_lane.state() == TurnLane::kInvalid || turn_lane.active_direction() != ind) {
+        return kTurnLaneNames.at(ind);
+      }
+      // Surround valid/active lanes with a '*'
+      std::stringstream ss;
+      ss << "*" << kTurnLaneNames.at(ind) << "*";
+      return ss.str();
+    };
 
     // Process the turn lanes - from left to right
     // empty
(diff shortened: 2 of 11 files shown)


proto/tripcommon.proto
```diff
@@ -133,6 +133,8 @@ message StreetName {
 message TurnLane {
   optional uint32 directions_mask = 1;
   optional bool is_active = 2;
+  optional bool is_possible = 3;
```
danpaz, line 136: What do you think about refactoring `is_active` and `is_possible` into an enum? That way it's clear that there's a hierarchy of states Active->Possible->Invalid.

src/odin/enhancedtrippath.cc
```diff
@@ -508,20 +508,27 @@ bool EnhancedTripLeg_Edge::HasTurnLane(uint16_t turn_lane_direction) const {
...
+      // activate upto activated_max lanes
+      if (activated_count < activated_max) {
         turn_lane.set_is_active(true);
         ++activated_count;
       }
+      // Mark non-active lanes as possible
+      turn_lane.set_is_possible(true);
+      // Set the active direction mask in the direction of the maneuver
+      // TODO: Highlighting lane markers for non-active but possible lanes
+      // seems like a questionable thing to do.
```
danpaz, line 530: Why questionable? I think you'd have this scenario if you're going straight through an intersection where there's multiple through lanes but one of the through lanes has an option to turn.

src/odin/enhancedtrippath.cc
```diff
@@ -918,6 +930,9 @@ std::string EnhancedTripLeg_Edge::TurnLanesToString() const {
     if (turn_lane.is_active()) {
       str += " ACTIVE";
     }
+    if (turn_lane.is_possible()) {
```
danpaz, line 933: Can this be `else if` so that we don't mark all active lanes as possible in RAD output? Since active implies possible as well.

src/odin/enhancedtrippath.cc
mandeepsandhu: Because I feel the UI will be a lot more distracting than if the lane direction markers were not highlighted (which is what this attribute is for). Imagine you're turning left, there are 3 lanes going in that direction and you have to be in the middle lane. Having the other lane's direction markers lit up is going to be visually confusing to the driver, and the only way to discern the preferred lane is to rely on the "background" being appropriately highlighted. We should consider making this an opt-in feature, maybe? (we don't have to do it in this PR, but something to consider at a later stage).

(I might be a bit biased here since my mind is trained to expect lanes in a certain way by "other" maps/nav applications 😉).

src/odin/enhancedtrippath.cc
mandeepsandhu: I had originally kept these 2 attributes mutually exclusive when setting the values in the protobuf. But yeah, you make a good point that RAD will get affected by this change (I haven't completed RAD testing yet, so didn't realize this will be an issue). Technically it will make the string representation of lanes not match the corresponding values in the protobuf, but I guess thats better than changing many tests & will make RAD easier.

src/tyr/route_serializer_osrm.cc
```diff
@@ -325,7 +324,7 @@ void route_geometry(json::MapPtr& route,
                     const valhalla::Options& options) {
   std::vector<PointLL> shape;
   if (options.has_generalize() && options.generalize() == 0.0f) {
-    shape = simplified_shape(directions, options);
```
mandeepsandhu, line 328: clang tidy complained about unused param.

proto/tripcommon.proto
mandeepsandhu: I'm not sure how an enum conveys a hierarchy. It would club them together under a single type to convey they are related. If you feel that makes it more readable, we can change it to an enum. What do you think?

src/odin/enhancedtrippath.cc
```diff
@@ -477,7 +477,7 @@ float EnhancedTripLeg_Edge::GetLength(const Options::Units& units) {
 
 bool EnhancedTripLeg_Edge::HasActiveTurnLane() const {
   for (const auto& turn_lane : turn_lanes()) {
-    if (turn_lane.is_active()) {
+    if (turn_lane.state() == ::valhalla::TurnLane::kActive) {
```
dgearhart, line 480: you could do the shorter `TurnLane::kActive`

src/tyr/route_serializer_osrm.cc
```diff
@@ -379,6 +378,30 @@ struct IntersectionEdges {
   }
 };
 
+std::string turn_lane_direction(uint16_t turn_lane) {
+  switch (turn_lane) {
+    case kTurnLaneReverse:
+      return "uturn";
```
dgearhart, line 384: should we expand to have `left uturn` and `right uturn` so user can provide proper icon?

src/tyr/route_serializer_osrm.cc
```diff
@@ -560,8 +583,15 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
       auto lanes = json::array({});
       for (const auto& turn_lane : prev_edge->turn_lanes()) {
         auto lane = json::map({});
-        // Process 'valid' flag
-        lane->emplace("valid", turn_lane.is_active());
+        // Process 'valid' & 'active' flags
+        bool is_valid = turn_lane.state() == ::valhalla::TurnLane::kValid;
```
dgearhart, line 587: the logic is reversed here - it can be valid if it is active OR the state is valid
bool is_active = (turn_lane.state() == ::valhalla::TurnLane::kActive);
bool is_valid = is_active || (turn_lane.state() == ::valhalla::TurnLane::kValid);

src/odin/enhancedtrippath.cc
```diff
@@ -915,8 +929,10 @@ std::string EnhancedTripLeg_Edge::TurnLanesToString() const {
     }
 
     // Output if marked as active
-    if (turn_lane.is_active()) {
+    if (turn_lane.state() == ::valhalla::TurnLane::kActive) {
       str += " ACTIVE";
```
dgearhart, line 933: should we add in the active direction string? so it can be part of the validation for both active and valid

src/tyr/route_serializer_osrm.cc
mandeepsandhu: I'm not sure I follow. Both implementations will be setting the 2 values exactly the same way. I put the valid check first because a lane cannot be active if its not also valid. Is there some other combination of active/valid that I'm missing currently?

src/odin/enhancedtrippath.cc
mandeepsandhu: yes! I forgot that we're already within the valhalla namespace here! thx.

src/odin/enhancedtrippath.cc
mandeepsandhu: Yes, thats a good idea. And that reminds me that I need to update tests to check for active_dirrection attribute! (sorry totally forgot about it)

src/tyr/route_serializer_osrm.cc
dgearhart: @mandeepsandhu there is only one state but multiple booleans to return - an `active` is also `valid` so you have to check `is_active` in that case

src/tyr/route_serializer_osrm.cc
mandeepsandhu: Won't that break parsing by existing clients that are expecting a `uturn`? The strings here need to be a sub-set of strings set in the [`intersection` function](https://github.com/valhalla/valhalla/blob/master/src/tyr/route_serializer_osrm.cc#L572-L608), which does not have a left/right uturn. 

I'm half a mind about changing the `intersections(...)` to use this new function so that we don't have to manage strings in 2 places. What do you think?

src/tyr/route_serializer_osrm.cc
mandeepsandhu: You are totally right. I don't know why I confused myself thinking this was like a "mask" where both attributes could be true. I'll fix this and add tests for serialized response to verify the behavior.

mandeepsandhu:
Sample RAD diffs:

```diff
<    3: TURN_LANES: [ left | left;through ACTIVE | right ] ACTIVE_TURN_LANES
---
>    3: TURN_LANES: [ left VALID, ACTIVE_DIR left | left;through ACTIVE, ACTIVE_DIR left | right ] ACTIVE_TURN_LANES
```

```diff
<    2: TURN_LANES: [ slight_left ACTIVE | slight_left ACTIVE | through | through | through ] ACTIVE_TURN_LANES
---
>    2: TURN_LANES: [ slight_left ACTIVE, ACTIVE_DIR slight_left | slight_left ACTIVE, ACTIVE_DIR slight_left | through | through | through ] ACTIVE_TURN_LANES
```

```diff
9c9
<    2: TURN_LANES: [ left | left ACTIVE | through | through | right ] ACTIVE_TURN_LANES
---
>    2: TURN_LANES: [ left VALID, ACTIVE_DIR left | left ACTIVE, ACTIVE_DIR left | through | through | right ] ACTIVE_TURN_LANES
15c15
<    3: TURN_LANES: [ left | through | through;right ACTIVE ] ACTIVE_TURN_LANES
---
>    3: TURN_LANES: [ left | through | through;right ACTIVE, ACTIVE_DIR right ] ACTIVE_TURN_LANES
```

danpaz:
@mandeepsandhu You should also try running RAD with the `--verbose-lanes` option to valhalla_run_route, this will print the turn_lanes at every edge rather than just at the maneuver point.

mandeepsandhu:
> try running RAD with the `--verbose-lanes` option to valhalla_run_route, this will print the turn_lanes at every edge rather than just at the maneuver point.

Thanks for pointing this out! I re-ran `turn_lane_routes.txt` with this option enabled. The results looked similar, just that there were more diffs of the same kind (as shown above).

src/tyr/route_serializer_osrm.cc
```diff
@@ -560,8 +583,16 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
...
-        // Process 'valid' flag
-        lane->emplace("valid", turn_lane.is_active());
+        // Process 'valid' & 'active' flags
+        bool is_active = turn_lane.state() == TurnLane::kActive;
+        // an active lane is also valid
+        bool is_valid = is_active || turn_lane.state() == TurnLane::kValid;
+        lane->emplace("active", is_active);
+        lane->emplace("valid", is_valid);
+        // Add active_indication for a valid lane
+        if (turn_lane.state() != ::valhalla::TurnLane::kInvalid) {
```
danpaz, line 593: `s/::valhalla::TurnLane/TurnLane`

danpaz (approved):
One minor comment. LGTM.

src/tyr/route_serializer_osrm.cc
mandeepsandhu: oops! thx for catching that.

test/turnlanes.cc
```diff
@@ -127,52 +127,52 @@ TEST(Turnlanes, validate_turn_lanes) {
...
                   expected_routes_size, expected_legs_size, expected_maneuvers_size, maneuver_index,
-                  "[ slight_left | slight_left | slight_right ACTIVE | right ]");
+                  "[ slight_left | slight_left | *slight_right* ACTIVE | right ]");
 
   // Test left most left u-turn active
   test_turn_lanes({VALHALLA_SOURCE_DIR
                    "test/pinpoints/turn_lanes/left_most_left_uturn_active_pinpoint.pbf"},
                   expected_routes_size, expected_legs_size, expected_maneuvers_size, maneuver_index,
-                  "[ left ACTIVE | left | left | through | through;right ]");
+                  "[ *left* ACTIVE | *left* VALID | *left* VALID | through | through;right ]");
```
dgearhart, line 152: @mandeepsandhu with a left u-tun only one lane can be active/valid - the other left turn lanes can not be valid

dgearhart (changes requested):
I commented on the left most left for a left u-turn - only one lane can be active - the others can not be active

test/turnlanes.cc
dgearhart: same for a right uturn

src/tyr/route_serializer_osrm.cc
```diff
@@ -560,8 +583,16 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
...
-        lane->emplace("valid", turn_lane.is_active());
+        // Process 'valid' & 'active' flags
+        bool is_active = turn_lane.state() == TurnLane::kActive;
+        // an active lane is also valid
+        bool is_valid = is_active || turn_lane.state() == TurnLane::kValid;
+        lane->emplace("active", is_active);
+        lane->emplace("valid", is_valid);
+        // Add active_indication for a valid lane
+        if (turn_lane.state() != TurnLane::kInvalid) {
+          lane->emplace("active_indication", turn_lane_direction(turn_lane.active_direction()));
```
dgearhart, line 594: Typo - should be `valid_indication`

test/gurka/test_turn_lanes.cc
```diff
@@ -57,7 +57,7 @@ TEST(Standalone, TurnLanes) {
 
   ASSERT_TRUE(prev_edge);
   EXPECT_EQ(prev_edge->turn_lanes_size(), 3);
-  EXPECT_EQ(prev_edge->TurnLanesToString(), "[ left | through | through;right ACTIVE ]");
+  EXPECT_EQ(prev_edge->TurnLanesToString(), "[ left | through | through;*right* ACTIVE ]");
```
dgearhart, line 60: this looks great and easy to read :pray:

dgearhart (approved):
thanks @mandeepsandhu :ship:

danpaz (approved):

PR #2660, opened 2020-10-20
bidirectional a* routing with specified time

kevinkreiser:
This is a first step towards using localtime to influence which speed information is used in bidirectional a*. At the moment we dont support any time dependence in bidirectional a*. this means that, we dont care about timed turn/access restrictions and we dont properly ask for historical speed.

This pr adds a new temporary date_time_type called `invariant` in addition to the existing ones (`current`, `depart_at`, `arrive_by`). `invariant` allows us to run a route at a specific time and not vary that time of day over the course of the route. maybe a good description would be "point in time routing". this case is striaght forward for bidirectional a* (the most performant routing algorithm). It allows bidirectional a* to become somewhat time aware in that it will now use the specified time to properly look at timed turn/access restrictions as well as properly index into the historical speed information.

this pr does not fix the `GetSpeed` function of graph tile to notice that we are not near the current time and not use current speed information if available. that will be handled in a future pr.


src/loki/trace_route_action.cc
```diff
@@ -174,8 +174,7 @@ void loki_worker_t::init_trace(Api& request) {
 
 void loki_worker_t::trace(Api& request) {
   init_trace(request);
-  const auto& costing = Costing_Enum_Name(request.options().costing());
-  if (costing == "multimodal") {
+  if (request.options().costing() == Costing::multimodal) {
```
kevinkreiser, line 177: simplification i noticed and just did

src/thor/bidirectional_astar.cc
```diff
@@ -126,10 +126,9 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
     return false;
   }
 
-  // Update the time information
+  // Update the time information even if time is invariant to account for timezones
   auto offset_time =
-      from_transition ? time_info
-                      : time_info.forward(pred.cost().secs, static_cast<int>(nodeinfo->timezone()));
+      from_transition ? time_info : time_info.forward(0.f, static_cast<int>(nodeinfo->timezone()));
```
kevinkreiser, line 131: previously it looked like we were varying the time, but there were no code paths that could get here where `time_info` would be `valid` and when its not valid then the `forward` function just exits immediately. ie this was a no-op before. now instead of varying the time we simply say `0.f` offset,  meaning same time just account for any changes in timezone.

in the future when bidirectional a* is time dependent and indeed varies the time as the route progresses we'll revert this change

src/thor/bidirectional_astar.cc
```diff
@@ -237,20 +236,22 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
     return true; // This is an edge we _could_ have expanded, so return true
   }
 
-  // TODO: actually use time_info
-  const uint64_t localtime = 0;
-  const uint32_t tz_index = 0;
+  // Skip this edge if no access is allowed (based on costing method)
+  // or if a complex restriction prevents transition onto this edge.
+  // if its not time dependent set to 0 for Allowed and Restricted methods below
+  const uint64_t localtime = time_info.valid ? time_info.local_time : 0;
```
kevinkreiser, line 242: we need to use local time for timed turn/access restrictions if this is a time dependent route. for now the only time dependnce that can make it to here is `date_time_type == invariant`

src/thor/bidirectional_astar.cc
```diff
@@ -237,20 +236,22 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
...
+                           &edgestatus_forward_, localtime, time_info.timezone_index)) {
     return false;
   }
 
   // Get cost. Separate out transition cost.
   Cost transition_cost = costing_->TransitionCost(meta.edge, nodeinfo, pred);
-  // TODO: actually use time_info
-  Cost newcost = pred.cost() + transition_cost + costing_->EdgeCost(meta.edge, tile);
+  Cost newcost =
+      pred.cost() + transition_cost + costing_->EdgeCost(meta.edge, tile, time_info.second_of_week);
```
kevinkreiser, line 254: actually use the time of the week to let the costing pick the right historical data to use

src/thor/bidirectional_astar.cc
```diff
@@ -317,10 +318,9 @@ bool BidirectionalAStar::ExpandReverse(GraphReader& graphreader,
     return false;
```
kevinkreiser, line 318: same changes in this function as in the forward directions function

src/thor/route_action.cc
```diff
@@ -231,39 +231,50 @@ void thor_worker_t::route(Api& request) {
...
+  for (auto* alg : std::vector<PathAlgorithm*>{
+           &multi_modal_astar,
+           &timedep_forward,
+           &timedep_reverse,
+           &astar,
+           &bidir_astar,
+           &bss_astar,
+       }) {
+    alg->set_interrupt(interrupt);
+  }
```
kevinkreiser, line 246: this seemed preferable to setting them all individually but honestly i dont like it that much either

src/thor/route_action.cc
```diff
@@ -231,39 +231,50 @@ void thor_worker_t::route(Api& request) {
...
   if (routetype == "bikeshare") {
-    bss_astar.set_interrupt(interrupt);
+    valhalla::midgard::logging::Log("algorithm::astar_bss ", " [ANALYTICS] ");
     return &bss_astar;
   }
 
   // If the origin has date_time set use timedep_forward method if the distance
   // between location is below some maximum distance (TBD).
-  if (origin.has_date_time()) {
+  if (origin.has_date_time() && options.date_time_type() != Options::invariant) {
```
kevinkreiser, line 262: so invariant types dont go to the time depenent algorithms anymore

src/thor/route_action.cc
```diff
@@ -276,14 +287,23 @@ thor::PathAlgorithm* thor_worker_t::get_path_algorithm(const std::string& routet
     for (auto& edge2 : destination.path_edges()) {
       if (edge1.graph_id() == edge2.graph_id() ||
           reader->AreEdgesConnected(GraphId(edge1.graph_id()), GraphId(edge2.graph_id()))) {
+
+        // We need time dependence so we use timedep_foward when its trivial (should be short so not a
+        // real issue that its not invariant)
+        if (options.has_date_time_type() && options.date_time_type() == Options::invariant) {
```
kevinkreiser, line 293: in the case that its a trivial route then we do let `timedep_forward` handle invariant as i didnt want to modify `astar` to do time dependence for just this tiny use-case

src/thor/route_action.cc
```diff
@@ -276,14 +287,23 @@ thor::PathAlgorithm* thor_worker_t::get_path_algorithm(const std::string& routet
...
+
+        // Otherwise we can use regular astar
         valhalla::midgard::logging::Log("algorithm::astar ", " [ANALYTICS] ");
-        astar.set_interrupt(interrupt);
         return &astar;
       }
     }
   }
+
+  // No other special cases we land on bidirectional a*
```
kevinkreiser, line 305: other than the special cases we fall through to bidirectional a*, specifically for `invariant` date_time types

src/thor/route_action.cc
```diff
@@ -356,7 +376,8 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
   for (auto origin = ++correlated.rbegin(); origin != correlated.rend(); ++origin) {
     // Get the algorithm type for this location pair
     auto destination = std::prev(origin);
-    thor::PathAlgorithm* path_algorithm = get_path_algorithm(costing, *origin, *destination);
+    thor::PathAlgorithm* path_algorithm =
+        get_path_algorithm(costing, *origin, *destination, api.options());
```
kevinkreiser, line 380: needed to pass the options so that the algorithm selection could check the date_time_type

src/worker.cc
```diff
@@ -293,7 +293,8 @@ rapidjson::Document from_string(const std::string& json, const valhalla_exceptio
 
 void add_date_to_locations(Options& options,
                            google::protobuf::RepeatedPtrField<valhalla::Location>& locations) {
-  if (options.has_date_time() && locations.size()) {
+  // otherwise we do what the person was asking for
+  if (options.has_date_time() && !locations.empty()) {
```
kevinkreiser, line 297: new protobuf has `empty`. it also has a lot of the stuff we have in our protoconversions.. we should probably delete some of those

src/worker.cc
```diff
@@ -304,6 +305,9 @@ void add_date_to_locations(Options& options,
       case Options::arrive_by:
         locations.Mutable(locations.size() - 1)->set_date_time(options.date_time());
         break;
+      case Options::invariant:
+        for (auto& loc : locations)
+          loc.set_date_time(options.date_time());
```
kevinkreiser, line 310: so, when its invariant, that means that all locations are computed at the same date_time

src/worker.cc
```diff
@@ -616,37 +620,50 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
   auto date_time_type = rapidjson::get_optional<unsigned int>(doc, "/date_time/type");
-  auto date_time_value = rapidjson::get_optional<std::string>(doc, "/date_time/value");
   if (date_time_type && Options::DateTimeType_IsValid(*date_time_type)) {
     // check the type is in bounds
     auto const v = static_cast<Options::DateTimeType>(*date_time_type);
     if (v >= Options::DateTimeType_ARRAYSIZE)
       throw valhalla_exception_t{163};
     options.set_date_time_type(static_cast<Options::DateTimeType>(v));
     // check the value exists for depart at and arrive by
+    auto date_time_value = v != Options::current
```
kevinkreiser, line 630: this works the same but seemed simplier. if its a current time type the string is current other wise its the user provided string

src/worker.cc
```diff
@@ -616,37 +620,50 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
         throw valhalla_exception_t{160};
       else if (v == Options::arrive_by)
         throw valhalla_exception_t{161};
+      else if (v == Options::invariant)
+        throw valhalla_exception_t{165};
     }
-    // check the value is sane for depart at and arrive by
-    if (v != Options::current && !baldr::DateTime::is_iso_valid(*date_time_value))
+    // check the value is sane
+    if (*date_time_value != "current" && !baldr::DateTime::is_iso_valid(*date_time_value))
```
kevinkreiser, line 642: works the same but is simpler

src/worker.cc
```diff
@@ -616,37 +620,50 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
-      options.set_date_time("current");
+    options.set_date_time(*date_time_value);
   } // not specified but you want transit, then we default to current
   else if (options.has_costing() &&
            (options.costing() == multimodal || options.costing() == transit)) {
     options.set_date_time_type(Options::current);
     options.set_date_time("current");
   }
 
+  // failure scenarios with respect to time dependence
```
kevinkreiser, line 652: a long time ago it seems 141 and 142 were removed from the code :grimacing: so i put them back and added the new temporary one for invariant

genadz (approved):

src/thor/bidirectional_astar.cc
PureW: Why make the change if the previous code was a no-op and the change should be reverted later?

proto/options.proto
```diff
@@ -155,6 +155,8 @@ message Options {
     current = 0;
     depart_at = 1;
     arrive_by = 2;
+    // Note: temporary until bidirectional a* does time dependence, does not vary the time as the route progresses
+    invariant = 3;
```
danpaz, line 159: Is this truly temporary or is there a possibility this would be a useful request option even after bidir a* gets time dependence? This could be useful to a user who wants to say "don't try to predict the future, just assume current conditions will stay the same throughout my route." Just a thought.

src/thor/bidirectional_astar.cc
```diff
@@ -939,8 +939,7 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
   std::vector<std::unordered_set<GraphId>> shared_edgeids;
 
   // get maximum amount of sharing parameter based on origin->destination distance
-  float max_sharing =
-      allow_alternates ? get_max_sharing(options.locations(0), options.locations(1)) : 0.f;
+  float max_sharing = allow_alternates ? get_max_sharing(origin, dest) : 0.f;
```
danpaz, line 942: Good catch 👍

proto/options.proto
kevinkreiser: good question! i dont know. you're right, it doesnt have to be temporary it could be useful indefinitely

src/thor/bidirectional_astar.cc
kevinkreiser: because the code wont work if we dont :smile: 

i could change it to this to make it future proof:

```c++
auto offset_time =
￼      from_transition ? time_info : time_info.forward(invariant ? 0.f, pred.cost().secs, static_cast<int>(nodeinfo->timezone()));
```

PureW: > i could change it to this to make it future proof

This sounds good I think. At least it signals to the reader that this functionality definitely depends on `invariant`

bench/CMakeLists.txt
```diff
@@ -17,14 +17,13 @@ add_dependencies(benchmarks utrecht_tiles)
...
-    ${target_file}.cc
-    ${VALHALLA_SOURCE_DIR}/third_party/microtar/src/microtar.c)
+    ${target_file}.cc)
   set_target_properties(${target_name} PROPERTIES FOLDER "Benchmarks")
   set_target_properties(${target_name}
     PROPERTIES
     COMPILE_DEFINITIONS
     VALHALLA_SOURCE_DIR="${VALHALLA_SOURCE_DIR}/")
-  target_link_libraries(${target_name} valhalla benchmark::benchmark)
+  target_link_libraries(${target_name} valhalla_test benchmark::benchmark)
```
kevinkreiser, line 26: this relies on some stuff in the new testlibrary. more about that below..

proto/trip.proto
```diff
@@ -325,6 +325,7 @@ message TripLeg {
   optional BoundingBox bbox = 9;
   optional ShapeAttributes shape_attributes = 10;
   repeated Incident incidents = 11;
+  repeated string algorithms = 12;
```
kevinkreiser, line 328: this is just a nice way to get insight into what algorithms were used to make a leg. we could live without it but its good book keeping and useful for integration style unitesting

bench/thor/routes.cc
```diff
@@ -9,7 +9,7 @@
 #include "midgard/pointll.h"
 #include "sif/autocost.h"
 #include "sif/costfactory.h"
-#include "test/util/traffic_utils.h"
```
kevinkreiser, line 12: this is all code motion, nothing to see here

src/thor/bidirectional_astar.cc
```diff
@@ -126,10 +127,11 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
...
 
-  // Update the time information
-  auto offset_time =
-      from_transition ? time_info
-                      : time_info.forward(pred.cost().secs, static_cast<int>(nodeinfo->timezone()));
+  // Update the time information even if time is invariant to account for timezones
+  auto seconds_offset = invariant ? 0.f : pred.cost().secs;
+  auto offset_time = from_transition
+                         ? time_info
+                         : time_info.forward(seconds_offset, static_cast<int>(nodeinfo->timezone()));
```
kevinkreiser, line 134: i made that change we came up with @PureW

src/thor/route_action.cc
```diff
@@ -425,11 +444,16 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
-                              path.end(), *origin, *destination, throughs, leg, interrupt, &vias);
+                              path.end(), *origin, *destination, throughs, leg, algorithms, interrupt,
+                              &vias);
         path.clear();
         vias.clear();
       }
     }
+
+    // if we just made a leg that means we are done recording which algorithms were used
+    if (path.empty())
```
kevinkreiser, line 455: this is a bit tricky because of alternates. we have to wait until that loop is finished above before we clear this or the first alternate wont have algorithms set

src/thor/trace_route_action.cc
```diff
@@ -162,7 +162,7 @@ void thor_worker_t::route_match(Api& request) {
     thor::TripLegBuilder::Build(options, controller, *reader, mode_costing, path.begin(), path.end(),
                                 *options.mutable_locations()->begin(),
```
kevinkreiser, line 163: this is all just adding algorithm names to the leg otherwise just some standardization of how we add a leg to a route

test/CMakeLists.txt
```diff
@@ -4,6 +4,19 @@ set_target_properties(gtest_main PROPERTIES FOLDER "Dependencies")
 set_target_properties(gmock PROPERTIES FOLDER "Dependencies")
 set_target_properties(gmock_main PROPERTIES FOLDER "Dependencies")
 
+# common things that the tests need
```
kevinkreiser, line 7: this was a nice little clean up. i took the 3 test utility headers we had and combined them into one header with an implementation file. i did the same with gurka. then i made a new test library out of them with this target which is really convenient because then all the test targets can just link this. cmake automatically handles giving them visibility of include directories and other dependent libraries (ie libvalhalla and the gtest libs). so now the unit test targets (and the bench targets) are much simplier. i measured a 6% speed up of a full build from no ccache using a single thread. margins will be very low for high core systems but our CI only has 2 cores so hopefuly this will speed up CI a tiny bit

test/gurka/gurka.cc
```diff
@@ -0,0 +1,1139 @@
+#include "baldr/directededge.h"
```
kevinkreiser, line 1: this was just copy paste from gurka.h

test/test.h
```diff
@@ -1,16 +1,25 @@
 // -*- mode: c++ -*-
+#pragma once
```
kevinkreiser, line 2: i combined all 3 test util headers into this header

test/gurka/gurka.h
```diff
@@ -786,14 +200,16 @@ valhalla::Api locate(const map& map,
...
-  return result;
-}
+// TODO: bidirectional edges overlap shapes so its hard to visualize them
+/**
+ * Dumps the gurka map to geojson including edges and nodes. Properties will be street names and node
+ * names. Edge ids are also included
+ * @param graph    the gurka map
+ * @return geojson string
+ */
+std::string dump_geojson_graph(const map& graph);
```
kevinkreiser, line 212: i added this new function to dump geojson of a graph from the information in the gurka::map. right now the information is sparse but over time we can add info to the properties. the idea is that we have a way to see what is in the graphs taht we create. i  understand the ascii is supposed to do that but frankly sometimes its very hard to reason about what edges are where in space. further more the properties allow us to see if the data is getting set on the edges correctly.

valhalla/thor/trafficalgorithm.h
```diff
@@ -1,67 +0,0 @@
-#ifndef VALHALLA_THOR_TRAFFICALGORITHM_H_
```
kevinkreiser, line 1: this was an old school POC we did for traffic influenced routing. it no longer serves a purpose so it gets the :axe:

src/worker.cc
```diff
@@ -616,37 +620,50 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+  if (options.has_date_time_type()) {
+    if (options.date_time_type() == Options::arrive_by) {
+      if (options.costing() == multimodal || options.costing() == transit)
+        throw valhalla_exception_t{141};
+      if (options.action() == Options::isochrone)
+        throw valhalla_exception_t{142};
+    } else if (options.date_time_type() == Options::invariant && options.action() != Options::route) {
```
kevinkreiser, line 659: we need to resolve this. map matching needs to at least get this date information and use it for costing

test/gurka/test_route.cc
```diff
@@ -215,11 +215,178 @@ TEST(AutoDataFix, deprecation) {
...
+  }
+};
+
+gurka::map AlgorithmTest::map = {};
+uint32_t AlgorithmTest::current = 0, AlgorithmTest::historical = 0, AlgorithmTest::constrained = 0;
+
+uint32_t speed_from_edge(const valhalla::Api& api) {
+  auto km = api.trip().routes(0).legs(0).node(0).edge().length();
+  auto h = api.trip().routes(0).legs(0).node(1).cost().elapsed_cost().seconds() / 3600.0;
+  return static_cast<uint32_t>(km / h + .5);
```
kevinkreiser, line 311: look at all edges to make sure they match

valhalla/thor/multimodal.h
```diff
@@ -57,12 +57,20 @@ class MultiModalPathAlgorithm : public PathAlgorithm {
...
               const sif::mode_costing_t& mode_costing,
               const sif::TravelMode mode,
-              const Options& options = Options::default_instance());
+              const Options& options = Options::default_instance()) override;
+
+  /**
+   * Returns the name of the algorithm
+   * @return the name of the algorithm
+   */
+  virtual std::string name() const override {
```
kevinkreiser, line 66: this should be const reference and then it should reference a static const string

src/worker.cc
kevinkreiser: for now we'll just allow trace_* and in a follow up pr we will actually recost the results of map matching based on whatever time information you have set in the route

proto/options.proto
kevinkreiser: lets keep it!

src/thor/bidirectional_astar.cc
```diff
@@ -516,16 +523,29 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
...
+  bool invariant = options.has_date_time_type() && options.date_time_type() == Options::invariant;
   auto forward_time_info = TimeInfo::make(origin, graphreader, &tz_cache_);
   auto reverse_time_info = TimeInfo::make(destination, graphreader, &tz_cache_);
 
+  // When a timedependent route is too long in distance it gets sent to this algorithm. It used to be
+  // the case that this algorithm called EdgeCost without a time component. This would result in
+  // timedependent routes falling back to time independent routing. Now that this algorithm is time
+  // aware we will be tracking time in one direction of the search. To revert to previous behavior
+  // you can uncomment the code below and get a time independent route in the fallback scenario.
+  //  if (!invariant) {
```
kevinkreiser, line 535: this is an important change in what happens when we fail over from timedependent to bidir a* due to exceeding max length. please read!

kevinkreiser:
@PureW @danpaz can i get another look at this. I did RAD and found that there were changes in timedependent routes. Ive put a comment on the pr and in bidirectional a* that explains why and tested that reverting that change removes the diff. Otherwise there were no diffs. THe performance for bidira* with historical traffic use is about a 70% performance penalty. It takes about 42 seconds to blow through 14k routes without historical traffic and about 72 seconds when you enable it. This is still vastly improved over the timedep algorithm which takes something like 160 seconds.

PureW (approved):
:tada: This is Great!

danpaz (approved):
👍

PureW (approved):

PR #2661, opened 2020-10-20
Recost shortcuts in bidastar: first approach

genadz:
##  Tasklist

 - [x] Add tests

## Requirements / Relations


genadz:
@kevinkreiser there are some open questions I wanted to discuss:
1. Should we skip path in case recosting fails ? or we should fallback to the current logic (right now is fallback).
2. Should we recost transition costs for recovered shortcuts ?

src/thor/bidirectional_astar.cc
```diff
@@ -916,10 +918,167 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader, const valhalla
...
+/**
+ * Recover shortcut edge and recost it
+ * @param  graphreader  Graph tile reader.
+ * @param  shortcut_id  Edge id of the shortcut to be recosted
+ * @param  costing      Costing function
+ * @param  transition_cost  Transition cost for the shortcut edge
+ * @param  previous_cost    Total cost from the previous edge
+ * TODO: add time information
+ */
+std::vector<PathInfo> recost_shortcut_forward(GraphReader& graphreader,
```
PureW, line 930: Is recosting shortcuts something that could benefit all A-star algorithms? If so, could this be implemented on the base-class?

genadz: I think yes, will check

PureW:
Could you add some gurka tests for this recovering of shortcuts?

kevinkreiser:
@PureW we already have tests for recovering shortcuts i dotn think we need more just for that. but we do need them specifically for this. the main point here is that the final path doesnt have shortcuts in it. so it would be enough to make a gurka map that has shortcuts, get a route that takes a shortcut, set the constrained speed on the non shortcut edges to a known value that is substantially different than the shortcut and then check that:

1. you get back no shortcut edges
2. your recosting works in that you get speeds/costs that look like the constrained speed you set

kevinkreiser:
@gena with respect to:

> Should we recost transition costs for recovered shortcuts ? (I mean, we have transition cost between some edge in path and shortcut edge -> should we replace it by transition cost between the edge and first superseded edge that was recovered from shortcut ?)

No, you dont need to recompute the transition cost for those as the value will be the same.

src/thor/bidirectional_astar.cc
kevinkreiser: heavy plus for this one

src/thor/bidirectional_astar.cc
```diff
@@ -916,10 +918,167 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader, const valhalla
...
+                                         uint32_t last_label_idx,
+                                         std::vector<PathInfo>& path) {
+  // we should go from the beginning to the end because of possible recostings;
+  // so, at first we should collect edge label indexes
+  std::stack<uint32_t> forward_idxs;
+  // Recover forward labels using predecessor indexes
+  for (auto edgelabel_index = last_label_idx; edgelabel_index != kInvalidLabel;
+       edgelabel_index = edgelabels_forward_[edgelabel_index].predecessor()) {
+    forward_idxs.push(edgelabel_index);
+  }
```
kevinkreiser, line 979: why make this change? its a bunch of extra allocation and a whole extra loop over the path. what does this give us other than a slightly clearer `while` loop below?

kevinkreiser: unless there is a clear reason to do this, i would highly suggest we revert this to using the label index and labelset directly in the while loop as was done before

src/thor/bidirectional_astar.cc
```diff
@@ -916,10 +918,167 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader, const valhalla
...
+ * @param  transition_cost  Transition cost for the shortcut edge
+ * @param  previous_cost    Total cost from the previous edge
+ * TODO: add time information
+ */
+std::vector<PathInfo> recost_shortcut_forward(GraphReader& graphreader,
+                                              const GraphId& shortcut_id,
+                                              const DynamicCost& costing,
+                                              Cost transition_cost,
+                                              Cost previous_cost) {
+  auto edges = graphreader.RecoverShortcut(shortcut_id);
```
kevinkreiser, line 935: as soon as this fails should we log and return?

src/thor/bidirectional_astar.cc
kevinkreiser: oh i see in the comment hints at it. you are saying that we have to start at the beginning to get the correct cost? i'm not sure its actually true though. i wonder what kind of penalty this adds. i think we can work around it by looking one label ahead and using that label as the "previous cost" for the recosting of the shortcut to avoid all this allocation

src/thor/bidirectional_astar.cc
```diff
@@ -916,10 +918,167 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader, const valhalla
...
+    forward_idxs.pop();
+    const BDEdgeLabel& edgelabel = edgelabels_forward_[edgelabel_idx];
+    // get current cost
+    Cost cost = path.empty() ? Cost() : path.back().elapsed_cost;
+
+    std::vector<PathInfo> recosted_segment;
+    // recost shortcut edge
+    if (edgelabel.shortcut()) {
+      recosted_segment = recost_shortcut_forward(graphreader, edgelabel.edgeid(), *costing_,
+                                                 edgelabel.transition_cost(), cost);
```
kevinkreiser, line 995: the last PathId stores boths the cost so far and the transition cost at the beginning of the PathID. so i think you might have the transition cost in the wrong place. i always confuse mysefl on this but thats whta it looks like to me here

src/thor/bidirectional_astar.cc
genadz: let's suppose we go from the end to the beginning (as before), and we already processed some edges: `edge_n, edge_n-1, ...., edge_k` and then meet shortcut and recost the edge - it means that costs for all edges from `n` to `k` will be invalid anymore and we must add difference between shortcut cost and recosted shortcut to each of them.
If we implement this approach directly - it will require approximately `n * (number of shortcuts)` additional operations (change costs) and no additional memory. We can optimize it a little: let's save all cost changes to stack (`~ (number of shortcuts)` additional memory) and apply these changes once on the second pass.

src/thor/bidirectional_astar.cc
genadz: hm, didn't get this note. could you run the details ?

src/thor/bidirectional_astar.cc
genadz: we already log errors in `RecoverShortcut` method, so I think we shouldn't log here. but, it might be reasonable not to continue and treat recovering fail as recosting fail - in this case no necessary to call `recost_forward` function becasue we can use known shortcut cost.

src/thor/bidirectional_astar.cc
genadz: all A-start algorithms inherit from `PathAlgorithm` interface. as for me, it's too common in order to add `recost_shortcut_forward` to this interface. instead of that, I will define this function in a separate *.h file (in `src/thor` folder) that can be used by everyone

alexey-gl:
Hello
While solving my #2692 problem I have tried to use this PR as possible solution.
Unfortunately it has not solved my issue completely.
I noticed when the route has several subsequent shortcuts only the first one was fully restored with original way ids. The rest still have way_id = 0 (i.e. they are still shortcuts). And I have to restored them during postprocessing with graph_reader.RecoverShortcut(...) stuff.
I suppose this PR should solve my issue.
Hope this comment will help with this PR testing.

src/thor/bidirectional_astar.cc
```diff
@@ -571,7 +573,7 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
 
         // Terminate if the cost threshold has been exceeded.
         if (fwd_pred.sortcost() + cost_diff_ > threshold_) {
-          return FormPath(graphreader, options, origin, destination);
+          return FormPath(graphreader, options, origin, destination, forward_time_info, invariant);
```
genadz, line 576: not sure if it's fully correct to consider only forward time and ignore reverse time:
- on the one hand we adjust times here https://github.com/valhalla/valhalla/blob/master/src/thor/route_action.cc#L409 and here https://github.com/valhalla/valhalla/blob/master/src/thor/route_action.cc#L512 ;
- but on the other hand we potentially can miss time restrictions in case we have predefined arrival time.

genadz:
Implemented new approach here https://github.com/valhalla/valhalla/pull/2711 . @kevinkreiser we should compare and choose the best one

genadz:
> Hello
> While solving my #2692 problem I have tried to use this PR as possible solution.
> Unfortunately it has not solved my issue completely.
> I noticed when the route has several subsequent shortcuts only the first one was fully restored with original way ids. The rest still have way_id = 0 (i.e. they are still shortcuts). And I have to restored them during postprocessing with graph_reader.RecoverShortcut(...) stuff.
> I suppose this PR should solve my issue.
> Hope this comment will help with this PR testing.

hm, interesting note about subsequent shortcuts. I added unit test and it works)

kevinkreiser:
@genadz should we close this since we went with the other approach?

PR #2662, opened 2020-10-21
Include turn durations in shortcut building

SvetlanaBayarovich:
# Issue

Calculate shortcut speed considering turn durations between contracted edges 

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/mjolnir/shortcutbuilder.cc
```diff
@@ -471,6 +457,15 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       bool diff_names = directededge->endnode().tileid() == edge_id.tileid() &&
                         !OpposingEdgeInfoMatches(tile, directededge);
 
+      // Recalculate speed limit to include turn durations
+      uint32_t average_speed = edgeinfo.speed_limit();
+      if (edgeinfo.speed_limit() != 0 && total_turn_duration > 0) {
+        // meters per second // TODO is there a converter? remove zeros
```
SvetlanaBayarovich, line 463: Suggestions are welcome

src/mjolnir/shortcutbuilder.cc
```diff
@@ -246,9 +224,9 @@ bool CanContract(GraphReader& reader,
 
   // Get the directed edges - these are the outbound edges from the node.
   // Get the opposing directed edges - these are the inbound edges to the node.
-  const DirectedEdge* edge1 = tile->directededge(edges[match.first]);
+  const DirectedEdge* edge1 = tile->directededge(edges[0]);
```
kevinkreiser, line 227: if you move this and the one below it up above the `EdgesMatch` then you can use it there as well.

SvetlanaBayarovich: Oh, got it. That's what you were talking about yesterday

src/mjolnir/shortcutbuilder.cc
SvetlanaBayarovich: Any

SvetlanaBayarovich: Alweays

src/mjolnir/shortcutbuilder.cc
```diff
@@ -471,6 +457,15 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       bool diff_names = directededge->endnode().tileid() == edge_id.tileid() &&
                         !OpposingEdgeInfoMatches(tile, directededge);
 
+      // Recalculate speed limit to include turn durations
+      uint32_t average_speed = edgeinfo.speed_limit();
```
kevinkreiser, line 461: shouldnt be messing with the speed limit here. speed limit is what is on the actual speed limit sign. what you want to do here is look at `directededge->speed` (you'll also want to recalculate for truck costing with `directededge->truckspeed`

we have a bunch of conversion constants in `midgard/constants.h` but we seem to be missing a few:
https://github.com/valhalla/valhalla/blob/master/valhalla/midgard/constants.h#L37

we should add something like in that header:

```c++
constexpr double kKMHtoMetersPerSec = 1000./3600.;
constexpr double kMetersPerSectoKMH = 3600./1000.;
```

then we'd have something like this in your new code:

```c++
double new_duration = length  / (directededge->speed()  * kKMHtoMetersPerSec) + total_turn_duration;
uint32_t new_speed = static_cast<uint32_t>(std::round((length / new_duration) * kMetersPerSectoKHM));

//TODO: actually go fix truck turn durations in sif and the other motor vehicle costing models (taxi, bus, etc)
double new_truck_duration = length  / (directededge->truck_speed()  * kKMHtoMetersPerSec) + total_truck_turn_duration;
uint32_t new_truck_speed = static_cast<uint32_t>(std::round((length / new_truck_duration) * kMetersPerSectoKHM));
```

src/mjolnir/shortcutbuilder.cc
```diff
@@ -480,7 +475,7 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       uint32_t idx = ((length & 0xfffff) | ((shape.size() & 0xfff) << 20));
       uint32_t edge_info_offset =
           tilebuilder.AddEdgeInfo(idx, start_node, end_node, 0, 0, edgeinfo.bike_network(),
-                                  edgeinfo.speed_limit(), shape, names, tagged_names, types, forward,
+                                  average_speed, shape, names, tagged_names, types, forward,
```
kevinkreiser, line 478: dont change the speedlimit

src/mjolnir/shortcutbuilder.cc
```diff
@@ -527,6 +522,9 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       // Compute the weighted edge density
       newedge.set_density(average_density / (static_cast<float>(length)));
 
+      // Update speed to the one that takes turn duretions into account
+      newedge.set_speed(average_speed);
```
kevinkreiser, line 526: add a `todo`:

```c++
//newedge.set_truck_speed(new_truck_speed);
```

src/mjolnir/shortcutbuilder.cc
```diff
@@ -471,6 +457,15 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       bool diff_names = directededge->endnode().tileid() == edge_id.tileid() &&
                         !OpposingEdgeInfoMatches(tile, directededge);
 
+      // Recalculate speed limit to include turn durations
+      uint32_t average_speed = edgeinfo.speed_limit();
+      if (edgeinfo.speed_limit() != 0 && total_turn_duration > 0) {
+        // meters per second // TODO is there a converter? remove zeros
+        auto const speed_mps = edgeinfo.speed_limit() * 1000.0 / 3600;
+        auto const actual_time = length / speed_mps + total_turn_duration;
+        average_speed = static_cast<uint32_t>(std::round(length / actual_time * 3600 / 1000));
```
kevinkreiser, line 466: i wouldnt call it average speed. its not an average

src/mjolnir/shortcutbuilder.cc
SvetlanaBayarovich: Oh, got it. It actually took me some time to understand that the speed I should update is the one in the `newedge` and is copied there implicitly from `directededge. But I was still wondering why there was also speed_limit in the EdgeInfo. Now it makes more sense

src/mjolnir/shortcutbuilder.cc
SvetlanaBayarovich: Naming is difficult 😞. "tuns_aware_speed" is also clumsy

src/mjolnir/shortcutbuilder.cc
```diff
@@ -315,11 +294,30 @@ uint32_t ConnectEdges(GraphReader& reader,
...
   // Get the tile and directed edge.
   const GraphTile* tile = reader.GetGraphTile(startnode);
   const DirectedEdge* directededge = tile->directededge(edgeid);
 
+  // Add edge and turn duration for car
+  auto const nodeinfo = tile->node(startnode);
+  auto const turn_duration = OSRMCarTurnDuration(directededge, nodeinfo, opp_local_idx);
+  total_duration += turn_duration;
+  auto const speed = directededge->speed();
+  // TODO: check speed for zero
```
SvetlanaBayarovich, line 309: In doubt if it's needed 🤔

kevinkreiser: add an assert here that its not 0, should be good enough

src/mjolnir/shortcutbuilder.cc
```diff
@@ -315,11 +294,30 @@ uint32_t ConnectEdges(GraphReader& reader,
...
+  auto const edge_duration = directededge->length() / (speed * kKPHtoMetersPerSec);
+  total_duration += edge_duration;
+
+  // Add edge and turn duration for truck
+  // Currently use the same as for cars. TODO: implement for trucks
+  total_truck_duration += turn_duration;
+  auto const truck_speed = tile->GetSpeed(directededge, kNoFlowMask, kInvalidSecondsOfWeek, true);
```
kevinkreiser, line 316: why dont you call `GetSpeed` above for the regular speed too?

src/mjolnir/shortcutbuilder.cc
```diff
@@ -527,6 +531,20 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       // Compute the weighted edge density
       newedge.set_density(average_density / (static_cast<float>(length)));
 
+      // Update speed to the one that takes turn durations into account
+      uint32_t new_speed = directededge->speed();
+      if (total_duration > 0) {
```
kevinkreiser, line 536: maybe change these ifs to assertions since its insane that it would be 0

src/mjolnir/shortcutbuilder.cc
```diff
@@ -432,7 +435,8 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       // Connect edges to the shortcut while the end node is marked as
       // contracted (contains edge pairs in the shortcut info).
       uint32_t rst = 0;
-      uint32_t opp_local_idx = 0;
+      // For turn duration calculation during contraction
+      uint32_t opp_local_idx = directededge->opp_local_idx();
```
kevinkreiser, line 439: this isnt actually needed right? it would be just as fine to set it to `-1` right?

src/mjolnir/shortcutbuilder.cc
```diff
@@ -315,11 +294,30 @@ uint32_t ConnectEdges(GraphReader& reader,
...
+  // Currently use the same as for cars. TODO: implement for trucks
+  total_truck_duration += turn_duration;
+  auto const truck_speed = tile->GetSpeed(directededge, kNoFlowMask, kInvalidSecondsOfWeek, true);
+  // TODO: check speed for zero
+  auto const edge_duration_truck = directededge->length() / (truck_speed * kKPHtoMetersPerSec);
+  total_truck_duration += edge_duration_truck;
+
   // Copy the restrictions and opposing local index. Want to set the shortcut
   // edge's restrictions and opp_local_idx to the last directed edge in the chain
   opp_local_idx = directededge->opp_local_idx();
```
kevinkreiser, line 323: i dont even know why we need to set this at all. why not use the value from the edge directly and remove this parameter from the function?

src/mjolnir/shortcutbuilder.cc
SvetlanaBayarovich: Yeah if speed can't be 0 on the edges then this should be some meaningful positive value. Actually the only incorrect case I can imagine is when edge speed is zero and the should result in zero speed on the whole shortcut. But we'll have division by zero first...

src/mjolnir/shortcutbuilder.cc
SvetlanaBayarovich: Great idea, done

src/mjolnir/shortcutbuilder.cc
SvetlanaBayarovich: I'm not sure I get your idea... When the cycle that does `ConnectEdges` finishes we know the last node but I don't think we know the last edge added to the shortcut in order to obtain `opp_local_idx` from it.
Also it is useful for me to know this previous edge idx to calculate the heading of the incoming edge. Alternatively I could certainly reverse `shape` and get "direction along polyline" but current approach looks cheaper to me

src/mjolnir/shortcutbuilder.cc
SvetlanaBayarovich: Not getting the idea 🤔

src/mjolnir/shortcutbuilder.cc
kevinkreiser: nevermind, i see now that it does the first one by iteslf and moves to the next, sorry about that

src/mjolnir/shortcutbuilder.cc
kevinkreiser: sorry i misread the code!

src/mjolnir/shortcutbuilder.cc
```diff
@@ -400,8 +398,15 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       DirectedEdge newedge = *directededge;
       uint32_t length = newedge.length();
 
-      // For computing weighted density along the shortcut
+      // For computing weighted density and total turn duration along the shortcut
       float average_density = length * newedge.density();
+      uint32_t const speed = tile->GetSpeed(directededge, kNoFlowMask);
```
genadz, line 403: do we really need to call `GetSpeed`? in this case it equals to `directededge->speed()`

kevinkreiser: we dont but its convenient to do so because it is what actual sif::DynamicCost and the others do. if for some reason we changed how GetSpeed worked underneath, then we'd get that change here as well as in the costing. the point is that we do the same thing the costing does so that this number and the number costing use are in sync. otherwise i agree, its very heavy for something so trivial. its more about making sure stuff stays in sync.

kevinkreiser (approved):
looks good to me just need to do some rounds of RAD. we should expect to see some differences in the duration metric that comes out of the `run_with_server.py` script and potentially (but hopefully not too many) path differences

src/mjolnir/shortcutbuilder.cc
```diff
@@ -527,6 +533,17 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       // Compute the weighted edge density
       newedge.set_density(average_density / (static_cast<float>(length)));
 
+      // Update speed to the one that takes turn durations into account
+      assert(total_duration > 0);
```
PureW, line 537: Would it make sense to be more direct here and either log an error or raise an exception? Since this is not in the hot-loop at runtime, we can take more care to find issues with the input data.

test/predictive_traffic.cc
```diff
@@ -105,15 +105,30 @@ TEST(PredictiveTraffic, test_get_speed) {
...
-  tile->GetSpeed(de, kConstrainedFlowMask, kConstrainedFlowSecondOfDay, &flow_sources);
+  tile->GetSpeed(de, kConstrainedFlowMask, kConstrainedFlowSecondOfDay, false, &flow_sources);
   EXPECT_FALSE(flow_sources & kPredictedFlowMask) << "Expected flow_sources not to include predicted";
 
-  tile->GetSpeed(de, kNoFlowMask, kConstrainedFlowSecondOfDay, &flow_sources);
+  tile->GetSpeed(de, kNoFlowMask, kConstrainedFlowSecondOfDay, false, &flow_sources);
 
   EXPECT_FALSE(flow_sources & kPredictedFlowMask) << "Expected flow_sources not to include predicted";
+
+  // Test case for turck speed
```
PureW, line 118:
```suggestion
  // Test case for truck speed
```

src/mjolnir/shortcutbuilder.cc
kevinkreiser: yeah i suggested to use the assert. i dont think there is any way that we set a 0 speed on directed_edges which is why i think an assert is ok. basically it says this is likely to never happen. remember that we only make shortcuts on level 0 and 1 edges which are all motorvehicle access only and im not sure we should ever get to the point where we are storing those types of edges with a speed of 0

kevinkreiser: oh i thought this was one of the other asserts above. i guess in this case it would only happen if the edges length were 0, which again is one of those things that we have a bit of code in there that says length must be at least 1

kevinkreiser (approved):
We reviewed the changes separately and as expected, there were a decent number of routes changed the bulk of which had just slightly higher weights. this makes a lot of sense. the original route got more expensive by a little bit and most have been close in cost to the other route which is now slightly cheaper than the original route but only slightly more expensive than the original cost of the original route :smile:

PR #2664, opened 2020-10-27
fix race condition in incident loading tests

kevinkreiser:
In the last round of extensive unit tests for incident loading, some platforms suffered from race conditions. It turns out unit testing threaded stuff is hard :smile: 

Anyway, the main gist here is that when you use a condition variable is almost never enough to use it without a signaling flag and a predicate that wraps it. I'll mark the spot in the code where you could make it so that no matter what the incident singleton would always fail to load in the specified amount of time.

### DIFF
--- a/src/baldr/incident_singleton.h
+++ b/src/baldr/incident_singleton.h
@@ -30,9 +30,11 @@ struct incident_singleton_t {
 protected:
   // parameter pack to share state between daemon thread and singleton instance
   struct state_t {
-    std::atomic<bool> lock_free;
-    std::condition_variable signal;
-    std::mutex mutex;
+    std::atomic<bool> initialized;  // whether or not the watcher thread has done 1 load of incidents
+    std::atomic<bool> lock_free;    // whether or not we can skip locking around cache operations
+    std::condition_variable signal; // how the watcher tells the main thread its done its first load
+    std::mutex mutex;               // for locking on cache operations
+    // the actual cache where tiles are stored
     std::unordered_map<uint64_t, std::shared_ptr<const valhalla::IncidentsTile>> cache;
   };
   // we use a shared_ptr to wrap the state between the watcher thread and the main threads singleton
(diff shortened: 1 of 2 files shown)


src/baldr/incident_singleton.h
```diff
@@ -46,23 +48,32 @@ struct incident_singleton_t {
...
-      : state{new state_t{}}, watcher(watch, config, tileset, state, interrupt()) {
+                       const std::unordered_set<valhalla::baldr::GraphId>& tileset,
+                       const watch_function_t& watch_func = incident_singleton_t::watch)
+      : state{new state_t{}}, watcher(watch_func, config, tileset, state, interrupt()) {
     // let the thread control its own lifetime
     watcher.detach();
     // check how long we should wait to find out if its initialized
     auto max_loading_latency =
         config.get<time_t>("incident_max_loading_latency", DEFAULT_MAX_LOADING_LATENCY);
+
```
kevinkreiser, line 72: if you added a bit of code here like:

```c++
std::this_thread::sleep_for(std::chrono::seconds(1));
```

This would cause the watcher thread in the background to finish initializing and to notify BEFORE we ever got to the `cv::wait` below. It turns out that does doesnt work :smile:  if the condition variable is notified before you wait on it you'll never hear about it. basically the guidance is, never wait on just the condition variable always add a flag and a predicate that wraps it. in this way the predicate will always be checked before waiting (and bail early if already true) and after the time out or notify occurs returning whatever value the predicate returns.

Now the only way we can hit that throw is if the incident watching thread actually takes longer than the configured timeout to say that its initialized, which should be never (unless you have billions of incident tiles and a really tight timeout, which the unti tests dont have).

src/baldr/incident_singleton.h
```diff
@@ -46,23 +48,32 @@ struct incident_singleton_t {
...
-   * @param tileset  an mmapped graph tileset (ie static) allows incident loading to be lock-free
+   * @param config      lets the daemon thread know where/how to look for incidents
+   * @param tileset     an mmapped graph tileset (ie static) allows incident loading to be lock-free
+   * @param watch_func  the function the background thread will run to keep incident caches up to date
    */
   incident_singleton_t(const boost::property_tree::ptree& config,
-                       const std::unordered_set<valhalla::baldr::GraphId>& tileset)
-      : state{new state_t{}}, watcher(watch, config, tileset, state, interrupt()) {
+                       const std::unordered_set<valhalla::baldr::GraphId>& tileset,
+                       const watch_function_t& watch_func = incident_singleton_t::watch)
```
kevinkreiser, line 65: i also added the ability to specify a watch function so that i could reenable the constructor test and test both outcomes (one where initialization passes and one where it fails)

genadz (approved):

PR #2665, opened 2020-10-27
Ambiguous Local Time when Within DST Change Hour

kevinkreiser:
If you select a local time that is within the DST change window and then apply a timezone to it it seems that the datetime library notices that it has two options for interpreting the time, either you meant the time BEFORE or AFTER the DST change and it doesnt know which. We might try asking Howard Hinnant if he can suggest a work around or a way of telling the library to choose between the two. As a last resort we can catch such an exception, move the time an hour and move it back after the transformation.

EDIT: Howard Hinnant already provided a solution for this which is to actually allow you to tell the library to choose which option you want :smile: so i'll update the PR with that and set this for review

### DIFF
--- a/valhalla/baldr/time_info.h
+++ b/valhalla/baldr/time_info.h
@@ -134,7 +134,7 @@ struct TimeInfo {
       LOG_ERROR("Could not parse provided date_time: " + date_time);
       return {false, 0, 0, kConstrainedFlowSecondOfDay, 0, false, nullptr};
     }
-    const auto then_date = date::make_zoned(tz, parsed_date);
+    const auto then_date = date::make_zoned(tz, parsed_date, date::choose::latest);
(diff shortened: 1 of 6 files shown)


danpaz (approved):
Nice find! LGTM

valhalla/baldr/time_info.h
```diff
@@ -134,7 +134,7 @@ struct TimeInfo {
       LOG_ERROR("Could not parse provided date_time: " + date_time);
       return {false, 0, 0, kConstrainedFlowSecondOfDay, 0, false, nullptr};
     }
-    const auto then_date = date::make_zoned(tz, parsed_date);
+    const auto then_date = date::make_zoned(tz, parsed_date, date::choose::latest);
```
kevinkreiser, line 137: unit test fails without this

danpaz (approved):

PR #2667, opened 2020-10-29
add top_speed for all motorized vehicles

### LINKED ISSUE #2666: include top_speed for all motorized vehicles
`motor_scooter` has a `top_speed` option. It'd be nice to have the same for `auto` & `truck`.

Will PR soon.

EDIT: added all other motorized vehicles as well.

nilsnolde:
fixes #2666 

Adds the ability for users to specify a maximum speed for all motorized vehicles. Logic is exactly copied from `motor_scooter`, i.e. also roads where `edge_speed > top_speed` are penalized.

## Tasklist

 - [x] Add tests


nilsnolde:
Should be good from my side now.

kevinkreiser:
@nilsnolde awesome!

src/sif/truckcost.cc
```diff
@@ -665,6 +675,11 @@ void ParseTruckCostOptions(const rapidjson::Document& doc,
     pbf_costing_options->set_use_tolls(
         kUseTollsRange(rapidjson::get_optional<float>(*json_costing_options, "/use_tolls")
                            .get_value_or(kDefaultUseTolls)));
+
+    // top_speed
+    pbf_costing_options->set_top_speed(
```
kevinkreiser, line 680: we should do this in dynamiccost for all the costings (slowly working towards one costing to rule them all)

kevinkreiser: we dont actually have to use it for all costings but if we do it in dynamic cost we can (and eventually should) have it work for all costings

proto/options.proto
```diff
@@ -119,6 +119,7 @@ message CostingOptions {
   optional bool ignore_access = 62;
   optional bool ignore_closures = 63;
   optional bool shortest = 64;
+  optional uint32 max_speed = 65;
```
kevinkreiser, line 122: can we name this top_speed like the parameter is called in the request? also maxspeed reminds me of the osm tag for speed limit so top speed is a good to differentiate

nilsnolde: Oops, didn't see that was still in there.. Yeah, I only realized that `top_speed` was already there after a few minutes.. Will revert.

src/sif/truckcost.cc
nilsnolde: yep, makes sense

proto/options.proto
kevinkreiser: yep no worries!

kevinkreiser:
i left some comments but ill try to make one cohesive comment here.

motorscooter already has top_speed. id like to see us move the parsing and setting of that into dynamic cost and then motorscooter can keep using it and other costing models can start using it as well. if you want to feel free to do them all but if you are lacking on time feel free to do just the ones you are interested it.

thanks again for the excellent contributions and always the attention to updating the docs (i always neglect that)

nilsnolde:
Right, I did use the existing `top_speed` stuff and I actually did add them already to all motorized vehicles. I'll move things to dynamiccost where I can, makes total sense. so much repetition right now..

kevinkreiser:
@nilsnolde yep one day we'll live in a world where all costing options are in dynamic cost and we just have thin implementations in the specific costing models, one tiny step at a time :smile:

src/sif/motorscootercost.cc
```diff
@@ -75,8 +75,6 @@ constexpr ranged_default_t<float> kCountryCrossingPenaltyRange{0, kDefaultCountr
 constexpr ranged_default_t<float> kUseFerryRange{0, kDefaultUseFerry, 1.0f};
 constexpr ranged_default_t<float> kUseHillsRange{0, kDefaultUseHills, 1.0f};
 constexpr ranged_default_t<float> kUsePrimaryRange{0, kDefaultUsePrimary, 1.0f};
-constexpr ranged_default_t<uint32_t> kTopSpeedRange{kMinimumTopSpeed, kDefaultTopSpeed,
-                                                    kMaximumTopSpeed};
```
nilsnolde, line 79: I'm not sure how to deal with this in c++.. Motorscooter needs `kDefaultTopSpeed` (45 kph) as fallback if the specified `top_speed` was outside the acceptable range.. Right now it'll default to `kMaxSpeedKph`.. Not sure how to make this available by default in `dynamiccost.h` so that motor scooter can override the range check..

src/sif/truckcost.cc
```diff
@@ -434,18 +434,23 @@ bool TruckCost::AllowedReverse(const baldr::DirectedEdge* edge,
 Cost TruckCost::EdgeCost(const baldr::DirectedEdge* edge,
                          const baldr::GraphTile* tile,
                          const uint32_t seconds) const {
-  auto speed = tile->GetSpeed(edge, flow_mask_, seconds);
+  auto edge_speed = tile->GetSpeed(edge, flow_mask_, seconds);
 
-  // Use the lower or truck speed (ir present) and speed
-  uint32_t s = (edge->truck_speed() > 0) ? std::min(edge->truck_speed(), speed) : speed;
+  std::cout << top_speed_;
```
nilsnolde, line 439: oops.. will revert

src/sif/motorscootercost.cc
kevinkreiser: we have a bit of an unfortunate situation here where all of our options parsing is not polymorphic. parsing is tied to specific classes but there is no class inheritance at play. so the standard object oriented approach wont work where the base class has functionality overridden in the child class. what that means is we're kind of screwed here without a redesign.

there is a work around where, in the parsing function for the class that needs to override this, you can modify `top_speed_` to be in range. so it starts out being wrong when DynamicCosts generic parsing function sets it but then in the motroscooter parse function we "fix" it.

id say for now that is an ok way to go. ill give some more thought about it in the next day or so in case there is another more elegant way.

nilsnolde: finally had some time again for this. I restored most of the code for motorscooter, so it's overriding the defaults by dynamiccost. If you have a better way, feel free to to override:)

test/gurka/test_top_speed.cc
```diff
@@ -0,0 +1,76 @@
...
+
+TEST_F(TopSpeedTest, AutoTopSpeed) {
+  doTests("auto", {{"/costing_options/auto/top_speed", "20"}});
+}
+
+TEST_F(TopSpeedTest, TruckTopSpeed) {
+  doTests("truck", {{"/costing_options/truck/top_speed", "20"}});
+}
+
+TEST_F(TopSpeedTest, ScooterTopSpeed) {
```
nilsnolde, line 58: Would like to test that `kDefaultTopSpeed` is taken when `top_speed` is outside the acceptable range.. how do I access `top_speed_` (or `options().top_speed()`) from a `valhalla::Api` object?

nilsnolde: Ehh, was covered by the failing inline test.. not sure I actually fixed the test..

nilsnolde:
tests are failing bcs `top_speed_` is not initialized somehow in the astar tests, always `0`.. about to fix that (after I know why..).

nilsnolde:
turned out to be an easy one: `top_speed` was only set when there was a costing_options object in the request. my gurka test didn't pick it up bcs there's always a costing_options object set, no matter what.. so thank god the astar tests issued "vanilla" requests, the only ones apparently..

NOW it's good to go @kevinkreiser :)

src/sif/autocost.cc
```diff
@@ -257,7 +257,7 @@ class AutoCost : public DynamicCost {
    * estimate is less than the least possible time along roads.
    */
   virtual float AStarCostFactor() const override {
-    return speedfactor_[kMaxAssumedSpeed];
+    return speedfactor_[top_speed_];
```
nilsnolde, line 260: `top_speed_` always defaults to `kMaxSpeedKph` (if not explicitly set) in all (but scooter) motorized vehicles. before some cost models used `kMaxAssumedSpeed`, others `kMaxSpeedKph`. not sure if this is a good change..

src/sif/dynamiccost.cc
```diff
@@ -224,6 +225,8 @@ void ParseSharedCostOptions(const rapidjson::Value& value, CostingOptions* pbf_c
     pbf_costing_options->set_name(*name);
   }
   pbf_costing_options->set_shortest(rapidjson::get<bool>(value, "/shortest", false));
+  pbf_costing_options->set_top_speed(
+      kVehicleSpeedRange(rapidjson::get<uint32_t>(value, "/top_speed", kMaxSpeedKph)));
```
nilsnolde, line 229: guard against too high/low values here

valhalla/sif/dynamiccost.h
```diff
@@ -40,6 +40,9 @@ constexpr float kMaxPenalty = 12.0f * midgard::kSecPerHour; // 12 hours
 // since a ferry is sometimes required to complete a route.
 constexpr float kMaxFerryPenalty = 6.0f * midgard::kSecPerHour; // 6 hours
 
+constexpr midgard::ranged_default_t<uint32_t> kVehicleSpeedRange{10, baldr::kMaxSpeedKph,
+                                                                 baldr::kMaxSpeedKph};
+
```
nilsnolde, line 45: does that make sense?

src/sif/autocost.cc
kevinkreiser: its not a good change as it will impact the astar heuristic in such a way as to assume a given node is closer to the goal (because you can travel faster). this will slow down the algorithm for sure. and...

just tested it, its about 17% slower which is defintiely not acceptable. we'll want to revert that change for sure

nilsnolde: my bad.. I might've only "seen" the previous truck implementation, which assumed `kMaxSpeedKph` here. My goal was actually the other way around, to make astar go faster (when a top_speed is actually set) :sweat_smile: Clamping the `top_speed` to `kMaxAssumedSpeed` will probably not be desirable (I guess since there's the `kMaxTrafficSpeed` of 252)?!

nilsnolde: this is what I should've done... what'ya think?
```
  virtual float AStarCostFactor() const override {
    uint32_t s = top_speed_ < kMaxAssumedSpeed ? top_speed_ : kMaxAssumedSpeed;
    return speedfactor_[s];
  }
```

kevinkreiser (changes requested):

src/sif/autocost.cc
```diff
@@ -420,8 +423,9 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
           ? ferry_factor_
           : (edge->use() == Use::kRailFerry) ? rail_ferry_factor_ : density_factor_[edge->density()];
 
+  float speed_penalty = (edge_speed > top_speed_) ? (edge_speed - top_speed_) * 0.05f : 0.0f;
```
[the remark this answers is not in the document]
nilsnolde: > practically speaking i guess we could say we're driving a trabi and so we kind of want to avoid the autobahnen.

exactly. I did clone that from motorscooter, but think that's a very sensible thing to do. mostly I imagine you'd limit the top_speed for this exact kind of situation, tractor, trabi or tank ;) 0.05 _sounds_ reasonable, a penalty of 1 for 20 kmh difference. Should I leave a TODO for now?

src/sif/motorcyclecost.cc
```diff
@@ -81,6 +81,8 @@ constexpr ranged_default_t<float> kUseTollsRange{0, kDefaultUseTolls, 1.0f};
 constexpr ranged_default_t<float> kUseTrailsRange{0, kDefaultUseTrails, 1.0f};
 constexpr ranged_default_t<float> kDestinationOnlyPenaltyRange{0, kDefaultDestinationOnlyPenalty,
                                                                kMaxPenalty};
+constexpr ranged_default_t<float> kTopSpeedRange{5, kMaxSpeedKph,
+                                                 kMaxSpeedKph}; // default to kMaxSpeedKph
```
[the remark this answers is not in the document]
nilsnolde: oops, relict

test/gurka/test_traffic.cc
```diff
@@ -87,7 +87,7 @@ TEST(Traffic, BasicUpdates) {
     auto result = gurka::route(map, "A", "C", "auto", {{"/date_time/type", "0"}}, clean_reader);
     gurka::assert::osrm::expect_steps(result, {"AB"});
     gurka::assert::raw::expect_path(result, {"AB", "BC"});
-    gurka::assert::raw::expect_eta(result, 14.303419);
+    gurka::assert::raw::expect_eta(result, 25.731991);
```
nilsnolde, line 90: So now an edge's speed can never be `kMaxTrafficSpeed` again. Is that gonna be a problem for traffic stuff?

src/sif/autocost.cc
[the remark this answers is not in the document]
kevinkreiser: i think its fine to leave in for now, im just saying that `.05` may need to change. the main reason for that is because the factor in auto costing is in different units than the factor for motorscooter (probably) so the scales might not make sense to be the same. in any case we can tune it later on. no big deal

test/gurka/test_traffic.cc
kevinkreiser: that is a very interesting question! the practical answer is no we likely dont care if a speed is above 87mph though we did used to allow it. i guess it would be nicer to do hard enforcement of the top speed only if it were set by the user and not defaulted but i can't practically think of why we wouldnt want to make a reasonable assumption like you're not going to drive faster than 90mph. i know there some rare places like in UAE where you can do 160kph but this seems like a pretty extreme case. maybe it'd be enough to up the maxassumed to 160kph and call it a day... let me think about that

nilsnolde: As long as absolute maxspeed is not country-specifically implemented (not sure about this, VS Code is balking this morning, can't track code..), I guess that's reasonable to default to the highest maxspeed of any country. With the understanding that you can go up to 252 KPH by setting higher `top_speed`s, e.g. in Germany where there's often (unfortunately) no real maxspeed on motorways, so some freaks might choose to actually go 250 KPH in real life. Then let's make it at least harder for them to plan routes like that ;)

src/sif/truckcost.cc
```diff
@@ -434,18 +434,20 @@ bool TruckCost::AllowedReverse(const baldr::DirectedEdge* edge,
 Cost TruckCost::EdgeCost(const baldr::DirectedEdge* edge,
                          const baldr::GraphTile* tile,
                          const uint32_t seconds) const {
-  auto speed = tile->GetSpeed(edge, flow_mask_, seconds);
+  auto edge_speed = tile->GetSpeed(edge, flow_mask_, seconds);
```
kevinkreiser, line 437: this will automatically give you the truck speed if the edge has it

```suggestion
  auto edge_speed = tile->GetSpeed(edge, flow_mask_, seconds, true);
```

src/sif/truckcost.cc
```diff
@@ -434,18 +434,20 @@ bool TruckCost::AllowedReverse(const baldr::DirectedEdge* edge,
...
 
-  // Use the lower or truck speed (ir present) and speed
-  uint32_t s = (edge->truck_speed() > 0) ? std::min(edge->truck_speed(), speed) : speed;
-  assert(s < speedfactor_.size());
+  // Use the lower of truck speed, top_speed (if present) and speed
+  uint32_t s = (edge->truck_speed() > 0) ? std::min({edge->truck_speed(), edge_speed, top_speed_})
+                                         : std::min(edge_speed, top_speed_);
```
kevinkreiser, line 441: since we got truck speed above we can simplify

```suggestion
  uint32_t s = std::min(edge_speed, top_speed_);
```

test/gurka/test_traffic.cc
kevinkreiser: on line 82, can you use `kMaxAssumedSpeed` in place of `MAX_TRAFFIC_SPEED_KPH`

nilsnolde: Thoughts on raising `kMaxAssumedSpeed` or we leave it be for now?

kevinkreiser: well i tested it with 160 and its a 3-4% performance penalty. so i think we probably shouldnt do that. which leaves the question of whether its ok to clamp traffic speeds down to 140kph in the case that they are higher than that. my gut reaction is yes its ok. @PureW @danpat any opinions on this? if traffic data says the road speed is >140kph should we actually prescribe that people drive that fast. my opinion is no with the caveat that people can opt into it by setting a top_speed = 160 or higher.

kevinkreiser (approved):
We need to delete the empty file `src/loki/edge_search.cc` but otherwise i think this is good to go. I'd like to have one other person say whether they agree or disagree about the clamp to 140kph though

kevinkreiser (approved):

danpaz (approved):
👍  I think it's ok to clamp top speed to 140kph by default.

mloskot:
> turned out to be an easy one: `top_speed` was only set when there was a costing_options object in the request. 

Yes, I have been caught by this too wondering why routing outputs very large leg time in (thousands of) seconds.
It was due to this line:
https://github.com/valhalla/valhalla/blob/3.1.4/src/sif/autocost.cc#L494

and `top_speed_=0` instead of the initialised to the default `kMaxAssumedspeed(140)`.

So, to properly initialise costing with *all* its defaults, here is the **minimal** workaround for C++ API users:

```cpp
valhalla::Api request;
valhalla::ParseApi("{}", valhalla::Options::route, request);  // `none` costing
// or
valhalla::ParseApi("{\"costing\":\"auto\"}", valhalla::Options::route, request);
// then
valhalla::sif::TravelMode mode;
auto costing = valhalla::sif::CostFactory{}.CreateModeCosting(request.options(), mode);
```

and now the `top_speed_=140` for meaningful times :)

PR #2669, opened 2020-10-29
Fix C++17 compatibility

averkhaturau:
Change deprecated std::random_shuffle calls to std::shuffle, use
std::bind instead of deprecated std::bind1st

# Issue

#2669

## Tasklist

## Requirements / Relations

Valhalla wasn't compile for C++17 because of usage of deprecated functions. This PR fixes it.

### DIFF
--- a/src/mjolnir/graphenhancer.cc
+++ b/src/mjolnir/graphenhancer.cc
@@ -666,7 +666,6 @@ bool IsNotThruEdge(GraphReader& reader,
 bool IsIntersectionInternal(const GraphTile* start_tile,
                             GraphReader& reader,
                             std::mutex& lock,
-                            const GraphId& startnode,
                             const NodeInfo& startnodeinfo,
                             const DirectedEdge& directededge,
                             const uint32_t idx) {
@@ -885,8 +884,7 @@ bool IsNextEdgeInternalImpl(const DirectedEdge directededge,
       if (!infer_internal_intersections)
         return diredge->internal();
       else
-        return IsIntersectionInternal(&end_node_tile, reader, lock, directededge.endnode(),
-                                      end_node_info, *diredge, i);
+        return IsIntersectionInternal(&end_node_tile, reader, lock, end_node_info, *diredge, i);
     }
   }
   return false;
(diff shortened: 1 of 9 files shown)


kevinkreiser:
@averkhaturau the project targest c++14 at the moment: https://github.com/valhalla/valhalla/blob/ac97edb83325133ed311c4e2197a887b356a171b/CMakeLists.txt#L19

that said though, if you can find a way to modify the code so that it would compile under c++17 we'll happly merge it. at this time though we arent ready to move to c++17. last i checked we had some older platforms whose compilers dont yet support c++17. i'll ask around to see if c++17 is possible

averkhaturau:
@kevinkreiser , I understand that it's necessary to support C++14, but as I found it's easy to build it with C++17 support with just few changes that I implemented in this PR. Valhalla is 99% ready to work with projects that use C++17, and it's easy to set it with CMake cache. C++14 (perhaps C++11 as well) build is still supported.

kevinkreiser:
@averkhaturau yep no problem i just wanted to be sure you knew we werent planning on migrating to c++17 just yet. but yeah as i said, we'll be happy to merge these preparations for migrating to c++17. thank you very much for your contribution!

kevinkreiser:
This looks good to me but is failing unrelated clang tidy issues. I'll take a pass at cleaning those up to make the build pass. Thanks again!

kevinkreiser (approved):

src/mjolnir/graphenhancer.cc
```diff
@@ -1891,7 +1888,8 @@ void GraphEnhancer::Enhance(const boost::property_tree::ptree& pt,
   for (const auto& tile_id : local_tiles) {
     tempqueue.emplace_back(tile_id);
   }
-  std::random_shuffle(tempqueue.begin(), tempqueue.end());
+  std::random_device rd;
+  std::shuffle(tempqueue.begin(), tempqueue.end(), std::mt19937(rd()));
```
kevinkreiser, line 1892: just noting that this is a change in behavior. from the docs: 

> 1) The random number generator is implementation-defined, but the function std::rand is often used.

PR #2670, opened 2020-10-29
fix reach counting of transition nodes

kevinkreiser:
There were two amateur level bugs (very likely my own) in loki.

1. when edge in spatial bin is filtered but opposing edge is not, we flip to creating a search candidate out of the opposing edge. we forgot to flip the edge_id to match the opposing edge. its unclear what impact this has had in the past (wrong edge with wrong id sent to the finalize method in loki::search). its possible there is no effect as the finalize method looks up both directions of an edge when making the final result. i'll review the code and run RAD
2. the major bug here.. in the naive reach calculation, it is possible expand from the same node multiple times so long as it was evaluated as a transition node. the code that enqueued those nodes didnt check if the node had alraedy been expanded. so you could visit the node and then if another edge saw the node from a transition it would get enqueued again. with the right graph and some bad luck this can make edges look reachable when they are not

### DIFF
--- a/src/loki/reach.cc
+++ b/src/loki/reach.cc
@@ -29,10 +29,15 @@ void Reach::enqueue(const baldr::GraphId& node_id,
   // otherwise we enqueue it
   queue_.insert(node_id);
   // and we enqueue it on the other levels
-  for (const auto& transition : tile->GetNodeTransitions(node))
+  for (const auto& transition : tile->GetNodeTransitions(node)) {
+    // skip nodes which are done already
+    if (done_.find(transition.endnode()) != done_.cend())
+      continue;
+    // otherwise we enqueue it
     queue_.insert(transition.endnode());
-  // and we remember how many duplicates we enqueued
-  transitions_ += node->transition_count();
+    // and we remember how many duplicates we enqueued
+    ++transitions_;
+  }
 }
 
 directed_reach Reach::operator()(const DirectedEdge* edge,
--- a/src/loki/reach.cc
+++ b/src/loki/reach.cc
@@ -193,11 +193,11 @@ directed_reach Reach::exact(const valhalla::baldr::DirectedEdge* edge,
 }
 
 // callback fired when a node is expanded from, the node will be the end node of the previous label
-void Reach::ExpandingNode(baldr::GraphReader& graphreader,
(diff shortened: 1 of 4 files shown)


src/loki/reach.cc
```diff
@@ -29,10 +29,15 @@ void Reach::enqueue(const baldr::GraphId& node_id,
   // otherwise we enqueue it
   queue_.insert(node_id);
   // and we enqueue it on the other levels
-  for (const auto& transition : tile->GetNodeTransitions(node))
+  for (const auto& transition : tile->GetNodeTransitions(node)) {
+    // skip nodes which are done already
+    if (done_.find(transition.endnode()) != done_.cend())
```
kevinkreiser, line 34: if the transition node was already seen and expanded we dont want to do it again or else that will over count

src/loki/reach.cc
```diff
@@ -29,10 +29,15 @@ void Reach::enqueue(const baldr::GraphId& node_id,
...
+      continue;
+    // otherwise we enqueue it
     queue_.insert(transition.endnode());
-  // and we remember how many duplicates we enqueued
-  transitions_ += node->transition_count();
```
kevinkreiser, line 35: since we now dont enqueue all the transition nodes we have to count them individually

src/loki/search.cc
```diff
@@ -585,6 +585,7 @@ struct bin_handler_t {
               opp_reach.inbound >= p_itr->location.min_inbound_reach_) {
             tile = opp_tile;
             edge = opp_edge;
+            edge_id = opp_edgeid;
```
kevinkreiser, line 588: ill review the code after this (finalize function) but this was also not right. when we bail on the main edge and swap in the opposing edge we were forgetting to swap in the id

kevinkreiser:
RAD showed only one diff where a route went from failing to now having a route!

src/loki/search.cc
kevinkreiser: yeah so the only thing the id gets used for in the finalize method is to look up the reach, which get set on the search result but isnt otherwised used later in the request processing. so this fix makes it so that you dont see the wrong reach value in locate for example but otherwise has no impact that i can see

danpaz (dismissed):

danpaz (approved):

PR #2675, opened 2020-11-03
Move turn_lane_direction helper to odin/util

danpaz:
# Issue

Small change to move the `turn_lane_direction` helper to odin/util.cc. 

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/odin/util.cc
+++ b/src/odin/util.cc
@@ -10,6 +10,7 @@
 #include <date/date.h>
 #include <date/tz.h>
 
+#include "baldr/turnlanes.h"
 #include "locales.h"
 #include "midgard/logging.h"
 #include "odin/util.h"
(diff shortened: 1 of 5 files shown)


src/odin/util.cc
```diff
@@ -134,5 +135,29 @@ const std::unordered_map<std::string, std::string>& get_locales_json() {
   return locales_json;
 }
 
+std::string turn_lane_direction(uint16_t turn_lane) {
+  switch (turn_lane) {
+    case baldr::kTurnLaneReverse:
+      return "uturn";
```
mandeepsandhu, line 141: The turn direction strings used here are repeated in 4 different places in the route serializer. Wonder if we can create constants that can be reused in multiple places?

If its trivial to do, I'd suggest doing it in this PR, else we can open a follow-up task to consolidate the various string constants used in `odin` and other modules.

danpaz: @mandeepsandhu I added a new header in https://github.com/valhalla/valhalla/pull/2675/commits/7485b30c5346f0654dae382de76450e5d1b32c58 what do you think of this approach?

mandeepsandhu (approved):
Thanks for the cleanup! 🙏

PR #2678, opened 2020-11-05
Edge length check

gknisely:
# Issue

If the length of the edge is greater than kMaxEdgeLength, then consider this a catastrophic error.  Need to do the check in the builder as transit edges could be large; however, they should not be for the graph.

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/mjolnir/bssbuilder.cc
```diff
@@ -197,7 +197,16 @@ DirectedEdge make_directed_edge(const GraphId endnode,
...
   directededge.set_endnode(endnode);
-  directededge.set_length(std::max(1.0f, valhalla::midgard::length(shape)));
+  float length = valhalla::midgard::length(shape);
+
+  if (length > kMaxEdgeLength) {
+    // Consider this a catastrophic error.  Need to do the check here as
+    // transit edges could be large; however, they should not be for the graph.
+    LOG_ERROR("Exceeding max. edge length: " + std::to_string(length));
+    throw std::runtime_error("DirectedEdgeBuilder: exceeded maximum edge length");
+  }
```
kevinkreiser, line 207: can we move this `if` block into `directededge::set_length` and replace the warning that is in there? then it is always going to happen in one place and we wont have to duplicate it whenever we make edges in some part of the code.

gknisely: @kevinkreiser the reason I didn't do this was because of transit and that we may have large edges.  Unless you think the max is too large even for transit.

test/graphbuilder.cc
```diff
@@ -89,6 +90,40 @@ TEST(graphbuilder, TestConstructEdgesSubset) {
   EXPECT_TRUE(reader.DoesTileExist(GraphId{5993698}));
 }
 
+TEST(Graphbuilder, TestDEBuilderLength) {
+
+  std::vector<PointLL> shape1{{-160.096619f, 21.997619f},
+                              {-90.037697f, 41.004531},
+                              {-160.096619f, 21.997619f}};
+  try {
+    DirectedEdgeBuilder edge_builder({}, GraphId(123, 2, 8), true, valhalla::midgard::length(shape1),
```
mandeepsandhu, line 99: You can use `ASSERT_NO_THROW`/`EXCEPT_NO_THROW` to check for exceptions

test/graphbuilder.cc
```diff
@@ -89,6 +90,40 @@ TEST(graphbuilder, TestConstructEdgesSubset) {
...
+    throw std::runtime_error(
+        "TestDEBuilderLength: DirectedEdgeBuilder: Should not have thrown an exception for the length");
+  }
+
+  std::vector<PointLL> shape2{{-160.096619f, 21.997619f},
+                              {-90.037697f, 41.004531},
+                              {-160.096619f, 21.997619f},
+                              {-90.037697f, 41.004531}};
+  try {
+    DirectedEdgeBuilder edge_builder({}, GraphId(123, 2, 8), true, valhalla::midgard::length(shape2),
```
mandeepsandhu, line 114: Consider `ASSERT_THROW`/`EXCEPT_THROW`

gknisely: sure.  will do

src/mjolnir/bssbuilder.cc
gknisely: updated to use bool in set_length

kevinkreiser (dismissed):

test/graphbuilder.cc
dnesbitt61: hmmm...won't assert stop execution? Since this is during data building we can perhaps expect times when a new OSM way / edge is added that might exceed this (e.g. long ferries). I was reading that asserts should be used when you know some condition that should never occur - here is a condition that could occur and I think should be an exception.

mandeepsandhu (approved):

PR #2685, opened 2020-11-17
Add support for ignoring live traffic closures for waypoints

mandeepsandhu:
Adds support for ignoring road closures due to traffic for location
waypoints, by way of a new search_filter option - "exclude_closures"

Roads which have live traffic speed of 0, are considered closed. If
search_filter.exclude_closures is set to false for a location, then any
outgoing edges that are marked closed, will be considered when loki
searches for start/end candidates.


valhalla/sif/dynamiccost.h
```diff
@@ -669,6 +669,16 @@ class DynamicCost {
 
   virtual Cost BSSCost() const;
 
+  /*
+   * Determine whether an edge is filtered due to traffic closure.
+   * @param  edgeid         GraphId of the opposing edge.
+   * @return  Returns true if the edge is filtered due to live traffic closure, false if not.
+   */
+  inline virtual bool FilterClosed(const baldr::DirectedEdge* edge,
```
kevinkreiser, line 677: im not a fan of making this public especially bcause it has some hidden state in there making use of `filter_closures_`

what if we make a generic public function like this:

```c++
inline virtual bool IsClosed(const baldr::DirectedEdge* edge,
                                   const baldr::GraphTile* tile) const {
￼    return (flow_mask_ & baldr::kCurrentFlowMask) && tile->IsClosed(edge);
￼  }
```
and inside of the `Filter` function we call it with the boolean `filter_closures_` and inside of the allowed function we call it with the boolean `ignore_closures`.

src/worker.cc
```diff
@@ -526,6 +536,16 @@ void parse_locations(const rapidjson::Document& doc,
     if (!had_date_time) {
       add_date_to_locations(options, *locations);
     }
+
+    // If any of the locations had search_filter.exclude_closures set to false,
+    // set the corresponding internal costing attribute, so that loki search
+    // does not drop a start/end candidate when an edge is marked as closed due
+    // to live traffic
```
kevinkreiser, line 543: NIT:

```suggestion
    // we tell the costing to let all closed roads through so that we can do
    // a secondary per-location filtering using lokis search_filter functionality
```

valhalla/sif/dynamiccost.h
mandeepsandhu: I don't see the `Allowed` function using `ignore_closures` currently. Looks like its used during a-star search? Not sure of its impact if we move the `ignore_closures` check there.

Even with that, the basic issue IMO, still stands i.e during search (specifically in `is_search_filter_triggered`), how do we know the state of `ignore_closures`? `filter_closures` is already appropriately set taking `ignore_closures` into account, so I thought it would be good to expose it indirectly via the `Filter` function. I don't think doing so exposes an internal state, but rather rather exposes an API that just uses it.

Not sure how making `IsClosed` public helps here, since we still need to access `filter_closures_` during search. Maybe I'm missing something.

mandeepsandhu: Actually, I don't need to make `FilterClosed` (or `IsClosed`) public. I guess I can call `costing.Filter(...)` inside `is_search_filter_triggered`:
```c++
if ((road_class > min_road_class || road_class < max_road_class) ||
     (search_filter.exclude_tunnel_ && edge->tunnel()) ||
     (search_filter.exclude_bridge_ && edge->bridge()) ||
     (search_filter.exclude_ramp_ && (edge->use() == Use::kRamp)) ||
     (search_filter.exclude_closures_ && costing.Filter(edge, tile))) {
      return true;
```

Here's how I think it will work:

1. `ignore_closures: false`, `exclude_closures: true` (both defaults): `Costing::Filter` filters candidates due to closures Search filter is never called for candidates that are already removed by `Costing::Filter`.
2. `ignore_closures: true`, `exclude_closures: true or false`: This sets `filter_closures_` to `false` and `Costing::Filter` does not filter any candidates due to closures. search filter is also not triggered, since `Costing::Filter` always returns `0.0/false`
3. `ignore_closures: false`, `exclude_closures: false`: This sets `filter_closures_` to `false` and `Costing::Filter` does not filter any candidates due to closures. search filter is also not triggered, since `exclude_closures` is `false`.

CHANGELOG.md
```diff
@@ -196,7 +196,8 @@
    * ADDED: Add annotations to osrm response including speed limits, unit and sign conventions [#2668](https://github.com/valhalla/valhalla/pull/2668)
    * ADDED: Added functions for predicted speeds encoding-decoding [#2674](https://github.com/valhalla/valhalla/pull/2674)
    * ADDED: Time invariant routing via the bidirectional algorithm. This has the effect that when time dependent routes (arrive_by and depart_at) fall back to bidirectional due to length restrictions they will actually use the correct time of day for one of the search directions [#2660](https://github.com/valhalla/valhalla/pull/2660)
-   * ADDED: If the length of the edge is greater than kMaxEdgeLength, then consider this a catastrophic error if the should_error bool is true in the set_length function. [2678](https://github.com/valhalla/valhalla/pull/2678)
+   * ADDED: If the length of the edge is greater than kMaxEdgeLength, then consider this a catastrophic error if the should_error bool is true in the set_length function. [#2678](https://github.com/valhalla/valhalla/pull/2678)a
```
danpaz, line 199:
```suggestion
   * ADDED: If the length of the edge is greater than kMaxEdgeLength, then consider this a catastrophic error if the should_error bool is true in the set_length function. [#2678](https://github.com/valhalla/valhalla/pull/2678)
```

src/worker.cc
```diff
@@ -825,7 +845,6 @@ void from_json(rapidjson::Document& doc, Options& options) {
 
   // Parse all of the costing options in their specified order
   sif::ParseCostingOptions(doc, "/costing_options", options);
-  options.set_costing(costing);
```
mandeepsandhu, line 828: Removed, since its being set [above](https://github.com/valhalla/valhalla/pull/2685/files#diff-dda1eacb6be34d7946468f09844f48890c1266509f634990ec27263271bd6f48R841)

src/loki/search.cc
```diff
@@ -36,7 +38,8 @@ bool is_search_filter_triggered(const DirectedEdge* edge,
   if ((road_class > min_road_class || road_class < max_road_class) ||
       (search_filter.exclude_tunnel_ && edge->tunnel()) ||
       (search_filter.exclude_bridge_ && edge->bridge()) ||
-      (search_filter.exclude_ramp_ && (edge->use() == Use::kRamp))) {
+      (search_filter.exclude_ramp_ && (edge->use() == Use::kRamp)) ||
+      (search_filter.exclude_closures_ && !costing.Filter(edge, tile))) {
```
kevinkreiser, line 42: it doesnt make sense for us to rely on the meaning of the costing filter here. for starters, this code path could never be hit if `!costing.Filter` is `true` because it means we would have filtered the edge out in the first place and never checked the search filters for it.

I suggest we just check if you want to exclude closures and if you are a costing that cares about current flow and current flow is 0. ie this:

```suggestion
      (search_filter.exclude_closures_ && (costing.flow_mask() & baldr::kCurrentFlowMask) && tile->IsClosed(edge))) {
```

mandeepsandhu: >for starters, this code path could never be hit if !costing.Filter is true because it means we would have filtered the edge out in the first place and never checked the search filters for it

You're right, in that `is_search_filter_triggered` is never called, **IF** `exclude_closures` is `true`. However, if `exclude_closure` is `false`, we internally make `filter_closures: true` which results in `!costing.Filter` returning `false` and we end up invoking the search filter.

Basically  there is never a case where we want `Costing::Filter()` to pass a candidate which might be filtered due to closures,  but then have `is_search_filter_triggered` filter it out. 

The change you suggested won't work if `ignore_closures: true` and `exclude_closures: true`, because the candidates will pass `Costing::Filter` but then fail in `is_search_filter_triggered`. So either we access the state of `ignore_closures` here, or use `DynmaicCost::IsClosed` by making it public (as that already takes `ignore_closures` into account).

I'm leaning towards using the `ignore_closures` value from the costing options pbf. Something like:
```
search_filter.exclude_closures = !costing_options.ignore_closures && location_options.exclude_closures
```
and then use it the way you suggested in `is_search_filter_triggered`. What do you think?

kevinkreiser: i guess im just saying that doing:

```c++
search_filter.exclude_closures_ && !costing.Filter(edge, tile)
```

this code says that if i want to exclude closures and the edge wasnt filtered before filter it now. it not being filtered before is not the same as saying it currently has a closure. there are many reasons an edge can be filtered, closures are not the only ones.

let me think about this a bit more. there simply has to be a way to make this less confusing :smile:

mandeepsandhu: > i guess im just saying that it makes no sense to do:
> 
> ```c++
> search_filter.exclude_closures_ && !costing.Filter(edge, tile)
> ```
> 
> this code says that if i want to exclude closures and the edge wasnt filtered before filter it now. it not being filtered before is not the same as saying it currently has a closure. there are many reasons an edge can be filtered, a closure is not the only reason.
> 

Understood. We don't have to use costing filter here. Thats why I suggested accessing `ignore_closures` from the costing options here (or make `DynmaicCost::IsClosed` public). We'll need it to respect its override behavior.

bench/thor/routes.cc
```diff
@@ -81,7 +82,8 @@ boost::property_tree::ptree build_config(const char* live_traffic_tar) {
 
 constexpr float kMaxRange = 256;
 
-static void BM_UtrechtBidirectionalAstar(benchmark::State& state) {
+template <class ALGO>
```
mandeepsandhu, line 85: Updated benchmark test to be able to run multiple algo's for the same test.

kevinkreiser:
@mandeepsandhu do you think we should consider an alternate implementation where we treat closures at the beginning and ending of routes similar to not_through regions. ie we can route to get out of them but we dont go in them in the middle of the route? i worry that a user getting a different edge candidate simply because they snapped to a multi edge closure would be unexpected/desired. was breifly discussing this with @danpaz

docs/api/turn-by-turn/api-reference.md
```diff
@@ -106,6 +106,7 @@ These options are available for `auto`, `auto_shorter`, `bus`, and `truck` costi
 | `country_crossing_cost` | A cost applied when encountering an international border. This cost is added to the estimated and elapsed times. The default cost is 600 seconds. |
 | `country_crossing_penalty` | A penalty applied for a country crossing. This penalty can be used to create paths that avoid spanning country boundaries. The default penalty is 0. |
 | `shortest` | Changes the metric to quasi-shortest, i.e. purely distance-based costing. Note, this will disable all other costings & penalties. Also note, `shortest` will not disable hierarchy pruning, leading to potentially sub-optimal routes for some costing models. The default is `false`. |
+| `ignore_closures` | If set to `true`, ignores all closures, marked due to live traffic closures, during routing. **Note:** This option cannot be set if `location.search_filter.exclude_closures` is also specified |
```
danpaz, line 109: The note here should explain also what happens if both options are specified (looking at the code in worker.cc it seems an exception is thrown).

src/worker.cc
```diff
@@ -179,6 +180,8 @@ const std::unordered_map<unsigned, std::string> OSRM_ERRORS_CODES{
      R"({"code":"InvalidValue","message":"The successfully parsed query parameters are invalid."})"},
     {142,
      R"({"code":"InvalidValue","message":"The successfully parsed query parameters are invalid."})"},
+    {143,
+     R"({"code":"InvalidValue","message":"The successfully parsed query parameters are invalid."})"},
```
danpaz, line 184: Can this error message specify what was invalid about the query?

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,388 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+      close_single_edge(reader, tile, index, current, "DE", closure_map);
+    };
+    test::customize_live_traffic_data(closure_map.config, close_edge);
+
+    auto result = gurka::route(closure_map, {"1", "2"}, costing, {{"/date_time/type", "0"}}, reader);
+    gurka::assert::osrm::expect_steps(result, {"HIC", "CFGD"});
+    gurka::assert::raw::expect_path(result, {"HIC", "CFGD"});
+
+    // Specify search filter to disable exclude_closures at departure. Due to
+    // consecutive closures, this should result in a "no route"
```
danpaz, line 483: I think this test captures the multi-edge closure case, which we do want to return a valid route.

danpaz: For completeness we should also try a test that has closures not just at the origin and destination but also on edge in between, to ensure we allow closures at the endpoints but still avoid them along the route.

mandeepsandhu: > For completeness we should also try a test that has closures not just at the origin and destination but also on edge in between, to ensure we allow closures at the endpoints but still avoid them along the route.

Intermediate closure scenario is covered in `IgnoreClosuresOverridesExcludeClosures` testcase. We route from `1 -> 2` and close edge `CD`. Without `ignore_closures` costing option we don't route through CD which is an intermediate edge.

Would you rather see this a separate standalone test (and named appropriately)?

danpaz: I was thinking of adding a new test case that closes the origin edge, destination edge, _and_ an intermediate edge, to exercise the case where a user can snap to a closure but avoid closures in between the snapped edges. Maybe this logic path is covered already though?

mandeepsandhu: > I think this test captures the multi-edge closure case, which we do want to return a valid route.

That would depend on whether there are suitable edges nearby, no? If loki's search results in only 1 candidate edge, then I guess if there are no reachable edges to it it would fail to find a route? CMIIW.

danpaz: But in this test the request has `exclude_closures=false`, so we know that there are suitable edges to snap to. Why then does it return a NoRoute?

kevinkreiser:
i think to get this to a point where we can ship id like to offer up the following idea. we add 2 test cases where we set filter_closures:false, they should look something like:

1. where there is a multiple edge closure but we set reachability so that the edge in the closure is not reachable but an edge outside of it is, creating a long snap anyway and routing around
image

2. where there is a multipe edge closure but we set a heading on the origin. the route will then hit the second edge and have to do a uturn to get around the closure
image

Once we can confirm that is the case and those test cases are failing, we can disable them and get thsi shipped with the idea that we will do smoe more work. specifically i think we should make starting and ending in a closure work like starting/ending in a not_thru area, where you are allowed to route inside of it until you leave it. as long as your predicessory label is marked that way you can keep expanding in there but as soon as the predecessor is not marked that way you can no longer expand edges like that. this will require us to know whether or not an edge is closed outside of the allowed method though which will be quite annoying...

src/worker.cc
mandeepsandhu: > Once we can confirm that is the case and those test cases are failing, we can disable them and get thsi shipped with the idea that we will do smoe more work. specifically i think we should make starting and ending in a closure work like starting/ending in a not_thru area, where you are allowed to route inside of it until you leave it. as long as your predicessory label is marked that way you can keep expanding in there but as soon as the predecessor is not marked that way you can no longer expand edges like that. this will require us to know whether or not an edge is closed outside of the allowed method though which will be quite annoying...

I get how continuing to expand on consecutive closures would work when doing bidir a-star (since we start expansion from both ends). But what would the behavior be for  the unidir a-star variants if there are multiple consecutive closures leading to the destination? When expanding from an intermediate open edge, we would avoid the closure, even though it might lead to the destination. If so, would it fail to find the route (what would happen if this was a not-thru edge instead)?

kevinkreiser: i cant remember if the forward/reverse only implement handle this, i have to check the code to refresh my memory on that

docs/api/turn-by-turn/api-reference.md
mandeepsandhu: The exception will get translated to an internal error code (`143` in this case), and that further translates to a [HTTP status 400 later](https://github.com/valhalla/valhalla/pull/2685/files#diff-dda1eacb6be34d7946468f09844f48890c1266509f634990ec27263271bd6f48R99).

Looking at the docs, it seems that all error codes are explained later in a separate section rather inline with options that can possibly create those errors.

docs/api/turn-by-turn/api-reference.md
```diff
@@ -494,6 +495,7 @@ The codes correspond to code returned from a particular [Valhalla project](https
 |140 | Action does not support multimodal costing |
 |141 | Arrive by for multimodal not implemented yet |
 |142 | Arrive by not implemented for isochrones |
+|143 | ignore_closure in costing and exclude_closure in search_filter cannot both be specified |
```
mandeepsandhu, line 498: This is the new exception type added which explains the error related to the new option.

src/worker.cc
mandeepsandhu: > Can this error message specify what was invalid about the query?

The specific error message is described in the [new internal error code](https://github.com/valhalla/valhalla/pull/2685/files#diff-7ab49622366d941710bed95da599b77e3bd3209104867f93e169fb9843dd4a8fR50). I think the error messages here are kept to match OSRM, which has this message for most error codes (@kevinkreiser CMIIW).

mandeepsandhu:
@kevinkreiser @danpaz I've added more tests for the following scenarios:
* Closures on trivial routes (src/dst on same edge & adjacent edges which would fallback to using uni-directional a*)
* Added tests with consecutive closures that can cause possible u-turns & distant edge snaps. Both tests are currently disabled.
* Added ability to test with different costing models & 2 date_time type (0, 3). date type 0 is currently disabled since that uses timedep-fwd astar which does not work for closures at destinations.

CC: @danpat

src/worker.cc
```diff
@@ -685,6 +720,43 @@ void from_json(rapidjson::Document& doc, Options& options) {
     options.set_linear_references(*linear_references);
   }
 
+  // costing defaults to none which is only valid for locate
```
mandeepsandhu, line 723: This code block was moved up before parsing locations so that we have the costing options parsed before hand. This helps us access `ignore_closures` option later.

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,840 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+  static void SetUpTestSuite() {
+    const std::string ascii_map = R"(
+              F----G
+              |    |
+    A--1--B---C----D-2-E--J--3--K
+             /
+        H---I
+
+
+   L-4---5-M-6-N
```
mandeepsandhu, line 235: Used for testing trivial cases which fallback to using timedep fwd

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,840 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+  // DE edge is closed in both directions. Route should avoid DE with
+  // exclude_closures set to true (default) & and use it otherwise
+  {
+    LiveTrafficCustomize close_edge = [](baldr::GraphReader& reader, baldr::TrafficTile& tile,
+                                         int index, baldr::TrafficSpeed* current) -> void {
+      close_single_edge(reader, tile, index, current, "JK", closure_map);
+    };
+    test::customize_live_traffic_data(closure_map.config, close_edge);
+
+    // TODO: 1->2 fails to get route without ignoring closure at 2. why?
```
mandeepsandhu, line 362: I found it odd that if I close edge `DE` and route to `2`, its unable to find the route. However the same does not happen when I close `JK` & route to `3`. I haven't found a good reason for this difference in behavior yet.

CC: @kevinkreiser @danpat @danpaz

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,840 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+    // params.emplace_back(std::make_tuple(costing, "0"));
+  }
+  return params;
+}
+
+INSTANTIATE_TEST_SUITE_P(SearchFilter,
+                         ExcludeClosuresOnWaypoints,
+                         ::testing::ValuesIn(buildParams()));
+
+class DISABLED_ExcludeConsecutiveEdgeClosures
```
mandeepsandhu, line 845: Opened https://github.com/valhalla/valhalla/issues/2732 to handle consecutive closures being tested here (currently disabled).

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,840 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+        (boost::format(
+             R"({"locations":[{"lat":%s,"lon":%s,"heading":90,"search_filter":{"exclude_closures":false}},{"lat":%s,"lon":%s,"search_filter":{"exclude_closures":false}}],"costing":"%s", "costing_options": {"%s": {"speed_types":["freeflow","constrained","predicted","current"]}}, "date_time":{"type":"%s", "value": "current"}})") %
+         std::to_string(closure_map.nodes.at("1").lat()) %
+         std::to_string(closure_map.nodes.at("1").lng()) %
+         std::to_string(closure_map.nodes.at("E").lat()) %
+         std::to_string(closure_map.nodes.at("E").lng()) % costing % costing % date_type)
+            .str();
+    result = gurka::route(closure_map, req_disable_exclude_closures, reader);
+    gurka::assert::osrm::expect_steps(result, {"CD", "AC", "AB", "BE"});
+    gurka::assert::raw::expect_path(result, {"CD", "CD", "AC", "AB", "BE"});
```
mandeepsandhu, line 965: This test currently fails if expecting a uturn on `CD` (the actual path returned is `"CD", "AC", "AB", "BE"`) even with a heading of 90º (going towards `D`).

Seems like the uturn wouldn't occur with left-side driving?

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,840 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+    const std::string& req_disable_exclude_closures =
+        (boost::format(
+             R"({"locations":[{"lat":%s,"lon":%s,"heading":0,"search_filter":{"exclude_closures":false}},{"lat":%s,"lon":%s,"search_filter":{"exclude_closures":false}}],"costing":"%s", "costing_options": {"%s": {"speed_types":["freeflow","constrained","predicted","current"]}}, "date_time":{"type":"%s", "value": "current"}})") %
+         std::to_string(closure_map.nodes.at("M").lat()) %
+         std::to_string(closure_map.nodes.at("M").lng()) %
+         std::to_string(closure_map.nodes.at("B").lat()) %
+         std::to_string(closure_map.nodes.at("B").lng()) % costing % costing % date_type)
+            .str();
+    auto result = gurka::route(closure_map, req_disable_exclude_closures, reader);
+    gurka::assert::osrm::expect_steps(result, {"HI", "FI", "EF", "BE"});
```
mandeepsandhu, line 1017: This test too fails with a no-route (both with `exclude_closure:true/false` at the start location). I've tried `minimum_reachability` of 0 & 1 but that didn't seem to affect anything.

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,840 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+std::vector<costing_and_datetype> buildParams() {
+  std::vector<costing_and_datetype> params;
+
+  std::vector<std::string> costings = {"auto",  "motorcycle", "motor_scooter", "bus",
+                                       "truck", "hov",        "taxi"};
+  for (const auto& costing : costings) {
+    // Add date_type:3 for time-invariant bidir a*
+    params.emplace_back(std::make_tuple(costing, "3"));
+    // Add date_type:0 for timedep-fwd a* using current time
+    // TODO: Currently, ignoring closures at destinations does not work for
```
mandeepsandhu, line 834: Opened https://github.com/valhalla/valhalla/issues/2733 for handling closures in unidirectional astar variants.

mandeepsandhu:
codecov is likely failing due to the disabled tests.

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,841 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+                                {"/date_time/value", "current"},
+                                {costing_speed_type, "current"}},
+                               reader);
+    gurka::assert::osrm::expect_steps(result, {"HIC", "CD"});
+    gurka::assert::raw::expect_path(result, {"HIC", "CD", "DE"});
+
+    // Specify search filter to disable exclude_closures at departure
+    const std::string& req_disable_exclude_closures =
+        (boost::format(
+             R"({"locations":[{"lat":%s,"lon":%s,"search_filter":{"exclude_closures":false}},{"lat":%s,"lon":%s}],"costing":"%s", "costing_options": {"%s": {"speed_types":["freeflow","constrained","predicted","current"]}}, "date_time":{"type":"%s", "value": "current"}})") %
```
danpaz, line 322: I think you can use gurka::route without pre-generating the request string like you do above

```
auto result = gurka::route(closure_map, {"1", "2"}, costing,
                               {{"/locations/0/search_filter/exclude_closures", false}},
                               {{"/date_time/type", date_type},
                                {"/date_time/value", "current"},
                                {costing_speed_type, "current"}},
                               reader);
```

test/gurka/test_search_filter.cc
danpaz: @mandeepsandhu I wonder if it's node snapping since DE is shorter than JK. Can you try increasing the gridsize from 10 to 100 and see if the same happens?

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,841 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+                                {"/date_time/value", "current"},
+                                {costing_speed_type, "current"}},
+                               reader);
+    gurka::assert::osrm::expect_steps(result, {"AB"});
+    gurka::assert::raw::expect_path(result, {"AB", "BC", "CD", "DE", "EJ"});
+
+    // Specify search filter to disable exclude_closures at destination
+    const std::string& req_disable_exclude_closures =
+        (boost::format(
+             R"({"locations":[{"lat":%s,"lon":%s},{"lat":%s,"lon":%s,"search_filter":{"exclude_closures":false}}],"costing":"%s", "costing_options": {"%s": {"speed_types":["freeflow","constrained","predicted","current"]}}, "date_time":{"type":"%s", "value": "current"}})") %
```
danpaz, line 374: Same as above, use cleaner gurka::route with options

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,841 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+
+  // None of the edges are closed. Route has multiple waypoints
+  {
+    auto result = gurka::route(closure_map, {"1", "2", "3"}, costing,
+                               {{"/date_time/type", date_type},
+                                {"/date_time/value", "current"},
+                                {costing_speed_type, "current"}},
+                               reader);
+    gurka::assert::osrm::expect_steps(result, {"AB", "DE"});
+    // Note: Due to the way 2 is placed, it causes a uturn at E
```
danpaz, line 403: Are you sure it's doing a u-turn? I don't think that's expected since it's just a straight route from 1->2->3
```
    A--1--B---C----D-2-E--J--3--K
```

and this path does look correct for a straight route with two legs

```
leg    1     1     1     1     2     2     2    
path {"AB", "BC", "CD", "DE", "DE", "EJ", "JK"}
```

test/gurka/test_search_filter.cc
```diff
@@ -182,3 +185,841 @@ TEST_F(SearchFilter, ExcludeRamp) {
...
+  }
+
+  {
+    LiveTrafficCustomize close_edge = [](baldr::GraphReader& reader, baldr::TrafficTile& tile,
+                                         int index, baldr::TrafficSpeed* current) -> void {
+      close_single_edge(reader, tile, index, current, "DE", closure_map);
+    };
+    test::customize_live_traffic_data(closure_map.config, close_edge);
+
+    // TODO: 1->2 fails to get route. Why?
```
danpaz, line 414: Maybe what's happening is 2 snaps to E and therefore there's no route possible from 1->2?

```
    A--1--B---C----D-2-E--J--3--K
```

test/gurka/test_search_filter.cc
danpaz: 👍 can you add a comment here linking to #2732?

danpaz (commented):
Few comments in the gurka tests to make sure I understand what's happening there. Otherwise looking good!

test/gurka/test_search_filter.cc
mandeepsandhu: Changing the gridsize to 100 does stop the no route exception. When you say "node snapping", do you mean its snapping to `E`? If so, shouldn't it still route upto `D`?

mandeepsandhu: Actually I also had to increase DE's length, otherwise its still unable to find the route

test/gurka/test_search_filter.cc
mandeepsandhu: Doh! 🤦 yes thats just the end of first leg & beginning of second leg...forgot 2 was a waypoint! Need coffee!

test/gurka/test_search_filter.cc
mandeepsandhu: Possibly. But its still failing after increasing length of `DE` and changing grid size to 100. Confusingly, the previous test case - `ExcludeClosuresAtDestination` no longer throws exception when routing between only 2 waypoints.

test/gurka/test_search_filter.cc
mandeepsandhu: Unfortunately this won't work because, in the `route()` overload accepts `unordered_map <string, string>`, which means that I have to pass-in `false` as `"false"`...and this causes this option to be a string in the resulting json (when what we want is a json boolean type).

This could possibly be fixed by passing in the `rapidjson::Value` variant (which can accommodate different types). But I'll have to test it out and see. For now, I'll keep it as-is.

test/gurka/test_search_filter.cc
mandeepsandhu: Cannot use as explained in previous comment.

test/gurka/test_search_filter.cc
danpaz: I think we should debug this -- start by dumping out the geojson of the map using `dump_geojson(map)` and the input/snapped locations and map those in geojson.io.

mandeepsandhu: @danpaz so this route wasn't failing from `1->2`, but rather from `2->3` (which is expected since there are no suitable edges that can be used to route out of `2`). I'll update the comments to reflect this.

test/gurka/test_search_filter.cc
kevinkreiser: ok yeah this is because the reach check doesnt do what we think it should do. the reach check has two phases, conservative fast mode, and hardcore full checking mode which it falls back to when the conservative estimate is a lower reach than we'd like.

i cant explain exclude_closure:true, but when its false, what happens is that conserative estimate keeps excluding closures whiel its doing the reach check and thinks that eveyrthing is reachable. to fix this we will need to add another check to the conservative check which tests whether the edge is currently closed. i can make this change and do a benchmark to see what that does to performance. the sad thing is it allready checks it in the filter but there is no clean way to find out about it (other than making the filter funciton take more options)

mandeepsandhu:
Performance impact of change:

```shell
branch:         |            50th              |             90th             |         99th
master          | 0.03990912437438965          | 0.07622202237447102          | 0.19320416450500488
this branch     | 0.04096802075703939 (+2.65%) | 0.07817689577738444 (+2.56%) | 0.19673490524291992 (+1.83%)
```
**Note:** These are average of 3 runs for each branch with 20k route requests (pre-warmed tile-cache by running 20k requests before measuring performance of each branch).

Also, for another set of test requests (`de_benchmark_routes.txt`) I got an _improvement_ in the 90th percentile of about 0.3%. But this seems t be within the margin of measurement noise. Although the 2.5% needs to be looked at. Thoughts @kevinkreiser ?

bench/thor/routes.cc
```diff
@@ -11,6 +11,7 @@
 #include "sif/costfactory.h"
 #include "test.h"
 #include "thor/bidirectional_astar.h"
+#include "thor/timedep.h"
```
kevinkreiser, line 14: can we revert this file? i know this was a long running pr with many iterations but thsi seems irrelevant now?

run_route_scripts/run_with_server.py
```diff
@@ -93,7 +92,7 @@ def make_request(post_body):
   parser.add_argument('--test-file', type=str, help='The file with the test requests', required=True)
   parser.add_argument('--url', type=str, help='The url to which you want to POST the request bodies', default='http://localhost:8002/route')
   parser.add_argument('--output-dir', type=str, help='The directory in which to place the result of each request')
-  parser.add_argument('--concurrency', type=int, help='The number of processes to use to make requests', default=psutil.cpu_count(logical=True))
+  parser.add_argument('--concurrency', type=int, help='The number of processes to use to make requests', default=multiprocessing.cpu_count())
```
kevinkreiser, line 95: was this something mac specific or?

test/gurka/test_instructions_small_end_ramp_forks.cc
```diff
@@ -62,7 +62,7 @@ class InstructionsSmallEndRampForks : public ::testing::Test {
     const auto layout =
         gurka::detail::map_to_coordinates(ascii_map, gridsize_metres, {5.1079374, 52.0887174});
 
-    map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_instructions_keep_highway_signs",
+    map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_instructions_small_end_ramp_forks",
```
kevinkreiser, line 65: this was a collision with another test, looks like it was a copy paste fail

run_route_scripts/run_with_server.py
mandeepsandhu: No. This happened on Linux as well. I didn't have `psutil` installed so this script was failing. Since `multiprocessing`  has the `cpu_count` I thought of using that instead of `psutil` since its already in use here.

bench/thor/routes.cc
mandeepsandhu: I was using bench for seeing the impact of the change on timedep a-star implementations. We're no longer doing that in this PR now, so I could revert it. Although I feel its okay to test the timedep implementations in the bench tests for completeness.

run_route_scripts/run_with_server.py
kevinkreiser: i mean you can just pip install it but sure, one less thing seems better (i cant remember why i used that in the first place, maybe one works better under docker or something...)

kevinkreiser (approved):
PR looks clean to me now. Sorry for the long slog on this one!

run_route_scripts/run_with_server.py
mandeepsandhu: yeah 1 less package to install. I _ think_ this should work inside docker too.

mandeepsandhu:
> PR looks clean to me now. Sorry for the long slog on this one!

Thanks for slogging with me on this one! :) Appreciate you taking the time to help me along the way :pray:

mandeepsandhu:
Performance impact after latest merge from master:
```shell
branch:         |            50th              |             90th             |         99th
master          | 0.03852168718973795          | 0.07410685221354167          | 0.18502410252888998
this branch     | 0.03876058260599772 (+0.62%) | 0.07470011711120605 (+0.8%)  | 0.18315505981445312 (*-1.01%*)
```
**Note:** These are average of 3 runs for each branch with 20k route requests (pre-warmed tile-cache by running 20k requests before measuring performance of each branch).

RAD tests showed 0 diffs. I think we're good to merge this change.

PR #2688, opened 2020-11-17
fix(openlr): Fixes km to meter conversion in OpenLr serializer

PureW:
There was a bug in the OpenLr serialization leading to the `dnp` value being wrong. A value in kilometer was treated as meter.

The change also renames TripLeg.Edge.length to TripLeg.Edge.length_km to avoid this issue in the future.

### DIFF
--- a/test/mapmatch.cc
+++ b/test/mapmatch.cc
@@ -1428,7 +1428,7 @@ TEST(Mapmatch, openlr_parameter_true_osrm_api) {
       "CwOduyULYiKJAAAV//0iGw==",
       "CwOdxCULYCKJAAAN//4iGw==",
       "CwOdySULXyKJAAAf//EiGw==",
-      "CwOd1yULVyKLAABV/9AiGw==",
+      "CwOd1yULVyKLAQBV/9AiGw==",
   };
   for (const auto& match : matches) {
     std::vector<std::string> references;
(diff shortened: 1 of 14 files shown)


test/mapmatch.cc
```diff
@@ -1428,7 +1428,7 @@ TEST(Mapmatch, openlr_parameter_true_osrm_api) {
       "CwOduyULYiKJAAAV//0iGw==",
       "CwOdxCULYCKJAAAN//4iGw==",
       "CwOdySULXyKJAAAf//EiGw==",
-      "CwOd1yULVyKLAABV/9AiGw==",
+      "CwOd1yULVyKLAQBV/9AiGw==",
```
danpaz, line 1431: Was this change intended for this PR?

PureW: I just updated the description/name of the PR. It actually fixes a bug in the openlr serialization at https://github.com/valhalla/valhalla/pull/2688/files#diff-dd901075ccbb89ba7900fb2348eac4f9da77488c897257ba9f657a10603531a8R80

The rename (most of the diff) is only made to avoid a similar issue in future.

PureW: So yes, this change in test-value is expected.

mandeepsandhu: Yeah I had the same confusion. I missed the line you pointed above, thinking the diff was simply a `s/length/length_km` 🙂

mandeepsandhu (dismissed):

test/mapmatch.cc
danpaz: 👍

mandeepsandhu (approved):

PR #2691, opened 2020-11-18
Improve speed checks during shortcut building

SvetlanaBayarovich:
# Issue

The idea is to create shortcuts only over the edges with the same speed, and do it whenever possible.
In order to achieve it two changes are made.
1. The check `(oppdiredge1->speed() != edge2->speed()) || (oppdiredge2->speed() != edge1->speed())` is added to ensure speed equality. Previously only outgoing edges of a contracted node were checked for equality which could result in undesirable shortcut, for example
image
Here speeds for BA and BC are equal, as a result AC and CA were created, regardless of different speed on CB. Now this is impossible
2. Remove the check for outgoing edges. All that is needed is to ensure that edges going the same direction have same speed - and that is point 1) from above. For example, here 2 shortcuts are created: AC with some speed and CA with some other.
image

Some context: to build a shortcut we look through the edges of some node and try to build it starting from each edge and going as far as possible. As a result, the shortcut in the opposite direction is built when the processing of the end node happens. GraphValidator checks that edges in both directions exist
## Tasklist

 - [x] Add tests

## Requirements / Relations


src/sif/CMakeLists.txt
```diff
@@ -7,7 +7,6 @@ set(sources
   motorcyclecost.cc
   motorscootercost.cc
   nocost.cc
-  osrm_car_duration.h
```
kevinkreiser, line 10: in the diff i dont see the file being actually  moved into the headers directory, how is this building?

kevinkreiser (approved):

src/sif/CMakeLists.txt
kevinkreiser: nevermind for some reason i just didnt notice it!

PR #2693, opened 2020-11-19
Double precision for lon lat

genadz:
Right now `float` is used all over the project for geocoordinates. This accuracy is not enough for some cases. Ideally, we should use `double` always.
What was done:
- where it's possible, `float` was changed to `double`;
- for other cases (where we're limited to 4 bytes) we store coordinates as integers with fixed (6 or 7 digits) precision;
- fixed a lot of unit tests.

## Tasklist

 - [ ] Add tests


genadz:
Performance testing (3 runs)
master branch:                    mean = `3m59s`, std = `6s`
ll_double_precition branch: mean = `4m2s`, std = `5s`

valhalla/baldr/nodeinfo.h
```diff
@@ -389,10 +385,12 @@ class NodeInfo {
...
-  // 26 bits for lat,lon offset allows 7 digit precision within 4 degree tiles. Note that
-  // this would require using double precision to actually achieve this precision.
-  uint64_t lat_offset_ : 26; // Latitude offset from tile base latitude
-  uint64_t lon_offset_ : 26; // Longitude offset from tile base longitude
+  // 26 bits for lat,lon offset allows 7 digits of precision even in 4 degree tiles
+  // to stay backwards compatible we have to break 6 digits and the 7th digit into two parts
+  uint64_t lat_offset_ : 22; // Latitude offset from tile base latitude in int 6 digit precision
+  uint64_t lat_offset7_ : 4; // Latitude offset 7th digit of precision
+  uint64_t lon_offset_ : 22; // Longitude offset from tile base longitude in int 6 digit precision
+  uint64_t lon_offset7_ : 4; // Longitude offset 7th digit of precision
```
kevinkreiser, line 393: this was a nice little trick i worked out to add a full 7 digits of precision to the nodes in the graph in a backwards compatible way. since previously we only did 6, it meant we only ever used 22 bit wide numbers at the most. since valhalla tiles assume little endian, i was able to just steal back the 4 unused bits and use them to store the 7th digit of precision.

src/baldr/tilehierarchy.cc
```diff
@@ -6,21 +6,25 @@ using namespace valhalla::midgard;
 namespace valhalla {
 namespace baldr {
 
-const std::map<uint8_t, TileLevel>& TileHierarchy::levels() {
+const std::vector<TileLevel>& TileHierarchy::levels() {
```
kevinkreiser, line 9: so i just couldnt let this stupidity linger. while chasing down a failing test, i decided to refactor the hierarchy static structure. there is no reason this should be a map. instead its a vector where the index is the level. why do binary search on something that you already know where the indices should be? this caused a lot of diffs in the rest of this PR. if you see anything with `TileHeirarchy::` it was this refactor

src/baldr/nodeinfo.cc
```diff
@@ -84,14 +84,21 @@ NodeInfo::NodeInfo(const PointLL& tile_corner,
...
-  lat_offset_ = ll.second < tile_corner.lat()
-                    ? 0
-                    : static_cast<uint32_t>((ll.second - tile_corner.lat()) / kDegreesPrecision);
-  lon_offset_ = ll.first < tile_corner.lng()
-                    ? 0
-                    : static_cast<uint32_t>((ll.first - tile_corner.lng()) / kDegreesPrecision);
+  lat_offset_ = 0;
+  if (ll.lat() > tile_corner.lat()) {
+    auto lat = std::round((ll.lat() - tile_corner.lat()) / 1e-7);
+    lat_offset_ = lat / 10;
```
kevinkreiser, line 92: old code will still work because the first 22 bits get the 6 digit lat/lon as they did before then below we also set the 4 extra bits with the 7th digit

kevinkreiser:
nice, the code somehow actually got 10 lines less as well!

test/mapmatch.cc
```diff
@@ -170,6 +170,8 @@ void compare_results(const valhalla::Api& expected, const valhalla::Api& result)
     for (int l = 0; l < route.legs_size(); ++l) {
       const auto& leg = route.legs(l);
 
+      std::cout << leg_answer->shape() << std::endl << leg.shape() << std::endl << std::endl;
```
danpaz, line 173: Remove print

valhalla/baldr/graphtileheader.h
```diff
@@ -162,15 +163,18 @@ class GraphTileHeader {
    * Get the base (SW corner) of the tile.
    * @return Returns the base lat,lon of the tile (degrees).
    */
-  const midgard::PointLL& base_ll() const {
-    return static_cast<const midgard::PointLL&>(base_ll_);
+  midgard::PointLL base_ll() const {
+    GraphId id(graphid_);
+    return TileHierarchy::levels()[id.level()].tiles.Base(id.tileid());
+    // return midgard::PointLL(base_ll_.first, base_ll_.second);
```
danpaz, line 169: Why did this change?

kevinkreiser: very insightful question! it has to do with getting full precision graph node lat lons. node lat lons are stored as an offset from the corner of the tile. for some reason we stored the corner of the tile in float in the tile header. the thing about a tile though is that we already know what the corner is by simple math on the tileid. so instead of storing it we just do a couple of multiplications here to get the corner of the tile in double precision rather than float that is stored. look  at the TODO i added at the very bottom of this file:

https://github.com/valhalla/valhalla/pull/2693/files#diff-02d2f61e72de41c5a1fb592eaf617f2e3c24c1882ed77791a4e876e9a76fd05eR600

danpaz: Ah nice. I wonder if we stored the corner ll in case we change tiling in the future and can no longer rely on the id? In anycase, sgtm 👍

valhalla/midgard/encoded.h
```diff
@@ -1,22 +1,110 @@
...
 
 #include <cmath>
+#include <stdexcept>
 #include <string>
 #include <type_traits>
 #include <vector>
 
+// we store 6 digits of precision in the tiles, changing to 7 digits is a breaking change
+// if you want to try out 7 digits of precision you can uncomment this definition
+//#define USE_7DIGITS_DEFAULT
```
kevinkreiser, line 11: this breaks compatibility but, for testing, if you want you can try out 7 digits of precision with this simple preprocessor macro

valhalla/midgard/encoded.h
```diff
@@ -1,22 +1,110 @@
...
+#else
+constexpr double DECODE_PRECISION = 1e-6;
+constexpr int ENCODE_PRECISION = 1e6;
+constexpr size_t DIGITS_PRECISION = 6;
+#endif
+
 namespace valhalla {
 namespace midgard {
 
+template <typename Point> class Shape7Decoder {
```
kevinkreiser, line 25: i ditched the other header that had this info and just moved it to this one. its nice to have all the polyline encoding/decoding in one place

danpaz (dismissed):
Approved pending some perf testing by @kevinkreiser @genadz

kevinkreiser:
i revalidated all the changes with the same perf test as @genadz a moment ago. on my machine master took 43.1s and this branch took 43.9 seconds, which is about a 1.8% penalty. we are good to go

PR #2697, opened 2020-11-24
Rework recost function to use TimeInfo

genadz:
This PR is helper for the https://github.com/valhalla/valhalla/pull/2661 .
Changes:
- use `TimeInfo` structure instead of `std::string` to represent date time in recost function;
- pass `invariant` option to recost function to supply static date time.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/thor/triplegbuilder.cc
```diff
@@ -1124,10 +1129,11 @@ void TripLegBuilder::Build(
...
     // we could cache the timezone and just add seconds when the timezone doesnt change
-    time_info = time_info.forward(trip_path.node_size() == 0
-                                      ? 0.0
-                                      : trip_path.node().rbegin()->cost().elapsed_cost().seconds(),
-                                  node->timezone());
+    const auto offset_time =
+        time_info.forward(trip_path.node_size() == 0
+                              ? 0.0
+                              : trip_path.node().rbegin()->cost().elapsed_cost().seconds(),
+                          static_cast<int>(node->timezone()));
```
kevinkreiser, line 1136: i guess this is a bug i introduced in the `invariant` PR. 

```suggestion
    const auto offset_time =
        time_info.forward(trip_path.node_size() == 0 || invariant
                              ? 0.0
                              : trip_path.node().rbegin()->cost().elapsed_cost().seconds(),
                          static_cast<int>(node->timezone()));
```

src/thor/triplegbuilder.cc
```diff
@@ -1376,7 +1382,7 @@ void TripLegBuilder::Build(
   }
 
   // Add that extra costing information if requested
-  AccumulateRecostingInfoForward(options, start_pct, end_pct, date_time, graphreader, trip_path);
+  AccumulateRecostingInfoForward(options, start_pct, end_pct, time_info, graphreader, trip_path);
```
kevinkreiser, line 1385: i see you want to use the unchanged one here. if that is the case then i suggest you make a copy of it to modify in the loop to avoid unneccesary timezone offsets and look ups

valhalla/baldr/time_info.h
```diff
@@ -41,6 +41,14 @@ struct TimeInfo {
   // a timezone offset cache because doing the offset math is expensive
   baldr::DateTime::tz_sys_info_cache_t* tz_cache;
 
+  /**
+   * Create TimeInfo object with default parameters.
+   * @return    TimeInfo structure
+   */
+  static TimeInfo empty() {
```
kevinkreiser, line 48: nitpick, id call this `invalid` because this disables any adjustments to the time.

kevinkreiser:
@genadz this looks really good, just some small changes required!

valhalla/baldr/time_info.h
genadz: ok, will rename

src/thor/triplegbuilder.cc
genadz: good catch! but I thinks your suggestion also contains a bug: why do you use such `trip_path.node().rbegin()->cost().elapsed_cost().seconds()` offset ? If we update `time_info` at each iteration we should move forward for time difference between 2 last nodes;
I think, we have similar bug here https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L167-L168 - pass `offset_time` to the method that expects time info for the start of the route

genadz: hm, just realized, If we update `time_info` on each iteration, it will look like:
```c++
    float seconds_diff = (is_first_edge || invariant)
                         ? 0.f
                         : std::prev(edge_itr)->elapsed_cost.secs;
    if (std::distance(path_begin, edge_itr) > 1)
      seconds_diff -= (edge_itr - 2)->elapsed_cost.secs;
    time_info = time_info.forward(seconds_diff, static_cast<int>(node->timezone()));
```
here `time_info` represents actual time for the current node. If we updated `time_info` on each iteration, we would accumulate rounding error for the time, due to the fact, that `time_info` keeps time as integer;

and, one more problem I noticed: here https://github.com/valhalla/valhalla/blob/master/src/thor/triplegbuilder.cc#L1129 we use offset from the previous node, why ? looks like a bug

src/thor/triplegbuilder.cc
```diff
@@ -1122,12 +1133,14 @@ void TripLegBuilder::Build(
...
     }
 
+    const bool is_first_edge = edge_itr == path_begin;
+    const bool is_last_edge = edge_itr == (path_end - 1);
+
     // have to always compute the offset in case the timezone changes along the path
     // we could cache the timezone and just add seconds when the timezone doesnt change
-    time_info = time_info.forward(trip_path.node_size() == 0
-                                      ? 0.0
-                                      : trip_path.node().rbegin()->cost().elapsed_cost().seconds(),
```
kevinkreiser, line 1129: @genadz good point! it looks like moving way into the future with this previous code. i can only imagine the code was supposed to be:

```c++
trip_path.node().rbegin()->cost().elapsed_cost().seconds() - (trip_path.node().rbegin()++)->cost().elapsed_cost().seconds()
```

kevinkreiser: concretely this meant that the edge speed attribute was incorrect when using historical speeds

kevinkreiser: @genadz ill quickly performance test this to make sure the timezone stuff doesnt matter and we'll get it shipped

kevinkreiser (approved):

kevinkreiser:
There was no difference in performance of timeless and time dependent routes

PR #2698, opened 2020-11-24
Fix polyline simplification float/double comparison

danpaz:
# Issue

#2693 updated coordinates to use double precision. In [polyline simplification code we were relying on a `>=` equality to compare distances](https://github.com/valhalla/valhalla/blob/master/src/midgard/polyline2.cc#L92), which tends to fail when one side of the comparison is float and the other is double. 

The result was we were almost always simplifying the entire polyline to 2 points, because the first iteration of the simplifier sets `dmax = epsilon`. The gurka test here captures that failure mode.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/midgard/polyline2.cc
+++ b/src/midgard/polyline2.cc
@@ -64,7 +64,7 @@ void Polyline2<coord_t>::Generalize(container_t& polyline,
   peucker = [&peucker, &polyline, epsilon, &indices](typename container_t::iterator start, size_t s,
                                                      typename container_t::iterator end, size_t e) {
     // find the point furthest from the line
-    float dmax = std::numeric_limits<typename coord_t::value_type>::lowest();
+    double dmax = std::numeric_limits<typename coord_t::value_type>::lowest();
     typename container_t::iterator itr;
     LineSegment2<coord_t> l{*start, *end};
     size_t j = e - 1, k;
--- a/src/midgard/polyline2.cc
+++ b/src/midgard/polyline2.cc
@@ -64,7 +64,7 @@ void Polyline2<coord_t>::Generalize(container_t& polyline,
   peucker = [&peucker, &polyline, epsilon, &indices](typename container_t::iterator start, size_t s,
                                                      typename container_t::iterator end, size_t e) {
     // find the point furthest from the line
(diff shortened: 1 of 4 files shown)


src/midgard/polyline2.cc
```diff
@@ -64,7 +64,7 @@ void Polyline2<coord_t>::Generalize(container_t& polyline,
   peucker = [&peucker, &polyline, epsilon, &indices](typename container_t::iterator start, size_t s,
                                                      typename container_t::iterator end, size_t e) {
     // find the point furthest from the line
-    float dmax = std::numeric_limits<typename coord_t::value_type>::lowest();
+    double dmax = std::numeric_limits<typename coord_t::value_type>::lowest();
```
danpat, line 67: This should probably be `coord_t::value_type` instead of `double`, or `auto` to avoid this happening again when we move to `bigdouble` :-P

kevinkreiser:
good catch. tbh id be surprised if there werent a few more of these sprinkled in. we should take another pass at looking how and when floats are used

src/midgard/polyline2.cc
danpaz: 👍  Fixed

kevinkreiser (approved):

src/midgard/polyline2.cc
```diff
@@ -64,7 +64,7 @@ void Polyline2<coord_t>::Generalize(container_t& polyline,
   peucker = [&peucker, &polyline, epsilon, &indices](typename container_t::iterator start, size_t s,
                                                      typename container_t::iterator end, size_t e) {
     // find the point furthest from the line
-    float dmax = std::numeric_limits<typename coord_t::value_type>::lowest();
+    coord_t::value_type dmax = std::numeric_limits<typename coord_t::value_type>::lowest();
```
kevinkreiser, line 67: use `auto` or prepend `typename` here

danpaz: :+1: Thanks, fixed

kevinkreiser (approved):

danpat (dismissed):

gknisely (dismissed):

danpat (approved):

PR #2699, opened 2020-11-27
Track Distance in Isochrones and Generate Distance Contours Optionally

kevinkreiser:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


nilsnolde:
@kevinkreiser thanks so much for starting this journey!!

So, I'm almost done with the implementation and on the last edits I realized we might be stuck here without starting to track distance as part of `sif::Cost`.. We don't need it for costing the edges (i.e. shortest) but (I think) we do need the aggregated distance so we can tell dijkstra to stop expanding when the maximum `distance` is reached. I'll comment more in the changeset..

nilsnolde:
On a general note: the way it works now is that each contour **can have** a `time` or `distance` member, but **must have** one of them.

src/thor/isochrone.cc
```diff
@@ -88,6 +76,28 @@ void Isochrone::ConstructIsoTile(
...
+    max_distance_est = max_seconds_ * 5.0f * kMPHtoMetersPerSec;
+  } else {
+    // A driving mode
+    max_time_est = max_distance / (5.0f * kMPHtoMetersPerSec);
+    max_distance_est = max_seconds_ * 70.0f * kMPHtoMetersPerSec;
+  }
+
+  max_distance = std::max(max_distance_est, max_distance);
+  max_seconds_ = std::max(max_seconds_, max_time_est);
+
```
nilsnolde, line 100: here I also estimate `max_seconds` similar to how it's been done before for `maxdistance`. I know this is sorta stupid, it's rather a side-effect of what's wrong with how we're currently doing it. `max_seconds_` is the criteria for the dijkstra to stop expanding. So it can't be 0 (as would be the case if there was no `time` contour member at all).

src/thor/isochrone.cc
```diff
@@ -298,6 +312,10 @@ ExpansionRecommendation Isochrone::ShouldExpand(baldr::GraphReader& graphreader,
   // Continue if the time interval has been met. This bus or rail line goes beyond the max
   // but need to consider others so we just continue here. Tells MMExpand function to skip
   // updating or pushing the label back
+
+  // TODO: cost should really also track distance for isodistances to work properly;
+  // right now they'll have an estimated time stop criteria for dijkstra..
+  // that's not the same as shortest
```
nilsnolde, line 318: I _think_ we do need the accumulated distance here from the Cost to say `|| pred.cost().dist > max_distance_`. Apart from a bigger effort regarding changing the Cost definition, this also would mean that both time and distance (if available) would be responsible to stop the dijkstra. If we had another API parameter for contours, e.g. `type` which could specify e.g. `time`, `distance` or `time_distance` (depending on what the user wants the stop criteria to be) we could do smth more flexible.

src/thor/isochrone_action.cc
```diff
@@ -13,11 +13,26 @@ std::string thor_worker_t::isochrones(Api& request) {
...
+    // time_distances.push_back({contour.time() * 60, contour.distance()});
+    contours_time.push_back(contour.time());
+    contours_distance.push_back(contour.distance());
+    colors[contour.time()] = contour.color();
+    // colors[contour.distance()] = contour.color();
+  }
+
+  // sort the contour vectors
+  std::sort(contours_time.begin(), contours_time.end());
+  std::sort(contours_distance.begin(), contours_distance.end());
```
nilsnolde, line 30: sorting here so we don't have to do it in any downstream method anymore..

src/thor/isochrone_action.cc
```diff
@@ -31,18 +46,25 @@ std::string thor_worker_t::isochrones(Api& request) {
...
+      grid->GenerateContours(contours_time, [](const time_distance_t& td) { return td.sec; },
+                             options.polygons(), options.denoise(), options.generalize());
+  auto isolines_dist =
+      grid->GenerateContours(contours_distance, [](const time_distance_t& td) { return td.dist; },
+                             options.polygons(), options.denoise(), options.generalize());
 
-  return tyr::serializeIsochrones<PointLL>(request, isolines, options.polygons(), colors,
-                                           options.show_locations());
+  return tyr::serializeIsochrones(request, isolines_dist, options.polygons(), colors,
+                                  options.show_locations());
```
nilsnolde, line 67: didn't add yet any merging for both metrics..

src/worker.cc
```diff
@@ -535,26 +535,27 @@ void parse_contours(const rapidjson::Document& doc,
...
     for (const auto& json_contour : *json_contours) {
-      // Grab contour time and validate that it is increasing
-      const float c = rapidjson::get_optional<float>(json_contour, "/time").get_value_or(NO_TIME);
-      if (c < prev || c == NO_TIME) {
+      // Grab contour time and distance
+      const float t =
+          rapidjson::get_optional<float>(json_contour, "/time").get_value_or(midgard::kNoIsoMetric);
+      const float d = rapidjson::get_optional<float>(json_contour, "/distance")
+                          .get_value_or(midgard::kNoIsoMetric);
+      if ((t == midgard::kNoIsoMetric && d == midgard::kNoIsoMetric)) {
```
nilsnolde, line 544: I allow unsorted contours, they will be sorted in isochrone_action.cc

nilsnolde: not sure yet if that has any downstream effects.. couldn't see any so far

valhalla/midgard/util.h
```diff
@@ -36,6 +36,9 @@ template <class T> struct ranged_default_t {
   }
 };
 
+// Applied if time or distance were omitted in the request
+constexpr float kNoIsoMetric = 0.0f;
```
nilsnolde, line 40: prob not the best place for it.. will change for sure

nilsnolde:
@kevinkreiser the implementation for isodistances is solid now:

- specify `time` **or** `distance` (in km) per contour
- GeoJSON features have an extra field `metric` which is either `time` or `distance`
- added a `max_distance_contours` limit to `valhalla_build_config`

It's not yet building everything, but `valhalla_service` target builds fine. I'll wait with the remaining tasks for a first review, I imagine you'd like to change a few things ;)

There's of course a performance penalty for adding this, but it's on the mild side I'd say.. For a 20 mins auto contour in Minneapolis it's around 10% (no benchmark..): 3.5 instead of 3.2 secs. IMHO still worth it.

Open questions:
- what kind of tests would you like to see here?

Open tasks:
- tests
- implement isodistances for multimodal as well
- adapt other scripts/executables (e.g. `valhalla_run_isochrone`)

src/thor/dijkstras.cc
```diff
@@ -157,10 +157,11 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
       }
     }
 
-    // Compute the cost to the end of this edge
+    // Compute the cost and path distance to the end of this edge
     Cost transition_cost = costing_->TransitionCost(directededge, nodeinfo, pred);
     Cost newcost = pred.cost() + costing_->EdgeCost(directededge, tile, offset_time.second_of_week) +
                    transition_cost;
+    uint32_t path_dist = pred.path_distance() + directededge->length();
```
nilsnolde, line 164: I chose `path_distance` instead of `distance`, I figured it's semantically more correct. I know we discussed adding this directly to the isochrone's `UpdateIsoTile()` and leaving the dijkstra alone.. But the edge labels have no way to set `path_distance` after they're created and supporting that would need quite a lot more change than simply tracking the path distance in the dijkstra.

nilsnolde: I did need to add an overload on the `BDEdgeLabel`s constructor to support this, see below..

src/thor/dijkstras.cc
```diff
@@ -183,7 +184,7 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
     uint32_t idx = bdedgelabels_.size();
     *es = {EdgeSet::kTemporary, idx};
     bdedgelabels_.emplace_back(pred_idx, edgeid, oppedgeid, directededge, newcost, newcost.cost, 0.0f,
-                               mode_, transition_cost, false, restriction_idx);
+                               mode_, path_dist, transition_cost, false, restriction_idx);
```
nilsnolde, line 187: I did need to add an overload on the `BDEdgeLabel`s constructor to support this, see below..

src/thor/isochrone.cc
```diff
@@ -51,21 +51,25 @@ Isochrone::Isochrone() : Dijkstras(), shape_interval_(50.0f) {
...
+  max_meters_ = max_km * kMetersPerKm;
   float max_distance;
-  max_seconds_ = max_minutes * 60;
   if (multimodal) {
     max_distance = max_seconds_ * 70.0f * kMPHtoMetersPerSec;
   } else if (mode == TravelMode::kPedestrian) {
     max_distance = max_seconds_ * 5.0f * kMPHtoMetersPerSec;
   } else if (mode == TravelMode::kBicycle) {
-    max_distance = max_seconds_ * 20.0f * kMPHtoMetersPerSec;
+    max_distance = max_seconds_ * 5.0f * kMPHtoMetersPerSec;
```
nilsnolde, line 66: oops, I'll revert

src/thor/isochrone.cc
```diff
@@ -217,17 +226,17 @@ void Isochrone::UpdateIsoTile(const EdgeLabel& pred,
     auto tile1 = isotile_->TileId(ll0);
     auto tile2 = isotile_->TileId(ll);
     if (tile1 == tile2) {
-      isotile_->SetIfLessThan(tile1, secs1 * kMinPerSec);
+      SetData(tile1, {secs1 * kMinPerSec, dist1 * kMetersPerKm});
```
nilsnolde, line 229: this new method will update the grid's time and/or distance value depending if it's smaller than the current value(s).

src/thor/isochrone.cc
```diff
@@ -295,10 +308,11 @@ ExpansionRecommendation Isochrone::ShouldExpand(baldr::GraphReader& /*graphreade
...
     }
   }
-  // Continue if the time interval has been met. This bus or rail line goes beyond the max
-  // but need to consider others so we just continue here. Tells MMExpand function to skip
+  // Continue if the time and distance intervals have been met. This bus or rail line goes beyond the
+  // max but need to consider others so we just continue here. Tells MMExpand function to skip
   // updating or pushing the label back
-  if (pred.cost().secs > max_seconds_ || pred.cost().cost > max_seconds_ * 4) {
+  if ((pred.cost().secs > max_seconds_ || pred.cost().cost > max_seconds_ * 4) &&
+      pred.path_distance() > max_meters_) {
```
nilsnolde, line 315: Only stop dijkstra expansion when both max time and max distance have been met.

src/thor/isochrone_action.cc
```diff
@@ -13,12 +13,18 @@ std::string thor_worker_t::isochrones(Api& request) {
   auto costing = parse_costing(request);
   auto& options = *request.mutable_options();
 
-  std::vector<float> contours;
-  std::unordered_map<float, std::string> colors;
+  std::vector<float> contours_time, contours_distance;
```
nilsnolde, line 16: so the way it works now:

- each PBF contour **always** has time and distance set (can of course be 0)
- collect them into separate vectors and sort descending
- only generate contours if the highest metric is greater than 0
- skip the geometry in the serialization if it's a 0 contour

src/tyr/isochrone_serializer.cc
```diff
@@ -18,71 +18,80 @@ using rgba_t = std::tuple<float, float, float>;
...
-      auto x = c * (1 - std::abs(std::fmod(h / 60.f, 2.f) - 1));
-      auto m = .25f;
-      rgba_t color = h < 60 ? rgba_t{m + c, m + x, m}
-                            : (h < 120 ? rgba_t{m + x, m + c, m} : rgba_t{m, m + c, m + x});
-      hex << "#" << std::hex << static_cast<int>(std::get<0>(color) * 255 + .5f) << std::hex
-          << static_cast<int>(std::get<1>(color) * 255 + .5f) << std::hex
-          << static_cast<int>(std::get<2>(color) * 255 + .5f);
-    }
-    ++i;
+  for (const auto& metric_contours : grid_contours) {
```
nilsnolde, line 31: Looks like a huge change, but really it's just adding an extra loop over the metrics and setting a few more high-level variables.

src/tyr/isochrone_serializer.cc
```diff
@@ -18,71 +18,80 @@ using rgba_t = std::tuple<float, float, float>;
...
                          })},
-      }));
+            {"properties", map({{"contour", static_cast<uint64_t>(interval.first)},
+                                {"color", hex.str()},            // lines
+                                {"fill", hex.str()},             // geojson.io polys
+                                {"fillColor", hex.str()},        // leaflet polys
+                                {"opacity", fp_t{.33f, 2}},      // lines
+                                {"fill-opacity", fp_t{.33f, 2}}, // geojson.io polys
+                                {"fillOpacity", fp_t{.33f, 2}},  // leaflet polys
+                                {"metric", std::string(metric_json)}})},
```
nilsnolde, line 92: distinguish between metrics in the output

src/worker.cc
```diff
@@ -100,7 +100,7 @@ const std::unordered_map<unsigned, unsigned> ERROR_TO_STATUS{
     {150, 400}, {151, 400}, {152, 400}, {153, 400}, {154, 400}, {155, 400}, {156, 400}, {157, 400},
     {158, 400}, {159, 400},
 
-    {160, 400}, {161, 400}, {162, 400}, {163, 400}, {164, 400}, {165, 400},
+    {160, 400}, {161, 400}, {162, 400}, {163, 400}, {164, 400}, {165, 400}, {166, 400},
```
nilsnolde, line 103: forgot to add `167` as well

src/worker.cc
```diff
@@ -535,26 +535,29 @@ void parse_contours(const rapidjson::Document& doc,
   // make sure the isoline definitions are valid
   auto json_contours = rapidjson::get_optional<rapidjson::Value::ConstArray>(doc, "/contours");
   if (json_contours) {
-    float prev = 0.f;
-    const float NO_TIME = -1.f;
     for (const auto& json_contour : *json_contours) {
-      // Grab contour time and validate that it is increasing
-      const float c = rapidjson::get_optional<float>(json_contour, "/time").get_value_or(NO_TIME);
-      if (c < prev || c == NO_TIME) {
+      // Grab contour time and distance
```
nilsnolde, line 539: we can allow random contours actually. the logic which depended on this was slightly changed now

valhalla/midgard/util.h
nilsnolde: any recommendations? btw, this goes for quite a few methods/constants.. not sure if I placed them right. though here I'm sure it's not the right place;)

valhalla/sif/edgelabel.h
```diff
@@ -446,6 +446,47 @@ class BDEdgeLabel : public EdgeLabel {
         opp_edgeid_(oppedgeid), not_thru_pruning_(not_thru_pruning) {
   }
 
+  /**
```
nilsnolde, line 449: overload to be able to track the path distance

valhalla/thor/isochrone.h
```diff
@@ -133,7 +138,26 @@ class Isochrone : public Dijkstras {
...
-                     const float secs0);
+                     const float secs0,
+                     const float dist0);
+
+  /**
+   * Updates data grid values for time and distance if present.
+   * @param tile_id Tile Id to set value for.
+   * @param value   Value to set at the tile/grid location.
+   */
+  inline void SetData(const int tile_id, const midgard::time_distance_t& value) {
```
nilsnolde, line 149: wrapper to update the grid's values

valhalla/worker.h
```diff
@@ -67,6 +67,8 @@ const std::unordered_map<unsigned, std::string>
                 {163, "Invalid date_type"},
                 {164, "Invalid shape format"},
                 {165, "Date and time required for destination for date_type of invariant"},
+                {166, "Exceeded max distance"},
+                {167, "Too many metrics in one contour"},
```
nilsnolde, line 71: don't forget to map 167 to HTTP status code...

src/thor/isochrone_action.cc
```diff
@@ -31,18 +37,35 @@ std::string thor_worker_t::isochrones(Api& request) {
...
-  return tyr::serializeIsochrones<PointLL>(request, isolines, options.polygons(), colors,
-                                           options.show_locations());
+  auto grid =
+      (costing == "multimodal" || costing == "transit")
+          ? isochrone_gen.ComputeMultiModal(*options.mutable_locations(),
+                                            contours_time.front() + 10.0f,
+                                            contours_distance.front() + 10.0f, *reader, mode_costing,
+                                            mode)
+          : isochrone_gen.Compute(*options.mutable_locations(), contours_time.front() + 10.0f,
+                                  contours_distance.front() + 10.0f, *reader, mode_costing, mode);
```
nilsnolde, line 47: Actually I'm not so happy with adding 10 km here.. For time it's not so terrible to add 10 mins as we modulate the grid size with the mode speed, so non-motor vehicles get smaller grids than motorized vehicles. but adding 10 km means a 1 km pedestrian iso gets an 11 km grid assigned, which makes the grid bounds much larger and increases the blockiness.. might be fine for now, but maybe the whole grid definition should change a little eventually?

src/thor/isochrone_action.cc
```diff
@@ -13,12 +13,18 @@ std::string thor_worker_t::isochrones(Api& request) {
...
-    contours.push_back(contour.time());
-    colors[contours.back()] = contour.color();
+    contours_time.push_back(contour.time());
+    colors[IsoMetrics::kTime][contour.time()] = contour.color();
+    contours_distance.push_back(contour.distance());
+    colors[IsoMetrics::kDistance][contour.distance()] = contour.color();
   }
+  // sort the contour vectors in descending order
+  std::sort(contours_time.begin(), contours_time.end(), std::greater<float>());
+  std::sort(contours_distance.begin(), contours_distance.end(), std::greater<float>());
```
nilsnolde, line 27: sorting the contour vectors here means we don't have to do it when validating the request params and we don't have to sort or check the sorting in other downstream functions

src/loki/isochrone_action.cc
```diff
@@ -55,11 +65,15 @@ void loki_worker_t::init_isochrones(Api& request) {
     throw valhalla_exception_t{152, std::to_string(max_contours)};
   }
 
-  // validate the contour time by checking the last one
-  const auto contour = options.contours().rbegin();
-  if (contour->time() > max_time) {
+  // check the contour metrics
+  float max_time_c, max_dist_c;
+  std::tie(max_time_c, max_dist_c) = get_max_metrics(options.contours());
```
nilsnolde, line 70: previously this depended on contours being sorted in ascending order. pulling out the max of 2 metrics from a flat "list" we can't do that anymore and might as well allow unsorted contours in the request params..

scripts/valhalla_build_config
```diff
@@ -236,7 +236,8 @@ config = {
       'max_contours': 4,
       'max_time': 120,
       'max_distance': 25000.0,
-      'max_locations': 1
+      'max_locations': 1,
+      'max_distance_contours': 120
```
nilsnolde, line 240: hmpf.. in the code I actually named it `max_kilometers` not `max_distance_contours`.. any preference? or smth else entirely?

kevinkreiser: im not sure we even need a new config parameter here? max_distance already specifies the largest distance contour value. max_contours still specifies the max number of contours regardless of type (time or distance you only get 4 total by default). maybe you want to change the default of the secone one but even that i think is fine to stay 4 by default

src/loki/isochrone_action.cc
```diff
@@ -31,6 +31,16 @@ void check_distance(const google::protobuf::RepeatedPtrField<valhalla::Location>
     }
   }
 }
+
+const std::tuple<float, float>
+get_max_metrics(const google::protobuf::RepeatedPtrField<valhalla::Contour>& contours) {
+  float max_dist, max_time = 0.0f;
+  for (auto& contour : contours) {
+    max_time = std::max(max_time, contour.time());
+    max_dist = std::max(max_dist, contour.distance());
```
kevinkreiser, line 40: i think a contour should have distance or time but not both.

src/thor/dijkstras.cc
kevinkreiser: there is an overload that does take this info already:

https://github.com/valhalla/valhalla/blob/master/valhalla/sif/edgelabel.h#L471

i think that one will work for you. i know its pretty annoying to figure out what is what especially with emplace_back

scripts/valhalla_build_config
nilsnolde: I would recommend a new one! max_distance specifies the largest distance between input locations (I guess to sorta limit the size of the data grid). one could use simply max_time, but that's a little confusing, and for slower costing models (bikes, ped) max_time and max_distance are quite a different concept. I think 4 input contours are fair still

src/loki/isochrone_action.cc
nilsnolde: If you go further in the review, you'll see that a request can have only time or distance. But when parsing it I found it more elegant to add 0 time or distance to the PBF object.

src/thor/dijkstras.cc
nilsnolde: yep you're right..

src/loki/isochrone_action.cc
kevinkreiser: you can and totally should have both in the same object that makes perfect sense but imho its way more standard in protobuf2 to check which one is set using `contour.has_time` or `contour .has_distance`

scripts/valhalla_build_config
kevinkreiser: ah right sorry for the confusion. its just an unfortunate naming clash. maybe we could rename `max_time` to `max_time_contour` and then add a `max_distance_contour`?

src/loki/isochrone_action.cc
nilsnolde: can change that, sure. then will it return false if `contour.time() == 0`?

scripts/valhalla_build_config
nilsnolde: yeah, agree, unfortunate naming.. won't that be a breaking change for all existing config files then? right now the `max_time` setting is required in loki_worker. we'd need to make it optional and provide a default in loki_worker if we'd want to rename right?

nilsnolde: or duplicate the max_time with a new name in the config. both a little ugly IMO. but maybe as ugly as the current naming;)

src/loki/isochrone_action.cc
kevinkreiser: sorry im suggesting this specifically:

```c++
max_time = std::max(max_time, contour.has_time() ? contour.time() : max_time);
max_dist = std::max(max_dist, contour.has_distance() ? contour.distance() : max_dist);
```

then you dont to to zero out the time if its missing or the distance if its missing in parsing code in worker.cc

scripts/valhalla_build_config
kevinkreiser: yes it would be but we break configs all the time by adding new required ones. there currently no real way around it so i think its the cost of doing business. the good news is that running code with a malformed config immediately fails and you get an error message why

src/loki/isochrone_action.cc
nilsnolde: done

scripts/valhalla_build_config
nilsnolde: done

src/thor/dijkstras.cc
nilsnolde: done

nilsnolde:
Ugh, smth is really wrong here again.. I'll investigate, no need to review yet..

nilsnolde:
Ok, found the error.. good for further review..

src/thor/dijkstras.cc
kevinkreiser: no this was the right fix. actually all of the algorithms should be tracking this. really what we should do is remove it from the  label and add it to the costing and let it be tracked in there.

src/thor/isochrone.cc
```diff
@@ -217,17 +239,17 @@ void Isochrone::UpdateIsoTile(const EdgeLabel& pred,
     auto tile1 = isotile_->TileId(ll0);
     auto tile2 = isotile_->TileId(ll);
     if (tile1 == tile2) {
-      isotile_->SetIfLessThan(tile1, secs1 * kMinPerSec);
+      isotile_->SetIfLessThan(tile1, {secs1 * kMinPerSec, dist1 * kMetersPerKm});
```
kevinkreiser, line 242: the constant looks wrong here, shouldnt it be something like `kKmPerMeter` and be like `0.001f`?

src/thor/dijkstras.cc
```diff
@@ -170,7 +171,7 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
       if (newcost.cost < lab.cost().cost) {
         float newsortcost = lab.sortcost() - (lab.cost().cost - newcost.cost);
         adjacencylist_->decrease(es->index(), newsortcost);
-        lab.Update(pred_idx, newcost, newsortcost, transition_cost, restriction_idx);
+        lab.Update(pred_idx, newcost, newsortcost, path_dist, transition_cost, restriction_idx);
```
kevinkreiser, line 174: order of the arguments is wrong here

src/thor/dijkstras.cc
```diff
@@ -865,8 +871,8 @@ void Dijkstras::SetDestinationLocations(
       // edge (edgeid) is set.
       uint32_t idx = bdedgelabels_.size();
       int restriction_idx = -1;
-      bdedgelabels_.emplace_back(kInvalidLabel, opp_edge_id, edgeid, opp_dir_edge, cost, cost.cost,
-                                 0., mode_, Cost{}, false, restriction_idx);
+      bdedgelabels_.emplace_back(kInvalidLabel, edgeid, opp_edge_id, directededge, cost, mode_,
```
kevinkreiser, line 874: copy paste fail here, using the wrong edge ids and edges

kevinkreiser:
As @nilsnolde noted I'm seeing approximately a 10% performance reduction by adding this extra dimension. I havent profiled it to see exactly where its coming from. I dont think tracking the distance in the labels or computing partial distance during the expansion is at all expensive but the way we are storing/accessing it in the grid potentially is as we've essentially doubled the memory footprint there. at any rate, i was hoping that the cost would be negligable but it seems like not :frowning_face:

nilsnolde:
>  the way we are storing/accessing it in the grid

would think so too.. so the assumption is then that any extra dimension will likely also add ~10% approx in the future.. And the way you changed my implementation it'll be quite easy to add other dimensions..

I personally don't think that's not too bad of a trade-off considering what we gain in terms of functionality. Can we gain some performance from other places? Didn't look into it, but what would it take to make the dijkstra use shortcuts?

nilsnolde:
I wonder why it's failing with `exceeding max distance:200`...  with no distance specified in the request it should default to 0, can't see how that's failing..

src/loki/isochrone_action.cc
```diff
@@ -31,6 +31,16 @@ void check_distance(const google::protobuf::RepeatedPtrField<valhalla::Location>
     }
   }
 }
+
+const std::tuple<float, float>
+get_max_metrics(const google::protobuf::RepeatedPtrField<valhalla::Contour>& contours) {
+  float max_dist, max_time = 0.0f;
```
kevinkreiser, line 37: @nilsnolde this is a bug, here max_dist undefined behavior. it will have random junk in it. when that random junk is a high value, this function will return that value

run_route_scripts/run_with_server.py
```diff
@@ -21,34 +21,54 @@ def get_post_bodies(filename):
       post_body['id'] = str(line_number)
       yield post_body
 
-# for persistant connections
-def initialize(url_, output_dir_):
+def initialize(args_):
```
kevinkreiser, line 24: i did a bunch of refactoring here to make this script more useful. first i moved all the program arguments to a global so that all processes could see them all. then i added the null format option which is useful when you just want to performance test an endpoint but dont care to parse the result (isochrone diffing makes no sense tbh)

src/thor/isochrone.cc
```diff
@@ -49,13 +49,33 @@ Isochrone::Isochrone() : Dijkstras(), shape_interval_(50.0f) {
...
-void Isochrone::ConstructIsoTile(
-    const bool multimodal,
-    const unsigned int max_minutes,
-    const google::protobuf::RepeatedPtrField<valhalla::Location>& locations,
-    const sif::TravelMode mode) {
+void Isochrone::ConstructIsoTile(const bool multimodal,
+                                 const valhalla::Api& api,
+                                 const sif::TravelMode mode) {
+
+  // Extend the times in the 2-D grid to be 10 minutes beyond the highest contour time.
```
kevinkreiser, line 56: instead of computing the size of the tile outside of the isochrone api it made sense to bring all that logic in here right next to where it is needed

test/actor.cc
```diff
@@ -38,7 +38,7 @@ boost::property_tree::ptree make_conf() {
         "bus": {"max_distance": 5000000.0,"max_locations": 50,"max_matrix_distance": 400000.0,"max_matrix_locations": 50},
         "hov": {"max_distance": 5000000.0,"max_locations": 20,"max_matrix_distance": 400000.0,"max_matrix_locations": 50},
         "taxi": {"max_distance": 5000000.0,"max_locations": 20,"max_matrix_distance": 400000.0,"max_matrix_locations": 50},
-        "isochrone": {"max_contours": 4,"max_distance": 25000.0,"max_locations": 1,"max_time": 120},
+        "isochrone": {"max_contours": 4,"max_distance": 25000.0,"max_locations": 1,"max_time_contour": 120, "max_distance_contour":200},
```
kevinkreiser, line 41: the bulk of all the test changes were changes to the config values the only test that actually changed was GriddedData

test/gridded_data.cc
```diff
@@ -11,18 +11,22 @@ namespace {
 
 TEST(GriddedData, Basic) {
   // fill this as distance from center
-  GriddedData<PointLL> g({-5, -5, 5, 5}, 1, std::numeric_limits<float>::max());
-  for (int i = 0; i < 10; ++i) {
-    for (int j = 0; j < 10; ++j) {
-      Tiles<PointLL> t({-5, -5, 5, 5}, 1);
+  GriddedData<1> g({-7, -7, 7, 7}, 1, {std::numeric_limits<float>::max()});
```
kevinkreiser, line 14: i made this area a bit bigger so i could get all the rings in it

test/gridded_data.cc
```diff
@@ -31,44 +35,52 @@ TEST(GriddedData, Basic) {
...
   std::cout << "{\"type\":\"FeatureCollection\",\"features\":[";
-  for(const auto& feature : contours) {
-    std::cout << "{\"type\":\"Feature\",\"properties\":{\"iso\": " << feature.first << "},";
-    std::cout << "\"geometry\":{\"type\":\"MultiLineString\",\"coordinates\":[";
-    for(const auto& line : feature.second) {
-      std::cout << "[";
-      for(const auto& coord : line) {
-        std::cout << "[" << coord.first << "," << coord.second << "]" << (&coord != &line.back() ?
-  "," : "");
+  for (const auto& feature_collection : contours) {
```
kevinkreiser, line 56: this is still useful for debugging but it looked like it hadnt been updated for a long time (probably before the first implementation was finished). so i updated it to work with the current implementation

test/isochrone.cc
```diff
@@ -123,31 +123,31 @@ void try_isochrone(loki_worker_t& loki_worker,
...
...3259,52.075937],[5.048321,52.081429],[5.047708,52.078324],[5.051237,52.075937],[5.050321,52.072423],[5.053321,52.077735],[5.052834,52.074450],[5.055660,52.074598],[5.054992,52.077937],[5.057797,52.078461],[5.059321,52.081150],[5.060329,52.078937],[5.054919,52.070937],[5.060321,52.069593],[5.067613,52.070937],[5.066321,52.075319],[5.062321,52.076429],[5.059287,52.073903],[5.061002,52.077256],[5.069321,52.075623],[5.071556,52.076937],[5.071321,52.081586],[5.054198,52.081937],[5.050907,52.083937],[5.051799,52.085415],[5.049795,52.085463],[5.049321,52.083135],[5.048774,52.085390],[5.045321,52.085851],[5.047321,52.087985],[5.048321,52.086876],[5.054985,52.088273],[5.058046,52.090212],[5.058693,52.092937],[5.063321,52.092392],[5.063768,52.093937],[5.062321,52.096482],[5.060270,52.094886],[5.057742,52.096937],[5.059005,52.098621],[5.057204,52.098937],[5.060839,52.098937],[5.060634,52.103937],[5.058321,52.102771],[5.053207,52.103823],[5.049845,52.106937],[5.049421,52.111037],[5.034321,52.116532],[5.029321,52.113579],[5.031766,52.118382],[5.026321,52.120692],[5.023255,52.123937],[5.033321,52.118969],[5.040321,52.125974]],"type":"LineString"},"type":"Feature"}],"type":"FeatureCollection"})";
```
kevinkreiser, line 134: the only changes here is that now in the properties we serialize which type of contour it was time or distance

src/midgard/gridded_data.cc
```diff
@@ -1,330 +0,0 @@
-#include "midgard/gridded_data.h"
```
kevinkreiser, line 1: this was all moved to the header file. the bulk of this class is a single function

valhalla/midgard/gridded_data.h
```diff
@@ -100,14 +116,325 @@ template <class coord_t> class GriddedData : public Tiles<coord_t> {
    *
    * @return contour line geometries with the larger intervals first (for rendering purposes)
    */
-  contours_t GenerateContours(const std::vector<float>& contour_intervals,
+  contours_t GenerateContours(std::vector<contour_interval_t>& intervals,
                               const bool rings_only = false,
                               const float denoise = 1.f,
-                              const float generalize = 200.f) const;
+                              const float generalize = 200.f) const {
+    // sort the contours first on the metric index then on the values with the bigger contours first
```
kevinkreiser, line 123: this used to have to be done outside of this class and so was kind of like an unspoken rule. now its part of the contract of this method to have sorted output

valhalla/midgard/gridded_data.h
```diff
@@ -100,14 +116,325 @@ template <class coord_t> class GriddedData : public Tiles<coord_t> {
...
+    std::vector<contour_lookup_t> lookups(intervals.size());
+    // TODO: preallocate the lookups for each interval
+
+    int tile_inc[4] = {0, 1, this->ncolumns_ + 1, this->ncolumns_};
+    int case_value;
+    int case_table[3][3][3] = {{{0, 0, 8}, {0, 2, 5}, {7, 6, 9}},
+                               {{0, 3, 4}, {1, 3, 1}, {4, 3, 0}},
+                               {{9, 6, 7}, {5, 2, 0}, {8, 0, 0}}};
+    // For each metric we tracked
+    for (const auto& metric : metrics) {
```
kevinkreiser, line 162: this loop wasnt here before since there was only one metric, time. now we have this external loop to loop over all the metrics for which this grid was created.

src/loki/isochrone_action.cc
nilsnolde: sure, I thought so, tested it separately and it returned fine locally. well, that syntax then doesn't initialize both vars..

kevinkreiser:
I was able to reclaim the performance loss via two optimizations i performed in: https://github.com/valhalla/valhalla/pull/2699/commits/18d83d415d6808108db0388a42bbba3c5421e777

I'm happy to say that a test set of 20 minut isochrones for 10k locations takes 29.5 seconds on this branch as well as 29.5 seconds on master. I'll explain the optimizations in the diff

valhalla/midgard/gridded_data.h
```diff
@@ -100,14 +116,339 @@ template <class coord_t> class GriddedData : public Tiles<coord_t> {
...
+              // the contour (e.g. 1 minute higher).
+              // TODO - the value 1 is a bit of a hack.
+              float nd = data_[newtileid][metric_index];
+              s[m] = nd < max_value_[metric_index] ? nd - contour_value : 1.0f;
+              tile_corners[m] = this->Base(newtileid);
+              sh[m] = (s[m] > 0.0f) - (s[m] < 0.0f); // pos = 1, neg = -1, 0 = 0
+            }
+            s[0] = 0.25 * (s[1] + s[2] + s[3] + s[4]);
+            tile_corners[0] = this->Center(tileid);
+            sh[0] = (s[0] > 0.0f) - (s[0] < 0.0f); // pos = 1, neg = -1, 0 = 0
```
kevinkreiser, line 265: previously this bit of code had 3 branches not including the loop condition. i rolled 0th element out of the loop to remove one and i removed the other 2 branches by using math to find the sign of the number

valhalla/midgard/gridded_data.h
```diff
@@ -100,14 +116,339 @@ template <class coord_t> class GriddedData : public Tiles<coord_t> {
...
+    // up the given tile. this works out to a number of discrete cases which we lookup using the table
+    // below. based on the case we perform the appropriate intersection. to avoid branching we store
+    // the intersection operation for each case in an array and perform the correct one by calling the
+    // function stored in the array
+    int case_table[3][3][3] = {
+        {{0, 0, 8}, {0, 2, 5}, {7, 6, 9}},
+        {{0, 3, 4}, {1, 3, 1}, {4, 3, 0}},
+        {{9, 6, 7}, {5, 2, 0}, {8, 0, 0}},
+    };
+    std::array<std::function<void()>, 10> cases{
```
kevinkreiser, line 151: this was the other optimization. previously the below operations in the cases array were each an element of a switch case. now the call to them is a little strange looking below but we avoid all those branches in the tight loop

nilsnolde:
:exploding_head: really impressive work @kevinkreiser.. thanks a ton!

dnesbitt61:
With this request:
isochrone?json={"locations":[{"lat":39.63147339293369,"lon":-76.14091873168945}],"costing":"bicycle","contours":[{"time":30},{"time":60},{"time":90},{"time":120}],"polygons":true,"denoise":0.2,"generalize":150}
I am only getting the 30 minute contour in the response. The 60, 90, 120 contours are present, but empty (no coordinates).
I get similar behavior if I change the contours to distance (only distance=30 has coordinates).

nilsnolde:
Ok, I'll give up for today.. couldn't find the issue yet, but I'll share what I found anyways:
- my last commit (https://github.com/valhalla/valhalla/pull/2699/commits/814f7ec48056447cb3553562b2c0d0e62d27e04a) was working as expected, the earliest commit (which successfully returns isochrones) showing this bug is https://github.com/valhalla/valhalla/pull/2699/commits/9c34e0655fbc68e0df6bd124ced76df5a4135370
- some more empirics: it **only** happens for isochrones/-distances if the biggest contour is bigger by > 10 than the smallest contour, i.e. times of `60, 65, 69, 70` work perfectly while `60, 65, 69, 71` fails. That happens to be the time we add to the biggest contour to make sure the gridded data is not too small, so there's smth flawed in that logic:
https://github.com/valhalla/valhalla/blob/440aa215bf45b6aa380f543f77014d07563ca992/src/thor/isochrone.cc#L66
- probably related: just found out that requesting time and distance in one request (but separate contours of course..) leads to gridded data not being initialized properly: while centering the grid it can't find a tile for the center LL, i.e. ID of -1:
https://github.com/valhalla/valhalla/blob/440aa215bf45b6aa380f543f77014d07563ca992/src/thor/isochrone.cc#L143
which leads to division by zero here:
```gdb
0x000055555572dffb in valhalla::midgard::Tiles<valhalla::midgard::GeoPoint<double> >::Base (this=0x7fffc00354f0, tileid=-1)
    at /home/nilsnolde/dev/test/valhalla/valhalla/midgard/tiles.h:222
222         int32_t row = tileid / ncolumns_;
(gdb) where
#0  0x000055555572dffb in valhalla::midgard::Tiles<valhalla::midgard::GeoPoint<double> >::Base (this=0x7fffc00354f0, tileid=-1)
    at /home/nilsnolde/dev/test/valhalla/valhalla/midgard/tiles.h:222
#1  0x000055555572e302 in valhalla::midgard::Tiles<valhalla::midgard::GeoPoint<double> >::Center (this=0x7fffc00354f0, tileid=-1)
    at /home/nilsnolde/dev/test/valhalla/valhalla/midgard/tiles.h:255
#2  0x00005555559d92b2 in valhalla::thor::Isochrone::ConstructIsoTile (this=0x7fffccd0d488, multimodal=false, api=..., mode=valhalla::sif::TravelMode::kDrive)
    at /home/nilsnolde/dev/test/valhalla/src/thor/isochrone.cc:143
#3  0x00005555559d9aa2 in valhalla::thor::Isochrone::Compute (this=0x7fffccd0d488, api=..., graphreader=..., mode_costing=..., mode=valhalla::sif::TravelMode::kDrive)
    at /home/nilsnolde/dev/test/valhalla/src/thor/isochrone.cc:169
#4  0x00005555559c0c41 in valhalla::thor::thor_worker_t::isochrones[abi:cxx11](valhalla::Api&) (this=0x7fffccd0cb70, request=...)
    at /home/nilsnolde/dev/test/valhalla/src/thor/isochrone_action.cc:38
```

nilsnolde:
Last but not least, the biggest isochrones (and especially the isodistances) have **very** different geometries between your last commit and my last commit @kevinkreiser , sanity checking some distances manually they were correct(er) before (though hopefully also related to all the above stuff) :
10 & 20 km Minneapolis: -93.267913, 44.855274
tip of this branch:
image

https://github.com/valhalla/valhalla/pull/2699/commits/814f7ec48056447cb3553562b2c0d0e62d27e04a:
image

nilsnolde:
Tiny bug, but squashed all the weird observations above.. Didn't quite understand the logic inside the comp function, so I simplified it in a way that I can get it ;) problem was it didn't actually return the max element and if both time & distance were in the request, it would only ever set one of them (whichever came first in the request). feel free to change to smth more c++ish.

The `isochrone` test might not be very robust btw.. It failed locally for me with the current tip of this branch for some mysterious reason (but doesn't on CI..), so I plotted the [last isochrone test](https://github.com/valhalla/valhalla/blob/33b2428a7c100ed89ba01cdb3cb30c7273814075/test/isochrone.cc#L150) and to me it seems like `master` shouldn't pass actually, the geometries seem to have evolved quite a bit from what the test expects (all isochrones run from the test's utrecht data), the right-mostern shows that this branch and master a slightly different as well:
black: expected geometry, blue: current master, yellow: current tip of the branch
isochrone_test

the test does seem to care about the exactish coordinates, so I guess it's broken?

nilsnolde:
ugh.. failing after all.. I'll have a deeper look..

nilsnolde:
ah well, no idea.. running the last manually `./build/test/isochrone '{"locations":[{"lat":52.078937,"lon":5.115321}],"costing":"bicycle","costing_options":{"bicycle":{"service_penalty":0}},"contours":[{"time":15}],"show_locations":true}'` gives a few hundred coords back. But when run without a request, i.e. `./build/test/isochrone` it returns 10 coords.. and at the same time the identical request in `master` returns a quite different geometry than the test actually requires but it still passes :man_shrugging:

I'm outta here for today, trying to get my sanity back..

nilsnolde:
couldn't let it go, would've haunted my dreams.. so it's reeeeally weird:
the test **only** pass if I revert to https://github.com/valhalla/valhalla/pull/2699/commits/c2ea486aa2aeb236831f0026dc5720f5dbf5c0ec and apply these changes:
```diff
diff --git a/src/thor/isochrone.cc b/src/thor/isochrone.cc
index d09e8678b..8ffaf70a5 100644
--- a/src/thor/isochrone.cc
+++ b/src/thor/isochrone.cc
@@ -58,7 +58,7 @@ void Isochrone::ConstructIsoTile(const bool multimodal,
   // time in seconds is used when terminating the search. The + 10 minutes adds a buffer for edges
   // where there has been a higher cost that might still be marked in the isochrone
   float max_minutes(0.0f), max_km(0.0f);
-  for (const auto c : api.options().contours()) {
+  for (const auto& c : api.options().contours()) {
     if (c.has_time() && c.time() > max_minutes) {
       max_minutes = c.time();
     } else if (c.has_distance() && c.distance() > max_km) {
@@ -66,10 +66,12 @@ void Isochrone::ConstructIsoTile(const bool multimodal,
     }
   }
   bool has_time = max_minutes > 0.0f;
+  max_minutes = has_time ? max_minutes + 10.0f : std::numeric_limits<float>::min();
   bool has_distance = max_km > 0.0f;
+  max_km = has_distance ? max_km + 10.0f : std::numeric_limits<float>::min();
 
-  max_seconds_ = has_time ? (max_minutes + 10.0f) * kSecPerMinute : std::numeric_limits<float>::min();
-  max_meters_ = has_distance ? (max_km + 10.0f) * kMetersPerKm : std::numeric_limits<float>::min();
+  max_seconds_ = has_time ? max_minutes * kSecPerMinute : max_minutes;
+  max_meters_ = has_distance ? max_km * kMetersPerKm : max_km;
   float max_distance;
   if (multimodal) {
     max_distance = max_seconds_ * 70.0f * kMPHtoMetersPerSec;
```

Aside from _not understanding at all_ (!!!) why this change actually _changes anything_:

When I merge current master on top of these changes, the isochrone fails the same way as it does now. If I don't apply these changes and simply revert back to https://github.com/valhalla/valhalla/pull/2699/commits/c2ea486aa2aeb236831f0026dc5720f5dbf5c0ec **without merging master it also fails the same way**! WHAAAT? HOW?

Really sorry for so much noise.. But this is just ridiculous..

kevinkreiser:
via the sleuthing of @nilsnolde this is fixed. it turns out my implementation wrongly asumed that std::max_element wanted a greater than comparator but really, like all other stl predicates, it wanted a less than.

time contours:
image

distance contours:
image

not sure why 45km and 60km smash together before hitting up against the bank of the susquhanna, we should look into that

nilsnolde:
thanks @kevinkreiser ! 

Yeah, there's smth conceptually wrong with the dijkstra for isodistances.. I did some more experimenting and yeah, you were on the right track there.. The core problem for **isodistances** with the current approach is exactly that we do consider **both** max time **and** max distance as the termination for dijkstra:

according to the priority queue it's flying on low-cost edges to areas which are far away (i.e. long distance) along the low-cost path, then labels those low-cost/far-distance edges as permanent. then a much later branch of the priority queue travelling along the smaller roads (high-cost edges) comes to those (initially low-cost, far-distance) edges but actually the path that late branch took was a much shorter distance and it could actually travel much longer. it won't travel any further though to update the grid with lower distances because those edges were already labeled as permanent by the earlier low-cost, higher priority branch of the queue.. see this image for some visuals, with isodistances of 1, 7, 8, 9 and 10 km and some routes to show what I mean:
image

At the end of the routes the 7 & 8 km isodistances are way too close together. My guess is that the algo first went along the purple route and updated the last edges to have > 9 km distance. The orange, actually closer, but higher-cost path, to the same location has no chance to see that anymore since those edges are labeled permanent.. 

### Suggested fix

I'd suggest to simply not allow mixing isodistances and isochrones. Then we can tell the `ShouldExpand()` function which type it's dealing with and choose its termination criteria accordingly. It's (IMO) mostly a gimmick anyways and I can't really see how it could be possible to do so without harming performance majorly..

What do you think @kevinkreiser ?

kevinkreiser:
@nilsnolde i think there are still a couple of problems.

1. as you noted the metric by which expansion proceeds is `cost` which is most heavily impacted by time and to a lesser degree by distance. because of that changing the should expand creiteria alone wont work. we would need to change the costing to track distance. the good news here is that we already shipped that `"costing_options":{"auto":{"shortest": true`}}`
2. the semantics of the graph traversal are controlled by costing. what does the user want when they ask for the distance contour? do they want the absolute smallest distance routes contours or do they want contours for the routes the router would actually suggest. i think its arguable and based on use case to be honest.

my suggestion is to leave the code as is. this means that you get time contours and distance contours that are based on the lowest cost which is whatever you define it to be in your costing models options. people then have the option to use the `shortest` costing option to force the cost to be distance based and get shortest distance isodistances if that is what they want. does that make sense or am i missing something?

nilsnolde:
aaand kevin changed my mind again.. should've invested some time learning about the implementation of the priority queue.. alright, that settles it and should be good to go as-is @dnesbitt61

dnesbitt61 (approved):

Issue #2122, opened 2019-12-20
SIGSEGV in valhalla_service when using synchronized global cache


purew:
I'm seeing segmentation faults in `valhalla_service` when using `global_synchronized_cache` and more than 1 worker thread _and_ issuing requests in parallel:
```
2019/12/20 13:48:29.034697 [ANALYTICS] total_location_distance::0.144637km
2019/12/20 13:48:29.057249 [ANALYTICS] travel_mode::2
Thread 10 "valhalla_servic" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffd3fff700 (LWP 2415821)]
0x00005555557738db in valhalla::baldr::GraphReader::GetOpposingEdgeId(valhalla::baldr::GraphId const&, valhalla::baldr::GraphTile const*&) ()
(gdb) 
```
Backtrace
```
(gdb) bt
#0  0x00005555557738db in valhalla::baldr::GraphReader::GetOpposingEdgeId(valhalla::baldr::GraphId const&, valhalla::baldr::GraphTile const*&) ()
#1  0x000055555579ac57 in (anonymous namespace)::bin_handler_t::handle_bin(__gnu_cxx::__normal_iterator<(anonymous namespace)::projector_wrapper*, std::vector<(anonymous namespace)::projector_wrapper, std::allocator<(anonymous namespace)::projector_wrapper> > >, __gnu_cxx::__normal_iterator<(anonymous namespace)::projector_wrapper*, std::vector<(anonymous namespace)::projector_wrapper, std::allocator<(anonymous namespace)::projector_wrapper> > >) ()
#2  0x00005555557a003f in valhalla::loki::Search(std::vector<valhalla::baldr::Location, std::allocator<valhalla::baldr::Location> > const&, valhalla::baldr::GraphReader&, valhalla::sif::DynamicCost const*) ()
#3  0x000055555565fdf6 in valhalla::loki::loki_worker_t::route(valhalla::Api&) ()
#4  0x000055555565730f in valhalla::loki::loki_worker_t::work(std::__cxx11::list<zmq::message_t, std::allocator<zmq::message_t> > const&, void*, std::function<void ()> const&) ()
[...]

PR #2340, opened 2020-04-27
Make GraphTile own their memory

kkaefer:
`GraphTile`s currently own their memory if they're loaded from disk as plain files, or if they're compressed. However, uncompressed tiles in a tar file must hope that the underlying mapped memory won't get unmapped.

This PR moves towards a model where all `GraphTile` objects explicitly own their memory. Simultaneously, `GraphTile`s can't be copied anymore. ~The various *Cache objects now use `shared_ptr`s for ownership as well.~

## Tasklist

 - [ ] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [ ] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #2703, opened 2020-12-01
Handle GraphTile-s as shared pointers

averkhaturau:
# Issue

#2703

## Tasklist

 - [ ] Add tests

## Requirements / Relations

 https://github.com/valhalla/valhalla/issues/2122

### DIFF
--- a/valhalla/baldr/double_bucket_queue.h
+++ b/valhalla/baldr/double_bucket_queue.h
@@ -100,7 +100,7 @@ public:
     if (buckets_.size() < bucketcount) {
       buckets_.resize(bucketcount);
     }
-    // Empty the overflow bucket and each bucket
+
     overflowbucket_.clear();
 
     // Set the current bucket to the lowest cost low level bucket
--- a/valhalla/baldr/double_bucket_queue.h
+++ b/valhalla/baldr/double_bucket_queue.h
@@ -1,5 +1,4 @@
-#ifndef VALHALLA_BALDR_DOUBLE_BUCKET_QUEUE_H_
-#define VALHALLA_BALDR_DOUBLE_BUCKET_QUEUE_H_
+#pragma once
 
 #include <algorithm>
 #include <cmath>
@@ -30,6 +29,11 @@ using buckets_t = std::vector<bucket_t>;
 class DoubleBucketQueue final {
 public:
   /**
+   * Default c-tor creates empty object that needs to be initializd with reuse()
+   */
+  DoubleBucketQueue() = default;
+
+  /**
    * Constructor given a minimum cost, a range of costs held within the
    * bucket sort, and a bucket size. All costs above mincost + range are
    * stored in an "overflow" bucket.
@@ -44,6 +48,27 @@ public:
                     const float range,
                     const uint32_t bucketsize,
                     const LabelCost& labelcost) {
+    reuse(mincost, range, bucketsize, labelcost);
+  }
+
+  DoubleBucketQueue(DoubleBucketQueue&&) = default;
+  DoubleBucketQueue& operator=(DoubleBucketQueue&&) = default;
+  DoubleBucketQueue(const DoubleBucketQueue&) = delete;
+  DoubleBucketQueue& operator=(const DoubleBucketQueue&) = delete;
+
+  /**
+   * The same as c-tor, but without buffers reallocation
+   * @param mincost    Minimum cost. Used to create the initial range for
+   *                   bucket sorting.
+   * @param range      Cost range for low-level buckets.
+   * @param bucketsize Bucket size (range of costs within same bucket).
+   *                   Must be an integer value.
+   * @param labelcost  Functor to get a cost given a label index.
+   */
+  void reuse(const float mincost,
+             const float range,
+             const uint32_t bucketsize,
+             const LabelCost& labelcost) {
     // We need at least a bucketsize of 1 or more
     if (bucketsize < 1) {
       throw std::runtime_error("Bucketsize must be 1 or greater");
@@ -55,7 +80,7 @@ public:
     }
 
     // Adjust min cost to be the start of a bucket
-    uint32_t c = static_cast<uint32_t>(mincost);
+    const uint32_t c = static_cast<uint32_t>(mincost);
     currentcost_ = (c - (c % bucketsize));
     mincost_ = currentcost_;
     bucketrange_ = range;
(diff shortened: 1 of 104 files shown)


kevinkreiser:
@averkhaturau this seems to possibly be a duplication of the work @kkaefer did in #2340 any chance you might update that PR instead?

kevinkreiser:
So i just performance tested this and its just shy of 6% slower. I run a multithreaded service with 24 threads and run 14k routes through it. Master takes about 44 seconds to complete while this brange takes 46.5 seconds to complete. Ive done this over multiple runs and its quite consistant. In my opinion 6% slow down is not worth the convenience of shared_ptr, though i must admit i really like the idea of using shared_ptr (because the caller of GetGraphTile can become the last owner of the tile if he wants).

So I guess my main question is this, @averkhaturau will you be doing some performance analysis yourself to see if you can reduce the impact that this branch has on performance?

kevinkreiser (changes requested):
We need to do something to lessen the performance impact of these changes. I personally really like the change but 6% slowdown is borderline significant.

One potentially crazy idea would be to use a preprocessor directive that controls whether its raw or shared_ptr and then you can configure your build for one or the other. im not sure if its feesible without lots of code duplication though, it might be very difficult to work with.

im not sure if there is another place where we can steal back some performance to pay for this maybe?

kevinkreiser:
just perf tested this here are the results:

```
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ grep -F response_time master/* | sed -e "s/.* //g" | stats
0.001992702 0.036702991 0.124625230 0.706006947 1.581996679 
[1] 0.126362
> 
> 
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ grep -F response_time shared_ptr/* | sed -e "s/.* //g" | stats
0.003233671 0.040534735 0.134189987 0.742965057 1.551311493 
[1] 0.1321535
> 
> 
```

kkaefer:
Something like [`intrusive_ptr`](https://www.boost.org/doc/libs/1_74_0/libs/smart_ptr/doc/html/smart_ptr.html#intrusive_ptr) could be an option here. Instead of allocating a separate control block, the control block is part of the object (so it needs to be aware of it) at the expense of not being able to create `weak_ptr`s (which this PR doesn’t seem to use anyway).

Another optimization might be to use shared pointer implementation that isn’t thread-safe. The STL implementation _is_ thread-safe and uses atomics to guarantee correctness. While this is great default behavior, we could sacrifice thread-safety of `GraphTile` pointers and use [non thread-safe variant](https://www.boost.org/doc/libs/1_74_0/libs/smart_ptr/doc/html/smart_ptr.html#local_shared_ptr).

Both of these approaches can be combined as well; e.g. [Boost’s `intrusive_ptr`](https://www.boost.org/doc/libs/1_74_0/libs/smart_ptr/doc/html/smart_ptr.html#intrusive_ref_counter) supports both thread-safe and -unsafe reference counting strategies.

valhalla/baldr/double_bucket_queue.h
```diff
@@ -66,8 +66,11 @@ class DoubleBucketQueue final {
     maxcost_ = mincost_ + bucketrange_;
 
     // Allocate the low-level buckets
-    size_t bucketcount = (range / bucketsize_) + 1;
+    const size_t bucketcount = (range / bucketsize_) + 1;
     buckets_.resize(bucketcount);
+    for (auto& b : buckets_) {
+      b.reserve(bucketsize);
```
averkhaturau, line 72: Not sure about bucketsize here.

kevinkreiser:
latest commit shows:

```
0.003687143 0.040842295 0.137289381 0.761976640 1.600770235 
[1] 0.1358664
```

averkhaturau:
Current performance experiment scenario:
1. Build valhalla_service
2. Download planet tiles `aws s3 cp s3://valhalla-production-eu-west-1/valhalla/planet_2020_12_01-03_00_00.tar . `,
3. Create valhalla service config with command like `../scripts/valhalla_build_config --mjolnir-tile-dir $(pwd)/valhalla_tiles_PA --mjolnir-tile-extract $(pwd)/planet_2020_12_01-03_00_00.tar --mjolnir-timezone ~/work/data/time_zone.sqlite --mjolnir-admin ~/work/data/admin_2020_09_20-03_00_00.sqlite > valhalla-PA.json`
4. Run valhalla_service with maximum available concurrency: `./valhalla_service ../test/valhalla.json 24`
5. From another console run requests `valhalla/run_route_scripts$ time ./run_with_server.py --test-file auto.txt --url http://localhost:8002/route --concurrency 24 --format json`. Here auto.txt is a set of route requests, 
[auto.txt](https://github.com/valhalla/valhalla/files/5664884/auto.txt)
6. `alias stats='R -q -e "x <- read.csv(\"stdin\", header = F); quantile(x[ ,1] , c(0, .5, .90, .99, 1)); sd(x[ , 1])"'`
7. `grep -F response_time 20201207_235004_auto/* | sed -e "s/.* //g" | stats` will output request - response time statistics. You have to install R to run this.

For different experiments repeat p.5 and p.7. '20201207_235004_auto' is an example of folder created by './run_with_server.py'

kevinkreiser:
I think our best bet here is to break out another PR where we try to find performance elsewhere. My suggestion would be to try to reduce allocations in the double bucket queue. Basically we call clear on the vectors there but we dont actually deallocate. This means that we will need to set an upper limit to resize back down to if an individual long route exceeds it but i think its ok for crazy long routes to hit this limit and pay the penalty.

mskurydin:
I also tried a bunch of performance runs on my local machine (same conditions as @kevinkreiser mentioned above). Also seeing a slowdown of 2.5-4%, but at the same time seeing a wide spread of the results both on master and on this branch (115 seconds vs 120 seconds). I also noticed that other processes on the machine might have a noticeable impact (not sure if it's CPU/memory or IO related). Can we reliably use this type of benchmarking to determine the exact impact of the changes on the performance?

Also observed https://github.com/valhalla/valhalla/pull/2719 shows from 5% to 9% improvement compared to the master.

kevinkreiser:
yeah i didnt mention my exact methodology but its approximatelyi as follows:

1. close everything (services are still running like ssh but i let those alone)
2. open a terminal, in one tab run valhalla_service in the other run the bench mark
3. run the benchmark 3 times without restarting the server taking the best score

its not the most scientific thing in the world but on my machine, runs of master do not fluctuate more than a half of a percent or something (excluding the first run with colder cache).

please understand im not just trying to be a pain in the butt for the sake of it. im trying to avoid an increase in latency across the board where all requests will suffer. for the individual user on the individual request they will not likely notice 5% especially when the route is something like 40ms to compute, but when you are running this in a server environment paying for the convenience of shared_ptr really has to be worth it, because that payment either means more machines or a looser gaurantee on latency/throughput. 

i really like @kkaefer suggestion about using a lighter weight auto_ptrs from boost. can you guys try that out or mention why such a pointer wouldnt fit your needs? hopefully that will bring the perf difference into the 1% range and we can merge it and move on.

kevinkreiser:
we went with another approach #2730

PR #2705, opened 2020-12-02
Fix bidirectional route failures at deadends

danpaz:
# Issue

This PR addresses two issues in bidirectional a* where the search paths in the forward and reverse directions fail to connect on deadend edges. This usually comes up when an intermediate waypoint is set on a deadend road with break_through set, which forces the route to start on an edge heading toward the deadend and requiring that the route make a uturn in order to leave the deadend.

## 1. Fix for deadends

The first failure scenario has to do with how bidirectional decides to prune not_thru edges. The changes in https://github.com/valhalla/valhalla/commit/6e0682b7f81f70c07099e0e92eb224dbb8df9257 allow the algorithm to enqueue the uturn edge in the case where the only other edge added was a not_thru.

```
Let's say we have a route that travels into a small not_thru region from the left towards A

...A--B--C

where
AB and BC are not_thru=true

And in order to complete the route we need to make a uturn out of the not_thru region.

Consider only the forward expansion, starting with predecessor = AB.

Logic before this PR:
1. Expand from B
2. Add BC
3. ExpandForwardInner returns true
4. Move on to next iteration
5. Expansion ends because BC is not expandable (not_thru)
(therefore uturn BA is never added)

Logic after this PR:
1. Expand from B
2. Add BC
3. ExpandInner returns false
4. Add BA (u-turn edge)
5. Expansion continues on A

```

## 2. Fix for deadends with a barrier

https://github.com/valhalla/valhalla/commit/888698743e6fb5d508fe91a08c398f95ada03cdb fixes an issue where expansion is immediately stopped when the node we're expanding from has an access restriction, such as a `barrier=block`. In the worst case this issue caused no route to be returned if a uturn at the barrier was the only option to escape the deadend.

```
A--B--C

where

C is barrier=block

The costing's Allowed(node) function returns false causing expansion to stop at this node.

In this PR we now expand from the opposing edge (do a uturn) if we encountered a barrier.

```

# Testing

**RAD**: this change affected 0.5% of user routes, and the changes there were caused by selecting a similar-cost alternative (within <1% in terms of route cost). Some example routes that are fixed by this change are in test_requests/deadends.txt.

**Performance**:  @kevinkreiser ran a performance benchmark and found no difference in total time to calculate routes.

<details>
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ time ./run_with_server.py --test-file auto.txt --url http://localhost:8002/route --concurrency 24 --format csv
Placing 13797 results in 20201202_152034_auto
17%
^[[A
^[[A29%
38%
48%
59%
71%
82%
93%
99%
100%
real	0m45.659s
user	0m28.267s
sys	0m2.361s
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ time ./run_with_server.py --test-file auto.txt --url http://localhost:8002/route --concurrency 24 --format csv
Placing 13797 results in 20201202_152120_auto
17%
31%
42%
52%
63%
73%
86%
96%
100%
real	0m44.577s
user	0m28.001s
sys	0m2.680s
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ time ./run_with_server.py --test-file auto.txt --url http://localhost:8002/route --concurrency 24 --format csv
Placing 13797 results in 20201202_152204_auto
17%
31%
41%
52%
63%
74%
86%
96%
100%
real	0m44.365s
user	0m28.268s
sys	0m2.442s
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ 
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ echo BRANCH
BRANCH
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ time ./run_with_server.py --test-file auto.txt --url http://localhost:8002/route --concurrency 24 --format csv
Placing 13797 results in 20201202_152358_auto
^[[A
^[[A
17%
31%
42%
52%
63%
74%
86%
96%
100%
real	0m43.977s
user	0m28.140s
sys	0m2.489s
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ time ./run_with_server.py --test-file auto.txt --url http://localhost:8002/route --concurrency 24 --format csv
Placing 13797 results in 20201202_152442_auto
17%
31%
42%
52%
63%
74%
86%
96%
100%
real	0m44.379s
user	0m28.215s
sys	0m2.469s
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/run_route_scripts$ time ./run_with_server.py --test-file auto.txt --url http://localhost:8002/route --concurrency 24 --format csv
Placing 13797 results in 20201202_152526_auto
17%
31%
42%
52%
64%
74%
86%
96%
100%
real	0m44.076s
user	0m28.054s
sys	0m2.490s

</details>

## Tasklist

 - [X] Add tests

## Requirements / Relations


src/thor/bidirectional_astar.cc
```diff
@@ -123,21 +123,28 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
...
   // Update the time information even if time is invariant to account for timezones
   auto seconds_offset = invariant ? 0.f : pred.cost().secs;
   auto offset_time = from_transition
                          ? time_info
                          : time_info.forward(seconds_offset, static_cast<int>(nodeinfo->timezone()));
 
-  uint32_t shortcuts = 0;
-  EdgeMetadata meta = EdgeMetadata::make(node, nodeinfo, tile, edgestatus_forward_);
+  // If we encounter a node with an access restriction like a barrier we allow a uturn
+  if (!costing_->Allowed(nodeinfo)) {
```
danpaz, line 137: This change handles the case where we encounter a disallowed node and instead of bailing we now want to consider making a uturn.

src/thor/bidirectional_astar.cc
```diff
@@ -297,7 +311,10 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
   if (expansion_callback_) {
     expansion_callback_(graphreader, "bidirectional_astar", meta.edge_id, "r", false);
   }
-  return true;
+
+  // we've just added this edge to the queue, but we won't expand from it if it's a not-thru edge that
+  // will be pruned. In that case we want to allow uturns.
+  return !(pred.not_thru_pruning() && meta.edge->not_thru());
```
danpaz, line 317: This change handles the case where we continue to enqueue not_thru edges like we did before, but we return false _if_ we're about to prune this edge. This allows the calling function to enqueue a u-turn edge in addition to the not_thru edge.

src/thor/isochrone.cc
```diff
@@ -285,7 +285,7 @@ void Isochrone::ExpandingNode(baldr::GraphReader& graphreader,
   UpdateIsoTile(current, graphreader, node->latlng(tile->header()->base_ll()), secs0);
 }
 
-ExpansionRecommendation Isochrone::ShouldExpand(baldr::GraphReader& graphreader,
+ExpansionRecommendation Isochrone::ShouldExpand(baldr::GraphReader& /*graphreader*/,
```
danpaz, line 288: clang-tidy nit

src/tyr/route_serializer_valhalla.cc
```diff
@@ -122,8 +125,9 @@ json::MapPtr summary(const valhalla::Api& api) {
   }
 
   auto route_summary = json::map({});
-  route_summary->emplace("time", json::fp_t{time, 3});
-  route_summary->emplace("length", json::fp_t{length, 3});
+  route_summary->emplace("time", json::fp_t{route_time, 3});
+  route_summary->emplace("length", json::fp_t{route_length, 3});
+  route_summary->emplace("cost", json::fp_t{route_cost, 3});
```
danpaz, line 130: Added `cost` to the summary part of the response. This was helpful to add in this PR to compare the costs of routes that changed in RAD tesing.

kevinkreiser:
> Let's say we have a route that travels into a small not_thru region from the left towards A

Its important to clarify this a bit I think for full understanding. It's not that we travel into a small `not_thru` region, its that the path starts out on an edge that *ISNT* a `not_thru` but its end node is only connected to `not_thru` edges. In this case we start out saying ok cool we will be pruning `not_thru` edges and so the search will immediately stop because thats all there is ahead of the start edge. one might say, WTF, why isnt the start edge classified as `not_thru` and the person saying that has a good point, so we should fix this as well in mjölnir. Anyway in this case we need to mark the start edge as a derived dead end and allow the uturn.

src/thor/bidirectional_astar.cc
kevinkreiser: the fallout from this is that we will enqueue more edges but practically speaking it doesnt really affect the number of iterations we run. so basically we'll have more labels but do pretty much the same amount of expansion

src/tyr/route_serializer_valhalla.cc
kevinkreiser: we should do this for the leg summary as well on line 327

danpaz: Done

kevinkreiser (approved):

dnesbitt61:
Will this now always allow a Uturn at the end of each initial not_thru edge? If so that will mess up routes that rely on not_thru pruning to avoid doing Uturns at cul-de-sacs

kevinkreiser:
@dnesbitt61 what this will do is allow a uturn of the edge *before* the not_through region, so the cul-de-sac uturn will still be avoided if its marked as not_through (which it very likely is)

dnesbitt61:
thanks. Tested this with: http://localhost:8002/route?json={"locations":[{"lat":39.650855,"lon":-76.052863},{"lat":39.653259,"lon":-76.049555,"type":"through"},{"lat":39.677466,"lon":-76.049938}],"costing":"bicycle","costing_options":{"bicycle":{"bicycle_type":"Road","use_roads":"0.5","cycling_speed":"25.0","use_hills":"0.5"}},"directions_options":{"language":"en-US"} and it works as before (continues on through roads and avoids a uturn).

dnesbitt61 (approved):
Tested performance with a long bicycle route and didn't make an appreciable difference (<2% time diff).

kevinkreiser (approved):

danpaz:
Thanks for testing @dnesbitt61 !

soheilarmin:
Hi,

Given the attached OSM, routing from [35.767956, 51.456357] to [35.767818, 51.457776] seems to be problematic. Actually it does not work as expected in many case. It seems that I the routing breaks at every node involving a turn restriction. 

[sample.zip](https://github.com/valhalla/valhalla/files/5720577/sample.zip)

PR #2278, opened 2020-03-18
Unify Time Tracking in all the Algorithms

kevinkreiser:
# Issue

This PR seeks to unify the way we track time during any given algorithms expansion in the graph. At the moment we use a bunch of the same patterns in all the different algorithms, and it would be great to encapsulate that into a single set of functionality that all the different code paths can re-use.

The eventual goal of this is to do bidirectional a* with time dependence but that functionality will be a completely separate PR. this is just the first steps

## Tasklist

 - [x] create and use struct in all algorithms
 - [x] add struct unit tests
 - [x] add integration (route) unit tests
 - [ ] Update the [changelog](CHANGELOG.md)

PR #2706, opened 2020-12-02
remove astar since its redundant with timedep_forward

kevinkreiser:
its annoying to maintain different algorithms for different occasions. depending on the options in your request, we choose which an algorithm out of 6 that are used for A to B routing. one of the distinctions in your request that makes it so that we use one algorithm over another is the presence of "time". if you want a time aware route we have a class of algorithms that does that, and another class that doesnt. in #2660 we added a form of time dependence to bidirectional a*, and we are trying to consolidate even further. to that end, this pr completely removes the regular (time independent) a* algorithm. this algorithm is redundent with the timedependent forward a* algorithm. a semi-recent refactor to the way we track time #2278 in our algorithms makes it so that the only thing stopping us is a single if that says, reject routes which dont have time at the origin. so i've removed that check and will make astar pure virtual.


src/thor/timedep_forward.cc
```diff
@@ -1,9 +1,9 @@
 #include "baldr/datetime.h"
+#include "baldr/graphconstants.h"
```
kevinkreiser, line 2: everything here was stuff that astar was filling in for timedep_forward. instead of keeping astar around as pure virtual i just moved its guts to here.

test/gurka/test_search_filter.cc
```diff
@@ -14,10 +14,10 @@ class SearchFilter : public ::testing::Test {
 
     const std::string ascii_map = R"(
```
kevinkreiser, line 15: these test changes are needed because the previous test was falling back to astar which didnt implement the derived uturn logic. once it started usting timedep_forward which does have this logic, it was allowing a uturn at C to come back to reach 2 (the original destination). this was apparently slightly cheaper than going around the block. to keep the spirit of the test going aroudn the block, i moved the origin/destination and changed the road class to lower the speed

valhalla/thor/timedep.h
```diff
@@ -1,8 +1,22 @@
-#ifndef VALHALLA_THOR_TIMEDEP_H_
-#define VALHALLA_THOR_TIMEDEP_H_
+#pragma once
```
kevinkreiser, line 1: again, this class derived from astar and used a lot of its methods. so ive copied them over here

src/thor/route_action.cc
```diff
@@ -336,7 +325,8 @@ std::vector<std::vector<thor::PathInfo>> thor_worker_t::get_path(PathAlgorithm*
 
     path_algorithm->Clear();
     cost->set_pass(1);
-    bool using_astar = (path_algorithm == &astar);
+    // TODO: is this a proxy for trivial route?
+    bool using_astar = (path_algorithm == &timedep_forward || path_algorithm == &timedep_reverse);
```
danpaz, line 329: Do we need to figure out this TODO before merging?

danpaz (dismissed):
Happy to have one less algorithm to maintain 👍 

One question below, otherwise LGTM.

danpaz (approved):

PR #2711, opened 2020-12-07
Recost shortcuts in bidastar: second approach

genadz:
## Tasklist

 - [x] Add tests

## Requirements / Relations


src/thor/bidirectional_astar.cc
```diff
@@ -1053,102 +1069,67 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
     }
 
     // Reverse the list
-    std::reverse(path.begin(), path.end());
-
-    // Special case code if the last edge of the forward path is the destination edge
-    // which means we need to worry about partial distance on the edge
-    if (edgelabels_reverse_[idx2].predecessor() == kInvalidLabel) {
```
kevinkreiser, line 1060: i know its a bit heavy to use recosting to fix this problem, but it feels really good to be able to remove this anyway :smile:

src/sif/recost.cc
```diff
@@ -121,6 +123,12 @@ void recost_forward(baldr::GraphReader& reader,
     // construct the label
     label = EdgeLabel(predecessor++, edge_id, edge, cost, cost.cost, 0, costing.travel_mode(), length,
                       transition_cost, time_restrictions_TODO);
+    // Set 'deadend' flag true only to allow to allow u-turns in DynamicCosting::Allowed method.
+    // In bidiastar algorithm 'deadend' flag can be set dynamically based on current costing and
+    // time restrictions, but we don't want to add additional logic here in order not to drop
+    // performance. So just use simple hack: set this flag true and thus allow u-turns.
+    // TODO: think about more explicit solution.
```
kevinkreiser, line 130: we can remove this and instead above we can do this:

```diff
diff --git a/src/sif/recost.cc b/src/sif/recost.cc
index 6e740a9a9..cb6e03bb1 100644
--- a/src/sif/recost.cc
+++ b/src/sif/recost.cc
@@ -72,17 +72,20 @@ void recost_forward(baldr::GraphReader& reader,
       throw std::runtime_error("Node cannot be found");
     }
 
-    // this node is not allowed
-    if (node && !costing.Allowed(node)) {
-      throw std::runtime_error("This path requires different node access than this costing allows");
-    }
-
     // grab the edge
     edge = reader.directededge(edge_id, tile);
     if (!edge) {
       throw std::runtime_error("Edge cannot be found");
     }
 
+    // re-derive uturns, would have been nice to return this but we dont know the next edge yet
+    label.set_deadend(label.opp_local_idx() == edge->localedgeidx());
+
+    // this node is not allowed, unless we made a uturn at it
+    if (node && !label.deadend() && !costing.Allowed(node)) {
+      throw std::runtime_error("This path requires different node access than this costing allows");
+    }
+
```

src/thor/bidirectional_astar.cc
```diff
@@ -1053,102 +1069,67 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
...
+      } else
+        path_edges.emplace_back(std::move(opp_edge_id));
 
       // Check if this is a ferry
       if (edgelabel.use() == Use::kFerry) {
         has_ferry_ = true;
       }
+    }
+
+    // don't care about trivial paths here, it will be handled correctly in recost function below
```
kevinkreiser, line 1096: the comment is wrong here. there is an existing bug with bidirectional a* where routes that start and end on the same edge fail when they need to go around the block.

```suggestion
    // bidirectional a* has a bug where it fails trivial routes in which you are on a one way edge and
    // the origin is near the end of the edge and the destination is near the beginning, in other words
    // a route that looks trivial but actually needs to go around the block to complete
```

test/astar.cc
```diff
@@ -1617,6 +1631,120 @@ TEST(Astar, BiDirTrivial) {
   EXPECT_LT(path.front().elapsed_cost.secs, 1);
 }
 
+TEST(BiDiAstar, test_recost_path) {
```
kevinkreiser, line 1634: one other option here would have been to subclass the bidirectional a* class and then call form path with some preconstructed labels that use the shortcut. then you can always be sure that the test is actually confirming that FormPath is undoing any labels with shortcuts. dont worry about making this change unless you really want to, just an observation from me

genadz: if you don't mind I will leave as is. But I completely agree, in unittests it's better to test smaller part than whole pipeline

src/sif/recost.cc
```diff
@@ -72,17 +72,21 @@ void recost_forward(baldr::GraphReader& reader,
...
-    }
-
     // grab the edge
     edge = reader.directededge(edge_id, tile);
     if (!edge) {
       throw std::runtime_error("Edge cannot be found");
     }
 
+    // re-derive uturns, would have been nice to return this but we dont know the next edge yet
+    if (label.opp_local_idx() == edge->localedgeidx())
```
kevinkreiser, line 82: why add a branch to this code path, we simply store the value of this comparison?

src/thor/alternates.cc
```diff
@@ -58,7 +58,8 @@ void filter_alternates_by_stretch(std::vector<CandidateConnection>& connections)
 
 // Limited Sharing. Compare duration of edge segments shared between optimal path and
 // candidate path. If they share more than kAtMostShared throw out this alternate.
-bool validate_alternate_by_sharing(GraphReader& graphreader,
+// Note that you should recover all shortcuts before call this function.
+bool validate_alternate_by_sharing(GraphReader& /*graphreader*/,
```
kevinkreiser, line 62: can you actually just remove the graphreader argument completely, now that we dont ahve to recover shortcuts its not used.

kevinkreiser (changes requested):
two things:

1. remove the branch in recosting
2. remove the graphreader arg in path sharing

src/sif/recost.cc
genadz: because by default we use `deadend` flag from the edge

genadz: hm. from the other hand, if it's deadend, this condition will be always true

kevinkreiser:
well crap... I performance tested this and saw that it added approximately a 25% slowdown. assuming it must be the shortcut recovery. i did this little bit of work: #2714 after testing with that enabled on this branch it was just as slow :smile: 

so i suspect some things..

1. i disabled recovering shortcuts on this branch and it was still about 12% slower, which must mean that recosting is slower
2. since having a cache of recovered shortcuts didnt speed it up it must be a function of having so many edges in the route, so maybe its in triplegbuilder or other serializers.

i'll look a bit closer to figure out where the performance drop lies

kevinkreiser:
Ok did a bit more digging. First I enabled the shortcut cache. Then I took the variable `path_edges` and made it a member variable `path_egdes_` (dont forget to clear it in the clear function). Then I wrote a bash alias like so: 

```bash
alias stats='R -q -e "x <- read.csv(\"stdin\", header = F); quantile(x[ ,1] , c(0, .5, .90, .99, 1)); sd(x[ , 1])"'
```

It seemed to me that the worst offenders in terms of routes were driving up my benchmarking scores to the point that i couldnt really make sense of it. Non insane routes were almost unaffected by this change. Anyway to prove out this thought I ran up my server with master. And did a run of RAD but with the json output format:

```bash
./run_with_server.py --test-file auto.txt --url http://localhost:8002/route --concurrency 24 --format json
```

then i did the same but ran this branch with my small changes added. then i compared the percentiles between the two:

master:
```bash
grep -F response_time 20201207_235004_auto/* | sed -e "s/.* //g" | stats
0.002395153 0.036434412 0.122441101 0.709217098 1.643116236 
[1] 0.1298356
> 
> 
```

this branch with my slight modifications:
```bash
grep -F response_time 20201207_235232_auto/* | sed -e "s/.* //g" | stats
0.002358675 0.038644910 0.150891209 0.788010578 1.731230021 
[1] 0.1442929
> 
> 
```

You can see that for the average case the performance is nearly unchanged where as the performance of the top 10 percent of the slowest routes is something like 10-15% worse. @genadz i tried to update your other branch where we dont recost the whole route but rather just the parts that need it as they are recovered. can you use the same testing method to see what kind of results you get?

genadz:
on my machine I got the following results:

master
```
grep -F response_time res_master/* | sed -e "s/.* //g" | stats 
0.003356218 0.108471870 0.408698320 0.995351839 2.934854984 
[1] 0.2041199
> 
> 
```
recost_shortcuts_new (second approach)
```
grep -F response_time res_recost/* | sed -e "s/.* //g" | stats 
0.00315094 0.13643813 0.55678535 1.43106830 3.76190114 
[1] 0.2874338
> 
> 
```
recost_shortcuts (first approach)
```
grep -F response_time res_first/* | sed -e "s/.* //g" | stats 
0.003399849 0.132326603 0.479046679 1.087140005 3.063050032 
[1] 0.2249207
> 
> 
```

hm, that's interesting. despite the fact that times for the first approach are closer to master than second approach, I got approximately the same total time for both recosting branches.

kevinkreiser:
first thing i would do is figure out which routes are failing and fix them in both branches so that at least correctnesswise the code is shippable. then we'll have to ponder about performance.

genadz:
fixed first branch. it's ~15% slower than master

recost_shortcuts (first approach)
```
grep -F response_time res_first/* | sed -e "s/.* //g" | stats
0.00292778 0.07737732 0.49527001 1.44068127 5.46187663 
[1] 0.2961684
> 
> 
```
master
```
grep -F response_time res_master/* | sed -e "s/.* //g" | stats
0.002770424 0.073756933 0.417046785 1.102108717 5.447159052 
[1] 0.2343167
> 
> 
```

genadz:
To check hypothesis of performance drop I excluded superseded edges from the final path and left shortcuts as in master; but the same time I didn't remove the code that recovers shortcuts. In other words I commented these https://github.com/genadz/valhalla/blob/recost_shortcuts_new/src/thor/bidirectional_astar.cc#L1061-L1062 lines and these https://github.com/genadz/valhalla/blob/recost_shortcuts_new/src/thor/bidirectional_astar.cc#L1086-L1087 . So, I got results (total time) similar to `master` branch. Here is response times:

modified **recost_shortcuts_new** (total time: `5m 5s`)
```
grep -F response_time res_recost_no_short_2/* | sed -e "s/.* //g" | stats
0.002896309 0.072885752 0.404890060 1.218031406 4.106480122 
[1] 0.2491634
```

**master** (total time: `4m 59s`)
```
grep -F response_time res_master/* | sed -e "s/.* //g" | stats
0.002793074 0.072911501 0.405596733 1.062811947 6.334811926 
[1] 0.2267943
```
@kevinkreiser it helps me to make an obvious conclusion: we loose performance in the later stages (triplegbuilder or so)

kevinkreiser:
@genadz yep i did the same experiment a few weeks ago and came to the same conclusion, do you have the stomache to go looking at triplegbuilder to see where it is slowest? i can tell you for a fact that our json serialization is slow. we should convert the route serializer to use rapidjson thats our best bet for performance there. that said though in that other branch i made where i measured the timings i seem to remember that the bulk was still in thor which would mean triplegbuilder

genadz:
ok, will look into triplegbuilder.

genadz:
@kevinkreiser you're right, if we optimize serialization it will not speed up much. I logged timings for some workers/functions and calculated some statistics (times in milliseconds):

**master**
```
percentiles	50%	90%	95%	100%	avg
loki::request	19.0	133.0	204.0	1187.0	49.0
get_path	25.0	165.0	301.0	5154.0	70.0
triplegbuilder	5.0	83.0	134.0	848.0	28.0
thor::route	36.0	257.0	427.0	5826.0	99.0
odin::serialize	3.0	10.0	15.0	83.0	4.0
```

**recost_shortcuts_new**
```
percentiles	50%	90%	95%	100%	avg
loki::request	21.0	146.0	218.0	1254.0	54.0
get_path	27.0	202.0	391.0	3863.0	86.0
triplegbuilder	5.0	129.0	261.0	2398.0	52.0
thor::route	39.0	356.0	668.0	4747.0	139.0
odin::serialize	3.0	14.0	25.0	108.0	6.0
```
- serialization became slower but it's a small part from the whole pipeline; the greatest deceleration in `triplegbuilder`, then `get_path`.

genadz:
@kevinkreiser this guy https://github.com/valhalla/valhalla/blob/master/src/thor/triplegbuilder_utils.h#L438 consumes most of the time in triplegbuilder

dnesbitt61:
That makes sense since great pains were made to add information so that lower level (local and arterial) tiles would not have to be accessed on long routes (except for near the origin and destination). Recovering shortcuts would lead to bringing in more tiles than are required to actually generate the path. I haven't followed the whole process of recovering shortcuts, but the main rationale for hierarchies and shortcuts was so that these lower level tiles would not need to be accessed.

 If code is triggered inside this loop:
// Add intersecting edges on different levels (follow NodeTransitions)
  if (node->transition_count() > 0) {
....
}
then I can see triplegbuilder taking much more time than before.

kevinkreiser:
@genadz @dnesbitt61 there are two ways to solve this then:

1. if its just accessing the graph on different levels that is causing the performance drop we can mark PathInfos with a "recovered" boolean to tell this method not to jump down levels to get intersecting edges, though this does kill some of the benefits of recovering shortcuts
2. if its just the fact that we are more often calling the method `AddTripIntersectingEdge` then maybe we can optimize it instead. what happens to performance when you comment out that method call. ie its just the jumping around of tile levels but not actually setting the data in the pbf?

genadz:
@kevinkreiser @dnesbitt61
After I commented out `AddTripIntersectingEdge` it became much faster, so yes, performance drops due to accessing the graph on lower levels.
Then I implemented the idea with not accessing lower levels for recovered edges. It also improved performance a lot, but still a little slower than master (~5%). 
```
grep -F response_time res_recost/* | sed -e "s/.* //g" | stats
0.002701759 0.073516130 0.426355600 1.261174417 3.951776743 
[1] 0.2577073
```
@kevinkreiser is it acceptable ? imho it's enough, and we can try more optimization tricks in a separate PR.

**One more important thing:** we have 126 "route not found" errors on this branch (76 on `master`). At least some of them happen because of this check https://github.com/genadz/valhalla/blob/recost_shortcuts_new/src/sif/recost.cc#L81 : node is rejected by access mode. I see several options here:
1. add new restriction to shortcut builder: all nodes along the shortcut must have the same access mode;
2. pass `recover` flag to recosting function and skip restrictions for such edges/nodes;
3. recover shortcuts inside of recosting function and , the same as in 2., skip restrictions for all superseded edges/nodes;
4. skip restrictions check at all because we already checked all of them in path algorithm; so care about only new costs.

I like the last one))

kevinkreiser:
@genadz i agree, we should do number 4. originally when i wrote the recosting i had assumed we'd use it for this but the first use-case i actually used it for was to get costing information for a different costing. in that case its not such a big deal when its not available but here its a show stopper. id rather recosting be permissive for both node and edge access so that we can get the information. i guess in that case we may as well just send a boolean to the function that says wither or not we should care about allowed?

src/thor/triplegbuilder.cc
```diff
@@ -943,12 +943,9 @@ void AccumulateRecostingInfoForward(const valhalla::Options& options,
...
   // setup a callback for the recosting to get each edge
-  auto in_itr = leg.node().begin();
-  sif::EdgeCallback edge_cb = [&in_itr]() -> baldr::GraphId {
-    auto edge_id = in_itr->has_edge() ? baldr::GraphId(in_itr->edge().id()) : baldr::GraphId{};
-    ++in_itr;
-    return edge_id;
-  };
+  std::vector<GraphId> edge_ids;
+  for (auto in_itr = leg.node().begin(); in_itr->has_edge(); ++in_itr)
+    edge_ids.emplace_back(in_itr->edge().id());
```
kevinkreiser, line 948: if switching from the callback model didnt help performance i would suggest we revert that change. do you konw if that change made a difference in performance?

kevinkreiser:
@genadz it looks liek you didnt push you latest changes up right? one thing i was trying to say with my second to last comment was that if you **ONLY** commented out the `AddTripIntersectingEdge` method then it was still doing the jumping between tiles/levels, whcih means its actually the work that its doing in that method to copy the attribution that is slow rather than the jumping around the graph. i have an idea to get a quick win for performance there. let me try an experiment...

kevinkreiser:
@genadz i tried an experiment here: https://github.com/valhalla/valhalla/tree/kk_reserve_pbf which basically focuses on a few things:

1. always reserve protobuf repeated objects if you know how many things you will put in them
2. always reserve both map and array json objects when you know at least approximately how many things you will put in them
3. always std::move json map and array objects (as move assignment is cheaper than copy assignment in shared_ptr)

in my tests this seemed to help a little bit for the 50% percentile but gave a decent performance boost in the worst case. i was hoping that maybe the bulk of the performance difference with this branch was happening in the worst 10% of routes or something and it was skewing the results. after posting so many different timings its hard for me to tell what is what at this point so i'm wondering if you try the changes in this branch above out, does it make a difference for you?

src/thor/triplegbuilder.cc
genadz: will check

genadz:
> @genadz it looks liek you didnt push you latest changes up right? one thing i was trying to say with my second to last comment was that if you **ONLY** commented out the `AddTripIntersectingEdge` method then it was still doing the jumping between tiles/levels, whcih means its actually the work that its doing in that method to copy the attribution that is slow rather than the jumping around the graph. i have an idea to get a quick win for performance there. let me try an experiment...

1. pushed the code that skip lower levels for recovered edges.
2. sorry, I misprinted, I commented out `AddTripIntersectingEdges` (not `AddTripIntersectingEdge`); this, obviously, improved performance a lot.

genadz:
@kevinkreiser thanks for the ideas/experiments! will try with your changes on Monday

src/thor/triplegbuilder_utils.h
```diff
@@ -486,6 +489,9 @@ void AddIntersectingEdges(const AttributesController& controller,
   if (node->transition_count() > 0) {
     const NodeTransition* trans = start_tile->transition(node->transition_index());
     for (uint32_t i = 0; i < node->transition_count(); ++i, ++trans) {
+      // Skip nodes from lower levels if the corresponding flag is enabled
+      if (!allow_downwards_transitions && !trans->up())
+        continue;
```
kevinkreiser, line 494: unfortunately we cant do this here. i think what we really want is to not call `AddIntersectingEdges` at all for nodes in the route which are between edges that were recovered from a shortcut. This will match the current implementation and not send confusing looking intersections to odin.

kevinkreiser:
@genadz testing the latest code here i still see very large performance difference, in absolute terms it takes me about 23% more time to complete a 14k set of routes. also there are about 10% diffs in a RAD of the same route set. i would have expected some diffs but 10% is staggering. we should that we are seeing diffs we expect.

@dgearhart would you be able to take a look? the summary here is that the code now turns shortcut edges in the path into the list of underlying edges. it doesnt add the intersecting edges at the nodes that were previously non-existant.

at first i thought at sharp turns and stuff, the narrative might get an extra maneuver but without in the intersecting edges that seems highly unlikely (i am pretty sure it wont happen). this leads me to believe we are somehow getting different paths but i fail to see how that is possible considering the change to recover the shortcuts is after the path is found!

dgearhart:
@kevinkreiser I can kick off in the background - i am assuming master vs this branch?

dgearhart:
> @dgearhart would you be able to take a look? the summary here is that the code now turns shortcut edges in the path into the list of underlying edges. it doesnt add the intersecting edges at the nodes that were previously non-existant.

@kevinkreiser 
I am seeing some time diffs - seem to be okay

I am seeing diffs like the following:
image
This looks good - we now have turn lanes with the `recost_shortcuts_new` branch

Why are we not adding the intersecting edge info? That would help with some missing maneuvers

I see a 32% delta for user routes - i do not have time to review all of them but someone should make a good pass at reviewing diffs

kevinkreiser:
@dgearhart i would love to add all the edges too but it turns out that costs a ton of CPU and therefore really makes the request slower. that would be the ultimate goal but the original goal of this work is just to be able to get the right costing/time for the edges along the path. i recently got my perf tools working in my IDE so i might have a look here. I'll say this, one of the big contributors is the state shield verbal regexs in odin. They alone are about 7% of the total request latency. I have an idea how to refactor them to make them less expensive but haven't worked on it yet.

mandeepsandhu:
>  I'll say this, one of the big contributors is the state shield verbal regexs in odin. They alone are about 7% of the total request latency. I have an idea how to refactor them to make them less expensive but haven't worked on it yet.

@kevinkreiser is this the shield name lookup in the shield manifest we do for creating banner instructions or is it part of building the voice instructions?

I would love to get rid of the shields manifest (its a big hashmap) altogether. We're working on changing the source of our shield images (from wikimedia to mapbox), so we might be able to get rid of it entirely in the near future.

genadz:
1. added commit with the logic that doesn't reject edges in recost function based on nodes/edges access restrictions.
2. don't know why, but I have different performance results: on my local machine current branch ~4% slower than `master`.

kevinkreiser:
@genadz ill test it again today maybe i didnt get the updates for some reason!

genadz:
> @genadz ill test it again today maybe i didnt get the updates for some reason!

thanks!

kevinkreiser:
ok looking at the 50th percentile yeah its about 5%. oddly the 90th is something like 15% slower but the 99th and 100th are 5 and 0 respectively. frankly its hard to tell what i should understand from this result. i guess mostly the 50th percentile is the most important since the bulk of requests will be in this range. when i look at the total time to complete the benchmark though its about a 7.5% performance drop. i think if we merge the optimizations branch and do a couple more optimizations we'll be able to pay for this with those :smile: 

i personally dont think we have to wait to merge this before we merge those, does anyone else have an opinion? @PureW or @danpat ?

```
branch:
0.002226353 0.038036823 0.144977236 0.778483784 1.566832542 

master:
0.002390385 0.036790848 0.125865579 0.713516281 1.550947189 
```

src/sif/recost.cc
```diff
@@ -94,10 +95,11 @@ void recost_forward(baldr::GraphReader& reader,
...
     // if its not time dependent set to 0 for Allowed method below
     const uint64_t localtime = offset_time.valid ? offset_time.local_time : 0;
-    // this edge is not allowed
+    // we should evaluate restrictions even if 'ignore_access' flag is true
     if (predecessor != baldr::kInvalidLabel &&
-        !costing.Allowed(edge, label, tile, edge_id, localtime, offset_time.timezone_index,
-                         time_restrictions_TODO)) {
+        (!costing.Allowed(edge, label, tile, edge_id, localtime, offset_time.timezone_index,
+                          time_restrictions_TODO) &&
+         !ignore_access)) {
```
kevinkreiser, line 102: move the flag up in the if so that allowed doesnt get called at all

```suggestion
        !ignore_access && !costing.Allowed(edge, label, tile, edge_id, localtime, offset_time.timezone_index,
                          time_restrictions_TODO)) {
```

genadz: we can't do this because we need to evaluate time restrictions.

genadz: or we can explicitly call `costing.EvaluateRestrictions` in case `ignore_access = true`

kevinkreiser: the restrictions are based on access though as well is there really a point of checking them? i guess im more thinking we should make the boolean called `throw_if_not_allowed` or something more generic like `strict` to just completely turn off checking allowed at all. maybe we can flip the meaning and say something like `bool allow_all = false`?

genadz: https://github.com/valhalla/valhalla/blob/master/test/astar.cc#L784 - this test fails if we skip `Allowed` call here

genadz: hm. what if we back to the first approach where we recost only shortcuts ? - in this case we will save all `time_restrictions` that we calculated for regular edges.
and, when recosting shortcuts, we can 1) use flag to turn off checking `Allowed` or 2) don't use this flag but in case recosting fails - we just don't expand shortcut edge, add shortcut to the final path

kevinkreiser: i personally prefer this implementation as its a lot less complex (no splicing). you have a good point about showing the restrictions. id say just let it as it is and maybe make a note that we have to put the flag second so we can get restriction information.

kevinkreiser:
@mandeepsandhu the main slowdown in odin is running over these regex's for every street name: https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/verbal_text_formatter_us.h#L37

i think we can rewrite this to do our own form of matching that isnt as smart as regex to speed this up.

kevinkreiser (approved):
thank you again so much for the very long path to getting here. i hope we can make a few more performance imporovements and then even remove the optimization for not adding intersecting edges.

PR #2712, opened 2020-12-07
timeinfo tracking in map matching

kevinkreiser:
Currently map matching (either fast or robust versions) dont use the TimeInfo object to track time along the route (once the match is found). This pr adds that tracking into both versions. It also makes some changes in the fast map matcher to prepare for supporting multileg matches. Right now i've gone a little too far with the multi leg support that sometimes the edge walk fails. So i'll either need to finish support for that or rollback some of the changes.


src/loki/height_action.cc
```diff
@@ -9,16 +10,6 @@ using namespace valhalla::midgard;
...
 
-namespace {
-PointLL to_ll(const valhalla::Location& l) {
-  return PointLL{l.ll().lng(), l.ll().lat()};
-}
-void from_ll(valhalla::Location* l, const PointLL& p) {
-  l->mutable_ll()->set_lat(p.lat());
-  l->mutable_ll()->set_lng(p.lng());
-}
-} // namespace
```
kevinkreiser, line 20: this existed in the header above so i did this little clean up

src/thor/map_matcher.cc
```diff
@@ -177,8 +174,8 @@ MapMatcher::FormPath(meili::MapMatcher* matcher,
 
   // We support either the epoch timestamp that came with the trace point or
   // a local date time which we convert to epoch by finding the first timezone
-  uint32_t origin_epoch = compute_origin_epoch(edge_segments, matcher, options);
-  std::string date_time = options.shape(0).has_date_time() ? options.shape(0).date_time() : "";
+  valhalla::baldr::DateTime::tz_sys_info_cache_t tz_cache;
+  auto time_info = init_time_info(edge_segments, matcher, options, &tz_cache);
```
kevinkreiser, line 178: this is the main point of this pr. we dont manually track second of week or the timezone etc. we do it with the time_info object as all the other time based algorithms do it

src/thor/route_matcher.cc
```diff
@@ -235,73 +237,85 @@ bool expand_from_node(const sif::mode_costing_t& mode_costing,
...
-      if (options.shape(0).has_date_time()) {
-        if (options.shape(0).date_time() == "current")
-          options.mutable_shape(0)->set_date_time(DateTime::iso_date_time(tz));
-        origin_epoch = DateTime::seconds_since_epoch(options.shape(0).date_time(), tz);
-      }
-      break;
-    }
-  }
+  valhalla::baldr::DateTime::tz_sys_info_cache_t tz_cache;
+  auto time_info = init_time_info(reader, options, &tz_cache);
```
kevinkreiser, line 318: so for the shape walking map matcher (ie the faster less robust one) we also switch to using time_info

src/thor/route_matcher.cc
```diff
@@ -235,73 +237,85 @@ bool expand_from_node(const sif::mode_costing_t& mode_costing,
...
+
 // For the route path using an edge walking method.
 bool RouteMatcher::FormPath(const sif::mode_costing_t& mode_costing,
                             const sif::TravelMode& mode,
-                            GraphReader& reader,
-                            const std::vector<meili::Measurement>& shape,
+                            baldr::GraphReader& reader,
                             valhalla::Options& options,
-                            std::vector<PathInfo>& path_infos) {
+                            std::vector<std::vector<PathInfo>>& legs) {
```
kevinkreiser, line 282: i started refactoring this method to support multileg routes but there needs to be a little bit more refactoring in the guts of these algorithm to make it possible. i started trying it but it was getting too hair so i aborted that for now

src/thor/trace_route_action.cc
```diff
@@ -147,26 +147,32 @@ void thor_worker_t::trace_route(Api& request) {
...
-      options.mutable_locations(0)->set_date_time(options.shape(0).date_time());
-
-    // Form the trip path based on mode costing, origin, destination, and path edges
-    auto& leg = *request.mutable_trip()->mutable_routes()->Add()->mutable_legs()->Add();
-    thor::TripLegBuilder::Build(options, controller, *reader, mode_costing, path.begin(), path.end(),
-                                *options.mutable_locations()->begin(),
-                                *options.mutable_locations()->rbegin(),
-                                std::list<valhalla::Location>{}, leg, {"edge_walk"}, interrupt);
-  } else {
+  std::vector<std::vector<PathInfo>> legs;
```
kevinkreiser, line 150: yeah so i refactored this so that when we finally get multileg support in the fast map matching algorithm this code will already be done, but as it stands today its still not supported

genadz (dismissed):

gknisely:
@kevinkreiser do we have existing tests already or do we need some

kevinkreiser:
@gknisely there are existing ones that cover depart_at but time invariant doesnt exist so far as i know. that behavior though is all wrapped into the timeinfo object. do you want me to add a few simple test cases just in case. should be quick

genadz (approved):

gknisely:
> @gknisely there are existing ones that cover depart_at but time invariant doesnt exist so far as i know. that behavior though is all wrapped into the timeinfo object. do you want me to add a few simple test cases just in case. should be quick

@kevinkreiser yes.  just to be safe

src/meili/candidate_search.cc
```diff
@@ -44,17 +44,17 @@ CandidateCollector::WithinSquaredDistance(const midgard::PointLL& location,
       continue;
     }
 
-    // Get the edge and its opposing edge. Transition edges are not
-    // allowed so we do not need to check node levels.
-    const auto opp_edgeid = reader_.GetOpposingEdgeId(edgeid, tile);
```
kevinkreiser, line 49: this was a bug. when you get the opposing id with the tile of the regular edge it can flip modify the tile to be the wrong tile which we use again to refer to the original edge. the code was throwing an exception, getting caught and failing the map match. this would happen on all map match input points which have radii large enough to intersect an edge that crosses a tile boundary

src/mjolnir/graphtilebuilder.cc
```diff
@@ -1021,11 +1021,8 @@ void GraphTileBuilder::AddBins(const std::string& tile_dir,
 
 // Add a predicted speed profile for a directed edge.
 void GraphTileBuilder::AddPredictedSpeed(const uint32_t idx,
-                                         const std::vector<int16_t>& profile,
+                                         const std::array<int16_t, kCoefficientCount>& coefficients,
```
kevinkreiser, line 1024: it was stupid to take a vector when we wanted the size of the thing to be exactly a certain length. std::array does this. so lets use that instead

test/gurka/CMakeLists.txt
```diff
@@ -17,9 +17,10 @@ if(ENABLE_DATA_TOOLS)
     file(RELATIVE_PATH FILENAME "${CMAKE_CURRENT_LIST_DIR}" ${FULLFILENAME})
     string(REGEX REPLACE "test_(.*).cc" "gurka_\\1" TESTNAME ${FILENAME})
     add_executable(${TESTNAME} EXCLUDE_FROM_ALL ${FILENAME})
-    set_target_properties(${TESTNAME} PROPERTIES
-      FOLDER "Tests"
-      COMPILE_DEFINITIONS VALHALLA_SOURCE_DIR="${VALHALLA_SOURCE_DIR}/")
+    set_target_properties(${TESTNAME} PROPERTIES FOLDER "Tests")
+    target_compile_definitions(${TESTNAME} PRIVATE
+      VALHALLA_SOURCE_DIR="${VALHALLA_SOURCE_DIR}/"
+      VALHALLA_BUILD_DIR="${VALHALLA_BUILD_DIR}/")
```
kevinkreiser, line 23: need this to reference properly the timezone sqlite that is built for use in timedependent tests

test/gurka/test_route.cc
```diff
@@ -247,7 +247,7 @@ TEST(AutoDataFix, deprecation) {
 class AlgorithmTest : public ::testing::Test {
 protected:
   static gurka::map map;
-  static uint32_t current, historical, constrained;
+  static uint32_t current, historical, constrained, freeflow;
```
kevinkreiser, line 250: i used this test as a model for what i did in the map matching test. they are basically idential and return identical results

test/test.h
```diff
@@ -84,7 +84,7 @@ using LiveTrafficCustomize = std::function<void(valhalla::baldr::GraphReader&,
 void customize_live_traffic_data(const boost::property_tree::ptree& config,
                                  const LiveTrafficCustomize& setter_cb);
 
-using HistoricalTrafficCustomize = std::function<std::vector<int16_t>(DirectedEdge&)>;
+using HistoricalTrafficCustomize = std::function<std::array<float, kBucketsPerWeek>(DirectedEdge&)>;
```
kevinkreiser, line 87: this is way more useful if you give it the raw speeds and let it convert it to the format stored in the tiles

test/gurka/test_match.cc
```diff
@@ -124,3 +125,156 @@ D--3--4--C--5--6--E)";
...
+    auto new_kmh = static_cast<uint32_t>(km / h + .5);
+    if (kmh != -1)
+      EXPECT_EQ(kmh, new_kmh);
+    kmh = new_kmh;
+  }
+  return kmh;
+}
+
+// map matching currently only allows forward time tracking and invariant so types 0, 1, 3.
+// type 2 arrive_by is not yet supported
```
kevinkreiser, line 237: we check all the different types of su pported timedependent map matching as well as all the different sources of speed data

gknisely (approved):

src/meili/candidate_search.cc
kevinkreiser: cc @PureW @mandeepsandhu @mookerji

PR #2714, opened 2020-12-07
shortcut recovery singleton and cache option

kevinkreiser:
Recovering shortcuts requires a graph traversal at the moment. Work was done in another branch to add superseded edges information to the tiles directly but that is currently not something we're persuing. As part of working towards recovering all shortcuts on a route #2661 and #2711 i had (wrongly) suspected that we didnt want to waste any additional time actually recovering shortcuts. so to that end i made a cached singleton to store recovery information. For a full planets worth of data this cache should be about 1gig in size. It also takes about 10 seconds on my machine to populate (further evidence for why its not the bottleneck on the other shortcut recovery prs.

anyway the work isnt completely useless but is likely practically not needed. it does ahve a few benefits:

1. it is handy though if you want to do planet wide operations and look up random shortcuts quickly
2. its nice to have moved shortcut recovery into its own testable file (makes graphreader source smaller)
3. changing the recovery method allows us to recover more of the shortcuts (the reverse oneway ones)


src/baldr/shortcut_recovery.h
```diff
@@ -0,0 +1,231 @@
...
+// TODO: break this out into a private header
+// a static cache for shortcut recovery that we can optionally pre-fill
+struct shortcut_recovery_t {
+protected:
+  /**
+   * Constructs a shortcut cache from a graphreaders tileset by recovering all shortcuts. If the
+   * graphreader passed in is null nothing is cached and revoery will happen on the fly
+   * @param reader
+   */
+  shortcut_recovery_t(valhalla::baldr::GraphReader* reader) {
```
kevinkreiser, line 22: on startup (when we first construct a graph reader) this gets called. if you wanted to turn caching on it will run through your tileset and cache all the superceded edges of all the shortcuts, if you didnt it will just bail immediately. in both cases it logs about what it did.

src/baldr/shortcut_recovery.h
```diff
@@ -0,0 +1,231 @@
...
+          decltype(recovered) opp_recovered = recovered;
+          std::reverse_copy(recovered.cbegin(), recovered.cend(), opp_recovered.begin());
+          // save some stats
+          bool failed = recovered.front() == shortcut_id;
+          unrecovered += failed;
+          superseded += failed ? 0 : recovered.size();
+          // cache it even if it failed (no point in trying the same thing twice)
+          shortcuts.emplace(shortcut_id, std::move(recovered));
+
+          // its cheaper to get the opposing without crawling the graph
```
kevinkreiser, line 62: instead of recovering the opposing edge separately later we just do it now

src/baldr/shortcut_recovery.h
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  /**
+   * Recovers the edges comprising a shortcut edge.
+   * @param reader       The graphreader for graph data ccess
+   * @param  shortcutid  Graph Id of the shortcut edge.
+   * @return Returns the edgeids of the directed edges this shortcut represents.
+   */
+  std::vector<valhalla::baldr::GraphId>
+  recover_shortcut(valhalla::baldr::GraphReader& reader,
```
kevinkreiser, line 97: this method was moved from graphreader into here, so that the cache can call it to fill the cache or call it on the fly in the event cache is disabled (the default)

src/baldr/shortcut_recovery.h
```diff
@@ -0,0 +1,231 @@
...
+      for (const DirectedEdge& edge : tile->GetDirectedEdges(node_index)) {
+        // are they the same enough that its part of the shortcut
+        // NOTE: this fails in about .05% of cases where there are two candidates and its not clear
+        // which edge is the right one. looking at shortcut builder its not obvious how this is
+        // possible as it seems to terminate a shortcut if more than one edge pair can be
+        // contracted... NOTE: because we change the speed of the edge in graph enhancer we cant use
+        // speed as a reliable determining factor
+        if (begin_node != edge.endnode() && !edge.is_shortcut() &&
+            edge.forwardaccess() == shortcut->forwardaccess() &&
+            edge.reverseaccess() == shortcut->reverseaccess() && edge.sign() == shortcut->sign() &&
```
kevinkreiser, line 154: previously we only recovered shortcuts on edges with forward auto access. ive changed that to just say that access is the same in both directions of the edge, whcih allows us to recover opposing shortcuts of oneways. in a recent planet dataset the stats were 

```
2020/12/07 21:23:37.570000 [INFO] Shortcut recovery cache enabled
2020/12/07 21:23:47.595293 [INFO] 14191294 shortcuts recovered as 90807070 superseded edges. 1776 shortcuts could not be recovered.
```

kevinkreiser: also notice recovery only takes 10 seconds, hence why we default to turning it off and doing on the fly recovery. caching of this costs approximately 1gig of ram per process. hardly worth the tiny amount of cpu it takes to recover on the fly

test/recover_shortcut.cc
```diff
@@ -153,7 +160,13 @@ TEST(RecoverShortcut, test_recover_shortcut_edges) {
...
 }
 
-} // namespace
+TEST(RecoverShortcut, test_recover_shortcut_edges_no_cache) {
+  recover(false);
+}
+
+TEST(RecoverShortcut, test_recover_shortcut_edges_cache) {
+  recover(true);
+}
```
kevinkreiser, line 169: we test with and without cache now

src/baldr/shortcut_recovery.h
```diff
@@ -0,0 +1,231 @@
...
+            edge.use() == shortcut->use() && edge.classification() == shortcut->classification() &&
+            edge.roundabout() == shortcut->roundabout() && edge.link() == shortcut->link() &&
+            edge.toll() == shortcut->toll() && edge.destonly() == shortcut->destonly() &&
+            edge.unpaved() == shortcut->unpaved() && edge.surface() == shortcut->surface()/* &&
+          edge.speed() == shortcut->speed()*/) {
```
PureW, line 159: Forgotten comment?

src/baldr/shortcut_recovery.h
```diff
@@ -0,0 +1,231 @@
...
+          current_edge = &edge;
+          begin_node = tile->header()->graphid();
+          begin_node.set_id(node_index);
+          accumulated_length += edge.length();
+          break;
+        }
+      }
+
+      // if we didnt add an edge or we went over the length we failed
+      if (current_edge == nullptr || accumulated_length > shortcut->length()) {
```
PureW, line 173: The reason it's not an "approximate equality" is because not all parts of the shortcut is required to be recovered?

PureW: Oh, the test for shorter length is below. I see.

src/baldr/shortcut_recovery.h
kevinkreiser: no that was there in the original. i left it unchanged. the comment above explains why its not turned on

src/baldr/shortcut_recovery.h
```diff
@@ -0,0 +1,231 @@
...
+public:
+  /**
+   * returns a static instance of the cache after prefilling it. if on the first call,
+   * the reader is nullptr then the cache will not be filled and recovery will be on the fly
+   *
+   * @param reader       the reader used to initialize the cache the first time
+   * @return a filled cache mapping shortcuts to superceeded edges
+   */
+  static shortcut_recovery_t& get_instance(valhalla::baldr::GraphReader* reader = nullptr) {
+    static shortcut_recovery_t cache{reader};
```
PureW, line 209: Should `cache` be `thread_local` so as to not break if used from multiple threads?

kevinkreiser: the cache is read-only which is completely fine, and in c++11 and on we are gauranteed that this initialization will happen only once. https://stackoverflow.com/a/19907903/5251867

PureW (approved):

PR #2717, opened 2020-12-08
Sanitize Logging and Add Timing

kevinkreiser:
This PR does 2 things:

1. it removes all the analytics logging that we had, this fell into disuse. in the future we'll add a nice clear place to put logging information for those people who want to run their own deploys and capture their own metrics
2. it adds a scoped timer template class which we use in all the major places to measure how much time it took to do that particular major chunk of the request/api. this info is then stored in the protobuf request object itself

I want to do a refactor of how we store the intermediate results before serialization so that we can optinally add this timing information and other statistics to the response of whatever api was called but for the sake of making the prs bite sized ill leave that for another pr


proto/info.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
+option optimize_for = LITE_RUNTIME;
+package valhalla;
+
+message Statistic {
+  optional string name = 1;  // the name of the statistic
+  optional double value = 2; // the value of the statistic
+}
+
+message Info{
```
kevinkreiser, line 10: this should be a place where we can add statistical information about the processing of an api call or request. for now its just key value pairs

proto/info.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
+option optimize_for = LITE_RUNTIME;
+package valhalla;
+
+message Statistic {
+  optional string name = 1;  // the name of the statistic
+  optional double value = 2; // the value of the statistic
```
kevinkreiser, line 7: with proto3 we could use OneOf here which might be a good idea (just add all the primative types). any thoughts?

src/loki/height_action.cc
```diff
@@ -79,12 +79,17 @@ std::vector<PointLL> loki_worker_t::init_height(Api& request) {
 }
 */
 std::string loki_worker_t::height(Api& request) {
+  // time this whole method and save that statistic
+  midgard::scoped_timer<> t([&request](const midgard::scoped_timer<>::duration_t& elapsed) {
+    auto e = std::chrono::duration_cast<std::chrono::duration<double, std::milli>>(elapsed).count();
+    auto* stat = request.mutable_info()->mutable_statistics()->Add();
+    stat->set_name("loki_worker_t::height");
+    stat->set_value(e);
+  });
```
kevinkreiser, line 88: this block is the same throughout, when the timer goes out of scope this lambda gets called back with the time that passed since the object was created. then we set the name and the value. more of an explanation of this pattern can be read here: https://stackoverflow.com/questions/2808398/easily-measure-elapsed-time/55265585#55265585

src/thor/route_action.cc
```diff
@@ -363,7 +380,7 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
         get_path_algorithm(costing, *origin, *destination, api.options());
     path_algorithm->Clear();
     algorithms.push_back(path_algorithm->name());
-    valhalla::midgard::logging::Log(path_algorithm->name(), " [ANALYTICS] algorithm::");
+    LOG_INFO(std::string("algorithm::") + path_algorithm->name());
```
kevinkreiser, line 383: for now i left this in as a LOG_INFO since its the biggest performance differentiator. i could be convinced to remove it though, thoughts?

src/worker.cc
```diff
@@ -1088,11 +1084,6 @@ void ParseApi(const http_request_t& request, valhalla::Api& api) {
     options.set_action(action);
   }
 
-  // disable analytics
```
kevinkreiser, line 1091: previously we had a request option relying on the do not track header being in the request. but now that we dont track anything anyway its superfluous

src/loki/isochrone_action.cc
```diff
@@ -64,6 +64,14 @@ void loki_worker_t::init_isochrones(Api& request) {
   parse_costing(request);
 }
 void loki_worker_t::isochrones(Api& request) {
+  // time this whole method and save that statistic
+  midgard::scoped_timer<> t([&request](const midgard::scoped_timer<>::duration_t& elapsed) {
+    auto e = std::chrono::duration_cast<std::chrono::duration<double, std::milli>>(elapsed).count();
+    auto* stat = request.mutable_info()->mutable_statistics()->Add();
+    stat->set_name("loki_worker_t::isochrones");
+    stat->set_value(e);
+  });
```
danpaz, line 73: Can we turn this repeated block into a macro?

proto/info.proto
PureW: Devils advocate: Is there a need for that?

Most usecases can make do with a floating point value. Do you have any examples that need something else?

src/loki/height_action.cc
PureW: Could the variable be named something clearer like `timer` instead of `t`?

src/thor/route_action.cc
PureW: I find it helpful when debugging issues to at least as a sanity check that the right a-star is handling the request. I'd vote for keeping it.

proto/info.proto
kevinkreiser: yeah i totally think in the context of metrics it makes sense to just have a single floating point value, i was just trying to suggest that maybe i didnt think of all the possible use-cases and that if there were more we could fix it with OneOf. currently i still cannot come up with a good use-case for anything but numbers though :smile:

kevinkreiser: i could kind of see a metric where an enum might make sense, in which case maybe strings would be better than numbers :man_shrugging:

src/loki/isochrone_action.cc
kevinkreiser: yeah i think we can make a function in worker.h or some other fundamental header and call it something like `make_timer(...)` should be straight forward!

proto/info.proto
mandeepsandhu: How is this proro msg intended to be consumed/used? Do you envision users using it to emit metrics to an external metrics collector (eg `statsd`/`prometheus` etc)?

If these statistics have to be emitted as metrics, then the user might also want to know what type of stat is within. Common types would be counters, gauges & timer (they all have different meaning for a metrics collector and hence have to be distinguished before being sent). Or is it intended to only have timing information?

We could also emit the metrics directly from the place of measurement, in which case we'd know what is it that we're measuring (timer/counts etc).

kevinkreiser: i didnt add it to this pr but what i want to do is refactor serialization to add this info to the serialized response (when a preprocessor define is turned on)

kevinkreiser:
@PureW @mandeepsandhu @danpaz this is ready for review

src/thor/route_action.cc
mandeepsandhu: yeah, +1 on keeping it.

proto/info.proto
mandeepsandhu: This could also be a place to add metadata about the type of statistic contained within. Eg if the stat is a timer, a count etc.

mandeepsandhu (approved):
LGTM :+1:

PR #2718, opened 2020-12-09
Replace all Python2 print statements with Python3 syntax

### LINKED ISSUE #2716: Adapt valhalla_build_elevation to python3
Still uses py2 syntax.. newer ubuntu versions only have py3.

nilsnolde:
fixes #2716 

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,7 @@
    * FIXED: Updated logic to call out a non-obvious turn [#2708](https://github.com/valhalla/valhalla/pull/2708)
    * FIXED: valhalla_build_statistics multithreaded mode fixed [#2707](https://github.com/valhalla/valhalla/pull/2707)
    * FIXED: If infer_internal_intersections is true then allow internals that are also ramps or TCs. Without this we produce an extra continue manuever.  [#2710](https://github.com/valhalla/valhalla/pull/2710)
+   * FIXED: Replace all Python2 print statements with Python3 syntax [#2716](https://github.com/valhalla/valhalla/issues/2716)
 
 * **Enhancement**
    * ADDED: Add ability to provide custom implementation for candidate collection in CandidateQuery. [#2328](https://github.com/valhalla/valhalla/pull/2328)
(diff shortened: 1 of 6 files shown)


PureW:
Are you aware of any division happening in the python scripts @nilsnolde ? I'm thinking about how that behaviour changed between python 2 and 3.

nilsnolde:
Very good point @PureW! Skimming through some of the py files: yes.. `scripts/list_tiles.py` and `scripts/transit_tiles.py` both do some integer division (though some might be floats..).. can't tell though if that's relevant or in which python version those were written in (they're not part of this PR, so no py2 print statements).. jeez, brings back some dark memories when migrating qgis stuff to py/qgis3 :sweat_smile: 

are they even used? list_tiles.py was written by @gknisely , maybe you got some insight?

dnesbitt61:
list_tiles.py is useful for anyone who wants to do pull a partial data set from a planet wide set of Valhalla tiles. I use code derived from this to create tar files that support "regions" defined by a bounding box. Useful for mobile/embedded uses of Valhalla.

PureW (approved):
LGTM

mandeepsandhu:
Why limit to only `print` statements? Curious if you considered doing a one-time migration using a tool like [`2to3`](https://docs.python.org/3/library/2to3.html) (kind of an official tool for migrating tp python 3)?

nilsnolde:
Yeah right, should've left some reasoning behind this:) we couldn't build elevation anymore after basing the valhalla docker image on 20.04, so my main purpose was to fix the elevation build script. And then I thought I should be nice and at least fix the other print statements as well. There might be some breaking imports left, but didn't see any obvious ones. 

Personally I don't like these kinda blackbox tools rewriting code, I'm a little neurotic there (and never bothered to see what 2to3 _exactly_ does). The most dangerous change is really what Anders pointed at IMO, which needs more of a deep dive than 2to3 could offer.

I'll leave it to someone else to merge:)

PureW:
@nilsnolde I'd go ahead and merge this, but it's out-of-date and I don't see the button for automatically bringing it up-to-date. Perhaps there's a conflict with master somewhere?

nilsnolde:
Ha weird, I do see it. I press and can do the merge then as well :)

Issue #1665, opened 2018-12-05
EdgeLabel allocation - shrink_to_fit


dnesbitt61:
EdgeLabels reserve space, and if needed will grow. A related issue (#272) discusses changing the reservation size based on predicted route length. This would reduce the initial allocaton of memory and help keep memory use lower in embedded environments.
For cases where path algorithms are persistent, the EdgeLabel allocation will grow to the largest size needed by any route, and currently will not reduce allocation afterwards. This issue recommends using vector::shrink_to_fit to reduce the size of the allocated EdgeLabel vectors so that the memory use of ath algorithms can be more strictly controlled.
Other temporary objects in path algorithm (DoubleBucketQueue and EdgeStatus) seem to clean up allocation, but this should be double checked to make sure they don't retain excess allocation.

Issue #272, opened 2017-02-02
Allow EdgeLabel reserved size to be configured


dnesbitt61:
(2016-06-02T16:37:54Z) @dnesbitt61 : originally created valhalla/thor#382

We configure in PathAlgorithm to be 500000 - this was optimized for long distance routes so when reallocation occurs it doesn't slow the algorithm so much. 

To support lower memory environments we should allow this to be configurable.  EdgeLabel size is about 72 bytes so this is 36 MB reserved. Systems only using shorter routes can get by with much less.
Also need to support reserving space for the reverse search EdgeLabels.

PR #2719, opened 2020-12-09
Optimize double_bucket_queue

averkhaturau:
# Issue

Optimization of routes calculation based on performance test profiling.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

https://github.com/valhalla/valhalla/pull/2703
 #1665


dnesbitt61:
This longstanding issue may be related: https://github.com/valhalla/valhalla/issues/1665.

valhalla/baldr/double_bucket_queue.h
```diff
@@ -82,16 +112,25 @@ class DoubleBucketQueue final {
   void clear() {
     // Empty the overflow bucket and each bucket
     overflowbucket_.clear();
-    while (currentbucket_ != buckets_.end()) {
```
kevinkreiser, line 85: this seems to suggest that we dont need to clear them all

valhalla/baldr/double_bucket_queue.h
```diff
@@ -82,16 +112,25 @@ class DoubleBucketQueue final {
...
-      currentbucket_->clear();
-      currentbucket_++;
+    for (auto& b : buckets_) {
+      b.clear();
```
kevinkreiser, line 116: i think we need to also need to reduce the size of the buckets and the overflowbucket if it goes over some configurable value otherwise they will grow infinitely

valhalla/baldr/double_bucket_queue.h
```diff
@@ -68,6 +93,11 @@ class DoubleBucketQueue final {
     // Allocate the low-level buckets
     size_t bucketcount = (range / bucketsize_) + 1;
     buckets_.resize(bucketcount);
+    for (auto& b : buckets_) {
+      b.clear();
+      b.reserve(bucketsize_ / 2 + 1);
```
dnesbitt61, line 98: What is the value of reserving here? The bucketsize is generally = 1 (meaning each bucket covers all edges that have cost within 1 second of each other). Maybe bucketsize is a misleading name but it is intended to mean the cost range held within a bucket. Ideally you would estimate the number of edges that might end up within the same bucket - but that is highly dependent on the density of the network touched by the actual route so I can't think of a metric to intelligently reserve an appropriate size. Many buckets could actually end up being empty so reserving a larger vector may be counter-productive.

valhalla/baldr/double_bucket_queue.h
dnesbitt61: Normal operation will end up pop'ing elements from all lower cost buckets so this means we need to only clear buckets >= the current cost bucket.

valhalla/baldr/double_bucket_queue.h
averkhaturau: Thank you, @dnesbitt61. I misunderstood meaning of bucketsize_. Perhaps you can suggest some better value for reserve, then just not reserving anything?

valhalla/baldr/double_bucket_queue.h
averkhaturau: Thank you, I will remove this change.

valhalla/baldr/double_bucket_queue.h
averkhaturau: I actually do the same thing in reuse() method. I'm not yet sure about the purpose of clear(), it seems it was never called, except tests. To free all allocated memory it's better to call `buckets_.clear()`, but in this case memory cannot be reused.

valhalla/baldr/double_bucket_queue.h
dnesbitt61: Well, the problem is that some buckets may be more populated than others (majority are likely to be 0). I think one would have to do some routes and log the sizes that occur (max size, perhaps the % of buckets that have any edges added). If this is only ever used in server applications then reserving a modest size for each bucket (32, 64?) makes sense but what about embedded applications? I'd hate to reserve a bunch of memory for sparsely populated buckets. Perhaps this behavior can be made optional so embedded applications can better control memory use?

averkhaturau: According to experiment with test set of routes on desktop, this reserve wins statistically insignificant performance improvement. Average value is 0.0, maximum value encountered was 14, but usually it's no more then 4. Requested bucketcount during the experiment was 2501 or 20001.

averkhaturau: `b.reserve(...)` is removed in last update of this PR

valhalla/baldr/double_bucket_queue.h
```diff
@@ -147,18 +179,18 @@ class DoubleBucketQueue final {
...
   }
 
 private:
-  float bucketrange_; // Total range of costs in lower level buckets
-  float bucketsize_;  // Bucket size (range of costs in same bucket)
-  float inv_;         // 1/bucketsize (so we can avoid division)
-  double mincost_;    // Minimum cost within the low level buckets
-  float maxcost_;     // Above this goes into overflow bucket
-  float currentcost_; // Current cost
+  float bucketrange_{}; // Total range of costs in lower level buckets
```
dnesbitt61, line 193: Are these brackets needed since default initialization will be 0 anyway?

averkhaturau: I added default c-tor, so I need to initialize all these values to 0

kevinkreiser:
master:
0.002086401 0.037312865 0.126078653 0.703841746 1.554239988 
[1] 0.1261245

this branch:
```
0.00281930 0.03657818 0.12528012 0.68923821 1.43198729 
[1] 0.1231092
```

Looks like an all around 2% boost in performance. I need to run tests to see what it does to the memory usage

src/thor/astar_bss.cc
```diff
@@ -41,11 +42,17 @@ AStarBSSAlgorithm::~AStarBSSAlgorithm() {
...
+  if (edgelabels_.size() > max_reserved_labels_count_) {
+    edgelabels_.resize(max_reserved_labels_count_);
+    edgelabels_.shrink_to_fit();
+  }
   // Clear the edge labels and destination list. Reset the adjacency list
   // and clear edge status.
   edgelabels_.clear();
   destinations_.clear();
-  adjacencylist_.reset();
+  if (adjacencylist_)
```
kevinkreiser, line 54: i didnt checking into it at all, but since we always want to reuse it via `clear` can we also make it nonpointer?

valhalla/thor/worker.h
```diff
@@ -113,13 +113,13 @@ class thor_worker_t : public service_worker_t {
...
-  TimeDepForward timedep_forward;
-  TimeDepReverse timedep_reverse;
+  std::unique_ptr<BidirectionalAStar> bidir_astar;
+  std::unique_ptr<AStarBSSAlgorithm> bss_astar;
+  std::unique_ptr<MultiModalPathAlgorithm> multi_modal_astar;
+  std::unique_ptr<TimeDepForward> timedep_forward;
+  std::unique_ptr<TimeDepReverse> timedep_reverse;
 
-  Isochrone isochrone_gen;
+  std::unique_ptr<Isochrone> isochrone_gen;
```
kevinkreiser, line 122: its not obvious to me the benefit of this change, what am i missing?

src/thor/astar_bss.cc
genadz: originally `DoubleBucketQueue` doesn't have a default constructor (because it doesn't make sense). that is the reason why I use a pointer. But from the other hand I agree with you @kevinkreiser , code looks more bulky with `unique_ptr`; will try to recollect it

valhalla/thor/worker.h
genadz: since we have to recreate all of these classes in `thor_worker` constructor (because of new parameter `max_reserved_labels_count`), it's not allowed to write `bidir_astar = BidirectionalAstar(...)` because it will cause copying.

idea: We can leave nonpointers here and explicitly forbid copy-constructor for these classes

src/thor/astar_bss.cc
genadz: ok. fixed it by introducing (actually, re-introducing) default constructor for DoubleBucketQueue

valhalla/thor/worker.h
genadz: there are 2 options I see:
1) add move constructor/operator= for each class derived from `PathAlgorithm` and `Dijkstras` and leave all destructors empty (because in case of `move` all states inside classes will be invalid).
2) explicitly forbid copy/move constructors for `PathAlgorithm` and `Dijkstras` (in this case we will be able to access the internal state in destructors) and leave `unique_ptr`s

bench/thor/routes.cc
```diff
@@ -178,6 +178,7 @@ static void BM_UtrechtBidirectionalAstar(benchmark::State& state) {
       // LOG_WARN("Running index "+std::to_string(i));
       auto result = astar.GetBestPath(origins[i], destinations[i], *clean_reader, costs,
                                       sif::TravelMode::kDrive);
+      astar.Clear();
```
genadz, line 181: we must clear the state before the next call

bench/thor/costmatrix.cc
```diff
@@ -105,6 +105,7 @@ static void BM_UtrechtCostMatrix(benchmark::State& state) {
   for (auto _ : state) {
     thor::CostMatrix matrix;
     auto result = matrix.SourceToTarget(sources, sources, reader, costs, mode, 100000.);
+    matrix.Clear();
```
genadz, line 108: we must clear the state before the next call

scripts/valhalla_build_config
```diff
@@ -91,7 +91,8 @@ config = {
     'source_to_target_algorithm': 'select_optimal',
     'service': {
       'proxy': 'ipc:///tmp/thor'
-    }
+    },
+    'max_reserved_labels_count': 1000000
```
genadz, line 95: according to https://github.com/valhalla/valhalla/issues/272#issue-204885361 it's about ~70 Mb

valhalla/baldr/double_bucket_queue.h
```diff
@@ -163,7 +192,7 @@ template <typename label_t> class DoubleBucketQueue final {
   bucket_t overflowbucket_;
 
   // Access to a container of labels to get cost given the label index.
-  const std::vector<label_t>& labelcontainer_;
+  const std::vector<label_t>* labelcontainer_;
```
kevinkreiser, line 195: why change this?

valhalla/baldr/double_bucket_queue.h
```diff
@@ -33,13 +39,35 @@ template <typename label_t> class DoubleBucketQueue final {
...
+   * The same as c-tor, but without buffers reallocation. Before call this
+   * method you should clean up the current state (call `clear`).
+   * @param mincost    Minimum cost. Used to create the initial range for
+   *                   bucket sorting.
+   * @param range      Cost range for low-level buckets.
+   * @param bucketsize Bucket size (range of costs within same bucket).
+   *                   Must be an integer value.
+   * @param labelcontainer  Container of labels with sortcosts.
+   */
+  void reuse(const float mincost,
```
kevinkreiser, line 66: maybe we should call this `reserve` to be more stl like

valhalla/thor/worker.h
kevinkreiser: you can just do it in the initializer list. first in the initializer list you set `reserved_label_count`, the you use it to initialize all the other algorithme members

valhalla/baldr/double_bucket_queue.h
genadz: hm, interesting idea. the only difference this is `must-have` function to use the queue. in stl it's not necessary to call `reserve` before usage

valhalla/baldr/double_bucket_queue.h
kevinkreiser: ahh that makes sense. its really hard to see a detail like that without it being spelled out

valhalla/baldr/double_bucket_queue.h
kevinkreiser: no worries just a suggestion

genadz:
`max_reserved_labels_count` parameter is used only to initialize `PathAlgorithm`s, it's not necessary to keep this parameter in thor_worker

kevinkreiser (approved):

PR #2724, opened 2020-12-11
Fixes problem with generating valid HGT file names (#2723)

### LINKED ISSUE #2723: Elevation tiles cannot be found
This is because the name_hgt function takes a signed int, but the index is unsigned so if it exceeds 32767 the resulting tile file name will be incorrect.

A pull request with a fix is forthcoming.

jfamiglietti:
# Issue

Fixes #2723. Some HGT files are not found

## Tasklist

 - [x] Add tests


valhalla/skadi/sample.h
```diff
@@ -45,6 +45,16 @@ class sample {
...
+  /**
+   * @return A tile index value from a coordinate
+   */
+  template <class coord_t> static uint16_t get_tile_index(const coord_t& coord);
+
+  /**
+   * @return The file name of a tile for a given index
+   */
+  static std::string get_hgt_file_name(uint16_t index);
+
```
kevinkreiser, line 57: since these are only used privately in the actual API and in the test can we hide them in the implementation by making them protected. we like to keep the public interface to our classes as minimal as possible.

to achieve this you can change your unit test to use a child class of `sample`. something like this potentially:

```c++
struct testable_sample : public sample<PointLL>{
  static uint16_t get_tile_index(const PointLL& coord){ return sample<PointLL>::get_tile_index(coord); }
  static std::string get_hgt_file_name(uint16_t index){return sample<PointLL>::get_hgt_file_name(index);}
};
```

jfamiglietti: Quite reasonable. Will do.

jfamiglietti: This is taken care of. Turns out there was already a `testable_sample_t` in that test class.

kevinkreiser (approved):

PR #2728, opened 2020-12-11
we should stop expanding if we hit our reach target

kevinkreiser:
it seems to me like we should tell dijkstras to stop instead of prune when we hit our reach target. to me the current code seem to suggest that instead of doing that, we make sure to first pop off and settle all of the outstanding labels in the label set, which is a huge waste of time!

this is only for exact reach which doesnt happen that often but when it does this should be more performant so long as i'm correct :wink:


kevinkreiser:
this was fixed as part of another pr #2734

Issue #1188, opened 2018-04-08
Time dependent routing - depart at


gknisely:
In order to be able to depart at a particular time and use predictive traffic we must use forward A*.  First phase should be just the implementation of forward A* that is time-dependent.  

- [ ] Design and implement time-dependent forward A* for depart-at routing.
- [ ] Test using real life routing.

PR #2730, opened 2020-12-12
Handle GraphTile objects as boost::intrusive_ptr

averkhaturau:
# Issue

An alternative to #1188 with attempts to improve performance

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/meili/candidate_search.cc
+++ b/src/meili/candidate_search.cc
@@ -134,9 +134,7 @@ void IndexBin(const graph_tile_ptr& tile,
               const int32_t bin_index,
               baldr::GraphReader& reader,
               CandidateGridQuery::grid_t& grid) {
-  if (!tile) {
-    return;
-  }
+  assert(tile);
 
   // Get the edges within the specified bin.
   auto edge_ids = tile->GetBin(bin_index);
@@ -189,9 +187,7 @@ CandidateGridQuery::GetGrid(const int32_t bin_id,
   int32_t tile_id = tiles.TileId(rc.second / ndiv, rc.first / ndiv);
   baldr::GraphId tileid(tile_id, bin_level_, 0);
   auto tile = reader_.GetGraphTile(tileid);
-  if (!tile) {
-    return nullptr;
-  }
+  assert(tile);
 
   // Compute bin index within the tile (row-ordered)
   int32_t bin_row = rc.first % ndiv;
--- a/src/meili/candidate_search.cc
+++ b/src/meili/candidate_search.cc
@@ -142,8 +142,8 @@ void IndexBin(const graph_tile_ptr& tile,
     // Get the right tile (edges in a bin can be in a different tile if they
     // pass through the tile but do not start or end in the tile). Skip if
     // tile is null.
-    auto bin_tile =
-        edge_id.tileid() == tile->header()->graphid().tileid() ? tile : reader.GetGraphTile(edge_id);
+    auto bin_tile = tile;
+    reader.GetGraphTile(edge_id, bin_tile);
     if (bin_tile == nullptr) {
       continue;
     }
--- a/src/meili/candidate_search.cc
+++ b/src/meili/candidate_search.cc
@@ -134,6 +134,10 @@ void IndexBin(const boost::intrusive_ptr<const baldr::GraphTile>& tile,
               const int32_t bin_index,
               baldr::GraphReader& reader,
               CandidateGridQuery::grid_t& grid) {
+  if (!tile) {
+    return;
+  }
+
   // Get the edges within the specified bin.
   auto edge_ids = tile->GetBin(bin_index);
   for (const auto& edge_id : edge_ids) {
--- a/src/meili/candidate_search.cc
+++ b/src/meili/candidate_search.cc
@@ -36,7 +36,7 @@ CandidateCollector::WithinSquaredDistance(const midgard::PointLL& location,
   std::vector<baldr::PathLocation> candidates;
   std::unordered_set<baldr::GraphId> visited_nodes;
   midgard::projector_t projector(location);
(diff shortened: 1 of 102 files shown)


kevinkreiser:
@averkhaturau let me know if/when you'd want me to test this out, thanks for giving it a shot!

valhalla/baldr/graphtile.h
```diff
@@ -45,7 +47,7 @@ class tile_getter_t;
 /**
  * Graph information for a tile within the Tiled Hierarchical Graph.
  */
-class GraphTile {
+class GraphTile : public boost::intrusive_ref_counter<GraphTile, boost::thread_unsafe_counter> {
```
kkaefer, line 50: I think it might be useful to make the constructors private and use a factory function to construct `boost::intrusive_ptr` objects to find uses of `GraphTile` that sidestep the new "official" way to construct them.

kevinkreiser:
Good news, this approach is about 1.1% slower than master which is completely reasonable for the benefits it brings. We just need to review the actual code changes here to get this shipped.

valhalla/baldr/graphtile.h
averkhaturau: Done.

src/mjolnir/graphtilebuilder.cc
```diff
@@ -890,13 +890,14 @@ void GraphTileBuilder::AddTileCreationDate(const uint32_t tile_creation_date) {
...
-std::array<std::vector<GraphId>, kBinCount> GraphTileBuilder::BinEdges(const GraphTile* tile,
-                                                                       tweeners_t& tweeners) {
+std::array<std::vector<GraphId>, kBinCount>
+GraphTileBuilder::BinEdges(const boost::intrusive_ptr<const GraphTile>& tile, tweeners_t& tweeners) {
   std::array<std::vector<GraphId>, kBinCount> bins;
   // we store these at the highest level
   auto max_level = TileHierarchy::levels().back().level;
   // skip transit or other special levels and empty tiles
-  if (tile->header()->graphid().level() > max_level || tile->header()->directededgecount() == 0) {
+  if (!tile || tile->header()->graphid().level() > max_level ||
```
kevinkreiser, line 899: we never call these functions with null tiles, if we do we dont want to hide a bug upstream. if you want to add stuff like this change it to an assertion that the tiles arent null

src/mjolnir/graphvalidator.cc
```diff
@@ -557,7 +553,17 @@ void GraphValidator::Validate(const boost::property_tree::ptree& pt) {
...
-  auto dataset_id = GraphTile(tile_dir, *tilequeue.begin()).header()->dataset_id();
+
+  uint64_t dataset_id = 0;
+  if (const auto& tile = GraphTile::Create(tile_dir, *tilequeue.begin())) {
+    if (const auto* header = tile->header()) {
+      dataset_id = header->dataset_id();
+    }
+  }
+  if (dataset_id == 0) {
+    LOG_ERROR("Tile isn't created from " + tile_dir + " with id " +
```
kevinkreiser, line 564: im not really sure this is needed but a more appropriate message would be "could not load tile from disk"

src/mjolnir/transitbuilder.cc
```diff
@@ -449,11 +449,14 @@ void FindOSMConnection(const PointLL& stop_ll,
 void AddOSMConnection(const GraphId& transit_stop_node,
                       const NodeInfo* transit_node,
                       const std::string& stop_name,
-                      const GraphTile* tile,
+                      const boost::intrusive_ptr<const GraphTile>& tile,
                       GraphReader& reader_local_level,
                       std::mutex& lock,
                       std::vector<OSMConnectionEdge>& connection_edges) {
-
+  if (!tile) {
```
kevinkreiser, line 456: assert instead?

averkhaturau: Done.

src/mjolnir/graphvalidator.cc
averkhaturau: Done.

src/mjolnir/graphtilebuilder.cc
averkhaturau: Done.

kevinkreiser:
@averkhaturau ive pushed a smal change to give the new GraphTile* an alias. i'm running testing on data building to see if there are any differences there, then ill run RAD to see if any differences there. i'll validate performance really quickly once again and then we can ship this

src/meili/candidate_search.cc
```diff
@@ -130,18 +130,22 @@ CandidateCollector::WithinSquaredDistance(const midgard::PointLL& location,
...
 // Add each road linestring's line segments into grid. Only one side
 // of directed edges is added
-void IndexBin(const baldr::GraphTile& tile,
+void IndexBin(const graph_tile_ptr& tile,
               const int32_t bin_index,
               baldr::GraphReader& reader,
               CandidateGridQuery::grid_t& grid) {
+  if (!tile) {
+    return;
+  }
```
kevinkreiser, line 139: there are still a lot of these new `if`s which i get but technically they are logically impossible. id prefer that these become `assert`s

kevinkreiser:
good news the created tiles are identical in single threaded mode. all i need to do yet is RAD testing. i can do that quickly tomorrow morning. almost there!

src/meili/candidate_search.cc
averkhaturau: Done.

kevinkreiser:
ok ive RAD test this and there are no differences. the performance difference between this and master is around 1-2% which is ok by me.

kevinkreiser (dismissed):

kevinkreiser (approved):

PR #2731, opened 2020-12-14
added sump buster logic.

gknisely:
# Issue

## Tasklist

 - [x] Add tests

## Requirements / Relations


test/gurka/test_node_access.cc
```diff
@@ -0,0 +1,73 @@
...
+      {"GK", {{"highway", "residential"}}}, {"HL", {{"highway", "residential"}}},
+      {"IJ", {{"highway", "residential"}}}, {"JK", {{"highway", "residential"}}},
+      {"KL", {{"highway", "residential"}}},
+  };
+
+  const gurka::nodes nodes = {{"F", {{"motor_vehicle", "no"}}}, {"G", {{"motorcar", "no"}}}};
+
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, 100);
+  auto map = gurka::buildtiles(layout, ways, nodes, {}, "test/data/gurka_node_access");
+  auto result = gurka::route(map, "A", "I", "auto");
```
mandeepsandhu, line 29: Suggestion: Instead of repeating the test for different costing models (where the expected result is the same), you can use [parametrized fixtures](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md) (see **Value-Parameterized Tests**) to pass-in the costing string as an arg to your test and then use that when making the route request.

test/gurka/test_node_access.cc
```diff
@@ -0,0 +1,73 @@
+#include "gurka.h"
+#include <gtest/gtest.h>
+
+using namespace valhalla;
+
+TEST(Standalone, NodeAccess) {
```
mandeepsandhu, line 6: IS this test related to the sump_buster change?

test/gurka/test_sump_buster.cc
```diff
@@ -0,0 +1,151 @@
...
+      {"CD", {{"highway", "residential"}}}, {"DE", {{"highway", "residential"}}},
+      {"BF", {{"highway", "residential"}}}, {"DG", {{"highway", "residential"}}},
+      {"EH", {{"highway", "residential"}}}, {"FI", {{"highway", "residential"}}},
+      {"GK", {{"highway", "residential"}}}, {"HL", {{"highway", "residential"}}},
+      {"IJ", {{"highway", "residential"}}}, {"JK", {{"highway", "residential"}}},
+      {"KL", {{"highway", "residential"}}},
+  };
+
+  const gurka::nodes nodes = {{"F",
+                               {{"barrier", "sump_buster"},
```
mandeepsandhu, line 96: This is more a question for my understanding: If a mode is not specified here (eg: `hov` or `motorcar`), does that mean this restriction does _not_ apply to it?

In the tests below, none of the modes use `F` (which has `motorcar: yes`). Does that mean `F` does NOT have the sump_buster restriction for anything except a `motorcar`?

mandeepsandhu: Also, this testcase can be shortened a quite a bit is parametrized fixtures are used (since the expected result is the same for all modes).

test/gurka/test_node_access.cc
gknisely: @mandeepsandhu no. This is an access bug that I found during the testing of sump_buster.

test/gurka/test_sump_buster.cc
gknisely: @mandeepsandhu For sump_buster we allow access for bus, truck, motorcycle, moped, foot, and bicycle and block access for auto, hov, and taxi for the defaults.  However, if a user adds a mode specific key then that overrides the default.  This same logic is used for gates, bollards, and barriers.  This is how I found the access issue and why I added a test for it above.

gknisely: > Also, this testcase can be shortened a quite a bit is parametrized fixtures are used (since the expected result is the same for all modes).

I will update.

mandeepsandhu: > @mandeepsandhu For sump_buster we allow access for bus, truck, motorcycle, moped, foot, and bicycle and block access for auto, hov, and taxi for the defaults. However, if a user adds a mode specific key then that overrides the default. This same logic is used for gates, bollards, and barriers. This is how I found the access issue and why I added a test for it above.

Got it. Thanks for the explanation 🙏

test/gurka/test_node_access.cc
gknisely: @mandeepsandhu I just simplified the tests since they are so simple.  Next time I will try out parameterized fixtures.  Ty for the suggestion.

mandeepsandhu (dismissed):
LGTM! 👍

mandeepsandhu (approved):

mandeepsandhu (approved):

PR #2743, opened 2020-12-18
optimizations

kevinkreiser:
1. where ever possible reserve space in pbf structures for the things that are about to go into them
2. convert valhalla format serialization to use rapidjson apis
3. templatize double bucket queue to access edge labels directly rather than via a lambda
4. instead of simple tile cache use flat tile cache which avoids hash lookups and memory indirection
5. templatize doublebucket queue to remove lambda/std::function invokation
6. reduce number of branches in expansion functions by simplifying derived uturn code
7. remove recursion from expansion functions to make call stack shallower

other than tidy this is ready for review. in my tests this nets a 2.5% performance increase in bidirectional a* routes. the only diffs in RAD were the changes in rounding the lengths in the serializer as mentioned in the diff below.


test/summary.cc
```diff
@@ -146,7 +146,7 @@ TEST(Summary, test_time_summary) {
       accumulated_time += maneuver["time"].GetDouble();
     }
     // make sure the end of the trip path is the same as the legs
-    EXPECT_NEAR(accumulated_time, leg["summary"].GetObject()["time"].GetDouble(), .01);
+    EXPECT_NEAR(accumulated_time, leg["summary"].GetObject()["time"].GetDouble(), .02);
```
kevinkreiser, line 149: rapidjsons decimal place logic truncates instead of rounds.

src/baldr/graphreader.cc
```diff
@@ -103,6 +103,67 @@ GraphReader::get_extract_instance(const boost::property_tree::ptree& pt) {
...
+  cache_.emplace_back(std::move(tile));
+  return &cache_.back();
+}
+
+void FlatTileCache::Trim() {
+  Clear();
+}
+
 // ----------------------------------------------------------------------------
 // SimpleTileCache implementation
```
kevinkreiser, line 168: TODO: delete this implementation

run_route_scripts/run_with_server.py
```diff
@@ -22,33 +22,36 @@ def get_post_bodies(filename):
       yield post_body
 
 # for persistant connections
-def initialize(url_, output_dir_):
+def initialize(args):
```
kevinkreiser, line 25: same changes made in the isodistances pr. just to streamline this implementation a bit

src/baldr/graphreader.cc
```diff
@@ -104,6 +104,67 @@ GraphReader::get_extract_instance(const boost::property_tree::ptree& pt) {
   return tile_extract;
 }
 
+// ----------------------------------------------------------------------------
+// FlatTileCache implementation
+// ----------------------------------------------------------------------------
+
+// Constructor.
+FlatTileCache::FlatTileCache(size_t max_size) : cache_size_(0), max_cache_size_(max_size) {
```
kevinkreiser, line 112: instead of hashing we use a vector to store a spot for all the tile pointers that could ever exist. this is slightly faster than hashing and the indirection in unordered_map but at the cost of a around 8 megabytes of ram iirc

src/meili/routing.cc
```diff
@@ -17,8 +17,7 @@ namespace valhalla {
 namespace meili {
 
 LabelSet::LabelSet(const float max_cost, const float bucket_size) {
-  const auto edgecost = [this](const uint32_t label) { return labels_[label].sortcost(); };
-  queue_.reset(new baldr::DoubleBucketQueue(0.0f, max_cost, bucket_size, edgecost));
+  queue_.reset(new baldr::DoubleBucketQueue<Label>(0.0f, max_cost, bucket_size, labels_));
```
kevinkreiser, line 20: lots of changes like this to tell doublebucketqueue what type of labels its being used with

src/proto_conversions.cc
```diff
@@ -138,6 +138,42 @@ const std::string& Options_Action_Enum_Name(const Options::Action action) {
   return i == actions.cend() ? empty : i->second;
 }
 
+bool Location_Type_Enum_Parse(const std::string& type, Location::Type* t) {
```
kevinkreiser, line 141: we were manually doing this before so i added these standard methods for going between strings and enums. choosing `k` named variables keeps us from using the built in protobuf implementations although older versions of protobuf didnt have these methods either in the message lite header

src/thor/costmatrix.cc
```diff
@@ -682,15 +682,11 @@ void CostMatrix::SetSources(GraphReader& graphreader,
...
-    // Set up lambda to get sort costs
-    const auto edgecost = [this, index](const uint32_t label) -> float {
-      return source_edgelabel_[index][label].sortcost();
-    };
-
     // Allocate the adjacency list and hierarchy limits for this source.
     // Use the cost threshold to size the adjacency list.
-    source_adjacency_[index].reset(
-        new DoubleBucketQueue(0, current_cost_threshold_, costing_->UnitSize(), edgecost));
+    source_adjacency_[index].reset(new DoubleBucketQueue<BDEdgeLabel>(0, current_cost_threshold_,
```
kevinkreiser, line 687: we no longer pass a lambda/function to the queue, isntead we give it access to the edge labels directly. one day we should merge the queue the labels and the status objects into one class called queue and do it all from there

src/thor/bidirectional_astar.cc
```diff
@@ -113,7 +104,6 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
                                        const GraphId& node,
                                        BDEdgeLabel& pred,
                                        const uint32_t pred_idx,
-                                       const bool from_transition,
```
kevinkreiser, line 116: we no longer do this recursively so there is no calls that are from transition nodes

src/thor/bidirectional_astar.cc
```diff
@@ -145,70 +133,67 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
...
     // Begin by checking if this is the opposing edge to pred.
     // If so, it means we are attempting a u-turn. In that case, lets wait with evaluating
     // this edge until last. If any other edges were emplaced, it means we should not
     // even try to evaluate a u-turn since u-turns should only happen for deadends
-    if (pred.opp_local_idx() == meta.edge->localedgeidx()) {
-      uturn_meta = meta;
-      found_uturn = true;
-      continue;
-    }
+    uturn_meta = pred.opp_local_idx() == meta.edge->localedgeidx() ? meta : uturn_meta;
```
kevinkreiser, line 146: this may still compile to a branch, but maybe not

src/thor/bidirectional_astar.cc
```diff
@@ -145,70 +133,67 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
...
 
-    disable_uturn = ExpandForwardInner(graphreader, pred, nodeinfo, pred_idx, meta, shortcuts, tile,
-                                       offset_time) ||
+    // Expand but only if this isnt the uturn, we'll try that later if nothing else works out
+    disable_uturn = (pred.opp_local_idx() != meta.edge->localedgeidx() &&
```
kevinkreiser, line 149: since we dropped the `continue` above we can get the same effect by shortcircuiting the call to `ExpandForwardInner` here

src/thor/bidirectional_astar.cc
```diff
@@ -145,70 +133,67 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
...
   }
 
   // Handle transitions - expand from the end node of each transition
-  if (!from_transition && nodeinfo->transition_count() > 0) {
+  if (nodeinfo->transition_count() > 0) {
     const NodeTransition* trans = tile->transition(nodeinfo->transition_index());
     for (uint32_t i = 0; i < nodeinfo->transition_count(); ++i, ++trans) {
-      if (trans->up()) {
-        hierarchy_limits_forward_[node.level()].up_transition_count++;
-        disable_uturn = ExpandForward(graphreader, trans->endnode(), pred, pred_idx, true,
```
kevinkreiser, line 176: here we remove recursion by looping over the edges at the node on other levels directly

src/thor/bidirectional_astar.cc
```diff
@@ -145,70 +133,67 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
...
     }
   }
 
-  if (!from_transition) {
-    // Now, after having looked at all the edges, including edges on other levels,
-    // we can say if this is a deadend or not, and if so, evaluate the uturn-edge (if it exists)
-    if (!disable_uturn && found_uturn) {
-      // If we found no suitable edge to add, it means we're at a deadend
-      // so lets go back and re-evaluate a potential u-turn
+  // Now, after having looked at all the edges, including edges on other levels,
```
kevinkreiser, line 181: this part of the diff looks like a mess but really we had a bunch of branching that was superfluous. instead ive just collapsed it down to only what is needed, a single `if`. the compiler was probably already doing that but honestly this reads way easier anyway

src/thor/bidirectional_astar.cc
```diff
@@ -113,7 +104,6 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
                                        const GraphId& node,
```
kevinkreiser, line 104: all the changes made to this `ExpandForward` were made in `ExpandReverse` and also in the other 2 algorithms `timedep_forward` and `timedep_reverse`

src/thor/route_action.cc
```diff
@@ -361,6 +361,7 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
   std::vector<thor::PathInfo> path;
   auto& correlated = *api.mutable_options()->mutable_locations();
   std::vector<std::string> algorithms;
+  api.mutable_trip()->mutable_routes()->Reserve(api.options().alternates() + 1);
```
kevinkreiser, line 364: just pbf reservations here

src/thor/triplegbuilder.cc
```diff
@@ -65,6 +65,7 @@ void AssignAdmins(const AttributesController& controller,
                   const std::vector<AdminInfo>& admin_info_list) {
   if (controller.category_attribute_enabled(kAdminCategory)) {
     // Assign the admins
+    trip_path.mutable_admin()->Reserve(admin_info_list.size());
```
kevinkreiser, line 68: more reservations here

src/tyr/serializers.cc
```diff
@@ -95,12 +95,28 @@ void route_references(json::MapPtr& route_json, const TripRoute& route, const Op
...
   for (const TripLeg& leg : route.legs()) {
+    references->reserve(references->size() + leg.node_size() - 1);
     for (const std::string& openlr : openlr_edges(leg)) {
       references->emplace_back(openlr);
     }
   }
   route_json->emplace("linear_references", references);
 }
+
+void openlr(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper_t& writer) {
```
kevinkreiser, line 106: since i converted the valhalla serializer to rapidjson i needed a separate function to do openlr additions to the response

valhalla/baldr/json.h
```diff
@@ -47,10 +48,10 @@ class Jmap : public std::unordered_map<std::string, Value> {
 };
 
 // the array value type in json
-class Jarray : public std::list<Value> {
+class Jarray : public std::vector<Value> {
 public:
   // just specialize vector
-  using std::list<Value>::list;
+  using std::vector<Value>::vector;
```
kevinkreiser, line 54: allows for reservation in the serializer. although we shoudl probably just convert the rest of the serializers to rapidjson

valhalla/baldr/rapidjson_utils.h
```diff
@@ -237,6 +237,161 @@ inline std::string serialize(const rapidjson::Document& doc) {
...
+/**
+ * A convenience class to the writer stringbuffer rapidjson pattern. This takes whatever
+ * data you want and serializes it directly to a json string rather than keeping it organized
+ * as json in an in memory object. If you are familiar with XML parsing/writing this is analogous
+ * to a SAX style writer/generator for json, as opposed to DOM (object model in memory) style
+ *
+ * The rapidjson write is pretty completely but its quite verbose for common operations like
+ * adding a key value pair. Because of that we've writting a small wrapper here to make its
+ * use less verbose
+ */
```
kevinkreiser, line 249: the built in rapidjson writer is verbose, this helper reduces that verbosity

valhalla/thor/pathalgorithm.h
```diff
@@ -151,10 +151,19 @@ struct EdgeMetadata {
...
+    return *this;
+  }
+
+  inline operator bool() const {
+    return edge;
+  }
+
+  inline bool operator!() const {
+    return !edge;
   }
```
kevinkreiser, line 167: these operators seemed more idiomatic

src/thor/route_action.cc
```diff
@@ -436,8 +437,10 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
         vias.swap(flipped);
 
         // Form output information based on path edges
-        if (api.trip().routes_size() == 0 || api.options().alternates() > 0)
+        if (api.trip().routes_size() == 0 || api.options().alternates() > 0) {
           route = api.mutable_trip()->mutable_routes()->Add();
+          route->mutable_legs()->Reserve(correlated.size());
```
danpaz, line 442: Could do a little better here by filtering out locations that are `via`.

valhalla/thor/timedep.h
```diff
@@ -248,6 +246,9 @@ class TimeDepReverse : public TimeDepForward {
   // bidirectional edge label structure.
   std::vector<sif::BDEdgeLabel> edgelabels_rev_;
 
+  // Adjacency list - approximate double bucket sort
+  std::shared_ptr<baldr::DoubleBucketQueue<sif::BDEdgeLabel>> adjacencylist_rev_;
```
danpaz, line 250: `adjacencylist_` above got moved to private, should this one also be made private?

src/thor/route_action.cc
kevinkreiser: that is true but what we are optimizing for here is avoiding reallocations. since the number of legs will always be less than or equal to correlated.size() this will accomplish that goal even if it does over estimate how much we need

valhalla/thor/timedep.h
kevinkreiser: i moved that one to private so that the reverse class couldnt see it. this is important because before this change forward and reverse algorithms shared the same adjacencylist. with the template they can no longer do that because they use 2 different types of labels (EdgeLabel and BDEdgeLabel). if anything i should probably make edgelabels_ private in the forward algorithm to prevent confusion there. basically the only thing they share now is edtge status.

i think its all not a big deal at the moment. my ultimate goal is to combine the 3 things that make up a "Queue/Visited" in the conventional sense of the words with respect to classical routing algorithms. we should come up with a decent name for a class that handles those functions then just do the refactor imho

PureW:
Hmm, any chance this could be split into smaller PRs?

src/proto_conversions.cc
```diff
@@ -310,4 +346,67 @@ bool RoadClass_Enum_Parse(const std::string& rc_name, valhalla::RoadClass* rc) {
...
+};
+
+const std::unordered_map<int, std::string> bicycle_to_string{
+    {static_cast<int>(DirectionsLeg_BicycleType_kRoad), "road"},
+    {static_cast<int>(DirectionsLeg_BicycleType_kCross), "cross"},
+    {static_cast<int>(DirectionsLeg_BicycleType_kHybrid), "hybrid"},
+    {static_cast<int>(DirectionsLeg_BicycleType_kMountain), "mountain"},
+};
+
+const std::unordered_map<int, std::string> transit_to_string{
```
PureW, line 371: If these are `switch` statements, the compiler can help you remember to add new enum values.

dnesbitt61 (approved):

src/baldr/graphreader.cc
kevinkreiser: yeah true, its 8.5MB for all levels:

(1036800 (level 3 tiles) + 1036800 (level 2 tiles) + 64800 (level 1)  + 4050 (level 0)) * 4 bytes

Do you think the extra work would be worth the 500k of RAM? imho modern devices are past the point of worrying about those types of ram requirements. I do agree though that we should have a config that lets you select simple if you dont want to pay the 8MB toll on mobile devices. I'll add an issue for that

valhalla/baldr/graphreader.h
```diff
@@ -88,6 +88,89 @@ class TileCache {
...
+    return offset < cache_indices_.size() ? cache_indices_[offset] : -1;
+  }
+
+  // The actual cached GraphTile objects
+  std::vector<graph_tile_ptr> cache_;
+
+  // Indicies into the array of actual cached items
+  std::vector<uint32_t> cache_indices_;
+
+  // Offsets in the idicides list for where a set of tile indices begin
```
dnesbitt61, line 164: spelling of indices

src/baldr/graphreader.cc
dnesbitt61: not worth worrying about 500K RAM, but agree on adding an issue to add a config setting to allow simple cache if desired.

src/proto_conversions.cc
genadz: agree, I'd also choose `switch` or even `std::array` of size `DirectionsLeg_TransitType_TransitType_ARRAYSIZE`

genadz (changes requested):

src/baldr/graphreader.cc
kevinkreiser: ive just pushed a commit to enable this. no issue needed!

src/proto_conversions.cc
kevinkreiser: these were copy pasted from the serializer, i just moved them here because the belonged here organization wise. if we wnat to change the pattern that we have going we should do it for all of them at once. imho we should save that for another PR

src/thor/bidirectional_astar.cc
```diff
@@ -145,70 +133,67 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
...
-        disable_uturn = true;
-      } else {
-        // We didn't add any shortcut of the uturn, therefore evaluate the regular uturn instead
-        bool uturn_added = ExpandForwardInner(graphreader, pred, nodeinfo, pred_idx, uturn_meta,
-                                              shortcuts, tile, offset_time);
-        disable_uturn = disable_uturn || uturn_added;
-      }
-    }
+    // Expand the uturn possiblity
+    bool uturn_added = ExpandForwardInner(graphreader, pred, nodeinfo, pred_idx, uturn_meta,
```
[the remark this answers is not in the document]
kevinkreiser: yep i totally prefer that as well. only reason i didnt do it was to leave it the way it was originally. ill change it!

src/thor/dijkstras.cc
```diff
@@ -57,24 +57,30 @@ void Dijkstras::Clear() {
 // Initialize - create adjacency list, edgestatus support, and reserve
 // edgelabels
 template <typename label_container_t>
-void Dijkstras::Initialize(label_container_t& labels, const uint32_t bucket_size) {
+void Dijkstras::Initialize(
+    label_container_t& labels,
+    std::shared_ptr<baldr::DoubleBucketQueue<typename label_container_t::value_type>>& queue,
```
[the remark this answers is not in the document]
kevinkreiser: the type depends on the method template. i dont think there is a way to do this because a type alias has to happen before the method prototype is declared, which means there is no template to use in the alias.

src/tyr/serializers.cc
```diff
@@ -95,12 +95,28 @@ void route_references(json::MapPtr& route_json, const TripRoute& route, const Op
   }
   json::ArrayPtr references = json::array({});
   for (const TripLeg& leg : route.legs()) {
+    references->reserve(references->size() + leg.node_size() - 1);
```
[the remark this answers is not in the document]
kevinkreiser: the only way they wouldnt be is if we try to serialize an empty leg. that is not possible unless there is a serious bug upstream from here, but i could put an assert if you like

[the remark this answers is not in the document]
kevinkreiser: reorganized it

kevinkreiser:
I've addressed pretty much all the comments here, anyone have time for a quick re-review, basically looking at the last 4 commits to see how the comments were addressed

dnesbitt61 (approved):

PR #2744, opened 2020-12-19
add shortcuts to bins

nilsnolde:
in preparation to work on #2570 

Simply adds shortcuts to the bins, so it's easier to find shortcuts when intersecting with avoid_polygons.

## Tasklist

 - [ ] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [ ] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #2750, opened 2020-12-25
avoid polygons

### LINKED ISSUE #2570: avoid polygons
We'd like to add the ability to avoid entire polygons, because the `avoid_locations` is often not the right thing if the client doesn't have access to road geometries to be able to set exact locations. Would that be interesting for you?

I guess it'd have to be an additional request parameter to not break the existing avoid_locations. E.g. passing a flat list of coordinates representing a ring? Or with the existing `avoid_locations` and a switch (e.g. avoid_polygon=true/false) to let valhalla know the locations represent a polygon?

Any red flags coming to mind? Also, if you're interested, a few pointers would be appreciated. 

I could imagine that an implementation isn't that straight-forward and possibly comes with (quite) a performance hit.

nilsnolde:
fixes #2570 

based on #2744. 

In the request specify `avoid_polygons` with multiple rings (actually MultiPolygon without inner rings..), e.g. `avoid_polygons=[[[0,0],[0,1],[1,1],[1,0],[0,0]]]`.

## Tasklist

- [x] avoid_polygons in request parameters
- [x] actual filtering of AvoidEdges


nilsnolde:
hmm.. I think geographic area calculation might not be supported for boost.geometry 1.65 in circle ci.. Not sure how to fix this, tried to specify a `geographic` strategy.. Would be nice to be able to get polygon/ring area easily. 

@mloskot maybe you have an idea (sometime after xmas;)))? https://github.com/valhalla/valhalla/blob/75ecbdd54b44dc5a8caf82e0fbca777d9a928258/src/loki/polygon_search.cc

mloskot:
@nilsnolde Given the amount of fixes and additions w.r.t. geographic calculations the Boost.Geometry received over last years, the 1.65 is pretty ancient, but I can not tell for sure. Is this still a probolem?

Looks like the latest failures are

```
/root/project/src/worker.cc: In function 'void {anonymous}::from_json(rapidjson::Document&, valhalla::Options&)':
/root/project/src/worker.cc:894:3: error: 'parse_polygons' was not declared in this scope; did you mean 'parse_locations'?
  894 |   parse_polygons(doc, options, "avoid_polygons", 137);
```

kevinkreiser:
im guessing the only thing holding us on an older version of boost is the x86 build, since we cant get newer x86 ubuntu images. perhaps we should just retire the x86 build. we could consider bringing in an arm build in its place that targets a 32bit arm architecture. That seems to be the only modern architecture where i would expect to see devices in the wild that might have to run a 32bit build of valhalla.

nilsnolde:
> Looks like the latest failures are
yeah I pushed smth pre-mature, apparently forgot to change one call.. I think I'll leave out area calculation anyways, as it turns out the better metric here is circumference. I'll see today or tmrw what CI thinks about that..

The old versions complicate dev too.. Guess I'll have to switch developing in a ubuntu container (on arch right now). Anyways, bumping versions would be nice of course. Did you ever try an arm build with valhalla @kevinkreiser ? Wondering if all dependencies are packaged at least..

nilsnolde:
finally works! 
image

removing bg::area might've done the trick for boost.geometry as well I hope.. few todos left, but almost there :champagne:

kevinkreiser:
@nilsnolde since we removed the x86 builds i think we should be safe to update to whatever boost we have in ubuntu 20.04 and whatever version of osx we are running with brew. we should probably do that in a separate pr as well

nilsnolde:
Ok sounds good @kevinkreiser . CI seems to only fail on some stupidity I added in one test. Shouldn’t be a blocker for this PR at least.

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,95 @@
...
+      }
+      for (auto bin_id : line_intersected[tile_id]) {
+        for (auto edge_id : tile->GetBin(bin_id)) {
+          // weed out duplicates early on
+          // TODO: don't iterate over the same bin twice, much more efficient
+          if (result_ids.find(edge_id) != result_ids.end()) {
+            continue;
+          }
+          if (edge_id.Tile_Base() != tile->header()->graphid().Tile_Base() &&
+              reader.GetGraphTile(edge_id, tile)) {
```
kevinkreiser, line 54: we should bail on an edge if we CANT get it because of the tile its in not being available to us
```suggestion
              !reader.GetGraphTile(edge_id, tile)) {
```

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,95 @@
...
+            continue;
+          }
+          const auto edge = tile->directededge(edge_id);
```
kevinkreiser, line 57: we can also avoid checking the shape of an edge for intersection if its an edge that the costing doesnt allow anyway. you'll need to add costing to this interface to do it but its well worth it
```suggestion
          const auto edge = tile->directededge(edge_id);
          
          // grab the opposing
          auto opp_tile = tile;
          baldr::DirectedEdge& opp_edge == nullptr;
          baldr::GraphId opp_id;
          
          // bail if we wouldnt be allowed on this edge anyway (or its opposing)
          if(!costing->Allowed(edge, tile) && (!(opp_id = reader->GetOpposingEdgeId(edge_id, opp_edge, opp_tile)).Is_Valid() || !costing->Allowed(opp_edge, opp_tile)) {
            continue;
          }
```

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,95 @@
...
+          auto edge_info = tile->edgeinfo(edge->edgeinfo_offset());
+          bool intersects = bg::intersects(ring, edge_info.shape());
+          if (intersects) {
+            result_ids.emplace(edge_id);
+            result_ids.emplace(reader.GetOpposingEdgeId(edge_id, tile));
```
kevinkreiser, line 62: if you took my suggestion above you will possibly already have the opposing edge id. also you shouldnt re-use `tile` here as it can be mutated (i think you mentioned that you found this out already, maybe you didnt push it up yet)
```suggestion
            result_ids.emplace(opp_id.Is_Valid() ? opp_id : reader->GetOpposingEdgeId(edge_id, opp_edge, opp_tile));
```

src/loki/polygon_search.cc
nilsnolde: where did the negation go?! weird.. thx!

nilsnolde:
Should be done now @kevinkreiser . I'll comment thoroughly on the PR in a few hours.

scripts/list_tiles.py
```diff
@@ -73,10 +73,10 @@ def Col(self, x):
 
   def Digits(self, number):
     digits = 1 if (number < 0) else 0
-    while long(number):
+    while number:
```
nilsnolde, line 76: the concept of `long` is deprecate in py3

scripts/valhalla_build_config
```diff
@@ -264,7 +264,8 @@ config = {
     'max_reachability': 100,
     'max_radius': 200,
     'max_timedep_distance': 500000,
-    'max_alternates': 2
+    'max_alternates': 2,
+    'max_avoid_polygons_length': 10
```
nilsnolde, line 268: The real driver for performance here is how many bins are intersected. With this max circumference one can intersect around max 10 bins (depending on latitude), which (in urban environments) adds a good 2-3 seconds to the request. Assuming it's a single polygon. One can also do 20 mini polygons each intersecting another bin. I don't know what you think, maybe we should add a max param for amount of polygons passed?

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,115 @@
...
+  // first pull out all *unique* bins which intersect the rings
+  for (size_t ring_idx = 0; ring_idx < rings.size(); ring_idx++) {
+    auto ring = rings[ring_idx];
+    auto line_intersected = tiles.Intersect(ring);
+    for (const auto& tb : line_intersected) {
+      for (const auto& b : tb.second) {
+        bins_intersected[tb.first][b].push_back(ring_idx);
+      }
+    }
+  }
```
nilsnolde, line 30: what the comment says. we do need to record for each bin which ring**s** it intersects, as that could be multiple and that's important for intersecting the edges within later on.

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,115 @@
...
+        // TODO: is this really worth it?
+        if (avoid_edge_ids.find(edge_id) != avoid_edge_ids.end()) {
+          continue;
+        }
+        // TODO: optimize the tile switching by enqueuing edges
+        // from other levels & tiles and process them after this big loop
+        if (edge_id.Tile_Base() != tile->header()->graphid().Tile_Base() &&
+            !reader.GetGraphTile(edge_id, tile)) {
+          continue;
+        }
```
nilsnolde, line 60: not sure how far one can optimise here.. as then the same optimization might be applied to getting the opp edges?

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,115 @@
...
+
+        // bail if we wouldnt be allowed on this edge anyway (or its opposing)
+        if (!costing->Allowed(edge, tile) &&
+            (!(opp_id = reader.GetOpposingEdgeId(edge_id, opp_edge, opp_tile)).Is_Valid() ||
+             !costing->Allowed(opp_edge, opp_tile))) {
+          continue;
+        }
+
+        // TODO: some logic to set percent_along for origin/destination edges
+        // careful: polygon can intersect a single edge multiple times
```
nilsnolde, line 74: I guess the logic for applying `percent_along` could be fairly easy.. switching `intersects` with `intersection`: https://www.boost.org/doc/libs/1_75_0/libs/geometry/doc/html/geometry/reference/algorithms/intersection/intersection_3.html

Not sure how much more expensive that would be though.. Any comments/ideas?

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,115 @@
...
+        // and needs to be applied to bins which are entirely inside a ring as well
+        auto edge_info = tile->edgeinfo(edge->edgeinfo_offset());
+        auto name = edge_info.GetNames()[0];
+        bool intersects = false;
+        for (const auto& ring_loc : bin.second) {
+          intersects = bg::intersects(rings[ring_loc], edge_info.shape());
+          if (intersects) {
+            break;
+          }
+        }
```
nilsnolde, line 84: ughh.. there's gotta be a nicer pattern for this in c++

src/loki/worker.cc
```diff
@@ -83,13 +84,35 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
...
-  // See if we have avoids and take care of them
-  if (static_cast<size_t>(options.avoid_locations_size()) > max_avoid_locations) {
-    throw valhalla_exception_t{157, std::to_string(max_avoid_locations)};
+  if (options.avoid_polygons_size()) {
+    double rings_length = 0;
+    std::vector<ring_bg_t> rings;
+    for (const auto& ring_pbf : options.avoid_polygons()) {
+      const ring_bg_t ring = PBFToRing(ring_pbf);
+      rings.push_back(ring);
+      rings_length += GetRingLength(ring) * kKmPerMeter;
```
nilsnolde, line 93: I could make those functions accept a `std::vector<ring_bg_t>`, but thought it might make them more generic this way.

src/loki/worker.cc
```diff
@@ -83,13 +84,35 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
...
+      avoid->set_percent_along(0);
+    }
   }
 
   // Process avoid locations. Add to a list of edgeids and percent along the edge.
   if (options.avoid_locations_size()) {
+    // See if we have avoids and take care of them
+    if (static_cast<size_t>(options.avoid_locations_size()) > max_avoid_locations) {
+      throw valhalla_exception_t{157, std::to_string(max_avoid_locations)};
+    }
```
nilsnolde, line 115: felt more appropriate to move it into the condition

src/loki/worker.cc
```diff
@@ -107,6 +130,7 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
             avoid->set_id(edge.id);
             avoid->set_percent_along(edge.percent_along);
 
+            // TODO: check if still necessary here since shortcuts are in bins now!
```
nilsnolde, line 133: didn't give that a whole lot of thought and to some degree I refused to understand the logic in loki::Search ;) I'm sure you could judge that easier

src/worker.cc
```diff
@@ -318,6 +318,35 @@ void add_date_to_locations(Options& options,
   }
 }
 
+// Parses JSON rings of the form [[lon1, lat1], [lon2, lat2], ...]] and operates on
+// PBF objects of the sort "repeated Location". Rings can be optionally closed.
+template <typename ring_pbf_t>
+void parse_ring(ring_pbf_t& ring,
+                rapidjson::Value& coord_array,
+                rapidjson::Document::AllocatorType& allocator,
+                bool close_ring = true) {
```
nilsnolde, line 327: again, tried to keep it somewhat generic

test/gurka/gurka.cc
```diff
@@ -117,7 +117,32 @@ std::string build_valhalla_request(const std::string& location_type,
...
+        a.PushBack(a1, allocator);
+      }
+      // Workaround end
+      // setting d directly somehow messes up the very first coordinate's X... bug in rapidjson? can't
+      // reproduce in isolated test..
+      // rapidjson::SetValueByPointer(doc, "/avoid_polygons", d.GetArray());
+      rapidjson::Pointer(kv.first).Set(doc, a);
+    } else {
+      rapidjson::Pointer(kv.first).Set(doc, kv.second);
+    }
```
nilsnolde, line 145: disgusting I know.. I tried the conventional rapidjson way, what happened then was that it for some weird reason turned a coordinate string of `[[[13.51031, 52.234141], [13.53159, 52.24115]...]]` into `[[[-300e-52, 52.234141], [13.53159, 52.24115]...]]`, i.e. the very first X coord got completely messed up.. I tried to debug that for half a day and eventually did this abomination..

Either I was doing smth really stupid or it's a bug in rapidjson. Maybe someone more familiar with rapidjson can have a look? (@mandeepsandhu or @danpat ?)

valhalla/loki/polygon_search.h
```diff
@@ -0,0 +1,67 @@
...
+#include <valhalla/sif/dynamiccost.h>
+
+#include <functional>
+
+// TODO: register PointLL in its own header, so it's available for bg across valhalla?
+BOOST_GEOMETRY_REGISTER_POINT_2D(valhalla::midgard::PointLL,
+                                 double,
+                                 boost::geometry::cs::geographic<boost::geometry::degree>,
+                                 first,
+                                 second)
```
nilsnolde, line 21: you wanna keep it contained here or put the registration into `PointLL` header?

nilsnolde:
slight bump here. maybe someone else than kevin could have a look (poor guy had to do all my reviews so far..)

src/loki/CMakeLists.txt
```diff
@@ -11,7 +11,8 @@ set(sources
   isochrone_action.cc
   trace_route_action.cc
   transit_available_action.cc
-  node_search.cc)
+  node_search.cc
+  polygon_search.cc)
```
mandeepsandhu, line 15: Since this is a new file, can you also add it to [`set_source_files_properties` below](https://github.com/valhalla/valhalla/pull/2750/files#diff-7552fd95e00220b1c45ebce4ff7c0c24e47586fa3feff85f77cbde5dcab65c79R20) so that we have stricter compiler checks enabled for it?

We're doing it piece-meal for existing files that we're making changes to (to reduce diff noise from just warning fixes)

CHANGELOG.md
```diff
@@ -258,6 +259,7 @@
    * ADDED: Added support to process the sump_buster tag.  Also, fixed a few small access bugs for nodes. [#2731](https://github.com/valhalla/valhalla/pull/2731)
    * ADDED: Log message if failed to create tiles directory. [#2738](https://github.com/valhalla/valhalla/pull/2738)
    * CHANGED: Tile memory is only owned by the GraphTile rather than shared amongst copies of the graph tile (in GraphReader and TileCaches). [#2340](https://github.com/valhalla/valhalla/pull/2340)
+   * CHANGED: Add shortcuts to the tiles' bins so we can treat them like regular edges. [#2744](https://github.com/valhalla/valhalla/pull/2744)
```
kevinkreiser, line 262: this was already merged so its probably a duplicate

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
    * FIXED: Loki was looking up the wrong costing enum for avoids [#2856](https://github.com/valhalla/valhalla/pull/2856)
 
 * **Enhancement**
+   * ADDED: avoid_polygons logic [#2750](https://github.com/valhalla/valhalla/pull/2750)
```
kevinkreiser, line 20: can we move this to the bottom of the list to m aintain the ordering

proto/options.proto
```diff
@@ -166,6 +166,10 @@ message Options {
     invariant = 3;
   }
 
+  message Polygon {
+    repeated Location coords = 1;
```
kevinkreiser, line 170: this is an important decision. the location object has with it all kinds of other options you can add. im pretty sure for this use-case we only ever just want the lat,lon but if we were to not use location we would need to do different parsing. anyway just something to think about

scripts/valhalla_build_config
kevinkreiser: in other files, i see this value defaulting to `100` can we make them all match, it creates less surprises

scripts/valhalla_build_config
```diff
@@ -523,7 +524,8 @@ help_text = {
     'max_reachability': 'Maximum reachability (number of nodes reachable) allowed on any one location',
     'max_radius': 'Maximum radius in meters allowed on any one location',
     'max_timedep_distance': 'Maximum b-line distance between locations to allow a time-dependent route',
-    'max_alternates': 'Maximum number of alternate routes to allow in a request'
+    'max_alternates': 'Maximum number of alternate routes to allow in a request',
+    'max_avoid_polygons_length': 'Maximum total circumference of all avoid_polygons in kilometers'
```
kevinkreiser, line 528: i would suggest the word `perimeter` in place of `circumference` here

src/loki/worker.cc
kevinkreiser: `loki::search` uses regular costing allowed function which by default disallowes shortcuts. i think this is still good because we actually want to find both the regular edges and the shortcuts and `search` wont do that unless you provide a radius and change the allowed function. so yeah no change here for now but somethjign we could think about encapulating rather than having this detail outside of the fucntion

proto/options.proto
kevinkreiser: hmm looks like you already have custom parsing. i would suggest changing this to LatLng object instead and not carry around all that baggage that comes with Location

nilsnolde: Call it `SimplePolygon` to leave the door open for more complex types?

test/gurka/gurka.cc
kevinkreiser: i think in the calling code you have to put something like 

```c++
{
  {"/avoid_polygons/0/0/0", lon0},
  {"/avoid_polygons/0/0/1", lat0},
  {"/avoid_polygons/0/1/0", lon1},
  {"/avoid_polygons/0/1/1", lat1},
  ...
}
```

nilsnolde: of course :facepalm:

kevinkreiser:
Sorry for the delay on this I have it on my list but havent gotten to it. The main thing I wanted to point out that if you have a decently large polygon and you put an origin or destination inside it you can fail the route. maybe we dont do it for the first pass but i think we need to actually also mark edges inside of the polygons. I have some ideas on how to do that.

I made a bunch of comments for cleanup and I still need to re-review the guts of the new intersection code after the first round but it shouldnt take long. I'll check back later tonight, sorry again for the delay

nilsnolde:
all good @kevinkreiser , thanks a bunch!

test/gurka/test_avoids.cc
```diff
@@ -0,0 +1,287 @@
...
+  rapidjson::Value locs_j(rapidjson::kArrayType);
+  for (auto& loc : locs) {
+    rapidjson::Value p(rapidjson::kObjectType);
+    p.AddMember("lon", loc.lng(), allocator);
+    p.AddMember("lat", loc.lat(), allocator);
+    locs_j.PushBack(p, allocator);
+  }
+
+  return locs_j;
+}
```
nilsnolde, line 31: the current logic for adding costing options doesn't accommodate for other value types than `string` (ie I need to add `double` as value with the pointer syntax). I didn't wanna mess with that, so I basically roll my own here. ain't saying it's prettier than before, but at least it's only ugly in a very confined space..

test/gurka/test_avoids.cc
```diff
@@ -0,0 +1,287 @@
...
+
+  // make sure the right exception is thrown
+  try {
+    gurka::do_action(Options::route, avoid_map, req);
+  } catch (const valhalla_exception_t& err) { EXPECT_EQ(err.code, 442); } catch (...) {
+    FAIL() << "Expected valhalla_exception_t.";
+  };
+}
+
+TEST_P(AvoidTest, TestAvoidLocation) {
```
nilsnolde, line 247: thought I'd throw that in too, while I'm at it..

test/gurka/gurka.h
```diff
@@ -97,6 +97,13 @@ void build_pbf(const nodelayout& node_locations,
  * @return list of edge names
  */
 std::vector<std::string> get_path(const valhalla::Api& result);
+
+/** Turns layout nodes into PointLLs.
+ * @param nodes the node layout map
+ * @param node_names the node names to associate with layout nodes
+ */
+std::vector<midgard::PointLL> to_lls(const nodelayout& nodes,
+                                     const std::vector<std::string>& node_names);
```
nilsnolde, line 106: needed to make this visible for the custom request builder in the test..

nilsnolde:
> maybe we dont do it for the first pass but i think we need to actually also mark edges inside of the polygons.

Did you have a chance to think about this @kevinkreiser ? As for the details: I think it makes sense to allow **origin** waypoints inside an avoid_polygon (case where your car is parked since eons and in the meantime someone decided to close off the entire area around it). However, all other waypoints (intermediate & destination): why you'd like to allow those to be inside avoid_polygons? If your origin is also inside a polygon, then it makes sense. But if not, IMHO it'd defy the purpose of avoid_polygons a little if you could still easily route inside of them from the outside. Bit of a style question as well I guess.

I think there'd be more to-dos:
- [ ] origin can be inside polygons
- [ ] higher accuracy setting `percent_along` properly
- [ ] possibly quite some performance enhancements

I'd like to help out on these as well, as far as I can. 

For now: would you be ok merging this PR in the state it is in right now? Not accounting for gross misunderstandings & bugs of course..

kevinkreiser:
@nilsnolde yep ive been slacking on this one. i'll review the guts again tonight and we'll get it merged barring any larger issues. sorry for the delay

test/matrix_bss.cc
```diff
@@ -48,47 +48,8 @@ valhalla::sif::cost_ptr_t create_costing() {
...
-      		"pedestrian": {"max_distance": 250000.0,"max_locations": 50,"max_matrix_distance": 200000.0,"max_matrix_locations": 50,"max_transit_walking_distance": 10000,"min_transit_walking_distance": 1},
-      		"skadi": {"max_shape": 750000,"min_resample": 10.0},
-      		"trace": {"max_distance": 200000.0,"max_gps_accuracy": 100.0,"max_search_radius": 100,"max_shape": 16000,"max_best_paths":4,"max_best_paths_shape":100},
-      		"transit": {"max_distance": 500000.0,"max_locations": 50,"max_matrix_distance": 200000.0,"max_matrix_locations": 50},
-      		"truck": {"max_distance": 5000000.0,"max_locations": 20,"max_matrix_distance": 400000.0,"max_matrix_locations": 50},
-	        "bikeshare": {"max_distance": 5000000.0,"max_locations": 20,"max_matrix_distance": 400000.0,"max_matrix_locations": 50}
-    		}
-	    })");
 
+const auto config = test::make_config("test/data/paris_bss_tiles");
```
kevinkreiser, line 52: how the hell did i miss this one, thanks!

src/worker.cc
```diff
@@ -94,7 +94,7 @@ const std::unordered_map<unsigned, unsigned> ERROR_TO_STATUS{
 
     {120, 400}, {121, 400}, {122, 400}, {123, 400}, {124, 400}, {125, 400}, {126, 400}, {127, 400},
 
-    {130, 400}, {131, 400}, {132, 400}, {133, 400}, {136, 400},
+    {130, 400}, {131, 400}, {132, 400}, {133, 400}, {136, 400}, {137, 400},
```
kevinkreiser, line 97: you need to also add mappings to the OSRM flavored errors below

kevinkreiser: also why isnt there a new entry for 167? its already got a mapping below but it looks new in the header file :thinking:

src/loki/worker.cc
```diff
@@ -83,13 +84,35 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
...
 
-  // See if we have avoids and take care of them
-  if (static_cast<size_t>(options.avoid_locations_size()) > max_avoid_locations) {
-    throw valhalla_exception_t{157, std::to_string(max_avoid_locations)};
+  if (options.avoid_polygons_size()) {
+    double rings_length = 0;
+    std::vector<ring_bg_t> rings;
+    for (const auto& ring_pbf : options.avoid_polygons()) {
+      const ring_bg_t ring = PBFToRing(ring_pbf);
+      rings.push_back(ring);
```
kevinkreiser, line 92: this avoids any copies

```suggestion
      rings.push_back(PBFToRing(ring_pbf));
      const ring_bg_t& ring = rings.back();
```

src/worker.cc
```diff
@@ -893,6 +922,18 @@ void from_json(rapidjson::Document& doc, Options& options) {
   // get the avoids in there
   parse_locations(doc, options, "avoid_locations", 133, ignore_closures);
 
+  // get the avoid polygons in there
+  auto rings_req = rapidjson::get_child_optional(doc, "/avoid_polygons");
+  if (rings_req) {
+    auto* rings_pbf = options.mutable_avoid_polygons();
+    try {
+      for (auto& req_poly : rings_req->GetArray()) {
```
kevinkreiser, line 930:
```suggestion
      for (const auto& req_poly : rings_req->GetArray()) {
```

kevinkreiser: why didnt the linter get that?

src/worker.cc
kevinkreiser, line 327:
```suggestion
                const rapidjson::Value& coord_array) {
```

src/worker.cc
```diff
@@ -318,6 +318,35 @@ void add_date_to_locations(Options& options,
...
+// PBF objects of the sort "repeated Location". Rings can be optionally closed.
+template <typename ring_pbf_t>
+void parse_ring(ring_pbf_t& ring,
+                rapidjson::Value& coord_array,
+                rapidjson::Document::AllocatorType& allocator,
+                bool close_ring = true) {
+  // optionally close the ring
+  if (coord_array[0] != coord_array[coord_array.Size() - 1] && close_ring) {
+    coord_array.PushBack(coord_array[0], allocator);
+  }
```
kevinkreiser, line 331: isntead of this we can at the end close the ring if its not closed
```suggestion
```

src/worker.cc
```diff
@@ -318,6 +318,35 @@ void add_date_to_locations(Options& options,
...
+    lon = midgard::circular_range_clamp<double>(lon, -180, 180);
+    double lat = coords[1].GetDouble();
+    if (lat < -90.0 || lat > 90.0) {
+      throw std::runtime_error("Latitude must be in the range [-90, 90] degrees");
+    }
+
+    auto* ll = ring->add_coords();
+    ll->set_lng(lon);
+    ll->set_lat(lat);
+  }
```
kevinkreiser, line 347:
```suggestion
  }
  
  // close the ring if it wasnt already
  if (!ring->coords().empty() && *ring->coords().begin() != *ring.coords()->rbegin()) {
    ring->add_coords()->CopyFrom(*ring->coords().begin());
  }
```

test/gurka/gurka.h
kevinkreiser: i would revert this change. i'll comment more later in the test

test/gurka/test_avoids.cc
```diff
@@ -0,0 +1,287 @@
...
+}
+
+TEST_F(AvoidTest, TestConfig) {
+  // Add a polygon with longer perimeter than the limit
+  std::vector<vl::ring_bg_t> rings{{{13.38625361, 52.4652558},
+                                    {13.38625361, 52.48000128},
+                                    {13.4181769, 52.48000128},
+                                    {13.4181769, 52.4652558}}};
+
+  auto lls = gurka::detail::to_lls(avoid_map.nodes, {"A", "D"});
```
kevinkreiser, line 149: i'm pretty sure the compiler will deduce a `std::vector` type here without you supplying it but if it complains just add that to the left of the first `{`

```suggestion
  auto lls = {avoid_map.nodes["A"], avoid_map.nodes["D"]};
```

valhalla/loki/polygon_search.h
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Finds all edge IDs which are intersected by the ring
+ *
+ * @param rings The (optionally closed) rings to intersect edges with
+ * @param reader GraphReader instance
+ *
+ */
+std::unordered_set<valhalla::baldr::GraphId>
+edges_in_rings(const std::vector<ring_bg_t>& rings,
```
kevinkreiser, line 44: it would be great if this interface were only this single function. here's what im thinking. im not a fan of having the guts of conversion in loki/worker.cc. aside: i know we do that for locations but its not because we wanted to, its because we didnt used to have protobuf when the project first started and we just havent gotten to converting loki::search to work with protobuf natively.

anyway.. what if we change this funciton signature to be:

```c++
std::unordered_set<valhalla::baldr::GraphId>
edges_in_rings(const protobuf::repeated<Ring>& rings, //that typename isnt right i just dont remember right now
               baldr::GraphReader& reader,
               float max_length);
```

then this function could do some checks at the beginning and throw if any ring is too crazy large. it can still throw the same valhalla exception. and we can take the other two functions that are here and move them into an anonymous name space in the cc file to remove them from the public interface. i just prefer to encapsulate related functionality as much as we can rather than spread it around. we want to have as small an public interface to each class/namespace etc as possible so it keeps the spider webbing down :smile:

src/loki/polygon_search.cc
kevinkreiser: its a good question. we can jump over to the source code to see if intersects calls intersection, if it does i assume its just the same cost. we would also need to know what sides of the intersection points are inside and outside of the polygon which is totally doable but feels pedantic at this point. my recommendation is TODO :smile:

src/worker.cc
nilsnolde: I think that's a (unnoticed) "artifact" from the shortcuts-to-bins PR

src/worker.cc
nilsnolde: > i assume boost geometry requires this like most systems?

Actually bg is surprisingly flexible with its inputs validation, it gives you the option if a ring should be closed or not: (https://www.boost.org/doc/libs/1_75_0/libs/geometry/doc/html/geometry/reference/concepts/concept_polygon.html):

We register our custom type with a macro though, I assume it takes the model's default values to do so. Meaning our ring type does expect to be closed I think.

valhalla/loki/polygon_search.h
nilsnolde: Ok understood! 
My reasoning was that these other 2 functions might be useful at some point so I tried to make them a little generic. In case there's other polygon/ring stuff you'd like to care of in the future. But of course it's easily refactored once that's actually the case.

kevinkreiser: yep exactly my thoughts, when it becomes useful we can expand the public interface

src/worker.cc
nilsnolde: Even though it should be possible to compare protobuf messages, it threw an error along the lines of `no conversion to protobuf::StringPiece` coupled with a `no operator==` error, so I resorted to comparing lat & long individually, see below

valhalla/loki/polygon_search.h
nilsnolde: Done

test/gurka/test_avoids.cc
nilsnolde: worked

nilsnolde:
ok @kevinkreiser included all your review comments finally. Let me know when you wanna talk about the actual work horse of the PR.

scripts/valhalla_build_config
```diff
@@ -527,7 +528,8 @@ help_text = {
     'max_reachability': 'Maximum reachability (number of nodes reachable) allowed on any one location',
     'max_radius': 'Maximum radius in meters allowed on any one location',
     'max_timedep_distance': 'Maximum b-line distance between locations to allow a time-dependent route',
-    'max_alternates': 'Maximum number of alternate routes to allow in a request'
+    'max_alternates': 'Maximum number of alternate routes to allow in a request',
+    'max_avoid_polygons_length': 'Maximum total perimeter of all avoid_polygons in kilometers'
```
kevinkreiser, line 532: nit: everything else is in meters except the output of the route response. we should probably go with meters units?

src/loki/worker.cc
```diff
@@ -83,13 +84,26 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
     }
   } catch (const std::runtime_error&) { throw valhalla_exception_t{125, "'" + costing_str + "'"}; }
 
-  // See if we have avoids and take care of them
-  if (static_cast<size_t>(options.avoid_locations_size()) > max_avoid_locations) {
-    throw valhalla_exception_t{157, std::to_string(max_avoid_locations)};
+  if (options.avoid_polygons_size()) {
+    // TODO: should the PBF edge assignment happen in edges_in_rings?
+    // feels like a waste to produce an unordered_set we just iterate over right after
```
kevinkreiser, line 89: i agree with the sentiment here but theres nothing else we can do, protobuf doesnt support hashsets natively and we kind of need the deduplication when we are doing the intersection. seems theres not a good clean solution here. i thikn what you have is reasonable tbh

src/worker.cc
```diff
@@ -325,6 +329,36 @@ void add_date_to_locations(Options& options,
   }
 }
 
+// Parses JSON rings of the form [[lon1, lat1], [lon2, lat2], ...]] and operates on
+// PBF objects of the sort "repeated Location". Rings can be optionally closed.
```
kevinkreiser, line 333: nit: comment is out of date now as its `repeated LatLng`

src/worker.cc
kevinkreiser: yeah thats another drawback with protobuf. since at the end of the day every type is inherited from the`message` object the most straight forward way to check equality would be serialize both and check the resulting strings are the same. failing that one would need to do some reflection. both of these options sound slow so i bet they opted just not to support it with out maybe an explicit conversion.

valhalla/loki/polygon_search.h
kevinkreiser: i would move as much as i can out of the header and into the source file (polygon_search.cc). ive not tried it but it would be great if it all could be sequestered to there.. same with line_bg_t and ring_bg_t. i havent tried this myself but it seems like the function below neither takes any of these as input nor returns them as output so you should be able to move them cleanly? im not exactly sure what the macro does but yeah put it in that single compilation unit would be the nicest way to manage it if it can still work that way

src/worker.cc
nilsnolde: doesn't sound like a todo here, so I left it as-is

nilsnolde:
Thanks for the further review @kevinkreiser! Think I could change it all as you recommended.. Let me know when you looked at the core functionality..

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,158 @@
...
+BOOST_GEOMETRY_REGISTER_RING(std::vector<vm::PointLL>)
+
+namespace {
+// register a few boost.geometry types
+using line_bg_t = bg::model::linestring<vm::PointLL>;
+using ring_bg_t = std::vector<vm::PointLL>;
+
+// map of tile for map of bin ids & their ring ids
+using bins_collector =
+    std::unordered_map<uint32_t, std::unordered_map<unsigned short, std::vector<size_t>>>;
```
kevinkreiser, line 29: i believe we could simplify this a bit but we can just leave that as a TODO for future work. fine the way it is for now

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,158 @@
...
+  std::vector<uint32_t> tile_keys(bins_intersected.size());
+  std::for_each(bins_intersected.begin(), bins_intersected.end(),
+                [&tile_keys](
+                    const std::pair<uint32_t,
+                                    std::unordered_map<unsigned short, std::vector<size_t>>>& e) {
+                  tile_keys.push_back(e.first);
+                });
+  std::sort(tile_keys.begin(), tile_keys.end());
+  for (const auto& tile_id : tile_keys) {
+    auto& tile_bins = bins_intersected[tile_id];
```
kevinkreiser, line 102: seems superfluous
```suggestion
```

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,158 @@
...
+    auto tile = reader.GetGraphTile({static_cast<uint32_t>(tile_id), bin_level, 0});
+    if (!tile) {
+      continue;
+    }
+    for (const auto& bin : tile_bins) {
```
kevinkreiser, line 107:
```suggestion
    for (const auto& bin : bins_intersected[tile_id]) {
```

kevinkreiser, line 107: or you could skip the sorting and just do:

```suggestion
  for (const auto& intersection : bins_intersected) {
    auto tile = reader.GetGraphTile({static_cast<uint32_t>(intersection.first), bin_level, 0});
    if (!tile) {
      continue;
    }
    for (const auto& bin : intersection.second) {
```

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,158 @@
...
+      // tile will be mutated most likely in the loop
+      if (tile->id().tileid() != tile_id) {
+        tile = reader.GetGraphTile({static_cast<uint32_t>(tile_id), bin_level, 0});
```
kevinkreiser, line 110: graphreader has a shortcut for this:

```suggestion
        reader.GetGraphTile({static_cast<uint32_t>(tile_id), bin_level, 0}, tile);
```

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,158 @@
...
+      }
```
kevinkreiser, line 111: graphreader has a shortcut for this
```suggestion
reader.GetGraphTile({static_cast<uint32_t>(tile_id), bin_level, 0}, tile);
```

src/loki/polygon_search.cc
```diff
@@ -0,0 +1,158 @@
...
+      for (const auto& edge_id : tile->GetBin(bin.first)) {
+        // weed out duplicates early on
+        // TODO: is this really worth it?
+        if (avoid_edge_ids.find(edge_id) != avoid_edge_ids.end()) {
```
kevinkreiser, line 115: i think its worth it as it should be O(1) and it avoids having to get the opposing edge below. for something more succinct you can use `avoid_edge_ids.count(edge_id) != 0` if you prefer

kevinkreiser:
this looks done to me. i just added a couple of tiny nits to the main algorithm. as  soon as you resolve those ill drop a ship here!

src/loki/polygon_search.cc
nilsnolde: jep, did that instead, much more elegant

kevinkreiser (approved):

PureW:
Just flagging that the unittests in this PR fail on newest Clang
```
[----------] 27 tests from AvoidPolyProfilesTest/AvoidTest                                                                             
[          ] generating map PBF at test/data/gurka_avoids/map.pbf                                                                      
[          ] building tiles in test/data/gurka_avoids                                                                                  
[ RUN      ] AvoidPolyProfilesTest/AvoidTest.TestAvoidPolygonWithDeprecatedParam/0                                                     
[          ] Valhalla request is: {"locations":[{"lon":0.00017966240894892044,"lat":0.0},{"lon":0.00017966240894892044,"lat":-0.0005389
872268353854}],"costing":"auto","avoid_polygons":[[[0.0,-0.00017966240895360996],[0.0003593248178978409,-0.00017966240895360996],[0.000
3593248178978409,-0.00035932481790721993],[0.0,-0.00035932481790721993]]]}                                                             
[       OK ] AvoidPolyProfilesTest/AvoidTest.TestAvoidPolygonWithDeprecatedParam/0 (400 ms)                                            
[ RUN      ] AvoidPolyProfilesTest/AvoidTest.TestAvoidPolygonWithDeprecatedParam/1                                                     
[          ] Valhalla request is: {"locations":[{"lon":0.00017966240894892044,"lat":0.0},{"lon":0.00017966240894892044,"lat":-0.0005389
872268353854}],"costing":"truck","avoid_polygons":[[[0.0,-0.00017966240895360996],[0.0003593248178978409,-0.00017966240895360996],[0.00
03593248178978409,-0.00035932481790721993],[0.0,-0.00035932481790721993]]]}                                                            
[       OK ] AvoidPolyProfilesTest/AvoidTest.TestAvoidPolygonWithDeprecatedParam/1 (21 ms)                                             
[ RUN      ] AvoidPolyProfilesTest/AvoidTest.TestAvoidPolygonWithDeprecatedParam/2                                                     
[          ] Valhalla request is: {"locations":[{"lon":0.00017966240894892044,"lat":0.0},{"lon":0.00017966240894892044,"lat":-0.0005389
872268353854}],"costing":"bicycle","avoid_polygons":[[[0.0,-0.00017966240895360996],[0.0003593248178978409,-0.00017966240895360996],[0.
0003593248178978409,-0.00035932481790721993],[0.0,-0.00035932481790721993]]]}                                                          
[       OK ] AvoidPolyProfilesTest/AvoidTest.TestAvoidPolygonWithDeprecatedParam/2 (21 ms)                                             
[ RUN      ] AvoidPolyProfilesTest/AvoidTest.TestAvoidPolygonWithDeprecatedParam/3
```

I haven't dug deeper into why this fails
```
clang++ --version
clang version 12.0.0
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

nilsnolde:
I don't see failing tests in the snippet.. sure you pasted the right part @PureW ?

this particular test `TestAvoidPolygonWithDeprecatedParam` was added only fairly recently with #3093 , though also by me..

PureW:
Woops, updated the snippet above.

PR #2752, opened 2020-12-28
fix python bindings installation

### LINKED ISSUE #2751: Python bindings are not properly installed
I forgot to change the installation stuff when I switched the python bindings to pybind11 in #2644 

When installing Valhalla, the python bindings are not importable `import valhalla` bcs the `__init__.py` is not installed.. Will PR a fix now..

nilsnolde:
fixes #2751 

## Tasklist


nilsnolde:
codecov stupid stuff again..

nilsnolde:
completely superseded by #2754

nilsnolde:
regardless of how #2754 will turn out, I think this should be merged. currently python won't get properly installed.

kevinkreiser:
@nilsnolde it looks ok ot me but id like to do this myself in a container to check out the site-packages dir and stuff just to see what it does in terms of installation. you wouldn't happen to already have like a directory listing and firing off a quick oneliner python command to load up the module?

nilsnolde:
@kevinkreiser It's still not perfect, it'll look like this:

```
/usr/lib/python3.9/site-packages
...
├── python_valhalla.cpython-39-x86_64-linux-gnu.so
└── valhalla
        └── __init__.py
```
and __init__.py does `from python_valhalla import *`. ideally python_valhalla.so lives inside `valhalla` as well and the init does the relative import. which is what I did in #2754 as well but didn't port here (yet?).

Once it's installed just do `import valhalla; valhalla.Configure('path/to/json') ; a = valhalla.Actor()`. 

You can try our image with build arg `VALHALLA_RELEASE=nn_python_install_fix`: https://github.com/gis-ops/docker-valhalla (ubuntu 20.04)

nilsnolde:
Now it's much better, after installation:
```
/usr/lib/python3.9/site-packages
...
└── valhalla
        ├── __init__.py
        ├── python_valhalla.cpython-39-x86_64-linux-gnu.so
        └── utils.py
```
I also included a utility to decode polylines, plus a test for it.

src/bindings/python/utils.py
```diff
@@ -0,0 +1,58 @@
...
+        byte = ord(value[index]) - 63
+        index += 1
+        result |= (byte & 0x1f) << shift
+        shift += 5
+        comp = result & 1
+
+    return ~(result >> 1) if comp else (result >> 1), index
+
+
+def _decode(expression, precision=5, order='lnglat', is3d=False):
```
kevinkreiser, line 33: we already have sample code for this in python in the docs in this repo, should probably go with that unless its not working for you: https://github.com/valhalla/valhalla/blob/master/docs/decoding.md#python

kevinkreiser: or better yet just expose the c++ functions in the python bindings and not have to maintain two functions for doing this :smile:

nilsnolde: right, I'd take the latter :) I'll remove it for now and save it for a future PR.

src/bindings/python/__init__.py
```diff
@@ -0,0 +1,4 @@
+try:
+    from .python_valhalla import *
+except ModuleNotFoundError:
+    from python_valhalla import *
```
kevinkreiser, line 4: is the module actually named `python_valhalla` or can a general user of this module `import valhalla` instead? i much prefer the latter. at least that is what it was called when we had the boost python bindings (iirc).

nilsnolde: It's a little convoluted tbh, but it's `import valhalla` still, since the module gets installed to `/usr/.../site-packages/valhalla`. The init.py just imports everything from the cpp lib `python_valhalla` (which resides in the same folder) and is just the target name in CMake. It was actually the same with boost.python I think.

nilsnolde: But before the `python_valhalla` cpp lib was installed directly to `site-packages` instead of `site-packages/valhalla`, so I threw in the try/except just in case.

kevinkreiser (approved):

PR #2754, opened 2020-12-30
Python bindings: build config and tiles

### LINKED ISSUE #2753: Make Python bindings more useful: build tiles & config from python API
I was about to write a blog post on how to use Valhalla with Python without going through the trouble of building it yourself, using our docker image, mapping some volumes etc. I quickly realized what a mess that'll be..

To start making the bindings more useful in a stand-alone manner (and eventually as PyPI wheels?), as a minimum, they should be able to:
- build a config file (sweet it's already a python script..)
- build tiles

Building admins and timezones would be nice too, but limited to mac/linux and a whole lot more dependencies.. Also those are almost static and could actually be downloaded by the python API if there was a central storage somewhere..

I looked into it: a few changes to python.cc and 1 more python module, that's it (hail pybind11!). I'll PR soon..

nilsnolde:
fixes #2753 

based on #2752 

Python bindings can now build a config and tiles, plus a few more helpful functions. It's not super pythonic yet, I took mostly care to not break backwards compatibility..

Typically it'd look like this:

```python
import json
from valhalla import Actor, BuildTiles, Configure, TarTiles, config

print(config.get_help()['mjolnir']['admin'])
Configure(
  './valhalla.json',
  config=config.get_default(),
  tile_dir='./valhalla_tiles',
  tile_extract='./valhalla_tiles.tar',
  verbose=False
)

BuildTiles(['./andorra-latest.osm.pbf'])
TarTiles()

actor = Actor()
route = actor.Route(json.dumps({"locations":[{"lat":42.555989,"lon":1.588898},{"lat":42.506021,"lon":1.534653}],"costing":"auto"}))
print(json.loads(route))
```


src/bindings/python/_utils.py
```diff
@@ -0,0 +1,8 @@
+import tarfile
+import os
+
+def _tar_tiles(source_dir: str, out_fp: str):
+    if not out_fp or not source_dir:
+        raise ValueError("Specify both mjolnir.tile_dir and mjolnir.tile_extract")
+    with tarfile.open(out_fp, "w") as tar:
+        tar.add(source_dir, arcname=os.path.basename(source_dir))
```
nilsnolde, line 8: will be called by c++

src/bindings/python/config.py
```diff
@@ -0,0 +1,34 @@
...
+    elif not conf:
+        raise ValueError("No local config file found, you need to specify a configuration")
+    
+    if tile_dir:
+        conf["mjolnir"]["tile_dir"] = tile_dir
+    if tile_extract:
+        conf["mjolnir"]["tile_extract"] = tile_extract
+
+    with open(path, 'w') as f:
+        json.dump(conf, f, indent=2)
```
nilsnolde, line 34: will also be called by c++

src/bindings/python/python.cc
```diff
@@ -12,34 +12,48 @@
...
+          std::string tile_dir = "",
+          std::string tile_extract = "",
+          bool verbose = true) {
   static boost::optional<boost::property_tree::ptree> pt;
-  // if we haven't already loaded one
-  if (config && !pt) {
+  if (config_path && !pt) {
+    // create the config via python
+    py::object create_config = py::module::import("valhalla.config").attr("_create_config");
+    create_config(config_path.get(), config, tile_dir, tile_extract);
```
nilsnolde, line 36: I think it's pretty cool that you can delegate stuff like that to python

src/bindings/python/python.cc
```diff
@@ -12,34 +12,48 @@
...
       pt = temp_pt;
 
       // configure logging
       boost::optional<boost::property_tree::ptree&> logging_subtree =
-          pt->get_child_optional("tyr.logging");
+          pt->get_child_optional("loki.logging");
       if (logging_subtree) {
+        // No logging if not wanted
+        if (!verbose) {
+          logging_subtree->put("type", "");
```
nilsnolde, line 49: convenience to not have to meddle with the right logging in the config

src/bindings/python/python.cc
```diff
@@ -48,12 +62,35 @@ configure(const boost::optional<std::string>& config = boost::none) {
...
+bool py_build_tiles(std::vector<std::string> input_pbfs) {
+  // make sure the service is configured
+  auto pt = configure();
+
+  pt.get_child("mjolnir").erase("tile_extract");
+  pt.get_child("mjolnir").erase("tile_url");
+
+  if (input_pbfs.empty()) {
+    throw std::invalid_argument("No PBF files specified");
+  }
```
nilsnolde, line 83: Will come out as a ValueError in Python

src/bindings/python/python.cc
```diff
@@ -93,7 +130,8 @@ struct simplified_actor_t : public valhalla::tyr::actor_t {
 };
 
 PYBIND11_MODULE(python_valhalla, m) {
-  m.def("Configure", py_configure);
+  m.def("Configure", py_configure, py::arg("config_file"), py::arg("config") = py::dict(),
+        py::arg("tile_dir") = "", py::arg("tile_extract") = "", py::arg("verbose") = true);
```
nilsnolde, line 134: what this'll do on the python side:
- if there's config_file and no config was specified nothing happens
- the config has to be a full valhalla config, `valhalla.config.get_default()` will load the config dict, and it'll overwrite an existing config file
- tile_dir and tile_extract are just convenience so you don't have to specify them manually in the config dict

test/bindings/python/utrecht-test.py
```diff
@@ -2,10 +2,42 @@
...
+valhalla.BuildTiles(['test/data/utrecht_netherlands.osm.pbf'])
+valhalla.TarTiles()
+
+assert all([os.path.isdir(p) for p in [
+    tiles_path,
+    os.path.join(tiles_path, '0'),
+    os.path.join(tiles_path, '1'),
+    os.path.join(tiles_path, '2')
+]])
+assert os.path.isfile(tar_path)
```
nilsnolde, line 39: test the whole pipeline.. not testing the exceptions yet..

nilsnolde:
it looks a bit wild maybe, c++ calling python, python calling c++.. open for discussion, but I find it nice to delegate some non-critical stuff to python which would be fairly verbose in c++.

nilsnolde:
ugh.. fedora has no mingw64 luajit package I just realized (not for the first time [apparently](https://github.com/valhalla/valhalla/pull/2619#issuecomment-700254605) :sweat_smile: ), I then don't build data_tools there and mjolnir is not available, CI fails..

fixing would be nice instead of not building python.. e.g. following the patches at https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-luajit. maybe I'll try fedora packaging in the new year;) guten rutsch allen!

src/bindings/python/CMakeLists.txt
```diff
@@ -4,7 +4,11 @@ pybind11_add_module(python_valhalla python.cc)
 target_link_libraries(python_valhalla PUBLIC valhalla)
 set_target_properties(valhalla PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
-configure_file(valhalla/__init__.py.in valhalla/__init__.py)
+# build the python lib in bin dir
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/__init__.py ${CMAKE_CURRENT_BINARY_DIR}/valhalla/__init__.py COPYONLY)
+configure_file(${VALHALLA_SOURCE_DIR}/scripts/valhalla_build_config ${CMAKE_CURRENT_BINARY_DIR}/valhalla/valhalla_build_config.py COPYONLY)
```
nilsnolde, line 9: valhalla_build_config now also acts as python module for the bindings btw..

nilsnolde:
Well, I disabled the python bindings for now in mingw CI until I (or someone who needs it for cross-compilation) creates the luajit package (opened #2758 for that).

Also, I let the user configure more than once.. actually I now doubt the usefulness of being able to configure multiple times, since the more useful things like `mjolnir.tile_extract` can't be configured more than once after an `Actor()` instance has been created. Though it would be nice to do so @kevinkreiser if possible?! What do you think?

test/CMakeLists.txt
```diff
@@ -311,14 +311,13 @@ if(ENABLE_PYTHON_BINDINGS AND ENABLE_DATA_TOOLS)
   add_custom_command(OUTPUT python_valhalla.log
     COMMAND
       LOCPATH=${VALHALLA_SOURCE_DIR}/locales
-      PYTHONPATH=src/bindings/python
+      PYTHONPATH=${CMAKE_BINARY_DIR}/src/bindings/python
       ASAN_OPTIONS=verify_asan_link_order=0:detect_leaks=0
-      /bin/bash -c "${PYTHON_EXECUTABLE} ${VALHALLA_SOURCE_DIR}/test/bindings/python/utrecht-test.py > test/python_valhalla.log"
+      /bin/bash -c "cd ${VALHALLA_SOURCE_DIR}/test/bindings/python/ && ${PYTHON_EXECUTABLE} -m unittest test_utrecht.py > ${CMAKE_BINARY_DIR}/test/python_valhalla.log"
```
nilsnolde, line 316: `unittest` can't import test_utrecht.py otherwise for some reason..

test/CMakeLists.txt
```diff
@@ -311,14 +311,13 @@ if(ENABLE_PYTHON_BINDINGS AND ENABLE_DATA_TOOLS)
   add_custom_command(OUTPUT python_valhalla.log
     COMMAND
       LOCPATH=${VALHALLA_SOURCE_DIR}/locales
-      PYTHONPATH=src/bindings/python
+      PYTHONPATH=${CMAKE_BINARY_DIR}/src/bindings/python
```
nilsnolde, line 314: this is also weird, since there's the `WORKING_DIRECTORY` directive, but wouldn't work otherwise..

nilsnolde:
More things TODO:
- transfer tarring to python completely by setting a `global` config dict when configuring
- include more utilities, e.g. decoding polyline5/6 (copy from https://github.com/gis-ops/routing-py/blob/master/routingpy/utils.py)
- BREAKING: accept/return a dict for all actions, by channeling the request/response through a python `json.dumps`/`json.loads()` interface

I'll do that once the questions above are cleared out.. not sure if you'd be fine with breaking some of the current python bindings code (e.g. the more pythonic accepting/returning dicts instead of strings)?!

nilsnolde:
Ok, another update and forget about the things above: it's perfectly possible to introduce more pythonic behavior without breaking existing clients. I overrode a few c++ things within python, so it's more flexible:

```
import json
from valhalla import Actor, Configure, BuildTiles
from valhalla.utils import decode_polyline

Configure(
  './valhalla.json',
  config=config.get_default(),
  tile_dir='./valhalla_tiles',
  tile_extract='./valhalla_tiles.tar',
  verbose=False'
)
BuildTiles(['./test/data/nyc.osm.pbf'])  # also creates a tar ball to mjolnir.tile_extract path

q = {"locations":[{"lat":52.08813,"lon":5.03231},{"lat":52.09987,"lon":5.14913}],"costing":"bicycle","directions_options":{"language":"ru-RU"}}
a = Actor()
r = a.Route(q)  # pass as dict, get a dict
coords = decode_polyline(r['trip']['legs'][0]['shape'])

r_str = a.Route(json.dumps(q))  # pass as str, get a str
assert json.dumps(json.loads(r_str)) == r
```

I think that's a good start.

If that's cool with you so far, the only discussion point would be:
Can we somehow reset the Actor interface when `Configure()` was called again (ideally only when the `tile_dir` or `tile_extract` changed), so it can reload the new tileset? I think that's the only missing piece to making the bindings maximum useful. For that I'm mostly missing what needs to be done on the c++ side..

src/bindings/python/CMakeLists.txt
```diff
@@ -4,7 +4,14 @@ pybind11_add_module(python_valhalla python.cc)
...
 
-configure_file(valhalla/__init__.py.in valhalla/__init__.py)
+# build the python lib in bin dir
+# needs to be configure_file, otherwise changes to py files are not detected
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/__init__.py ${CMAKE_CURRENT_BINARY_DIR}/valhalla/__init__.py COPYONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/utils.py ${CMAKE_CURRENT_BINARY_DIR}/valhalla/utils.py COPYONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/config.py ${CMAKE_CURRENT_BINARY_DIR}/valhalla/config.py COPYONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/_actor.py ${CMAKE_CURRENT_BINARY_DIR}/valhalla/_actor.py COPYONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/buildtiles.py ${CMAKE_CURRENT_BINARY_DIR}/valhalla/buildtiles.py COPYONLY)
+configure_file(${VALHALLA_SOURCE_DIR}/scripts/valhalla_build_config ${CMAKE_CURRENT_BINARY_DIR}/valhalla/valhalla_build_config.py COPYONLY)
```
nilsnolde, line 14: didn't find a better to do this.. could be a cmake macro I guess?

nilsnolde: just needs to be able to deploy the files again when smth changed

src/bindings/python/CMakeLists.txt
```diff
@@ -39,6 +46,10 @@ set(_packages_path "${_path}/Lib/site-packages")
 unset(_path)
 endif()
 message(STATUS "Installing python modules to ${CMAKE_INSTALL_PREFIX}/${_packages_path}")
-install(TARGETS python_valhalla
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/valhalla
```
nilsnolde, line 49: install the `python_valhalla.so` also in `<py_dir>/valhalla`. will be important later, when wheels will be built, where it'll need to pull all non-standard dynamic libs next to the valhalla.so, which then should happen in its own namespace..

src/bindings/python/__init__.py
```diff
@@ -0,0 +1,6 @@
+try:
+    from .python_valhalla import Configure
+except ModuleNotFoundError:
+    from python_valhalla import Configure
```
nilsnolde, line 4: this is only for tests. I couldn't figure out how to tell cmake to build the .so to the `./build/src/bindings/python/valhalla` folder so the tests can also the relative import `from .python_valhalla`

src/bindings/python/_actor.py
```diff
@@ -0,0 +1,60 @@
...
+        return json.loads(func(json.dumps(req)))
+    elif not isinstance(req, str):
+        raise ValueError(f"Request must be either of type str or dict")
+
+    return func(req)
+
+
+class Actor(_Actor):
+
+    def Route(self, req: Union[str, dict]) -> Union[str, dict]:
```
nilsnolde, line 22: subclass c++ `actor` so we have some more flexibility on the python side which c++ doesn't need to care about

src/bindings/python/buildtiles.py
```diff
@@ -0,0 +1,45 @@
...
+    tile_dir = Path(_global_config['mjolnir']['tile_dir'])
+    if not tile_dir.is_dir():
+        raise ValueError(f"mjolnir.tile_dir={tile_dir} is not a directory")
+    tile_extract = Path(_global_config['mjolnir']['tile_extract'])
+    if not tile_extract.parent.exists():
+        raise ValueError(f"mjolnir.tile_extract={tile_extract} is not inside an existing directory")
+
+    with tarfile.open(tile_extract, 'w') as tar:
+        tar.add(tile_dir, arcname=os.path.basename(tile_dir))
+
```
nilsnolde, line 44: I also wanted to remove the `tile_dir` files here (figured they're needlessly consuming space), but for some reason an Actor does not read from tar files but from filesystem after building tiles.. No idea why..

src/bindings/python/config.py
```diff
@@ -0,0 +1,50 @@
+import json
+import os
+
+from .valhalla_build_config import config as _config, help_text as _help_text, optional as _optional
+
+_global_config = dict()
```
nilsnolde, line 6: once configured we don't have to rely on the c++ property tree anymore for doing simple things like tarring up tiles..

src/bindings/python/utils.py
```diff
@@ -0,0 +1,61 @@
+from typing import List, Tuple
+
+
+def decode_polyline(polyline, precision=6, order='lnglat'):
```
nilsnolde, line 4: personally I find [lon, lat] nicer for coords as default

nilsnolde:
closing for now.. need to think this over again..

PR #2760, opened 2021-01-06
Reject using std::locale for tiles path

averkhaturau:
# Issue

Improve no-RTTI compatibility. In GCC std::local c-tor may throw unhandled exception if compiled with -fno-rtti

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/baldr/graphtile.cc
```diff
@@ -450,32 +428,29 @@ GraphTile::FileSuffix(const GraphId& graphid, const std::string& fname_suffix, b
...
+  auto get_num_digits = [&level]() {
+    auto max_id = level.tiles.ncolumns() * level.tiles.nrows() - 1;
+    size_t max_length = static_cast<size_t>(std::log10(std::max(1, max_id))) + 1;
+    const size_t remainder = max_length % 3;
+    if (remainder) {
+      max_length += 3 - remainder;
+    }
+    return max_length;
+  };
+  const size_t num_digits = get_num_digits();
```
kkaefer, line 440: Why is this a separate lambda rather than just having the code in-line?

src/baldr/graphtile.cc
```diff
@@ -450,32 +428,29 @@ GraphTile::FileSuffix(const GraphId& graphid, const std::string& fname_suffix, b
...
 
-  // make a locale to use as a formatter for numbers
-  std::ostringstream stream;
-  if (is_file_path) {
-    stream.imbue(dir_locale);
-  } else {
-    stream.imbue(url_locale);
+  std::string tile_id_str(num_digits + num_digits / 3, '0');
+  assert(tile_id_str.size() % 4 == 0);
+  for (int sep_index = tile_id_str.size() - 4; sep_index >= 0; sep_index -= 4)
```
kkaefer, line 445: Nit: Add braces around for loop

src/baldr/graphtile.cc
```diff
@@ -450,32 +428,29 @@ GraphTile::FileSuffix(const GraphId& graphid, const std::string& fname_suffix, b
...
+    tile_id_str[sep_index] = separator;
+  size_t ind = tile_id_str.size();
+  for (auto tile_id = graphid.tileid(); tile_id != 0; tile_id /= 1000, ind -= 4) {
+    std::string tree_digits = std::to_string(tile_id % 1000);
+    std::copy(tree_digits.begin(), tree_digits.end(), tile_id_str.begin() + ind - tree_digits.size());
```
kkaefer, line 450: Nit: This could be replaced with `snprintf` (which supports zero fill, e.g. with `"%03u"`)) and could also encompass printing the separator to the correct place to eliminate the previous for loop.

src/baldr/graphtile.cc
averkhaturau: Idea here is to limit scope of temporary variables like `remainder` and `max_id` and make sure they are used only for calculation of string length. But ok, I will remove lambda.

src/baldr/graphtile.cc
averkhaturau: I will rewrite it more optimal, without using other functions.

src/baldr/graphtile.cc
averkhaturau: Thank you.

src/baldr/graphtile.cc
kkaefer: ah; could just be `{}` scope. alternatively, you could use an immediately invoked lambda, i.e. don't give it a name and instead call it immediately.

averkhaturau: Lambda removed.

src/baldr/graphtile.cc
```diff
@@ -449,33 +427,35 @@ GraphTile::FileSuffix(const GraphId& graphid, const std::string& fname_suffix, b
...
+  for (uint32_t tile_id = graphid.tileid(); tile_id != 0; tile_id /= 10) {
+    tile_id_str[ind--] = '0' + static_cast<char>(tile_id % 10);
+    if ((tile_id_strlen - ind) % 4 == 0) {
+      tile_id_str[ind--] = separator;
+    }
+  }
+  // add missing separators
+  for (size_t sep_ind = 0; sep_ind < ind; sep_ind += 4) {
+    tile_id_str[sep_ind] = separator;
   }
```
kevinkreiser, line 456: im not sure you actually need two loops here. i think if you just change to while part of the loop to be `ind >= 0` it should still work. feel free to ignore this suggestion. otherwise the code looks good. thanks for simplifying this

kevinkreiser (approved):

src/baldr/graphtile.cc
averkhaturau: I tried to do it with one loop, but then rewrite it with two loops. And imho it looks clearer this way, putting zeros to the places when they are already seemed not good to me.

averkhaturau: It's possible to simplify this code with initializing `tile_id_str` with `"/000/000/.../000"` with maximum possible size, but I'm not sure what is maximum size here.

PR #2765, opened 2021-01-08
Fix Reach for Exclude Closures

kevinkreiser:
This pr was broken out from #2685 

In implementing that pr we found that the reach metric was incorrect when certain permutations of perameters were used. This has been illustrated by a test that is now enabled in this pr. I'll describe the changes in the code


valhalla/sif/dynamiccost.h
```diff
@@ -186,6 +197,32 @@ class DynamicCost {
...
+   *
+   * This method is to be used by loki::search and loki::reach and the base class implementation is
+   * used to do basically accessibility and adherence to the disallow mask
+   *
+   * @param edge           the edge that should or shouldnt be allowed
+   * @param tile           the tile which contains the edge (for traffic lookup)
+   * @param disallow_mask  a mask that controls additional properties that should disallow the edge
+   * @return true if the edge is allowed to be used (either as a candidate or a reach traversal)
+   */
+  inline virtual bool Allowed(const baldr::DirectedEdge* edge,
```
kevinkreiser, line 213: i renamed the `Filter` function to be `Allowed` like the other functions for nodes and pairs of edges on a path. I've also provided a default imlementation that does some basic things like checking access and doing the conservative rejection that that reach method needs. This allowed me to greatly simplify downstream costings' implementations, one day we wont even need to implement them but thats a story for another time.

test/gurka/test_search_filter.cc
```diff
@@ -584,17 +584,19 @@ TEST_P(ExcludeClosuresOnWaypoints, DISABLED_ConsecutiveClosuresWithLowReachabili
 
     // Specify minimum reachability so that nearby edges (which were previously
     // rejected due to low-reachbility), can be selected
+    // Note: we must turn off candidate ranking because of a bug in bidira* that
+    // doesnt continue the other direction of the search if one becomes exhausted
     const std::string& req_low_reachbility =
         (boost::format(
-             R"({"locations":[{"lat":%s,"lon":%s,"search_filter":{"exclude_closures":false}, "minimum_reachability": 5},{"lat":%s,"lon":%s}],"costing":"%s", "costing_options": {"%s": {"speed_types":["freeflow","constrained","predicted","current"]}}, "date_time":{"type":"%s", "value": "current"}})") %
+             R"({"locations":[{"lat":%s,"lon":%s,"search_filter":{"exclude_closures":false}, "minimum_reachability": 2, "rank_candidates":false},{"lat":%s,"lon":%s}],"costing":"%s", "costing_options": {"%s": {"speed_types":["freeflow","constrained","predicted","current"]}}, "date_time":{"type":"%s", "value": "current"}})") %
```
kevinkreiser, line 591: the problem we were encountering was that the reach was wrong when we tell the library to ignore closures for locations. in this case the reach computation would end up thinking that an edge behind a multi edge closure was reachable because it uses the, previusly named, `Filter` function, to do the quick conservative reach measurement. The problem is this conservative check was no longer conservative because we bundled the exclude closure flag into the `Filter` function. That was a mistake because it made locations that werent reachable look like they were. And would therefor fail the route instead of snapping somewhere else

src/sif/autocost.cc
```diff
@@ -275,17 +275,15 @@ class AutoCost : public DynamicCost {
...
-      // TODO - use classification/use to alter the factor
-      return 1.0f;
-    }
+  virtual bool Allowed(const baldr::DirectedEdge* edge,
+                       const graph_tile_ptr& tile,
+                       uint16_t disallow_mask = kDisallowNone) const override {
+    bool allow_closures = (!filter_closures_ && !(disallow_mask & kDisallowClosure)) ||
+                          !(flow_mask_ & kCurrentFlowMask);
+    return DynamicCost::Allowed(edge, tile, disallow_mask) && !edge->bss_connection() &&
+           (allow_closures || !tile->IsClosed(edge));
```
kevinkreiser, line 284: notice here we check if the conservative reachability algorithm asked us to always adhere to closures (we need to be conservative).

src/loki/reach.cc
```diff
@@ -53,6 +53,12 @@ directed_reach Reach::operator()(const DirectedEdge* edge,
     return reach;
   max_reach_ = max_reach;
 
+  // these are used below to get conservative estimates of forward and reverse reach
+  constexpr uint16_t forward_disallow_mask =
+      sif::kDisallowEndRestriction | sif::kDisallowSimpleRestriction | sif::kDisallowClosure;
+  constexpr uint16_t reverse_disallow_mask =
+      sif::kDisallowStartRestriction | sif::kDisallowSimpleRestriction | sif::kDisallowClosure;
```
kevinkreiser, line 60: when doing the reach check we tell the `Allowed` (filter) method what type of conservative access we want. We used to do this manually but now we let the method do it for us

src/loki/search.cc
```diff
@@ -35,15 +35,12 @@ bool is_search_filter_triggered(const DirectedEdge* edge,
...
-    return true;
-  }
-
-  return false;
+  return (road_class > min_road_class || road_class < max_road_class) ||
+         (search_filter.exclude_tunnel_ && edge->tunnel()) ||
+         (search_filter.exclude_bridge_ && edge->bridge()) ||
+         (search_filter.exclude_ramp_ && (edge->use() == Use::kRamp)) ||
+         (search_filter.exclude_closures_ && (costing.flow_mask() & kCurrentFlowMask) &&
+          tile->IsClosed(edge));
```
kevinkreiser, line 43: im not sure why there was an `if` here, its gone now. also i reverted the changes to calling isclosed directly on the costing. i didnt like that it hides state from us. instead i made sure that `exclude_closures` is always properly set at request parsing time so we can rely on it there (wther the user set it or it was overriden by ignore_closures in the costing options)

src/worker.cc
```diff
@@ -512,21 +513,25 @@ void parse_locations(const rapidjson::Document& doc,
...
-              ignore_closures ? !(*ignore_closures) : (exclude_closures ? *exclude_closures : true));
-          // set exclude_closures_disabled if any of the locations has the
-          // search_filter.exclude_closures set as false
-          if (!location->search_filter().exclude_closures()) {
-            exclude_closures_disabled = true;
-          }
+        }
+
+        // search_filter.exclude_closures must always be set because ignore_closures overrides it
+        // so if only ignore_closures is set we still need to set the search filter
```
kevinkreiser, line 519: this block needed to be moved out of the if to make sure that exclude_closures is always properly set.

src/sif/autocost.cc
```diff
@@ -716,25 +714,6 @@ class BusCost : public AutoCost {
...
                               int& restriction_idx) const override;
-
-  /**
-   * Function to be used in location searching which will
-   * exclude and allow ranking results from the search by looking at each
-   * edges attribution and suitability for use as a location by the travel
-   * mode used by the costing method. It's also used to filter
-   * edges not usable / inaccessible by bus.
-   */
-  float Filter(const baldr::DirectedEdge* edge, const graph_tile_ptr& tile) const override {
```
kevinkreiser, line 727: this and the others below were redundant, they were the same as autocost so i just let them use that one

src/thor/route_action.cc
```diff
@@ -192,7 +192,11 @@ std::string thor_worker_t::expansion(Api& request) {
   }
 
   // track the expansion
-  route(request);
+  try {
+    route(request);
+  } catch (...) {
+    // we swallow exceptions because we actually want to see what the heck the expansion did anyway
```
kevinkreiser, line 198: this was a nice addition to make the `/expansion` endpoint more useful. you can now "debug" failed routes

src/loki/reach.cc
```diff
@@ -62,7 +68,7 @@ directed_reach Reach::operator()(const DirectedEdge* edge,
   // seed the expansion with a place to start expanding from
   Clear();
   graph_tile_ptr tile, start_tile = reader.GetGraphTile(edge_id);
-  if ((tile = start_tile) && costing->Filter(edge, tile) > 0.f && !edge->restrictions())
+  if ((tile = start_tile) && costing->Allowed(edge, tile, sif::kDisallowSimpleRestriction))
```
kevinkreiser, line 71: tons of name changes like this all throughout the PR. i couldnt live with the difference in naming anymore

src/sif/nocost.cc
```diff
@@ -192,7 +192,9 @@ class NoCost : public DynamicCost {
    * mode used by the costing method. It's also used to filter
    * edges not usable / inaccessible by automobile.
    */
-  float Filter(const baldr::DirectedEdge* edge, const graph_tile_ptr&) const override {
+  bool Allowed(const baldr::DirectedEdge* edge,
+               const graph_tile_ptr&,
+               uint16_t disallow_mask = kDisallowNone) const override {
     return !(edge->is_shortcut() || edge->IsTransitLine());
```
kevinkreiser, line 198: the point of this costing is that it doesnt need to be conservative which is why its Implementation didnt change

test/gurka/gurka.cc
```diff
@@ -805,9 +805,12 @@ std::string dump_geojson_graph(const map& graph) {
...
       }
-      properties.AddMember("edge_id", std::to_string(edge_id), doc.GetAllocator());
-      properties.AddMember("opp_edge_id", std::to_string(reader.GetOpposingEdgeId(edge_id)),
-                           doc.GetAllocator());
+
+      // TODO: why on earth do we need decltype?
+      properties.AddMember(decltype(doc)::StringRefType(edge.forward() ? "edge_id" : "opp_edge_id"),
+                           std::to_string(edge_id), doc.GetAllocator());
+      properties.AddMember(decltype(doc)::StringRefType(edge.forward() ? "opp_edge_id" : "edge_id"),
+                           std::to_string(reader.GetOpposingEdgeId(edge_id)), doc.GetAllocator());
```
kevinkreiser, line 813: its more intuitive to name the edge the one in the forward direction and the opp edge the backward direction (against the grain of the shape)

test/gurka/test_search_filter.cc
kevinkreiser: also note we need to turn of `rank_candidates` here because of a bug in bidirectional a* where it runs out of stuff in one of the search branches but doesnt continue in the other even if it has stuff

PR #2769, opened 2021-01-12
Add 'use_tracks' option to avoid highway tracks

genadz:
Added new option `use_tracks` for `auto,bus,taxi,hov,truck` costings.
If it's `0` we will try to avoid tracks (add high penalty), if it's `1` we will favor tracks a little.

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/sif/autocost.cc
```diff
@@ -76,6 +77,10 @@ constexpr float kLeftSideTurnCosts[] = {kTCStraight,         kTCSlight,  kTCUnfa
 constexpr float kMinFactor = 0.1f;
 constexpr float kMaxFactor = 100000.0f;
 
+// min and max factors to apply when use tracks
+constexpr float kMinTrackFactor = 0.8f;
+constexpr float kMaxTrackFactor = 10.f;
```
genadz, line 82: I used the same factors as for alleys https://github.com/mapbox/api-valhalla/blob/589913ce8df30a6a8948df026d10ccdd3e0b80b1/src/lib/buildDirectionsDrivingRequest.js#L58 .

src/sif/autocost.cc
```diff
@@ -341,6 +348,15 @@ AutoCost::AutoCost(const CostingOptions& costing_options, uint32_t access_mask)
   toll_factor_ = use_tolls < 0.5f ? (4.0f - 8 * use_tolls) : // ranges from 4 to 0
                      (0.5f - use_tolls) * 0.03f;             // ranges from 0 to -0.15
 
+  // Preference to use tracks. Is a value from 0 to 1.
+  float use_tracks = costing_options.use_tracks();
```
genadz, line 352: I'd rather use boolean to guarantee that tracks will be excluded from the route. But in this case we will fail in cases where tracks is the only option to complete the route.

src/sif/autocost.cc
```diff
@@ -866,6 +890,13 @@ class HOVCost : public AutoCost {
     if ((edge->forwardaccess() & kHOVAccess) && !(edge->forwardaccess() & kAutoAccess)) {
       factor *= kHOVFactor;
     }
+
+    if (edge->use() == Use::kAlley) {
+      factor *= alley_factor_;
```
genadz, line 895: looks like we forgot about `alley` for trucks

src/sif/autocost.cc
```diff
@@ -1021,6 +1052,13 @@ class TaxiCost : public AutoCost {
     if ((edge->forwardaccess() & kTaxiAccess) && !(edge->forwardaccess() & kAutoAccess)) {
       factor *= kTaxiFactor;
     }
+
+    if (edge->use() == Use::kAlley) {
+      factor *= alley_factor_;
```
genadz, line 1057: the same as for `hov`. we have `alley_factor` but don't use it

src/sif/autocost.cc
```diff
@@ -45,6 +45,7 @@ constexpr float kDefaultUseFerry = 0.5f;     // Factor between 0 and 1
 constexpr float kDefaultUseRailFerry = 0.4f; // Factor between 0 and 1
 constexpr float kDefaultUseHighways = 1.0f;  // Factor between 0 and 1
 constexpr float kDefaultUseTolls = 0.5f;     // Factor between 0 and 1
+constexpr float kDefaultUseTracks = 0.f;     // Avoid tracks by default. Factor between 0 and 1
```
genadz, line 48: apply high penalty for `track`-edges by default

genadz:
I ran RAD tests and got the following results (comparing to master branch):
```
total number of routes: 13797
number of routes with different instructions: 495
number of routes with different time:         738
number of routes with different length:       504
```
Then I grabbed several top different routes and looked at them in directions-debug tool and openstreetmap. As expected, routes from `master` contains track roads, but routes from the current branch avoid them. So, at first sight everything looks ok, but for more confidence I want to look at some more routes.

dnesbitt61:
any reason not to add this to other costing? bicycle, pedestrian, motorcycle, motor_scooter all seem like they could benefit (perhaps with different default settings)

kevinkreiser:
i agree with @dnesbitt61 ive been recently taking the track of alwayys adding everything to the base class DynamicCost and then using it in all the costings. i hope that one day we can have potentially most of the code in DynamicCost and just very small changes in the individual costings.

genadz:
> any reason not to add this to other costing? bicycle, pedestrian, motorcycle, motor_scooter all seem like they could benefit (perhaps with different default settings)

imho, for all these costings it's not very important if it's a `track` road or not (because any case this road will be passable), it's more important surface factor (that is used). (moreover, for bicycle we already use `track` tag in edgecosting method https://github.com/valhalla/valhalla/blob/master/src/sif/bicyclecost.cc#L617-L618).

`auto` case (and derived) is more sensitive to the road types, it's not enough to use some implicit parameters (as surface) to detect if this road is passable or not. so, for this case we introduced `use_tracks` parameter

dnesbitt61:
Adding to bicycle costing may be less important than the other costing cases I mentioned, but I don't think it hurts to add to the base DynamicCost class and then perhaps use a different default value for bicycle. I can see possible cases where I would want to avoid tracks more or less than what the current bicycle costing does.

genadz:
> Adding to bicycle costing may be less important than the other costing cases I mentioned, but I don't think it hurts to add to the base DynamicCost class and then perhaps use a different default value for bicycle. I can see possible cases where I would want to avoid tracks more or less than what the current bicycle costing does.

@dnesbitt61 @kevinkreiser got your point. will try to move `use_tracks` option to the base class

genadz:
@kevinkreiser @dnesbitt61 updated pr to use `use_tracks` in all costings

src/sif/bicyclecost.cc
```diff
@@ -670,6 +672,13 @@ Cost BicycleCost::EdgeCost(const baldr::DirectedEdge* edge,
   // Create a final edge factor based on total stress and the weighted grade penalty for the edge.
   float factor = 1.0f + grade_penalty[edge->weighted_grade()] + total_stress + surface_factor;
 
+  // Apply factor that defines willingness to use track roads. Pay attention that `track` tag is
+  // already used when calculate `roadway_stress` parameter.
+  // TODO - work with `track` tag in one place
+  if (edge->use() == Use::kTrack) {
+    factor *= track_factor_;
```
genadz, line 679: we check for tracks in the code above https://github.com/valhalla/valhalla/blob/master/src/sif/bicyclecost.cc#L617-L618 . I didn't find the right way how to introduce `track_factor_` here and don't change current behavior, so I decided to apply `track_factor_` to the final `factor` value to be consistent with the logic in other costings

kevinkreiser: that makes me think we should just let bicycle costing alone for the moment. what this is saying is that `use_tracks` collides with `use_roads`. in my opinion `use_roads` is probably the most important bicycle factor (other than what bicycle type you pick) in the costing. @dnesbitt61 opinions?

src/sif/dynamiccost.cc
```diff
@@ -213,7 +222,16 @@ void DynamicCost::AddUserAvoidEdges(const std::vector<AvoidEdge>& avoid_edges) {
...
-};
+}
+
+void DynamicCost::set_use_tracks(float use_tracks) {
+  // Calculate factor value based on use preference. Return value
+  // in range [kMaxTrackFactor; 1], if use < 0.5; or
+  // in range [1; kMinTrackFactor], if use > 0.5
+  track_factor_ = use_tracks < 0.5f
+                      ? (kMaxTrackFactor - 2.f * use_tracks * (kMaxTrackFactor - 1.f))
+                      : (kMinTrackFactor + 2.f * (1.f - use_tracks) * (1.f - kMinTrackFactor));
```
kevinkreiser, line 233: ok cool so this works out to `.5` being a factor of `1`, which is good. then costings like auto (and probably other motor vehicles) have a default `use_tracks` that is a lower number. makes sense to me!

test/isochrone.cc
```diff
@@ -131,7 +131,7 @@ TEST(Isochrones, Basic) {
...
...2771,52.072487],[5.043259,52.075937],[5.048321,52.081429],[5.047708,52.078324],[5.051237,52.075937],[5.050321,52.072423],[5.053321,52.077735],[5.052834,52.074450],[5.055660,52.074598],[5.054992,52.077937],[5.057797,52.078461],[5.059321,52.081150],[5.060329,52.078937],[5.054919,52.070937],[5.060321,52.069593],[5.067613,52.070937],[5.066321,52.075319],[5.062321,52.076429],[5.059287,52.073903],[5.061002,52.077256],[5.069321,52.075623],[5.071556,52.076937],[5.071321,52.081586],[5.054198,52.081937],[5.050907,52.083937],[5.051766,52.085382],[5.049795,52.085463],[5.049321,52.083135],[5.048774,52.085390],[5.045321,52.085851],[5.047321,52.087985],[5.048321,52.086876],[5.054985,52.088273],[5.058693,52.092937],[5.063321,52.092392],[5.063768,52.093937],[5.062321,52.096482],[5.060270,52.094886],[5.057742,52.096937],[5.059005,52.098621],[5.057204,52.098937],[5.060839,52.098937],[5.060634,52.103937],[5.058321,52.102771],[5.053207,52.103823],[5.049845,52.106937],[5.049421,52.111037],[5.034321,52.116532],[5.029321,52.113579],[5.031766,52.118382],[5.026321,52.120692],[5.023255,52.123937],[5.033321,52.118969],[5.040321,52.125974]],"type":"LineString"},"type":"Feature"}],"type":"FeatureCollection"})";
```
genadz, line 134: testing graph contains tracks. as far as I changed default `tracks` preference for auto costing, some resulting coordinates were changed

kevinkreiser: makes sense!

src/sif/bicyclecost.cc
genadz: okay, let's remove it for now.

kevinkreiser:
other than just leaving bicycle alone for now this looks good to me!

src/sif/bicyclecost.cc
dnesbitt61: I would leave bicycle as is and set the default use_tracks factor to whatever leads to no costing change (0.5 as you have done). A user can override the current behavior to favor/avoid tracks more so than currently if they desire but unless they do so, bicycle routes will not be changed.

kevinkreiser (approved):
thanks for removing bicycle for now

PR #2771, opened 2021-01-12
Add use_distance to auto cost to allow choosing between two primary cost components, time or distance

kevinkreiser:
In our previous work we removed `auto_shorter` and replaced it with a boolean option to give back the shortest route. I propose removing that (from the api) and replacing it with `use_distance` so that the user can control how much they favor time vs distance as the main factor of their cost units. This allows them to still specify other penalties like those for ferry or highway etc.

So far I only quickly hacked it in for auto cost so i could test it but I'd rather add it to the base class and use it the same everywhere.

EDIT: I'd like to reduce the scope of this pr just to adding `use_distance` to `autocost` and letting everything else intact. to properly replace shortest (which exists on all costings) I'll need to move use_distance to the base costing class. which is what i want to do but there isnt a really clean way of doing that at the moment. i propose that first we do a refactor which moves all the `factors` into the base class and then have the children classes tell the base class what defaults it should use. Its up in the air how they should do that but @genadz had some ideas on it.


nilsnolde:
It’s been bothering me a lot too that shortest just completely disregards other factors.. 👍

kevinkreiser:
@nilsnolde any problem if i make another pr for replacing `shortest` with this in all the costings? id like to get this idea merged for autocost and then do a larger refactor, where we pull all the "factors" into DynamicCost and have the subclasses pass in their defaults for parsing. @genadz is dealing with the same kind of annoyance on #2769 and suggested a work around that i think would be best done separately.

nilsnolde:
Totally agree as well, more flexibility is always welcome:)

proto/options.proto
```diff
@@ -121,6 +121,7 @@ message CostingOptions {
   optional bool ignore_closures = 63;
   optional bool shortest = 64;
   optional float service_penalty = 65;
+  optional float use_distance = 66;
```
kevinkreiser, line 124: new costing option from 0 to 1, how much do you care about minimizing over all distance

src/sif/autocost.cc
```diff
@@ -45,6 +45,7 @@ constexpr float kDefaultUseFerry = 0.5f;     // Factor between 0 and 1
 constexpr float kDefaultUseRailFerry = 0.4f; // Factor between 0 and 1
 constexpr float kDefaultUseHighways = 1.0f;  // Factor between 0 and 1
 constexpr float kDefaultUseTolls = 0.5f;     // Factor between 0 and 1
+constexpr float kDefaultUseDistance = 0.f;   // Factor between 0 and 1
```
kevinkreiser, line 48: the default is to care only about minimizing time (as it always was) which means there are zero RAD differences

src/sif/autocost.cc
```diff
@@ -118,6 +120,15 @@ constexpr float kSurfaceFactor[] = {
...
 };
 
+// The basic costing for an edge is a trade off between time and distance. We allow the user to
+// specify which one is more important to them and then we use a linear combination to combine the two
+// into a final metric. The problem is that time in seconds and length in meters have two wildly
+// different ranges, so the linear combination always favors length vs time. What we do to combat this
+// is to change length units into time units by multiplying by the reciprocal of a constant speed.
+// This means basically changes the units of distance to be more in the same ballpark as the units of
+// time and makes the linear combination make more sense.
+constexpr float kInvMedianSpeed = 1.f / 16.f; // about 37mph
```
kevinkreiser, line 130: please read the comment above :smile:

src/sif/autocost.cc
```diff
@@ -289,11 +300,13 @@ class AutoCost : public DynamicCost {
...
-  float alley_factor_;       // Avoid alleys factor.
-  float toll_factor_;        // Factor applied when road has a toll
-  float surface_factor_;     // How much the surface factors are applied.
+  float density_factor_[16];  // Density factor
+  float highway_factor_;      // Factor applied when road is a motorway or trunk
+  float alley_factor_;        // Avoid alleys factor.
+  float toll_factor_;         // Factor applied when road has a toll
+  float surface_factor_;      // How much the surface factors are applied.
+  float distance_factor_;     // How much distance factors in overall favorability
+  float inv_distance_factor_; // How much time factors in overall favorability
```
kevinkreiser, line 309: only these two are new the rest is just whitespace

src/sif/autocost.cc
```diff
@@ -333,6 +346,10 @@ AutoCost::AutoCost(const CostingOptions& costing_options, uint32_t access_mask)
   float use_highways_ = costing_options.use_highways();
   highway_factor_ = 1.0f - use_highways_;
 
+  // Preference for distance vs time
+  distance_factor_ = costing_options.use_distance() * kInvMedianSpeed;
+  inv_distance_factor_ = 1.f - costing_options.use_distance();
```
kevinkreiser, line 351: the two coefficents of the linear combination

src/sif/autocost.cc
```diff
@@ -408,32 +425,28 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
...
-
-  if (edge->use() == Use::kAlley) {
-    factor *= alley_factor_;
-  }
-
-  return Cost(sec * factor, sec);
+  factor += (highway_factor_ * kHighwayFactor[static_cast<uint32_t>(edge->classification())] +
+             surface_factor_ * kSurfaceFactor[static_cast<uint32_t>(edge->surface())] +
+             speed_penalty + edge->toll() * toll_factor_) *
+            (edge->use() == Use::kAlley ? alley_factor_ : 1.f);
```
kevinkreiser, line 445: i removed all the branches except 1 in favor of a multiplcation trick

src/sif/autocost.cc
```diff
@@ -408,32 +425,28 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
...
+
+  // base cost before the factor is a linear combination of time vs distance, depending on which one
+  // the user thinks is more important to them
+  return Cost((sec * inv_distance_factor_ + edge->length() * distance_factor_) * factor, sec);
```
kevinkreiser, line 449: here is where we use a linear combination for the edges time vs the edges length

src/sif/autocost.cc
```diff
@@ -473,9 +486,12 @@ Cost AutoCost::TransitionCost(const baldr::DirectedEdge* edge,
     if (!edge->has_flow_speed() || flow_mask_ == 0)
       seconds *= trans_density_factor_[node->density()];
 
-    c.cost += shortest_ ? 0.f : seconds;
+    c.cost += seconds;
   }
 
+  // Account for the user preferring distance
+  c.cost *= inv_distance_factor_;
```
kevinkreiser, line 493: this is also a linear combination for the transition cost however we can simplify it because the distance of a transition is always 0 which simplifies away the second half of the formula

src/sif/pedestriancost.cc
```diff
@@ -415,98 +415,51 @@ class PedestrianCost : public DynamicCost {
...
-   * country crossing, boarding a ferry, toll booth, gates, entering destination
-   * only, alleys, and maneuver penalties. Each costing method can provide different
-   * costs for these transitions (via costing options).
-   * @param node Node at the intersection where the edge transition occurs.
-   * @param edge Directed edge entering.
-   * @param pred Predecessor edge information.
-   * @param idx  Index used for name consistency.
-   * @return Returns the transition cost (cost, elapsed time).
-   */
-  sif::Cost base_transition_cost(const baldr::NodeInfo* node,
```
kevinkreiser, line 430: this method is completely redundant with the method below so it bit the dust

src/sif/pedestriancost.cc
```diff
@@ -415,98 +415,51 @@ class PedestrianCost : public DynamicCost {
...
-      c.cost += destination_only_penalty_;
-    }
-    if (edge->use() == baldr::Use::kAlley && pred->use() != baldr::Use::kAlley) {
-      c.cost += alley_penalty_;
-    }
-    if (!edge->link() && edge->use() != Use::kEgressConnection &&
-        edge->use() != Use::kPlatformConnection && !edge->name_consistency(idx)) {
-      c.cost += maneuver_penalty_;
-    }
+    // Additional penalties without any time cost
```
nilsnolde, line 455: duplicate

nilsnolde (approved):

kevinkreiser:
perf tested this and there is no discernable difference:

master:
```
0.001742601 0.036118269 0.124286747 0.718150904 1.567052364 
```

branch:
```
0.002444983 0.036208153 0.123085022 0.718498600 1.613869429
```

src/sif/autocost.cc
```diff
@@ -408,32 +425,28 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
...
   if (shortest_) {
     return Cost(edge->length(), sec);
   }
 
-  float factor =
-      (edge->use() == Use::kFerry)
-          ? ferry_factor_
-          : (edge->use() == Use::kRailFerry) ? rail_ferry_factor_ : density_factor_[edge->density()];
+  // base factor is either ferry, rail ferry or density based
+  float factor = (edge->use() == Use::kFerry) * ferry_factor_ +
```
ktatterso, line 435: Previously the factor was exclusive to either ferry, rail-ferry, or neither.  Now it appears that the logic is written to allow for an edge to be both ferry and rail-ferry - in which case their factors are combined. 

Based on how the code is written tho, I'm guessing edge->use() can only be one value.  (e.g., there aren't any bitmasks involved).

I'm not a huge fan of using the result of the boolean condition as false=0 and true=1 in math below.  How would you feel about using a switch statement?

src/sif/autocost.cc
ktatterso: "trick" == using a boolean as a double?  :).  I wonder if that generates a compiler warning...

src/sif/autocost.cc
kevinkreiser: ah yes, i wanted to comment on this, thanks for the reminder! an edge can only have one use, so while i did use addition here only one of the terms can be non-zero, so its really like saying `ferry || rail_ferry || density`

src/sif/autocost.cc
ktatterso: If this is a commonly used convention that simplifies the logic throughout, I can learn to live with promoting bools to doubles (or ints).  Just an observation.

src/sif/autocost.cc
kevinkreiser: i wanted to experiment with what a more drastic removal of branch statements would do to this function as its called very very many times in the main loop of path finding. in my tests the performance wasnt measureably different but i did appreciate the simplification of the code so i left it.

src/sif/autocost.cc
kevinkreiser: bool gets upcasted to float yes. since float has more precision than bool there is no warning on standard warning levels though `-Wall` or `-Wpedantic` may warn about this i havent checked. we do have `-Werror` enabled for `sif` in the build so we are fine there for the moment imho

kevinkreiser: locally ive add both `-Wall` and `-Wpedantic` and indeed it doesnt complain (though pedantic did find an extra semicolon in some other code :wink:

src/sif/autocost.cc
PureW: Could the comment or variable name itself mention that it weights between minimizing time or distance?

Feel free to ignore if it doesn't make sense.

kevinkreiser: sure!

src/sif/autocost.cc
danpat: @ktatterso it *is* commonly used in this codebase, but I tend to agree with you - it makes for terser code, but I don't think that always means simpler logic or improved readability.

I would like to see a benchmark to prove out that removal of branches here is actually significant.  I know the *theory* is to reduce branching -> improved pipelining -> faster code, but I have my doubts that these kinds of branches actually end having a significant impact in many places.

src/sif/autocost.cc
ktatterso: Ah!  If performance is the goal then I highly recommend a switch statement.  A switch would work well here.  I think it would also work well in base_transition_cost().

A switch statement compiles to a jump-table, avoiding the cascading conditionals.

src/sif/autocost.cc
PureW: The compiler _should_ be able to optimize the simple branches away

kevinkreiser: yeah i recently made some performance improvements to isochrones that resulted in something like a 10% gain by removing a cascade of if elses. we got a smaller gain by reducing branching and recursion in bidirectional_astar as well. i just wanted to do an experiment here to see if we could get anything out of it.

the compiler will likely be able to predict a lot of these that occur with relatively low frequency i agree. in c++20 we can even influence that with the `unlikely` keyword

src/sif/autocost.cc
kevinkreiser: i dont think we can use it in `base_transition_cost` because value for the switch is likely too complex. we could use it here but im not sure a jump table wouldnt still stall the pipeline anyway. what is suspect is that the branch predictor will realize that most of the time the density based factor will apply and it will make no performance impact. ill give it a quick test, if there is no difference ill just use your suggestion!

src/sif/autocost.cc
ktatterso: @PureW I'm not sure the compiler can optimize away any condition unless it is programmatically impossible, e.g., "if ( false )".  Maybe you can clarify?

Compiler optimized branch "prediction" can be baked into our binaries if we can employ some sort of "profile guided optimization".  I can speak to this a bit further if there is interest.

In the end tho, the switch statement is readable (imo), doesn't rely on compiler optimization or branch prediction of any sort, and provably faster than cascading conditionals (tho as @danpat points out, not always measurably faster).

src/sif/autocost.cc
```diff
@@ -650,6 +690,8 @@ void ParseAutoCostOptions(const rapidjson::Document& doc,
     pbf_costing_options->set_use_tracks(kDefaultUseTracks);
     pbf_costing_options->set_flow_mask(kDefaultFlowMask);
     pbf_costing_options->set_top_speed(kMaxAssumedSpeed);
+    pbf_costing_options->set_use_tracks(kDefaultUseTracks);
```
kevinkreiser, line 693: @genadz we missed this in your last pr!

ktatterso (approved):

PR #2774, opened 2021-01-14
adding an action to check the changelog has been modified on PRs

kevinkreiser:
we semi-recently removed the changelog validation from our CI. this was because some changes didnt really warrent it. i want to add it back but make it optional. this way we can see that the job fails but we have to mentally evaluate whether that is ok or not, which should make us think twice about whether we actually need a changelog entry or not for a particular change

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -10,7 +10,7 @@ jobs:
       - name: Validate
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


PureW (approved):
:+1:

kevinkreiser:
i will merge this while its failing and then i will make the CI job optional for merge

PR #2780, opened 2021-01-18
Add potential tile to all GetGraphTile calls

keygang:
I change `GetGraphTile(graph_id)` to `GetGraphTile(graph_id, potential_tile)` to reduce time for searching required tile

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -168,6 +168,7 @@ void BidirectionalAStar::Init(const PointLL& origll, const PointLL& destll) {
 //
 template <const ExpansionType expansion_direction>
 inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
+                                            graph_tile_ptr& potential_tile,
                                             const sif::BDEdgeLabel& pred,
                                             const baldr::DirectedEdge* opp_pred_edge,
                                             const baldr::NodeInfo* nodeinfo,
@@ -183,9 +184,10 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
 
   graph_tile_ptr t2 = nullptr;
   baldr::GraphId opp_edge_id;
-  const auto get_opp_edge_data = [&t2, &opp_edge_id, &graphreader, &meta, &tile]() {
+  const auto get_opp_edge_data = [&t2, &opp_edge_id, &graphreader, &potential_tile, &meta, &tile]() {
     // Get end node tile, opposing edge Id, and opposing directed edge.
-    t2 = meta.edge->leaves_tile() ? graphreader.GetGraphTile(meta.edge->endnode()) : tile;
+    t2 = meta.edge->leaves_tile() ? graphreader.GetGraphTile(meta.edge->endnode(), potential_tile)
+                                  : tile;
     if (t2 == nullptr) {
       return false;
     }
@@ -370,6 +372,7 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
 
 template <const ExpansionType expansion_direction>
 bool BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
+                                graph_tile_ptr& potential_tile,
                                 const baldr::GraphId& node,
                                 sif::BDEdgeLabel& pred,
                                 const uint32_t pred_idx,
@@ -379,7 +382,7 @@ bool BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
   constexpr bool FORWARD = expansion_direction == ExpansionType::forward;
   // Get the tile and the node info. Skip if tile is null (can happen
   // with regional data sets) or if no access at the node.
-  graph_tile_ptr tile = graphreader.GetGraphTile(node);
+  graph_tile_ptr tile = graphreader.GetGraphTile(node, potential_tile);
   if (tile == nullptr) {
     return false;
   }
@@ -405,11 +408,11 @@ bool BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
     // is labelled
     pred.set_deadend(true);
     // Check if edge is null before using it (can happen with regional data sets)
-    return opp_edge &&
-           ExpandInner<expansion_direction>(graphreader, pred, opp_pred_edge, nodeinfo, pred_idx,
-                                            {opp_edge, opp_edge_id,
-                                             edgestatus.GetPtr(opp_edge_id, tile)},
-                                            shortcuts, tile, offset_time);
+    return opp_edge && ExpandInner<expansion_direction>(graphreader, potential_tile, pred,
+                                                        opp_pred_edge, nodeinfo, pred_idx,
+                                                        {opp_edge, opp_edge_id,
+                                                         edgestatus.GetPtr(opp_edge_id, tile)},
+                                                        shortcuts, tile, offset_time);
   }
 
   bool disable_uturn = false;
@@ -428,8 +431,8 @@ bool BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
     // Expand but only if this isnt the uturn, we'll try that later if nothing else works out
     disable_uturn =
         (pred.opp_local_idx() != meta.edge->localedgeidx() &&
-         ExpandInner<expansion_direction>(graphreader, pred, opp_pred_edge, nodeinfo, pred_idx, meta,
-                                          shortcuts, tile, offset_time)) ||
+         ExpandInner<expansion_direction>(graphreader, potential_tile, pred, opp_pred_edge, nodeinfo,
+                                          pred_idx, meta, shortcuts, tile, offset_time)) ||
         disable_uturn;
   }
 
@@ -443,7 +446,7 @@ bool BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
       graph_tile_ptr trans_tile = nullptr;
       if ((!trans->up() && !ignore_hierarchy_limits_ &&
            hierarchy_limits[trans->endnode().level()].StopExpanding(pred.distance())) ||
-          !(trans_tile = graphreader.GetGraphTile(trans->endnode()))) {
+          !(trans_tile = graphreader.GetGraphTile(trans->endnode(), potential_tile))) {
         continue;
       }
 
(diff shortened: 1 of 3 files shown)


kevinkreiser:
@keygang i had thought about doing this a few times but i suspected that since the queue is doing pulling off edges in a best first fashion it wouldnt matter that much (it would be jumping all over the place in terms of what tiles its looking at). i guess for shorter routes you'd only be switching between 3 levels of tiles. i wonder if we could make an array to hold the 3 tiles (1 per level) of the most recently used graphtiles, that might make this even more likely to not have to go to the cache.

also, @keygang i recently added a flat simple tilecache which avoids the indirection of the unordered_map at the cost of a small amount of ram. i dont suppose a similar optimization could be done on the synchronized LRU cache (i've not spent any time thinking about it)?

keygang:
> @keygang i had thought about doing this a few times but i suspected that since the queue is doing pulling off edges in a best first fashion it wouldnt matter that much (it would be jumping all over the place in terms of what tiles its looking at). i guess for shorter routes you'd only be switching between 3 levels of tiles. i wonder if we could make an array to hold the 3 tiles (1 per level) of the most recently used graphtiles, that might make this even more likely to not have to go to the cache.
> 
> also, @keygang i recently added a flat simple tilecache which avoids the indirection of the unordered_map at the cost of a small amount of ram. i dont suppose a similar optimization could be done on the synchronized LRU cache (i've not spent any time thinking about it)?

@kevinkreiser i checked it locally for my tests and If I'm not mistaken, on my laptop performance enhanced by 25%

kevinkreiser:
@keygang wow, ill check it myself as well!

kevinkreiser:
@keygang i checked it myself and in memory map mode i had one run that was slightly faster than master. i assume your configuration likely not memory map mode but either way this is a fine change to make. we should make it in timedep_forward as well as in timedep_reverse. would you like to add that to this pr or do it in separate pr?

keygang:
> @keygang i checked it myself and in memory map mode i had one run that was slightly faster than master. i assume your configuration likely not memory map mode but either way this is a fine change to make. we should make it in timedep_forward as well as in timedep_reverse. would you like to add that to this pr or do it in separate or?

@kevinkreiser Thanks! i will add it in this PR)

src/thor/bidirectional_astar.cc
```diff
@@ -228,6 +229,7 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
 // edge is a not-thru edge that will be pruned.
 //
 inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
+                                                   graph_tile_ptr& potential_tile,
```
kevinkreiser, line 232: this one and the ExpandReverseInner dont need to change they already send the cached tile along

src/thor/timedep_forward.cc
```diff
@@ -223,8 +226,9 @@ inline bool TimeDepForward::ExpandForwardInner(GraphReader& graphreader,
   float dist = 0.0f;
   float sortcost = newcost.cost;
   if (dest_edge == destinations_percent_along_.end()) {
-    graph_tile_ptr t2 =
-        meta.edge->leaves_tile() ? graphreader.GetGraphTile(meta.edge->endnode()) : tile;
+    graph_tile_ptr t2 = meta.edge->leaves_tile()
```
kevinkreiser, line 229: same with this inner function, the change isnt needed

src/thor/timedep_reverse.cc
```diff
@@ -169,6 +171,7 @@ bool TimeDepReverse::ExpandReverse(GraphReader& graphreader,
 //
 // Returns true if any edge _could_ have been expanded after restrictions etc.
 inline bool TimeDepReverse::ExpandReverseInner(GraphReader& graphreader,
+                                               baldr::graph_tile_ptr& potential_tile,
```
kevinkreiser, line 174: also not needed

src/thor/bidirectional_astar.cc
```diff
@@ -289,8 +291,9 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
   }
 
   // Get end node tile (skip if tile is not found) and opposing edge Id
-  graph_tile_ptr t2 =
```
kevinkreiser, line 292: here we can just make a copy of the tile that was passed in and use that as the cached tile in the GetGraphTile call. no need to add more tiles to the signature of ExpandForwardInner

kevinkreiser (dismissed):
everythign looks good except i really dont think we need to add new parameters to the `Inner` functions. We can either:

1. inside the function make a copy of the tile that was passed into it OR
2. pass the tile by value to the Inner function so that it gets its own copy which it can use in GetGraphTile

keygang:
> everythign looks good except i really dont think we need to add new parameters to the `Inner` functions. We can either:
> 
> 1. inside the function make a copy of the tile that was passed into it OR
> 2. pass the tile by value to the Inner function so that it gets its own copy which it can use in GetGraphTile

I've made option number 2

src/thor/bidirectional_astar.cc
```diff
@@ -562,6 +564,9 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
   auto forward_time_info = TimeInfo::make(origin, graphreader, &tz_cache_);
   auto reverse_time_info = TimeInfo::make(destination, graphreader, &tz_cache_);
 
+  baldr::graph_tile_ptr origin_potential_tile = nullptr;
+  baldr::graph_tile_ptr destination_potential_tile = nullptr;
```
kevinkreiser, line 568: I want to try one more thing here to squeeze every last drop of performance out of this. Here we only cache whatever tile we are using on a particular iteration, that is already very good but remember that we have multiple levels of tiles. every time the iteration jumps to a different heirarchy level we get a cache miss. we can avoid this cache miss as well. I'll push up a commit that adds this extra optimization later this afternoon!

kevinkreiser:
@keygang i tried to further optimize this but my tests showed that master was faster than this branch with the following additional changes:

```diff
 void BidirectionalAStar::SetOrigin(GraphReader& graphreader,
-                                   baldr::graph_tile_ptr& potential_tile,
+                                   std::array<baldr::graph_tile_ptr, 4>& forward_tile_cache,
                                    valhalla::Location& origin,
                                    const TimeInfo& time_info) {
   // Only skip inbound edges if we have other options
@@ -857,12 +858,13 @@ void BidirectionalAStar::SetOrigin(GraphReader& graphreader,
     }
 
     // Get the directed edge
-    graph_tile_ptr tile = graphreader.GetGraphTile(edgeid, potential_tile);
+    graph_tile_ptr tile = graphreader.GetGraphTile(edgeid, forward_tile_cache[edgeid.level()]);
     const DirectedEdge* directededge = tile->directededge(edgeid);
 
     // Get the tile at the end node. Skip if tile not found as we won't be
     // able to expand from this origin edge.
-    graph_tile_ptr endtile = graphreader.GetGraphTile(directededge->endnode(), potential_tile);
+    graph_tile_ptr endtile =
+        graphreader.GetGraphTile(directededge->endnode(), forward_tile_cache[edgeid.level()]);
     if (!endtile) {
       continue;
     }
@@ -912,7 +914,7 @@ void BidirectionalAStar::SetOrigin(GraphReader& graphreader,
 
 // Add destination edges to the reverse path adjacency list.
 void BidirectionalAStar::SetDestination(GraphReader& graphreader,
-                                        baldr::graph_tile_ptr& potential_tile,
+                                        std::array<baldr::graph_tile_ptr, 4>& reverse_tile_cache,
                                         const valhalla::Location& dest,
                                         const TimeInfo& time_info) {
   // Only skip outbound edges if we have other options
@@ -938,7 +940,7 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader,
       continue;
     }
     // Get the directed edge
-    graph_tile_ptr tile = graphreader.GetGraphTile(edgeid, potential_tile);
+    graph_tile_ptr tile = graphreader.GetGraphTile(edgeid, reverse_tile_cache[edgeid.level()]);
     const DirectedEdge* directededge = tile->directededge(edgeid);
 
     // Get the opposing directed edge, continue if we cannot get it
@@ -969,7 +971,8 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader,
     // edge (edgeid) is set.
     uint32_t idx = edgelabels_reverse_.size();
     edgestatus_reverse_.Set(opp_edge_id, EdgeSet::kTemporary, idx,
-                            graphreader.GetGraphTile(opp_edge_id, potential_tile));
+                            graphreader.GetGraphTile(opp_edge_id,
+                                                     reverse_tile_cache[opp_edge_id.level()]));
     edgelabels_reverse_.emplace_back(kInvalidLabel, opp_edge_id, edgeid, opp_dir_edge, cost, sortcost,
                                      dist, mode_, c, !opp_dir_edge->not_thru(), -1);
     adjacencylist_reverse_.add(idx);
diff --git a/valhalla/thor/bidirectional_astar.h b/valhalla/thor/bidirectional_astar.h
index ac84a2cc5..07f3132db 100644
--- a/valhalla/thor/bidirectional_astar.h
+++ b/valhalla/thor/bidirectional_astar.h
@@ -200,7 +200,7 @@ protected:
    * @param time_info    What time is it when we start the route
    */
   void SetOrigin(baldr::GraphReader& graphreader,
-                 graph_tile_ptr& potential_tile,
+                 std::array<baldr::graph_tile_ptr, 4>& potential_tile,
                  valhalla::Location& origin,
                  const baldr::TimeInfo& time_info);
 
@@ -212,7 +212,7 @@ protected:
    * @param time_info    What time is it when we end the route
    */
   void SetDestination(baldr::GraphReader& graphreader,
-                      graph_tile_ptr& potential_tile,
+                      std::array<baldr::graph_tile_ptr, 4>& potential_tile,
                       const valhalla::Location& dest,
                       const baldr::TimeInfo& time_info);
```

kevinkreiser:
and when i test this branch without my additional optimization master is still faster! im not sure what is going on here...

mskurydin:
Is there a noticeable difference between master and this change? (just curious why)

keygang:
> Is there a noticeable difference between master and this change? (just curious why)

Weird. If the tile is always the wrong one for `graph_id` I can't understand why master would be faster

```
graph_tile_ptr& GetGraphTile(const GraphId& graphid, graph_tile_ptr& tile) {
    return !tile || tile->id() != graphid.Tile_Base() ? tile = GetGraphTile(graphid) : tile;
  }
  ```

File: src/thor/bidirectional_astar.cc [hunk 6]
```diff
@@ -529,8 +533,10 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
   // PathLocation using edges.front here means we are only setting the
   // heuristics to one of them alternate paths using the other correlated
   // points to may be harder to find
-  SetOrigin(graphreader, origin, forward_time_info);
-  SetDestination(graphreader, destination, reverse_time_info);
+  baldr::graph_tile_ptr origin_potential_tile = nullptr;
```
genadz, line 536: I would rename it to `forward_tile` or `forward_tile_hint`; and the same for `destination_potential_tile`

genadz:
@keygang @kevinkreiser I tested it locally. Didn't notice any changes in performance

keygang:
also Didn't notice any performance changes locally
@kevinkreiser

kevinkreiser:
if it doesnt help then i would suggest we close this as an experiment and forget the idea. adding the complexity to the code is not worth it unless there are performance gains especially because this kind of thing has been something that is very common for people to make mistakes on (by using the wrong tile).

PR #2784, opened 2021-01-20
GraphTile::FileSuffix Should not Allow Invalid Ids

kevinkreiser:
Previously this method allowed you to use invalid tile ids so long as the the tile level was valid. Now we reject invalid tile ids as well (the rest of the math in the function relies on that so without it we can attempt to read memory that doesnt belong to us).


src/baldr/graphtile.cc
```diff
@@ -429,6 +429,10 @@ GraphTile::FileSuffix(const GraphId& graphid, const std::string& fname_suffix, b
 
   // figure out how many digits in tile-id
   const auto max_id = level.tiles.ncolumns() * level.tiles.nrows() - 1;
+  if (graphid.tileid() > max_id) {
```
kevinkreiser, line 432: if we dont do this, `ind` below can wrap around 0 and cause all kinds of problems

PureW (approved):

averkhaturau (approved):

PR #2786, opened 2021-01-20
chore: Updates libosmium

PureW:
The previous version led to compile errors when `-Wall` was propagated down to libosmium. Updating libosmium fixes this.

A more long-term fix would be to configure cmake to not propagate `-Wall` down to what should be imported as a system dependency.


mandeepsandhu:
How do we test this change to make sure the new version works as expected? Do we need to run RAD tests to make sure there are no noticeable changes?

PureW:
Looking at the [changes](https://github.com/osmcode/libosmium/compare/01bf91a4593c740352550a8811213f7141d857f1...336001aba7f29d65f24a53799895b9c7f16ee8a2) it doesn't seem like any change in functionality, mostly adding new features and fixing compiler warnings and deprecations.

gknisely (approved):

PR #2787, opened 2021-01-20
Added unit test for BuildAdminFromPBF() to test the GEOS 3.9 update.

ktatterso:
# Issue

Added unit test for BuildAdminFromPBF() to test the GEOS 3.9 update.

## Tasklist

 - [x] Add tests

## Requirements / Relations


test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,168 @@
...
+
+    const std::string ascii_map = R"(
+          A-------B-------C
+          |       |       |
+          |       |       |
+          |       |       |
+          F-------E-------D
+    )";
+
+    // To define an administrative boundary, the nodes must for a closed polygon.
```
mandeepsandhu, line 31: s/must for a closed polygon/must **form** a closed polygon ?

ktatterso: "Form".  Thanks.

kevinkreiser:
@ktatterso  you'll want to run the formatter on your code so CI is happy. you should be able to just run `scripts/format.sh`

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,128 @@
...
+    \"id_table_size\": 1000, \
+    \"tile_dir\": \"test/data/admin\", \
+    \"admin\": \"test/data/admin.sqlite\", \
+    \"timezone\": \"test/data/not_needed.sqlite\" \
+    } \
+  }";
+  file.close();
+
+  boost::property_tree::ptree pt;
+  rapidjson::read_json(config_file, pt);
```
kevinkreiser, line 89: you can create the config directly without writing a file, which i would recommend. it looks lsomething like th is:

```
boost::property_tree::ptree pt;
pt.put("mjolnir.concurrency", 1);
pt.put("mjolnir.tile_dir", "foo/bar");
...
```

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,128 @@
...
+                                        },
+                                        {{{{gurka::way_member, "ABCDEFA", "outer"}}},
+                                         {
+                                             {"type", "boundary"},
+                                             {"boundary", "administrative"},
+                                             {"admin_level", "2"},
+                                         }}};
+
+    const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize);
+    admin_map = gurka::buildtiles(layout, ways, {}, relations, "test/data/admin");
```
kevinkreiser, line 65: i hadnt thought about this before but its quite odd to have to actually build tiles just so you can get the pbf written. can we expose the methods in gurka to write just the pbf to file so you can reference that in the test below?

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,128 @@
...
+          |       |       |
+          |       |       |
+          |       |       |
+          F-------E-------D
+    )";
+
+    // To define an administrative boundary, the nodes must form a closed polygon.
+    const gurka::ways ways = {{"ABCDEFA",
+                               {
+                                   {"highway", "primary"},
```
kevinkreiser, line 32: i dont think the ways need tags at all

test/gurka/test_build_admin.cc
ktatterso: That would look cleaner, yes. I will do that. (I was sort-of modeling my code after other unit tests, we may want to apply your suggestion to those tests, too, eventually)

test/gurka/test_build_admin.cc
ktatterso: Lol, yea, I wasn't sure.  It did seem odd to define an admin boundary as a highway. :). I'll look into it.

ktatterso: I tried removing the tags and I get a seg fault.  For reasons I haven't figured out yet, despite having a debug build, CLion doesn't break on the seg fault, it just runs/reports it.  Frustrating.

I tried an empty map, an empty value in the map, and a nonsense value in the map - all crash.  But the test is happy with "highway", "primary".

Thoughts?

test/gurka/test_build_admin.cc
ktatterso: I have no idea.  Lemme see.

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,128 @@
...
+                              {"BCDEB",
+                               {
+                                   {"highway", "primary"},
+                               }}};
+
+    const gurka::relations relations = {{{{{gurka::way_member, "ABEFA", "outer"}}},
+                                         {
+                                             {"type", "boundary"},
+                                             {"boundary", "administrative"},
+                                             {"admin_level", "4"},
```
kevinkreiser, line 47: it seems to me like we should also add the name of the admin. and then build tiles with this admin sqlite. then we can look at the nodes in the graph and say hey what admin does this have it should be X and actually then you can completely avoid the whole cracking open the sql below and verify that the whole pipeline works (or you could keep the sqlite checks below as extra checks too doesnt hurt!)

test/gurka/test_build_admin.cc
mandeepsandhu: > I tried removing the tags and I get a seg fault. For reasons I haven't figured out yet, despite having a debug build, CLion doesn't break on the seg fault, it just runs/reports it. Frustrating.

@ktatterso are you able to get a backtrace (from coredump maybe)?

test/gurka/test_build_admin.cc
ktatterso: Ok, I tried to remove the logic that builds the tiles - only to find out I actually need the tile data.  GetAdminInfo() requires the caller to provide a GraphTileBuilder, whose constructor requires me to pass in a folder to tile data.

test/gurka/test_build_admin.cc
ktatterso: We discussed and realized the mock map needed one route in it, so I added a highway that spans the two states.

test/gurka/test_build_admin.cc
ktatterso: Named the admins. Added some containment tests.

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,183 @@
...
+TEST(AdminTest, TestAdminPolygonBasic) {
+  // Creates test/data/admin/map.pbf and tile data
+  const std::string workdir = "test/data/admin";
+  auto admin_map = BuildPBFandTiles(workdir);
+
+  boost::property_tree::ptree pt;
+  pt.put("mjolnir.concurrency", 1);
+  pt.put("mjolnir.id_table_size", 1000);
+  pt.put("mjolnir.tile_dir", "test/data/admin");
+  pt.put("mjolnir.admin", "test/data/admin.sqlite");
```
kevinkreiser, line 117: you should probably nest this inside the admin dir so as not to collide with other tests

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,183 @@
...
+// admin via the returned Admin object.
+//
+// As you can see in the ascii-map, there is a simple two node highway that spans
+// the two states. I thought it'd be interesting to see if I could prove that the
+// node "G" lives in "Colorado" and that "H" lives in "Utah".
+
+TEST(AdminTest, TestAdminPolygonBasic) {
+  // Creates test/data/admin/map.pbf and tile data
+  const std::string workdir = "test/data/admin";
+  auto admin_map = BuildPBFandTiles(workdir);
```
kevinkreiser, line 111: if you do this first, before you build admins then the tiles themselves will not have admin information. we need to build the admins first and then build the tiles. can you explain a little more about why we need to do this first? im pretty sure all we need is to build the pbf using the gurka::detail bits. should be approximately like this:

1. make ascii nodes ways relations
2. gurka map_to_coords
3. gurka::detail::build_pbf
4. make config file
5. build admins
6. check that the admin information in the sqlite file looks good
7. build tiles (i get that this still will recreate the pbf, we could either live with that inefficiency or we could add a parameter to specify the existing path to the pbf)
8. check that the admin names on the nodes in the graph are correct

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,183 @@
...
+      const midgard::PointLL& node_latlon = node.second;
+      uint32_t state_offset = GetContainingState(tilebuilder, polys, node_latlon);
+      ASSERT_EQ(state_offset, co_offset);
+    } else if (node_name == "H") {
+      uint32_t ut_offset = tilebuilder.AddName("Utah");
+      const midgard::PointLL& node_latlon = node.second;
+      uint32_t state_offset = GetContainingState(tilebuilder, polys, node_latlon);
+      ASSERT_EQ(state_offset, ut_offset);
+    }
+  }
```
kevinkreiser, line 182: i think we can do something like the following to test that the admins made it into the tiles where they belong with the right values:

```c++
  // get the edge in the direction of G to H and the opp edge in the direction of H to G
  const DirectedEdge* edge, *opp_edge;
  std::tie(std::ignore, edge, std::ignore, opp_edge) = gurka::findEdge(*reader, map.nodes, "GH", "H");
  // grab the end node of the edge (that should be H)
  auto tile = reader->GetGraphTile(edge->endnode());
  const auto* H = tile->node(edge->endnode());
  // get also the end node of the opp edge that should be G
  auto opp_tile = reader->GetGraphTile(opp_edge->endnode());
  const auto* G = opp_tile->node(opp_edge->endnode());
  // figure out what the admin info for H and G
  auto H_admin = tile->admininfo(H->admin_index());
  auto G_admin = tile->admininfo(H->admin_index());
  //TODO: some asserts that it matches what was in the pbf
  ```

File: test/gurka/test_build_admin.cc (hunk 6 above)
Comment (ktatterso) [reply test/gurka/test_build_admin.cc]: Ok, did it.

File: test/gurka/test_build_admin.cc (hunk 7 above)
Comment (ktatterso) [reply test/gurka/test_build_admin.cc]: Yep, we discussed.  Took your suggestions.

File: test/gurka/test_build_admin.cc (hunk 8 above)
Comment (ktatterso) [reply test/gurka/test_build_admin.cc]: Thanks, suggestions taken.

File: test/gurka/test_build_admin.cc [hunk 9]
```diff
@@ -0,0 +1,245 @@
...
+  std::unordered_map<uint32_t, bool> drive_on_right;
+  std::unordered_map<uint32_t, bool> allow_intersection_names;
+  AABB2<PointLL> world_box(-180.0f, -90.0f, 180.f, 90.f);
+  polys = GetAdminInfo(db_handle, drive_on_right, allow_intersection_names, world_box, tilebuilder);
+  sqlite3_close(db_handle);
+
+  // For two states part of one country, there should be three polys, etc.
+  EXPECT_EQ(polys.size(), 3);
+  EXPECT_EQ(drive_on_right.size(), 3);
+  EXPECT_EQ(allow_intersection_names.size(), 3);
```
kevinkreiser, line 210: im not sure this is needed. below where you test the contents of the tile you can assert that everything you're testing here also happened correctly. we have other tests that exorcise tilebuilder, i would suggest adding a few bits below to cover what is covered here. 

1. you can check if its drive on right or left down below and check the actual value rather than just the existance of the property. have a look at nodeinfo::drive_on_right
2. you can check the number of polygons above where you have your sets of names above
3. the allow intersections names boolean is a bit trickier. first you have to modify your json configuration to tell which countries should have named intersections, maybe you need to add another country into the test to really see both allowed and disallowed. then you'll simply give the nodes names in your osm tags above. finally you can then check nodeinfo::named_intersection and if you really want you can even go fetch the name to make sure it matches your tag value

let me know if any of this is unclear!

test/gurka/gurka.cc
```diff
@@ -507,7 +507,7 @@ findEdge(valhalla::baldr::GraphReader& reader,
             auto forward_edge_id = tile_id;
             forward_edge_id.set_id(i);
             auto reverse_edge_id = tile->GetOpposingEdgeId(forward_directed_edge);
-            auto* reverse_directed_edge = tile->directededge(i);
+            auto* reverse_directed_edge = tile->directededge(reverse_edge_id.id());
```
kevinkreiser, line 510: yep you found a bug here, i have no clue how the hell this is possible to have been here so long but anyway...

i would suggest we change this whole bit to be more like:

```c++
const DirectedEdge* reverse_directed_edge = nullptr;
graph_tile_ptr reverse_tile = tile;
auto reverse_edge_id = GetOpposingEdgeId(forward_edge_id, reverse_directed_edge, reverse_tile);
```

Edges are stored in teh tiles where they originate, that is where their begin node is. For example: imagine we have 2 tiles that are adjacent call them A and B. If forward edge here originates in A and ends in B then its opposing edge will originate in B and end in A. That means that forward edge is sorted in A and reverse edge is stored in B. One cannot ask tile A for reverse edge and get the right answer. Now the fix you did would actually fix this crazy bug in the case that forward edge and reverse edge were in the same tile, so its definitely better, but its stil not right. The above snippet I posted should do the trick and im guessing other unit test fall out will ensue...

kevinkreiser: it looks like it only affected that time tracking test regardless of the fix so yeah i think you can just make the change i suggest here and we should be good to go on that front!

ktatterso: Thanks for the explanation, I'm on it.

test/gurka/test_build_admin.cc
ktatterso: Thanks for our chat today.  I think I've got everything wrapped.  Let me know.

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,281 @@
...
+                             {
+                                 {"highway", "primary"},
+                             }},
+                            {"XZ",
+                             {
+                                 {"highway", "primary"},
+                             }}};
+
+  // X lives Japan which allows named intersections - and is named.
+  const gurka::nodes nodes = {{"X", {{"junction", "named"}}}};
```
kevinkreiser, line 61: i think it might be `junction=yes` and `name=Shibuya Crossing` something like that. im surprised that it sets the boolean even though it doesnt have a name!? that is blowing my mind. cc @gknisely

kevinkreiser: https://wiki.openstreetmap.org/wiki/Tag:junction%3Dyes#Usage_examples

ktatterso: Lol I dunno, I just reverse engineered this code:  https://github.com/valhalla/valhalla/blob/52885542d9bd5718d54f9ae8d01de072859cbab8/src/mjolnir/pbfgraphparser.cc#L872

ktatterso: Sorry I don't know how to grab three lines in a link yet... its 872-874.

kevinkreiser: no worries at all but i think you found another mild bug! notice the boolean `has_junction_name` regardless of whether or not it actually has a name? that doesnt seem right! and your test proves that the flag makes it all the way through to the nodeinfo after the tiles are created. i think we dont want to do that if there isnt actually a name. and oh man looking at that method it really needs the refactor stick. i'm going to push up a bit of a change to that part of the code, bear with me!

src/mjolnir/adminbuilder.cc
```diff
@@ -0,0 +1,651 @@
...
+  }
+
+  // loading SpatiaLite as an extension
+  if (!valhalla::mjolnir::load_spatialite(db_handle)) {
+    sqlite3_close(db_handle);
+    return;
+  }
+
+  /* creating an admin POLYGON table */
+  sql = "SELECT InitSpatialMetaData(1); CREATE TABLE admins (";
```
kevinkreiser, line 283: adding the argument 1 here makes this like 100x faster

src/mjolnir/pbfgraphparser.cc
```diff
@@ -865,13 +865,14 @@ struct graph_callback : public OSMPBF::Callback {
       results = results ? results : empty_node_results_;
     }
 
-    const auto& highway_junction = results->find("highway");
```
kevinkreiser, line 868: this looks like a lot of diffs but its really the same code just simplified. there were:

1. tons of copy paste of checking the length of the value, i made a single varaible for it and did it once at the top of the loop
2. an insane amount of ternaries around things there were already true/false values (wtf?)
3. lots of nested `ifs` where were very simply combined into a single `if` with some `&&`s
4. iterators declared as const reference (wtf?)
5. main change was that we store a bool for when a junction could be named and then at the end, only once outside of the loop, if the junction actually had a name we set the flag that eventually gets written into nodeinfo (removing false positives as a possibility)

test/gurka/test_build_admin.cc
```diff
@@ -0,0 +1,286 @@
...
+                                 {"highway", "primary"},
+                             }}};
+
+  // X lives Japan which allows named intersections - and is named.
+  // gurka automatically names the nodes by their name in the ascii map
+  // if you want to make sure there is no name you need to send empty string
+  const gurka::nodes nodes = {
+      {"X", {{"junction", "named"}, {"name", "namae wa desu ka"}}},
+      {"Y", {{"junction", "yes"}, {"name", ""}}},
+  };
```
kevinkreiser, line 66: @ktatterso i made this explicit. gurka does stuff for us automatically but that ends up masking the behavior we wanted to test a little bit. i found setting it directly here is easier to reason about (beacuse you dont have to know that the gurka method is actually always adding a name to every node)

test/gurka/test_time_tracking.cc
```diff
@@ -218,7 +218,8 @@ TEST(TimeTracking, routes) {
       }
     }
   }
-  std::vector<double> expected = {0, 17.1429, 34.2857, 176.789, 228.2844, 252.2844, 276.2844};
+  std::vector<double> expected = {0,        17.1429,  34.2857,  51.4286,
```
kevinkreiser, line 221: these were the ones that changed as a result of fixing the bug in the gurka findEdge function

kevinkreiser (dismissed):

kevinkreiser (approved):

test/gurka/test_time_tracking.cc
ktatterso: correct.

PR #2788, opened 2021-01-20
Implement living street avoidance

SvetlanaBayarovich:
# Issue

Introduces `use_living_streets` parameter which allows to tune living streets preferences during routing with different costings.
The impact on route selection consists of 2 parts:

- penalize entering living street areas, if parameter value is in range [0, 0.5]
- increase or decrease costs for living street edges according to parameter value.  

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/sif/autocost.cc
```diff
@@ -435,6 +444,8 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
     factor *= alley_factor_;
   } else if (edge->use() == Use::kTrack) {
     factor *= track_factor_;
+  } else if (edge->use() == Use::kLivingStreet) {
```
SvetlanaBayarovich, line 447: (self) Add tests

test/astar_bss.cc
```diff
@@ -323,11 +323,8 @@ TEST(AstarBss, test_Auto) {
       R"({"locations":[{"lat":48.859895,"lon":2.3610976338},{"lat":48.86271911,"lon":2.367111146}],"costing":"auto"})";
   std::vector<valhalla::DirectionsLeg_TravelMode> expected_travel_modes{
       valhalla::DirectionsLeg_TravelMode::DirectionsLeg_TravelMode_kDrive};
-  std::vector<std::string> expected_route{"Rue de la Perle",   "Rue Vieille du Temple",
-                                          "Rue Barbette",      "Rue Elzévir",
-                                          "Place de Thorigny", "Rue de Thorigny",
-                                          "Rue Debelleyme",    "Rue de Turenne",
-                                          "Rue Commines",      "Rue Amelot"};
+  std::vector<std::string> expected_route{"Rue de la Perle", "Rue des Archives", "Rue Pastourelle",
```
SvetlanaBayarovich, line 326: `Rue de Thorigny` is a living street, so avoid it now

src/sif/autocost.cc
```diff
@@ -291,11 +296,12 @@ class AutoCost : public DynamicCost {
...
-  float highway_factor_;     // Factor applied when road is a motorway or trunk
-  float alley_factor_;       // Avoid alleys factor.
-  float toll_factor_;        // Factor applied when road has a toll
-  float surface_factor_;     // How much the surface factors are applied.
+  float density_factor_[16];   // Density factor
+  float highway_factor_;       // Factor applied when road is a motorway or trunk
+  float alley_factor_;         // Avoid alleys factor.
+  float toll_factor_;          // Factor applied when road has a toll
+  float surface_factor_;       // How much the surface factors are applied.
+  float living_street_factor_; // Avoid living streets factor.
```
SvetlanaBayarovich, line 304: Added as an autocost-only member for now. Maybe should be moved to DynamicCost. Not sure which profiles should also be affected.

src/sif/autocost.cc
```diff
@@ -66,6 +66,9 @@ constexpr float kTaxiFactor = 0.85f;
 // Do not avoid alleys by default
 constexpr float kDefaultAlleyFactor = 1.0f;
 
+// Avoid living street by default
+constexpr float kDefaultLivingStreetFactor = 3.0f;
```
SvetlanaBayarovich, line 70: Need to understand, what should be used as a default value. 3.0 solves Barcelona cases from the issue

SvetlanaBayarovich:
Some test cases from initial issue:

|prod|result|data|
|---|---|---|
|image|image|image|
|image|image|image|

src/sif/dynamiccost.cc
```diff
@@ -233,6 +237,17 @@ void DynamicCost::set_use_tracks(float use_tracks) {
                       : (kMinTrackFactor + 2.f * (1.f - use_tracks) * (1.f - kMinTrackFactor));
 }
 
+void DynamicCost::set_use_living_streets(float use_living_streets) {
+  // Calculate factor value based on use preference. Return value
+  // in range [kMaxLivingStreetFactor; 1], if use < 0.5; or
+  // in range [1; kMinLivingStreetFactor], if use > 0.5
```
PureW, line 243: Maybe mention that `living_street_factor` is inversely proportional to `use_living_streets`?

test/gurka/test_use_living_streets.cc
```diff
@@ -0,0 +1,94 @@
...
+  for (const auto& c : costing)
+    if (c == "bicycle" || c == "pedestrian")
+      // living_street tag shouldn't affect these costings too much
+      validate_path(gurka::route(use_living_streets_map, "1", "2", c), {"AB", "BE", "EF"});
+    else
+      // avoid living_streets by default; use living_streets only if the route starts or ends at
+      // 'living_street' edge
+      validate_path(gurka::route(use_living_streets_map, "1", "2", c),
+                    {"AB", "BC", "CD", "DE", "EF"});
+}
```
PureW, line 65: This test is great, easy to read and follow.

PureW (dismissed):
Nice, LGTM

test/gurka/test_use_living_streets.cc
SvetlanaBayarovich: All credit goes to @genadz and his tests on track avoidance :)

src/sif/dynamiccost.cc
SvetlanaBayarovich: done

test/astar_bss.cc
kevinkreiser: but this is a bike share route which means its just pedestirans and bike riders, we shouldnt be avoiding living streets for those modes of travel should we?

src/sif/pedestriancost.cc
```diff
@@ -59,7 +59,8 @@ constexpr float kDefaultSideWalkFactor = 1.0f;     // Neutral value for sidewalk
 constexpr float kDefaultAlleyFactor = 2.0f;        // Avoid alleys
 constexpr float kDefaultDrivewayFactor = 5.0f;     // Avoid driveways
 constexpr float kDefaultUseFerry = 1.0f;
-constexpr float kDefaultUseTracks = 0.5f; // Factor between 0 and 1
+constexpr float kDefaultUseTracks = 0.5f;        // Factor between 0 and 1
+constexpr float kDefaultUseLivingStreets = 0.6f; // Factor between 0 and 1
```
kevinkreiser, line 63: should this be 1.f? isnt the point of most of these living streets that pedestrians get right of way at or above other modes of travel?

test/astar_bss.cc
SvetlanaBayarovich: Yeah, but this is the test case for auto costing, as the line 270 says, so I thought this was expected

src/sif/pedestriancost.cc
SvetlanaBayarovich: I agree with your idea, but in my opinion pedestrians are more sensitive to route distance/duration increase which can happen if we favor living streets more. That's why I didn't want to mess with this costing a lot

PureW: Interesting, so a customer who would like more scenic and clean routes may want to crank this up at the potential expense of time and distance. Makes sense.

SvetlanaBayarovich:
According to the testing, the amount of routes that changed significantly is relatively small. There are only 62 routes with relative difference in distance greater than 1%. As for the weight, the routes with high difference are all of a certain type: short (from 300-400 meters to couple of kilometers) and with mandatory living street edges on them, so that the route itself does not change even change wrt master, only the cost increases. Thus when living streets are not mandatory, we manage to avoid them without significant weight increase. 

The most interesting cases I came across are


1)Well, we do avoid this living street. The result just does not look nice.
|master|living street avoidance|data|
|---|---|---|
|image|image|image|

2)More relevant for the large areas of living streets. We add sth like 400 meters to the total distance to save like ~50 meters of riding living streets (because 'living street' edges are really costly)
|master|living street avoidance|data|
|---|---|---|
|image|image|image|

gknisely (approved):

SvetlanaBayarovich:
Testing another approach with high penalty for transition into living street and not such a big factor (previously I was betting everything on large factor) in order to avoid situation 2) from my comment above. The idea is to minimize the part of the route that goes through living streets, but not to add to much to the route length if we need to use it as an origin or destination

gknisely (approved):

PR #2789, opened 2021-01-21
Add boolean parameter to clear memory for edge labels from thor

keygang:
Add boolean parameter to clear memory for edge labels from thor path algorithms.

It will be good for mobile platforms to free extra memory because building a route is a rare task. It frees up about 50mb.


scripts/valhalla_build_config
```diff
@@ -92,7 +92,8 @@ config = {
     'service': {
       'proxy': 'ipc:///tmp/thor'
     },
-    'max_reserved_labels_count': 1000000
+    'max_reserved_labels_count': 1000000,
+    'clear_reserved_memory': True
```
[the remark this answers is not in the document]
keygang: > what is the difference between setting max_reserved_labels_count=0 and clear_reserved_memory=True ?

@genadz it will be no reallocations after push backing in the vector

src/thor/astar_bss.cc
```diff
@@ -29,10 +29,11 @@ constexpr uint32_t kInitialEdgeLabelCount = 500000;
 constexpr uint32_t kMaxIterationsWithoutConvergence = 200000;
 
 // Default constructor
-AStarBSSAlgorithm::AStarBSSAlgorithm(uint32_t max_reserved_labels_count)
+AStarBSSAlgorithm::AStarBSSAlgorithm(uint32_t max_reserved_labels_count, bool clear_reserved_memory)
```
kevinkreiser, line 32: i was just talking to @genadz about this the other day. we should just pass the config ptree here and then pull the values out that we need otherwise the constructors arguments will get really long. what does everyone else think about that? the default can be an empty ptree and then we can have defaults for these values in the initializer list someting like:

```c++
AStarBSSAlgorithm::AStarBSSAlgorithm(const boost::property_tree::ptree& config = {}): 
      PathAlgorithm(), max_label_count_(std::numeric_limits<uint32_t>::max()),
      mode_(TravelMode::kDrive), travel_type_(0),
      max_reserved_labels_count_(config.get<uint32_t>("max_reserved_labels_count", kMaxReservedLabelsCountDefault),
      clear_reserved_memory_(config.get<bool>("clear_reserved_memory", false)) {
...
```

scripts/valhalla_build_config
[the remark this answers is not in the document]
kevinkreiser: `false` should be the default

src/thor/astar_bss.cc
keygang: > what does everyone else think about that?

@kevinkreiser yes, i like it too. Make it in this PR?

keygang: > Make it in this PR?

done)

src/thor/astar_bss.cc
```diff
@@ -29,10 +29,12 @@ constexpr uint32_t kInitialEdgeLabelCount = 500000;
...
 
 // Default constructor
-AStarBSSAlgorithm::AStarBSSAlgorithm(uint32_t max_reserved_labels_count)
+AStarBSSAlgorithm::AStarBSSAlgorithm(const boost::property_tree::ptree& config)
     : PathAlgorithm(), max_label_count_(std::numeric_limits<uint32_t>::max()),
       mode_(TravelMode::kDrive), travel_type_(0),
-      max_reserved_labels_count_(max_reserved_labels_count) {
+      max_reserved_labels_count_(
+          config.get<uint32_t>("max_reserved_labels_count", kInitialEdgeLabelCount)),
+      clear_reserved_memory_(config.get<bool>("clear_reserved_memory", false)) {
```
kevinkreiser, line 37: looks perfect, thank you!

src/thor/astar_bss.cc
```diff
@@ -42,8 +44,13 @@ AStarBSSAlgorithm::~AStarBSSAlgorithm() {
...
   // Reduce edge labels capacity if it's more than limit
-  if (edgelabels_.size() > max_reserved_labels_count_) {
-    edgelabels_.resize(max_reserved_labels_count_);
+  if (!clear_reserved_memory_) {
+    if (edgelabels_.size() > max_reserved_labels_count_) {
+      edgelabels_.resize(max_reserved_labels_count_);
+      edgelabels_.shrink_to_fit();
+    }
+  } else {
+    edgelabels_.clear();
```
kevinkreiser, line 53: I think all of these can be simplified to this:

```suggestion
  edgelabels_.resize(clear_reserved_memory ? 0 : max_reserved_labels_count_);
  edgelabels_.shrink_to_fit();
```

see that below we are already clearing the labels no matter what. so the question is only really how much reservation do we want to have in the end? either we want 0 reservation when that boolean is set OR we want the max_reserved.

src/thor/astar_bss.cc
```diff
@@ -42,8 +44,13 @@ AStarBSSAlgorithm::~AStarBSSAlgorithm() {
...
     edgelabels_.shrink_to_fit();
   }
```
kevinkreiser, line 55: I think all of these can be simplified to this:

```suggestion
  edgelabels_.resize(clear_reserved_memory ? 0 : max_reserved_labels_count_);
  edgelabels_.shrink_to_fit();
```

see that below we are already clearing the labels no matter what. so the question is only really how much reservation do we want to have in the end? either we want 0 reservation when that boolean is set OR we want the max_reserved.

kevinkreiser:
looks like the tests are running into what i expected: https://github.com/valhalla/valhalla/issues/2808

I would suggest we reserve max_reach in the reach constructor by passing it to the dijkstra constructor.

kevinkreiser:
@keygang i cleaned up that issue with the resevering too much by default in the reach check on this branch. with any luck the performacne will be unimpacted and we can merge after checking that

kevinkreiser:
bad news! performance is absolutely aweful compared to master... im not quite sure where the hell its going though... @genadz any thoughts?

src/thor/astar_bss.cc
```diff
@@ -42,10 +44,9 @@ AStarBSSAlgorithm::~AStarBSSAlgorithm() {
 // Clear the temporary information generated during path construction.
 void AStarBSSAlgorithm::Clear() {
   // Reduce edge labels capacity if it's more than limit
-  if (edgelabels_.size() > max_reserved_labels_count_) {
-    edgelabels_.resize(max_reserved_labels_count_);
-    edgelabels_.shrink_to_fit();
-  }
+  edgelabels_.resize(clear_reserved_memory_ ? 0 : max_reserved_labels_count_);
+  edgelabels_.shrink_to_fit();
```
kevinkreiser, line 48: my suggestion was wrong here. when you dont have clear_reserved_memory_ set to true and the edges labels container is smaller than max_reserved_labels_count_ then you dont actually have to resize and shrink_to_fit.\



```c++
auto reservation = clear_reserved_memory_ ? 0 : max_reserved_labels_count_; // do this in the constructor and save it as a member variable?
if (edgelabels_forward_.size() > reservation) {
    edgelabels_forward_.resize(reservation);
    edgelabels_forward_.shrink_to_fit();
}

```

kevinkreiser: i tried fixing this in bidirectional a* and dijkstras but its still slow... where else are we running into this, it clearly has to be this bit of the code that is causign the performane bottleneck. i'll  look at it with fresh eyes in a day or so

keygang: @kevinkreiser 
you haven't watched this PR again?

genadz:
> @keygang id love to merge this but there is something wrong with the performance. its incredibly slow. we need to concentrate on what these changes are doing to make the code slower. maybe start with master and add one change at a time to see which one is doing it. i had some th oughts about where it could be comign from but wasnt able to figure out why. maybe @genadz could also take a quick look since hes also been in there recently?

@keygang @kevinkreiser I found a reason of bad performance for default case (`clear_reserved_memory=false`). Function `Algorithm::Clear()` is called on each request. In that function we initialize (`vector::resize`) labels vector with `max_reserved_labels_count` elements and then immediately clear it. So, we shouldn't call `resize` if number of elements less than allowed capacity.

keygang:
> > @keygang id love to merge this but there is something wrong with the performance. its incredibly slow. we need to concentrate on what these changes are doing to make the code slower. maybe start with master and add one change at a time to see which one is doing it. i had some th oughts about where it could be comign from but wasnt able to figure out why. maybe @genadz could also take a quick look since hes also been in there recently?
> 
> @keygang @kevinkreiser I found a reason of bad performance for default case (`clear_reserved_memory=false`). Function `Algorithm::Clear()` is called on each request. In that function we initialize (`vector::resize`) labels vector with `max_reserved_labels_count` elements and then immediately clear it. So, we shouldn't call `resize` if number of elements less than allowed capacity.

https://github.com/valhalla/valhalla/pull/2789#discussion_r567346651

genadz:
> yes, but @kevinkreiser checked without resize when size is small the max_reserved_labels

> i tried fixing this in bidirectional a* and dijkstras but its still slow... where else are we running into this, it clearly has to be this bit of the code that is causign the performane bottleneck. i'll look at it with fresh eyes in a day or so

you should fix this in **all** algorithms, because you call `Clear()` here https://github.com/valhalla/valhalla/blob/master/src/thor/worker.cc#L343-L356 per each request

kevinkreiser:
@genadz yeah i saw that too and made the change locally but still saw poor performance. I'm not sure what the heck i was doing wrong (probably just too many things at once). If someone has the time that this no longer shows the performance regression i think we can get it reviewed and merged

CHANGELOG.md
```diff
@@ -496,6 +497,7 @@
    * CHANGED: Optimize double bucket queue to reduce memory reallocations. [#2719](https://github.com/valhalla/valhalla/pull/2719)
    * CHANGED: Collapse merge maneuvers [#2773](https://github.com/valhalla/valhalla/pull/2773)
    * CHANGED: Add shortcuts to the tiles' bins so we can find them when doing spatial lookups. [#2744](https://github.com/valhalla/valhalla/pull/2744)
+   * CHANGED: Add shortcuts to the tiles' bins so we can find them when doing spatial lookups. [#2744](https://github.com/valhalla/valhalla/pull/2344)
```
genadz, line 500: @keygang looks like you've added it by accident

keygang: ops, yes

src/loki/reach.cc
```diff
@@ -67,7 +67,9 @@ directed_reach Reach::operator()(const DirectedEdge* edge,
   // and we can expand from the start of a complex restriction because only the end would mark a
   // potential stopping point (maybe a path followed the restriction)
 
-  // seed the expansion with a place to start expanding from
+  // seed the expansion with a place to start expanding from, set the number of labels we want to use
+  // we're finding nodes here so we'll double it assuming we queue less edges than nodes we see
+  max_reserved_labels_count_ = max_reach * 2;
```
genadz, line 72: did you find it experimentally ?

keygang: To be honest, I don't quite understand why this is. Kevin seemed to be doing it.
This is related to this issue https://github.com/valhalla/valhalla/issues/2808

valhalla/thor/bidirectional_astar.h
```diff
@@ -43,7 +43,7 @@ class BidirectionalAStar : public PathAlgorithm {
 public:
   /**
    * Constructor.
-   * @param config A config object of key, value pairs
+   * @param config   Config with std::string for key and data
```
genadz, line 46: was the previous description confusing ?

keygang: No, This is rebase's problem. Bring it back?

src/loki/reach.cc
genadz: ok. I think rough theoretical result can be obtained from here https://en.wikipedia.org/wiki/Planar_graph

kevinkreiser: it was just an attempt to right size the labels allocation for the reach computation. as the comment says, since the reach is based on nodes reached, i couldnt use max_reach directly to guess at how many edges i would ahve to see to hit the node reach threshold so i guessed that i'd see about 2 edges for every new node that i found. we could arrive at this empirically if we wanted to

valhalla/thor/bidirectional_astar.h
genadz: I don't see any reason to change this line. so, I'd propose change it back

valhalla/thor/bidirectional_astar.h
```diff
@@ -112,6 +112,9 @@ class BidirectionalAStar : public PathAlgorithm {
   std::vector<sif::BDEdgeLabel> edgelabels_reverse_;
   uint32_t max_reserved_labels_count_;
 
+  // if `true` clean reserved memory for edge labels
+  bool clear_reserved_memory_;
```
kevinkreiser, line 116: all of these algorithms derive from pathalgorithm. rather than duplicating this maybe we should just move it up there and have less code in total? although i guess thats the same for all of the already duplicated stuff wrt edge labels handling.

genadz: if it's possible we should move `max_reserved_labels_count_` also

genadz:
@keygang @kevinkreiser I tested this branch (with `clear_reserved_memory=false`) locally against master and found out that this branch is **slower by ~2%**. interesting why

kevinkreiser:
yeah i would expect no change for defaults. maybe its the loki thing

genadz:
> yeah i would expect no change for defaults. maybe its the loki thing

Checking...

genadz:
> > yeah i would expect no change for defaults. maybe its the loki thing
> 
> Checking...

@keygang @kevinkreiser sorry, I checked timings again and didn't notice any difference. so, probably some other local stuff hit timings when I measured them first time.

dnesbitt61:
I have found that performance drops if EdgeLabels needs to be resized during route computation. One can check the reserved size against the # of labels at the end of computation (I always look at the log statement in FormPath).

keygang:
> I have found that performance drops if EdgeLabels needs to be resized during route computation. One can check the reserved size against the # of labels at the end of computation (I always look at the log statement in FormPath).

Yes, when the flag is `true`, we need to resize labels during route computation. But we need to clean up labels in order not to take up extra memory on mobile platforms which we don't use, because these labels take up a lot of space

valhalla/thor/pathalgorithm.h
```diff
@@ -32,8 +32,10 @@ class PathAlgorithm {
   /**
    * Constructor
    */
-  PathAlgorithm()
-      : interrupt(nullptr), has_ferry_(false), not_thru_pruning_(true), expansion_callback_() {
+  PathAlgorithm(uint32_t max_reserved_labels_count, bool clear_reserved_memory)
+      : interrupt(nullptr), has_ferry_(false), not_thru_pruning_(true), expansion_callback_(),
+        max_reserved_labels_count_(std::move(max_reserved_labels_count)),
```
genadz, line 37: `std::move` - what for ? integer types don't have move-constructors, it will be always copying

valhalla/thor/pathalgorithm.h
```diff
@@ -32,8 +32,10 @@ class PathAlgorithm {
   /**
    * Constructor
    */
-  PathAlgorithm()
-      : interrupt(nullptr), has_ferry_(false), not_thru_pruning_(true), expansion_callback_() {
+  PathAlgorithm(uint32_t max_reserved_labels_count, bool clear_reserved_memory)
```
genadz, line 35: imho, it's better to pass property tree into the constructor. in that case we would have common parameters parsing
```c++
config.get<uint32_t>("max_reserved_labels_count", kInitialEdgeLabelCountBD),
config.get<bool>("clear_reserved_memory", false)
```
only once here

keygang: yeah good, remark, but how set `kInitialEdgeLabelCountBD` beautifully in `PathAlgorithm`? In different algos we set different `kInitialEdgeLabelCountBD`, for example for [Astar we set 1000000](https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L18) but for [multimodal we set 200000](https://github.com/valhalla/valhalla/blob/master/src/thor/multimodal.cc#L41)
@genadz

keygang: we can update config by default value and forward it to `PathAlgoritm` as a copy, is it good?

genadz: hm, I don't see a good way how to set different `max_reserved_labels_count` for different algorithms and keep parsing only in `PathAlgorithm`. so, probably for now it's okay to have constructor as you proposed
```c++
PathAlgorithm(uint32_t max_reserved_labels_count, bool clear_reserved_memory)
```
@kevinkreiser WDYT ?

kevinkreiser: @genadz @keygang i agree, its completley fine to keep it that way until we think of something cleaner

genadz (approved):

scripts/valhalla_build_config
```diff
@@ -363,7 +364,8 @@ help_text = {
     'service': {
       'proxy': 'IPC linux domain socket file location'
     },
-    'max_reserved_labels_count': 'Maximum capacity for edge labels reserved in path algorithm',
+    'max_reserved_labels_count': 'Maximum capacity that allowed to keep reserved in path algorithm.',
+    'clear_reserved_memory': 'if `True` clean reserved memory in path algorithms',
```
kevinkreiser, line 368:
```suggestion
    'clear_reserved_memory': 'If True clean reserved memory in path algorithms',
```

genadz (approved):

genadz (approved):

kevinkreiser (approved):

PR #2802, opened 2021-01-25
Do not skip connections in bidiastar

genadz:
# Issue

This PR should fix 2 bugs:
**1) for some cases bidirastar builds extra detours.**

After debugging:
in bidirectional astar: we mark edge as `permanent` only before call expansion method. But it's not correct, because we can mark edge earlier right after we popped this edge from the queue.

Current behavior may loose some connections in the following case: let's suppose `A->B->C->D` is the part of the shortest route. On some iteration we popped `B->C` label as `fwd_pred` from the forward queue. Then we applied reverse search from the node `C` (suppose cost for `D->C` less than cost for `B->C`). Then we popped `C->B` label as `rev_pred` from the reverse queue. And, looks like now we should set the connection `B->C`, but we don't do it because the edge `B->C` hasn't marked as `permanent` for now. So, we skip this and continue forward expansion from the node `C`.

So, now imagine that we have 2 not-thru regions `A` and `B` and one connection edge between these regions. In some cases if we skip this connection path cannot be found or may be found with extra detour.

**2) For some cases bidirastar returns pathes with double u-turns.**

It happens due to the following: let's suppose we have a complex restriction `no_u_turn: from AB via BC to CD`. Then, forward search reaches `BC`, checks that it's a deadend (because of the restriction) and makes a u-turn `CB`. After that reverse search reaches `CB`, checks that it's a deadend (because of the restriction), makes a u-turn `BC` and meets `CB` edge from forward search. So, now we have the following path: `AB -> BC -> CB -> BC -> CD` - there is no rule to forbid such path so we accept it. 

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -603,6 +603,9 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
       if (forward_pred_idx != kInvalidLabel) {
         fwd_pred = edgelabels_forward_[forward_pred_idx];
 
+        // Forward path to this edge can't be improved, so we can settle it right now.
+        edgestatus_forward_.Update(fwd_pred.edgeid(), EdgeSet::kPermanent);
+
         // Terminate if the cost threshold has been exceeded.
         if (fwd_pred.sortcost() + cost_diff_ > threshold_) {
           return FormPath(graphreader, options, origin, destination, forward_time_info, invariant);
@@ -632,6 +635,9 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
       reverse_pred_idx = adjacencylist_reverse_.pop();
       if (reverse_pred_idx != kInvalidLabel) {
         rev_pred = edgelabels_reverse_[reverse_pred_idx];
+
+        // Reverse path to this edge can't be improved, so we can settle it right now.
+        edgestatus_reverse_.Update(rev_pred.edgeid(), EdgeSet::kPermanent);
 
         // Terminate if the cost threshold has been exceeded.
         if (rev_pred.sortcost() > threshold_) {
@@ -665,9 +671,6 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
       expand_forward = true;
       expand_reverse = false;
 
-      // Settle this edge.
-      edgestatus_forward_.Update(fwd_pred.edgeid(), EdgeSet::kPermanent);
-
       // setting this edge as settled
       if (expansion_callback_) {
         expansion_callback_(graphreader, "bidirectional_astar", fwd_pred.edgeid(), "s", false);
@@ -687,9 +690,6 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
       // Expand reverse - set to get next edge from reverse adj. list on the next pass
       expand_forward = false;
       expand_reverse = true;
-
-      // Settle this edge
-      edgestatus_reverse_.Update(rev_pred.edgeid(), EdgeSet::kPermanent);
 
       // setting this edge as settled, sending the opposing because this is the reverse tree
       if (expansion_callback_) {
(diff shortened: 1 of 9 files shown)


genadz:
When I made changes some tests failed. The reason: we have at least one more bug. This bug should be fixed here https://github.com/valhalla/valhalla/pull/2796 . So, this PR is blocked right now

genadz:
RAD tests passed. It was only one different route, I checked it manually, no worries.

src/thor/bidirectional_astar.cc
```diff
@@ -758,7 +758,7 @@ bool BidirectionalAStar::SetForwardConnection(GraphReader& graphreader, const BD
 
   // Set a threshold to extend search
   if (threshold_ == std::numeric_limits<float>::max()) {
-    threshold_ = (pred.sortcost() + cost_diff_) + kThresholdDelta;
+    threshold_ = std::max(pred.sortcost() + cost_diff_, opp_pred.sortcost()) + kThresholdDelta;
```
genadz, line 761: as far as we settle edge right after we pop it from the queue, `pred.sortcost() + cost_diff` may be lower than `opp_pred.sortcost()`. so, we should choose bigger in order no to prune search too early

test/gurka/test_route.cc
```diff
@@ -535,3 +535,104 @@ TEST(Standalone, BridgingEdgeIsRestricted) {
...
+  const std::string ascii_map = R"(
+                             F
+                             2
+                             |
+                             E
+                             |
+      A-1--B--------C--------D---------H---------------------------------------I
+                                       |                                       |
+                                       K---------------------------------------J
+  )";
```
genadz, line 551: special testcase that reproduces bug in master

test/gurka/test_route.cc
```diff
@@ -535,3 +535,104 @@ TEST(Standalone, BridgingEdgeIsRestricted) {
...
+  // This test checks that bidirectional astar doesn't allow double uturns.
+  // It might happens if we handle connection edges incorrectly.
+  const std::string ascii_map = R"(
+  J---------------2---------I-------------------------------F
+                            |
+                            |
+  A---1-------B-------------C------D------------------------E
+              |                    |
+              K--------------------L
+  )";
```
genadz, line 599: special testcase that reproduce bug in master

test/gurka/test_route.cc
```diff
@@ -535,3 +535,104 @@ TEST(Standalone, BridgingEdgeIsRestricted) {
...
+  not_thru_edgeids.push_back(std::get<0>(gurka::findEdgeByNodes(*reader, nodes, "D", "C")));
+  not_thru_edgeids.push_back(std::get<0>(gurka::findEdgeByNodes(*reader, nodes, "C", "B")));
+  not_thru_edgeids.push_back(std::get<0>(gurka::findEdgeByNodes(*reader, nodes, "B", "A")));
+  not_thru_edgeids.push_back(std::get<0>(gurka::findEdgeByNodes(*reader, nodes, "D", "E")));
+  not_thru_edgeids.push_back(std::get<0>(gurka::findEdgeByNodes(*reader, nodes, "E", "F")));
+
+  test::customize_edges(map.config, [&not_thru_edgeids](const GraphId& edgeid, DirectedEdge& edge) {
+    if (std::find(not_thru_edgeids.begin(), not_thru_edgeids.end(), edgeid) != not_thru_edgeids.end())
+      edge.set_not_thru(true);
+  });
```
kevinkreiser, line 583: its too bad this didnt happen automatically in the graphenhancer. i suppose you had to do this as a workaround?

kevinkreiser:
@genadz im sorry i have to ask it, hows the performance difference :smile:

test/gurka/test_route.cc
genadz: right. I needed `not_thru` attributes to reproduce the bug

genadz:
> @genadz im sorry i have to ask it, hows the performance difference

the same as `master`

kevinkreiser (approved):

genadz:
@kevinkreiser thanks for review!

PR #2803, opened 2021-01-25
Update visual compare script

kevinkreiser:
The main purpose of this PR is to change the default preference between time and distance in `AutoCost`. At the moment the default is to prefer above all else the shortest time route (barring other attirbute based penalties). This can lead to routes that stick very heavily to highways or higher class roads at the expense of taking a much more direct path to the destiation. I'll update this description with as many anecdotal changes as I can to give a flavor of what this change does in practice.

In addition to that change I wrote a small script that can use the differences found in RAD to generate urls for viewing the route paths on the same map. This makes use of the tool i made in: https://github.com/valhalla/demos/pull/236 

On to the visual diffs (the script makes the previous route red and the new route green):

image

image

image

image

image

image

image

image

image

image

image

image


src/sif/autocost.cc
```diff
@@ -46,7 +46,7 @@ constexpr float kDefaultUseRailFerry = 0.4f; // Default preference of using a ra
 constexpr float kDefaultUseHighways = 1.0f;  // Default preference of using a motorway or trunk 0-1
 constexpr float kDefaultUseTolls = 0.5f;     // Default preference of using toll roads 0-1
 constexpr float kDefaultUseTracks = 0.f;     // Default preference of using tracks 0-1
-constexpr float kDefaultUseDistance = 0.f;   // Default preference of using distance vs time 0-1
+constexpr float kDefaultUseDistance = 0.5f;  // Default preference of using distance vs time 0-1
```
mandeepsandhu, line 49: Dumb q, but wouldn't `0.5` indicate equal preference to both?

mandeepsandhu:
Thanks for making the visual-diff tool! Those before-after paths really show the difference the weightage is making to path selection!

dgearhart:
We will have to revisit when we have more time

gknisely (approved):

PR #2807, opened 2021-01-26
thor::serialize_to_pbf should only be enabled if HAVE_HTTP 

danpat:
...to avoid 'defined but not used' compiler errors.

cc @kevinkreiser @PureW


PureW (approved):

kevinkreiser:
ah crap that was my fault, apologies!

kevinkreiser (approved):

PR #2810, opened 2021-01-27
Racing Directory Creation is OK So Long As the Directory Exists

kevinkreiser:
Through @ktatterso 's tireless effort we have finally found the underlying (at least i think so) issue causing tile building to sometimes fail. Let me explain..

When we are in the first stage of tile building we will spawn a bunch of threads to build tiles at the same time. These threads are all creating the directory structure as defined in GraphTile.h/cc. If there are enough tiles to be created and enough threads its possible that they are both competing to create the same directories at the same time. Lets look at the filename patterns structure:

    level/3_digits/.../3_digits.gph

Now the level determins how many subdirectories you need to make. In practice there are is at least one 3 digit sub directory (for levels 0 and 1) and up to two 3 digit sub directories (for levels 2 and 3). @ktatterso and i theorized that something was clashing when trying to make directories for these tiles and we were right, but at first it didnt seem like it should matter.

Assume two threads are trying to write tiles in the same directory, ie they both race to create the same directory tree. The thing about the directory creation is it doesnt happen all in one go, each sub directory is created one at a time. If any one of those creations fails the function returns false signaling it couldnt make the directory. At first this seems ok, surely the other thread succeded which is why this thread failed BUUUUTTT... you can have two threads trying to write tiles that look like this:

    Thread 1: 2/452/423/987.gph
    Thread 2: 2/452/999/234.gph

Here we have the case that thread 1 races thread 2 for the creation of the subdirectory `452` and wins. Thread 2 then gives up up completely and says well I couldn't do it. But no one else tries to make the next sub directory for thread 2, the `999` part. Because of this when thread 2 tries to open up the file for writing it cant because the deepest subdirectory doesnt exist.

The good news is the fix should be easy. We can simply ignore when we fail to make the directory if it already existed and move on with the rest of the directory creation as normal. I've forgone writing a test for this as its super hard to do with the whole threading concerns.


valhalla/filesystem.h
```diff
@@ -376,9 +376,9 @@ inline bool create_directories(const path& p) {
     if (stat(partial.c_str(), &s) != 0) {
       // create this piece with filesystem::permissions::all
 #ifdef _WIN32
-      if (_mkdir(partial.c_str()) != 0) {
+      if (_mkdir(partial.c_str()) != 0 && errno != EEXIST) {
 #else
-      if (mkdir(partial.c_str(), S_IRWXU | S_IRWXG | S_IRWXO) != 0) {
+      if (mkdir(partial.c_str(), S_IRWXU | S_IRWXG | S_IRWXO) != 0 && errno != EEXIST) {
 #endif
         return false;
```
kevinkreiser, line 383: we only give up if we failed to create a sub directory in the tree AND it was for some other reason than the directory already existed.

merkispavel:
wow! Kevins 1 : 0 bug

ktatterso (dismissed):
I merged these changes and can confirm I no longer see any issues when running valhalla_build_tiles.

valhalla/filesystem.h
```diff
@@ -416,7 +419,11 @@ inline void resize_file(const path& p, std::uintmax_t new_size) {
 }
 
 inline bool remove(const path& p) {
-  return ::remove(p.c_str()) == 0;
+  bool success = ::remove(p.c_str()) == 0;
+  if (!success) {
+    success = errno == ENOENT;
+  }
+  return success;
```
kevinkreiser, line 426: i think this breaks compatibility with stls implementation. i'd prefer that not to be the case. eventually we're going to move to c++17 and this will be a rude gotcha, what do you think @ktatterso ? is this fixing a bug that we are also running into? also it looks like technically we should implement something like this to meet the spec:

```suggestion
  // if it failed for any reason other than the path not existing then throw
  if (remove(p.c_str()) != 0 && errno != ENOENT) {
    throw std::runtime_error("filesystem_error filesytem::remove");
  }
  // otherwise no error means success or not existing means failure to remove
  return errno == 0;
```

ktatterso: I added this check because of a couple reasons.  First, it mirrors the check you added for mkdir, which basically says "if the folder already exists return success".  Hence, the remove logic says, "if you want to remove a directory that doesn't exist return success."

The other reason was I wrote the unit test for concurrent subdirectory remove to expect all threads to succeed.  That expectation could be adapted to whatever we desire, though.

Honestly, I like approach of returning success if asked to remove something that doesn't exist. But writing C++17 compatible code is more important than my feelings :).  As a result, I think these changes to remove() should be reverted - the original implementation was "correct" from the C++17 perspective.

@kevinkreiser I can make these changes if you want?  And update the unit test accordingly.

kevinkreiser: the orginal was correct except it didnt throw when it had an error other than enoent. it was missing the `throw`. :+1: on making the changes if you're good with it!

valhalla/filesystem.h
```diff
@@ -416,22 +422,53 @@ inline void resize_file(const path& p, std::uintmax_t new_size) {
...
+        tries++;
+    } else
+      throw std::runtime_error("filesystem_error filesystem::remove");
+  }
+
+  return ret;
 }
 
-inline bool remove_all(const path& p) {
+inline std::uintmax_t remove_all(const path& p) {
```
ktatterso, line 450: Modified prototype/impl to match C++17 ::remove_all().

valhalla/filesystem.h
ktatterso: Sounds good.

test/filesystem.cc
```diff
@@ -135,6 +137,70 @@ TEST(Filesystem, extension) {
...
+
+    // populate each subfolder with num_files_per_folder text files.
+    {
+      const int num_files_per_folder = 10;
+      std::string folder_name = "test";
+      for (char subdir = 'a'; subdir < 'k'; subdir++) {
+        folder_name = folder_name + "/" + subdir;
+        for (size_t i = 0; i < num_files_per_folder; i++) {
+          std::string filename = folder_name + "/" + std::to_string(i) + ".txt";
+          FILE* f = fopen(filename.c_str(), "w");
```
mookerji, line 182: Nit: we should probably use `std::ofstream` here, which is consistent with using the C++ standard library for I/O throughout the project.

test/filesystem.cc
```diff
@@ -135,6 +137,70 @@ TEST(Filesystem, extension) {
...
+      }
+    }
+
+    // Tell every thread to delete the same "test/a" folder. Each will
+    // take some portion of the effort decided simply by who gets there first.
+    // The key aspect is that this succeeds without error and the "test/a"
+    // folder is successfully deleted.
+    {
+      std::vector<std::unique_ptr<std::thread>> threads(num_threads);
+      for (size_t i = 0; i < num_threads; i++)
```
mookerji, line 195: Nit: surprised this wasn't caught by our linters or formatters, but I'm pretty sure we use still use curly braces for these single-line expression blocks.

valhalla/filesystem.h
```diff
@@ -416,22 +422,53 @@ inline void resize_file(const path& p, std::uintmax_t new_size) {
...
+    // otherwise its a file or link try to delete it
+    else {
+      if (remove(i->path()))
+        num_removed++;
+    }
   }
+
   // delete the root
-  return remove(p);
+  if (remove(p))
```
mookerji, line 468: See prior comment about braces and these single-line expression blocks.

valhalla/filesystem.h
mookerji, line 450: Does this need to be inline?

test/filesystem.cc
```diff
@@ -135,6 +137,70 @@ TEST(Filesystem, extension) {
 TEST(Filesystem, file_size) {
 }
 
+TEST(Filesystem, concurrent_folder_create_delete) {
+
+  const char* create_subdir = "test/a/b/c/d/e/f/g/h/i/j";
```
mookerji, line 142: Can we use `std::string` instead of C-style strings here?

valhalla/filesystem.h
```diff
@@ -416,22 +422,53 @@ inline void resize_file(const path& p, std::uintmax_t new_size) {
...
+  // it can occur that another thread is removing the same file
+  // object at the same time. Occasionally it will occur that
+  // while one thread is deleting (thread 1), the other thread
+  // (thread 2) will receive -1 from ::remove() and (errno==ENOENT
+  // or errno==EINVAL). And yet, the file still exists until thread 1
+  // is finished deleting. In this exact moment, if thread 2 calls
+  // stat() on the file object in question it will get 0 which means
+  // "file object exists". We kinda want thread 2 to wait for the
+  // file object to truly go out of existence. Hence, this spin loop.
+  if (ret == false) {
```
mookerji, line 436: `if (!ret)`?

mookerji:
Made a few comments, annoyingly largely around syntax and formatting. A few of these I could have expected to be caught by `clang-format` and then rejected during CI, so that was a bit surprising.

valhalla/filesystem.h
mookerji: ... and great comment!

danpat:
@mookerji I think the things you're flagging are only caught by `clang-tidy` (for example, [braces around statements](https://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html)), and we only have a [limited number of checks enabled](https://github.com/valhalla/valhalla/blob/master/.clang-tidy#L2), IIRC mostly because `clang-tidy` is historically slooooow.

test/filesystem.cc
PureW: It's not configured for our linters, the usage is mixed throughout the codebase with many single line if-expressions without braces.

test/filesystem.cc
PureW: Nit: `create_subdir` is framed as a verb, I usually think of verbs as functions or similar

test/filesystem.cc
```diff
@@ -135,6 +137,70 @@ TEST(Filesystem, extension) {
...
+
+  size_t num_threads = std::thread::hardware_concurrency();
+
+  const int num_iters = 50;
+  for (int j = 0; j < num_iters; j++) {
+    // concurrent folder creation
+    {
+      std::vector<std::unique_ptr<std::thread>> threads(num_threads);
+      for (size_t i = 0; i < num_threads; i++)
+        threads[i].reset(new std::thread(create_folder));
```
PureW, line 168: There's a risk that no concurrency at all happens if each thread finishes before the next is created since there is no synchronization. I guess the filesystem operations likely prevent this from happening.

test/filesystem.cc
```diff
@@ -135,6 +137,70 @@ TEST(Filesystem, extension) {
...
+  auto remove_folder = [&]() {
+    std::uintmax_t delete_count = filesystem::remove_all(base_subdir);
+    // I've found that two+ threads will claim that they've deleted the same
+    // file object. This results in double+ counting and prevents me from
+    // asserting anything related to the delete_count. All we can be sure
+    // is the following:
+    EXPECT_TRUE(!filesystem::exists(base_subdir));
+  };
+
+  size_t num_threads = std::thread::hardware_concurrency();
```
PureW, line 160: Curious what other people think about `hardware_concurrency` in a test?

Since we're not compute bound here, there's not really an use in using the native number of threads, and it likely leads to the test behaving differently when running on a 2 core intel vs a 32 core amd.

kevinkreiser: its not good. if CI doesnt have the resources to do it will pretty much fork bomb the machine.

test/filesystem.cc
kevinkreiser: yeah lets not go through another round of format the whole codebase :smile:

test/filesystem.cc
```diff
@@ -135,6 +137,70 @@ TEST(Filesystem, extension) {
...
+    }
+
+    // Tell every thread to delete the same "test/a" folder. Each will
+    // take some portion of the effort decided simply by who gets there first.
+    // The key aspect is that this succeeds without error and the "test/a"
+    // folder is successfully deleted.
+    {
+      std::vector<std::unique_ptr<std::thread>> threads(num_threads);
+      for (size_t i = 0; i < num_threads; i++)
+        threads[i].reset(new std::thread(remove_folder));
```
PureW, line 196: nit: Similar caveat as earlier about no synchronization to start the race.

valhalla/filesystem.h
kevinkreiser: yes because its a free function in the header

test/filesystem.cc
kevinkreiser: perhaps we can give it the disabled label so that the test exists but we dont run it in CI

PureW: I was thinking of a fixed number of threads so it's independent of the machine it runs on.

Is resources an issue for this test? It's not forking the process, simply running the filesystem operation in a lightweight thread in this test, right?

test/filesystem.cc
ktatterso: Probably so.  Y'all are gonna get a kick out of my old school programming.  :)

test/filesystem.cc
ktatterso: I don't conventionally use braces for single-line blocks, but I can.

test/filesystem.cc
ktatterso: Here are my thoughts.  These are just threads - so quite lightweight.  Each thread is doing a trivial number of trivial things on disk.  The only memory consumption of note is the weight of each thread.  Creating 100 threads is only a touch slower than 8.  (I timed it.)

Anders' suggestion to "start the race" is a good one.  I mean, there are a ton of variables outside our control to force/guarantee concurrency - but we can control that one.  I'll implement that.

I don't have strong feelings either way regarding using a fixed number of threads vs the number available in the system.

test/filesystem.cc
ktatterso: Good suggestion, thanks.

test/filesystem.cc
ktatterso: Right, I'll implement a "starting gun" event.  Thanks.

test/filesystem.cc
ktatterso: The decision to use a const std::string vs const char *... it depends on a lot of factors.  I generally program in a manner that is "as simple as possible".  I don't need the features offered by std::string in this case, so I decided to use a const char * string literal.  However, I can sense this team is big on using C++ across the board.  

My philosophy/coding style is to best match the existing conventions within the team/project - so I'm happy to use std::string in these instances.

Fair point about the name, Anders.

test/filesystem.cc
kevinkreiser: yeah i didnt literally mean fork bomb but more so the idea. i would agree if we can rewrite it some how to force trhe race that would be better.

there is a  huge problem with creating large numbers of threads which is that ive seen it cause problems on CI. even with larger machines on CI we use full parallelism when we run `make check` this means that all the threads that we could be using, we are already using. if you then start this test you are doubling that, which causes cpu contention and can lead to CI stalling out. it has happend to us many times in the past. since these are short lived threads maybe it wont be a problem but random CI failures isnt fun for anybody.

ktatterso: Fair points K1.  Maybe we should run "make check" with less than all available cores/threads?  The runtime of make-check might increase a bit, but having the extra processing headroom should theoretically help cut down on stalls.

Honestly, in my experience, hyperthreads do help with scaling but not as much as you'd wish/want.  I might even go so far as to suggest running make-check with half the number of available system "cores", which is (at least in this day/age) the actual number of real cpu cores.

Combining K1's and my thoughts, I'm thinking that this test should only use half of the available system cores.

Curious everyone's thoughts on configuring CI's make-check to use half the available system cores?  Or other thoughts?

PureW: > Curious everyone's thoughts on configuring CI's make-check to use half the available system cores? Or other thoughts?

Hmm maybe? I don't have enough to form an opinion here. As long as we don't decrease the number of processes used for compiling code.

test/filesystem.cc
mandeepsandhu: I'm inclined to agree with @ktatterso . For this specific use-case, i.e string constants, `const char*` (or maybe `const char[]` if we want to use size, which is compile time optimized)) is more light-weight than `std::string`. If the string needs to be modified or accepted as a function arg (even a const arg), then I'd prefer to use `std::string` (because of its implicit memory mgmt). Just my 2¢.

test/filesystem.cc
mandeepsandhu: I think since the thread is making a syscall (to do file IO), its highly likely that other processes (and in this case threads) get scheduled in during that time.

valhalla/filesystem.h
mandeepsandhu: also, :heart:  the 80-char wrapping :slightly_smiling_face:

test/filesystem.cc
mookerji: To be clear, I think the issue here is around style/consistency of our type choices over performance concerns. The latter isn't really an issue, since this is a unit test. On the other hand, we seem more likely to use an STL string container over C-style strings in many places where we have string constants (there are, of course, exceptions to this).

Just pointing this out as a theme, I have no strong preferences about what we're doing here.

test/filesystem.cc
mookerji: @danpat called this out elsewhere: it's actually caught by `clang-tidy`. In any case, I thought this was being enforced by CI, feel free to ignore.

test/filesystem.cc
ktatterso: Mandeep is probably right - the file i/o slows things down enough that concurrency is almost guaranteed.  That said, Anders point is still valid - and there's no harm in sophisticating the test a bit to maximize concurrency.  Hence my addition of the "start_race_event" condition variable.

test/filesystem.cc
ktatterso: I've reduced the number of worker threads to num-available / 2.  On a separate channel we can continue thoughts on how many cores/threads to use when CI is running tests.

test/filesystem.cc
```diff
@@ -135,6 +137,139 @@ TEST(Filesystem, extension) {
...
+      start_race_mutex.lock();
+      start_race = true;
+      start_race_mutex.unlock();
+      start_race_event.notify_all();
+
+      for (size_t i = 0; i < num_threads; i++) {
+        threads[i]->join();
+      }
+
+      EXPECT_TRUE(filesystem::exists(nested_subdir));
```
mandeepsandhu, line 222: I guess this should be an `ASSERT_TRUE` (instead of `EXPECT_...`) since we don't want to proceed ahead if the dir creation did not fully succeed (else the test might fail when trying to create files below).

I guess we can look at other `EXPECT_...` statements and change them to `ASSERT_...` wherever applicable.

test/filesystem.cc
```diff
@@ -135,6 +137,139 @@ TEST(Filesystem, extension) {
 TEST(Filesystem, file_size) {
 }
 
+TEST(Filesystem, concurrent_folder_create_delete) {
```
mandeepsandhu, line 140: Great test! Very clean and readable. Thank you!

BTW, do we need a test-level cleanup function to remove any leftover artifacts of the test? Or would rather have the directories around in case its useful for debugging test failures?

valhalla/filesystem.h
```diff
@@ -416,22 +422,53 @@ inline void resize_file(const path& p, std::uintmax_t new_size) {
...
+
+  // it can occur that another thread is removing the same file
+  // object at the same time. Occasionally it will occur that
+  // while one thread is deleting (thread 1), the other thread
+  // (thread 2) will receive -1 from ::remove() and (errno==ENOENT
+  // or errno==EINVAL). And yet, the file still exists until thread 1
+  // is finished deleting. In this exact moment, if thread 2 calls
+  // stat() on the file object in question it will get 0 which means
+  // "file object exists". We kinda want thread 2 to wait for the
+  // file object to truly go out of existence. Hence, this spin loop.
```
kevinkreiser, line 435: while i get that this is a nice feature i really dont think it should be the filesystem libraries concern if you write code that is racing to delete the same paths. that synchronization should be handled in the calling application. indeed i had a look at both the boost implementation as well as another filesystem implementation (https://github.com/gulrak/filesystem) and neither of them implement such a tactic: 
https://github.com/gulrak/filesystem/blob/de57485877865156370f5e9878592293645b207e/include/ghc/filesystem.hpp#L4396
https://github.com/boostorg/filesystem/blob/develop/src/operations.cpp#L229

@ktatterso what was the issue that triggered you looking into this anyway? i've kind of lost the plot at this point :smile:  i cant think of any multithreaded calls to `remove`. in mjolnir at the beginning of the process we `remove_all` existing tile directories in the main thread... what am i missing here?

kevinkreiser: yeah it seems like its not something we really need to worry about at this time. the only call was the one that i remembered off the top of my head, the rest are all in tests which definitely shouldnt race because they definitely shouldnt be using the same tile directories:

image

now i get that you could be racing at the process level but again should filesystem really be the one protecting against that? i guess since it has such a small impact its probably ok but its again likely a deviation from the standard implimentation. i'm ok with whichever direction you want to go. want to just ship what we've got?

test/filesystem.cc
ktatterso: Agree, thanks.

test/filesystem.cc
ktatterso: Thanks and good thought.  However, as written, the test (concurrently) creates the nested subfolder (starting at "test/a"), populates it, then concurrently removes the "test/a" folder.  If the test is working, there shouldn't be any artifacts leftover.

valhalla/filesystem.h
ktatterso: Multithreaded folder creation motivated the creation of the unit test.  Deciding to test multithreaded directory removal was simply my idea.  I figured if we want to be robust in multithreaded directory creation we would also want to be robust in multithreaded directory removal.  In a sense, valhalla's filesystem::remove_all() is the "calling application" - so imo we should take responsibility for it being robust.

In my mind, I wrote a test that should work (threads concurrently calling filesystem::remove_all()).  It exposed an issue and I've resolved the issue.  No need to wait for a customer to find this issue, we can fix it now.

kevinkreiser:
@ktatterso can you put a ship on this so we can merge it when the build finishes?

ktatterso (approved):

ktatterso:
As requested I switched the EXPECT_* macros to ASSERT_*.  Then I merged with master.  And although the pipeline is still running I've approved the PR.

PR #2811, opened 2021-01-27
Fix Recently Added Valhalla Alternates Serialization

kevinkreiser:
In adding #2734 we needed a way to represent alternate routes in the valhalla serializer. In the first iteration of that PR it was added while we were still using the old json serializer. Before I merged the PR though many other PRs made it into master which caused a huge merge parade. As a part of that merge process I needed to reconcile the new json serialization (using rapidjson) with the old code. I thought I had done that but infact I was wrong and there wasnt any test coverage of it. I've added test coverage in this pr and also fixed the serialization.

### DIFF
--- a/src/tyr/route_serializer_valhalla.cc
+++ b/src/tyr/route_serializer_valhalla.cc
@@ -549,16 +549,16 @@ std::string serialize(const Api& api) {
     writer.start_object("trip");
 
     // the locations in the trip
-    locations(api, 0, writer);
+    locations(api, i, writer);
 
     // the actual meat of the route
-    legs(api, 0, writer);
+    legs(api, i, writer);
 
     // openlr references of the edges in the route
-    valhalla::tyr::openlr(api, 0, writer);
+    valhalla::tyr::openlr(api, i, writer);
 
     // summary time/distance and other stats
-    summary(api, 0, writer);
+    summary(api, i, writer);
 
     writer.end_object(); // trip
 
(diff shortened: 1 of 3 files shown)


gknisely (dismissed):

src/tyr/route_serializer_valhalla.cc
```diff
@@ -538,15 +537,15 @@ void legs(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper_t
...
 
   // for each route
   for (int i = 0; i < api.directions().routes_size(); ++i) {
     if (i == 1) {
       writer.start_array("alternates");
     }
 
-    // the main route
+    // the route itself
+    writer.start_object();
```
kevinkreiser, line 548: every route needs to be wrapped in an object `{` we had only one for the outer object in the response but for each of the alternates we didnt add one. this lead to json that looked like:

```json
{
  "trip": .... main route here,
  "alternates": [
    "trip",
    {guts of alternate here},
```

test/alternates.cc
```diff
@@ -50,12 +52,21 @@ void test_alternates(int num_alternates) {
...
+  rapidjson::Document doc;
+  doc.Parse(json);
+  ASSERT_FALSE(doc.HasParseError());
+  ASSERT_TRUE(doc.HasMember("trip"));
+  ASSERT_TRUE(doc.HasMember("alternates") == (num_alternates > 0));
+  if (num_alternates > 0) {
+    for (const auto& alt : doc["alternates"].GetArray()) {
+      ASSERT_TRUE(alt.HasMember("trip"));
+    }
+  }
```
kevinkreiser, line 69: here we actually check that the json isnt bunk

dnesbitt61 (approved):

PR #2813, opened 2021-01-27
nit: Links debug build with lld

PureW:
Lld (from clang) is a _much_ faster linker which
improves our `check` target build a lot.

As each test-suite is its own target, linking faster
leads to markedly faster test-times.

From my local test of a incremental `ninja check` after
touching triplegbuilder.cc I run all the tests
in 146 seconds, compared to 275 seconds without lld


.circleci/config.yml
```diff
@@ -92,8 +92,10 @@ jobs:
       - run: |
           # NOTE: -Werror disabled in CI, as we currently have >4k warnings.
           cd build \
+            && apt-get update && apt-get install --assume-yes lld \
```
kevinkreiser, line 95: should add this to the docker build container dependencies script

.circleci/config.yml
```diff
@@ -92,8 +92,10 @@ jobs:
...
             && cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=On -DCPACK_GENERATOR=DEB \
                 -DENABLE_COMPILER_WARNINGS=On -DENABLE_WERROR=Off -DCMAKE_EXPORT_COMPILE_COMMANDS=On \
+                -DCMAKE_CXX_FLAGS="-fuse-ld=lld" \
```
kevinkreiser, line 98: should we default to this inside of cmake if we detect its presence, also why not the other builds?

PureW: It seemed far from obvious how to do it inside `cmake` except for manually enabling it for every single target. [Stackoverflow thread](https://stackoverflow.com/questions/1867745/cmake-use-a-custom-linker)

 A workaround was to use this compiler flag which works if you haven't already made specific linker configuration inside cmake

.circleci/config.yml
PureW: Yup, I added this under the commit marked as revert. For now wanted to test it in ci and get PR feedback before tweaking the dockerfile (which ideally is in a separate PR merged first).

Flow could be:
 1) Merge second PR that adds `lld` to dockerfile
 2) That triggers an update of the base-image on merge to master
 3) Use updated baseimage with `lld` in this PR.

PureW: Once https://github.com/valhalla/valhalla/pull/2819 is merged I can rebase and remove this `apt-get install`

PureW:
Ok, rebased on master where `lld` now should be part of the docker base image. Also enabled `lld` for the other linux targets. The greatest ROI will still be in testing though where the linker is called for each test-file which is its own binary.

PureW:
Hmm why did geos stop linking in release :confused:

kevinkreiser:
yeah i havent the slightest clue why lld cant find the symbols for geos in release mode that doesnt make sense to me. regardless of the build, the geos library is the same one. i would suggest we do `VERBOSE=1 make` to see whats on the link line and see why it cant find geos

.circleci/config.yml
mandeepsandhu: For some reason I wasn't able to see a speedup when testing with lld locally.

Without `lld`:
```shell
$ cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -DLOGGING_LEVEL=ERROR -DENABLE_NODE_BINDINGS=Off -DENABLE_PYTHON_BINDINGS=Off -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DBUILD_SHARED_LIBS=On
...
$ make clean
$ time make -j(nproc) check
...
Executed in   67.45 secs   fish           external 
   usr time  267.32 secs  1355.00 micros  267.32 secs 
   sys time   52.82 secs  397.00 micros   52.82 secs
```

With `lld`:
```shell
$ cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -DLOGGING_LEVEL=ERROR -DENABLE_NODE_BINDINGS=Off -DENABLE_PYTHON_BINDINGS=Off -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DBUILD_SHARED_LIBS=On -DCMAKE_CXX_FLAGS="-fuse-ld=lld"
...
$ make clean
$ time make -j(nproc) check
Executed in   67.14 secs   fish           external 
   usr time  266.98 secs  1792.00 micros  266.98 secs 
   sys time   52.11 secs    0.00 micros   52.11 secs 
```

mandeepsandhu: ok, so the reason I dont see a speedup is because I have `-DBUILD_SHARED_LIBS=On` which builds a shared lib (and does not have as much a hit on link as a static lib).

@PureW has confirmed that speedup with static lib is significant. Lets roll this change! :+1:

mandeepsandhu (dismissed):

.circleci/config.yml
mandeepsandhu: @PureW I've had trouble using `lld` to link valhalla as a **shared lib** after the changes introduced in https://github.com/valhalla/valhalla/pull/2787 (I can't see any issue with the PR itself though, but the link failure happens after this change was introduced).

Can you try it on your setup and see if valhalla links successfully with `-DBUILD_SHARED_LIBS=On`?

PureW: @mandeepsandhu simply from turning on `BUILD_SHARED_LIBS` gives me linking errors with geos which seem to be caused by cmake not passing geos to the linker command AFAICT

> FAILED: valhalla_add_predicted_traffic   
> : && /usr/bin/c++ -fuse-ld=lld -fdiagnostics-color=auto  -g  CMakeFiles/valhalla_add_predicted_traffic.dir/src/mjolnir/valhalla_add_predicted_traffic.cc.o -o valhalla_add_predicted_traffic  -Wl,-rpath,/home/anders/Documents/Programming/mapbox/valhalla/build-clang/src:  src/libvalhalla.so.3.1.0  /usr/lib/libboost_program_options.so.1.75.0  /usr/lib/libprotobuf-lite.so  /usr/lib/libcurl.so  /usr/lib/libz.so  /usr/lib/libprime_server.so  /usr/lib/libspatialite.so  /usr/lib/libsqlite3.so  /usr/lib/libluajit-5.1.so  -lm  -pthread && :
> ld.lld: error: src/libvalhalla.so.3.1.0: undefined reference to geos::geom::GeometryFactory::destroy() [--no-allow-shlib-undefined]

 :disappointed: 

I began this work to bring down the test-times initially. Maybe we should instead simply turn on `BUILD_SHARED_LIBS` and stay with `ld` for the tests as shared libs avoid the expensive static link time.

mandeepsandhu: yeah I assume it has something to do with how we're [importing libgeos into cmake](https://github.com/valhalla/valhalla/blob/master/CMakeLists.txt#L308). Maybe `lld` doesn't like the generated link command for it. Yeah switching to shared lib for tests seems like a low-hanging fruit.

PureW:
Hmm something changed with geos and now CI passes again. @mandeepsandhu Want to approve again?

mandeepsandhu:
> Hmm something changed with geos and now CI passes again. @mandeepsandhu Want to approve again?

@PureW last time I tried locally, `lld` was having issues linking in shared lib mode. Has that changed on CI? And does `lld`link shared libvalhalla successfully?  (I saw another PR trying to enable shared lib, but looks like it was abandoned)

kevinkreiser (approved):

PureW:
> > Hmm something changed with geos and now CI passes again. @mandeepsandhu Want to approve again?
> 
> @PureW last time I tried locally, `lld` was having issues linking in shared lib mode. Has that changed on CI? And does `lld`link shared libvalhalla successfully? (I saw another PR trying to enable shared lib, but looks like it was abandoned)

Shared lib and lld seem to not play together. I opened the other PR to see if shared lib with regular linking would be faster than this but I got another obscure link error in CI.

Given that the end-goal is to make CI faster, and everyone is free to link however they want locally, I'd say we should merge this.

PR #2815, opened 2021-01-27
fix a regresion introduced in #2743

kevinkreiser:
I did some optimizations in #2743 but made a very important typo when doing so. I missed that the function call we used for timedependent algorithms to stop downward expansion was the one that took the predecessory labels distance. This caused the router to cull search space that was needed for it to find the destination (or origin in the reverse search)

### DIFF
--- a/src/thor/timedep_forward.cc
+++ b/src/thor/timedep_forward.cc
@@ -104,7 +104,8 @@ bool TimeDepForward::ExpandForward(GraphReader& graphreader,
       // if this is a downward transition (ups are always allowed) AND we are no longer allowed OR
       // we cant get the tile at that level (local extracts could have this problem) THEN bail
       graph_tile_ptr trans_tile = nullptr;
-      if ((!trans->up() && hierarchy_limits_[trans->endnode().level()].StopExpanding()) ||
+      if ((!trans->up() &&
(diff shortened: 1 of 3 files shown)


gknisely (approved):

src/thor/timedep_forward.cc
```diff
@@ -104,7 +104,8 @@ bool TimeDepForward::ExpandForward(GraphReader& graphreader,
       // if this is a downward transition (ups are always allowed) AND we are no longer allowed OR
       // we cant get the tile at that level (local extracts could have this problem) THEN bail
       graph_tile_ptr trans_tile = nullptr;
-      if ((!trans->up() && hierarchy_limits_[trans->endnode().level()].StopExpanding()) ||
+      if ((!trans->up() &&
+           hierarchy_limits_[trans->endnode().level()].StopExpanding(pred.distance())) ||
```
kevinkreiser, line 108: in the previous PR i accidentally removed the distance argument :facepalm:

kevinkreiser: https://github.com/valhalla/valhalla/pull/2743/files#diff-1d57d4cfc830f3baf54a6c0378f8fb3c54aefcd414483503b2a02d8da591c280L115

PR #2816, opened 2021-01-28
Add support for closure annotations

mandeepsandhu:
Adds a new shape attribute filter for getting closure locations in the
directions response. Similar to incidents, each closure annotation has
a start & end geometry index, which indexes into the coordinates list.

### DIFF
--- a/src/thor/triplegbuilder.cc
+++ b/src/thor/triplegbuilder.cc
@@ -125,6 +125,17 @@ void UpdateIncident(const std::shared_ptr<const valhalla::IncidentsTile>& incide
   }
 }
 
+valhalla::TripLeg_Closure* fetch_last_closure_annotation(TripLeg& leg) {
+  return leg.closures_size() ? leg.mutable_closures(leg.closures_size() - 1) : nullptr;
+}
+
+valhalla::TripLeg_Closure* fetch_or_create_closure_annotation(TripLeg& leg) {
+  valhalla::TripLeg_Closure* closure = fetch_last_closure_annotation(leg);
+  // If last closure annotation has its end index populated, create a new
+  // closure annotation
+  return (!closure || closure->has_end_shape_index()) ? leg.add_closures() : closure;
+}
+
 /**
  * Chops up the shape for an edge so that we have shape points where speeds change along the edge
  * and where incidents occur along the edge. Also sets the various per shape point attributes
@@ -172,6 +183,7 @@ void SetShapeAttributes(const AttributesController& controller,
     double speed; // meters per second
     uint8_t congestion;
     std::vector<const valhalla::IncidentsTile::Location*> incidents;
+    bool closed;
   };
 
   // A list of percent along the edge, corresponding speed (meters per second), incident id
@@ -187,15 +199,20 @@ void SetShapeAttributes(const AttributesController& controller,
       cuts.emplace_back(cut_t{traffic_speed.breakpoint1 / 255.0,
                               speed,
                               static_cast<std::uint8_t>(traffic_speed.congestion1),
-                              {}});
+                              {},
+                              traffic_speed.closed(0)});
       if (traffic_speed.breakpoint2 > 0) {
         cuts.emplace_back(cut_t{traffic_speed.breakpoint2 / 255.0,
                                 speed,
                                 static_cast<std::uint8_t>(traffic_speed.congestion2),
-                                {}});
+                                {},
+                                traffic_speed.closed(1)});
         if (traffic_speed.speed3 != UNKNOWN_TRAFFIC_SPEED_RAW) {
-          cuts.emplace_back(
-              cut_t{1, speed, static_cast<std::uint8_t>(traffic_speed.congestion3), {}});
+          cuts.emplace_back(cut_t{1,
+                                  speed,
+                                  static_cast<std::uint8_t>(traffic_speed.congestion3),
+                                  {},
+                                  traffic_speed.closed(2)});
         }
       }
     }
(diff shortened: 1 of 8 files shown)


src/thor/triplegbuilder.cc
```diff
@@ -289,6 +307,25 @@ void SetShapeAttributes(const AttributesController& controller,
...
     }
+    if (controller.attributes.at(kShapeAttributesClosure)) {
+      // Process closure annotations
+      if (cut_itr->closed) {
+        // Found a closure. Fetch a new annotation, or the last closure
+        // annotation if it does not have an end index set (meaning the shape
+        // is still within an existing closure)
+        ::valhalla::TripLeg_Closure* closure = fetch_or_create_closure_annotation(leg);
+        if (!closure->has_begin_shape_index()) {
+          closure->set_begin_shape_index(i - 1);
```
PureW, line 318: How come it's `i - 1` and not just `i`?

test/gurka/test_closure_annotations.cc
```diff
@@ -0,0 +1,327 @@
...
+
+#include <boost/format.hpp>
+#include <gtest/gtest.h>
+
+using namespace valhalla;
+using LiveTrafficCustomize = test::LiveTrafficCustomize;
+
+namespace {
+
+inline void SetLiveSpeedFrom(baldr::TrafficSpeed* live_speed, uint64_t speed, uint8_t subsegment) {
```
PureW, line 12: Any chance `subsegment` argument can be renamed to `breakpoint1` just to decrease number of terms describing the same thing?

Also, you're not setting the speed of the first subsegment here even though you do set the breakpoint1, is that intended? You may want to set `speed1` to either unknown or 0 (or something else) for clarity

test/gurka/test_closure_annotations.cc
```diff
@@ -0,0 +1,327 @@
...
+  live_speed->breakpoint1 = subsegment;
+  live_speed->breakpoint2 = 255;
+  live_speed->speed2 = speed >> 1;
+  live_speed->speed3 = UNKNOWN_TRAFFIC_SPEED_RAW;
+}
+
+inline void SetLiveSpeedUpto(baldr::TrafficSpeed* live_speed, uint64_t speed, uint8_t subsegment) {
```
PureW, line 19: same note about name here

PureW: The speed datatype should probably be uint8_t too

PureW: How about `overall_speed`?

PureW: `SetLiveSpeed` below can probably be collapsed into this function if you set `subsegment/breakpoint1` argument to default to 255?

test/gurka/test_closure_annotations.cc
```diff
@@ -0,0 +1,327 @@
...
+                                 const gurka::map& map) {
+  baldr::GraphId tile_id(tile.header->tile_id);
+  auto edge = std::get<0>(gurka::findEdge(reader, map.nodes, edge_name, end_node));
+  if (edge.Tile_Base() == tile_id && edge.id() == index) {
+    uint8_t subsegment = static_cast<uint8_t>(255 * percent_along);
+    SetLiveSpeedFrom(current, 0, subsegment);
+  }
+}
+
+void close_partial_dir_edge_upto(baldr::GraphReader& reader,
```
PureW, line 46: Just an FYI:

 There are some utility functions already for customizing live traffic like `test::customize_live_traffic_data(map.config, cb_setter_max);` as used in https://github.com/valhalla/valhalla/blob/master/test/gurka/test_traffic.cc#L56

No

PureW: Ohh I see that you are in fact using that below.

test/gurka/test_closure_annotations.cc
```diff
@@ -0,0 +1,327 @@
+#include "gurka.h"
```
PureW, line 1: Could we apply more compiler warnings to this file at https://github.com/valhalla/valhalla/blob/master/test/gurka/CMakeLists.txt#L14 ?

src/thor/triplegbuilder.cc
mandeepsandhu: Since this loops [starts from the 2nd shape index](https://github.com/valhalla/valhalla/pull/2816/files#diff-fd61a9dafa6119b300d3fcc49f59456388c00d6851fb8cc325141be44daf58c4R295). It does that since we create cuts only if speed changes in subsequent shape segments.

test/gurka/test_closure_annotations.cc
mandeepsandhu: totally! will add it

test/gurka/test_closure_annotations.cc
mandeepsandhu: yup :slightly_smiling_face: 

I added these functions here since I wanted partial closures. If this is reused elsewhere in the future, we could move it to the gurka module itself.

test/gurka/test_closure_annotations.cc
mandeepsandhu: > Any chance `subsegment` argument can be renamed to `breakpoint1` just to decrease number of terms describing the same thing?
> 

Yeah, I can change `subsegment` to `breakpoint1` (this was just me trying to reverse engg the `TrafficSpeed` struct and trying to make sense of what breakpoint meant :stuck_out_tongue: ).

> Also, you're not setting the speed of the first subsegment here even though you do set the breakpoint1, is that intended? You may want to set `speed1` to either unknown or 0 (or something else) for clarity

So this function intends to set live speed to `0` **starting from** a particular subsegment/breakpoint of the edge. Anything before that should remain whatever value it previously had. TBH I arrived at the vlaues being set after a few iterations. Let me know if there's a better/correct/simpler way to do it. I want to partially close edges upto/from a certain percentage of the edge.

test/gurka/test_closure_annotations.cc
mandeepsandhu: I'm not sure what `overall_speed` should be set to if its already populated (I do that when constructing the gurka map in the setup function) and we want to close only part of the edge. Thats why I left it as is

PureW: The most important is probably to set `overall_speed` to 0 if any of the subsegments have zero speed. That marks the edge as closed if a subsegment is closed. Apart from that, it's not used if you have more than one subsegment.

PureW (approved):

PureW (approved):

PR #2826, opened 2021-02-01
Wrap traffic.coefficients with optional

merkispavel:
The reason is we always add traffic data to the tile because the condition `coefficients.size() == kCoefficientCount` is always true in case with std::array

# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


gknisely (approved):

kevinkreiser:
Good catch sorry for the trouble there!

danpat:
Adding for the record that this PR fixes a bug introduced here https://github.com/valhalla/valhalla/pull/2712/files/b9cc04125755d3c7230557521682ca7709206d80#r566566669

PR #2829, opened 2021-02-02
more test fixes for historical traffic regression

kevinkreiser:
this is an addendum to a previous fix that @merkispavel did in #2826 for the regression that i introduced in #2712 basically it does the same thing but also for the test helper functions


merkispavel (approved):
thx

merkispavel:
It looked safe to merge it so I did. Let me know if I shouldn't do it in the future ;)

kevinkreiser:
@merkispavel nope thats great thank you. if its not ready ill put it in draft or say something explicitly in the pr description.

Issue #699, opened 2017-05-05
Allow GTFS Import into Valhalla


dnesbitt61:
Several changes are needed to allow import of GTFS into Valhalla. Here is a proposed design:

- Publish the transit tile protocol buffer format (pbf).
- Program an application to convert GTFS into pbf tiles.
- Split the _valhalla_build_transit_ application into 2 parts: 1st part queries Transitland and generates transit pbf tiles. 2nd part creates Valhalla transit graph tiles from the transit pbf tiles.

araichev:
Can some at Valhalla do the first task above: publish the transit tile protocol buffer format?
Then i could look into the second task.

dnesbitt61:
We hope to have a proto specification in the next week or tow, but then we have to write the importer from that specification into Valhalla tiles. This is on our near term plan so we hope to get to that soon.

phildonovan:
@dnesbitt61 Any progress regarding the protocols? It would be fantastic to be able to do this, and I'm happy to help Alex. Of course, if I could help with the protocols then I'd be happy to as well.

dnesbitt61:
We are starting work on this after we finish a couple of Open Traffic tasks. Hopefully have something in November. Ping us in a couple of weeks and we can see if there are any areas we need assistance.

araichev:
Hi @dnesbitt61 , any word on this issue?

dnesbitt61:
Well, we have moved to Mapbox now that Mapzen is shutting down. So there has been no progress. Will have to see where transit falls in our priority list when things settle. Transitland is also in a holding pattern, though I think a couple folks are going to continue to maintain that.

Rui-Santos:
> Transitland is also in a holding pattern, though I think a couple folks are going to continue to maintain that.

Well, if you could make Valhalla not download the whole world transit data from the Transitland API I'm sure that would help them keep it live longer. It doesn't even make any sense. Why not use the same restriction that are used to build the admins DB? I understand the feed from transitland gives the whole data they have, but why not pick it by what osm.pbf files are loaded?

dnesbitt61:
Transit data is retrieved from Transitland using the valhallla_fetch_transit application. This is run independently and prior to valhalla_build_transit and valhalla_build_tiles. There is logic in this application to specify a single feed. If you would like to work on some updates to this process to accept a bounding box or some other geographic constraint to limit the volume of data downloaded from Transitland we would be likely to accept such an update.

Rui-Santos:
[...]

PR #2835, opened 2021-02-04
feat: GTFS support

pranavpandey1998official:
# Issue

GTFS #699
## Tasklist

 - [ ] Move gtfs parser to separate module
 - [ ] support for frequency based trips

## Requirements / Relations

note -
this is my first PR ever in c++ so bear with me
tested with http://www.bart.gov/dev/schedules/google_transit.zip

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,7 +266,7 @@ set(valhalla_programs valhalla_run_map_match valhalla_benchmark_loki valhalla_be
 ## Valhalla data tools
 set(valhalla_data_tools valhalla_build_statistics valhalla_ways_to_edges valhalla_validate_transit
   valhalla_benchmark_admins valhalla_build_connectivity	valhalla_build_tiles valhalla_build_admins
-  valhalla_convert_transit valhalla_fetch_transit valhalla_query_transit valhalla_add_predicted_traffic)
+  valhalla_convert_transit valhalla_fetch_transit valhalla_gtfs_transit valhalla_query_transit valhalla_add_predicted_traffic)
 
 ## Valhalla services
 set(valhalla_services valhalla_loki_worker valhalla_odin_worker valhalla_thor_worker)
--- a/src/mjolnir/valhalla_gtfs_transit.cc
+++ b/src/mjolnir/valhalla_gtfs_transit.cc
@@ -1,17 +1,17 @@
 #include <iostream>
-#include <string>
-#include <vector> 
 #include <queue>
+#include <string>
 #include <thread>
+#include <vector>
 
-#include <boost/property_tree/ptree.hpp>
 #include <boost/algorithm/string.hpp>
+#include <boost/property_tree/ptree.hpp>
 
-#include "baldr/rapidjson_utils.h"
-#include "baldr/gtfs.h"
 #include "baldr/graphid.h"
-#include "baldr/tilehierarchy.h"
 #include "baldr/graphtile.h"
+#include "baldr/gtfs.h"
+#include "baldr/rapidjson_utils.h"
+#include "baldr/tilehierarchy.h"
 #include "midgard/logging.h"
 
 #include "mjolnir/transitpbf.h"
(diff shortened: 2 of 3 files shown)


kevinkreiser:
ill check this out and give it a try. at the end of the day, id like to suck this logic out into a function so that we can call it in a unit test with a fake gtfs feed to verify the functionality. i might have some time this weekend to do that

nilsnolde:
@pranavpandey1998official can you ping me on nils@gis-ops.com pls? I'd like to talk about this PR with you in private, we might be able to support here. Sorry to do it this way, but there's no public email associated with your account.

farfromrefug:
@kevinkreiser @nilsnolde any news on this? it is really hard to build transit tiles with transitland now because of the api rate limit

kevinkreiser:
@farfromrefug no news but i recently did a survey of the gtfs spec in which i cataloged all of the bits we need to write the ingestion tool. i can document that here and then come up with a design. i know @pranavpandey1998official did some work on it already so we'll have to see how it lines up but i think this is a big enough thing that we might want a little bit of a design layout first before we implement it

nilsnolde:
That’d be beautiful @kevinkreiser .. unfortunately that potential client never contacted us again, and we had to prioritize other features. But still very much up for a mid-term collab on GTFS support!

farfromrefug:
@kevinkreiser thanks could try to help too. Would love to see this coming to rest transit in valhalla

pranavpandey1998official:
I don't think I had the band width needed to finish this now, as I am occupied with a lot of stuff also the code is 6 months old and so its better for me to just write new one with a fresh look into the problem,

few findings that I have which I would love the community to know from which they can start and my takes on the correct approach, 

first use the lib (https://github.com/mapsme/just_gtfs ) I had already merged a branch in the repo which is compatible for Valhalla (i.e. it uses boost to port c++ 17 to c++11) this will make life a lot easier so we don't have to deal with GTFS parsing

second thing we have to understand about transit land is that they also uses GTFS and all other format and came up with their own format to combine everything (like a lot of format ) into one umbrella, so because it is possible to convert GTFS to transit land it is also possible to do vise-versa

what Valhalla currently do is fetch the transit land feed , then convert that feed into a proto file which is a representation of that transit feed, and then use that to build a graph which we call transit graph

transit feed -> proto -> graph

there are two approaches that we can 
1) replace transit feed with GTFS and convert GTFS to proto
GTFS -> proto -> graph

2) to convert GTFS to graph
GTFS -> graph

now if you look at the code it seems like a no brainer to go with the first approach and that is what this PR initially tried to do

but I soon find out that taking that second approach is the correct way to move forward for future endevors, let me convince you why

1) Its easier to debug and make changes , less steps to go to the final step the easier it is to add new feature and debug, also GTFS is very vast and extensible as compared to transit feed like GTFS-RT and GTFS flexible which will be much easier to add if we take the second step

2) transit land actually utalises the OSM data to make its transit feed, which makes transit feed dependent on OSM data , GTFS is independent of OSM data , now you may think its a good thing but no, as how tiles are build in Valhalla you must know that during building of transit tiles we don't supply OSM data and because of it we must do the matching step in post tile step, which is a major issue why this PR is now merged( I am now getting into detail contact me or @nilsnolde )

3) the graph which vallhala uses to store the transit data is actually very similar to GTFS , so what happens is transit converts GTGS to tansit feed and vallhala again convert that feed to GTFSish, if we take the 1st approach we would be doing is first we convert our GTFS to transit tile and then vallhala that tranit tile to GTFSish which is pointless to do and can be very hard to do right a lot of data will be lost in converting between these steps ,

If I will get some time to do I will take the second approach and finish it, and must say I have read the codebase very extensively and I can see a lot of patterns already to think that the first approach will be the right thing to do, someone picking up this thing which is new to codebase should first read the codebase thoroughly specially the tile building part, for the veterans of the codebase, I believe you will take the best approach to solve this issue as you have much more insight  than I ever will,

I am closing this PR because I don’t  think I will ever finish this, If I get motivated enough  I will be taking the second approach and making a new PR, hope someone will find this PR useful

nilsnolde:
well thanks a ton for your work so far @pranavpandey1998official!

farfromrefug:
@pranavpandey1998official i have a question about the point of osm data dependance. I am not sure i understand what your approach implies in term of building step and dependance of gtfs data on the valhalla tiles. 
Recently i have been thinking that it would.be amazing if the valhalla transit data could be separated from the actual valhalla tiles. The reason is that the update of transit data can happen a lot more often that valhalla data. Plus you might want to add gtfs without  the need of rebuilding valhalla tiles everytimes. Wondering if your approach could allow (if actually possible) this?
To put my point even more into context i use valhalla directly embedded in a mobile app with valhalla tiles packaged in a sqlite database. I would like to allow user to download gtfs and "build" them on the fly without changing the valhalla sqlite database.

pranavpandey1998official:
I don't understand the first question, but Iet me try , so transitland data are build using the osm data , so they have an dependence on OSM data , now GTFS is completely independent of osm so , to convert GTFS to transit land data you need OSM data , 
Currently valhalla fetches the transitland data and convert that into a protobuf (lets call it X), and then that X is converted into layer 4 graph tile
now what this PR tries to do is convert GTFS to X and then in theory the code already written for converting X to layer 4 graphtile should work

what what is happening is that becasue transitland data already have dependency on OSM we cannot convert GTFS to X without OSM data,

make sense?


second vallhala transit tiles are already separated from the rest of the tiles, they are level 4 in tile hierarchy , so when you build them they are already different file, if you can change that file you can hot reload the transit graph without effecting anything, there is an issue that will make that thing possible (https://github.com/valhalla/valhalla/issues/3117)

how are you storing the tiles in sqlite database, did you build that yourself ? or there is an proven approach already present

farfromrefug:
@pranavpandey1998official thanks a lot yet it makes more sense now. 
I use valhalla within `https://github.com/CartoDB/mobile-sdk`.  They have a script to build sqlite database for valhalla tiles here https://github.com/nutiteq/mobile-sdk-scripts/blob/master/scripts/build_valhalla_packages.py. 
To integrate that into valhalla/carto they had to change Valhalla a bit https://github.com/nutiteq/valhalla/commit/7dd5ae8fe04edc45fc046b948dfd509755818ad2. I recently asked them to move to a submodule so that in the future we can hopefully have that merge in some way inside Valhalla.
I have been using that for a while and it works amazingly well.
I am hopping i could in some way build separate sqlite databases for GTFS transit data, and hopefully i can do that on the phone directly. That way i could create a something like [transportr](https://github.com/grote/Transportr) but totally offline and not dependent on navitia or other services.

kevinkreiser:
@farfromrefug thats right, i remember @mtehver working on that way back in the day, i think it got pr'd but it quickly fell out of date before we could get it merged. i still feel bad about that, many apologies @mtehver if you are ever up for it again, adding support for mbtiles to mainline valhalla id be happy to review (so you dont have to resolve conflicts on your end when we change interfaces etc).

as far as GTFS in sqlite is concerned, i dont see why you couldn't import each of the feeds files into a like-named table, the problem would of course be how to merge many feeds into one. and then finally for valhalla to work with it we would still need to convert these dbs into valhalla tiles. i had assumed, although ive not written up the proposal yet, that we would go directly from a gtfs feeds source files into our internal tile format.

PR #2839, opened 2021-02-04
Fix compression_utils.cc::inflate(...) throw - make it catchable

DmitryAk:
# Issue

In the `compression_utils.cc::inflate(...)` body:
```
try {
  if (stream.avail_in == 0)
    src_func(stream);
  if (stream.avail_in == 0)
    throw;
} catch (...) {
  inflateEnd(&stream);
  return false;
}
```
The `throw;` expression cannot be caught in further `catch (...)` block - it rethrows exception to the next level (https://en.cppreference.com/w/cpp/language/throw - "its compound-statement is considered to have been 'exited' ").
But from the `inflate(...)` declaration and definition it seems like the intention is to handle this exception and return `false`, isn't it?
If the expression `throw std::exception();` is used instead - this exception is catchable in further `catch (...)` block, it can be handled, so the `inflate(...)` returns `false` without exception rethrowing.

The added unit test expectation (test/compression.cc) fails on old version of the `compression_utils.cc::inflate(...)` (exception is thrown outside), but succeeds with current update.

## Tasklist

 - [x] Add tests

### DIFF
--- a/src/baldr/compression_utils.cc
+++ b/src/baldr/compression_utils.cc
@@ -83,7 +83,7 @@ bool inflate(const std::function<void(z_stream&)>& src_func,
       if (stream.avail_in == 0)
         src_func(stream);
       if (stream.avail_in == 0)
-        throw;
+        throw std::exception();
     } catch (...) {
       inflateEnd(&stream);
       return false;
--- a/test/compression.cc
+++ b/test/compression.cc
@@ -95,6 +95,9 @@ TEST(Compression, fail_inflate) {
     s.next_in = static_cast<Byte*>(static_cast<void*>(&bad[0]));
     s.avail_in = static_cast<unsigned int>(bad.size() * sizeof(std::string::value_type));
   };
+  auto inflate_src_fail3 = [](z_stream& s) -> void {
+    /* Nothing to do, simulates 'cannot inflate' - reproducible if no disk space. */
+  };
   auto inflate_dst_fail = [](z_stream& s) -> int {
     throw std::runtime_error("im the gingerbread man");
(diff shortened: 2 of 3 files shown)


src/baldr/compression_utils.cc
```diff
@@ -83,7 +83,7 @@ bool inflate(const std::function<void(z_stream&)>& src_func,
       if (stream.avail_in == 0)
         src_func(stream);
       if (stream.avail_in == 0)
-        throw;
+        throw std::exception();
```
mandeepsandhu, line 86: From the [docs](https://en.cppreference.com/w/cpp/language/throw), `This form is only allowed when an exception is presently being handled (it calls std::terminate if used otherwise). `

I think this fix is good :+1:

test/compression.cc
```diff
@@ -95,6 +95,9 @@ TEST(Compression, fail_inflate) {
     s.next_in = static_cast<Byte*>(static_cast<void*>(&bad[0]));
     s.avail_in = static_cast<unsigned int>(bad.size() * sizeof(std::string::value_type));
   };
+  auto inflate_src_fail3 = [](z_stream& s) {
```
mandeepsandhu, line 98: minor nit: maybe also specify the return explicitly `[](z_stream& s) -> void` (I've seen `void` type being mentioned in other lambdas).

mandeepsandhu (dismissed):
1 minor comment otherwise LGTM

test/compression.cc
DmitryAk: Yes, I'm adding it right now

Issue #2834, opened 2021-02-04
Track Penalty Requires More Iterations to Converge in Time Dependent Routes


kevinkreiser:
in #2769 we added the ability for the user to specify how much they can tollerate using highway=track while at the same time adding a penalty to those "uses" by default. this causes such search paths to sort toward the end of the queue. this means that if it is required to use such paths, the algorithm will need to exhaust more of the road network before it will evaluate such paths further.

all of the algorithms have a cut off point at which they give up in trying to find a solution (it doesnt make sense to search forever). sadly the change above makes it so that at least some routes fail more than they did before (because they dont get enough iterations). specifically this route:

    http://valhalla.github.io/demos/routing/simple.html#[{%22costing%22:%22auto%22,%22locations%22:[{%22lat%22:48.784,%22lon%22:2.199},{%22lat%22:45.931767,%22lon%22:6.118683}],%22date_time%22:{%22type%22:2,%22value%22:%222021-02-04T07:00%22}}]

is failing on master. if you increase the iterations cutoff here: https://github.com/valhalla/valhalla/blob/master/src/thor/timedep_reverse.cc#L18 the route will succeed.

i think we need to reevaluate these cutoffs based on the change in the use_tracks PR. what i would do is take the 14k routes, regex them to add timedependence and see how many failures we get now vs before the use_tracks change. id then try to find a new cut off to get the failures back to the same level as before the change. naturally this will cost CPU and therefore latency but its the only thing we can do other than reducing the overall magnitude of the use_tracks penalty (which is probably something else we should consider).

cc @genadz @kshehadeh

PR #2845, opened 2021-02-08
Increase limits for timedep_* algorithms to fix "no route found"

genadz:
# Issue

the problem was described here https://github.com/valhalla/valhalla/issues/2834 . So, I did 2 things:
1) split track penalty into 2: edge factor and transition penalty. This change allowed to decrease `track_factor` (it boosted convergence) and the same time to keep avoid tracks.
2) found new limits for timedep_* algorithms to save "no routes" rate the same as before.

## Tasklist

 - [x] Add tests

## Requirements / Relations


genadz:
on `auto.txt` routes (with time) performance decreased on ~5%.

src/sif/dynamiccost.cc
```diff
@@ -53,9 +53,12 @@ namespace sif {
 // default options/parameters
 namespace {
 
+// max penalty to apply when use tracks
+constexpr float kMaxTrackPenalty = 300.f; // 5 min
```
kevinkreiser, line 57: are you going to add a costing option to all the user to set this? a lot of our other penalties work like this like gate penalty and alley penalty

src/thor/timedep_forward.cc
```diff
@@ -14,7 +14,7 @@ namespace thor {
 constexpr uint32_t kInitialEdgeLabelCount = 500000;
 
 // Number of iterations to allow with no convergence to the destination
-constexpr uint32_t kMaxIterationsWithoutConvergence = 800000;
+constexpr uint32_t kMaxIterationsWithoutConvergence = 1800000;
```
kevinkreiser, line 17: oof more than double :(   i guess the good news is that we rarely reach this limit except for some extreme cases?

test/gurka/test_use_tracks.cc
```diff
@@ -22,18 +22,19 @@ class UseTracksTest : public ::testing::Test {
     A-1----B--------E-----2-F
            |        |       |
            |        |       |
-           |        |       3
            |        |       |
-           C--------D       G
```
kevinkreiser, line 27: can you briefly describe the test change, its not really obvious from the diff

kevinkreiser:
this looks good to me. myy only question is about adding the penality as a parameter in the request like the other ones are

src/sif/dynamiccost.cc
genadz: I've implemented the same logic as @SvetlanaBayarovich for `living_street`. reasons:
1. for external users it's more convenient to have one clear parameter `use_tracks` to represent preference of track roads;
2. right now `track_penalty` is measured in seconds that we're adding to the transition cost; if we change it in the future, we will have to notify users about that. In case of one `use_tracks` parameter we can tweak `track_factor` and `track_penalty` without affecting users

test/gurka/test_use_tracks.cc
genadz: yeah. as far as I introduced `track_penalty`, route `ABEF` became cheaper and route `ABCDEF` became more expensive (because of high penalty in transition cost). So, in order to demonstrate track avoidance feature, we should have route completely without tracks as `ABCDEF` and with tracks as `ABEF`

genadz:
Performance test. I ran `auto.txt` requests with time parameter on my local machine. Below is results I got:

set `use_tracks=0.5` (do not use penalties for tracks)
```
percentiles	5%	10%	20%	50%	90%	95%	99%	100%
response_time	0.0s	0.0s	0.0s	0.2s	1.4s	2.1s	4.1s	21.7s
```
set `use_tracks=0` (max penalties for track roads)
```
percentiles	5%	10%	20%	50%	90%	95%	99%	100%
response_time	0.0s	0.0s	0.0s	0.2s	1.4s	2.0s	3.7s	26.3s
```

as we can see, for `99%` of routes timings remained almost the same (with current branch became even less, but probably this is because of measurement error). For most slowest routes performance decreased on ~`18%`
@kevinkreiser @kshehadeh

src/thor/timedep_forward.cc
genadz: right, it happens not very often

kevinkreiser (dismissed):

src/sif/dynamiccost.cc
```diff
@@ -232,6 +235,10 @@ Cost DynamicCost::BSSCost() const {
 }
 
 void DynamicCost::set_use_tracks(float use_tracks) {
+  // Calculate penalty value based on use preference. Return value
+  // in range [kMaxTrackPenalty; 0], if use < 0.5; or
+  // 0, if use > 0.5.
+  track_penalty_ = use_tracks < 0.5f ? (kMaxTrackPenalty * (1.f - 2.f * use_tracks)) : 0.f;
```
mandeepsandhu, line 241: Is this parameter documented somewhere? I couldn't find it in the [API docs](https://github.com/valhalla/valhalla/blob/master/docs/api/turn-by-turn/api-reference.md).

Its math for generating the track penalty is not very straight-forward. Eg: setting a value of `0.4` would mean applying a penalty of 1 min (?) to each track edge. CMIIW. Makes it a bit hard to reason as to what affect the value has without looking at the code.

mandeepsandhu (approved):
Minor commetn about docs. Rest LGTM!

src/sif/dynamiccost.cc
genadz: good catch, I forgot to add `use_tracks` option to docs( will do. the reasons why we use single option I described here https://github.com/valhalla/valhalla/pull/2845#discussion_r572716957

genadz:
@kevinkreiser do you know why `build-docker-images` job failed ? I see the error `Must provide --username with --password-stdin` but don't understand why it happened, because I see that variables `DOCKERHUB_PASS, DOCKERHUB_USERNAME` exist in project settings

kevinkreiser:
@genadz its because you forked the repository and you dont have the dockerhub credentials. i'l work on making that build robust to that. you can ignore it for now

kevinkreiser (approved):

kevinkreiser:
should be fixed now

genadz:
> should be fixed now

thanks a lot!

PR #2846, opened 2021-02-09
Ignore consecutive closures

mandeepsandhu:
If search_filter->exclude_closures is false, the router now ignores
multiple consecutive closures start from or leading into the start
or end location respectively.


src/thor/timedep_reverse.cc
```diff
@@ -276,7 +276,8 @@ inline bool TimeDepReverse::ExpandReverseInner(GraphReader& graphreader,
   uint32_t idx = edgelabels_rev_.size();
   edgelabels_rev_.emplace_back(pred_idx, meta.edge_id, oppedge, meta.edge, newcost, sortcost, dist,
                                mode_, transition_cost,
-                               (pred.not_thru_pruning() || !meta.edge->not_thru()), restriction_idx);
+                               (pred.not_thru_pruning() || !meta.edge->not_thru()), true,
```
genadz, line 279: don't we want to ignore consecutive closures for single directional search algorithms ? why the logic is different from bidirastar ?

mandeepsandhu: Not yet @genadz . Ignore closures does not yet work for unidirectional astar algorithms. The main issue is that we have to selectively ignore closures only if they lead up to the destination edge (but we won't know that while expanding in 1 directions, unlike bidir astar where the expansion starts from both ends).

There's work to be done to make sure that we ignore a closure only if it leads upto the destination.

kevinkreiser: honestly im hoping to eventually just remove the unidirectional astar algorithms and move the time dependent stuff to bidirectional a*.

src/thor/bidirectional_astar.cc
```diff
@@ -306,6 +306,7 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
   edgelabels_forward_.emplace_back(pred_idx, meta.edge_id, opp_edge_id, meta.edge, newcost, sortcost,
                                    dist, mode_, transition_cost,
                                    (pred.not_thru_pruning() || !meta.edge->not_thru()),
+                                   (pred.closure_pruning() || !costing_->IsClosed(meta.edge, tile)),
```
kevinkreiser, line 309: its unfortunate to have to check this twice. basically we already called IsClosed above potentially inside of the allowed function and now we call it again here. it will be interesting to see the performance impact of this in a system that is actually hooked up with live traffic (ie when calling this is non trivial)

valhalla/sif/edgelabel.h
```diff
@@ -399,7 +410,8 @@ class EdgeLabel {
   uint64_t not_thru_ : 1;
   uint64_t deadend_ : 1;
   uint64_t on_complex_rest_ : 1;
-  uint64_t spare_0 : 2;
+  uint64_t closure_pruning_ : 1;
```
kevinkreiser, line 413: since we dont do this for the unidirectional algorithms, why not add this only to the BDEdgelabel like not_thru_pruning is. you can take some bits from the oppedgeid it doesnt need 63bits, the max id is only 21+22+3 bits wide.

kevinkreiser:
other than the comment about the bit being on the BDEdgeLabel the rest of this PR looks as i would have expected it. would you mind moving the bit to that child class instead?

src/thor/timedep_reverse.cc
genadz: @mandeepsandhu got it! thanks

genadz (approved):

src/thor/dijkstras.cc
```diff
@@ -199,7 +199,8 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
     uint32_t idx = bdedgelabels_.size();
     *es = {EdgeSet::kTemporary, idx};
     bdedgelabels_.emplace_back(pred_idx, edgeid, oppedgeid, directededge, newcost, mode_,
-                               transition_cost, path_dist, false, restriction_idx, pred.path_id());
+                               transition_cost, path_dist, false, true, restriction_idx,
```
kevinkreiser, line 202: the changes here dont look right. you shouldnt hardcode this to `true`. you'll also want to make the changes here to look like the ones you did in bidirectional a*. if you dont do this the reachability computation will be technically wrong. speaking of which we need to think about the cheap reachability check and whether or not it should bail early and fall back ot the full check via dijkstras

kevinkreiser (changes requested):

src/thor/bidirectional_astar.cc
mandeepsandhu: I have a static dump of traffic that I'm running perf tests with. Will post back the results here shortly.

valhalla/sif/edgelabel.h
mandeepsandhu: Yeah I started off by putting it in the `BDEdgelLabel` class. But since `DynamicCost::Allowed()` accepts an `EdgeLabel` (its also used in unidirectional astar variants which don't use `BDEdgeLabel`), I kept it in the base class.

src/thor/dijkstras.cc
mandeepsandhu: I hard-coded it here (and other unidirectional variants) since I dont want to change the existing behavior for them. This flag switched on closure pruning which is what we do currently, i.e we don't consider closed edges. Or maybe there's a gap in my understanding about how reachability works. Let me know.

mandeepsandhu:
Performance test results (avg of 3 runs of [`de_benchmark_routes.txt`](https://github.com/valhalla/valhalla/blob/master/test_requests/de_benchmark_routes.txt) between this branch and `master`):
 percentile | 50th | 90th | 99th
---------------|-------|--------|-------
`master` | 0.10986 | 0.19896 | 0.33849
this branch | 0.10969 (**-0.15%**) | 0.19967 (**+0.35%**) | 0.34198 (**+1.02%**)

CC @kevinkreiser

valhalla/sif/edgelabel.h
kevinkreiser: yeah i was thinking that since we dont care about it in unidirectional we didnt need it there but you're right since the allowed method has to take the base class, its easier to access it there. the other option is to just put the function for getting teh value from the label in the base class but actually store the boolean in the BDEdgeLabel.

then the base class always returns `true` but the BDEdgeLabel returns the actual value of the boolean/bit by overriding the getter function

src/thor/dijkstras.cc
kevinkreiser: yeah so there are two things here. we have a quck reach check, where we want allowed to return false right away when an edge is closed to set the reachability low. when this happens, ie the quick reach check fails, the edge candidate gets kicked out to dijkastas to compute the actual reachability. but to do that we need to actually properly handle whether the predecessor was or wasnt closed and set pruning appropriately. the way this is here any closed edge will have low reachability and force us to search for another candidate edge. which was the case before, but when the user says hey i want to route through consecutive closures at the beginning/end of the route we need dijkstras here to actually do that.

valhalla/sif/edgelabel.h
mandeepsandhu: Added a `TODO` to move the `not_thru_pruning` flag to the base class.

test/gurka/test_reach.cc
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  // Check reach for AB ignoring closures
+  {
+    co.set_filter_closures(false);
+    auto costing = factory.Create(co);
+    auto edge = gurka::findEdgeByNodes(*reader, closure_map.nodes, "A", "B");
+    auto reach = reach_checker(std::get<1>(edge), std::get<0>(edge), 50, *reader, costing);
+
+    EXPECT_EQ(reach.inbound, 8);
```
mandeepsandhu, line 192: The in/outbound reach should be `9` (the same as with no closures). However what I'm observing is that edges leaving `A` are not looked at when expanding in forward & reverse directions. I have not yet figured why the expansion is omitting these edges. Let me know if you an idea as to why it would be doing that.

test/gurka/test_reach.cc
```diff
@@ -0,0 +1,233 @@
...
+    auto costing = factory.Create(co);
+    auto edge = gurka::findEdgeByNodes(*reader, closure_map.nodes, "H", "I");
+    auto reach = reach_checker(std::get<1>(edge), std::get<0>(edge), 50, *reader, costing);
+
+    EXPECT_EQ(reach.inbound, 7);
+    EXPECT_EQ(reach.outbound, 7);
+  }
+}
+
+TEST_F(TestReach, DISABLED_ReachWithClosures2) {
```
mandeepsandhu, line 207: Currently disabled. This tests the scenario where we're currently under-counting the inbound reach

test/gurka/test_reach.cc
kevinkreiser: if u-turns are disallowed, i could see how that would happen. consider the forward direction of the reach check. it would start in the closure get out of it but not be able to get back in. in this case it wouldnt be able to reach the A node. but that only works if its not able to make a uturn at B, stay in the closure and reach A. to visualize this we could modify the reach algorithm to print out the node lat lons as they are visited just to be sure. or we could just do a one time debug session to watch what nodes are being seen when.

mandeepsandhu: Okay, so this is what I figured is going on: When we're expanding forward (similar thing happens for reverse expansion too), at some some point we go from consecutive closed edge(s) to an open edge, eg AB -> BE , which enables closure pruning. Now when we expand further from BE, one path goes to DE (another open edge), and then DE tries expanding to AD which is closed. Since pruning is on we never expand from DE into AD which explains why the reach is off by 1. CMIIW.

src/sif/recost.cc
```diff
@@ -125,7 +125,7 @@ void recost_forward(baldr::GraphReader& reader,
     length += edge->length() * edge_pct;
     // construct the label
     label = EdgeLabel(predecessor++, edge_id, edge, cost, cost.cost, 0, costing.travel_mode(), length,
-                      transition_cost, time_restrictions_TODO);
+                      transition_cost, time_restrictions_TODO, true);
```
kevinkreiser, line 128: i think this would be more correct as `false` so that you dont recost cost a multiple edge closer and fail the allowed check because pruning is turned on. maybe it might be better instead of hardcoding here to actually use `!ignore_access`

src/thor/costmatrix.cc
```diff
@@ -328,7 +328,7 @@ void CostMatrix::ForwardSearch(const uint32_t index, const uint32_t n, GraphRead
       *es = {EdgeSet::kTemporary, idx};
       edgelabels.emplace_back(pred_idx, edgeid, oppedge, directededge, newcost, mode_, tc,
                               pred.path_distance() + directededge->length(),
-                              (pred.not_thru_pruning() || !directededge->not_thru()),
+                              (pred.not_thru_pruning() || !directededge->not_thru()), true,
```
kevinkreiser, line 331: why hardcode this? i think this algorithm should work the same as bidirectional_astar, indeed you can see the not_thru stuff which is an analog in the other arguments to the edgelabel constructor

src/thor/dijkstras.cc
```diff
@@ -199,7 +199,9 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
     uint32_t idx = bdedgelabels_.size();
     *es = {EdgeSet::kTemporary, idx};
     bdedgelabels_.emplace_back(pred_idx, edgeid, oppedgeid, directededge, newcost, mode_,
-                               transition_cost, path_dist, false, restriction_idx, pred.path_id());
+                               transition_cost, path_dist, false,
+                               (pred.closure_pruning() || !costing_->IsClosed(directededge, tile)),
```
kevinkreiser, line 203: this coverage check cannot be right. how frustrating

src/thor/timedep_reverse.cc
kevinkreiser: yeah the best we could do here is for forward it would work only at the origin and for reverse it would work only at the destination. i suppose we should implement that just to remove a gotcha from the code?

valhalla/sif/dynamiccost.h
```diff
@@ -696,6 +696,15 @@ class DynamicCost {
 
   virtual Cost BSSCost() const;
 
+  /*
+   * Determine whether an edge is currently closed due to traffic.
+   * @param  edgeid         GraphId of the opposing edge.
+   * @return  Returns true if the edge is closed due to live traffic constraints, false if not.
+   */
+  inline virtual bool IsClosed(const baldr::DirectedEdge* edge, const graph_tile_ptr& tile) const {
```
kevinkreiser, line 704: hahah i just hate this. expanding the public part of the interface but there is no cleaner way unfortunately :frowning_face:

src/sif/recost.cc
mandeepsandhu: Yeah I didn't fully grasp what recost costing was used for. So wherever it wasn't clear to me what the costing was doing, I defaulted to `true` to keep it same as before.

Can using `!ignore_access` cause issues because it will also ignore intermediate closures?

src/thor/costmatrix.cc
mandeepsandhu: I'll fix this (oversight on my end).

src/thor/costmatrix.cc
```diff
@@ -703,7 +703,7 @@ void CostMatrix::SetSources(GraphReader& graphreader,
       // Set the initial not_thru flag to false. There is an issue with not_thru
       // flags on small loops. Set this to false here to override this for now.
       BDEdgeLabel edge_label(kInvalidLabel, edgeid, oppedge, directededge, cost, mode_, ec, d, false,
-                             -1);
+                             true, -1);
```
mandeepsandhu, line 706: This is the pruning state when setting the source and destination locations. There's a comment here which I don't quite understand the reason for:
```
// Set the initial not_thru flag to false. There is an issue with not_thru
```
I didn't find any tests that highlight the issue alluded to here.

Should we follow the same for closure pruning, i.e start in disabled state (`false`)?

src/thor/timedep_reverse.cc
mandeepsandhu: @kevinkreiser do you know why `not_thru` pruning is implemented for time-dependent reverse but not forward? I feel this asymmetry is causing more confusion and I don't want to add the closure pruning without fully understanding why it wasn't done for the time-dep fwd algo. Was it simply an oversight?

src/sif/recost.cc
kevinkreiser: TLDR; basically what it says is "be permissive" when recomputing the cost of the route. 

recosting is used to compute cost/time of a path using a different costing. it does call the "allowed" function using the labels it generates while recomputing the cost of a path with a different costing. this means that if you turn on pruning the allowed function will return `false` if the original path was going through a multiple edge closure. while i think it is more correct to set pruning to `false` when someone sets `ignore_access` it turns out that it wont matter because the `if` above checks if that boolean was set. so with the change you just made you'll have made it the case that `allowed` will return `true` and not try to fail the route at a closure. honestly it might be pretty silly at all to even check the allowed method in recosting. i did that in the original implementation but the more times we revisit it the less i like that it exists :smile: 

in the future though it will be important because its the way that we get information about any kinds of restrictions the path might go through which we will want to bubble up to the user.

src/thor/costmatrix.cc
kevinkreiser: this is essentially the same as SetOrigin in the bidirectional algorithm we should do the same thing here.  a copy paste wil suffice and the same in the destination (targets) function :smile:

src/thor/timedep_reverse.cc
```diff
@@ -511,7 +513,8 @@ void TimeDepReverse::SetOrigin(GraphReader& graphreader,
     // DO NOT SET EdgeStatus - it messes up trivial paths with oneways
     uint32_t idx = edgelabels_rev_.size();
     edgelabels_rev_.emplace_back(kInvalidLabel, opp_edge_id, edgeid, opp_dir_edge, cost, sortcost,
-                                 dist, mode_, c, false, -1);
+                                 dist, mode_, c, false, !(costing_->IsClosed(opp_dir_edge, tile)),
```
kevinkreiser, line 516: im pretty sure this should be `directededge` from above. its the state of the destination edge the route will actually end on that we need to use. `opp_dir_edge` is the opposing edge which we use to carry out the backwards search.

src/thor/timedistancematrix.cc
```diff
@@ -121,7 +121,7 @@ void TimeDistanceMatrix::ExpandForward(GraphReader& graphreader,
     // Add to the adjacency list and edge labels.
```
kevinkreiser, line 121: this algorithm is the matrix equivalent of timedep_forward/rev for completeness we should copy paste the same changes we did elsewhere to here. we can add matrix unit testing to confirm it later on.

test/gurka/gurka.cc
```diff
@@ -464,7 +464,7 @@ map buildtiles(const nodelayout& layout,
   detail::build_pbf(result.nodes, ways, nodes, relations, pbf_filename);
   std::cerr << "[          ] building tiles in " << result.config.get<std::string>("mjolnir.tile_dir")
             << std::endl;
-  midgard::logging::Configure({{"type", ""}});
+  midgard::logging::Configure({{"type", "std_out"}});
```
kevinkreiser, line 467: you might have accidently left this on for testing? honestly i kind of like this change now that the logging is vastly reduced for route requests, however the logging for tile creation is still very verbose (as it should be)

kevinkreiser (changes requested):
Just a couple minor things:

1. the one bug in the timedep reverse setorigin function using the wrong edge (im guessing your test is closing both directions of the edge and so you didnt notice it)
2. and some copy paste missing from time distance matrix

test/gurka/gurka.cc
mandeepsandhu, line 467: Oops, good catch! yeah I had switched it on for debugging.

src/thor/timedep_reverse.cc
mandeepsandhu: Ah, yes. the directed edge is create just a few lines above. I didn't see it when doing copy-pasta!

kevinkreiser (approved):
thanks again for enduring the long and arduous journey!

mandeepsandhu:
> thanks again for enduring the long and arduous journey!

Thanks for the ride-along! :slightly_smiling_face:

gknisely (approved):

kevinkreiser (approved):

PR #2847, opened 2021-02-09
Binary Artifacts Docker Image

kevinkreiser:
Many others around the internet have made some nice docker images of valhalla for various purposes (heres one from @nilsnolde https://github.com/gis-ops/docker-valhalla/blob/master/Dockerfile). I'd like to make their lives a bit easier by making a generic image that has all of the binaries installed but doesnt actually run anything. Then all they have to do is grab this image to start with and figure out what exactly they want to run whether its data making or a server or CI or analytical stuff.

To that end a made a basic dockerfile here that accomplishes that. I tried to make the image reasonably small but still include debug information in case someone needs to triage a crash. Its a bit cumbersome as you have to gunzip the debug symbols but lets hope its few and far between. The final size of the image is 1.2gb. I'm sure more could be trimmed but we can get it in a follow up if anyone has ideas.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,48 +13,28 @@ commands:
       - run: brew install protobuf cmake ccache libtool boost-python libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
 
 jobs:
-  build-base:
+  build-docker-images:
     docker:
       - image: docker:18.06.3-ce-git
     steps:
       - setup_remote_docker
       - checkout
       - run:
-          name: Build Valhalla base Dockerfile
+          name: Build Valhalla Docker Images
           command: |
             docker build -f ./docker/Dockerfile-build --tag valhalla/valhalla:build-latest ./docker
-      - run:
-          name: Push Valhalla base image
-          command: |
-            if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
-              docker push valhalla/valhalla:build-latest
-            elif [[ ! -z "${CIRCLE_TAG}" ]]; then
-              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
-              docker push valhalla/valhalla:build-${CIRCLE_TAG}
-            fi
-
-  build-binary:
-    requires: build-base
-    docker:
-      - image: docker:18.06.3-ce-git
-    steps:
-      - setup_remote_docker
-      - checkout
-      - run:
-          name: Build Valhalla binary artifacts Dockerfile
-          command: |
             if [[ "${CIRCLE_BRANCH}" == "master" ]] || [[ ! -z "${CIRCLE_TAG}" ]]; then
               docker build -f ./docker/Dockerfile-run --tag valhalla/valhalla:run-latest .
             fi
       - run:
-          name: Push Valhalla binary artifacts image
+          name: Push Valhalla Docker Images
           command: |
+            echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+              docker push valhalla/valhalla:build-latest
               docker push valhalla/valhalla:run-latest
             elif [[ ! -z "${CIRCLE_TAG}" ]]; then
-              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+              docker push valhalla/valhalla:build-${CIRCLE_TAG}
               docker push valhalla/valhalla:run-${CIRCLE_TAG}
             fi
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -29,6 +29,33 @@ jobs:
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
               echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
               docker push valhalla/valhalla:build-latest
+            elif [[ ! -z "${CIRCLE_TAG}" ]]; then
```
kevinkreiser, line 32: circle changes are a work in progress. probably need to spend a few rounds on this yet

.circleci/config.yml
```diff
@@ -29,6 +29,33 @@ jobs:
...
+              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+              docker push valhalla/valhalla:build-${CIRCLE_TAG}
+            fi
+
+  build-binary:
+    requires: build-base
```
kevinkreiser, line 38: i want this one to depend on the build job but it seems circle doesnt like this syntax (maybe its from an older version)

nilsnolde (changes requested):
Good effort! And I'd like to not build it all from scratch! Few ideas to bring down final image size (probably in decreasing effect order):
- in the final image, do just one `RUN` statement with all cmd concatenated, that probably already reduces it by half (reason is each RUN creates one layer, literally a _system_ diff from the previous layer)
- I think the install of `software-properties-common` is surprisingly huge as well. For the final image's sake you could build & install prime_server to /usr/local in the builder phase?
- try `python3.8-minimal` instead of python, good enough as base I think. Or even not at all and leave it to downstream images to decide if they want it. Same goes for gdb/spatialite-bin/curl/wget (whichever is not needed by the various scripts)

Just as reference: our final image is currently 470 MB on disk, 180 MB compressed for pulling. And tbh, I wouldn't like the size to increase even further when basing on this one. Not a fair comparison, but OSRM is a tiny 100/38 MB

kevinkreiser:
@nilsnolde i made the changes you suggested except for removing the various tools that the scripts rely on. this reduced the size by about 100mb. at the end of the day the difference in size between your image and mine i think is explainable by me keeping the debug info around for triaging crashes (this is about 475mb) plus the few extra programs that i have installed.

ihmo for the sake of people reporting problems or triaging what the heck is wrong, having this extra info is useful if you want it. if you dont want it you should be able to make a minimal dockerfile that just removes that debug file and `apt purge` the packages you dont want, right? the idea was that you would need to make your  own image anyway so you could do whatever final `CMD` you need to for your use case. am i missing something here that would make that general idea invalid?

.circleci/config.yml
```diff
@@ -13,22 +13,29 @@ commands:
...
       - image: docker:18.06.3-ce-git
     steps:
       - setup_remote_docker
       - checkout
       - run:
-          name: Build Valhalla base Dockerfile
+          name: Build Valhalla Docker Images
           command: |
             docker build -f ./docker/Dockerfile-build --tag valhalla/valhalla:build-latest ./docker
+            if [[ "${CIRCLE_BRANCH}" == "master" ]] || [[ ! -z "${CIRCLE_TAG}" ]]; then
```
kevinkreiser, line 26: if we are merging master or tagging then we also building the binary artifacts image

.circleci/config.yml
```diff
@@ -13,22 +13,29 @@ commands:
...
       - run:
-          name: Push Valhalla base image
+          name: Push Valhalla Docker Images
           command: |
+            echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
               docker push valhalla/valhalla:build-latest
+              docker push valhalla/valhalla:run-latest
+            elif [[ ! -z "${CIRCLE_TAG}" ]]; then
```
kevinkreiser, line 36: i figured tagging was a good thing to do too. we do this rarely but when it fails we'll get an email and can fix it manually

docker/Dockerfile-run
```diff
@@ -0,0 +1,51 @@
+####################################################################
```
kevinkreiser, line 1: i tried to comment this pretty well just have a read over it.  the main idea is we use the multistage build to save space. i keep the debug info around but compressed in a targz. this brings the image down in size by about 1.5gb to a total of 1.1gb. downstream consumers of this image can always remove this extra debug info to make their image even slimmer.

also note that there is no `CMD` in the final image. the idea here is that this is the base for people to use to do whatever use-case they want to use valhalla for (building tiles, running a service, doing analytics, who knows).

nilsnolde:
as usual you're right I guess;) was hoping for more reduction by combining the RUN statements but fair enough. Will try it soon as base image for ours to see where we land with this.

nilsnolde (approved):

kevinkreiser:
@nilsnolde one other thing i forgot i also did `locale-gen` for all the locales we support which adds another 40MB to the image. i believe people can use `localepurge` to remove these too if they are not wanted

kevinkreiser:
@nilsnolde @PureW all the builds are done, i renamed the docker images one so that one will never finish until this is merged and the name change is complete.

kevinkreiser:
going to merge and get the ci workflow name change updated on master

PR #2848, opened 2021-02-09
Implement depth-first-search to match way_ids with edge_ids

merkispavel:
# Issue

This implementation handles the case when you might choose wrong direction more than once while matching way_ids with graph_ids. See the [gurka](https://github.com/valhalla/valhalla/pull/2848/commits/5505587bf9bfdd9a7a5a640099b3907b52f0b8e5#r572726931) test for better understanding 

## Tasklist

 - [x] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Update the docs with any new request parameters or changes to behavior described
 - [x] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #2849, opened 2021-02-09
Honor access mode while matching OSMRestriction with graph

merkispavel:
# Issue

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/test/gurka/test_route.cc
+++ b/test/gurka/test_route.cc
@@ -679,3 +679,43 @@ TEST(Standalone, OneWayIdToManyEdgeIds) {
     EXPECT_STREQ(e.what(), "No path could be found for input");
   }
 }
+
+TEST(Standalone, HonorAccessPropertyWhenConstructingRestriction) {
+  const std::string ascii_map = R"(
+        A<-----B<----D
+               ^
+               |
+        E----->F---->G
+)";
+
+  const gurka::ways ways = {
+      {"EF", {{"highway", "secondary"}, {"oneway", "yes"}, {"osm_id", "1"}}},
+      {"FG", {{"highway", "secondary"}, {"oneway", "yes"}, {"osm_id", "1"}}},
+      {"FB", {{"highway", "primary"}, {"oneway", "yes"}, {"osm_id", "2"}}},
+      {"DB", {{"highway", "primary"}, {"oneway", "yes"}, {"osm_id", "3"}}},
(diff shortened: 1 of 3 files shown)


test/gurka/test_route.cc
```diff
@@ -679,3 +679,42 @@ TEST(Standalone, OneWayIdToManyEdgeIds) {
...
+      {"DB", {{"highway", "primary"}, {"oneway", "yes"}, {"osm_id", "3"}}},
+      {"BA", {{"highway", "primary"}, {"oneway", "yes"}, {"osm_id", "3"}}},
+  };
+
+  const gurka::relations relations = {
+      {{
+           {gurka::way_member, "EF", "from"},
+           {gurka::way_member, "FB", "via"},
+           {gurka::way_member, "BA", "to"},
+       },
```
merkispavel, line 703: `EF` and `FG` edges have the same `way_id`. So `GetGraphIds` return `GFBA` instead of `EFBA`. We must look at the access in my opinion. Let me know if the suggested solution is correct or not: it's the first that came to my mind

merkispavel:
I compared the number of restrictions between master and this branch: the PR "removes" some of the restrictions(5-10 restrictions of ~2700). So need more time to investigate it

gknisely:
see comment [here](https://github.com/valhalla/valhalla/pull/2848#issuecomment-776419124)

merkispavel:
Checked master vs pr diff in restrictions in the Florida. Looks good

List of restrictions that are removed due to access flags:
https://www.openstreetmap.org/relation/10618528
https://www.openstreetmap.org/relation/10618529
https://www.openstreetmap.org/relation/9768860
https://www.openstreetmap.org/relation/9063828
https://www.openstreetmap.org/relation/10368658
https://www.openstreetmap.org/relation/10368657
https://www.openstreetmap.org/relation/9949592
https://www.openstreetmap.org/relation/7509315
In most cases there's at least one oneway way(heh) among the restriction

The last two restrictions are actually buggy: the via way is incorrect
https://www.openstreetmap.org/relation/9949592
https://www.openstreetmap.org/relation/7509315

gknisely (dismissed):

merkispavel:
Example of the fixed case: osm origin restriction, valhalla before, valhalla after

Screen Shot 2021-02-15 at 21 44 57

gknisely (approved):

PR #2856, opened 2021-02-10
Fix costing enum for avoids in loki

### LINKED ISSUE #2853: Avoid locations and truck profile
avoid_locations option works fine with auto profile. But with truck profile it is completely ignored. Maybe that is by design, and should be mentioned in documentation.
P.S. Did not test with other profiles.

nilsnolde:
fixes #2853 

"by accident" car and bike worked out, all the other profiles not so much..


danpat (approved):

danpat:
🤦

src/loki/worker.cc
```diff
@@ -94,7 +94,7 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
       auto avoid_locations = PathLocation::fromPBF(options.avoid_locations());
       auto results = loki::Search(avoid_locations, *reader, costing);
       std::unordered_set<uint64_t> avoids;
-      auto* co = options.mutable_costing_options(static_cast<uint8_t>(costing->travel_mode()));
```
kevinkreiser, line 97: yeah this is easy to confuse with mode costing which has a costing object per mode of travel (for multimodal). tbh i started working on a refactor of all of this stuff to make it less confusing but i never got it finished. we kind of made a mistake in the original api design that forced us into multimodal making everything miserable for us

PR #2857, opened 2021-02-10
Conditional access restriction update

gknisely:
# Issue
added support for destination for conditional access restrictions

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/test/gurka/test_conditional_restrictions.cc
+++ b/test/gurka/test_conditional_restrictions.cc
@@ -30,37 +30,45 @@ protected:
 
     const std::string timeDenied =
         "no @ (Mar 00:00-07:00;Mar 17:30-24:00;Apr 00:00-07:00;Apr 19:00-24:00;Aug 00:00-07:00;Aug 19:00-24:00)";
-    const std::string timeAllowed =
+    const std::string yesAllowed =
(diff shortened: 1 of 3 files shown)


test/gurka/test_conditional_restrictions.cc
```diff
@@ -30,37 +30,45 @@ class ConditionalRestrictions : public ::testing::Test {
...
+        "delivery @ (Mar 00:00-07:00;Mar 17:30-24:00;Apr 00:00-07:00;Apr 19:00-24:00;Aug 00:00-07:00;Aug 19:00-24:00)";
     const gurka::ways ways = {
         {"AD",
          {{"highway", "service"},
           {"motorcar", "no"},
           {"bicycle", "yes"},
           {"foot", "yes"},
-          {"bicycle:conditional", timeDenied},
-          {"foot:conditional", timeAllowed}}},
+          {"bicycle:conditional", yesAllowed},
```
mandeepsandhu, line 49: The restriction changed from "denied" to "allowed" but none of the tests below needed to be changed? So for the time it was denied would need to have its test changed, no?

gknisely: @mandeepsandhu before we only tested for the `yes` attribute.  So I added the others in randomly for the tests regardless of the mode as it hits the same code.  If we don't encounter one of those values we default to denied.

mandeepsandhu: Got it. thx

mandeepsandhu (approved):

PR #2859, opened 2021-02-11
nit: Fixes compiler errors

PureW:
Renames the Isochrone/Centroid `Expand` function which overloads `Dijkstra::Expand` while changing the type signature:

```
../valhalla/thor/isochrone.h:58:50: error: 'valhalla::thor::Isochrone::Expand' hides overloaded virtual function [-Werror,-Woverloaded-virtual]
  std::shared_ptr<const midgard::GriddedData<2>> Expand(const ExpansionType& expansion_type,
                                                 ^
../valhalla/thor/dijkstras.h:68:16: note: hidden overloaded virtual function 'valhalla::thor::Dijkstras::Expand' declared here: type mismatch at 1st parameter ('valhalla::thor::ExpansionType' vs 'const valhalla::thor::ExpansionType &')
  virtual void Expand(ExpansionType expansion_type,
               ^
In file included from ../src/thor/worker.cc:13:
In file included from ../valhalla/thor/worker.h:24:
```

### DIFF
--- a/src/thor/centroid.cc
+++ b/src/thor/centroid.cc
@@ -93,12 +93,12 @@ bool PathIntersection::operator==(const PathIntersection& i) const {
 }
 
 // main entry point to the functionality
-std::vector<std::vector<PathInfo>> Centroid::Expand(const ExpansionType& expansion_type,
-                                                    valhalla::Api& api,
-                                                    baldr::GraphReader& reader,
-                                                    const sif::mode_costing_t& costings,
-                                                    const sif::TravelMode mode,
-                                                    valhalla::Location& centroid) {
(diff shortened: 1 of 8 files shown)


kevinkreiser:
Instead of all of this could we just remove the `virtual` keyword from the function in the dijsktras class? seemed like we had hoped the signatures would be the same but in practice it wasnt the case

kevinkreiser:
i just went ahead and did that. we dont make use of polymorphism in terms of an internface to a caller for this class anyway we are really just using inheritance for the sake of sharing base functionality

kevinkreiser (approved):

PureW:
> i just went ahead and did that. we dont make use of polymorphism in terms of an internface to a caller for this class anyway we are really just using inheritance for the sake of sharing base functionality

:+1: sounds ok to me.

PR #2861, opened 2021-02-11
Remove stop impact when flow info is used on an edge

dgearhart:
# Issue

Flow information already incorporates the inference of stop impact therefore, we do not need to include stop impact when flow info is used.

## Tasklist

 - [ ] Add tests


src/sif/autocost.cc
```diff
@@ -484,12 +484,15 @@ Cost AutoCost::TransitionCost(const baldr::DirectedEdge* edge,
...
   c.secs = OSRMCarTurnDuration(edge, node, pred.opp_local_idx());
 
-  // Intersection transition time = factor * stopimpact * turncost. Factor depends
-  // on density and whether traffic is available
+  // Transition time = turncost * stopimpact * densityfactor
   if (edge->stopimpact(idx) > 0 && !shortest_) {
     float turn_cost;
     if (edge->edge_to_right(idx) && edge->edge_to_left(idx)) {
-      turn_cost = kTCCrossing;
+      if (!edge->has_flow_speed() || flow_mask_ == 0) {
```
kevinkreiser, line 491: i think there are a couple of pretty big pitfalls here.

1. having flow doesnt mean you have constrained so you might miss that (i think the add historical traffic doesnt let you add one without the other but in the future it might decouple those)
2. if you had a live speed but didnt have historical you wont know it here

i suggest that we take a bit from the predecessor label and if when we added that it had any of the traffic sources set we set that bit to true. then we can change all these ifs in here to be `if(pred.measured_speed())`

@dgearhart let me know if you need more info, but basically when we call EdgeCost (before we add the label) we can get back a mask that says what speed sources were used to compute teh cost. we can then use that to set the boolean on the label and hten use that boolean here

dgearhart:
replaced with https://github.com/valhalla/valhalla/pull/2891

PR #2866, opened 2021-02-15
nit: Fixes compiler warning/error in incident-loading

PureW:


src/baldr/incident_singleton.h
```diff
@@ -265,7 +265,7 @@ struct incident_singleton_t {
           valhalla::baldr::GraphId tile_id(((uint64_t(1) << 25) - 1) & entry);
           seen.insert(tile_id);
           // spare last 39 bits are the timestamp, leaves us with something like 17k years
-          uint64_t timestamp = (entry >> 25) & ((uint64_t(1) << 39) - 1);
+          int64_t timestamp = (entry >> 25) & ((uint64_t(1) << 39) - 1);
```
PureW, line 268: Resolve comparison against signed `last_scan` down below.

ktatterso (approved):

src/baldr/incident_singleton.h
kevinkreiser: why not `time_t` to match exactly the thing its being compared against? on some systems `time_t` might not be `int64_t` right?

PR #2026, opened 2019-11-05
Many Refactorings, Bug Fixes and Quality of Life Improvements

kevinkreiser:
Refactor costing to better handle multiple speed datasources
Better usability of curl for fetching tiles on the fly
LRU cache scheme for tile storage
GraphTile size check
Pick more sane values for highway and toll avoidance
Refactor adding predicted speed info to speed up process
Allow selecting speed data sources at request time
Allow disabling certain neighbors in connectivity map

Co-authored-by: Maksim Skurydzin <maksim.skurydzin@mapbox.com>
Co-authored-by: Daniel Paz-Soldan <daniel.pazsoldan@gmail.com>
Co-authored-by: Pavel Merkis <pavel.merkis@mapbox.com>
Co-authored-by: vmihaylenko <vxmihaylenko@gmail.com>

PR #2869, opened 2021-02-16
revert highway factor typo

kevinkreiser:
Once upon a time, i made a horrible horrible mistake:
https://github.com/valhalla/valhalla/pull/2026/files#diff-3b95c3b6cad47a8b18380f94765b22cb96a80fdbc8e2def2ed090756ac2cdca8R87

Which causes us to favor highways above anything else... It was so long ago that even though I've noticed it several times since then I had assumed it was always that way. :facepalm: 

Thanks to @dgearhart and @gknisely for noticing it and asking me about it!


gknisely (dismissed):
we all make mistakes :-)

gknisely (approved):

PR #2870, opened 2021-02-17
Avoid service roads

SvetlanaBayarovich:
# Issue

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/sif/autocost.cc
```diff
@@ -483,6 +493,8 @@ Cost AutoCost::TransitionCost(const baldr::DirectedEdge* edge,
   uint32_t idx = pred.opp_local_idx();
   Cost c = base_transition_cost(node, edge, &pred, idx);
   c.secs = OSRMCarTurnDuration(edge, node, pred.opp_local_idx());
+  c.cost += service_penalty_ *
```
SvetlanaBayarovich, line 496: The reason it is not in the `base_transition_cost` is because bicycle costing already uses such penalty, but in combination with some other multipliers. Also `truck` profile has `low_class_penalty_` which affects transition to residential and service roads

src/sif/autocost.cc
```diff
@@ -354,6 +358,9 @@ AutoCost::AutoCost(const CostingOptions& costing_options, uint32_t access_mask)
   distance_factor_ = costing_options.use_distance() * kInvMedianSpeed;
   inv_distance_factor_ = 1.f - costing_options.use_distance();
 
+  service_penalty_ = costing_options.service_penalty();
+  service_factor_ = 2;
```
SvetlanaBayarovich, line 362: Still thinking if it's worth using factor for the task

src/sif/autocost.cc
dnesbitt61: I think it would be fine to add to base transition cost and remove from bicyclecost. I added the penalty recently and I don't think it will impact much moving it into transition cost. Not sure about the truck profile - others @gknisely ? are better able to comment on that.

SvetlanaBayarovich:
A couple of cases to show the effect of changes. 1 - 3 were cutting corners through parking lots, 4th used bus lane
|master|with avoidance|satellite|
|---|---|---|
|image|image|image|

SvetlanaBayarovich:
One more case that was routed through the service road before
|master|with avoidance|satellite|
|---|---|---|
|image|image|image|

docs/api/turn-by-turn/api-reference.md
```diff
@@ -105,6 +105,8 @@ These options are available for `auto`, `bus`, and `truck` costing methods.
...
 | `use_living_streets` | This value indicates the willingness to take living streets. This is a range of values between 0 and 1. Values near 0 attempt to avoid living streets and values near 1 will favor living streets. The default value is 0 for trucks, 0.1 for cars, buses, motor scooters and motorcycles. Note that sometimes living streets are required to complete a route so values of 0 are not guaranteed to avoid living streets entirely. |
 | `use_tracks` | This value indicates the willingness to take track roads. This is a range of values between 0 and 1. Values near 0 attempt to avoid tracks and values near 1 will favor tracks a little bit. The default value is 0 for autos, 0.5 for motor scooters and motorcycles. Note that sometimes tracks are required to complete a route so values of 0 are not guaranteed to avoid tracks entirely. |
+| `service_penalty` | A penalty applied for transition to generic service road. The default penalty is 0 trucks and 15 for cars, buses, motor scooters and motorcycles. |
```
SvetlanaBayarovich, line 108: 0 for trucks because of 'low_class_penalty' truck option which covers this case to a certain extent

src/sif/bicyclecost.cc
```diff
@@ -373,7 +373,6 @@ class BicycleCost : public DynamicCost {
   float use_roads_;                // Preference of using roads between 0 and 1
   float road_factor_;              // Road factor based on use_roads_
   float avoid_bad_surfaces_;       // Preference of avoiding bad surfaces for the bike type
-  float service_penalty_;          // Penalty (seconds) to use a generic service road
```
SvetlanaBayarovich, line 376: Moved to baas class

src/sif/bicyclecost.cc
```diff
@@ -876,7 +862,7 @@ void ParseBicycleCostOptions(const rapidjson::Document& doc,
         kAlleyPenaltyRange(rapidjson::get_optional<float>(*json_costing_options, "/alley_penalty")
                                .get_value_or(kDefaultAlleyPenalty)));
 
-    // alley_penalty
+    // service_penalty
```
SvetlanaBayarovich, line 865: Just typo fixed

test/gurka/test_route_incidents.cc
```diff
@@ -29,10 +29,10 @@ class IncidentsTest : public ::testing::Test {
 
     // connect the ways via the nodes
     const gurka::ways ways = {
-        {"AB", {{"highway", "tertiary"}}},  {"BC", {{"highway", "tertiary"}}},
+        {"AB", {{"highway", "tertiary"}}},  {"BC", {{"highway", "service"}}},
         {"DEF", {{"highway", "primary"}}},  {"GHI", {{"highway", "primary"}}},
         {"ADG", {{"highway", "motorway"}}}, {"BE", {{"highway", "secondary"}}},
-        {"EH", {{"highway", "service"}}},
+        {"EH", {{"highway", "tertiary"}}},
```
SvetlanaBayarovich, line 35: With new behavior route tries to avoid service roads, so it caused changes for the cases that included `EH` edge. If there is no specific need to have the `EH` edge as a service I suggest placing that `service` class on some dead-end edge so that it does not affect routing, but still has incidents placed on it in some test cases. The alternative would be changing costings (to nullify service penalties) in all the test cases affected

dgearhart:
@SvetlanaBayarovich @dnesbitt61 do we need to include `Use::kParkingAisle` too? To avoid cutting through parking lots
/cc @gknisely

docs/api/turn-by-turn/api-reference.md
stari4ek: do we need this context (`low_class_penalty` for truck) in the doc itself?

SvetlanaBayarovich:
@dgearhart I think this should be investigated separately. As I remember, the initial complaint was for the cases that are service but not parking for some reason although they should, according to the real world surroundings. And I'd rather not mix these changes in one

dnesbitt61:
One possibility could be to create a method, is_service that returns true for any service uses (generic service road, parking aisle, driveway, alley) and have avoid_service avoid any/all of them. Maybe treat alleys separately as they currently are, but lump all other service roads together.

I would think most parking_aisles are not directly connected to through roads but rather are connected to service roads which then connect to the through road.

docs/api/turn-by-turn/api-reference.md
gknisely: I think it is ok to leave out.  Should be noted that we are considering adding a `residential_penalty` and the `low_class_penalty` may get removed.  I think it is better to have a penalty supported by all modes vs just one or the other.

gknisely (approved):

gknisely (approved):

PR #2871, opened 2021-02-17
Check parent dir exists before building live traffic extract

mandeepsandhu:
fopen (used within mtar_open) will fail if the parent directories don't
exist. Throw an exception if traffic extract directory does not exist.

### DIFF
--- a/test/gurka/test_match.cc
+++ b/test/gurka/test_match.cc
@@ -168,7 +168,7 @@ protected:
                                 {"mjolnir.shortcuts", "false"},
                                 {"mjolnir.timezone", VALHALLA_BUILD_DIR "test/data/tz.sqlite"},
                             });
-    map.config.put("mjolnir.traffic_extract", "test/data/algorithm_selection/traffic.tar");
+    map.config.put("mjolnir.traffic_extract", "test/data/match_timedep/traffic.tar");
 
     // add live traffic
     test::build_live_traffic_data(map.config);
--- a/test/test.cc
+++ b/test/test.cc
@@ -21,6 +21,7 @@
 
 #include <boost/algorithm/string.hpp>
 
+#include "filesystem.h"
 #include "microtar.h"
 
 namespace {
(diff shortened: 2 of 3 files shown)


test/gurka/test_match.cc
```diff
@@ -168,7 +168,7 @@ class TrafficBasedTest : public ::testing::Test {
                                 {"mjolnir.shortcuts", "false"},
                                 {"mjolnir.timezone", VALHALLA_BUILD_DIR "test/data/tz.sqlite"},
                             });
-    map.config.put("mjolnir.traffic_extract", "test/data/algorithm_selection/traffic.tar");
+    map.config.put("mjolnir.traffic_extract", "test/data/traffic_test/traffic.tar");
```
danpat, line 171: The path here should be `test/data/match_timedep/traffic.tar`

Gurka creates `test/data/match_timedep` and puts `map.pbf` in there - the directory will exist to also put your `traffic.tar` file in.

mandeepsandhu: updated it. I'm keeping the change to create dir if it doesn't exist, unless you think its better to throw an exception. let me know

test/test.cc
```diff
@@ -438,6 +439,14 @@ void build_live_traffic_data(const boost::property_tree::ptree& config,
   std::string tile_dir = config.get<std::string>("mjolnir.tile_dir");
   std::string traffic_extract = config.get<std::string>("mjolnir.traffic_extract");
 
+  // Create parent path before creating live traffic extract
+  filesystem::path parent_dir = filesystem::path(traffic_extract).parent_path();
+  if (!filesystem::exists(parent_dir)) {
```
mandeepsandhu, line 444: @danpat I've changed it to throw exception if the traffic extract's parent directory(s) does not exist. That will force the user to specify the same dir as that for tiles, instead of creating new directories everywhere.

test/test.cc
```diff
@@ -438,6 +439,14 @@ void build_live_traffic_data(const boost::property_tree::ptree& config,
   std::string tile_dir = config.get<std::string>("mjolnir.tile_dir");
   std::string traffic_extract = config.get<std::string>("mjolnir.traffic_extract");
 
+  // Create parent path before creating live traffic extract
```
danpat, line 442: Comment here is stale.

danpat (dismissed):

danpat (approved):

PR #2872, opened 2021-02-18
Waypoints near edges with poor reachability leads to no route

danpat:
# Issue

When a route waypoint that is not the source or origin is near an edge with dubious connectivity (like oneways-to-nowhere), we are still snapping to it.  This leads to us being able to route *to* that point from the origin, but not onward from that point to the next waypoint in a multi-waypoint route, leading to an overall routing failure.

What we should be doing here is snapping to a different location with better reachability.

So far, this PR just adds a test case that describes the problem.  Fix is still TODO.

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/test/gurka/test_reachability.cc
+++ b/test/gurka/test_reachability.cc
@@ -1,0 +1,43 @@
+#include "gurka.h"
+#include <gtest/gtest.h>
+
+using namespace valhalla;
+
+TEST(Reachability, dont_snap_waypoints_to_deadends) {
+  const std::string ascii_map = R"(
+  A--1-B-2--C
+      / \
+     D   E)";
+
+  // BE and EH are highway=path, so no cars
+  // EI is a shortcut that's not accessible to bikes
+  const gurka::ways ways = {{"AB", {{"highway", "primary"}}},
+                            {"BC", {{"highway", "primary"}}},
+                            {"BD", {{"highway", "secondary"}, {"oneway","yes"}}},
+                            {"BE", {{"highway", "secondary"}, {"oneway","-1"}}}};
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, 25);
+
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/reachability");
+
+  {
+    // Verify simple waypoint routing works
+    auto result1 = gurka::do_action(valhalla::Options::route, map, {"A", "1", "C"}, "auto");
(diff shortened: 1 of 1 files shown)


kevinkreiser:
Yeah the big problem here is the **multipoint** route component. We do indeed also select the low reachability candidate in case we are routing inside of a disconnected island. The algorithm will also get a reachable candidate and both will be used for the route. The problem though is that, we compute the first leg and end up using the low reachability candidate. Then we fail the second leg because we have to start from that candidate. If we really want to preserve the behavior of routing within islands (dubious importance but should be considered) we would need to, after the second leg fails, remove the unreachable candidate, run the first leg again and then continue with the second leg. This can be hell because you can imagine having to back track at every intermediate location on the multipoint route. Another option is to remove all low reachability candidates for intermediate locations and just say screw it, we wont worry about routing inside of disconnected islands on multipoint routes.

danpat:
> Another option is to remove all low reachability candidates for intermediate locations and just say screw it, we wont worry about routing inside of disconnected islands on multipoint routes.

Right, that was my first thought here.  Especially if the waypoints are "far" apart, the odds of them belonging to the same isolated network component diminish.

What I'd ideally like to do here is have the reachability test output the set of edges reached - then before routing, we can look for intersection, and select the set of candidates that are either:

  a. low reachability, but the reachable edge sets intersect, or
  b. all have max reachability (i.e. belong to the large road network).

OSRM's approach to this is to return 1 or 2 candidates - a low reachability candidate if that's the closest thing, and then the nearest candidate with "max reachability", and nothing in between.  If the "max reachability" candidate is closest, then it is returned alone.

Once these 2 options are returned for each waypoint, there's a routine that decides which ones to use - the low reachability candidates if they all touch the same subnetwork, otherwise use all the max reachability candidates.

IMO, OSRM's behaviour regarding this is quite nice - it almost always gives you roughly what you want (a route), even if you drop coarsely placed coordinates (e.g. geocoding city names landing you in the middle of major parks).

The two differences we have to figure out here are:

  1. OSRM's reachability threshold is 1000 edges instead of 100.
  2. OSRM has the ability to pre-calculate the strongly connected networks and number them, so calculating reachability test isn't expensive at query time, and testing for common connectivity of low reachability candidates is trivial.

danpat:
Deprecated by https://github.com/valhalla/valhalla/pull/2914

PR #2873, opened 2021-02-18
Decide at Runtime (based on storage config) What Thread Saftey is Required

### LINKED ISSUE #2867: Valhalla 3.1 is not thread safe
After updating our embedded Valhalla application to Valhalla 3.1, we started seeing intermittent crashes. This happened during multi-threaded route calls. As many threading issues are, this issue was intermittent (hard to replicate) and demonstrated different crash symptoms. After a thorough code-review I found PR #2772 which offers a fix. The problem is, this refers to an undocumented build option and the default configuration builds Valhalla in a non thread safe manner. So my conclusion is that Valhalla 3.1 is not thread-safe by default. Fixes include:

1. This option (if kept) must be documented in the build instructions.
2. The default configuration MUST be with thread safe reference counters enabled. Valhalla is an open source project and its use should be safe with default build configurations. A note on when it is safe to set this to OFF can be attached for those who are sure they do not require thread safety. 
3. Is this build option even necessary? What is the penalty for always using a thread safe reference counter? It seems to me that unless there is a real performance penalty that the code should always be configured to be thread safe.

kevinkreiser:
fixes #2867

The idea here is that we want graphtile to know whether its reference counting should be threadsafe or not based on what tile cache created it. For virtually all tile caches except the `globally_synchronized_cache` thread unsafe reference counting is ok because there is only ever one reference to the graph tile.

At the moment i still need to figure out a way to keep the `GraphTIle` move assignable but the guts of the thread safe counting is prohibiting that..

### DIFF
--- a/valhalla/baldr/graphtile.h
+++ b/valhalla/baldr/graphtile.h
@@ -42,17 +42,14 @@ namespace baldr {
 const std::string SUFFIX_NON_COMPRESSED = ".gph";
 const std::string SUFFIX_COMPRESSED = ".gph.gz";
 
-#ifdef ENABLE_THREAD_SAFE_TILE_REF_COUNT
-using GraphTileRefCounter = boost::thread_safe_counter;
-#else
-using GraphTileRefCounter = boost::thread_unsafe_counter;
-#endif // ENABLE_THREAD_SAFE_TILE_REF_COUNT
+void intrusive_ptr_add_ref(const GraphTile* p) BOOST_SP_NOEXCEPT;
+void intrusive_ptr_release(const GraphTile* p) BOOST_SP_NOEXCEPT;
 
 class tile_getter_t;
 /**
  * Graph information for a tile within the Tiled Hierarchical Graph.
  */
-class GraphTile : public boost::intrusive_ref_counter<GraphTile, GraphTileRefCounter> {
+class GraphTile {
 public:
   static const constexpr char* kTilePathPattern = "{tilePath}";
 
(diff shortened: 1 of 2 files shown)


valhalla/baldr/graphtile.h
```diff
@@ -833,7 +830,37 @@ class GraphTile : public boost::intrusive_ref_counter<GraphTile, GraphTileRefCou
...
+
+  // because boost intrusive pointer uses templates in its provided reference counting implementations
+  // we cannot switch between thread safe and unsafe at run time. this sucks because the only mode of
+  // storage operation that really needs thread saftey is the global_synchronized_cache option.
+  // normally having the extra thread saftey isnt a problem but it costs about 5% performance so here
+  // we have the best of both worlds by allowing  you to pick which kind of reference counting you
+  // want at instantiation
+
+  /*const bool thread_safe_{true};
+  mutable std::atomic<int> atomic_ref_count_{0};*/
```
kevinkreiser, line 842: these things mess up move assignment :frowning:

valhalla/baldr/graphtile.h
```diff
@@ -829,11 +815,52 @@ class GraphTile : public boost::intrusive_ref_counter<GraphTile, GraphTileRefCou
...
+  static std::unique_ptr<const GraphMemory> DecompressTile(const GraphId& graphid,
+                                                           const std::vector<char>& compressed);
+
+  // we encapsulate tile pointers inside of boost::intrusive_ptr which requires reference counting
+  // the default reference counting provided allows for both thread safe and unsafe modes of operation
+  // but.. its templated which makes it impossible to decide at runtime which you need
+  // global_synchronized_cache requires thread safe tiles when they manage their own lifetime
+  // this tile cache tells graphtile to use an atomic counter (which is about %5 more expensive)
+  // all other tile caches are used by individual threads so they dont need to do atomic operations
+  std::unique_ptr<ref_counter> ref_counter_;
```
kevinkreiser, line 830: this is getting close. all that is left to do is to change graphreader to properly construct safe or unsafe variants based on the the t ype of tile cache we have. @dnesbitt61 ill ping you for testing as soon as i get that in

src/baldr/graphtile.cc
```diff
@@ -155,19 +141,18 @@ graph_tile_ptr GraphTile::Create(const GraphId& graphid,
 GraphTile::GraphTile(const GraphId& graphid,
                      std::unique_ptr<const GraphMemory> memory,
                      std::unique_ptr<const GraphMemory> traffic_memory)
-    : header_(nullptr), traffic_tile(std::move(traffic_memory)) {
+    : header_(nullptr), traffic_tile(std::move(traffic_memory)),
+      ref_counter_(new ref_counter{false}) {
```
kevinkreiser, line 145: @dnesbitt61 if you dont want to wait for me to finish this pr before you test, you can also change this to `true` and all the  graph tile ptrs should be thread safe in terms of managing their lifetimes

kevinkreiser:
meh actually its leaking. seems like one of the ways to make a graphtile pointer ends up not getting a ref counter... debugging

PR #2874, opened 2021-02-18
Ensure route/leg summaries are unique

ktatterso:
# Issue

Summary:  When multiple routes are returned, depending on the nature of the routes chosen, each route might have the same route summary.  This is because the original logic would limit the number of route-names in the summary to the two longest legs.

To make the route/leg summaries unique, the new logic keeps adding the next longest named-segment into the summary until that summary is unique among all route/leg summaries.

### Real world example #1
coordinates (lon,lat):
10.169593009081325,50.342127671261125;8.633053521619985,49.918773879486025

Curl request:
curl http://localhost:8002/route --data '{"locations":[{"lat":50.342127671261125,"lon":10.169593009081325,"type":"break"},{"lat":49.918773879486025,"lon":8.633053521619985,"type":"break"}],"costing":"auto","alternates":3,"directions_options":{"units":"kilometers","format":"osrm"}}'

image

Two routes are returned.  Previously, these summaries were returned:
route 1 summary:  B 279, A 66
route 2 summary: B 279, A 66

With my fix these summaries are returned:
route 1 summary: B 279, A 66, A 661
route 2 summary: B 279, L 2304, A 66

### Real world example #2
coordinates (lon,lat):
-115.46257799894421,32.67751487912436;-121.48215941379783,38.55516780677243

Curl request:
curl http://localhost:8002/route --data '{"locations":[{"lat":32.67751487912436,"lon":-115.46257799894421,"type":"break"},{"lat":38.55516780677243,"lon":-121.48215941379783,"type":"break"}],"costing":"auto","alternates":3,"directions_options":{"units":"kilometers","format":"osrm"}}'

image

Two routes are returned.  Previously these summaries were returned:
route 1 summary:  CA 86 North, I 5 North
route 2 summary:  CA 86 North, I 5 North

With my fix these summaries are returned:
route 1 summary: CA 86 North, Foothill Freeway, I 5 North
route 2 summary: CA 86 North, CA 58, I 5 North

### Real world example #3
coordinates (lon,lat):
0.697933819865483,47.3949235422551;7.510451765734274,51.58151535543996

Curl request:
curl http://localhost:8002/route --data '{"locations":[{"lat":47.3949235422551,"lon":0.697933819865483,"type":"break"},{"lat":51.58151535543996,"lon":7.510451765734274,"type":"break"}],"costing":"auto","alternates":3,"directions_options":{"units":"kilometers","format":"osrm"}}'

image

Two routes are returned.  Previously these summaries were returned:
route 1 summary:  A 10, A 1
route 2 summary:  A 10, A 1

With my fix these summaries are returned:
route 1 summary:  A 10, A 1, A 2, E40
route 2 summary:  A 10, A 1, A 2, E314

## Tasklist

 - [x] Add tests

## Requirements / Relations


dgearhart:
@ktatterso can you supply some before and after examples?

merkispavel (commented):
1) A few minor suggestions to use std::move() where we don't need the object anymore
2) I left a comment `src/tyr/route_serializer_osrm.cc` with another possible solution within your approach. Let me know what you think about it

src/tyr/route_serializer_osrm.cc
```diff
@@ -1598,10 +1547,273 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
...
+      std::string leg_summary =
+          rscache.get_n_segment_summary(route_i, leg_idx, num_named_segs_needed);
+
+      leg_summaries.emplace_back(leg_summary);
+    }
+
+    all_summaries.emplace_back(leg_summaries);
+  }
+
+  return all_summaries;
```
merkispavel, line 1814:
```
            |----B---|
-----A------N        |---C------D--
            |                |
         --E|                |
         |                   |
         |                   |
         |                   G
         |                   |
         --------F------------
```
I guess it's reasonable to generate summaries +- like this
1)"A, (any of BC), D"
2)"A, (any of EFG), D"

imho we should look at the nodes where routes fork(as this is what makes them different) and join back.
Not sure if it's possible to implement having current "routes-response", but just my view.

Anyway the current solution code gets rid of summary duplication so I'm okay to merge it

test/gurka/test_route_summary.cc
```diff
@@ -0,0 +1,133 @@
...
+      {"BCDE", {{"highway", "primary"}, {"name", "RT 3"}, {"foot", "yes"}, {"bicycle", "no"}}},
+      {"BGHE", {{"highway", "primary"}, {"name", "RT 4"}, {"foot", "no"}, {"bicycle", "yes"}}},
+      {"EF", {{"highway", "primary"}, {"name", "RT 5"}}},
+  };
+
+  const auto node_layout = gurka::detail::map_to_coordinates(ascii_map, 100, PointLL{5.108, 52.01});
+
+  std::unordered_map<std::string, std::string> config_map =
+      {{"mjolnir.data_processing.use_direction_on_ways", "true"},
+       {"mjolnir.admin", VALHALLA_SOURCE_DIR "test/data/netherlands_admin.sqlite"}};
```
ktatterso, line 46: By specifying an admin we get drive-on-right== true.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1598,10 +1547,273 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
...
+              NamedSegment{map_item.first, map_item.second.first, map_item.second.second});
+        }
+
+        // Sort list by descending maneuver distance
+        std::sort(segs_by_dist.begin(), segs_by_dist.end(),
+                  [](const NamedSegment& a, const NamedSegment& b) {
+                    return b.distance < a.distance;
+                  });
+
+        leg_segs_by_dist.emplace_back(segs_by_dist);
```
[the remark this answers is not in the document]
ktatterso: Good points re std::move(), thanks.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1598,10 +1547,273 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
...
+      }
+
+      route_leg_segs_by_dist.emplace_back(leg_segs_by_dist);
+    }
+
+    // fill the cache with empty results so we can safely index it with the same
+    // dimensionality as the route_leg_segs_by_dist object.
+    cache.reserve(route_leg_segs_by_dist.size());
+    for (size_t i = 0; i < route_leg_segs_by_dist.size(); i++) {
+      cache.emplace_back(std::vector<std::vector<std::string>>{});
```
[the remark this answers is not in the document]
ktatterso: Clean suggestion, thanks.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1598,10 +1547,273 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
...
+
+  // vector 1: routes
+  // vector 2: legs
+  // string: unique summary for the route/leg
+  std::vector<std::vector<std::string>> all_summaries;
+  all_summaries.reserve(routes.size());
+
+  // Find the simplest summary for every leg of every route. Important note:
+  // each route should have the same number of legs. Hence, we only need to make
+  // unique the same leg (leg_idx) between all routes.
```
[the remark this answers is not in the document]
ktatterso: One (admittedly solvable) challenge with having leg-iteration as the outer loop is you don't know how many legs there are until you're on the route.  Otherwise, I don't see an algorithmic advantage, but maybe I'm missing it.  Keep in mind that even though the algorithm is O(n^2) the n's are small (n==number of routes) - and any route summary that is generated is cached, avoiding repeat "heavy" lifting (not really _that_ heavy).

src/tyr/route_serializer_osrm.cc
ktatterso: You are correct.  The nice thing about appending the "next longest leg" into the summary is it is quite deterministic.  The (any of ...) approach could be made deterministic by some heuristic... say, by choosing the longest.  :)

I appreciate the thoughts, thanks.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1655,7 +1874,7 @@ std::string serialize(valhalla::Api& api) {
     route_summary(route, api, imperial, i);
 
     // Serialize route legs
-    route->emplace("legs", serialize_legs(api.directions().routes(i).legs(),
+    route->emplace("legs", serialize_legs(api.directions().routes(i).legs(), route_leg_summaries[i],
```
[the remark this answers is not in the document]
ktatterso: Its splitting hairs I think.  Passing by const-ref is effectively like passing a pointer - very cheap.  The std::move() is likely a few instructions more... it would be an interesting thing to study and see which came out as more efficient.  In this usage/scenario I think I'll just leave as it.  An interesting insight to be sure though, thanks.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1598,10 +1547,273 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
...
+    for (size_t i = 0; i < route_leg_segs_by_dist.size(); i++) {
+      cache.emplace_back(std::vector<std::vector<std::string>>{});
+      cache[i].reserve(route_leg_segs_by_dist[i].size());
+      for (size_t j = 0; j < route_leg_segs_by_dist[i].size(); j++) {
+        cache[i].emplace_back(std::vector<std::string>{});
+        cache[i][j].reserve(route_leg_segs_by_dist[i][j].size());
+        for (size_t k = 0; k < route_leg_segs_by_dist[i][j].size(); k++) {
+          cache[i][j].emplace_back(std::string{});
+        }
+      }
```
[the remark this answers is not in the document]
ktatterso: That's a bit too slick for me... :). I think I'll simplify all the emplace_back()'s to pass nothing... that reads easier to me at least.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1616,10 +1588,272 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
     // Keep the leg
     output_legs->emplace_back(std::move(output_leg));
     leg++;
+    leg_index++;
   }
   return output_legs;
 }
 
+//=============================================================================
```
kevinkreiser, line 1596: this is getting large, what do you think about moving it to a private header since its a nice little encapsulated chunk of functionality?

also yeah the cache is good because of osrm style map matching defaulting to break points for everyy intput location, makes for a ton of legs. you might try benchmarking map matching if you are concerned at all about performance

src/tyr/route_summary_cache.h
```diff
@@ -0,0 +1,105 @@
+#pragma once
```
kevinkreiser, line 1: i wasnt suggesting adding it as a public header i was suggesting taking this and the implementation and putting it in a header inside of the source directory and including it only from the serializer (private header). i think its good to keep implementation details like this out of the top level apis (since we already have a ton of stuff there)

src/tyr/route_serializer_osrm.cc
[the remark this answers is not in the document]
merkispavel: > the n's are small (n==number of routes) 
definitely makes sense 👍

src/tyr/route_serializer_osrm.cc
```diff
@@ -1563,7 +1508,7 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
     // Get a summary based on longest maneuvers.
     double duration = leg->summary().time();
     double distance = units_to_meters(leg->summary().length(), !imperial);
-    output_leg->emplace("summary", summarize_leg(leg));
+    output_leg->emplace("summary", leg_summaries[leg_index]);
```
merkispavel, line 1511: `output_leg` is an `unordered_map` under the hood so it's better to move arguments

src/tyr/route_serializer_osrm.cc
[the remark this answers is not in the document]
merkispavel: Sorry that I didn't make it clear from the start :) Check it out https://github.com/valhalla/valhalla/pull/2874/files#r584849492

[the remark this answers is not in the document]
merkispavel: So my idea is to move `route_leg_summaries[i]` so that we could move its elements inside `serialize_legs` function

src/tyr/route_summary_cache.h
ktatterso: Hmm.  I saw the route_summary_cache as only ever meant to be utilized by one code consumer - hence why it was implemented in the source file.

Moving it into a header implies the possibility that another code consumer could use it.  I didn't really see it being the case, but that's how I read your suggestion.  In which case, the methods were too big to justify being inlined, hence why I out-of-lined them.

But this comes down to a design preference and I'm starting to read between the lines.  I think you're suggesting that the the route_summary_cache 1) still just have one code customer, 2) be fully inlined which is in effect the same as having the route_summary_cache implemented in the source, but 3) allows the code to be modularized into its own home.

All that being said, I'm glad to take your suggestion.  Following existing design precedents is important to me.

src/tyr/route_serializer_osrm.cc
ktatterso: I'm not sure what you're suggesting.  As you say output_leg is an unordered_map - and the emplace_back() method will use the move operator if possible.  What is being passed to it is a std::string, which has a move operator.

Within CLion I did a "go to definition" on that line of code and sure enough, it takes me to:
`pair<iterator, bool> emplace(_Args&&... __args)`

I believe that means the move operator will be employed.

src/tyr/route_summary_cache.h
kevinkreiser: yeah sorry i didnt spell out why i was asking that very well, apologies for that. number 3 is probably the main reason why i suggested it. giving it a small footprint but letting it be sequestered into its own area seems important to me. it also makes it really straight forward to unit test directly.

src/tyr/route_serializer_osrm.cc
merkispavel: yeah, exactly. I believe so as well

ktatterso: thanks

src/tyr/route_summary_cache.h
ktatterso: thanks, suggestion taken.

merkispavel (approved):
well done! thx

dgearhart:
@ktatterso I will check out branch and try some more examples. I will post what I was mentioning the other day

nilsnolde:
@ktatterso re format.sh script: I know the pain.. I added `cat scripts/format.sh > .git/hooks/pre-commit && tail -n +7 scripts/error_on_dirty.sh >> .git/hooks/pre-commit` to have it as pre-commit hook and error out when it's dirty before committing

dgearhart (commented):
Works nicely on most of the test cases I ran. My only comment was what I mentioned the other day - since we collapsed some maneuvers with the obvious maneuver logic - summary may not always reflect primary street names based on the edges. Thanks @ktatterso

ktatterso:
Thanks @dgearhart

PR #2875, opened 2021-02-18
U-turns being forbidden at barriers, leading to no-routes when inputs force conditions that would require a u-turn

danpat:
# Issue

Not 100% sure what the issue here is yet, but this test case demonstrates the problem.  When you use `type=break_through` at waypoints, and a u-turn is required at a dead-end, a NoRoute result is the outcome.

I'm not sure whether it's the `break_through` that's the root cause, or the `access=no` on the necessary u-turn location in this test, but the test shows the problem.

cc @gknisely 

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -154,6 +154,10 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
   if (!costing_->Allowed(nodeinfo)) {
     const DirectedEdge* opp_edge = nullptr;
     const GraphId opp_edge_id = graphreader.GetOpposingEdgeId(pred.edgeid(), opp_edge, tile);
+    // Mark the predecessor as a deadend to be consistent with how the
+    // edgelabels are set when an *actual* deadend (i.e. some dangling OSM geometry)
+    // is labelled
+    pred.set_deadend(true);
     // Check if edge is null before using it (can happen with regional data sets)
     return opp_edge &&
            ExpandForwardInner(graphreader, pred, nodeinfo, pred_idx,
@@ -362,6 +366,10 @@ bool BidirectionalAStar::ExpandReverse(GraphReader& graphreader,
   if (!costing_->Allowed(nodeinfo)) {
     const DirectedEdge* opp_edge = nullptr;
     const GraphId opp_edge_id = graphreader.GetOpposingEdgeId(pred.edgeid(), opp_edge, tile);
+    // Mark the predecessor as a deadend to be consistent with how the
+    // edgelabels are set when an *actual* deadend (i.e. some dangling OSM geometry)
+    // is labelled
+    pred.set_deadend(true);
     // Check if edge is null before using it (can happen with regional data sets)
     return opp_edge &&
            ExpandReverseInner(graphreader, pred, opp_pred_edge, nodeinfo, pred_idx,
--- a/test/gurka/test_barrier_uturns.cc
+++ b/test/gurka/test_barrier_uturns.cc
@@ -1,0 +1,65 @@
+#include "gurka.h"
+#include <gtest/gtest.h>
+
+using namespace valhalla;
+
+class BarrierUturns : public testing::TestWithParam<int> {
+protected:
+  static gurka::map map;
+
+  static void SetUpTestSuite() {
+    const std::string ascii_map = R"(
+  A----B----C
+       1    |
+       D    /
+       |   /
+       E--F)";
+
+    // BD and BE are oneways that lead away, and toward the main road A-B-C
+    const gurka::ways ways = {{"AB", {{"highway", "primary"}}}, {"BC", {{"highway", "primary"}}},
+                              {"BD", {{"highway", "primary"}}}, {"DE", {{"highway", "primary"}}},
+                              {"EF", {{"highway", "primary"}}}, {"CF", {{"highway", "primary"}}}};
+    const gurka::nodes nodes = {{"D", {{"barrier", "fence"}, {"access", "no"}}}};
+    const auto layout = gurka::detail::map_to_coordinates(ascii_map, 25);
+
+    map = gurka::buildtiles(layout, ways, nodes, {}, "test/data/break_through");
+  }
+};
+
+gurka::map BarrierUturns::map = {};
+
+TEST_F(BarrierUturns, break_waypoint) {
+  // Check with type="break" for middle waypoint (the default)
+  auto result1 = gurka::do_action(valhalla::Options::route, map, {"A", "1", "F"}, "auto");
+  gurka::assert::raw::expect_path(result1, {"AB", "BD", "BD", "BC", "CF"});
+}
+
+TEST_F(BarrierUturns, break_thruough_waypoint_into_barrier) {
+  // Check with type="break_through" for middle waypoint
+  auto result1 = gurka::do_action(valhalla::Options::route, map, {"A", "1", "F"}, "auto",
+                                  {{"/locations/1/type", "break_through"}});
+  gurka::assert::raw::expect_path(result1, {"AB", "BD", "BD", "BD", "BC", "CF"});
+}
+
+// Test various algorithms with forced heading
+TEST_P(BarrierUturns, forced_heading_into_barrier_with_algorithm) {
+  const std::string& request =
+      (boost::format(
+           R"({"locations":[{"lat":%s,"lon":%s,"heading":180,"heading_tolerance":45},{"lat":%s,"lon":%s}],
+             "costing":"auto",
+             "date_time": { "type": %d, "value": "2020-10-30T09:00"},
+             "costing_options": { "auto": { "speed_types": ["constrained"]}}})") %
+       std::to_string(map.nodes.at("1").lat()) % std::to_string(map.nodes.at("1").lng()) %
+       std::to_string(map.nodes.at("F").lat()) % std::to_string(map.nodes.at("F").lng()) % GetParam())
+          .str();
+  auto result1 = gurka::do_action(valhalla::Options::route, map, request);
+  gurka::assert::raw::expect_path(result1, {"BD", "BD", "BC", "CF"});
+}
+
+/**
+ * 0 = current time (timedep_forward)
+ * 1 = departure time (timedep_forward)
+ * 2 = arrival time (timedep_reverse)
+ * 3 = invariant time (bidirectional_astar)
+ */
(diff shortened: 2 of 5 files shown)


kevinkreiser:
yeah `break_through` means you cant turn around at the snap point. it means you at least have to go to the end of the edge, ie continue through the snap point. its basically a way of saying "break" (i want you to make a new leg here) but "go through" (i want you to keep going in the same direction, no immediate uturn at the snap point). but yeah if that leads you into a black hole then i could see why that'd be a problem :smile:

kevinkreiser:
@danpat im a bit confused though. i think recently we (@danpaz)changed this behavior to allow a uturn even where the node has no access: https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L142

:thinking:

danpaz:
The change in bidirectional at https://github.com/valhalla/valhalla/pull/2705/files#r534545579 was meant to address cases like these, by allowing a uturn at nodes where access is disallowed. Essentially we now enqueue the reverse edge, where before we were bailing and returning a NoRoute. Now it seems we're back to the old behavior.

It seems like there was a regression around that code, or around the special case handling uturns just below it.

mandeepsandhu:
When routing with multiple waypoints, do we have to start at the same location the previous leg ended on? Is it not acceptable to, say snap to a different edge if its not possible to continue on the edge the previous leg ended on?

kevinkreiser:
so currently the code will look at the edge the last leg ended on and remove all candidates but that one when its a "*_through" location. for other types i forget if it does weed them out or not. but if it didnt the route would be pretty wild in that there would be a teleport between snap locations

danpat:
To be clear - the test isn't expecting a u-turn at the middle waypoint (1) here - it's expecting to continue through the waypoint, then perform the u-turn at the `barrier=fence,access=node` at the end of the street.  For some reason, it's not doing that.

I'm wondering whether use of `heading` and `heading_tolerance` could be used to force the same situation without the use of a `break_through` waypoint.  I will try, and push up an additional test if so.

kevinkreiser:
@danpat yep that is what i understood from your description. i think the changes @danpaz and i linked should account for that but im not sure why it still fails!

danpat:
Just pushed an additional test using `heading` - confirms it's not the `type=break_through`, but the u-turn at the barrier.

src/thor/bidirectional_astar.cc
```diff
@@ -143,6 +143,10 @@ bool BidirectionalAStar::ExpandForward(GraphReader& graphreader,
   if (!costing_->Allowed(nodeinfo)) {
     const DirectedEdge* opp_edge;
     const GraphId opp_edge_id = graphreader.GetOpposingEdgeId(pred.edgeid(), opp_edge, tile);
+    // Mark the predecessor as a deadend to be consistent with how the
+    // edgelabels are set when an *actual* deadend (i.e. some dangling OSM geometry)
+    // is labelled
+    pred.set_deadend(true);
```
kevinkreiser, line 149: nice catch, we totally forgot to do that!

danpat: Took a little digging to understand the usage of the deadend flag, but it was a one-liner (per algorithm!) in the end.

@kevinkreiser I also fixed timedep_forward/reverse - are there any other instances that likely need this?

When is `dijkstra.cc` used?  It doesn't have any of the work from https://github.com/valhalla/valhalla/pull/2705 - was that deliberate?

kevinkreiser: i was going to mention those two but saw you already got them :+1: 

dijkstras is used in the reach check when the faster conservative reach check fails to find minimum reachability. the lack of derived deadend support in that algorithm is definitely an oversight.

i think its ok to note it in an issue though and come back to it in a separate PR if you would prefer it (tests will change because of it so it might be a decent sized rabbit hole). specifically we now have two types of deadend per search tree:

1. deadends because of node access which you just fixed or completed :wink: 
2. deadends because of edge access which was already there from a long time ago refactor @PureW and i did to fix other no route scenarios
 
both should make it into dijkstras to make the algorithm robust in the way the others are but like i said it might be a bit of a task to wade through all the test changes so maybe its better as a separate issue/PR?

kevinkreiser (dismissed):

danpaz (approved):
Thanks @danpat !

src/thor/bidirectional_astar.cc
```diff
@@ -351,6 +355,10 @@ bool BidirectionalAStar::ExpandReverse(GraphReader& graphreader,
   if (!costing_->Allowed(nodeinfo)) {
     const DirectedEdge* opp_edge = nullptr;
     const GraphId opp_edge_id = graphreader.GetOpposingEdgeId(pred.edgeid(), opp_edge, tile);
+    // Mark the predecessor as a deadend to be consistent with how the
+    // edgelabels are set when an *actual* deadend (i.e. some dangling OSM geometry)
+    // is labelled
+    pred.set_deadend(true);
```
danpat, line 361: @kevinkreiser I'm not sure my tests hit the reverse case here, so a sanity check would be good: is `pred` in the reverse test supposed to be marked as a dead-end?  I assume it's the directed edge heading *away* from the deadend point?

The logic in `AutoCost::AllowedReverse()` simply looks at `pred.deadend()`, which wasn't the inverted logic I was expecting.

Is `deadend` supposed to be set on *both* directed edges connected to a deadend?  If yes, then all is well here, but it's a bit confusing as you can't tell without further digging whether an edge is coming *from* a deadend or going *to* one.

danpat: Same question for `timedep_reverse` I guess - I copy-pasta'd the one-line "fix", but is that actually what we want?

kevinkreiser: `DirectedEdge::deadend` is deprecated, we should note that in its javadoc comment. when we made the original change to the algorithms to find derived deadends we considered marking it spare but we couldnt because then old code using new tiles wouldnt work properly. so we have to keep properly setting that flag/bit at tile creation but the current algorithms shouldnt use it they should derive deadend-ness as you and dan and anders and i have been adding to the algorithms.

an easy way to test that its doing what you want in reverse is to set and arrive by time and put the destination behind a uturn at a node. this will trigger timedep_reverse which should work given your changes. the reverse search uses the opposing edges of the real edges we use on the path, thats why those methods take an opposing edge for allowed so it can check the real edge the path would be taking

danpat: @kevinkreiser Hmm, not quite deprecated.  `DirectedEdge::deadend()` is still used in the constructor of `EdgeLabel` (I tried removing the getter function from DirectedEdge and it leads to compilation failure) where labels inherit whatever value is set on the directededge.  I will look and see if that is easy to remove so we can fully deprecate it to avoid future confusion.

test/gurka/test_barrier_uturns.cc
```diff
@@ -0,0 +1,131 @@
...
+         std::to_string(map.nodes.at("1").lat()) % std::to_string(map.nodes.at("1").lng()) %
+         std::to_string(map.nodes.at("F").lat()) % std::to_string(map.nodes.at("F").lng()))
+            .str();
+    auto result1 = gurka::do_action(valhalla::Options::route, map, request);
+
+    gurka::assert::raw::expect_path(result1, {"BD", "BD", "BC", "CF"});
+  }
+
+  {
+    // Force heading due-south for the initial coordinate, which should run the route into the barrier
```
mandeepsandhu, line 51: This test looks like a repeat of the previous test?

test/gurka/test_barrier_uturns.cc
```diff
@@ -0,0 +1,131 @@
...
+        (boost::format(
+             R"({"locations":[{"lat":%s,"lon":%s,"heading":180,"heading_tolerance":45},{"lat":%s,"lon":%s}],"costing":"auto"})") %
+         std::to_string(map.nodes.at("1").lat()) % std::to_string(map.nodes.at("1").lng()) %
+         std::to_string(map.nodes.at("F").lat()) % std::to_string(map.nodes.at("F").lng()))
+            .str();
+    auto result1 = gurka::do_action(valhalla::Options::route, map, request);
+    gurka::assert::raw::expect_path(result1, {"BD", "BD", "BC", "CF"});
+  }
+
+  // Cover all the routing algorithms (bidirectional astar, astar, timedep forward, timedep reverse)
```
mandeepsandhu, line 63: Looks like there's only 1 parameter `date_time:type` thats changing between tests. You can use [parametrized fixtures](https://github.com/google/googletest/blob/master/docs/advanced.md#how-to-write-value-parameterized-tests) to pass-in that value and avoid repeating the test body.

src/thor/bidirectional_astar.cc
kevinkreiser: yeah it was many months ago, i remember looking at it with @PureW but i couldnt remember the outcome exactly. its probably better to leave it for now until dijkstras (and the matrix algorithms and multimodal and bss) get derived deadend support anyway.

danpat: @kevinkreiser ok, for this PR I'll leave it - but boy, there sure is a lot of logic duplication in here.  I would expect a lot of this stuff to either live in a base class, or all these algorithms to derive from a template interface that supplied much of this kind of detail.

kevinkreiser: ive eye'd that up too a couple of times but it just doesnt work out for slightly annoying reasons/differences between the algorithms. there might be a way to slice it just right so the code can be shared but it might involve using the same edge labels everywhere or some tricks with the astar heuristic. its definitely tempting but the devils in the details :smile:

danpat: Well, the downside to *not* doing it is that the implementations in these algorithms are slowly diverging over time, leading to inconsistent results between actions (e.g. many_to_many between A->B could return different results to a route from A->B).

test/gurka/test_barrier_uturns.cc
danpat: Good idea - cleaned this up.

test/gurka/test_barrier_uturns.cc
danpat: It was - good 👀

danpat:
Did some cleanup on the test - @mandeepsandhu @danpaz @kevinkreiser anyone mind 👍 on this again?

kevinkreiser (approved):
thanks again!

src/thor/bidirectional_astar.cc
danpat: Opened a ticket to review all this stuff - I think we need some consolidation. https://github.com/valhalla/valhalla/issues/2885

Issue #2830, opened 2021-02-02
Compilation errors with ENABLE_SERVICES=Off


xlqian:
Hello, 
We struggled a little bit with the compilation of 3.1.0. Our app does not require the service, so we compiled with `ENABLE_SERVICES=Off`, it seems a chunk of codes is deactivated by the macro and BOOM!

```
/valhalla/src/thor/worker.cc:55:13: error: ‘std::__cxx11::string {anonymous}::serialize_to_pbf(valhalla::Api&)’ defined but not used [-Werror=unused-function]
 std::string serialize_to_pbf(Api& request) {                                                                          
             ^~~~~~~~~~~~~~~~                                                                   
cc1plus: all warnings being treated as errors                                       
src/thor/CMakeFiles/valhalla-thor.dir/build.make:614: recipe for target 'src/thor/CMakeFiles/valhalla-thor.dir/worker.cc.o' failed
make[2]: *** [src/thor/CMakeFiles/valhalla-thor.dir/worker.cc.o] Error 1                  
make[2]: *** Waiting for unfinished jobs....                                
CMakeFiles/Makefile2:1535: recipe for target 'src/thor/CMakeFiles/valhalla-thor.dir/all' failed
make[1]: *** [src/thor/CMakeFiles/valhalla-thor.dir/all] Error 2                           
Makefile:151: recipe for target 'all' failed                                                                                                                        
make: *** [all] Error 2        
``` 

The flag is activated here: https://github.com/valhalla/valhalla/blob/3.1.0/src/thor/CMakeLists.txt#L38

Is that a "bug"?

best regards

kevinkreiser:
That's been fixed in master. We'll do a new release shortly

erdmanncross:
Hello.
I am trying to compile using the script:
https://github.com/valhalla/valhalla/blob/master/scripts/Ubuntu_Bionic_Install.sh
however is giving this error:
```
[ 70%] Built target valhalla-loki
[ 70%] Building CXX object src/thor/CMakeFiles/valhalla-thor.dir/route_matcher.cc.o
[...]

PR #2878, opened 2021-02-19
fix compilation error when boost < 1.68 and libprotobuf < 3.6 

xlqian:
# Issue https://github.com/valhalla/valhalla/issues/2830#issuecomment-772468377

1. Fix when libprotobuf is of a lower version
2. Fix when boost is of a lower version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 * **Removed**
 
 * **Bug Fix**
-   * FIXED: Fix compilation errors when boost < 1.68 and libprotobuf < 3.6  [#2878](https://github.com/valhalla/valhalla/pull/2878)
    * FIXED: Fix compression_utils.cc::inflate(...) throw - make it catchable [#2839](https://github.com/valhalla/valhalla/pull/2839)
    * FIXED: Fix compiler errors if HAVE_HTTP not enabled [#2807](https://github.com/valhalla/valhalla/pull/2807)
    * FIXED: Fix alternate route serialization [#2811](https://github.com/valhalla/valhalla/pull/2811)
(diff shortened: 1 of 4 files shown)


src/mjolnir/valhalla_add_predicted_traffic.cc
```diff
@@ -173,7 +173,7 @@ void update_tile(const std::string& tile_dir,
   size_t pred_count = 0;
   for (uint32_t j = 0; j < tile_builder.header()->directededgecount(); ++j) {
     auto found = speeds.find(j);
-    pred_count += found != speeds.cend() && found->second.coefficients.has_value();
+    pred_count += found != speeds.cend() && bool(found->second.coefficients);
```
xlqian, line 176: https://www.boost.org/doc/libs/1_68_0/libs/optional/doc/html/boost_optional/reference/header__boost_optional_optional_hpp_/detailed_semantics___optional_values.html

basically, `bool()` and `has_value` is the same thing

kevinkreiser: yep i noticed this too. would you mind using `static_cast<bool>()` for these?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 * **Removed**
 
 * **Bug Fix**
+   * FIXED: Fix compilation errors when boost < 1.68 and libprotobuf < 3.6  [#2878](https://github.com/valhalla/valhalla/pull/2878)
```
kevinkreiser, line 5: can you move this to the bottom of the list, we've been following that pattern. basically inside of a release the oldest stuff is at the top and the youngest stuff is at the bottom. i can see why that is somewhat unappealing though.... maybe we can change it in the next release and follow that from now on

src/mjolnir/valhalla_add_predicted_traffic.cc
xlqian: yeah sure, but for me, `bool()` is a defined specialized  explicit operator from `boost::optional`(even in lower version) class... it's not a "just" casting :)

src/thor/route_matcher.cc
```diff
@@ -191,7 +191,7 @@ bool expand_from_node(const mode_costing_t& mode_costing,
         elapsed += cost;
         // overwrite time with timestamps
         if (use_timestamps)
-          elapsed.secs = shape[index].time() - shape[0].time();
+          elapsed.secs = shape.Get(index).time() - shape.Get(0).time();
```
xlqian, line 194: hmmm, how can I test this so that codecov would be contented?

kevinkreiser: that shouldnt be your burden. we can merge and just ignore codecov in this instance. this is for the fast exact map matching endpoint which as a whole could use more structured testing.

kevinkreiser (approved):

PR #2879, opened 2021-02-21
Added documentation example of decoding encoded path shapes using R.

chris31415926535:
This PR adds an example of how to decode Valhalla's encoded paths using R to the documentation. It's a straightforward (i.e. not optimized) R implementation of the path decoding algorithm used in the examples for the other languages. I've tested it and it worked fine.

Since this is just a minor addition to the documentation, please let me know if you need anything else from me.

### DIFF
--- a/docs/decoding.md
+++ b/docs/decoding.md
@@ -161,3 +161,65 @@ def decode(encoded):
 
 print(decode(sys.argv[1]))
 ```
+
+## R
+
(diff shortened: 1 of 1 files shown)


kevinkreiser (approved):
I'll have to take your word for this one, im no expert on R. if anyone else has more experience please chime in!

PR #2890, opened 2021-02-25
DNM: nit: Serializes incident-id as integer rather than string

PureW:


mandeepsandhu (approved):
Are incidence IDs supposed to be always numeric? Is the ID something we internally generate or does it come from an external system?

kevinkreiser:
the one idea was that we wanted to generate them based on immutable properties of the incident. ive made a few suggestions on how to do this in PRs past.... but i'll be darned if i can find them. the basic idea is we make a bitfield and encode some part of the incident (including location and time) into the 64bit id. the trick is picking features of the incident that reduce id collisions

kevinkreiser (approved):
yeah we went back and forth on this a couple of times. im happy that we ended on int

PureW:
Closing this due to it being a breaking api change with clients already consuming strings.

PR #2891, opened 2021-02-25
Reduce stop impact cost when flow data is present

gknisely:
# Issue

This PR(collaboration between @dgearhart  and @gknisely ) is the first attempt at reducing the favoring of longer routes or highways for very little time improvement.  Basically, we reduced transition cost when edge has flow information.  A bit has been added to the edgelabel which indicates that we have flow information or traffic set from one of the flow sources: kFreeFlowMask, kConstrainedFlowMask, kPredictedFlowMask, or kCurrentFlowMask.

Note: This update is for only the auto costing for now.   

## Examples
| BEFORE | AFTER |
|---|---|
| image | image |
| image | image |
| image | image |

## Performance
No performance impact with these changes
### Average cross country route
| MASTER | BRANCH |
|---|---|
| 219 ms | 219 ms |

## Tasklist

 - [x] Add tests


dnesbitt61:
can you provide a description of the reason for the changes and the expected impact?

gknisely:
> can you provide a description of the reason for the changes and the expected impact?

Yes.  Still in draft mode for this PR

src/sif/autocost.cc
```diff
@@ -262,12 +266,14 @@ class AutoCost : public DynamicCost {
...
    * @param  edge  the opposing predecessor in the reverse tree
+   * @param  has_flow_speed Do we have any of the measured speed types set?
    * @return  Returns the cost and time (seconds)
    */
   virtual Cost TransitionCostReverse(const uint32_t idx,
                                      const baldr::NodeInfo* node,
                                      const baldr::DirectedEdge* pred,
-                                     const baldr::DirectedEdge* edge) const override;
+                                     const baldr::DirectedEdge* edge,
+                                     const bool has_flow_speed) const override;
```
gknisely, line 276: This is a TODO:  We should try to update TransitionCostReverse with `const EdgeLabel& pred` vs `const baldr::DirectedEdge* pred`; however, we did not have time to make this change.  

@dnesbitt noted that when expanding in reverse the edge added to the edge label is in the direction of expansion.  This is the reverse of the direction of travel. This is why the opposing predecessor directed edge is passed into the TransitionCostReverse method rather than the predecessor EdgeLabel. With some thought and care it may be possible to refactor this but I'm not sure. We need to make sure idx is computed correctly is one complexity

src/sif/autocost.cc
```diff
@@ -109,6 +109,9 @@ constexpr ranged_default_t<float> kUseLivingStreetsRange{0.f, kDefaultUseLivingS
 constexpr ranged_default_t<float> kServicePenaltyRange{0.0f, kDefaultServicePenalty, kMaxPenalty};
 constexpr ranged_default_t<float> kServiceFactorRange{kMinFactor, kDefaultServiceFactor, kMaxFactor};
 
+// Maximum highway avoidance bias (modulates the highway factors based on road class)
```
gknisely, line 112: Ported this over from the motorcycle costing.

src/sif/autocost.cc
```diff
@@ -353,8 +359,18 @@ AutoCost::AutoCost(const CostingOptions& costing_options, uint32_t access_mask)
   alley_factor_ = costing_options.alley_factor();
 
   // Preference to use highways. Is a value from 0 to 1
-  float use_highways_ = costing_options.use_highways();
-  highway_factor_ = 1.0f - use_highways_;
+  // Factor for highway use - use a non-linear factor with values at 0.5 being neutral (factor
+  // of 0). Values between 0.5 and 1 slowly decrease to a maximum of -0.125 (to slightly prefer
+  // highways) while values between 0.5 to 0 slowly increase to a maximum of kMaxHighwayBiasFactor
+  // to avoid/penalize highways.
+  float use_highways = costing_options.use_highways();
```
gknisely, line 366: Again, ported from motorcycle costing

src/sif/autocost.cc
```diff
@@ -432,9 +448,10 @@ bool AutoCost::AllowedReverse(const baldr::DirectedEdge* edge,
 // Get the cost to traverse the edge in seconds
 Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
                         const graph_tile_ptr& tile,
-                        const uint32_t seconds) const {
+                        const uint32_t seconds,
+                        uint8_t& flow_sources) const {
   // either the computed edge speed or optional top_speed
-  auto edge_speed = tile->GetSpeed(edge, flow_mask_, seconds);
+  auto edge_speed = tile->GetSpeed(edge, flow_mask_, seconds, false, &flow_sources);
```
gknisely, line 454: Adding `&flow_sources` here allows us to save off which speed sources were used in this speed calculation.  This is flattened down to a single bit on the edgelabel.  Which basically indicates if we have some sort of traffic data for this edge.

src/thor/bidirectional_astar.cc
```diff
@@ -322,6 +323,9 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
                                    dist, mode_, transition_cost,
                                    (pred.not_thru_pruning() || !meta.edge->not_thru()),
                                    (pred.closure_pruning() || !costing_->IsClosed(meta.edge, tile)),
+                                   static_cast<bool>(flow_sources &
+                                                     (kFreeFlowMask | kConstrainedFlowMask |
+                                                      kPredictedFlowMask | kCurrentFlowMask)),
                                    restriction_idx);
```
gknisely, line 329: Saving the flow information to the edgelabel. If we have one of the flow sources: kFreeFlowMask, kConstrainedFlowMask, kPredictedFlowMask, or kCurrentFlowMask then we have flow data and set the bit to true.

src/thor/bidirectional_astar.cc
```diff
@@ -512,9 +516,12 @@ inline bool BidirectionalAStar::ExpandReverseInner(GraphReader& graphreader,
   // Get cost. Use opposing edge for EdgeCost. Separate the transition seconds so we
   // can properly recover elapsed time on the reverse path.
   const Cost transition_cost =
-      costing_->TransitionCostReverse(meta.edge->localedgeidx(), nodeinfo, opp_edge, opp_pred_edge);
-  const Cost newcost =
-      pred.cost() + costing_->EdgeCost(opp_edge, t2, time_info.second_of_week) + transition_cost;
+      costing_->TransitionCostReverse(meta.edge->localedgeidx(), nodeinfo, opp_edge, opp_pred_edge,
+                                      pred.has_flow_speed());
+  uint8_t flow_sources;
+  const Cost newcost = pred.cost() +
```
gknisely, line 522: Added flow_sources to TransitionCostReverse so that we can save it off in the edge label.

src/thor/bidirectional_astar.cc
```diff
@@ -285,8 +285,9 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
 
   // Get cost. Separate out transition cost.
   Cost transition_cost = costing_->TransitionCost(meta.edge, nodeinfo, pred);
-  Cost newcost =
-      pred.cost() + transition_cost + costing_->EdgeCost(meta.edge, tile, time_info.second_of_week);
+  uint8_t flow_sources;
+  Cost newcost = pred.cost() + transition_cost +
+                 costing_->EdgeCost(meta.edge, tile, time_info.second_of_week, flow_sources);
```
gknisely, line 291: Added flow_sources to EdgeCost so that we can save it off in the edge label.

valhalla/sif/edgelabel.h
```diff
@@ -411,7 +424,7 @@ class EdgeLabel {
   uint64_t deadend_ : 1;
   uint64_t on_complex_rest_ : 1;
   uint64_t closure_pruning_ : 1;
-  uint64_t spare_0 : 1;
+  uint64_t has_flow_speed_ : 1;
```
gknisely, line 428: Our new bit for flow.

src/sif/autocost.cc
```diff
@@ -353,8 +359,18 @@ AutoCost::AutoCost(const CostingOptions& costing_options, uint32_t access_mask)
...
-  float use_highways_ = costing_options.use_highways();
-  highway_factor_ = 1.0f - use_highways_;
+  // Factor for highway use - use a non-linear factor with values at 0.5 being neutral (factor
+  // of 0). Values between 0.5 and 1 slowly decrease to a maximum of -0.125 (to slightly prefer
+  // highways) while values between 0.5 to 0 slowly increase to a maximum of kMaxHighwayBiasFactor
+  // to avoid/penalize highways.
+  float use_highways = costing_options.use_highways();
+  if (use_highways >= 0.5f) {
+    float f = (0.5f - use_highways);
+    highway_factor_ = f * f * f;
```
mandeepsandhu, line 369: dumb q: whats the reasoning for the math here (why the cube)?

valhalla/sif/edgelabel.h
kevinkreiser: i would suggest we name that `has_measured_speed_` since we really want this for any edge that has measured traffic speed (flow/constrained/historical/realtime)

src/sif/autocost.cc
kevinkreiser: obviously because @gknisely only has 3 F's to give! badum-cha!

gknisely: copied from motorcycle costing.  we were seeing better results.  @dnesbitt61 originally wrote this logic.  Do you know why you did this?

dnesbitt61: some secret sauce likely arrived at doing some route testing with motorcycles routes to see what gave results that seemed good. Luckily I commented my approach:
  // Factor for trail use - use a non-linear factor with values at 0.5 being neutral (factor
  // of 0). Values between 0.5 and 1 slowly decrease to a maximum of -0.125 (to slightly prefer
  // trails) while values between 0.5 to 0 slowly increase to a maximum of the surfact_factor_
  // to avoid/penalize trails.
Now this was applied to motorcycles where there were 2 different route case: first being for road motorcycling and second more adventure motorcyling. So I cannot vouch for its use in auto costing

src/sif/autocost.cc
```diff
@@ -564,11 +593,23 @@ Cost AutoCost::TransitionCostReverse(const uint32_t idx,
...
+    bool is_turn = false;
+    if (edge->turntype(idx) == baldr::Turn::Type::kLeft ||
+        edge->turntype(idx) == baldr::Turn::Type::kSharpLeft ||
+        edge->turntype(idx) == baldr::Turn::Type::kRight ||
+        edge->turntype(idx) == baldr::Turn::Type::kSharpRight) {
+      seconds *= edge->stopimpact(idx);
+      is_turn = true;
+    }
 
+    // Apply density factor and stop impact penalty if there isn't traffic on this edge or you're not
```
dgearhart, line 606: This is where reduced stop impact cost when using measured speed

src/sif/autocost.cc
```diff
@@ -516,11 +532,23 @@ Cost AutoCost::TransitionCost(const baldr::DirectedEdge* edge,
...
```
dgearhart, line 545: This is where reduced stop impact cost when using measured speed

dgearhart: More work to be done in the future to balance cost with measured speed

valhalla/sif/edgelabel.h
gknisely: @kevinkreiser updated

run_route_scripts/batch.sh
```diff
@@ -58,7 +58,7 @@ cat "${TMP}" | parallel --progress -k -C '\|' -P "${CONCURRENCY}" "valhalla_run_
 # "-e" suffix, even though -i arg is empty. The wildcard below will ensure that
 # that file is deleted as well.
 rm -f ${TMP}*
-echo "orgLat, orgLng, destLat, destLng, result, #Passes, runtime, trip time, length, arcDistance, #Manuevers" > ${RESULTS_OUTDIR}/statistics.csv
+echo "orgLat, orgLng, destLat, destLng, result, #Passes, runtime, trip time, length, arcDistance, #Manuevers, elapsedCostSeconds, elapsedCostCost" > ${RESULTS_OUTDIR}/statistics.csv
```
dgearhart, line 61: Header updates fro elapsed cost statistics

run_route_scripts/results/combine_route_stats.py
```diff
@@ -5,7 +5,7 @@
 import sys
 
 
-STATS_TO_DIFF = ['#Passes', 'runtime', 'trip time', 'length', '#Manuevers']
+STATS_TO_DIFF = ['#Passes', 'runtime', 'trip time', 'length', '#Manuevers', 'elapsedCostSeconds', 'elapsedCostCost']
```
dgearhart, line 8: enable elapsed cost stats for delta and % delta of baseline vs latest

src/valhalla_run_route.cc
```diff
@@ -71,11 +71,13 @@ class PathStatistics {
   float trip_dist;
   float arc_dist;
   uint32_t manuevers;
+  double elapsed_cost_seconds;
```
dgearhart, line 74: add elapsed cost into stats

kdiluca (approved):
Knowing that the RAD results were good for this then I approve however we need to add more tests!

src/sif/autocost.cc
```diff
@@ -564,11 +593,23 @@ Cost AutoCost::TransitionCostReverse(const uint32_t idx,
...
+    // using traffic
+    if (!has_measured_speed) {
```
mandeepsandhu, line 608: Why don't we use the predecessor label's `has_measure_speed()` instead of passing-in the value explicitly here (similar to whats done in `TransitionCost()` ?

dnesbitt61: the predecessor EdgeLabel is in reverse direction of actual travel. I imagine the 2 directed edges in opposing directions could have different speeds / measured speed flags.

src/sif/autocost.cc
mandeepsandhu: Yeah, I guess RAD tests will show if this is generating good results for auto or not (which I believe is good, right @gknisely ?)

src/sif/autocost.cc
mandeepsandhu: Any reason why this isn't a pointer like the equivalent param in `GraphTile::GetSpeed()`? Pointer is good if not everyone is expected to use this param (eg in `DynamicCost::EdgeCost` we're creating an unused var because of the need for pass by ref).

src/sif/autocost.cc
mandeepsandhu: ah ok, that makes sense. thx @dnesbitt61

src/sif/motorcyclecost.cc
```diff
@@ -245,12 +246,14 @@ class MotorcycleCost : public DynamicCost {
...
    * @param  edge  the opposing predecessor in the reverse tree
+   * @param  has_measured_speed Do we have any of the measured speed types set?
    * @return  Returns the cost and time (seconds)
    */
   virtual Cost TransitionCostReverse(const uint32_t idx,
                                      const baldr::NodeInfo* node,
                                      const baldr::DirectedEdge* pred,
-                                     const baldr::DirectedEdge* edge) const override;
+                                     const baldr::DirectedEdge* edge,
+                                     const bool has_measured_speed) const override;
```
mandeepsandhu, line 256: Is `has_measured_speed` unused here?

src/sif/motorscootercost.cc
```diff
@@ -285,12 +286,14 @@ class MotorScooterCost : public DynamicCost {
...
    * @param  edge  the opposing predecessor in the reverse tree
+   * @param  has_measured_speed Do we have any of the measured speed types set?
    * @return  Returns the cost and time (seconds)
    */
   virtual Cost TransitionCostReverse(const uint32_t idx,
                                      const baldr::NodeInfo* node,
                                      const baldr::DirectedEdge* pred,
-                                     const baldr::DirectedEdge* edge) const override;
+                                     const baldr::DirectedEdge* edge,
+                                     const bool has_measured_speed) const override;
```
mandeepsandhu, line 296: Is `has_measured_speed` unsed here? If so, maybe make it an unnamed param to avoid compiler warning.

mandeepsandhu: Same comment for all other places where we have this param unused.

src/sif/autocost.cc
gknisely: @mandeepsandhu due to this https://github.com/valhalla/valhalla/issues/2896

src/sif/motorcyclecost.cc
gknisely: @mandeepsandhu we are missing RAD tests for this and it is noted as a TODO

mandeepsandhu: ignore this comment, I didn't realize this was the function declaration.

src/sif/motorscootercost.cc
mandeepsandhu: ignore these too. sorry about the noise!

gknisely: @mandeepsandhu updated

valhalla/sif/dynamiccost.h
```diff
@@ -904,6 +907,10 @@ class DynamicCost {
     c.cost += service_penalty_ *
               (edge->use() == baldr::Use::kServiceRoad && pred->use() != baldr::Use::kServiceRoad);
     // shortest ignores any penalties in favor of path length
+
+    c.cost += service_penalty_ *
```
mandeepsandhu, line 911: Why are we adding `service_penalty_` twice?

valhalla/sif/edgelabel.h
mandeepsandhu: no more spare bits left :slightly_frowning_face:

mandeepsandhu:
> Knowing that the RAD results were good for this then I approve however we need to add more tests!

:100:  we need more tests!

valhalla/sif/dynamiccost.h
gknisely: @mandeepsandhu merge issue.  Thank you for seeing this one

gknisely: updated.

mandeepsandhu: Since this affected the base transition cost, should we RAD test a few routes to see the result is still as expected? Although if this was only due to a recent merge and the RAD results in PR were performed before this change, then we can skip it. I'll leave it to you

mandeepsandhu (approved):
Changes look good. Approving assuming RAD with latest changes look good as well.

src/sif/autocost.cc
PureW: @gknisely Can you copy that code comment and modify if necessary to this code? I`m with @mandeepsandhu here :)

src/thor/bidirectional_astar.cc
genadz:
```c++
kFreeFlowMask | kConstrainedFlowMask | kPredictedFlowMask | kCurrentFlowMask
```
- this is used several times, probably would be better to change it with a new constants ? `kFlowMask`

PureW: Yes, plus one on this

kevinkreiser: we have that already: `kDefaultFlowMask`

valhalla/sif/edgelabel.h
kevinkreiser: there are 17 more spare below. there just arent anymore spare in _this_ 64bit value

src/thor/bidirectional_astar.cc
gknisely: @genadz @PureW @kevinkreiser thank you i missed that one

mandeepsandhu (approved):

PR #2905, opened 2021-02-28
Add edgeinfo overload

merkispavel:
# Issue

The main change is in the `graphtile.h` file: `edgeinfo` overload is added. This is done for convenience to get rid of redundancy in the valhalla code

Not sure if it should be tracked in the changelog as it's minor "refactoring". But I don't mind if it should be there

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/mjolnir/graphfilter.cc
+++ b/src/mjolnir/graphfilter.cc
@@ -143,7 +143,7 @@ void FilterTiles(GraphReader& reader,
         // encoded shape plus way Id.
         bool added;
         uint32_t idx = directededge->edgeinfo_offset();
-        auto edgeinfo = tile->edgeinfo(idx);
+        auto edgeinfo = tile->edgeinfo(directededge);
         std::string encoded_shape = edgeinfo.encoded_shape();
         uint32_t w = hasher(encoded_shape + std::to_string(edgeinfo.wayid()));
         uint32_t edge_info_offset =
--- a/valhalla/baldr/graphtile.h
+++ b/valhalla/baldr/graphtile.h
@@ -327,6 +327,12 @@ public:
   EdgeInfo edgeinfo(const size_t offset) const;
 
   /**
+   * Get a pointer to edge info.
+   * @return  Returns edge info.
+   */
+  EdgeInfo edgeinfo(const DirectedEdge* edge) const;
+
+  /**
    * Get the complex restrictions in the forward or reverse order.
    * @param   forward - do we want the restrictions in reverse order?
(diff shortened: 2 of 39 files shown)


valhalla/baldr/graphtile.h
```diff
@@ -326,6 +326,12 @@ class GraphTile : public boost::intrusive_ref_counter<GraphTile, GraphTileRefCou
    */
   EdgeInfo edgeinfo(const size_t offset) const;
 
+  /**
+   * Get a pointer to edge info.
+   * @return  Returns edge info.
+   */
+  EdgeInfo edgeinfo(const DirectedEdge* edge) const;
```
merkispavel, line 333: here it is

valhalla/baldr/graphtile.h
```diff
@@ -326,6 +326,12 @@ class GraphTile : public boost::intrusive_ref_counter<GraphTile, GraphTileRefCou
    */
   EdgeInfo edgeinfo(const size_t offset) const;
```
kevinkreiser, line 327: are there any places in the code left that use this? i was trying to think of some in my head and my thought was that it happens in mjolnir somewhere but i couldnt come up with the exact location.

kevinkreiser:
despite the number of files changed, i agree its minor and doesnt warrent a changelog entry. looks good to me

kevinkreiser (approved):

valhalla/baldr/graphtile.h
merkispavel: I replaced all places where the old one overload is used(see new commit). Only a few `GraphTile` methods are left that uses this overload: `GraphTile::GetNames` and `GraphTile::GetTypes`
So if it's okay to "break" `GraphTile` api i would do one of the two things
1. Remove this overload and fix ^ 2 methods.
2. Move this overload to the protected/private section

src/mjolnir/graphfilter.cc
```diff
@@ -143,7 +143,7 @@ void FilterTiles(GraphReader& reader,
         // encoded shape plus way Id.
         bool added;
         uint32_t idx = directededge->edgeinfo_offset();
```
kevinkreiser, line 145: this line makes me think that you should do the same with `GraphTile::GetNames`

kevinkreiser (approved):

src/mjolnir/graphfilter.cc
merkispavel: yea, agreed

PR #2910, opened 2021-03-02
Legal max speed parameter

### LINKED ISSUE #2840: Avoid streets entirely with (legal) maxspeed higher than threshold
We recently added `top_speed` to all (motorized) models where roads with higher _estimated_ (not max) speeds are slightly penalized.

However, there are instances where it's mandatory that vehicles don't enter roads which have a legal speed limit above a certain threshold. Apparently that's applicable for electric bikes/other "lower" electric vehicle types in US/CA. In parameter terms this could be `legal_speed` or similar.

Would you be generally okay with a contribution like that?

I can see `speed_limit` is already collected from `maxspeed` OSM tag and put into `EdgeInfo`.

There a few questions:
- I think `maxspeed:forward/backward` should be collected too (?!)
- Would the check go into `DynamicCost::Allowed(Reverse)`? I can't see `EdgeInfo` being used there (or in any other part of the actual path finding), but I'm quite sure you'd not like to extend the `DirectedEdge` interface with that info!?
- Obviously `maxspeed` (and more so their directional siblings) are still fairly rare, so country-/roadclass-specific defaults would be nice to get some better coverage of `speed_limit` (I know it's been on your TODO for a while..). Would you accept a contribution in a first step without that enhancement? I'd be happy to look into it at some point, but I couldn't take care of that right now tbh.

Any thoughts/opinions?

nilsnolde:
fixes #2840 

As described in the issue.. Pretty special use case I'd say, but since I implemented it for a client already, maybe you have some use for it too? The client's use case was e-(cargo-)bikes somewhere in Michigan, which are apparently not allowed on roads which have a max speed higher than ~50 KPH. No idea if that's a common thing anywhere else, but here it goes..

## Tasklist

 - [x] Add tests


lua/graph.lua
```diff
@@ -1460,17 +1460,18 @@ function filter_tags_generic(kv)
   kv["alt_name"] = kv["alt_name"]
   kv["official_name"] = kv["official_name"]
 
-  if kv["maxspeed"] == "none" then
-    --- special case unlimited speed limit (german autobahn)
-    kv["max_speed"] = "unlimited"
-  else
-    kv["max_speed"] = normalize_speed(kv["maxspeed"])
+  -- max speeds are allowed to be unlimited
+  local speeds_tags = { max_speed="maxspeed", forward_speed="maxspeed:forward", backward_speed="maxspeed:backward" }
```
nilsnolde, line 1464: I think it'd be more consistent to also allow the directional variants to be unlimited.. Shouldn't really happen, but I've seen it in at least synthetic OSM data

valhalla/baldr/directededge.h
```diff
@@ -1121,7 +1127,8 @@ class DirectedEdge {
   uint64_t seasonal_ : 1;       // Seasonal access (ex. no access in winter)
   uint64_t deadend_ : 1;        // Leads to a dead-end (no other driveable roads) TODO
   uint64_t bss_connection_ : 1; // Does this lead to(come out from) a bike share station?
-  uint64_t spare4_ : 9;
+  uint64_t max_speed_ : 8;      // speed_ can be maxspeed:advisory or maxspeed:practical
+  uint64_t spare4_ : 1;
```
kevinkreiser, line 1131: oh man... this is really rough. i'm having a hard time making a case for this just because of advisory or practical. i'd almost rather just stop parsing advisory or practical so we can use `speed_` instead.

@dnesbitt61 or @gknisely your thoughts?

nilsnolde: Haha no worries, I figured so! An alternative could be to change the `SpeedType` enum to `kTagDefault`, `kTagPractical`, `kTagMax`, `kClassified`:

https://github.com/valhalla/valhalla/blob/nn_legal_limit/valhalla/baldr/graphconstants.h#L365-L369

Plus the code to parse them from the PBF.. And reverse however `speed` was enhanced after parsing. That way we could at least detect if the tagged speed is from a `maxspeed` tag and not one of the others and not write too low speeds to `max_speed_`..

Or chop default & practical ;) Or a no-merge.. up to you.

src/mjolnir/shortcutbuilder.cc
```diff
@@ -323,6 +324,10 @@ uint32_t ConnectEdges(GraphReader& reader,
   auto const edge_duration_truck = directededge->length() / (truck_speed * kKPHtoMetersPerSec);
   total_truck_duration += edge_duration_truck;
 
+  // update the minimum max speed
+  min_speed_limit =
+      directededge->max_speed() < min_speed_limit ? directededge->max_speed() : min_speed_limit;
```
nilsnolde, line 329: this is bs anyways.. need to think about that again..

dnesbitt61:
I have a hard time approving adding yet another speed to DirectedEdge (speed, free-flow, constrained-flow, truck already exist) but I will defer to others. When speed_type == SpeedType::kTagged isn't the speed value equal to the speed limit? It looks like there are a couple cases where speed is reduces on rough surfaces, but otherwise speed is usually the speed limit.

valhalla/baldr/directededge.h
dnesbitt61: speed_type in DirectedEdge is only 1 bit so this doesn't seem practical.

kevinkreiser: i had this thought too. actually i think we **can** do it by taking a single free bit, shifting it one and oring it with the existing bit. effectively making this speed type field 2 bits wide. it would almost be backwards compatible in that old data would work with the new code. but new data with old code would mean that anything that were advisory or practical speed types would look like categorized speed types. i feel like this is a livable compromise since since they are sparsely populated in the data.

dnesbitt61:
Motor scooter costing has similar limitations but it seems that rather than not allowing on roads exceeding some speed limit, the approach was to penalize roads with speeds above the scooter's top speed. This perhaps could be revisited to use similar logic to e-bikes (which perhaps should be its own costing profile rather than adding yet another condition to bicycle costing).

valhalla/baldr/directededge.h
nilsnolde: > i feel like this is a livable compromise

totally agreed!

nilsnolde:
> When speed_type == SpeedType::kTagged isn't the speed value equal to the speed limit?

 @dnesbitt61 in the case where practical_speed or default_speed are also set on a OSM way. 

> Motor scooter costing has similar limitations

actually we added it to all motorized vehicles a few PRs back. As you say, the logic is penalization, but if you're talking about a legal speed limit I think relevant edges should be absolutely avoided.

I'm hoping their project continues so we could look at developing a proper e-bike profile..

valhalla/baldr/directededge.h
nilsnolde: Shortcuts would also need to know their speed type then and so far that's not recorded.

I guess a shortcut should only get a speed limit assigned if all participating edges had one, otherwise default to 0 as now. If all had one, the speed limit should be the minimum of all edges' maxspeed. Correct me if I'm wrong though.

nilsnolde: Though I'm not sure if the edges making up a shortcut are even available to the routing algos? In the case where a shortcut speed limit would exceed a `legal_speed` one might still legally travel _some_ root edges of the shortcut since they could have higher speed limits. Should be pretty rare and can probably be disregarded, but it'd also help my general understanding..

nilsnolde: Ah forget it.. just found the spot that skips building shortcuts if their speeds are different: 

https://github.com/valhalla/valhalla/blob/d3c4eece264ab4ca846fb373cf75ef14dcd1e1d2/src/mjolnir/shortcutbuilder.cc#L255-L258

dnesbitt61: would driving routes need to be impacted by the max speed logic? If not then it seems unnecessary to check speed limits when creating shortcuts. But then, truck routing (for instance) uses shortcuts but doesn't have a similar check for truck speed here - seems like it should. Problem is the more attributes you use in costing, the less effective shortcuts become.

nilsnolde: Ugh sorry, I'm getting confused here anyways.. With the new approach (2 bit SpeedType) `speed` is what will hold the maxspeed, not `speed_limit` (not accessible to DirectedEdge). So shortcuts are fine actually..

Tbh, I don't really understand why different speeds are a reason to not create a shortcut @dnesbitt61 .. Couldn't the `top_speed' costing work with average speeds on the accumulated shortcut? The same for setting a `speed_` on the final shortcut DirectedEdge, no? Probably I'm missing a crucial step here, but I can't see it.. Supporting a legal speed limit for motorized costings would need some logic like that though, but it could be limited to `maxspeed` tagged ones.

dnesbitt61: I take back my comment on truck routing and truck speed - it doesn't look like truck speed is used when doing truck routing which seems wrong.
Wouldn't any costing that uses some penalty or avoidance when speed is above some threshold need to make sure that shortcuts don't filter out such speeds? A connected set of edges that are 45MPH, then 55MPH, then back to 45MPH would need to keep the 55MPH section separate if any costing penalizes edges above 45MPH. Alternatively, any costing which penalizes or avoids edges by speed might need to avoid using shortcut edges.

nilsnolde: > it doesn't look like truck speed is used when doing truck routing which seems wrong

actually it's used when `GetSpeed()` is called in the costing, there's a (fairly new I think) `is_truck` bool arg. 

> A connected set of edges that are 45MPH, then 55MPH, then back to 45MPH would need to keep the 55MPH section separate if any costing penalizes edges above 45MPH

yeah, of course..

dnesbitt61: been doing only bicycle routing recently and haven't followed recent updates to speeds and their use (or shortcuts for that matter). Sorry for wading into this discussion without full knowledge!

nilsnolde: All good, you cleared up some misconceptions on my side:)

PR #2913, opened 2021-03-04
default whitelist status endpoint [ci skip]

kevinkreiser:
in #2907 we added the status endpoint but forgot to default whitelist in the available service actions in the config. this does that :smile:


kevinkreiser:
i skipped CI for this because its not something that is even tested in CI

danpaz (approved):

PR #2914, opened 2021-03-04
Fix NoRoute for waypoints with low reachability edges

genadz:
# Issue

The main idea and unittest described here https://github.com/valhalla/valhalla/pull/2872 . This PR fixes problem with "no route found" for multipoint requests with intermediate points on isolated (unreachable) islands. In case of fail, we remove all low reachable edge candidates for all intermediate waypoints. It forces routing algorithm to use edges from large network that should be the same with a high probability.

## Tasklist

 - [x] Add tests

## Requirements / Relations


valhalla/loki/search.h
```diff
@@ -13,22 +13,27 @@
 namespace valhalla {
 namespace loki {
 
+struct SearchOptions {
+  uint32_t max_reachability_ = 100;
```
kevinkreiser, line 17: why do we need this and why do we need to store an extra bool on the candidates we find? each location already has the max_reach assigned to it (minimum_reachabilty). cant we just add a member to that class that says:

```c++
bool high_reach() const {
  return outbound >= min_reach && inbound >= min_reach;
}
```

I fail to see why its needed to keep this information and really make any of the changes in loki search when the candidate is part of the location which already knows what the max reach could be and the candidate knows what the actual reach was.

src/thor/route_action.cc
```diff
@@ -489,50 +507,87 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
+
+  // For each pair of locations
+  auto origin = ++correlated.rbegin();
+  while (origin != correlated.rend()) {
+    auto destination = std::prev(origin);
+    if (!route_two_points(origin, destination)) {
+      // if routing failed because an intermediate waypoint was snapped to the low reachability road
+      // (such road lies in a small connectivity component that is not connected to other locations)
+      // we should leave only high reachability candidates and try to route again
+      if (allow_retry && is_through_point(*destination) &&
```
kevinkreiser, line 525: i believe this still wont fix all cases. if you have a multipoint route that has lets say 5 locations, you could get stuck in a local maxima. imagine and island where your first 4 locations are, they all route fine but the 5th one is on the outer larger network. for htis to work you have to actually pop off all the first 3 legs you made with the first 4 low reach locations in order to find a path. though i agree solving n=1 is better than not solving it at all. i just wanted to point out that we could still have failures if we dont solve n=m.

valhalla/loki/search.h
genadz: yeah, I also don't like this too much. but, current solutions assumes that we detect big connectivity components. so, if we got location with already set `minimum_reachability` value that is actually low, in this case `high_reach` will return `true` even in cases when this edge lies in a small component.

genadz: hm, so, probably for route requests we should override `minimum_reachability` with the `max_reachability` value from the config

src/thor/route_action.cc
genadz: seems like we will find a route in your example: if we fail at 4th leg (route from 4th location that lies in a small cc to 5th location from a big cc) we will remove path edges that belong to a small cc and find some route that lies in a larger network. did I miss something ? what is the expected result here ?

genadz: the only case that is not covered here: if we route in an island and have unreachable in one direction waypoints (like oneways that lead to deadends). here all locations will be in a small cc, so we can't retry. Such case requires additional reachability analysis: I think for now it's okay

valhalla/loki/search.h
genadz: @kevinkreiser `outbound, inbound` values are stored in `PathEdge` class but `minumun_reachability` is kept in `PathLocation`. there are 2 possible solutions I see:
1. define `bool high_reach(uint32_t min_inbound_reach, uint32_t min_outbound_reach) const` method in `PathEdge` class and use `min_outbound_reach_,min_inbound_reach_` values from `Location` class to pass to this method;
2. define `bool high_reach(const PathEdge& path_edge) const` method in `PathLocation` class and use it to detect if an edge is high reachable or not

src/thor/route_action.cc
kevinkreiser: > did I miss something 

i think so, if you redo the 4th leg with only large components then you'll have disconnected legs. the first 3 legs will be in the small component and the last one will be out of it. im not sure what havoc that would wreak downstream

valhalla/loki/search.h
kevinkreiser: yep either of those would be good, that is what i was thinking too :+1:

src/thor/route_action.cc
genadz: I remove all low reachable edge candidates for all intermediate locations, not only for last one

genadz: @kevinkreiser don't you think that we should always use `max_reachability` value from the config to set `minimun_reachabilty` filed for all locations in order to ensure
1. consistency;
2. `is_high_reach` will return true only for really big components ? because right now we use `default_reachabilty` parameter to initialize `minimum_reachability` for locations, and it equals to 50

kevinkreiser: the way it works is that per location in the request the user can specify the min reach they want to see on a location before they consider it "highly reachable". if they dont specify this value, it gets filled out from config with a default, and yes that default is 50. in any case the location will have the number of reach which we should consider as highly reachable if an edge meets or exceeds this number. ie `location.minimum_reachability` should be the number the user or our default config thinks is highly reachable.

1. sure it would be more consistent but it would also remove the ability for the user to choose! we cant know what data the user is working with, or what their threshold is.
2. yes im saying we dont need that static high reach value that you hardcoded. the definition of high reach is already existing in the minimum_reachabiility on each location

genadz:
@kevinkreiser as expected, performance hasn't changed.

merkispavel (commented):
I looked through the changes. At first sight I haven't found any bugs but I'll leave 🚢  to be Kevin's responsibility ;) 
p.s seeing this code for the first time, so I might miss some details

test_requests/through_routes.txt
```diff
@@ -4,3 +4,6 @@
...
 -j '{"costing":"auto","locations":[{"lon":-122.42256,"heading":81,"lat":37.75679,"type":"break"},{"lat":37.75863,"lon":-122.41907,"heading":175,"type":"through"},{"lat":37.75543,"lon":-122.41877,"type":"break"}]}'
 -j '{"costing":"auto","locations":[{"lon":-122.42256,"heading":81,"lat":37.75679,"type":"break"},{"lat":37.75863,"lon":-122.41907,"heading":174,"heading_tolerance":60,"type":"through"},{"lat":37.75543,"lon":-122.41877,"type":"break"}]}'
+-j '{"costing":"auto","locations":[{"lat":38.245509,"lon":-85.75095,"type":"break"},{"lat":38.251,"lon":-85.763,"type":"through"},{"lat":38.258075,"lon":-85.744407,"type":"break"}]}'
+-j '{"costing":"auto","locations":[{"lat":52.027886,"lon":5.077448,"type":"break"},{"lat":52.030559,"lon":5.078592,"type":"through"},{"lat":52.030178,"lon":5.080847,"type":"through"},{"lat":52.025977,"lon":5.08614,"type":"break"}]}'
+-j '{"costing":"auto","locations":[{"lat":39.821226,"lon":-84.317642,"type":"break"},{"lat":39.828069,"lon":-84.314242,"type":"through"},{"lat":39.832658,"lon":-84.318762,"type":"through"},{"lat":39.828843,"lon":-84.323137,"type":"break"}]}
```
genadz, line 9: test routes that fail right now

src/baldr/pathlocation.cc
```diff
@@ -61,5 +61,9 @@ bool PathLocation::shares_edges(const PathLocation& other) const {
   return true;
 }
 
+bool PathLocation::is_high_reachable(const PathEdge& edge) const {
```
kevinkreiser, line 64: nit: should be `is_highly_reachable` or `has_high_reach`

src/loki/search.cc
```diff
@@ -407,7 +407,7 @@ struct bin_handler_t {
       auto opposing_edge_id = reader.GetOpposingEdgeId(candidate.edge_id, other_edge, other_tile);
 
       if (other_edge && costing->Allowed(other_edge, other_tile)) {
-        auto reach = get_reach(opposing_edge_id, other_edge);
```
kevinkreiser, line 410: hahaha oops! thanks for removing the shadowing!

test/gurka/test_reachability.cc
```diff
@@ -0,0 +1,51 @@
...
+                                    nullptr, "break_through");
+    gurka::assert::raw::expect_path(result2, {"AB", "BC", "BC"});
+  }
+  {
+    // BD is a oneway leading away from ABC with no escape
+    // input coordinate at D should get snapped to the ABC way
+    // Should give the same result as A->1->C
+    auto result = gurka::do_action(valhalla::Options::route, map, {"A", "D", "C"}, "auto", {}, {},
+                                   nullptr, "break_through");
+    gurka::assert::raw::expect_path(result, {"AB", "AB", "AB", "BC"});
```
kevinkreiser, line 41: why are there three AB edges? im sure there is some detail i just cant remember at the moment... i get why there is one from A to the snap point at or near 1 and that then there is another for the next leg from that point to B but i cant figure out where the 3rd one is coming from at the moment. can you jog my memory?

src/thor/route_action.cc
```diff
@@ -395,38 +427,33 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
     // Get the algorithm type for this location pair
-    auto destination = std::prev(origin);
-    thor::PathAlgorithm* path_algorithm =
-        get_path_algorithm(costing, *origin, *destination, api.options());
+    thor::PathAlgorithm* path_algorithm = get_path_algorithm(costing, *origin, *destination, options);
     path_algorithm->Clear();
     algorithms.push_back(path_algorithm->name());
     LOG_INFO(std::string("algorithm::") + path_algorithm->name());
 
     // TODO: delete this and send all cases to the function above
```
kevinkreiser, line 442: so since we deleted this, then can we delete that function i had lying around above and also delete this comment?

src/thor/route_action.cc
```diff
@@ -80,6 +80,42 @@ void via_discontinuity(
...
+      loc.mutable_path_edges()->SwapElements(0, loc.path_edges_size() - 1);
+    }
+    loc.mutable_path_edges()->RemoveLast();
+  }
+  loc.mutable_filtered_edges()->Clear();
+}
+
 /**
 // removes any edges from the location that aren't connected to it (because of radius)
 void remove_edges(const GraphId& edge_id, valhalla::Location& loc, GraphReader& reader) {
```
kevinkreiser, line 121: delete this function i mean, since you are doing it a different way

valhalla/baldr/pathlocation.h
```diff
@@ -78,6 +78,13 @@ struct PathLocation : public Location {
    */
   bool shares_edges(const PathLocation& other) const;
 
+  /**
+   * Check if path edge has high reachability for this location.
+   * @param edge PathEdge that should be checked
+   * @return     true if the edge is high reachable
+   */
+  bool is_high_reachable(const PathEdge& edge) const;
```
kevinkreiser, line 86: seems this method is actually unused. we instead use the free function inside thor route_action

src/thor/route_action.cc
```diff
@@ -80,6 +80,42 @@ void via_discontinuity(
...
 
+inline bool is_through_point(const valhalla::Location& l) {
+  return l.type() == valhalla::Location::kThrough || l.type() == valhalla::Location::kBreakThrough;
+}
+
+inline bool is_break_point(const valhalla::Location& l) {
+  return l.type() == valhalla::Location::kBreak || l.type() == valhalla::Location::kBreakThrough;
+}
+
+inline bool is_high_reachable(const valhalla::Location& loc,
```
kevinkreiser, line 91: to avoid mistakes, maybe instead of sending the path edge to this function perhaps send the index and get the path edge from the passed in location or at least validate that this location owns this path edge

src/thor/route_action.cc
```diff
@@ -80,6 +80,42 @@ void via_discontinuity(
...
+  int start_idx = std::distance(loc.mutable_path_edges()->begin(), new_end);
+  loc.mutable_path_edges()->DeleteSubrange(start_idx, loc.path_edges_size() - start_idx);
+
+  new_end = std::remove_if(loc.mutable_filtered_edges()->begin(), loc.mutable_filtered_edges()->end(),
+                           [&loc](const auto& pe) { return !is_high_reachable(loc, pe); });
+  start_idx = std::distance(loc.mutable_filtered_edges()->begin(), new_end);
+  loc.mutable_filtered_edges()->DeleteSubrange(start_idx, loc.filtered_edges_size() - start_idx);
+}
+
+inline void remove_all_edges_except_one(valhalla::Location& loc, const GraphId& edgeid) {
```
kevinkreiser, line 109: if its a special case node snap, should allow both the edge ending at the node and the edge beginning at the node?

kevinkreiser: i see this is just copy paste from below with also removing the filtered edges. best to keep it the same to avoid any unforeseen issues

src/thor/route_action.cc
```diff
@@ -455,32 +482,29 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
-        while (destination->type() != valhalla::Location::kBreak &&
-               destination->type() != valhalla::Location::kBreakThrough) {
+        while (!is_break_point(*destination)) {
           throughs.push_back(*destination);
           --destination;
         }
 
         // We have to flip the via indices because we built them in backwards order
-        decltype(vias) flipped;
+        std::remove_reference<decltype(vias)>::type flipped;
```
kevinkreiser, line 494: was `decltype` previously declaring the new variable `flipped` as a reference?

src/thor/route_action.cc
```diff
@@ -395,38 +427,33 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
-      destination->mutable_path_edges()->RemoveLast();
+    if (is_through_point(*destination) && first_edge.Is_Valid()) {
+      remove_all_edges_except_one(*destination, first_edge);
     }
 
     // Get best path and keep it
-    auto temp_paths = get_path(path_algorithm, *origin, *destination, costing, api.options());
+    auto temp_paths = get_path(path_algorithm, *origin, *destination, costing, options);
+    if (temp_paths.empty())
+      return false;
```
kevinkreiser, line 452: this takes the place of the throw that was in `get_path` before hand so you can signal below to retry

src/thor/route_action.cc
```diff
@@ -489,50 +513,87 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
+        }
+        // reset state and try to route again
+        route = nullptr;
+        first_edge = {};
+        vias.clear();
+        path.clear();
+        algorithms.clear();
+        trip.mutable_routes()->Clear();
+        origin = ++correlated.rbegin();
+        continue;
```
kevinkreiser, line 551: resets the entire state of all the legs of the route and starts completely over from the beginning doing all the legs over.

src/thor/route_action.cc
```diff
@@ -489,50 +513,87 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
+
+  // For each pair of locations
+  auto origin = ++correlated.rbegin();
+  while (origin != correlated.rend()) {
+    auto destination = std::prev(origin);
+    if (!route_two_points(origin, destination)) {
+      // if routing failed because an intermediate waypoint was snapped to the low reachability road
+      // (such road lies in a small connectivity component that is not connected to other locations)
+      // we should leave only high reachability candidates and try to route again
+      if (allow_retry && is_through_point(*destination) &&
```
kevinkreiser, line 531: cant this happen on other locations types than `through` though? like even the `break` or `via` types could fail right?

src/thor/route_action.cc
```diff
@@ -489,50 +513,87 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
+          !is_high_reachable(*destination, destination->path_edges(0))) {
```
kevinkreiser, line 532: the reason why `0` works here is only because its a `through` which means we've culled all other edges but the matching ones? what about on the first leg of the route where we havent dont any culling yet, why does `0` always work here?

src/thor/route_action.cc
genadz: it was a compilation error on windows

valhalla/baldr/pathlocation.h
genadz: will remove

src/thor/route_action.cc
genadz: `Location` has two vectors: `path_edges` and `filtered_edges`. So , if we send an index it will be not clear with edges to use

genadz: I'd leave it as is

kevinkreiser: ah! i hadnt noticed that you did this to the filtered edges as well. i guess worrying about accidentally passing the wrong edge location combo isnt so likely so we probably dont need an `assert` or something

src/thor/route_action.cc
kevinkreiser: because you captured it by reference when you moved the code into a lambda. i see

src/thor/route_action.cc
genadz: yeah, make sense. will revert

src/thor/route_action.cc
genadz: is it proposition for comment ? to make more detailed ?

src/thor/route_action.cc
genadz: > i see this is just copy paste from below with also removing the filtered edges. best to keep it the same to avoid any unforeseen issues

agree, will replace with one function.

> if its a special case node snap, should allow both the edge ending at the node and the edge beginning at the node?

why is it important ? if we choose some edge and location has type `through`, we should be able to go through this edge

src/thor/route_action.cc
kevinkreiser: Sorry I wasn't asking to revert I was just adding clarifications to confirm my understanding

src/thor/route_action.cc
kevinkreiser: It's just me clarifying my understanding or anyone else's who is following along haha

src/thor/route_action.cc
genadz: ))) I thought you proposed to leave exception and handle it in retry section

src/thor/route_action.cc
genadz: for `break` and `via` types we use all path edges. it means if we failed to find a route it's definitely not a connectivity problem, so it doesn't make sense to remove low reachable edges and retry

src/thor/route_action.cc
genadz: hm, good question. I though that the first (last) point can't be a `through` type

kevinkreiser: youre right it cant!

genadz: should we handle it somehow ?

src/thor/route_action.cc
kevinkreiser: haha i made the mistake of leaving this comment on the arrive_by which is harder to think about. let me see if i can mentally find a crack in the forward direction (depart_at). i just need to mull it over a bit to see if it handles everything we're worried about sorry for the confusion, just want to really understand the changes to this critical section ;)

src/thor/route_action.cc
```diff
@@ -590,7 +649,50 @@ void thor_worker_t::path_depart_at(Api& api, const std::string& costing) {
...
+  // For each pair of locations
+  auto destination = ++correlated.begin();
+  while (destination != correlated.end()) {
+    auto origin = std::prev(destination);
+    if (!route_two_locations(origin, destination)) {
+      // if routing failed because an intermediate waypoint was snapped to the low reachability road
+      // (such road lies in a small connectivity component that is not connected to other locations)
+      // we should leave only high reachability candidates and try to route again
+      if (allow_retry && is_through_point(*origin) &&
+          !is_high_reachable(*origin, origin->path_edges(0))) {
```
genadz, line 668: I think such condition
```c++
if (allow_retry && origin != correlated.begin() && is_through_point(*origin) && origin->path_edges_size() > 0 &&
std::all_of(origin->path_edges()->begin(), origin->path_edges()->end(), [&origin](const auto& edge) { return !is_highly_reachable(*origin, edge);}))
```

should handle all cases

src/thor/route_action.cc
kevinkreiser: haha sorry about that )

test/gurka/test_reachability.cc
genadz: ha, great question!) actually, the first `AB` edge is reversed. I think there is a small bug in `TimeDepForward` algorithm. I debugged it, there is a combination of conditions in `TimeDepForward::SetOrigin` method that led to this:
1. looks like we should remove all inbound edges in case of the origin was snapped to the node, but we didn't do ti because of this https://github.com/valhalla/valhalla/blob/master/src/thor/timedep_forward.cc#L432 : origin and destination is on the same edge;
2. when we got forward `AB` edge we found that the destination also on this edge, and added forward `AB` edge to the priority queue with cost equal to `A -> 1`, BUT didn't settle this edge because of this comment https://github.com/valhalla/valhalla/blob/master/src/thor/timedep_forward.cc#L530 ;
3. last step, we have edges `ABr` and `ABf` in the queue and sortcost for `ABr` edge is less than sortcost for `ABf`. So, pop edge `ABr` from the queue, start expansion from the node `A` and found destination edge `ABf`. The end)

genadz: I think, this should be fixed in another PR.

src/thor/route_action.cc
```diff
@@ -590,7 +642,52 @@ void thor_worker_t::path_depart_at(Api& api, const std::string& costing) {
...
+  // For each pair of locations
+  auto destination = ++correlated.begin();
+  while (destination != correlated.end()) {
+    auto origin = std::prev(destination);
+    if (!route_two_locations(origin, destination)) {
+      // if routing failed because an intermediate waypoint was snapped to the low reachability road
+      // (such road lies in a small connectivity component that is not connected to other locations)
+      // we should leave only high reachability candidates and try to route again
+      if (allow_retry && origin != correlated.begin() && is_through_point(*origin) &&
+          origin->path_edges_size() > 0 && !is_highly_reachable(*origin, origin->path_edges(0))) {
```
kevinkreiser, line 661: ok so we dont have to solve this in this PR but... it is possible that the router will use a different edge at the end of one leg than the beginning of the next one. so while you were right that we only need to currently care about handling a failure for through locations (because only they widdle down the edges to just the one that was used in the adjacent leg) it is the case that the router could have a situation like i described in my previous comment.

lets go with the simple case: break->break->break

assume that the middle break point has two pairs of edge candidates one inside a small component and one outside. assume also that the first break point is inside the small component. what happens is then the fist leg is in the component and the second leg is outside of it. makign the returned route, invalid. so you are right, it wont fail the route but it will return something unusable to the user. this is a long standing issue and a rare occurance, but i wanted to point it out to you here.

we probably should be removing some of the pairs of candidates even for breaks and vias if they arent directly connected to the last edge in the previous edge. thats i think what i planned to do with that commented out function above but it was so long ago i dont quite remember ))

src/thor/route_action.cc
```diff
@@ -395,38 +417,33 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
+    if (is_through_point(*destination) && first_edge.Is_Valid()) {
+      remove_path_edges(*destination,
+                        [&first_edge](const auto& edge) { return edge.graph_id() != first_edge; });
     }
 
     // Get best path and keep it
-    auto temp_paths = get_path(path_algorithm, *origin, *destination, costing, api.options());
+    auto temp_paths = get_path(path_algorithm, *origin, *destination, costing, options);
+    if (temp_paths.empty())
+      return false;
```
kevinkreiser, line 442: nit: this line isnt covered because your unit test doesnt use date_time type=2 (arrive_by). you could increase coverage by adding another permutation just in case to make sure that the fix works for arrive_by as well.

src/thor/route_action.cc
```diff
@@ -489,50 +503,90 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
+        correlated = options.locations();
+        for (auto loc = std::next(correlated.begin()); loc != std::prev(correlated.end()); ++loc) {
+          remove_path_edges(*loc,
+                            [&loc](const auto& edge) { return !is_highly_reachable(*loc, edge); });
+          // it doesn't make sense to continue if there are no more path edges
+          if (loc->path_edges_size() == 0)
+            // no route found
+            throw valhalla_exception_t{442};
+        }
+        // resets the entire state of all the legs of the route and starts completely
```
kevinkreiser, line 535: all the coverage checks above are the same thing, if you add an arrive_by route that needs a retry then these will all be resolved

test/gurka/test_reachability.cc
```diff
@@ -0,0 +1,51 @@
+#include "gurka.h"
```
kevinkreiser, line 1: nit: there is already a `test_reach` and a `test_route` gurka test, maybe this should just go in one of those?

test/gurka/test_reachability.cc
kevinkreiser: interesting, we have a bit of code that is supposed to do this if i can remember where it is. it should say somethign like if its node snapped and its an origin no point in using the edge that is snapped to its end node...

kevinkreiser: this bit of code: https://github.com/valhalla/valhalla/blob/master/src/thor/timedep_forward.cc#L424-L429

kevinkreiser: the purpose is to ignore end node snaps for origins of the route unless that is all we have

genadz: https://github.com/valhalla/valhalla/blob/master/src/thor/timedep_forward.cc#L432 - we also use this condition that just checks that origin and destination on the same edge. so , probably it's needless

src/thor/route_action.cc
genadz: it's very useful comment, thank you! we should rethink this logic and fix in a separate issue. I'll crate after this one gets merged

genadz:
@kevinkreiser updates tests to check both depart_at and arrive_by attributes.

kevinkreiser (approved):

PR #2915, opened 2021-03-04
Map match snap

ktatterso:
# Issue

I investigated a segfault and landed on this fix.

When map matching we call `CandidateCollector::WithinSquaredDistance()`.  This routine projects input (gps) points onto map/graph edges.  

I think this is what is happening:  very occasionally, under the right geometric conditions, adjacent gps points will project very near a vertex but not quite on it (e.g., 1e-9 m).  I think the logic expects the projections to land in an ordered manner, but in this case the projections are very near one another (again 1e-9 m) and are projecting slightly out-of-order - the latter projecting on the first edge and the former on the second - just barely on either side of their shared vertex.  Picture two gps points being orthogonal to the street edges and projecting on either side of their shared vertex.

To address this I am passing a tolerance of `0.001 m` into helper::Project() which will snap projections to the respective endpoints if they are within the tolerance.

Even if my interpretation of what is happening isn't exactly correct, I'd argue that this is a good health change.  The code that follows the call to `helper::Project()` has these types of conditions:

```
        if (dist == 1.f) {
          snapped_node = edge->endnode();
        } else if (dist == 0.f) {
          snapped_node = opp_edge->endnode();
        }

```

While not well named, ```dist``` represents the percentage along the segment's length where the gps point projected - hence it ranges from [0.0,1.0].  Without the tolerance, projections (like in this case) can land very close to a vertex, resulting in a ```dist == 1e-9```.  Very close to zero - but not quite.  By employing the tolerance, we snap to the vertex and return ```dist == 0.0```.

As a rule, exact floating-point comparisons should be avoided.  However, because I've employed the tolerance, these comparisons can remain.

Why was the code segfaulting?  Without the tolerance fix above, we later call ```cut_segments()``` and land on this line of code:

```
    bool loop = prev_match.edgeid == curr_match.edgeid &&
                prev_match.distance_along > curr_match.distance_along;
    auto last_segment = std::find_if(first_segment + static_cast<size_t>(loop), segments.end(),
                                     [&curr_match](const EdgeSegment& segment) {
                                       return (segment.edgeid == curr_match.edgeid);
                                     });
```
In the segfault scenario, ```prev_match.distance_along==1e-9``` and ```curr_match.distance_along==0.0```.  Very close!  But as a result, ```loop==true``` and we start the search at ```first_segment+1``` instead of ```first_segment```.  Consequently, we don't find a matching ```edgeid``` and ```last_segment``` is the end iterator which is used in an unchecked manner and we segfault.

Refactoring:  I renamed ```MatchResult::distance_along``` to ```MatchResult::percent_along```, since in fact, it is a percentage from [0.0,1.0].  This is more consistent with other parts of valhalla. 

Testing notes:  I was unable to manufacture a scenario (in gurka) that would replicate the subtle geometry that would lead to this issue. Further, pulling the same real world section from both protomaps and geofabrik failed to reproduce this issue. Even pulling a different vintage of the planet data failed to reproduce this issue. To reproduce this issue and author a test case I pulled a single tile (3.6MB) from a specific vintage no-traffic planet tar and based my test on that.  Question for reviewers:  is it worth adding 3.6MB of test data for this single unit test?  I'm torn.

Another testing note:  RAD testing this change isn't meaningful because RAD tests don't test 'trace_route' (which is map matching).

## Tasklist

 - [x] Add tests

## Requirements / Relations


PureW:
> Question for reviewers: is it worth adding 3.6MB of test data for this single unit test? I'm torn.

How large is the file if gzipped? Something I've done in the past is check-in a gzipped test-file for hard-to-mock error-states. The build-tool would then unpack the file before running tests, which would bring down the data checked into git.

PureW:
Thanks for the nice commenting and clarification of the code :+1:

ktatterso:
> > Question for reviewers: is it worth adding 3.6MB of test data for this single unit test? I'm torn.
> 
> How large is the file if gzipped? Something I've done in the past is check-in a gzipped test-file for hard-to-mock error-states. The build-tool would then unpack the file before running tests, which would bring down the data checked into git.

gzipped is 1.36MB

src/meili/candidate_search.cc
```diff
@@ -38,6 +38,13 @@ CandidateCollector::WithinSquaredDistance(const midgard::PointLL& location,
   midgard::projector_t projector(location);
   graph_tile_ptr tile;
 
+  // Employ a snap distance when calling helpers::Project(). Small amounts of
+  // floating-point noise can naturally creep into our computations; snapping
+  // helps keep things well conditioned. What this represents is, "if our point
+  // projection onto a segment is within this distance of either segment end,
+  // just snap to that end."
+  const float snap_tol = 0.001; // 1 mm
```
kevinkreiser, line 46: nit: `constexpr`

src/meili/match_route.cc
```diff
@@ -141,15 +141,21 @@ void cut_segments(const std::vector<MatchResult>& match_results,
...
     }
-    // we want to handle to loop by locating the correct target edge by comparing the distance alone
-    bool loop = prev_match.edgeid == curr_match.edgeid &&
-                prev_match.distance_along > curr_match.distance_along;
+    // we want to handle to loop by locating the correct target edge by comparing the distance alone.
+    // Note: there should be a proper tolerance check to see if the 'percent_along's are close
+    // to one another, but comparing percentages isn't geometrically correct. Some refactoring
+    // would be required if we want to strenghen this condition.
+    bool loop =
+        prev_match.edgeid == curr_match.edgeid && prev_match.percent_along > curr_match.percent_along;
```
kevinkreiser, line 149: i dont understand the comment about correctness and i dont understand the significance of the change. can you elaborate on what this fixes?

src/meili/match_route.cc
```diff
@@ -141,15 +141,21 @@ void cut_segments(const std::vector<MatchResult>& match_results,
...
     // if it is a loop, we start the search after the first edge
     auto last_segment = std::find_if(first_segment + static_cast<size_t>(loop), segments.end(),
                                      [&curr_match](const EdgeSegment& segment) {
                                        return (segment.edgeid == curr_match.edgeid);
                                      });
+
     assert(last_segment != segments.cend());
+    if (last_segment == segments.cend())
+      continue;
```
kevinkreiser, line 158: the assert above is basically saying this should never happen. the continue feels like a band aide. can you explain the purpose. i havent looked at this code since we refactored it last summer so id need to spend a day coming back up to speed

valhalla/meili/match_result.h
```diff
@@ -21,7 +21,7 @@ struct MatchResult {
   // Which edge this match point stays
   baldr::GraphId edgeid;
   // Percentage distance along the edge
-  double distance_along;
+  double percent_along;
```
kevinkreiser, line 24: oh i see you just renamed this

kevinkreiser:
projecting out of order along an edge is allowed. in fact we have a parameter that controls how far in distance they can be out of order before it matters (ie makes the match do a uturn). that parameter is called `interpolation_distance`. This makes me wonder if `cur_match` was an interpolated match and if so perhaps the interpolation code has an issue with setting the distance/percent properly to make sure it occurs after the previous match.

so yeah long story short, can you confirm that the test case you have has interpolation_distance turned on or not? osrm has something called `tidy` which when disabled might be similar to interpolation_distance=0. its possible your are testing with a request like that. if that is the case i would argue that we should indeed see whatever out of order stuff that we get in the input. but then the question becomes how did the two edge ids match. the matcher shouldnt have allowed the two edge ids to match if the current match was at 0% along and the previous was at 0.0000000001% along, it should have forced the match to go around the block (ie make a loop, which is what the code above is detecting).

valhalla/meili/match_result.h
ktatterso: right.  I mean, its a percentage.  :).  Valhalla uses the naming convention 'percent_along' in a number of other places.  same context.

src/meili/match_route.cc
ktatterso: Ah yes.  A couple things going on here.  Remember the nature of the problem leads us to this condition being met.  The assert does fire - but only in a debug build.  As you say, the assert is catching a logic condition that is truly unexpected.  My fix - using ```snap_tol``` when snapping - addresses the problem.  As a result, the assert no longer occurs and you could argue that this condition is no longer needed.

However!  You know how code evolves... an upstream algorithm changes, an unexpected geometry arises... and boom, the unexpected thing starts happening again.  The assert will only "help" in a debug build.  I feel like it's a good practice to be defensive in production builds too, especially if we are avoiding a segfault.  That said, don't get me wrong, I'm not a fan of hyper-defensive code either (you know, checking for null from every call to ```new```).  This just seems like one of those places where having both an assert and a runtime check is appropriate.

src/meili/match_route.cc
ktatterso: right, the rename.

kevinkreiser: i tried to explain a bit below but this check is important. its possible to have a sequence of match points such that the match can go around the block and come back to the same edge but behind the previous match point. that is completely valid. this bit of code detects that and makes sure not to again split the same edge but rather find the next occurance of that edge in the sequence of matched edges.

src/meili/match_route.cc
kevinkreiser: my point is that the assertion is a precondition. this `if` statement directly contradicts that. the assertion is like saying, this cannot be the case if the code up to this point is working properly. but then directly after that you say actually its cool we'll just continue.

you can think of the assertion in this case is like `std::logic_error`

genadz: @ktatterso you're right from the one side: we definitely don't want segfaults. But, in case we continue after `assert` actually we implicitly hide the real problem. so, I'd rather remove these 2 lines or throw an exception instead of `assert`

kevinkreiser: I agree we should throw logic error

PR #2918, opened 2021-03-05
add min_level argument to valhalla_ways_to_edges

nilsnolde:
I wanted to export all edges to way IDs, so I made the tile level configurable, defaulting to local level as before.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/mjolnir/valhalla_ways_to_edges.cc
+++ b/src/mjolnir/valhalla_ways_to_edges.cc
@@ -142,7 +142,8 @@ int main(int argc, char** argv) {
       for (uint32_t n = 0; n < tile->header()->directededgecount(); n++, ++edge_id) {
         const DirectedEdge* edge = tile->directededge(edge_id);
         if (edge->IsTransitLine() || edge->use() == Use::kTransitConnection ||
-            edge->use() == Use::kEgressConnection || edge->use() == Use::kPlatformConnection) {
+            edge->use() == Use::kEgressConnection || edge->use() == Use::kPlatformConnection ||
+            edge->is_shortcut()) {
           continue;
         }
 
--- a/src/mjolnir/valhalla_ways_to_edges.cc
+++ b/src/mjolnir/valhalla_ways_to_edges.cc
@@ -90,8 +90,8 @@ bool ParseArguments(int argc, char* argv[]) {
     if (min_level > max_level) {
       std::cerr << "min_level " << min_level << " exceeds max level of " << max_level << std::endl;
       std::cerr << "\n" << options << std::endl;
+      return false;
     }
-    return false;
   }
 
   if (vm.count("config")) {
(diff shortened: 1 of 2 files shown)


src/mjolnir/valhalla_ways_to_edges.cc
```diff
@@ -44,15 +45,17 @@ bool ParseArguments(int argc, char* argv[]) {
...
       "\n"
       "\n");
 
   options.add_options()("help,h", "Print this help message.")("version,v",
                                                               "Print the version of this software.")(
       "config,c", boost::program_options::value<filesystem::path>(&config_file_path)->required(),
-      "Path to the json configuration file.")
+      "Path to the json configuration file.")(
+      "min_level", boost::program_options::value<uint16_t>(&min_level),
+      "Min level to export: 0 will export all levels, 1 will export all levels but 0 and so on.")
```
kevinkreiser, line 58: Its a bit weird because we say min level but the smaller levels have larger numbers. Honestly i wouldnt be against just changing this thing to default to going through all the levels and just ignoring the break in compatibility but if you think the flexibility is needed maybe we can just parse comma separate level numbers as a string and then validate them as you do below after turning them into an unordered_set of uint8_t. there is a snippet for doing soemthign like this in prime_server that you could re-use if it sounds interesting to you: https://github.com/kevinkreiser/prime_server/blob/master/src/prime_helpers.hpp#L7-L25

if not not worries just a suggestion to reduce confusion

src/mjolnir/valhalla_ways_to_edges.cc
```diff
@@ -115,29 +127,35 @@ int main(int argc, char** argv) {
...
-          edge->use() == Use::kEgressConnection || edge->use() == Use::kPlatformConnection) {
+  for (auto& c = min_level; c <= TileHierarchy::levels().size(); c++) {
+    const auto level = TileHierarchy::levels()[c].level;
+    const auto tiles = TileHierarchy::levels()[c].tiles;
+    for (uint32_t id = 0; id < tiles.TileCount(); id++) {
+      // If tile exists add it to the queue
+      GraphId edge_id(id, level, 0);
+      if (!reader.DoesTileExist(edge_id)) {
         continue;
       }
```
kevinkreiser, line 139:
```suggestion
  for(auto tile_id : reader.GetTileSet() {
```

kevinkreiser, line 139:
```suggestion
for(auto tile_id : reader.GetTileSet() {
```

src/mjolnir/valhalla_ways_to_edges.cc
nilsnolde: Ah no, totally fine for _me_ to drop that sort of flexibility/backwards compatibility.. Maybe someone else has some strong feeling about this? Otherwise I'll drop it..

nilsnolde: Seems no one cares too much, let's do all levels by default then.

kevinkreiser (approved):

kevinkreiser (approved):

kevinkreiser:
trivial, merging

PR #2920, opened 2021-03-05
Cache predicted speeds

merkispavel:
# Issue

While running valhalla_service with perf I noticed that `predicted speed decompressing` takes noticeable amount of time. So the first thing I thought about was to cache it. My testing shows 5%+ performance boost

1) AFAIK Now `GraphTile` is kinda static object. The only thing that you might care about is `GraphTile` related issues: ref_counter and tilecache thread safety. PredictedSpeed caching introduces one more place where you can mess up. I wrapped it with ifdef - let me know if this is enough

2) Would be cool if you double check my performance test results

cc @genadz @kevinkreiser 

## Tasklist

 - [ ] Add tests

## Requirements / Relations


valhalla/baldr/graphtile.h
```diff
@@ -588,7 +588,20 @@ class GraphTile : public boost::intrusive_ref_counter<GraphTile, GraphTileRefCou
...
       uint32_t idx = de - directededges_;
-      float speed = predictedspeeds_.speed(idx, seconds);
+      uint32_t bucket = seconds / kSpeedBucketSizeSeconds;
+#ifdef ENABLE_THREAD_SAFE_TILE_REF_COUNT
+      float speed = predictedspeeds_.speed(idx, bucket);
+#else
+      float speed;
+      {
+        BucketSpeed& entry = predicted_speed_cache[idx];
+        if (entry.last_bucket != bucket) {
```
kevinkreiser, line 598: i wonder why this is showing the performance improvement. why are we computing the speed for the same edge that often? i mean i know we do it at least a couple times, once during the route, once during recosting for shortcuts removal, once during shape attributes (actually not that last one i guess). i assume those are the culprets??

when i was previously looking into how to speed up the predicted speed look up one of the thoughts i had was vectorizing the decompression algorithm. im not exactly sure if it can be done because the types arent exactly the same but SIMD seemed like an obvious avenue for something like inverse dct.

valhalla/baldr/graphtile.h
```diff
@@ -588,7 +588,20 @@ class GraphTile : public boost::intrusive_ref_counter<GraphTile, GraphTileRefCou
...
       seconds %= midgard::kSecondsPerWeek;
       uint32_t idx = de - directededges_;
-      float speed = predictedspeeds_.speed(idx, seconds);
+      uint32_t bucket = seconds / kSpeedBucketSizeSeconds;
+#ifdef ENABLE_THREAD_SAFE_TILE_REF_COUNT
+      float speed = predictedspeeds_.speed(idx, bucket);
+#else
+      float speed;
+      {
+        BucketSpeed& entry = predicted_speed_cache[idx];
```
kevinkreiser, line 597: and yeah as you point out its very unfortunate that we remove the "read-only" nature of graphtile here. if we want to do something like this maybe we need to keep the cache in the graph reader or somehwere else so that we can preserve graphtiles `const` interface without resorting to `mutable`

valhalla/baldr/graphtile.h
merkispavel: > i wonder why this is showing the performance improvement. why are we computing the speed for the same edge that often? i mean i know we do it at least a couple times, once during the route, once during recosting for shortcuts removal, once during shape attributes (actually not that last one i guess). i assume those are the culprets??

Interesting.. The only thing I can think of is we reset edges if we match a restriction https://github.com/valhalla/valhalla/blob/6901099c3f6aea3b930f2a79bc11df694d80e1e4/valhalla/sif/dynamiccost.h#L381 (if I'm not mistaken the edge can be visited more than once in this case).
But  I'll check my requests one more time. There's is a chance that they were close to each other(in sense of distance) so they used the same tiles

kevinkreiser: i guess during the actual routing we will hit the same edge some number of times and call UpdateLabel so maybe that is where most of the savings are

merkispavel:
Closing not to spam open PR list with inactive PRs

PR #2922, opened 2021-03-05
nit: Enables compiler warnings in part of mjolnir module

PureW:


src/mjolnir/graphenhancer.cc
```diff
@@ -831,92 +831,6 @@ bool IsIntersectionInternal(const graph_tile_ptr& start_tile,
   return true;
 }
 
-// Get the Headings for the node.
-void GetHeadings(const graph_tile_ptr& tile, NodeInfo& nodeinfo, uint32_t ntrans) {
```
mandeepsandhu, line 835: were these functions unused?

gknisely: @mandeepsandhu We used to enhance the turnlanes based on if it was an internal edge or not.  That logic was removed and I guess we forgot to remove the functions.

src/mjolnir/graphenhancer.cc
```diff
@@ -1904,7 +1818,7 @@ void GraphEnhancer::Enhance(const boost::property_tree::ptree& pt,
   LOG_DEBUG("Turn Channel Count = " + std::to_string(stats.turnchannelcount));
   LOG_DEBUG("Ramp Count = " + std::to_string(stats.rampcount));
   LOG_DEBUG("Pencil Point Uturn count = " + std::to_string(stats.pencilucount));
-  for (auto density : stats.density_counts) {
+  for ([[maybe_unused]] auto density : stats.density_counts) {
```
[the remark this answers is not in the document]
PureW: It's apparently a c++17 type hint to mark a variable as not used. `density` is not used when the debug statement is a NOOP-statement. I tried a bunch of things here but apparently nothing is implemented by all compilers we use unless you go to c++ 17 extensions.

src/mjolnir/graphenhancer.cc
```diff
@@ -1904,7 +1818,7 @@ void GraphEnhancer::Enhance(const boost::property_tree::ptree& pt,
...
     LOG_DEBUG("Density: " + std::to_string(density));
   }
```
kevinkreiser, line 1823:
```suggestion
#ifdef LOGGING_LEVEL_DEBUG
  for (auto density : stats.density_counts) {
    LOG_DEBUG("Density: " + std::to_string(density));
  }
#endif
```

src/mjolnir/graphenhancer.cc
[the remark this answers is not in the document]
PureW: Fun fact, compiler is already warning about c++17 usage in other places of the code :)

```
../test/parse_request.cc:2240:20: warning: decomposition declarations are a C++17 extension [-Wc++17-extensions]              
  for (const auto& [costing, default_value] : costing_with_defaults) {                                                                 
                   ^~~~~~~~~~~~~~~~~~~~~~~~ 
```

PureW:
@mandeepsandhu @kevinkreiser I addressed the c++17 extension. Should be good to go now.

mandeepsandhu (dismissed):
lgtm

src/mjolnir/graphenhancer.cc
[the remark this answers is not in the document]
kevinkreiser: yes i complained about this one to @genadz if you check the git history :smile: but i only complained once and didnt bother him again

kevinkreiser (changes requested):
sorry to be the bearer of bad news but the windows build is broken. msvc complains with:

    D:\a\1\s\valhalla/mjolnir/osmaccess.h(22): error C7555: use of designated initializers requires at least '/std:c++latest'
    D:\a\1\s\valhalla/mjolnir/osmaccess.h(32): error C7555: use of designated initializers requires at least '/std:c++latest'

valhalla/mjolnir/osmaccess.h
```diff
@@ -19,7 +19,7 @@ struct OSMAccess {
   OSMAccess() {
     memset(this, 0, sizeof(OSMAccess));
     osmwayid_ = 0;
-    attributes_ = {0};
+    attributes_ = {.v = 0};
```
PureW, line 22: This is annoying. I had to tweak this syntax to satisfy gcc but now msvc instead complains :crying_cat_face:

kevinkreiser: i think the constructor can be vastly simplified and work on all platforms:

```c++
  OSMAccess() {
    memset(this, 0, sizeof(OSMAccess));
    osmwayid_ = 0;
    attributes_.v = 0;
  }
```

valhalla/mjolnir/osmnode.h
```diff
@@ -94,7 +94,8 @@ struct OSMNode {
    */
   midgard::PointLL latlng() const {
     // if either coord is borked we return invalid pointll
-    if (lng7_ == -1 || lat7_ == -1)
+    if (lng7_ == std::numeric_limits<uint32_t>::max() ||
+        lat7_ == std::numeric_limits<uint32_t>::max())
```
kevinkreiser, line 98: this is not right! look at the method above, it explicitly sets them to `-1` and if they were out of range they stay `-1`

kevinkreiser: its odd that assignment is fine but comparison is not

PureW: Odd indeed. Updated the assignment as well.

valhalla/midgard/sequence.h
```diff
@@ -433,6 +433,10 @@ template <class T> class sequence {
     return memmap.size() + write_buffer.size();
   }
 
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-copy"
```
kevinkreiser, line 438: could you post the compiler warning for this just so we can reason about what is going on here

PureW: The osx compiler in CI does not recognize `-Wdeprecated-copy` and so bails here. The if-def dance is avoid this on osx. https://app.circleci.com/pipelines/github/valhalla/valhalla/5844/workflows/1a663e84-beef-4ef3-8283-b5017c0d261d/jobs/39679

The warning about the deprecated copy comes from
```
/root/project/src/mjolnir/graphbuilder.cc:257:58: warning: implicitly-declared 'constexpr valhalla::midgard::sequence<valhalla::mjolnir::Node>::iterator::iterator(const valhalla::midgard::sequence<valhalla::mjolnir::Node>::iterator&)' is deprecated [-Wdeprecated-copy]
  257 |           sequence<Node>::iterator element = --nodes.end();
      |                                                          ^
In file included from /root/project/valhalla/baldr/graphreader.h:20,
                 from /root/project/src/mjolnir/graphbuilder.cc:14:
```
https://app.circleci.com/pipelines/github/valhalla/valhalla/5842/workflows/fd3c3901-38a9-490c-880a-06ecbec2ea5e/jobs/39674

kevinkreiser: instead of this `ifdef` dance we could just explicitly implement the copy constructor. would you mind if we tried that out?

kevinkreiser: maybe even this will work?

```c++
iterator(const iterator&) = default;
```

PureW: Seems like that may work :+1:

kevinkreiser (approved):
looks good to me! im glad we could get rid of the complicated workarounds for different platforms

PR #2925, opened 2021-03-08
nit: Link with shared libs

PureW:
Testing CI build times compared to lld in https://github.com/valhalla/valhalla/pull/2813


PureW:
Weird link errors.

PR #2927, opened 2021-03-09
Fix polygon area calculation: use Shoelace formula

merkispavel:
https://en.wikipedia.org/wiki/Shoelace_formula

# Issue

Isochrone api creates polygon away from the origin location. The reason is the broken `polygon_area` function. The area of the small polygon turnes out to be larger than the whole convex-hull-ish polygon. So if you set `denoise = 1` you get the issue itself.

Screen Shot 2021-03-09 at 08 44 26

Screen Shot 2021-03-09 at 08 28 33

## If you want to reproduce

request: `'{"locations":[{"lat":51.2491202,"lon":0}],"costing":"pedestrian","contours":[{"time":20}], "show_locations":true, "polygons":true, "denoise":1}'`

[planet.tar.zip](https://github.com/valhalla/valhalla/files/6106155/planet.tar.zip)

## Tasklist

 - [ ] Add tests

## Requirements / Relations


merkispavel:
I noticed that inner polygons have positive area sometimes. I think this might be the cause of the failed tests. Anyway, investigating..

dnesbitt61:
I think the standard convention is for counter-clockwise orientations to produce positive area and clockwise orientations should yield negative area.  A simple test (rectangle) could be added to test/util_midgard.cc:

```
TEST(UtilMidgard, Area) {
  // Clockwise polygon (rectangle). Should return negative area.
  using poly_t = std::vector<Point2>;
  poly_t poly{ {0, 0}, {0, 20}, {20, 20}, {20, 0}, {0, 0} };
  auto area = polygon_area(poly);
  EXPECT_EQ(area, -400.0);

  // Reverse the polygon (ccw) - should be positive area
  std::reverse(std::begin(poly), std::end(poly));
  area = polygon_area(poly);
  EXPECT_EQ(area, 400.0);
}
```

merkispavel:
> Polygon area (line 552 in midgard/util.cc) I believe should be:
> 
> area += (p1->first * p2->second) - (p1->second * p2->first)
> 
> This should work for any simple polygon.



> I verified that the change described above is correct. I wrote a simple test with a rectangle and the current polygon_area computation fails and the update is correct. Problem is this breaks the isochrone tests.

Right, these formulas are equivalent. I was too lazy to edit more than 2 symbols :) sorry

merkispavel:
> I think the standard convention is for counter-clockwise orientations to produce positive area and clockwise orientations should yield negative area. A simple test (rectangle) could be added to test/util_midgard.cc:
> 
> ```
> TEST(UtilMidgard, Area) {
>   // Clockwise polygon (rectangle). Should return negative area.
>   using poly_t = std::vector<Point2>;
>   poly_t poly{ {0, 0}, {0, 20}, {20, 20}, {20, 0}, {0, 0} };
>   auto area = polygon_area(poly);
>   EXPECT_EQ(area, -400.0);
> 
>   // Reverse the polygon (ccw) - should be positive area
>   std::reverse(std::begin(poly), std::end(poly));
>   area = polygon_area(poly);
>   EXPECT_EQ(area, 400.0);
> }
> ```

Totally agreed! I have more changes locally with unittests + comments, haven't pushed it yet
Moreover, some changes are needed not to lose polygon orientation in the isochrone code. This is done as well(haven't pushed it yet). I just need to figure out whether to push it to this branch or split to ease its review. 
Anyway, I'll ping you when its ready(probably tomorrow)

merkispavel:
Another PR should be merged first in order to fix tests https://github.com/valhalla/valhalla/pull/2932

src/midgard/util.cc
```diff
@@ -543,13 +543,14 @@ x_intercept<GeoPoint<double>>(const GeoPoint<double>&,
...
   typename container_t::value_type::first_type area =
       polygon.back() == polygon.front() ? 0.
-                                        : (polygon.back().first + polygon.front().first) *
-                                              (polygon.back().second + polygon.front().second);
+                                        : (polygon.back().first * polygon.front().second -
+                                           polygon.back().second * polygon.front().first);
   for (auto p1 = polygon.cbegin(), p2 = std::next(polygon.cbegin()); p2 != polygon.cend();
        ++p1, ++p2) {
-    area += (p1->first + p2->first) * (p1->second + p2->second);
+    area += p1->first * p2->second - p1->second * p2->first;
```
merkispavel, line 553: @dnesbitt61 I changed the formula to its classic form not to confuse people in the future

dnesbitt61: thanks!

dnesbitt61 (commented):
Looks good, thanks!

kevinkreiser:
@merkispavel if you could visually inspect the differences in the isochrone unit test and then update the values then we can get CI passing and we can merge it. looks good to me

merkispavel:
@kevinkreiser @dnesbitt61 the tests are fixed. Looks shippable in my view

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
    * FIXED: Allow u-turns at no-access barriers when forced by heading [#2875](https://github.com/valhalla/valhalla/pull/2875)
    * FIXED: Fixed "No route found" error in case of multipoint request with locations near low reachability edges [#2914](https://github.com/valhalla/valhalla/pull/2914)
    * FIXED: Python bindings installation [#2751](https://github.com/valhalla/valhalla/issues/2751)
+   * FIXED: Skip bindings if there's no Python development version [#2893](https://github.com/valhalla/valhalla/pull/2893)
    * FIXED: Skip bindings if there's no Python development version [#2893](https://github.com/valhalla/valhalla/pull/2878)
```
kevinkreiser, line 27: this seems wrong :smile:

merkispavel: damn :)

merkispavel: The wrong link was used before so automerge decided to keep both entries. Fixed, thanks

valhalla/midgard/gridded_data.h
```diff
@@ -98,7 +98,7 @@ template <std::size_t dimensions_t> class GriddedData : public Tiles<PointLL> {
     typename PointLL::first_type s[5]; // Values at the tile corners and center
```
kevinkreiser, line 98: why is the code from https://github.com/valhalla/valhalla/pull/2932 in this PR as well? something seems messed up here?

merkispavel: Originally i found 2 issues at once(polygon area bug and orientation bug) so I decided to make a hierarchy `master <- 2927 <- 2932` for some reason 🤦(probably because of the unit tests)  Now I think It brought more mess than clearness. Anyway, I had to fix unit tests so it didn't bring any profit
Would you like to split them @kevinkreiser ? (I guess I do, it'll just take a little more time and a few more approves)

kevinkreiser: oh no, no problem at all, i just missed a very important detail:

image

the previous one was merged into this branch not master. i completely missed that. this is fine!

valhalla/midgard/gridded_data.h
```diff
@@ -115,55 +115,77 @@ template <std::size_t dimensions_t> class GriddedData : public Tiles<PointLL> {
...
+    // bounds, i.e., exterior rings are counterclockwise, and holes are clockwise."  (c)
+    // (c) https://tools.ietf.org/html/rfc7946#section-3.1.6
+    bool swap_table[3][3][3] = {
+        {{false, false, true}, {false, true, true}, {true, false, false}},
+        {{false, true, false}, {true, false, false}, {true, false, false}},
+        {{true, true, false}, {false, false, false}, {false, false, false}},
+    };
     std::array<std::function<void()>, 10> cases{
-        []() {},
+        [&]() { from_pt = to_pt = {}; },
```
kevinkreiser, line 144: do you want to change that back to a no-op?

kevinkreiser (dismissed):
LGTM, previous review was the difficult part. fixing the area calculation was straight forward

kevinkreiser (approved):

PureW (approved):

PR #2931, opened 2021-03-11
replace the python version finding mess with cmake builtins

### LINKED ISSUE #2929: Broken Python Installation
unfortunately it seems that #2752 broke the python installation. in docker when i test this and do `make install` i get the following:

```
root@e611d5dd51ad:/valhalla/build# find /usr/local/lib/python*
/usr/local/lib/python.
/usr/local/lib/python./dist-packages
/usr/local/lib/python./dist-packages/valhalla
/usr/local/lib/python./dist-packages/valhalla/__init__.py
/usr/local/lib/python./dist-packages/valhalla/python_valhalla.cpython-38-x86_64-linux-gnu.so
/usr/local/lib/python2.7
/usr/local/lib/python2.7/site-packages
/usr/local/lib/python2.7/dist-packages
/usr/local/lib/python3.8
/usr/local/lib/python3.8/dist-packages
```

wtf is `python.` seems like somewhere something is trying to figure out the python major and minor version but doesnt have it? so it becomes `python${major}.{$minor}` where both variables are empty.. this is breaking the docker build at the moment so im going to work around it and we'll need to debug and fix whatever happened here

cc @nilsnolde

nilsnolde:
fixes #2929 

it uses the "new" CMake builtin approach to get the right python lib path from the actual python installation used to build the bindings. Meaning it'll install now to `/usr/lib/python3/dist-packages/valhalla` instead of `/usr/local...`, but that's actually better IMO

### DIFF
--- a/docker/Dockerfile-run
+++ b/docker/Dockerfile-run
@@ -30,7 +30,7 @@ RUN rm -f valhalla_*.debug
 RUN strip --strip-debug --strip-unneeded valhalla_* || true
 RUN strip /usr/local/lib/libvalhalla.a
 # TODO: fix https://github.com/valhalla/valhalla/issues/2929 and update this line
-#RUN strip /usr/local/lib/python3.8/dist-packages/python_valhalla.cpython-38-x86_64-linux-gnu.so
+RUN strip /usr/lib/python3/dist-packages/valhalla/python_valhalla.cpython-38-x86_64-linux-gnu.so
 
 ####################################################################
 # copy the important stuff from the build stage to the runner image
--- a/docker/Dockerfile-run
+++ b/docker/Dockerfile-run
@@ -29,7 +29,6 @@ RUN tar -cvf valhalla.debug.tar valhalla_*.debug && gzip -9 valhalla.debug.tar
 RUN rm -f valhalla_*.debug
 RUN strip --strip-debug --strip-unneeded valhalla_* || true
 RUN strip /usr/local/lib/libvalhalla.a
-# TODO: fix https://github.com/valhalla/valhalla/issues/2929 and update this line
 RUN strip /usr/lib/python3/dist-packages/valhalla/python_valhalla.cpython-38-x86_64-linux-gnu.so
 
 ####################################################################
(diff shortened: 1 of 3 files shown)


docker/Dockerfile-run
```diff
@@ -30,7 +30,7 @@ RUN rm -f valhalla_*.debug
 RUN strip --strip-debug --strip-unneeded valhalla_* || true
 RUN strip /usr/local/lib/libvalhalla.a
 # TODO: fix https://github.com/valhalla/valhalla/issues/2929 and update this line
```
kevinkreiser, line 32:
```suggestion
```

kevinkreiser:
i kind of prefer the `/usr/local` prefix as that is the standard location to install when you build something from source. ie i think its the canonical location for from source builds. but at the same time i like the extreme simplification in the cmakefile. maybe we can merge this can get the /usr/local back later

nilsnolde:
Well, yeah I agree for valhalla, but if you think about it more, it actually makes sense to let the python framework you're using to build the bindings tell you where to install them. It installs to /usr/local if your python installation is also at /usr/local. And if you force the install to /usr/local, a regular python won't be able to find it by default, you'd have to alter the `sys.path`. An analogy would be if someone wanted to build a valhalla add-on, but strictly installs it to /usr instead of letting the already (wherever) installed valhalla decide where to put the plugin so it can even find it. Plus the really nice simplification:)

kevinkreiser:
i still think `/usr/local` is quite standard:

```
Python 3.6.9 (default, Jan 26 2021, 15:33:00) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
Could not open PYTHONSTARTUP
FileNotFoundError: [Errno 2] No such file or directory: '/home/kkreiser/.pythonrc'
>>> import sys
>>> print(sys.path)
['', '/usr/lib/python36.zip', '/usr/lib/python3.6', '/usr/lib/python3.6/lib-dynload', '/usr/local/lib/python3.6/dist-packages', '/usr/lib/python3/dist-packages']
```

but as i mentioned im totally cool with tackling that later

kevinkreiser (approved):

nilsnolde:
Huh, you're right, thought I checked it on ubuntu. Maybe I did it on my arch, where /usr/local is not in the sys.path (or in the default linker paths, guess they go together).. I have less strong feelings about it in that case:)

PR #2932, opened 2021-03-11
Orient segments/rings according to right-hand rule

merkispavel:
# Issue

At the moment we "forget" about segment orientation and it messes up signed area of the rings. This PR fixes such behaviour

"A linear ring MUST follow the right-hand rule with respect to the
      area it bounds, i.e., exterior rings are counterclockwise, and
      holes are clockwise." (c) https://tools.ietf.org/html/rfc7946#section-3.1.6

## Tasklist

 - [x] Add tests

## Requirements / Relations


test/isochrone.cc
```diff
@@ -101,23 +100,23 @@ TEST(Isochrones, Basic) {
...
...5.141321,52.106324],[5.140321,52.108765],[5.133321,52.105371],[5.134213,52.106937],[5.131321,52.107349],[5.129321,52.104532],[5.115321,52.106504],[5.111321,52.104524],[5.110531,52.108147],[5.107983,52.107599],[5.107321,52.109955],[5.103321,52.106677],[5.100321,52.107739],[5.097321,52.105768],[5.09847,52.109086],[5.094321,52.107837],[5.092321,52.110126],[5.080321,52.112602],[5.081321,52.107534],[5.084321,52.106528],[5.087321,52.107447],[5.089651,52.105607],[5.085321,52.103602],[5.082408,52.106024],[5.077321,52.106621],[5.077321,52.104335],[5.081656,52.102272],[5.082616,52.099641],[5.078321,52.10046],[5.077321,52.098424],[5.071321,52.09879],[5.070321,52.097363],[5.055535,52.110937],[5.059321,52.112586],[5.063239,52.110019],[5.057779,52.115937],[5.058321,52.117587],[5.062832,52.117937],[5.062321,52.119025],[5.057532,52.119937],[5.058321,52.118287],[5.05593,52.119546],[5.047291,52.116937],[5.049872,52.114937],[5.055683,52.115299],[5.053596,52.111662],[5.051321,52.111621],[5.038321,52.116248],[5.035835,52.118451],[5.037321,52.121382],[5.039321,52.119996],[5.041321,52.120514],[5.038507,52.122937],[5.040321,52.125974]],"type":"LineString"},"type":"Feature"}],"type":"FeatureCollection"})";
```
merkispavel, line 103: I checked iou metric(intersection over union) of the polygons: it is about 0.9999. 
In details a few points shifted slightly(1e-5-1e-6 delta) so I just set new `expected_value`

merkispavel, line 103: Changed order according to the right-hand rule

merkispavel, line 103: I checked iou metric(intersection over union) of the polygons: it is about 0.9999. 
In details a few points shifted slightly(1e-5-1e-6 delta) so I just set new `expected_value`

valhalla/midgard/gridded_data.h
```diff
@@ -115,55 +115,77 @@ template <std::size_t dimensions_t> class GriddedData : public Tiles<PointLL> {
     // function stored in the array
     int case_table[3][3][3] = {
         {{0, 0, 8}, {0, 2, 5}, {7, 6, 9}},
-        {{0, 3, 4}, {1, 3, 1}, {4, 3, 0}},
+        {{0, 3, 4}, {1, 0, 1}, {4, 3, 0}},
```
merkispavel, line 118: this is the case when all three vertexes lie on the contour level(so sh[] == 0). As segment orientation is introduced, it's unclear how to orient the segment in this case. It looks safe just to omit such case(when all sh[] == 0) so i did it.

The picture from the original source also relates this case to the "do-nothing" case. (see "g" case) 
http://paulbourke.net/papers/conrec/
conrec3

valhalla/midgard/gridded_data.h
```diff
@@ -395,15 +390,18 @@ template <std::size_t dimensions_t> class GriddedData : public Tiles<PointLL> {
...
+        //          line.reverse();
+        //        }
         // sampling the bottom left corner means everything is skewed, so unskew it
         for (auto& coord : line) {
           coord.first += h;
           coord.second += h;
         }
       }
+      // remove points and lines
+      contour.remove_if([](const contour_t& line) { return line.size() < 4; });
```
merkispavel, line 399: This has nothing to with current PR, just sanity improvement.
When `generalize` != 0 some of the inner holes turn into lines and points. I think we should remove them 
Screen Shot 2021-03-11 at 13 39 21

valhalla/midgard/gridded_data.h
```diff
@@ -394,16 +389,15 @@ template <std::size_t dimensions_t> class GriddedData : public Tiles<PointLL> {
         if (gen_factor > 0.f) {
           Polyline2<PointLL>::Generalize(line, gen_factor, {});
         }
-        // if this ends up as an inner we'll undo this later
```
merkispavel, line 397: not sure what "we'll undo this later" means but now everything works as expected

merkispavel:
@kevinkreiser @dnesbitt61 may I ask for a review? 🙏

valhalla/midgard/gridded_data.h
kevinkreiser: agree

valhalla/midgard/gridded_data.h
```diff
@@ -181,8 +203,10 @@ template <std::size_t dimensions_t> class GriddedData : public Tiles<PointLL> {
     contours_t contours(intervals.size(), std::list<feature_t>{feature_t{}});
 
     // and something to find them quickly
-    using contour_lookup_t = std::unordered_map<PointLL, typename feature_t::iterator>;
-    std::vector<contour_lookup_t> lookups(intervals.size());
+    using contour_lookup_t = std::map<PointLL, typename feature_t::iterator>;
+    // store begins and ends of the segments separately not to loose segment orientation
+    std::vector<contour_lookup_t> begin_lookups(intervals.size());
+    std::vector<contour_lookup_t> end_lookups(intervals.size());
```
kevinkreiser, line 209: we should quantify the performance loss due to these changes. i suppose we hope that its mostly outweighed by the actual expansion but in practice, i semi-recently refactored a bunch of this stuff to win back 10% performance in this area of the code (using the lambdas above)

merkispavel: > I suppose we hope that its mostly outweighed by the actual expansion

I did some performance testing recently. Indeed, it's different orders of magnitude(i.e expansion takes > 90% comparing to non-expansion steps)

kevinkreiser: can you post the results here. its important to know the changes in performance characteristics before we merge such changes imho

merkispavel: I tested performance with `costing=auto&contour=60min` requests. I can barely see the difference in performance: it's close to 1-2% so looks mergeable to me

kevinkreiser: Excellent, that sounds reasonable, thanks!

valhalla/midgard/gridded_data.h
```diff
@@ -115,55 +115,77 @@ template <std::size_t dimensions_t> class GriddedData : public Tiles<PointLL> {
...
+    // bounds, i.e., exterior rings are counterclockwise, and holes are clockwise."  (c)
+    // (c) https://tools.ietf.org/html/rfc7946#section-3.1.6
+    bool swap_table[3][3][3] = {
+        {{false, false, true}, {false, true, true}, {true, false, false}},
+        {{false, true, false}, {true, false, false}, {true, false, false}},
+        {{true, true, false}, {false, false, false}, {false, false, false}},
+    };
     std::array<std::function<void()>, 10> cases{
-        []() {},
+        [&]() { from_pt = to_pt = {}; },
```
kevinkreiser, line 144: this is just to be more explicit i guess, since it will be skipped later?

valhalla/midgard/gridded_data.h
kevinkreiser: it was so long ago i cant remember what i was thinking about there :frowning_face:

kevinkreiser: maybe it was a TODO and then I did the todo? but didnt remove the comment? sorry about that

kevinkreiser (approved):

valhalla/midgard/gridded_data.h
merkispavel: ah damn, I didn't notice it  :) https://github.com/valhalla/valhalla/blob/99893c90aa80857739bb812050439be1110fee22/valhalla/midgard/gridded_data.h#L270

I guess we can leave this change, anyway it doesn't change anything(as you noticed)

Issue #2919, opened 2021-03-05
Ferry edges sometimes not connected to high-level road network


SvetlanaBayarovich:
As I was looking at the ferries that router avoids and one it eventually uses, I noticed that ferry routes are high-level edges but sometimes they are not connected to the rest of high-level network. If I understand correctly the purpose of `ReclassifyFerryConnections` it should connect ferry edge to that network , but somehow it does not happen. As a result, by the moment such ferry is reached, the limit for hierarchy transitions (down) is exhausted, which results in either attempts to find a route using "connected" ferries or no route at all. 

@dnesbitt61 Could you comment on intended behavior for ferry connections?

Below are some examples that illustrate ferry connections. 

Legend: **red** denotes L0 edges, **blue** - L1, **orange** - L2

The ferries that are (eventually) used. They are all connected to L0 roads (and later L1 as well)

| West Sicily | Alicante Spain |Palma de Mallorca |
|--|--|--|
|![image](https://user-images.githubusercontent.com/50918075/110137294-bd2eb300-7de1-11eb-9299-58a3a5358154.png)|![image](https://user-images.githubusercontent.com/50918075/110137848-5d84d780-7de2-11eb-9375-2965c685ea04.png)|![image](https://user-images.githubusercontent.com/50918075/110138261-d5530200-7de2-11eb-83c9-da885a002868.png)|

The ferries that are avoided
| Messina, Sicily | Melilla, North of Africa | Denia, Spain |
|-|-|-|
|![image](https://user-images.githubusercontent.com/50918075/110149950-b444de00-7def-11eb-9aea-90107cfa1f23.png)| ![image](https://user-images.githubusercontent.com/50918075/110150553-7300fe00-7df0-11eb-882a-59af35759234.png)|![image](https://user-images.githubusercontent.com/50918075/110150893-d68b2b80-7df0-11eb-8dc8-c957940a4dc1.png)|
||![image](https://user-images.githubusercontent.com/50918075/110150375-4220c900-7df0-11eb-9396-3757a4a8cc11.png)||

dnesbitt61:
In many locations, ferries connect to highway=service ways (which are the lowest road_class). The intent of ReclassifyFerryConnections is to create the shortest path from the terminal node of a ferry to a high class road. It has been quite a while since this code was written and last tested so I cannot verify its current behavior.

SvetlanaBayarovich:
[...]

PR #2933, opened 2021-03-11
Use kServiceRoad edges while reclassifying ferry connections

SvetlanaBayarovich:
# Issue

At some point the enum value `Use::kServiceRoad` was added but the code part that looks for ferry connection wasn't updated
Partially solves ferry connection problems described in https://github.com/valhalla/valhalla/issues/2919

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/src/mjolnir/ferry_connections.cc
+++ b/src/mjolnir/ferry_connections.cc
@@ -119,7 +119,7 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
 
       // Skip uses other than road / other (service?)
       const OSMWay w = *ways[edge.wayindex_];
-      if (w.use() != baldr::Use::kOther &&
+      if (w.use() != baldr::Use::kOther && w.use() != baldr::Use::kServiceRoad &&
           static_cast<int>(w.use()) > static_cast<int>(baldr::Use::kTurnChannel)) {
         continue;
       }
--- a/test/gurka/test_ferry_connections.cc
+++ b/test/gurka/test_ferry_connections.cc
@@ -1,0 +1,74 @@
+#include "gurka.h"
+#include <gtest/gtest.h>
+
+using namespace valhalla;
+
+baldr::RoadClass
+ReclassifyFerryConnectionEdge(std::map<std::string, std::string> const& way_description) {
+  constexpr double gridsize_metres = 1000;
+
+  const std::string ascii_map = R"(
+          A--B--b--C----D--E
+    )";
+
+  const gurka::ways ways = {{"AB", {{"highway", "trunk"}, {"name", ""}}},
+                            {"Bb", way_description},
+                            {"bC", way_description},
+                            {"CD",
+                             {{"motor_vehicle", "yes"},
+                              {"motorcar", "yes"},
+                              {"bicycle", "yes"},
+                              {"foot", "yes"},
+                              {"horse", "no"},
+                              {"duration", "01:25"},
+                              {"route", "ferry"},
+                              {"operator", "Cape May-Lewes Ferry"},
+                              {"name", "Cape May-Lewes Ferry"},
+                              {"destination:forward", "Cape May"},
+                              {"destination:backward", "Lewes"}}},
+                            {"DE", {{"highway", "trunk"}, {"name", ""}}}};
+
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize_metres);
+
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_reclassify_ferry_connections");
+  baldr::GraphReader graph_reader(map.config.get_child("mjolnir"));
+
+  auto edge = std::get<1>(gurka::findEdgeByNodes(graph_reader, layout, "B", "b"));
+  return edge->classification();
+}
+
+TEST(Standalone, ReclassifyFerryConnection) {
+  // for these values of 'highway' tag edge class is upgraded in order to connect ferry to a
+  // high-class road
+  const std::vector<std::string> reclassifiable_ways = {"secondary",      "tertiary",
+                                                        "unclassified",   "service",
+                                                        "secondary_link", "tertiary_link"};
+  for (const auto& cls : reclassifiable_ways) {
+    std::map<std::string, std::string> desc = {{"highway", cls}, {"name", ""}};
+    auto edge_class = ReclassifyFerryConnectionEdge(desc);
+    EXPECT_EQ(edge_class, baldr::RoadClass::kPrimary);
+  }
+}
+
+TEST(Standalone, DoNotReclassifyFerryConnection) {
+  // roads with these values of 'highway' tag do not participate in search for ferry connection so
+  // edge class remains low
+  const std::vector<std::string> not_reclassifiable_ways = {"track", "living_street"};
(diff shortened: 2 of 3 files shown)


dnesbitt61:
good catch! Seems that I messed this up when I added the generic service use.

SvetlanaBayarovich:
Yeah, this is one of the issues with ferry connections. Probably the most influential one, but definitely not the only one 😞

SvetlanaBayarovich:
Some pictures to compare
| Before | After |
|-|-|
|image||
|||

src/mjolnir/ferry_connections.cc
```diff
@@ -119,7 +119,7 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
 
       // Skip uses other than road / other (service?)
       const OSMWay w = *ways[edge.wayindex_];
-      if (w.use() != baldr::Use::kOther &&
+      if (w.use() != baldr::Use::kOther && w.use() != baldr::Use::kServiceRoad &&
```
PureW, line 122: Is there any gurka regression test that can be added for this?

dnesbitt61, line 122: Is Use::kOther needed now that this explicitly looks for service road use?

dnesbitt61 (commented):
With the sac_scale PR (which should soon be merged as it was just approved) I added a gurka test for trace_attributes. I think one could create a simple test with service roads leading to a ferry, then use the trace_attributes test and check that road_class is set appropriately.

src/mjolnir/ferry_connections.cc
gknisely: @SvetlanaBayarovich we should check on this.  looking at https://github.com/valhalla/valhalla/blob/master/lua/graph.lua#L1332 if not one of the services above and it is not empty it will be set to other

dnesbitt61: But this code above:

```
elseif use == nil and kv["highway"] == "service" then
        use = 11
```

is what makes highway=service (without the explicit service=... in the use method) have use = 11 (kServiceRoad).

SvetlanaBayarovich: If I understand correctly, the whole picture is like 
```
  if kv["highway"] then
     if
     ...
     elseif use == nil and kv["highway"] == "service" then
        use = 11
     ...    
     end
  end

  if use == nil and kv["service"] then
    use = 40 --other
  ...
```
So we only get assigned use = 40 if there is a `service` tag but no `highway` in the item (way?) being parsed. I don't have a feeling if it's possible for the OSM data. On one hand, looks like any combination of tags is theoretically possible, but maybe there are some constraints that I'm not aware of.

@gknisely  @dnesbitt61 Is there a way to understand that? My only idea at the moment is to run over all the edges created and check if use = 40 was ever assigned

SvetlanaBayarovich: But still such check guarantees nothing for future

dnesbitt61: I do not know if this tagging combination happens. Perhaps taginfo could reveal if this combination is possible? I know that service=driveway says that highway=service is required but I don't know if there is anything that prevents a service=XXXX without highway=service. It probably can't hurt to keep the Use::kOther for cases that don't fall into any of the existing service uses (generic, alley, driveway, parking_aisle).

Do you think that Use::kServiceRoad is missing on some edges leading to ferries?

SvetlanaBayarovich: I'm not sure I got your question...
But I have a related question. There is a case when one can get to a ferry only if one uses a drive-through edge. And it obviously gets Use::kDriveThru. Currently such edges are also excluded from connection search. Is it legal to use such label here at all?
image

dnesbitt61: What I was trying to ask is whether you were seeing some roads that were not properly marked as Use::kServiceRoad leading to a ferry. Sounds like you found a case with highway:service and service:drive-through that should be reclassified as it leads to a ferry.

https://wiki.openstreetmap.org/wiki/Tag:service%3Ddrive-through - That seems to indicate that perhaps this other service road type needs to be considered when reclassifying edges leading to ferries. I imagine this is a case where "where customers can visit a business without leaving their cars." (the business is the ferry itself).

SvetlanaBayarovich: Hmm, maybe I'd rather have this change as a separate update in order not to slow down this pr which actually only brings back what was already there, as I think it'll take some time to test

dnesbitt61: makes sense to me

test/gurka/test_ferry_connections.cc
```diff
@@ -0,0 +1,74 @@
...
+
+  const gurka::ways ways = {{"AB", {{"highway", "trunk"}, {"name", ""}}},
+                            {"Bb", way_description},
+                            {"bC", way_description},
+                            {"CD",
+                             {{"motor_vehicle", "yes"},
+                              {"motorcar", "yes"},
+                              {"bicycle", "yes"},
+                              {"foot", "yes"},
+                              {"horse", "no"},
```
PureW, line 22: :D

test/gurka/test_ferry_connections.cc
```diff
@@ -0,0 +1,74 @@
...
+}
+
+TEST(Standalone, DoNotReclassifyFerryConnection) {
+  // roads with these values of 'highway' tag do not participate in search for ferry connection so
+  // edge class remains low
+  const std::vector<std::string> not_reclassifiable_ways = {"track", "living_street"};
+  for (const auto& cls : not_reclassifiable_ways) {
+    std::map<std::string, std::string> desc = {{"highway", cls}, {"name", ""}};
+    auto edge_class = ReclassifyFerryConnectionEdge(desc);
+    EXPECT_GE(edge_class, baldr::RoadClass::kPrimary);
```
PureW, line 60: Is `EXPECT_GE` (>=) correct or would you want `EXPECT_GT` (>)?

test/gurka/test_ferry_connections.cc
```diff
@@ -0,0 +1,74 @@
...
+  // roads with these values of 'service' tag do not participate in search for ferry connection so
+  // edge class remains low
+  const std::vector<std::string> not_reclassifiable_use = {"parking_aisle", "driveway", "alley",
+                                                           "emergency_access", "drive-through"};
+  for (const auto& use : not_reclassifiable_use) {
+    std::map<std::string, std::string> desc = {{"highway", "tertiary"},
+                                               {"service", use},
+                                               {"name", ""}};
+    auto edge_class = ReclassifyFerryConnectionEdge(desc);
+    EXPECT_GE(edge_class, baldr::RoadClass::kPrimary);
```
PureW, line 72: Some question here

test/gurka/test_ferry_connections.cc
SvetlanaBayarovich: Sure! It was `_NE` first, but then I realized it would be better to have gt, but messed up :(. Fixed now

PureW (commented):
LGTM, I'll leave final approve to @gknisely or @dnesbitt61

gknisely (approved):

PR #1960, opened 2019-10-04
updated private access logic.

gknisely:
toss way where access=private and highway=service and service != driveway

# Issue

Seeing issues where we are routing on roads where access is private.  For example, service roads at airports.  Logic tosses them if access=private and highway=service and service != driveway.

## Before

image

## After

image

## Tasklist

 - [x] Review - you must request approval to merge any PR to master
 - [x] Generally use squash merge to rebase and clean comments before merging
 - [x] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #2934, opened 2021-03-11
Revert tossing of private roads

gknisely:
# Issue

The tossing of private roads in https://github.com/valhalla/valhalla/pull/1960 was too aggressive and resulted in a lot of no routes.  This PR reverts that logic and a test was added.   Internally, we will have folks check the access around airports, bus terminals, etc.

## Tasklist

 - [x] Add tests

## Requirements / Relations


test/gurka/test_route_on_private_access.cc
```diff
@@ -0,0 +1,53 @@
...
+using namespace valhalla;
+
+const std::unordered_map<std::string, std::string> build_config{
+    {"mjolnir.admin", {VALHALLA_SOURCE_DIR "test/data/netherlands_admin.sqlite"}}};
+
+const std::vector<std::string>& costing = {"auto",          "hov",        "taxi",
+                                           "bus",           "truck",      "bicycle",
+                                           "motor_scooter", "motorcycle", "pedestrian"};
+
+TEST(Standalone, AvoidPencilPointUturns) {
```
dgearhart, line 17: Forgot to rename AvoidPencilPointUturns

gknisely: @dgearhart thanks.  fixed.

dgearhart (approved):
LGTM :shipit:

abhi2039:
@gknisely What do you mean by  "Internally, we will have folks check the access around airports, bus terminals, etc."? How do you avoid situations like these: https://www.cnbc.com/2013/09/26/apple-maps-app-sends-drivers-to-alaska-airport-runway.html

@dgearhart any thoughts on this?

danpat:
@abhi2039 That kind of problem is an interesting intersection between geocoding and routing.  In many situations, it can be solved by ensuring that the target routing point that you get from your geocoding solution puts you at the desired terminal dropoff point, not the middle of the runway.

Issue #2923, opened 2021-03-07
Build failure on non-x86*: "comparison is always true due to limited range of data type [-Werror=type-limits]"


PureTryOut:
```
In file included from /builds/PureTryOut/aports/testing/valhalla/src/valhalla-3.1.0_git20210304/valhalla/baldr/graphid.h:11,
                 from /builds/PureTryOut/aports/testing/valhalla/src/valhalla-3.1.0_git20210304/valhalla/baldr/directededge.h:6,
                 from /builds/PureTryOut/aports/testing/valhalla/src/valhalla-3.1.0_git20210304/valhalla/sif/pedestriancost.h:5,
                 from /builds/PureTryOut/aports/testing/valhalla/src/valhalla-3.1.0_git20210304/src/sif/pedestriancost.cc:1:
/builds/PureTryOut/aports/testing/valhalla/src/valhalla-3.1.0_git20210304/valhalla/baldr/json.h: In member function 'std::ostream& valhalla::baldr::json::OstreamVisitor::operator()(const string&) const':
/builds/PureTryOut/aports/testing/valhalla/src/valhalla-3.1.0_git20210304/valhalla/baldr/json.h:97:17: error: comparison is always true due to limited range of data type [-Werror=type-limits]
   97 |           if (c >= 0 && c < 32) {
      |               ~~^~~~
```

```
/builds/PureTryOut/aports/testing/valhalla/src/valhalla-3.1.0_git20210304/third_party/rapidjson/include/rapidjson/writer.h:430:67: error: comparison is always true due to limited range of data type [-Werror=type-limits]
  430 |             else if ((sizeof(Ch) == 1 || static_cast<unsigned>(c) < 256) && RAPIDJSON_UNLIKELY(escape[static_cast<unsigned char>(c)]))  {
      |                                          ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
```

This only seems to happen on non-x86 and x86_64 arches like aarch64 and ppc64le.

PR #2937, opened 2021-03-12
Allow disabling Werror

PureTryOut:
Mainly useful for distribution packaging where we don't care about
warnings as long as the resulting package works properly. Make sure it
defaults to ON so packaging will have to turn it OFF and warnings
won't accidentally be missed

# Issue

#2923. This doesn't _solve_ it but helps packagers experiencing the same problem.


PureW:
Looks good to me. I re-triggger the Azure pipeline as the error seemed unrelated to the change.

PureTryOut:
@PureW sorry seems I forgot a spot, I think I caught every place now.

mookerji:
@PureTryOut @PureW we already have this exact cmake flag. see just a few lines up: https://github.com/valhalla/valhalla/blob/d658151cbed872f3606c5c3dfce4583c0c083133/CMakeLists.txt#L33-L37.

We have this default to `OFF` because Valhalla currently has too many warnings to fix ... the build will always fail.

PureTryOut:
Ah yes I missed that. However, it wasn't used everywhere and the build was still failing for me because of it. This PR fixes that, but I will remove that extra line.

mookerji:
@PureTryOut forgot to say: yes, great catch. You're totally right that it wasn't being used for all the targets.

PureTryOut:
@mookerji removed that line, should be good to go now.

PureW:
> @PureTryOut @PureW we already have this exact cmake flag. see just a few lines up:
> 
> https://github.com/valhalla/valhalla/blob/d658151cbed872f3606c5c3dfce4583c0c083133/CMakeLists.txt#L33-L37
> .
> 
> We have this default to `OFF` because Valhalla currently has too many warnings to fix ... the build will always fail.

It's not the exact same thing though. The existing switch enables warnings everywhere which we didn't manage to enable due to existing code.

So we decided to try to migrate piecemeal, file-by-file, to a warning free build. What @PureTryOut is doing is disabling those new warnings/errors on the file-by-file basis.

PureW (changes requested):
The change as is right now disables the file-by-file warnings. See my comment below.

PureTryOut:
Named it `ENABLE_SINGLE_FILES_WERROR` for now, will that do @PureW?

PureW (approved):
Great, thanks!

PureW:
Looks like Azure is broken again...

PureW (approved):

PR #2943, opened 2021-03-17
Fix cutting of response when traffic splits an edge into three parts

danpat:
# Issue

When traffic splits an edge into *three* parts, we were looking at the wrong field to determine if the third segment needs to exist.  Edges are not often split this many times, and when the bug occurred, it resulted a zero-length geometry, so not a huge impact.

## Tasklist

 - [X] Add tests

## Requirements / Relations


test/gurka/test_traffic.cc
```diff
@@ -701,6 +701,94 @@ TEST(Traffic, CutGeoms) {
...
+        ts.speed1 = 20 >> 1;
+        ts.congestion1 = baldr::MAX_CONGESTION_VAL - 1;
+        ts.breakpoint1 = 100;
+
+        ts.speed2 = 40 >> 1;
+        ts.congestion2 = 1;
+
+        // Regression is when breakpoint2 is set to 255,
+        // but speed3 is not set to UNKNOWN - we would incorrectly
+        // emit an additional 0-length slice in the API response
```
PureW, line 730: For my understanding, the case of `speed3` or `congestion3` being set while `breakpoint2` is 255 is still an invalid state, but you're making the code more defensive to gracefully handle such an invalid state?

PureW (approved):
LGTM

kevinkreiser (approved):

kevinkreiser:
feel free to ignore the non required windows build, theres a problem with CI rather than the code

test/gurka/test_traffic.cc
danpat: Yeah - the *old* logic allowed for doing something in that case (breakpoint=255, but still emitting the third split).  The code now correctly does not make the third split if breakpoint=255, so set values in speed3/congestion3 will be properly ignored.

Issue #2896, opened 2021-02-26
Proposal to Simplify TransitionCost


dnesbitt61:
@gknisely keeps bugging me that the 2 transition cost methods (TransitionCost and TransitionCostReverse) have different signatures. Why is that?

The reason there are 2 signatures is that when expanding the search in the reverse direction, the predecessor EdgeLabel is the opposing direction as the directed edge that is traversed along the actual route path. This means that the proper index (idx) used in turn costing and name consistency is not contained in the EdgeLabel, so it must be passed into TransitionCostReverse. Also, TransitionCostReverse has the predecessor directed edge passed into the method rather than the predecessor EdgeLabel so that attributes are correct for the actual direction of travel along the predecessor edge.

To me it doesn't make sense to make the arguments the same in the 2 methods - what makes sense is to simplify so that only a single TransitionCost method is needed. Is this possible? Could we make a single method with the following signature:

Cost TransitionCost(const baldr::DirectedEdge* edge,
                                 const baldr::NodeInfo* node,
                                 const EdgeLabel& pred,
                                 const uint32_t idx)

idx must be passed into the transition cost method. For forward expansion, this is pred.opp_local_idx. For reverse expansion this is meta.edge->localedgeidx().

The other potential issue is that any directed edge attributes used in TransitionCost would need to be the same in both directions (since the predecessor EdgeLabel is the reverse of the actual path directed edge when expanding in the reverse direction). It seems that current transition costs use attributes like classification, use, toll, dest_only which are the same in both directed edges of the actual edge.

The question becomes - would simplification of TransitionCost into a single method outweigh potential risks and possible future needs to use attributes that apply only in 1 direction of an "edge". If so, then this proposal may be possible, but would need significant testing to make sure current routing behavior isn't changed.

dnesbitt61:
Some not so quick changes make this seem improbable. TransitionCostReverse uses the predecessor opposing DirectedEdge as the "to" edge in the transition and the current opposing DirectedEdge as the "from" edge. SO a quick and simple change in bidirectional_astar.cc to use the "common" TransitionCost method (with idx added) in place of TransitionCostReverse doesn't work. The to and from cases are reversed (e.g. transition from a non-ramp onto a ramp is actually reversed to be transition from ramp to non-ramp). Perhaps a "trick" like was done for base_transition_cost to allow interchanging DirectedEdge and EdgeLabel might work - but this maybe obscures some important parts of the logic.

Issue #224, opened 2017-02-02
Penalize Uturns at doubly digitized intersections


dnesbitt61:
(2016-01-25T14:12:15Z) @dnesbitt61 : originally created valhalla/thor#304

This approach should work: 1) Identify and mark left turns onto internal intersection edges (right turns in UK)
2) Add a flag to EdgeLabel for such cases.
3) In TransitionCost - add penalty if prior turn was marked as left onto internal and this transition is left from an internal edge.

NOTE: not seeing this happen much. Perhaps wait until we note some good examples.

kevinkreiser:
in #3886 we have found another pathalogical doubly digitized uturn. the basic issue is that we need to know about name consistancy for N labels back. my thought is similar to daves:

1. when we expand onto an internal edge we crack open the edgeinfo of the predecessor edge (that isnt internal) and gets its nameinfo offset
2. we put this nameinfo offset into the edge label
3. while we keep expanding internal edges we simply propogate this nameinfo offset forward
4. as soon as we expand from an internal to a non internal we check if the nameinfo offsets match. if they do its a uturn if they dont its (propbably, caveats below) not a uturn

i know opening up the edgeinfo during path traversal is a sin but im hoping that if we only do it for those expansions transition to and from internal edges it will be so small as to be unnoticable. and by propogating the offset we can also solve the cases where there are multiple internal edges back to back.

anyway this is just an idea. as dave pointed out originally, its unclear as to the prevelence of this issue whther or not its worth it. one could certainly re-digitize the road network in some small number of these places (tag for the router) to avoid these issues but in other cases the router would indeed have to handle it (eg. the standard `H` and `#` style doubly digitized crossings)

kevinkreiser:
closing this in favor of the discussion over in #3886 this issue specifically was fixed by adding code to detect doubly digitized uturns however that code finds some false positives

PR #2944, opened 2021-03-17
penalize uturns at pencil points and short internal edges

gknisely:
# Issue

Penalize pencil point uturns and uturns at short internal edges.  Note: `motorcycle` and `motor_scooter` do not penalize on 
short internal edges.  No uturn penalty logic has been added to the `pedestrian` and `bicycle` costing models.   

Normal uturns at multi-digitized roads is still allowed.  For an internal edge to be considered short, it must be less than `kShortInternalLength` or 8.0 meters.  Notice no change:

Screenshot from 2021-03-17 11-55-43

## Examples

(1) before - short internal edge
Screenshot from 2021-03-17 12-00-35

(1) after - short internal edge
Screenshot from 2021-03-17 12-00-52

(2) before - uturn at pedestrian crossing
Screenshot from 2021-03-17 12-03-11

(2) after - uturn at pedestrian crossing
Screenshot from 2021-03-17 12-03-26

(3) before - pencil point uturn
Screenshot from 2021-03-17 12-09-44

(3) after - pencil point uturn
Screenshot from 2021-03-17 12-10-01

(4) before - pencil point uturn
Screenshot from 2021-03-17 12-12-57

(4) after - pencil point uturn
Screenshot from 2021-03-17 12-13-11

(5) before - pencil point uturn and short internal edge uturn
Screenshot from 2021-03-17 12-14-49

(5) after - pencil point uturn and short internal edge uturn
Screenshot from 2021-03-17 12-15-02

(6) before - pencil point uturn at destination
Screenshot from 2021-03-17 12-20-39

(6) after - pencil point uturn at destination
Screenshot from 2021-03-17 12-20-58

## Tasklist

 - [x] Add tests

## Requirements / Relations

Added tests for related PR https://github.com/valhalla/valhalla/pull/2891


src/thor/bidirectional_astar.cc
```diff
@@ -317,13 +317,24 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
   float sortcost =
       newcost.cost + astarheuristic_forward_.Get(t2->get_node_ll(meta.edge->endnode()), dist);
 
+  InternalTurn turn = InternalTurn::kNoTurn;
```
kevinkreiser, line 320: this feels like costing bleeding into the algorithms. is there anyway we can turn this into a function in costing so that the costings that dont care about uturns can do nothing and then motor vehicles can share a base class function that does this?

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1479,8 +1479,7 @@ struct graph_callback : public OSMPBF::Callback {
     }
 
     // Infer cul-de-sac if a road edge is a loop and is low classification.
-    if (!way_.roundabout() && loop_nodes_.size() != nodes.size() &&
-        (way_.use() == Use::kRoad || way_.use() == Use::kServiceRoad) &&
+    if (!way_.roundabout() && loop_nodes_.size() != nodes.size() && way_.use() == Use::kRoad &&
```
gknisely, line 1482: this was a data bug that marks drive throughs as cul-de-sacs, which can cause dive bombs

Screenshot from 2021-03-17 11-39-35

src/sif/autocost.cc
```diff
@@ -529,19 +533,40 @@ Cost AutoCost::TransitionCost(const baldr::DirectedEdge* edge,
...
     }
 
+    float seconds = turn_cost;
+    bool is_turn = false;
+    bool has_left = (edge->turntype(idx) == baldr::Turn::Type::kLeft ||
+                     edge->turntype(idx) == baldr::Turn::Type::kSharpLeft);
+    bool has_right = (edge->turntype(idx) == baldr::Turn::Type::kRight ||
+                      edge->turntype(idx) == baldr::Turn::Type::kSharpRight);
+    bool has_reverse = edge->turntype(idx) == baldr::Turn::Type::kReverse;
+
```
gknisely, line 543: moved these to bools as they are used for the uturn logic for short, internal edges and pencil points too

src/sif/bicyclecost.cc
```diff
@@ -336,13 +336,15 @@ class BicycleCost : public DynamicCost {
...
+   * @param  internal_turn  Did we make an turn on a short internal edge.
    * @return  Returns the cost and time (seconds)
    */
   virtual Cost TransitionCostReverse(const uint32_t idx,
                                      const baldr::NodeInfo* node,
                                      const baldr::DirectedEdge* pred,
                                      const baldr::DirectedEdge* edge,
-                                     const bool /*has_measured_speed*/) const override;
+                                     const bool /*has_measured_speed*/,
+                                     const InternalTurn /*internal_turn*/) const override;
```
gknisely, line 347: not used for now

src/sif/motorcyclecost.cc
```diff
@@ -487,15 +490,42 @@ Cost MotorcycleCost::TransitionCost(const baldr::DirectedEdge* edge,
         turn_cost += 0.5f;
     }
 
+    float seconds = turn_cost;
```
gknisely, line 493: moved the logic from autocost.cc to finish the logic for motor cycle, motor scooter, and truck costing.  Ref https://github.com/valhalla/valhalla/pull/2891  I also added tests as they were missing.

src/thor/astar_bss.cc
```diff
@@ -453,7 +454,8 @@ void AStarBSSAlgorithm::SetOrigin(GraphReader& graphreader,
     // of the path.
     uint32_t d = static_cast<uint32_t>(directededge->length() * (1.0f - edge.percent_along()));
     EdgeLabel edge_label(kInvalidLabel, edgeid, directededge, cost, sortcost, dist,
-                         TravelMode::kPedestrian, d, Cost{}, baldr::kInvalidRestriction, true, false);
+                         TravelMode::kPedestrian, d, Cost{}, baldr::kInvalidRestriction, true, false,
```
gknisely, line 457: Not used is bss as it is only bicycle and pedestrian costing

src/thor/bidirectional_astar.cc
gknisely: @kevinkreiser lost me a bit here.  please explain

src/thor/bidirectional_astar.cc
```diff
@@ -760,6 +783,11 @@ bool BidirectionalAStar::SetForwardConnection(GraphReader& graphreader, const BD
   EdgeStatusInfo oppedgestatus = edgestatus_reverse_.Get(oppedge);
   auto opp_pred = edgelabels_reverse_[oppedgestatus.index()];
 
+  // Disallow connections that are part of an uturn on an internal edge
+  if (pred.internal_turn() != InternalTurn::kNoTurn) {
```
gknisely, line 787: we can't allow connections here as the full uturn needs to be checked; otherwise, an edge case like a short route would be allowed.

test/gurka/README.md
```diff
@@ -221,3 +221,18 @@ low-level helper functions available in case you want to do something a little m
   - `gurka::detail::build_config(workdir);` - builds a `boost::property_tree` for tile generation in `workdir`
   - `gurka::detail::map_to_coordinates(ascii_map, gridsize);` - calculates coordinates for all the A-Za-z0-9 nodes in the `ascii_map` given the `gridsize`
   - `gurka::detail::build_pbf(node_locations, ways, nodes, relations, pbf_filename);` - generates an OSM PBF for the nodes, ways, and relations you've defined.  The `nodemap` is the result of `gurka::detail::map_to_coordinates`
+  
+## Debuging help
```
gknisely, line 225: @dgearhart is this enough for the readme for gurka

test/gurka/test_costing_with_traffic.cc
```diff
@@ -0,0 +1,248 @@
+#include "gurka.h"
+#include "test.h"
+#include <gtest/gtest.h>
```
gknisely, line 3: This is the test that is for porting the logic from autocost to the other costing models.  Tests with and without traffic.  

https://github.com/valhalla/valhalla/pull/2944/files#r596212523

gknisely: also @mandeepsandhu thank you for the help with the tests.

src/thor/bidirectional_astar.cc
browndp08: Forgive my ignorance here, but does this impact every travel mode and if so, is that desired? For example, could this negatively impact pedestrian routes?

gknisely: @browndp08 nope.  I noted it in the change log.  `Penalize pencil point uturns and uturns at short internal edges. Note: `motorcycle` and `motor_scooter` models do not penalize on short internal edges. No new uturn penalty logic has been added to the pedestrian and bicycle costing models.`

src/thor/bidirectional_astar.cc
gknisely: @kevinkreiser I am setting `InternalTurn::kLeftTurn` value on the edge label because I need to know if we made a left onto a small internal edge and in the costing we check if this is set and check if we are transitioning off of the internal edge via another left.  This is the type of uturn that we are trying to avoid.  And vice versa for driving on the right.  So I am confused what you want to change to a function.

kevinkreiser: sorry poor suggestion. i see this is only marking what case we have rather than making a decision about it, the costing does that later.

so i guess what im really getting at is, can we encapsulate this into a function on the pathalgorithm class and call this same function from all the different algorithms? it feels bad to copy paste what looks like similar code into every algorithm or am i missing some subtleties?

src/thor/bidirectional_astar.cc
kevinkreiser: this seems strange. why would we totally disallow this rather than penalize it? think of the flip side of the problem where the next opportunity to get to your destination is way way way down the road. uturns are bad for sure, but they arent as bad as driving an extra mile to make a loop you know what i mean?

kevinkreiser: oh and also i get that there is a special consideration that its an internal edge but im not sure that that really matters. maybe you could explain a bit more why this is so undesirable that we would never want to suggest it

src/thor/bidirectional_astar.cc
gknisely: @kevinkreiser so after thinking about this, I realized this morning what you meant.  I will move the following logic to the costing models where it is needed.  

```
  InternalTurn turn = InternalTurn::kNoTurn;
  uint32_t opp_local_idx = pred.opp_local_idx();
  baldr::Turn::Type turntype = meta.edge->turntype(opp_local_idx);
  if (nodeinfo->drive_on_right()) {
    if (meta.edge->internal() && meta.edge->length() <= kShortInternalLength &&
        (turntype == baldr::Turn::Type::kSharpLeft || turntype == baldr::Turn::Type::kLeft))
      turn = InternalTurn::kLeftTurn;
  } else if (meta.edge->internal() && meta.edge->length() <= kShortInternalLength &&
             (turntype == baldr::Turn::Type::kSharpRight || turntype == baldr::Turn::Type::kRight))
    turn = InternalTurn::kRightTurn;
```

gknisely: @kevinkreiser updated.

gknisely: I think it is better in the costing as if we are not interested in this we just return `InternalTurn::kNoTurn` and don't override the TurnType function.  I think in the future we should consider adding additional logic for trucks.  Like what do we consider a short internal edge for trucks vs autos.

src/thor/bidirectional_astar.cc
gknisely: hey @kevinkreiser
I do just penalize it.  Not sure why you thought I don't.  We just can't allow the connection there if is is not equal to `InternalTurn::kNoTurn`  We have to evaluate the full uturn

kevinkreiser: oh im sorry i didnt pay attention to the precise wording there. you are saying we dont allow connecting there because we we cant consider both sides of the internal intersection (since it would be both search paths). i misunderstood that you were saying with the picture that we should never do a uturn over internal intersections. sorry about that

src/sif/autocost.cc
```diff
@@ -590,19 +653,40 @@ Cost AutoCost::TransitionCostReverse(const uint32_t idx,
...
-    if (edge->turntype(idx) == baldr::Turn::Type::kLeft ||
-        edge->turntype(idx) == baldr::Turn::Type::kSharpLeft ||
-        edge->turntype(idx) == baldr::Turn::Type::kRight ||
-        edge->turntype(idx) == baldr::Turn::Type::kSharpRight) {
+    if (has_left || has_right || has_reverse) {
       seconds *= edge->stopimpact(idx);
       is_turn = true;
     }
 
+    if (node->drive_on_right()) {
```
kevinkreiser, line 672: instead of duplicating this a bunch, could we move it into `base_transition_cost` and then have different costings opt in and out of it by setting a boolean member variable that controls whether it happens or not. then costings like pedestrian can set the flag `DynamicCost::penalize_uturns = false` and so when they call `base_transition_cost` it doesnt get applied but when AutoCost calls it it will have the member variable set to true and it will be applied. i'm just trying to lessen the burdon of maintaining this complexity by sequestering it to one place in the code base :smile:

src/thor/bidirectional_astar.cc
gknisely: oh no.  If we don't do this then that edge case could happen.

src/sif/autocost.cc
gknisely: @kevinkreiser hmmm may make `TransitionCostReverse` confusing cause it does not have the same arguments at TransitionCost.  We still have not resolved https://github.com/valhalla/valhalla/issues/2896 yet.  

```
Cost AutoCost::TransitionCostReverse(const uint32_t idx,
                                     const baldr::NodeInfo* node,
                                     const baldr::DirectedEdge* pred,
                                     const baldr::DirectedEdge* edge,
                                     const bool has_measured_speed,
                                     const InternalTurn internal_turn) const {
```

gknisely: If we ever get that resolved then maybe we could move it there.

kevinkreiser: ok cool well then instead of base_transition_cost could we at least turn it into a method on DynamicCost for forward and reverse and have everything share it so at least its all in one place?

src/thor/bidirectional_astar.cc
kevinkreiser: yep i got ya, sorry for the confusion

src/mjolnir/pbfgraphparser.cc
dgearhart: nice

test/gurka/README.md
dgearhart: @gknisely yes, thank you

src/sif/autocost.cc
gknisely: @kevinkreiser ok updated.

src/sif/autocost.cc
```diff
@@ -529,19 +533,40 @@ Cost AutoCost::TransitionCost(const baldr::DirectedEdge* edge,
...
+      // Did we make a uturn on a short, internal edge?
+      if (has_reverse || (pred.internal_turn() == InternalTurn::kLeftTurn && has_left))
```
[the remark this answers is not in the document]
dgearhart: might have to include slight turns if we find any issues - for now good

src/sif/motorcyclecost.cc
```diff
@@ -487,15 +491,48 @@ Cost MotorcycleCost::TransitionCost(const baldr::DirectedEdge* edge,
...
-    if (!edge->has_flow_speed() || flow_mask_ == 0)
-      seconds *= trans_density_factor_[node->density()];
+    if (has_left || has_right || has_reverse) {
+      seconds *= edge->stopimpact(idx);
+      is_turn = true;
+    }
+
+    if (node->drive_on_right()) {
+      // Did we make a uturn at a node.
+      if (has_reverse)
```
gknisely, line 512: Paired with @dgearhart on this.  Decided we want to penalize a reverse just like auto and others for motorcycle and motorscooters

dgearhart (dismissed):
Just had code walk through with @gknisely 
This fixes many u-turn issues - nice job @gknisely 
LGTM for now 
In the future we can refactor to simplify and avoid copy/paste
@kevinkreiser do you agree?

src/sif/autocost.cc
kevinkreiser: i just wish there were a way we could take this block out of TransitionCostReverse and TransitionCost and put them into DynamicCost and use them from there. id really like it if we could whittle down the costings into very slight tweaks on DynamicCost base class functionality so there weren't so much copy pasta and code to maintain, so that when we add a costing feature it gets added to all places more easily etc.

could we pull this and the forward direction out into methods like:

```c++
//pass seconds by reference so it can be modified and inline the functions
AddUturnPenalty(node, edge, has_reverse, has_left, has_right, internal_turn seconds);
AddUturnPenaltyReverse(node, edge, has_reverse, has_left, has_right, internal_turn, seconds);
```

gknisely: @kevinkreiser sure will do

gknisely: @kevinkreiser was able to do this with one function.  if we need to split it out later we can do so.

``
  /**
   * Adds a penalty to 3 types of uturns.  1) uturn on a short, internal edge 2) uturn at a node
   * 3) pencil point uturn. Note that motor_scooter and motorcycle costing models do not penalize
   * uturns on a short, internal edge; hence, the boolean penalize_internal_uturns.
   * @param  idx          Directed edge local index
   * @param  node         Node (intersection) where transition occurs.
   * @param  edge         Directed edge (the to edge)
   * @param  has_reverse  Did we perform a reverse?
   * @param  has_left     Did we make a left (left or sharp left)
   * @param  has_right    Did we make a right (right or sharp right)
   * @param  penalize_internal_uturns   Do we want to penalize uturns on a short, internal edge
   * @param  internal_turn              Did we make an turn on a short internal edge.
   * @param  seconds      Time.
   */
  inline void AddUturnPenalty(const uint32_t idx,
                              const baldr::NodeInfo* node,
                              const baldr::DirectedEdge* edge,
                              const bool has_reverse,
                              const bool has_left,
                              const bool has_right,
                              const bool penalize_internal_uturns,
                              const InternalTurn internal_turn,
                              float& seconds) const
```

kevinkreiser (dismissed):

kevinkreiser:
i know its a bit late, but what does this cost us in performance? anything? looks like its really only a couple more ifs in transition cost which doesnt scare me much..

gknisely:
> i know its a bit late, but what does this cost us in performance? anything? looks like its really only a couple more ifs in transition cost which doesnt scare me much..

We tested it before the code motion, but not after.  @dgearhart what was it

kevinkreiser:
yeah im just gut checking just in case. if you looked at it and it wasnt materially different (which i expect) then no worries from me just curious

dgearhart:
> We tested it before the code motion, but not after. @dgearhart what was it

We will retest performance

gknisely:
Run 1)
master

0.002900362 0.063055992 0.281685424 1.557780247 4.179844141
[1] 0.2677605
>
>

branch
0.003243685 0.065886378 0.294086933 1.561512096 4.541202545
[1] 0.2712838
>
>

Run 2)
master
0.002973318 0.060310364 0.265932035 1.480785275 3.168209553
[1] 0.255456
>
>

branch
0.00328660 0.06354356 0.27470057 1.56310110 3.19098520
[1] 0.2663576
>
>

dgearhart:
## Cross country route performance
image

dgearhart (approved):
thank you @gknisely - fixed many u-turn issues :ship:

kevinkreiser:
fixed #224

PR #2946, opened 2021-03-17
Floating-point precision fix in meili::Project() fixes map-match issue

ktatterso:
# Issue

Map-match segfault.  

Investigation revealed that a very particular geometry led to gps-point projections to land very near to nodes which created a situation that downstream code was not prepared to handle (downstream code being: `meili::CandidateCollector::WithinSquaredDistance(), meili::cut_segments()`).

# How to reproduce

These are the two gps points:  13.2626372011, 38.159659599999999;13.262609100000001, 38.159699600000003

# Resolution

To detect that a gps-point has projected exactly onto a node, code in `meili::CandidateCollector::WithinSquaredDistance()` looks at the `percentage_along` value returned from `meili::helpers::Project()`.  If the `percentage_along` is exactly 0.0 or exactly 1.0 - it projected onto one side of the segment or the other.

However, if the edge is not forward, the `percentage_along` is inverted, e.g., `1.f - percentage_along`.  When `percentage_along` is very small (e.g., less than 1e-8), the computation of `1.f - percentage_along` will be exactly `1.0`.  When we project the same point along the opposing edge, it lands at `0.99998`, which when subtracted from `1.f` yields a somewhat small but non-zero result.  This combination of results is unexpected:  that the projection onto the edge `percentage_along` is `1.0` and onto the the opp-edge is small but non-zero.

There are a number of ways to fix this, but I decided to make the fix in `meili::helpers::Project()`.  I simply check if the `percentage_along` is quite small (less than `1e-7`) and if so snap it to `0.f`.  To be consistent/symmetric, I also check if `percentage_along` is within `1e-7` of `1.f` and if so snap it to `1.f`.  Consequently, the projections will also snap to the respective segment end.

# Thoughts

- This isn't `float`'s fault.  This could also happen with `double`.  It is simply the nature of floating-point.
- I don't like exact comparison's of floats/doubles - like the ones in `meili::CandidateCollector::WithinSquaredDistance()` that check for exact node projection by checking if `percentage_along` is exactly `0.0` or `1.0`.  The snapping fix I made makes me feel a bit better about these particular comparisons, but as a rule I still don't like exact comparisons of floats/doubles. Float/double's are inexact things - and geometric/mathematical code frequently leads to results that are very close to a value but rarely exact.
- Percentages along segments aren't geometrically rigorous (e.g., 10% of a 10 m segment is quite different than 10% of a 1 m segment). I'd rather employ distance tolerances. In fact, in some ways I like my previous fix better. For those who recall, I passed in a `snap_distance=0.001` to `meili::helpers::Project()`.
- Hey @kevinkreiser, I know I indicated this might have something to do with interpolation but I was mistaken.  No interpolation is occurring. Sorry about that.

## Tasklist

 - [x] Add tests


src/meili/geometry_helpers.cc
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  // percent_along is a float between 0 and 1 representing the location of
+  // the closest point on LineString to the given Point, as a fraction
+  // of total 2d line length.
+  closest_partial_length += closest_segment_point.Distance(closest_point);
+  float percent_along =
+      total_length > 0.f ? static_cast<float>(closest_partial_length / total_length) : 0.f;
+
+  // Not so much "snapping" as recognizing that floating-point has limited precision.
```
ktatterso, line 60: This is the fix.

src/meili/geometry_helpers.cc
```diff
@@ -0,0 +1,101 @@
...
+using namespace valhalla::meili;
+using namespace valhalla::midgard;
+
+namespace valhalla {
+namespace meili {
+namespace helpers {
+
+// snapped point, squared distance, segment index, offset
+std::tuple<PointLL, float, typename std::vector<PointLL>::size_type, float>
+Project(const projector_t& p, Shape7Decoder<midgard::PointLL>& shape, float snap_distance) {
```
kevinkreiser, line 19: i know you said we'll have to deal with this with double precision too, but actually this method is behind the times. once upon a time everything was single precision floats, but now the main geometric classes are using double precision. specifically PointLL and its various functions are now in `double` including the `Distance` function below. maybe we should update this code as we did with the rest of the code to use double precision so that at least its up to date as well.

src/meili/geometry_helpers.cc
```diff
@@ -0,0 +1,101 @@
+#include "meili/geometry_helpers.h"
```
kevinkreiser, line 1: i like that you made this a source file but it makes it impossible to tell exactly what was changed. is there any way you could prepare a diff so that its easier for reviewers to tell what the actual changes were?

src/meili/geometry_helpers.cc
ktatterso: I think that's a great idea.  I'd love to take that up in a separate "tech debt" ticket.

src/meili/geometry_helpers.cc
ktatterso: I hear ya... what is the best way to do that?

kevinkreiser:
i forget, is the opposing edge projected separately, ie do we call this method twice with each direction of the edge and for one direction we get a 1 where as the other direction we get a 0.00001 but not 0.0 as expected? it seems incredibly wasteful to project the same geometry twice so i dont suspect we are doing that but i didnt go check to refresh my memory. is it possible that there are more than one pair of edges at the node in question and that we get a discrepancy there about percent along as you discribed, not with an edge and is opposing edge, but with two edges that share a node but arent opposing of each other?

src/meili/geometry_helpers.cc
kevinkreiser: locally you could paste the body of the method back into the header and then do something like `git diff origin/master -- valhalla/meili/geometry_helpers.h` and then post that diff here. that should show just the changes in the header wrt to what is in master (i think, this is off the top of my head)

test/gurka/test_geometry.cc
```diff
@@ -0,0 +1,153 @@
...
+     -81,  1,    -86,  1,    17,   108, 26,   -84,  1,   -100, 1,    54,   -62, 1,   -111, 2,    -120,
+     3,    -127, 2,    -16,  1,    127, -62,  1,    -43, 1,    -32,  1,    -67, 2,   38,   -49,  1,
+     -100, 1,    64,   14,   -74,  1,   105,  -100, 2,   -125, 1,    -44,  3,   -81, 1,    -126, 3,
+     -93,  1,    -64,  1,    -67,  2,   -82,  3,    -85, 1,    -36,  1,    -75, 1,   48,   -67,  1,
+     -128, 1,    -111, 1,    -72,  2,   -59,  2,    -8,  3,    -113, 3,    -40, 3,   -85,  2,    -126,
+     2,    -43,  2,    -68,  2,    -17, 1,    -46,  1,   -35,  2,    -4,   2,   42,  -106, 1,    -96,
+     1,    48,   -82,  7,    -38,  4,   -102, 2,    -90, 1,    -74,  3,    -36, 1,   -16,  1,    47,
+     -38,  2,    -105, 2,    -70,  1,   -37,  1,    -34, 2,    -85,  2,    -44, 1,   117,  -106, 2,
+     91,   -42,  2,    -68,  1,    -18, 2,    -126, 1,   -62,  7,    -62,  2,   -4,  2,    -106, 1,
+     -2,   3,    -118, 1,    -78,  1,   -118, 1};
```
kevinkreiser, line 28: why not a string literal?

kevinkreiser: oh i see you are using the raw 7bit encoded shape we have in the tiles

test/gurka/test_geometry.cc
```diff
@@ -0,0 +1,153 @@
...
+
+//---------------------------------------------------------------------------
+// Convert an encoded shape into a vector of points.
+//---------------------------------------------------------------------------
+void get_shape(const char* enc_shape, size_t size_enc_shape, std::vector<PointLL>& shape_points) {
+  midgard::Shape7Decoder<midgard::PointLL> shape(enc_shape, size_enc_shape);
+  shape_points.clear();
+  while (!shape.empty())
+    shape_points.emplace_back(shape.pop());
+}
```
kevinkreiser, line 38: try using this instead of duplicating: `midgard::decode`

src/meili/geometry_helpers.cc
```diff
@@ -0,0 +1,101 @@
...
+  // Float has "7.2" decimal digits of precision. Hence, we will consider 1e-7 as our
+  // representative small figure, below which we snap to 0.0 and within that distance
+  // of 1.0 will snap to 1.0.
+  constexpr float float_precision_at_one = 1e-7;
+  constexpr float very_close_to_one = 1.0 - float_precision_at_one;
+  if (percent_along < float_precision_at_one) {
+    percent_along = 0.0;
+  } else if (percent_along > very_close_to_one) {
+    percent_along = 1.0;
+  }
```
kevinkreiser, line 81: other than the comments the only change is really this block yes?

src/meili/geometry_helpers.cc
```diff
@@ -0,0 +1,101 @@
...
+// snapped point, squared distance, segment index, offset
+std::tuple<PointLL, float, typename std::vector<PointLL>::size_type, float>
+Project(const projector_t& p, Shape7Decoder<midgard::PointLL>& shape, float snap_distance) {
+  PointLL first_point(shape.pop());
+  auto closest_point = first_point;
+  auto closest_segment_point = first_point;
+  float closest_distance = p.approx.DistanceSquared(closest_segment_point);
+  size_t closest_segment = 0;
+  float closest_partial_length = 0.f;
+  float total_length = 0.f;
```
kevinkreiser, line 26: are you sure you dont want to at least try changing those floats to double and the one on line 57 just to at least see what it does with the test case? it seems pretty low effort to just look, in my opinion

kevinkreiser: ive just tried updating this method on master to use `double` and all the tests pass so i think it would be pretty reasonable to also fix that in this PR. if you disagree i can just PR what ive done separately. what do you think?

kevinkreiser: for reference

```diff
diff --git a/valhalla/meili/geometry_helpers.h b/valhalla/meili/geometry_helpers.h
index 39e0235fc..f90da6c7b 100644
--- a/valhalla/meili/geometry_helpers.h
+++ b/valhalla/meili/geometry_helpers.h
@@ -20,14 +20,14 @@ namespace helpers {
 inline std::tuple<midgard::PointLL, float, typename std::vector<midgard::PointLL>::size_type, float>
 Project(const midgard::projector_t& p,
         midgard::Shape7Decoder<midgard::PointLL>& shape,
-        float snap_distance = 0.f) {
+        double snap_distance = 0) {
   midgard::PointLL first_point(shape.pop());
   auto closest_point = first_point;
   auto closest_segment_point = first_point;
-  float closest_distance = p.approx.DistanceSquared(closest_point);
+  double closest_distance = p.approx.DistanceSquared(closest_point);
   size_t closest_segment = 0;
-  float closest_partial_length = 0.f;
-  float total_length = 0.f;
+  double closest_partial_length = 0;
+  double total_length = 0;
 
   // for each segment
   auto u = first_point;
@@ -57,23 +57,23 @@ Project(const midgard::projector_t& p,
   // the closest point on LineString to the given Point, as a fraction
   // of total 2d line length.
   closest_partial_length += closest_segment_point.Distance(closest_point);
-  float offset = total_length > 0.f ? static_cast<float>(closest_partial_length / total_length) : 0.f;
-  offset = std::max(0.f, std::min(offset, 1.f));
+  double offset = total_length > 0 ? static_cast<float>(closest_partial_length / total_length) : 0;
+  offset = std::max(0.0, std::min(offset, 1.0));
 
   // Snap to vertices if it's close
   if (total_length * offset <= snap_distance) {
     closest_point = first_point;
     closest_distance = p.approx.DistanceSquared(closest_point);
     closest_segment = 0;
-    offset = 0.f;
-  } else if (total_length * (1.f - offset) <= snap_distance) {
+    offset = 0;
+  } else if (total_length * (1 - offset) <= snap_distance) {
     closest_point = u;
     closest_distance = p.approx.DistanceSquared(closest_point);
     closest_segment = i - 1;
     offset = 1.f;
   }
 
-  return std::make_tuple(std::move(closest_point), closest_distance, closest_segment, offset);
+  return std::make_tuple(std::move(closest_point), closest_distance, closest_segment, static_cast<float>(offset));
 }
 
 } // namespace helpers
```

test/gurka/test_geometry.cc
```diff
@@ -0,0 +1,153 @@
...
+    const PointLL* prev_pt = &shape[0];
+    for (size_t i = 1; i < shape.size(); i++) {
+      const PointLL* curr_pt = &shape[i];
+      dist += curr_pt->Distance(*prev_pt);
+      prev_pt = curr_pt;
+    }
+  }
+
+  return dist;
+}
```
kevinkreiser, line 57: there is :smile: its https://github.com/valhalla/valhalla/blob/master/valhalla/midgard/polyline2.h#L64

test/gurka/test_geometry.cc
```diff
@@ -0,0 +1,153 @@
+#include "gurka.h"
```
kevinkreiser, line 1: this test doesnt use gurka at all, so i dont think it needs to be a gurka test. in fact maybe it should be added to an existing test?

test/gurka/test_geometry.cc
ktatterso: The string literal has non-printable characters (and unsigned chars).

src/meili/geometry_helpers.cc
ktatterso: Almost.  There was this line of code previously that was occurring right before this newly added code:
`offset = std::max(0.f, std::min(offset, 1.f));`
It was basically bounding offset to [0.0,1.0].  My condition above achieves the same so I removed that previous logic.

test/gurka/test_geometry.cc
ktatterso: Correct.  I can move it.

src/meili/geometry_helpers.cc
ktatterso: Like I said, the issue can occur with double's too.  It just so happens that double's have enough precision to handle this case.  I'd be happy to switch my logic over to double's.  But I will keep my snapping logic - just adapt it to double's.

src/meili/geometry_helpers.cc
ktatterso: > i forget, is the opposing edge projected separately, ie do we call this method twice with each direction of the edge and for one direction we get a 1 where as the other direction we get a 0.00001 but not 0.0 as expected? it seems incredibly wasteful to project the same geometry twice so i dont suspect we are doing that but i didnt go check to refresh my memory. is it possible that there are more than one pair of edges at the node in question and that we get a discrepancy there about percent along as you discribed, not with an edge and is opposing edge, but with two edges that share a node but arent opposing of each other?

The issue is on this line:
https://github.com/valhalla/valhalla/blob/712824602bff7ec89f24a04c03980ecb38f9a07c/src/meili/candidate_search.cc#L84

The edge is not forward, so `1.f-percentage_along` becomes `1.f` when `percentage_along=1e-9`.

As to whether we are projecting twice, it seems to be the case. However, this PR is largely meant to fix the segfault and improve the mathematical conditioning of the algorithm.  Avoiding double projecting can be taken up in a separate PR I think.

src/meili/geometry_helpers.cc
ktatterso: Ok sure.  But like I said double's will have this problem too.  Switching to double might fix _this_ problem, but isn't the correct long term fix.

src/meili/geometry_helpers.cc
kevinkreiser: > Avoiding double projecting can be taken up in a separate PR I think.

yep im not suggesting you fix that in this pr if thats the case. im trying to make sure i understand correctly what the series of events is that leads to this happening

src/meili/geometry_helpers.cc
kevinkreiser: yeah im not arguing that switching to double is the long term fix, im saying why not also do that clean up while we are here. it sucks that it invalidates the test but id say just keep the test and the changes but also update the function to use doubles

test/gurka/test_geometry.cc
ktatterso: thanks

src/meili/match_route.cc
```diff
@@ -149,7 +149,11 @@ void cut_segments(const std::vector<MatchResult>& match_results,
                                      [&curr_match](const EdgeSegment& segment) {
                                        return (segment.edgeid == curr_match.edgeid);
                                      });
-    assert(last_segment != segments.cend());
+
+    if (last_segment == segments.cend()) {
+      throw std::logic_error("In meili::cutsegments(), unexpectedly unable to locate target edge.");
+      assert(false);
```
kevinkreiser, line 155: this seems strange. you want to crash if we are in debug but in release you want to throw. id say just throw in all cases. it results in a failed map match either way. if we find a failed one that we want to investigate we can hop into the debugger and `catch throw` to see the details at the time of the throw

kevinkreiser: actually no. this assert is dead code, as the coverage test is telling us

kevinkreiser, line 155:
```suggestion
```

src/meili/geometry_helpers.cc
ktatterso: Sure, I can move things to double's.

src/meili/match_route.cc
ktatterso: fair enough... not sure why I did that...

ktatterso: right.

test/gurka/test_geometry.cc
ktatterso: right didn't know about that routine, thanks

src/meili/geometry_helpers.cc
```diff
@@ -0,0 +1,101 @@
...
+namespace meili {
+namespace helpers {
+
+// snapped point, squared distance, segment index, offset
+std::tuple<PointLL, double, typename std::vector<PointLL>::size_type, double>
+Project(const projector_t& p, Shape7Decoder<midgard::PointLL>& shape, double snap_distance) {
+  PointLL first_point(shape.pop());
+  auto closest_point = first_point;
+  auto closest_segment_point = first_point;
+  double closest_distance = 1e100;
```
kevinkreiser, line 23: if you want to change this behavior i suggest you use `std::numeric_limits<double>::max` rather than a hardcoded value, its easier for the reader to understand.

src/meili/match_route.cc
```diff
@@ -141,15 +141,22 @@ void cut_segments(const std::vector<MatchResult>& match_results,
     if (!curr_match.is_break_point && curr_idx != last_idx) {
       continue;
     }
+
+    // Allow for some fp-fuzz in this gt comparison
+    bool prev_gt_curr = prev_match.distance_along > curr_match.distance_along + 1e-3;
```
kevinkreiser, line 146: i still feel that the precondition for calling this function is that things are in order. ie we shouldnt have to account for this because upstream callers were already made to do so. that said its not a huge deal but also kind of contradicts the `throw` on line 158.

valhalla/meili/geometry_helpers.h
```diff
@@ -16,65 +15,11 @@ namespace valhalla {
...
-  } else if (total_length * (1.f - offset) <= snap_distance) {
-    closest_point = u;
-    closest_distance = p.approx.DistanceSquared(closest_point);
-    closest_segment = i - 1;
-    offset = 1.f;
-  }
-
-  return std::make_tuple(std::move(closest_point), closest_distance, closest_segment, offset);
-}
+        double snap_distance = 0.f);
```
kevinkreiser, line 22: nit: 0 or 0.0

kevinkreiser:
other than a few nits and formatting/linting the code this looks pretty much ready

src/meili/match_route.cc
ktatterso: Yea, not sure if you saw my commit comment.  After switching `helpers::Project()` to `double`'s the unit test `MapMatch::interpolation()` was hitting the `throw::logic_error()`.  Here are the values:
```
prev_match.distance_along:  0.85524916907388349
curr_match.distance_along:  0.85524916648864746
```
The difference starts on the 9th decimal.  Float's only have 7.2 decimal digits of precision so, in a sense, we were already rounding to 7 decimal digits by using float.  Given that our projection logic employs both flat-earth approximation for projecting and haversine for distances - and then we combine those computations to calculate `percentage_along` - I think we have to anticipate we will see wobbles like this.

src/meili/geometry_helpers.cc
ktatterso: ok

valhalla/meili/geometry_helpers.h
ktatterso: missed that one, thx

src/meili/match_route.cc
kevinkreiser: yeah my point was only that the idea was that we would fix the wobbliness upstream from this function by fixing the distance along in the match objects before this function ever gets called. but its fine to not widen the scope of this PR (which is what i meant by "not a huge deal").

why do you say the projection logic uses flat-earth approximation (https://github.com/valhalla/valhalla/blob/master/valhalla/midgard/util.h#L593)? the distance approximator certainly does do that, but its used just for the check of whether one point is closer than another. its implementation, while inaccurate for absolute distance, should produce distances that sort properly when comparing whether one pair of points is closer or further than another pair of points.

kevinkreiser (approved):

src/meili/match_route.cc
ktatterso: Correct, we use `projector_t::operator()` for projection, which is a flattened-earth approach.  However, inside `helpers::Project()` we use `PointLL::Distance()` which uses law-of-cosines (I was mistaken when I said haversine). However, the point remains that `PointLL::Distance()` attempts to take into account earth curvature, whereas `projector_t::operator()` does not.

Point being, the `percentage_along` is based on a projected point (using flattened earth) as a percentage of a distance along a segment's length (using curved earth). I don't have a problem with the approach, just noting that results derived by combining these two systems will likely have slight variances against ideal expectations (e.g., a pure cartesian 2D approach) depending on where we are on the globe, segment lengths, etc. Not to mention law-of-cosines known weakness for shorter distances.

Hence the need for tolerances.

kevinkreiser: maybe i misunderstood what you meant by flat earth, but i had assumed you meant cartesian or strictly 2d. you'll notice we do at least scale delta X based on the latitude to account for longitunial geodesics converging at the poles. i wouldnt consider this "flat earth" or "cartesian", but i agree its not a proper projection function that would intersect a perpendicular geodesic with the geodesic between to points.

perhaps your larger point though is actually that because the projection doesnt perfectly (it cant anyway, but even less perfectly) place the projected point along the geodesic and therefore using law of cosines to measure distance will give different results based on how off the geodesic the projected point is?

ktatterso: Correct.

PR #2949, opened 2021-03-17
Allow config object to be passed-in to path algorithms

mandeepsandhu:
This allows us to pass additional configuration values into path
algorithm classes without changing the signature of the constructor
each time.

### DIFF
--- a/src/thor/worker.cc
+++ b/src/thor/worker.cc
@@ -72,15 +72,12 @@ namespace thor {
 thor_worker_t::thor_worker_t(const boost::property_tree::ptree& config,
                              const std::shared_ptr<baldr::GraphReader>& graph_reader)
     : mode(valhalla::sif::TravelMode::kPedestrian),
-      bidir_astar(config.get<uint32_t>("thor.max_reserved_labels_count", kMaxReservedLabelsCount)),
-      bss_astar(config.get<uint32_t>("thor.max_reserved_labels_count", kMaxReservedLabelsCount)),
-      multi_modal_astar(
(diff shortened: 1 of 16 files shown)


mandeepsandhu:
The changes have been borrowed from https://github.com/valhalla/valhalla/pull/2789 . I made a separate PR so that its changes can also be used in https://github.com/valhalla/valhalla/pull/2936 (which will be adding another config option to `thor`).

CC: @keygang

src/thor/worker.cc
```diff
@@ -71,17 +71,11 @@ namespace thor {
...
-      multi_modal_astar(
-          config.get<uint32_t>("thor.max_reserved_labels_count", kMaxReservedLabelsCount)),
-      timedep_forward(
-          config.get<uint32_t>("thor.max_reserved_labels_count", kMaxReservedLabelsCount)),
-      timedep_reverse(
-          config.get<uint32_t>("thor.max_reserved_labels_count", kMaxReservedLabelsCount)),
-      isochrone_gen(config.get<uint32_t>("thor.max_reserved_labels_count", kMaxReservedLabelsCount)),
-      matcher_factory(config, graph_reader), reader(graph_reader), controller{} {
+    : mode(valhalla::sif::TravelMode::kPedestrian), bidir_astar(config.get_child("thor")),
+      bss_astar(config.get_child("thor")), multi_modal_astar(config.get_child("thor")),
```
mandeepsandhu, line 75: Instead of passing the entire config tree, we're passing-in `thor`s configuration object.

kevinkreiser (approved):

averkhaturau (dismissed):

danpat (approved):

Issue #2942, opened 2021-03-16
Pedestrian crossing should be a separate TripLeg_Use


abhi2039:
Currently, pedestrian crossings are treated as regular walkways in Valhalla. Making them separate so that an edge can be identified if it is a pedestrian crossing can have a few benefits:
1. More detailed navigation instructions for pedestrians. 
2. Potentially we could use this information to adjust cost of pedestrian crossings if needed.

I am happy to implement it if folks think it is a good idea.

dnesbitt61:
I take it that there are already edges created for pedestrian crossings and this would just make them a new Use?

abhi2039:
> I take it that there are already edges created for pedestrian crossings and this would just make them a new Use?

That's correct.

dnesbitt61:
This sounds like a good idea to me. From a costing standpoint I could see some slight penalty to prevent too many crossings so a route would tend to stay on one side of a street. 

From a guidance/navigation standpoint it would be nice if there was a way to determine the name of the street that is crossed. I'm not sure this would be feasible for a crossing that does from a sidewalk to another sidewalk (where sidewalks are separate ways).

abhi2039:
@dnesbitt61 Please review my PR if you get a chance: https://github.com/valhalla/valhalla/pull/2950

nilsnolde:
was fixed in #2950

PR #2950, opened 2021-03-17
Pedestrian crossing should be a separate TripLeg_Use

abhi2039:
# Pedestrian crossing should be a separate TripLeg_Use 

What issue is this PR targeting?  Issue #2942 

Added a separate use for pedestrian crossings. Updated narrative dictionary (English only for now and default to old behavior for other languages) to have instructions like "Turn right on the pedestrian crossing"

## Tasklist

 -  Added tests

## Requirements / Relations


dnesbitt61 (dismissed):
Looks good to me from a data building side. I'll let others comment on the narrative/guidance updates.

dgearhart (changes requested):
@abhi2039 I will take a look

proto/trip.proto
```diff
@@ -57,6 +57,7 @@ message TripLeg {
     kPathUse = 27;
     kPedestrianUse = 28;
     kBridlewayUse = 29;
+    kPedestrianCrossing = 32;
```
dgearhart, line 60: Add `Use` suffix to keep consistent

src/odin/narrativebuilder.cc
```diff
@@ -4071,9 +4071,18 @@ NarrativeBuilder::FormStreetNames(const Maneuver& maneuver,
...
     if ((maneuver.travel_mode() == TripLeg_TravelMode_kPedestrian) && maneuver.unnamed_walkway()) {
-      street_names_string = empty_street_name_labels->at(kWalkwayIndex);
+      auto num_empty_street_labels = empty_street_name_labels->size();
+
+      // Pedestrian crossing string is only available for english locale. Default to kWalkwayIndex
+      // for other languages
+      bool apply_pedestrian_crossing_label =
+          maneuver.pedestrian_crossing() && kPedestrianCrossingIndex < num_empty_street_labels;
+      auto dictionary_index =
+          apply_pedestrian_crossing_label ? kPedestrianCrossingIndex : kWalkwayIndex;
```
dgearhart, line 4084: Don't need to do this - just have English placeholder in other language files

locales/en-US.json
```diff
@@ -44,7 +44,8 @@
       "empty_street_name_labels": [
         "the walkway",
         "the cycleway",
-        "the mountain bike trail"
+        "the mountain bike trail",
+        "the pedestrian crossing"
```
dgearhart, line 48: Thinking the en-US (the default) phrase should be `"the crosswalk"` as per: https://en.wikipedia.org/wiki/Pedestrian_crossing

test/narrative_dictionary.cc
```diff
@@ -14,7 +14,8 @@ namespace {
 
 // Expected strings
 const std::vector<std::string> kExpectedEmptyStreetNameLabels = {"the walkway", "the cycleway",
-                                                                 "the mountain bike trail"};
+                                                                 "the mountain bike trail",
+                                                                 "the pedestrian crossing"};
```
dgearhart, line 18: Thinking the en-US (the default) phrase should be "the crosswalk" as per: https://en.wikipedia.org/wiki/Pedestrian_crossing

dgearhart:
@abhi2039 nice job! I left a few comments
Could you please add a gurka test demonstrating the crosswalk use?
[Valhalla Gurka Tests](https://github.com/valhalla/valhalla/blob/master/test/gurka/README.md)
Thank you for contributing

src/odin/narrativebuilder.cc
abhi2039: Thanks for reviewing. Yeah, it does not make sense anymore. I had added it before I realized that if I don't add the corresponding strings in other locales, the tests will fail. I will get rid of it.

abhi2039:
> @abhi2039 nice job! I left a few comments
> Could you please add a gurka test demonstrating the crosswalk use?
> [Valhalla Gurka Tests](https://github.com/valhalla/valhalla/blob/master/test/gurka/README.md)
> Thank you for contributing

Thank you, @dgearhart! Addressed your comments.

abhi2039:
> @abhi2039 nice job! I left a few comments
> Could you please add a gurka test demonstrating the crosswalk use?
> [Valhalla Gurka Tests](https://github.com/valhalla/valhalla/blob/master/test/gurka/README.md)
> Thank you for contributing

Thank you, @dgearhart ! Addressed your comments.

test/gurka/test_crosswalk.cc
```diff
@@ -0,0 +1,23 @@
...
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, 50);
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/crosswalk");
+  auto result = gurka::do_action(valhalla::Options::route, map, {"A", "B"}, "pedestrian");
+
+  EXPECT_EQ(result.directions().routes_size(), 1);
+  EXPECT_EQ(result.directions().routes(0).legs_size(), 1);
+  gurka::assert::raw::expect_path(result, {"AB"});
+
+  auto leg = result.trip().routes(0).legs(0);
+  EXPECT_EQ(leg.node(0).edge().use(), valhalla::TripLeg_Use::TripLeg_Use_kPedestrianCrossingUse);
```
dgearhart, line 22: could we test a path we traverse a crosswalk in the middle - walk down street and then turn left into the crosswalk, etc
You can have test check the actual instructions

abhi2039: Yeah, I wanted to do that but unfortunately couldn't figure out a way to make the ways unnamed in this test. The changes I made are for only unnamed pedestrian crossings. For named pedestrian crossings, nothing changes.

dgearhart: should be able to use `{"name", ""}`

abhi2039: Cool! I will give that a go.

abhi2039: All set now!

abhi2039:
@dgearhart Does this look good to go?

dgearhart:
> @dgearhart Does this look good to go?

I think so - I will try to take final pass today

dgearhart:
@abhi2039 sorry ran out time Friday - I grabbed your branch and will review and run some tests this afternoon

dgearhart:
@abhi2039 I did a small data conversion with the first route in the [pedestrian_routes.txt](https://github.com/valhalla/valhalla/blob/master/test_requests/pedestrian_routes.txt#L1) test file
image

```
1: Walk west on Kingston Street.
2: Turn right onto the crosswalk. 
```
This looks odd to me - since the user might not take the crosswalk in this scenario

Did you [run](https://github.com/valhalla/valhalla/blob/master/run_route_scripts/README.md) the pedestrian test files with new data and diff the results? If so, what were you seeing?

abhi2039:
> @abhi2039 I did a small data conversion with the first route in the [pedestrian_routes.txt](https://github.com/valhalla/valhalla/blob/master/test_requests/pedestrian_routes.txt#L1) test file
> image
> 
> ```
> 1: Walk west on Kingston Street.
> 2: Turn right onto the crosswalk. 
> ```
> 
> This looks odd to me - since the user might not take the crosswalk in this scenario
> 
> Did you [run](https://github.com/valhalla/valhalla/blob/master/run_route_scripts/README.md) the pedestrian test files with new data and diff the results? If so, what were you seeing?

I have not run the pedestrian test files with new data. I have tested it out on the internal fork of Valhalla (at work) which already has these changes. 
The right turn maneuver on San Jose Avenue is starting on the crosswalk and then moves to the sidewalk. Is the concern that the path is transitioning from the road to sidewalk and there is no need to call out crosswalk?

dgearhart:
> The right turn maneuver on San Jose Avenue is starting on the crosswalk and then moves to the sidewalk. Is the concern that the path is transitioning from the road to sidewalk and there is no need to call out crosswalk?

Yes

dgearhart:
> Yeah, I think that's more intuitive. In that case, the instruction will be "Turn right onto the walkway". Basically, I need to add some logic to detect if the maneuver is from a road to a sidewalk. In those cases, we don't need to call out crosswalks. When the maneuver is from a sidewalk to a crosswalk, then the instruction "Turn right onto the crosswalk" makes more sense.
> I think we are on the same page but let me know if I misunderstood something.

yes - exactly.
Also, what are your thoughts on how explicit the instructions should be for each crosswalk traversal? If a user was walking straight down a sidewalk and traversed 3 crosswalks at 3 different intersections - should each one be called out?

abhi2039:
> > Yeah, I think that's more intuitive. In that case, the instruction will be "Turn right onto the walkway". Basically, I need to add some logic to detect if the maneuver is from a road to a sidewalk. In those cases, we don't need to call out crosswalks. When the maneuver is from a sidewalk to a crosswalk, then the instruction "Turn right onto the crosswalk" makes more sense.
> > I think we are on the same page but let me know if I misunderstood something.
> 
> yes - exactly.
> Also, what are your thoughts on how explicit the instructions should be for each crosswalk traversal? If a user was walking straight down a sidewalk and traversed 3 crosswalks at 3 different intersections - should each one be called out?

@dgearhart I updated the logic and added a couple of test points. Please take a look whenever you get a chance. Thanks again for your thorough review.

dgearhart (commented):
@abhi2039 looks good - just need to get CI passing

abhi2039:
> @abhi2039 looks good - just need to get CI passing

@dgearhart Thanks! Just to be sure - is there anything you need from me?

kevinkreiser:
@abhi2039 looks like you just need to resolve the conflict with master. in a recent pr we changed `Maneuver` constructor in the same place you did so you'll just have to resolve that yet!

dgearhart (approved):
Thank you @abhi2039  :ship:

grzezlo:
Thanks for introducing this great feature.
Unfortunately, I found some problem related to detecting the crosswalks.
If the crosswalk is the line, tagged as highway=footway and highway=crossing, everything works well, but often the pedestrian crossing is a part of longer footway.
Example:
at lat:52.321434, lon:20.971978 there is a node[1] in OpenStreetMap tagged highway=crossing.
Way[2] which crosses this node is tagged as highway=footway.
So the line 1312 of graph.lua can not detect this situation, because one of required tags is on the node, and other, on the way.
```
   elseif (kv["highway"] == "footway" and kv["footway"] == "crossing") then
     use = 32
```     
Can it be addressed somehow, or the limitations of processing in graph.lua makes it impossible?
[1]: https://www.openstreetmap.org/node/1459336494
[2]: https://www.openstreetmap.org/way/176939492

abhi2039:
> Thanks for introducing this great feature.
> Unfortunately, I found some problem related to detecting the crosswalks.
> If the crosswalk is the line, tagged as highway=footway and highway=crossing, everything works well, but often the pedestrian crossing is a part of longer footway.
> Example:
> at lat:52.321434, lon:20.971978 there is a node[1] in OpenStreetMap tagged highway=crossing.
> Way[2] which crosses this node is tagged as highway=footway.
> So the line 1312 of graph.lua can not detect this situation, because one of required tags is on the node, and other, on the way.
> 
> ```
>    elseif (kv["highway"] == "footway" and kv["footway"] == "crossing") then
>      use = 32
> ```
> 
> Can it be addressed somehow, or the limitations of processing in graph.lua makes it impossible?
> [1]: https://www.openstreetmap.org/node/1459336494
> [2]: https://www.openstreetmap.org/way/176939492

Sorry for the late response. I cannot think of a way to do this. @dgearhart , @kevinkreiser any suggestions?

PR #2953, opened 2021-03-19
restore compatibility with gcc 6.3.0, libprotobuf 3.0.0, boost v1.62.0

nilsnolde:
I wanted to build python packages on debian 9 (for compatibility with [PEP600](https://www.python.org/dev/peps/pep-0600/)), which has a bunch of fairly old versions.. Note, the build was with `-DENABLE_SERVICES=OFF -DENABLE_TESTS=OFF -DENABLE_BENCHMARKS=OFF`.. 

I would've built at least with SERVICES=ON, but I couldn't get prime_server to be discovered by CMake (v3.18.3), which could be an issue with `pkg-config` which is on `v0.29` on debian 9.. 

```
-- Checking for module 'libprime_server>=0.6.3'
--   Requested 'libprime_server >= 0.6.3' but version of libprime_server is
CMake Error at /usr/local/share/cmake-3.18/Modules/FindPkgConfig.cmake:545 (message):
  A required package was not found
```

### DIFF
--- a/src/loki/polygon_search.cc
+++ b/src/loki/polygon_search.cc
@@ -38,8 +38,6 @@ ring_bg_t PBFToRing(const valhalla::Options::Ring& ring_pbf) {
   for (const auto& coord : ring_pbf.coords()) {
     new_ring.push_back({coord.lng(), coord.lat()});
   }
-  // corrects geometry and handedness as expected by bg for rings
-  bg::correct(new_ring);
   return new_ring;
 }
 
--- a/src/thor/route_action.cc
+++ b/src/thor/route_action.cc
@@ -424,7 +424,8 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
 
   auto route_two_locations = [&, this](auto& origin, auto& destination) -> bool {
     // Get the algorithm type for this location pair
-    thor::PathAlgorithm* path_algorithm = get_path_algorithm(costing, *origin, *destination, options);
+    thor::PathAlgorithm* path_algorithm =
+        this->get_path_algorithm(costing, *origin, *destination, options);
     path_algorithm->Clear();
     algorithms.push_back(path_algorithm->name());
     LOG_INFO(std::string("algorithm::") + path_algorithm->name());
@@ -437,7 +438,7 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
     }
 
     // Get best path and keep it
-    auto temp_paths = get_path(path_algorithm, *origin, *destination, costing, options);
+    auto temp_paths = this->get_path(path_algorithm, *origin, *destination, costing, options);
     if (temp_paths.empty())
       return false;
 
@@ -567,7 +568,8 @@ void thor_worker_t::path_depart_at(Api& api, const std::string& costing) {
 
   auto route_two_locations = [&, this](auto& origin, auto& destination) -> bool {
     // Get the algorithm type for this location pair
-    thor::PathAlgorithm* path_algorithm = get_path_algorithm(costing, *origin, *destination, options);
+    thor::PathAlgorithm* path_algorithm =
+        this->get_path_algorithm(costing, *origin, *destination, options);
     path_algorithm->Clear();
     algorithms.push_back(path_algorithm->name());
     LOG_INFO(std::string("algorithm::") + path_algorithm->name());
@@ -580,7 +582,7 @@ void thor_worker_t::path_depart_at(Api& api, const std::string& costing) {
     }
 
     // Get best path and keep it
-    auto temp_paths = get_path(path_algorithm, *origin, *destination, costing, options);
+    auto temp_paths = this->get_path(path_algorithm, *origin, *destination, costing, options);
     if (temp_paths.empty())
       return false;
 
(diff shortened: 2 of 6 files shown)


src/loki/polygon_search.cc
```diff
@@ -38,8 +38,6 @@ ring_bg_t PBFToRing(const valhalla::Options::Ring& ring_pbf) {
   for (const auto& coord : ring_pbf.coords()) {
     new_ring.push_back({coord.lng(), coord.lat()});
   }
-  // corrects geometry and handedness as expected by bg for rings
-  bg::correct(new_ring);
```
nilsnolde, line 42: `correct()` is not supporting `geographic` CS at lower versions it seems. but actually, let me look at this again in bit more detail.. it corrects a bunch of things, but the handedness check is probably most useful. we register a valhalla type as ring via a macro, where I guess the expected handedness is clockwise (the default for builtin ring type) and all tests are on clockwise rings as well. I'm not sure what'll happen if I add a counterclockwise test, so I'll do that real quick

nilsnolde: ok, seems to be fine.. I altered the avoids test to also test a counterclockwise ring and still passes

src/thor/route_action.cc
```diff
@@ -424,7 +424,8 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
 
   auto route_two_locations = [&, this](auto& origin, auto& destination) -> bool {
     // Get the algorithm type for this location pair
-    thor::PathAlgorithm* path_algorithm = get_path_algorithm(costing, *origin, *destination, options);
+    thor::PathAlgorithm* path_algorithm =
+        this->get_path_algorithm(costing, *origin, *destination, options);
```
kevinkreiser, line 428: can you share the compiler messages that are making you use this syntax, is it something about being ambiguous? we dont use `this` to access members anywhere else in the code base that im aware of so itd be sad to start doing that stylistically if we can instead just rename some other methods that the compiler is complaining about?

src/tyr/route_serializer_osrm.cc
```diff
@@ -1582,7 +1582,7 @@ summarize_route_legs(const google::protobuf::RepeatedPtrField<DirectionsRoute>&
   // unique the same leg (leg_idx) between all routes.
   for (size_t route_i = 0; route_i < routes.size(); route_i++) {
 
-    size_t num_legs_i = routes[route_i].legs_size();
+    size_t num_legs_i = routes.Get(route_i).legs_size();
```
kevinkreiser, line 1585: classic old protobuf :smile:

test/gurka/test_avoids.cc
```diff
@@ -176,20 +176,24 @@ TEST_P(AvoidTest, TestAvoid2Polygons) {
...
+
+  // one clockwise ring
   std::vector<ring_bg_t> rings;
   rings.push_back({{node_a.lng() + 0.1 * dx, node_a.lat() - 0.01 * dy},
                    {node_a.lng() + 0.1 * dx, node_a.lat() - 0.1 * dy},
                    {node_a.lng() - 0.1 * dx, node_a.lat() - 0.1 * dy},
                    {node_a.lng() - 0.1 * dx, node_a.lat() - 0.01 * dy},
                    {node_a.lng() + 0.1 * dx, node_a.lat() - 0.01 * dy}});
+
+  // one counterclockwise ring
```
kevinkreiser, line 192: nice!

src/thor/route_action.cc
nilsnolde: It was smth along the lines of `Missing object` on those lines where `this->` was missing. It's not a general access member thing, rather smth to do with capturing `this` inside the lambda. I can reproduce tmrw and paste the trace if you want.

nilsnolde: here it is:

```make
/valhalla/src/thor/route_action.cc: In instantiation of ‘valhalla::thor::thor_worker_t::path_arrive_by(valhalla::Api&, const string&)::<lambda(auto:2&, auto:3&)> [with auto:2 = std::reverse_iterator<google::protobuf::internal::RepeatedPtrIterator<valhalla::Location> >; auto:3 = std::reverse_iterator<google::protobuf::internal::RepeatedPtrIterator<valhalla::Location> >]’:
/valhalla/src/thor/route_action.cc:518:49:   required from here
/valhalla/src/thor/route_action.cc:441:87: error: cannot call member function ‘std::vector<std::vector<valhalla::thor::PathInfo> > valhalla::thor::thor_worker_t::get_path(valhalla::thor::PathAlgorithm*, valhalla::Location&, valhalla::Location&, const string&, const valhalla::Options&)’ without object
     auto temp_paths = get_path(path_algorithm, *origin, *destination, costing, options);
                                                                                       ^
/valhalla/src/thor/route_action.cc:445:5: error: unable to deduce ‘auto&&’ from ‘temp_paths’
     for (auto& temp_path : temp_paths) {
     ^~~
```

nilsnolde: In a quick search I couldn't find any other similar error report, so I'd blame the old GCC?! In any case, it seems capturing `[&, this]` is redundant and could be simply `[&]`, which compiles both without the above fix on a recent GCC (v10.2) and with the above fix on v6.3.0.

kevinkreiser: ah man that does suck, very unfortunate old compiler bug

kevinkreiser (approved):

src/thor/route_action.cc
nilsnolde: jeez, you're up early on a sat ;)

nilsnolde: ah DST hit in..

kevinkreiser:
no need to merge master and wait for another round of CI changes are localized

PR #2962, opened 2021-03-23
Remove duplicate names being used across different gurka tests

mandeepsandhu:
These would cause random failures when running multiple tests together
as 1 test would overwrite the other tests map dat and fail in
unexpected ways.

### DIFF
--- a/test/gurka/test_avoids.cc
+++ b/test/gurka/test_avoids.cc
@@ -98,7 +98,7 @@ protected:
                               {"BE", {{"highway", "residential"}, {"name", "2nd"}}}};
     const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
     // Add low length limit for avoid_polygons so it throws an error
-    avoid_map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_shortcut",
+    avoid_map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_avoids",
                                   {{"service_limits.max_avoid_polygons_length", "1000"}});
   }
 };
--- a/test/gurka/test_centroid.cc
+++ b/test/gurka/test_centroid.cc
@@ -11,7 +11,7 @@ TEST(centroid, minimal) {
       {"CD", {{"highway", "residential"}}},
   };
   const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
-  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_flat_loop");
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_centroid_minimal");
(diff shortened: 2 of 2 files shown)


PureW (approved):
LGTM

ktatterso (approved):

kevinkreiser:
classic copy paste fail. i which we wouldnt make so many separate gurka test files. surely some of these could be put inside the same test suite, its getting kind of unmanageable large at some point!

test/gurka/test_centroid.cc
```diff
@@ -11,7 +11,7 @@ TEST(centroid, minimal) {
       {"CD", {{"highway", "residential"}}},
   };
   const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
-  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_flat_loop");
```
kevinkreiser, line 14: that was mine, sorry and thanks for getting it!

mandeepsandhu:
> classic copy paste fail. i which we wouldnt make so many separate gurka test files. surely some of these could be put inside the same test suite, its getting kind of unmanageably large at some point!

Yeah, or have gurka create a unique data dir per test (using the test name) and tests can use it when calling `buildtiles`. That way we avoid name collisions.

danpat:
> surely some of these could be put inside the same test suite, its getting kind of unmanageably large at some point!

@kevinkreiser I think this is a symptom of readability - folks will be more likely to re-use existing tests that they can quickly comprehend.  See https://en.wikipedia.org/wiki/Mere-exposure_effect - it is hard to get familiar with some of the tests because they are hard to read, leading to https://en.wikipedia.org/wiki/IKEA_effect :-P

Gurka was somewhat modeled after OSRM's Cucumber test suite that used the Gherkin DSL for defining tests.  There was resistance to adding a DSL to Valhalla, so I came up with the gurka framework to mirror the functionality in a language familiar to the team.  In hindsight, I think a test-specific DSL would have helped reduce the copy-paste behaviour we're seeing in many tests, although it may have hindered adoption of the gurka framework in general.

</programming philosophy rant off>

test/gurka/test_avoids.cc
```diff
@@ -98,7 +98,7 @@ class AvoidTest : public ::testing::TestWithParam<std::string> {
                               {"BE", {{"highway", "residential"}, {"name", "2nd"}}}};
     const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
     // Add low length limit for avoid_polygons so it throws an error
-    avoid_map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_shortcut",
+    avoid_map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_avoids",
```
nilsnolde, line 101: oops, and there's mine.. thanks!

PR #2965, opened 2021-03-24
build: Bail with error if non-existant build-type specified

PureW:
Avoids issues with silently falling back to a debug build

Example:

```
cmake .. -DCMAKE_BUILD_TYPE=RelWithDebug ...  -GNinja
CMake Error at CMakeLists.txt:83 (message):
  Unrecognized build type.  Use one of Debug, Release, RelWithDebInfo,
  MinRelSize
```


danpat (approved):

PureTryOut:
Hmm, not sure if I'm a fan of this. At Alpine Linux we build with `-DCMAKE_BUILD_TYPE=None` to make sure our downstream compiler options are used rather than upstream. Valhalla is now the only package where this doesn't work and we are required to use a different build type instead.

PureW:
This PR was implemented as a way to error out if the user did some kind of mistyping of the build-type. 

I was not aware of the usecase for requesting `None`  but it sounds valid. Why not simply go ahead and open a PR to add `None` to the list of valid build-types introduced in this PR?

PureTryOut:
Ah if you're ok with it, then gladly! Coming in in a bit...

PR #2966, opened 2021-03-24
Fixes CMake if-clauses

PureW:
Non-matching `endif` clauses are triggering

```
CMake Warning (dev) in src/CMakeLists.txt:
  A logical block opening on the line

    /home/anders/Documents/Programming/mapbox/valhalla/src/CMakeLists.txt:186 (if)

  closes on the line

    /home/anders/Documents/Programming/mapbox/valhalla/src/CMakeLists.txt:190 (endif)

  with mis-matching arguments.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Introduced in 1696d9ad1b

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -187,7 +187,7 @@ if (UNIX AND ENABLE_SINGLE_FILES_WERROR)
   # Enables stricter compiler checks on a file-by-file basis
   # which allows us to migrate piecemeal
   set_source_files_properties(proto_conversions.cc PROPERTIES COMPILE_FLAGS "-Wall -Werror")
-endif (UNIX)
+endif (UNIX AND ENABLE_SINGLE_FILES_WERROR)
 
 if (ENABLE_SERVICES)
   set(valhalla_hdrs ${valhalla_hdrs} ${VALHALLA_SOURCE_DIR}/valhalla/tile_server.h)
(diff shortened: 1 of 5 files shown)


mandeepsandhu (dismissed):
good catch!

src/CMakeLists.txt
```diff
@@ -187,7 +187,7 @@ if (UNIX AND ENABLE_SINGLE_FILES_WERROR)
   # Enables stricter compiler checks on a file-by-file basis
   # which allows us to migrate piecemeal
   set_source_files_properties(proto_conversions.cc PROPERTIES COMPILE_FLAGS "-Wall -Werror")
-endif (UNIX)
+endif (UNIX AND ENABLE_SINGLE_FILES_WERROR)
```
kevinkreiser, line 190: why not make all of these just `endif ()`? indentation is surely enough to match these up

kevinkreiser: see here: https://cmake.org/cmake/help/latest/command/endif.html

> The optional <condition> argument is supported **for backward compatibility only**. If used it must be a verbatim repeat of the argument of the opening if clause.

I highly suggest we dont put anything in the condition as its meaningless and extra crap to maintain

PureW: Fully agree. I didn't know that was an option!

kevinkreiser (approved):

Issue #2885, opened 2021-02-24
Inconsistent handling of edge-traversal between routing algorithms


danpat:
Per https://github.com/valhalla/valhalla/pull/2875#discussion_r580720480, we have some divergence in the handling of dead-ends between different algorithms.  This is just a small example of a bigger problem.

We have the following code that all implements slight variants of the graph search:

  - multimodal
  - bidirectional_astar
  - timedep_forward
  - timedep_reverse
  - astar_bss
  - loki/reach
  - centroid
  - isochrone
  - timedistancematrix
  - timedistancematrixbss
  - others??? (map matching?)

Each of these implement *largely* similar code.  When adding things like [fixing uturns at deadends](https://github.com/valhalla/valhalla/pull/2875), [starting routes from closed points](https://github.com/valhalla/valhalla/pull/2846), and others, one ends up needing to duplicate code in multiple places.

We're seeing quite a bit of drift - folks are only making changes to the algorithm that affects the feature they're working on, without realizing (or perhaps being daunted by) how many other places their change needs to be made.

Each of the routing algorithms above should be reviewed.  Common logic needs to be moved into a common base class or template, and algorithm-specific code needs to be kept separate.

kevinkreiser:
@danpat merging as much base functionality into a base class will be a gargantuan refactor considering:

1. use or disuse of hierarchy culling
2. using the correct heuristics depending on uni vs bidirectional
3. different types of edge labels
4. different stopping criteria

I think the end goal is excellent but the journey may be quite arduous :smile: As you noted in #2967 regression tests will be key.

I also wonder if we can somehow move a function at a time into path algorithm? I've tried this a bit in the past (i guess you were alluding to that) and i've found that an approach that worked best was to edit the methods in place (without moving them) until they were nearly identical and everythign was still working, and then finally moving them as the last step. ive also found that there are a decent amount of small differences among implementations which makes this work doubly annoying because the small differences are often edge cases which may not be adequately covered by unit tests.

anyway all that to say i hope we can get some serious consolidation over time :+1:

danpat:
@kevinkreiser yep, I've been staring at the code for a couple of days to try and get enough of a picture to start to take some steps.

I've been doing exactly what you suggested, starting with the isochrone plugin - functions like `ExpandForward` and `ExpandReverse` are relatively straightforward to templatize and condense.

[...]

PR #2967, opened 2021-03-25
Call correct isochrone compute method depending on requested expansion type

danpat:
# Issue

Looks like this was a typo introduced in the refactoring done [here](https://github.com/valhalla/valhalla/commit/d7a4e787615c54e2bf72fd452ba47d59051005c1#diff-c9004ab6fb25d790ff2f519dc4d705e5d336521654a5b74dba0bb5f18d2c703aL207)

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/dijkstras.cc
+++ b/src/thor/dijkstras.cc
@@ -235,10 +235,10 @@ void Dijkstras::Expand(ExpansionType expansion_type,
       Compute(*api.mutable_options()->mutable_locations(), reader, costings, mode);
       break;
     case ExpansionType::reverse:
-      Compute(*api.mutable_options()->mutable_locations(), reader, costings, mode);
+      ComputeReverse(*api.mutable_options()->mutable_locations(), reader, costings, mode);
       break;
     case ExpansionType::multimodal:
-      Compute(*api.mutable_options()->mutable_locations(), reader, costings, mode);
+      ComputeMultiModal(*api.mutable_options()->mutable_locations(), reader, costings, mode);
       break;
     default:
       throw std::runtime_error("Unknown expansion type");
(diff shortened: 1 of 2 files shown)


src/thor/dijkstras.cc
```diff
@@ -235,10 +235,10 @@ void Dijkstras::Expand(ExpansionType expansion_type,
       Compute(*api.mutable_options()->mutable_locations(), reader, costings, mode);
       break;
     case ExpansionType::reverse:
-      Compute(*api.mutable_options()->mutable_locations(), reader, costings, mode);
+      ComputeReverse(*api.mutable_options()->mutable_locations(), reader, costings, mode);
```
mandeepsandhu, line 238: Is it possible to add tests (new or update existing) to catch such issues? (haven't looked at current isochrone tests)

kevinkreiser: its hard to do the multimodal one unless we just test that it works for a short pedestrian only route. one day we need to figure out how to generate transit pbfs from gurka that would be awesome.

as for reverse we can totally test it but it will be just as unscientific of a test as the existing isochrone test in that we'll just check that a polygon comes out. its very hard to really accurately test this type of thing where the result is a geometric shape.

kevinkreiser (approved):

kevinkreiser:
This is a good breakfix. We should add testing but it seems unfair to put that burden on this PR given there was no testing on it before.

danpat:
@kevinkreiser 👍 thanks - I have some refactoring PRs coming to start to address https://github.com/valhalla/valhalla/issues/2885, and I'm going to start with aligning the isochrone and bidir_astar code.  I'll include some tests with that, because as this PR shows, refactoring can be risky :-)

PR #2969, opened 2021-03-26
WIP: Consolidate forward/reverse edge traversal in isochrone/reach code.

danpat:
# Issue

This PR is the first of what I expect to be a series that work towards addressing https://github.com/valhalla/valhalla/issues/2885

There's lots of work to do, so in order to mimize conflict with other concurrent work, I plan to push early and often, and get these refactors into `master` iteratively, rather than in one big refactor nuclear explosion.

This PR addresses some of the code duplication within `dijkstra.cc` - in particular, the `Compute/ComputeReverse` functions.  Specifically, here's what's happened;

  - `Dijkstra::ExpandForward/ExpandReverse` have been consolidated into a single template function called `ExpandTemplate<direction>` whose body lives in `dijkstra.h` - there are `if` statements within that to handle the slight differences in forward/reverse expansion - final naming of this function is TBD, I just picked one to make it obvious.
  - Same for `Compute/ComputeReverse` - consolidated into `Compute<direction>` in `dijkstra.h`
  - All callers of these functions have been updated to call the correct templated version

I did *not* touch the `ComputeMultiModal` function yet - there's a lot more diff in there than between the forward/reverse expansion code.  I'd like to do it as a separate PR once we've settled this one.

This is a fairly small refactor - I expect the first few of these will trigger numerous discussions, and as we sort through the concerns, I'll be able to move through the remaining code more quickly.  With that in mind, I'm going to try to keep the PRs small and contained so we don't have to review/discuss huge sweeping sets of changes.

As I go through the code, I'm absolutely sure I'm going to find inconsistencies - that's the whole reason for undergoing this refactor in the first place.  I'm going to do my best to properly merge duplicated code and incorporate features that are implemented in one place but not in another, but when I'm unsure, I will put explicit questions/comments in the PR to ensure that the correct behaviour is implemented.

The overall intent here is pure refactoring.  I expect that some bugfixes that exist in one routing algorithm will become available in another routing algorithm that was stale, but other than that, my goal is to have no impact on performance or functionality with these changes.  As folks review these PRs, if you notice things that need to be fixed, let's make issues for them and keep these PRs focused on the refactoring process.

How did I do this?

I moved the functions that look like they had a lot of shared code into separate files, then diffed them.  I carefully reviewed the diffs, change-by-change, and created new consolidated versions of the functions with compile-time branches instead of separate mostly-duplicated copies.  Here's a sample diff of the `Expand/ExpandReverse` functions that got merged here:

Screen Shot 2021-03-25 at 6 42 31 PM

The actual functional differences are fairly small - some flipped variables in the reverse case, and some unnecessary different naming, and the odd special-case logic, mostly for the reverse case.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/dijkstras.cc
+++ b/src/thor/dijkstras.cc
@@ -213,16 +213,14 @@ void Dijkstras::ExpandInner(baldr::GraphReader& graphreader,
       newcost = pred.cost() +
                 costing_->EdgeCost(directededge, tile, offset_time.second_of_week, flow_sources) +
                 transition_cost;
-    } else if (expansion_direction == ExpansionType::reverse) {
+    } else {
       transition_cost =
           costing_->TransitionCostReverse(directededge->localedgeidx(), nodeinfo, opp_edge,
                                           opp_pred_edge, pred.has_measured_speed(),
                                           pred.internal_turn());
-      newcost =
-          pred.cost() + costing_->EdgeCost(opp_edge, t2, offset_time.second_of_week, flow_sources);
-      // TODO(danpat): why is this done differently in the reverse direction? This was refactored
-      //               to here from older code.
-      newcost.cost += transition_cost.cost;
+      newcost = pred.cost() +
+                costing_->EdgeCost(opp_edge, t2, offset_time.second_of_week, flow_sources) +
+                transition_cost;
     }
     uint32_t path_dist = pred.path_distance() + directededge->length();
 
--- a/src/thor/dijkstras.cc
+++ b/src/thor/dijkstras.cc
@@ -109,13 +109,16 @@ Dijkstras::SetTime(google::protobuf::RepeatedPtrField<valhalla::Location>& locat
   return infos;
 }
 
-// Expand from a node in the forward direction
-void Dijkstras::ExpandForward(GraphReader& graphreader,
-                              const GraphId& node,
-                              const EdgeLabel& pred,
-                              const uint32_t pred_idx,
-                              const bool from_transition,
-                              const TimeInfo& time_info) {
+template <const ExpansionType expansion_direction>
+void Dijkstras::ExpandInner(baldr::GraphReader& graphreader,
+                            const baldr::GraphId& node,
+                            const decltype(Dijkstras::bdedgelabels_)::value_type& pred,
+                            const uint32_t pred_idx,
+                            const baldr::DirectedEdge* opp_pred_edge,
+                            const bool from_transition,
+                            const baldr::TimeInfo& time_info) {
+
+  constexpr bool FORWARD = expansion_direction == ExpansionType::forward;
   // Get the tile and the node info. Skip if tile is null (can happen
   // with regional data sets) or if no access at the node.
   graph_tile_ptr tile = graphreader.GetGraphTile(node);
@@ -141,8 +144,10 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
 
   // Update the time information
   auto offset_time =
-      from_transition ? time_info
-                      : time_info.forward(pred.cost().secs, static_cast<int>(nodeinfo->timezone()));
+      from_transition
+          ? time_info
+          : (FORWARD ? time_info.forward(pred.cost().secs, static_cast<int>(nodeinfo->timezone()))
+                     : time_info.reverse(pred.cost().secs, static_cast<int>(nodeinfo->timezone())));
 
   // Expand from end node in forward direction.
   GraphId edgeid = {node.tileid(), node.level(), nodeinfo->edge_index()};
@@ -154,8 +159,23 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
     // (based on the costing method) or if a complex restriction exists for
     // this path.
     if (directededge->is_shortcut() || es->set() == EdgeSet::kPermanent ||
-        !(directededge->forwardaccess() & access_mode_)) {
+        !((FORWARD ? directededge->forwardaccess() : directededge->reverseaccess()) & access_mode_)) {
       continue;
+    }
+
+    // Get end node tile, opposing edge Id, and opposing directed edge.
+    graph_tile_ptr t2;
+    baldr::GraphId oppedgeid;
+    const baldr::DirectedEdge* opp_edge;
+    // Only populate these at this point for the reverse search - the forward search
+    // may exit before needing them, so we can defer some work until later
+    if (not FORWARD) { // aka reverse
+      t2 = tile;
+      oppedgeid = graphreader.GetOpposingEdgeId(edgeid, t2);
+      if (t2 == nullptr) {
+        continue;
+      }
+      opp_edge = t2->directededge(oppedgeid);
     }
 
     // Check if the edge is allowed or if a restriction occurs
@@ -163,25 +183,47 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
     uint8_t restriction_idx = -1;
     if (offset_time.valid) {
       // With date time we check time dependent restrictions and access
-      if (!costing_->Allowed(directededge, pred, tile, edgeid, offset_time.local_time,
-                             nodeinfo->timezone(), restriction_idx) ||
-          costing_->Restricted(directededge, pred, bdedgelabels_, tile, edgeid, true, todo,
-                               offset_time.local_time, nodeinfo->timezone())) {
+      const bool allowed =
+          FORWARD ? costing_->Allowed(directededge, pred, tile, edgeid, offset_time.local_time,
+                                      nodeinfo->timezone(), restriction_idx)
+                  : costing_->AllowedReverse(directededge, pred, opp_edge, t2, oppedgeid,
+                                             offset_time.local_time, nodeinfo->timezone(),
+                                             restriction_idx);
+      if (!allowed || costing_->Restricted(directededge, pred, bdedgelabels_, tile, edgeid, true,
+                                           todo, offset_time.local_time, nodeinfo->timezone())) {
         continue;
       }
     } else {
-      if (!costing_->Allowed(directededge, pred, tile, edgeid, 0, 0, restriction_idx) ||
-          costing_->Restricted(directededge, pred, bdedgelabels_, tile, edgeid, true)) {
+      const bool allowed =
+          FORWARD ? costing_->Allowed(directededge, pred, tile, edgeid, 0, 0, restriction_idx)
+                  : costing_->AllowedReverse(directededge, pred, opp_edge, t2, oppedgeid, 0, 0,
+                                             restriction_idx);
+
+      if (!allowed || costing_->Restricted(directededge, pred, bdedgelabels_, tile, edgeid, true)) {
         continue;
       }
     }
 
     // Compute the cost and path distance to the end of this edge
-    Cost transition_cost = costing_->TransitionCost(directededge, nodeinfo, pred);
+    Cost transition_cost, newcost;
     uint8_t flow_sources;
-    Cost newcost = pred.cost() +
-                   costing_->EdgeCost(directededge, tile, offset_time.second_of_week, flow_sources) +
-                   transition_cost;
+
+    if (FORWARD) {
+      transition_cost = costing_->TransitionCost(directededge, nodeinfo, pred);
+      newcost = pred.cost() +
+                costing_->EdgeCost(directededge, tile, offset_time.second_of_week, flow_sources) +
+                transition_cost;
+    } else if (expansion_direction == ExpansionType::reverse) {
+      transition_cost =
+          costing_->TransitionCostReverse(directededge->localedgeidx(), nodeinfo, opp_edge,
+                                          opp_pred_edge, pred.has_measured_speed(),
+                                          pred.internal_turn());
+      newcost =
+          pred.cost() + costing_->EdgeCost(opp_edge, t2, offset_time.second_of_week, flow_sources);
+      // TODO(danpat): why is this done differently in the reverse direction? This was refactored
+      //               to here from older code.
+      newcost.cost += transition_cost.cost;
+    }
     uint32_t path_dist = pred.path_distance() + directededge->length();
 
     // Check if edge is temporarily labeled and this path has less cost. If
@@ -197,9 +239,12 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
       continue;
     }
 
-    // Only needed if you want to connect with a reverse path
-    graph_tile_ptr t2 = tile;
-    GraphId oppedgeid = graphreader.GetOpposingEdgeId(edgeid, t2);
+    // Only needed if you want to connect with a reverse path - for reverse mode,
+    // these were populated earlier
+    if (expansion_direction == ExpansionType::forward) {
+      t2 = tile;
+      oppedgeid = graphreader.GetOpposingEdgeId(edgeid, t2);
+    }
 
     // Add edge label, add to the adjacency list and set edge status
     uint32_t idx = bdedgelabels_.size();
@@ -208,23 +253,46 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
                                transition_cost, path_dist, false,
                                (pred.closure_pruning() || !costing_->IsClosed(directededge, tile)),
                                static_cast<bool>(flow_sources & kDefaultFlowMask),
-                               costing_->TurnType(pred.opp_local_idx(), nodeinfo, directededge),
+                               (expansion_direction == ExpansionType::forward)
+                                   ? costing_->TurnType(pred.opp_local_idx(), nodeinfo, directededge)
+                                   : costing_->TurnType(directededge->localedgeidx(), nodeinfo,
+                                                        opp_edge, opp_pred_edge),
                                restriction_idx, pred.path_id());
     adjacencylist_.add(idx);
   }
 
   // Handle transitions - expand from the end node of each transition
   if (!from_transition && nodeinfo->transition_count() > 0) {
-    const NodeTransition* trans = tile->transition(nodeinfo->transition_index());
+    const baldr::NodeTransition* trans = tile->transition(nodeinfo->transition_index());
     for (uint32_t i = 0; i < nodeinfo->transition_count(); ++i, ++trans) {
-      ExpandForward(graphreader, trans->endnode(), pred, pred_idx, true, offset_time);
+      ExpandInner<expansion_direction>(graphreader, trans->endnode(), pred, pred_idx, opp_pred_edge,
+                                       true, offset_time);
     }
   }
 }
 
+// Note: multimodal not ported yet, so there is no implementation for ExpansionType::multimodal
+template void Dijkstras::ExpandInner<ExpansionType::forward>(
+    baldr::GraphReader& graphreader,
+    const baldr::GraphId& node,
+    const decltype(Dijkstras::bdedgelabels_)::value_type& pred,
+    const uint32_t pred_idx,
+    const baldr::DirectedEdge* opp_pred_edge,
+    const bool from_transition,
+    const baldr::TimeInfo& time_info);
+template void Dijkstras::ExpandInner<ExpansionType::reverse>(
+    baldr::GraphReader& graphreader,
+    const baldr::GraphId& node,
+    const decltype(Dijkstras::bdedgelabels_)::value_type& pred,
+    const uint32_t pred_idx,
+    const baldr::DirectedEdge* opp_pred_edge,
+    const bool from_transition,
+    const baldr::TimeInfo& time_info);
+
 // performs one of the types of expansions
-// TODO: reduce code duplication between forward, reverse and multimodal as they are nearly identical
-void Dijkstras::Expand(ExpansionType expansion_type,
+// TODO: reduce code duplication between forward, reverse and multimodal as they are nearly
+// identical
+void Dijkstras::Expand(const ExpansionType expansion_type,
                        valhalla::Api& api,
                        baldr::GraphReader& reader,
                        const sif::mode_costing_t& costings,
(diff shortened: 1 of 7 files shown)


valhalla/thor/dijkstras.h
```diff
@@ -194,29 +234,186 @@ class Dijkstras {
...
+                  transition_cost;
+      } else if (expansion_direction == ExpansionType::reverse) {
+        transition_cost =
+            costing_->TransitionCostReverse(directededge->localedgeidx(), nodeinfo, opp_edge,
+                                            opp_pred_edge, pred.has_measured_speed(),
+                                            pred.internal_turn());
+        newcost =
+            pred.cost() + costing_->EdgeCost(opp_edge, t2, offset_time.second_of_week, flow_sources);
+        // TODO(danpat): why is this done differently in the reverse direction?
+        newcost.cost += transition_cost.cost;
```
danpat, line 368: Does anyone know why the reverse implementation does this separate addition of the `.cost` field, rather than doing the same as the forward search does?

valhalla/thor/dijkstras.h
```diff
@@ -194,29 +234,186 @@ class Dijkstras {
...
+  template <const ExpansionType expansion_direction,
+            // As multimodal isn't ported yet, but we're using the 3-value enum, check that
+            // the caller is using a supported expansion mode (this is a compile time check to
+            // catch this mistake early).  Once multimodal is supported, this can be removed.
+            class = typename std::enable_if<expansion_direction == ExpansionType::forward || expansion_direction == ExpansionType::reverse>::type>
+  void ExpandTemplate(baldr::GraphReader& graphreader,
                      const baldr::GraphId& node,
                      const sif::EdgeLabel& pred,
                      const uint32_t pred_idx,
+                     const baldr::DirectedEdge *opp_pred_edge,
```
danpat, line 246: Templatizing this consolidates the signature, and requires callers in the forward case to supply something for `opp_pred_edge` - nullptr in most cases.  Does anyone have an issue with that general pattern?

valhalla/thor/dijkstras.h
```diff
@@ -194,29 +234,186 @@ class Dijkstras {
...
    */
-  void ExpandForward(baldr::GraphReader& graphreader,
+  template <const ExpansionType expansion_direction,
+            // As multimodal isn't ported yet, but we're using the 3-value enum, check that
+            // the caller is using a supported expansion mode (this is a compile time check to
+            // catch this mistake early).  Once multimodal is supported, this can be removed.
+            class = typename std::enable_if<expansion_direction == ExpansionType::forward || expansion_direction == ExpansionType::reverse>::type>
+  void ExpandTemplate(baldr::GraphReader& graphreader,
                      const baldr::GraphId& node,
                      const sif::EdgeLabel& pred,
```
danpat, line 244: There are no uses of the `BDEdgelabel` fields inside this function, but some variants of `ExpandInnerReverse` in other classes take `BDEdgeLabel` - can anyone anticipate any issues with having a single signature here?  In what cases is `BDEdgeLabel` required?  Is there any harm in deprecating it if necessary and moving its fields into `EdgeLabel`?

valhalla/thor/dijkstras.h
```diff
@@ -194,29 +236,187 @@ class Dijkstras {
...
+    baldr::GraphId edgeid = {node.tileid(), node.level(), nodeinfo->edge_index()};
+    EdgeStatusInfo* es = edgestatus_.GetPtr(edgeid, tile, pred.path_id());
+    const baldr::DirectedEdge* directededge = tile->directededge(edgeid);
+    for (uint32_t i = 0; i < nodeinfo->edge_count(); ++i, ++directededge, ++edgeid, ++es) {
+      // Skip this edge if permanently labeled (best path already found to this
+      // directed edge). skip shortcuts or if no access is allowed to this edge
+      // (based on the costing method) or if a complex restriction exists for
+      // this path.
+      if (expansion_direction == ExpansionType::forward) {
+        if (directededge->is_shortcut() || es->set() == EdgeSet::kPermanent ||
```
danpat, line 294: @dnesbitt61 In the original code, you had authored this `if` statement and the one below it (line 299) with a different order of operations.  I've normalized them for readability, but was there a reason for the different ordering of conditions?

valhalla/thor/dijkstras.h
kevinkreiser: BD edge label is required for the algorithms that do reverse searches. for forward only algorithms like dijkstras compute and timedep_forward you can get away with just EdgeLabel. i would have like dot completely delete EdgeLabel and have everything use BDEdgelabel But the difference in size actually does have a performance impact. I too did consolidation on dijkstras when i made it work for reach as well and when i did that i converted the forward direction of dijstras to also use BDEdgeLabel which came with a small penalty. When we get to the timedep forward algorithm and consolidating that with reverse (even more) we should check what we pay to move it to BDEdgeLabel but I'm hoping its small enough to be worth the refactor

valhalla/thor/dijkstras.h
```diff
@@ -78,22 +78,64 @@ class Dijkstras {
    * @param  mode_costing List of costing objects
    * @param  mode         Travel mode
    */
-  virtual void Compute(google::protobuf::RepeatedPtrField<valhalla::Location>& origin_locs,
```
kevinkreiser, line 81: can you move these back to the `cc` file and use explicit template instantiations? i like hiding the details of the algorithms in the src files if we can. you can then also remove your `std::enable_if` because the explicit instantiation takes the place of it. or is there an other reason to have it in the header that im missing?

valhalla/thor/dijkstras.h
danpat: Ok - I'll see if it's practical to turn the label type into a template parameter too - that might be the best of both worlds.

On my list today is to add a benchmark for this in `bench/` - I'll do one for `reach` as well.

valhalla/thor/dijkstras.h
danpat: No, just expedience - will move.

valhalla/thor/dijkstras.h
PureW: Works for me.

src/thor/dijkstras.cc
```diff
@@ -154,34 +159,71 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
...
-      if (!costing_->Allowed(directededge, pred, tile, edgeid, offset_time.local_time,
-                             nodeinfo->timezone(), restriction_idx) ||
-          costing_->Restricted(directededge, pred, bdedgelabels_, tile, edgeid, true, todo,
-                               offset_time.local_time, nodeinfo->timezone())) {
+      const bool allowed =
+          FORWARD ? costing_->Allowed(directededge, pred, tile, edgeid, offset_time.local_time,
+                                      nodeinfo->timezone(), restriction_idx)
+                  : costing_->AllowedReverse(directededge, pred, opp_edge, t2, oppedgeid,
+                                             offset_time.local_time, nodeinfo->timezone(),
+                                             restriction_idx);
```
kevinkreiser, line 191: this almost makes me want to see the templatization trickle down into costing in the future. one step at a time though haha

src/thor/dijkstras.cc
```diff
@@ -693,7 +624,8 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
       continue;
     }
 
-    // Make the label in advance, we may not end up using it but we need it for the expansion decision
+    // Make the label in advance, we may not end up using it but we need it for the expansion
+    // decision
```
kevinkreiser, line 628: nit: saw this above as well, does the linter have different widths in different peoples editors?

src/thor/dijkstras.cc
danpat: Yeah, I had exactly the same though.  I've bookmarked it for later in the refactor :-)

kevinkreiser:
got to love that diff ratio:
image

so far this looks good to me. once CI is good with everything. i think all we have to do is run RAD and perf tests on it to confirm its within measurement error. since its mainly isochrones it can be a little bit less straightforward but there are modes on the `run_with_server.py` that will allow you to do both checking for diffs in results as well as a general performance comparison. just let me know if that makes sense as a way to confirm that the changes are good

src/thor/dijkstras.cc
danpat: Fixed this - it was because I moved the code into the `.h` file initially that had a slightly different indentation depth, then moved it back.

danpat:
@kevinkreiser 👍 I've got a `bench/thor/isochrone.cc` in the works to check the core algorithm - I'm not expecting any impact, most of the new branches are evaluated at compile time and disappear.  RAD tests TBD but I agree they should be done.

valhalla/thor/dijkstras.h
kevinkreiser: FWIW i had already turned the labels into a template parameter in the initialization function when i took a pass at reduce reuse recycle on dijkstras (or actually when i created it out of isochrones?) for the purposes of the reach computation.

valhalla/thor/dijkstras.h
danpat: @kevinkreiser Do you have any clue on this one?  Do these additions need to be done differently ?

kevinkreiser: yeah so i see that in the reverse direction we only add the cost of the transition to the costing object but not the elapsed time of the transition... off the top of my head i dont know why that is... i'll have to look around at the code a bit more tonight to refresh m y memory unless @dnesbitt61 remembers?

kevinkreiser: more succinctly, why in the forward direction to we accumulate both cost and secs for transitions: 
https://github.com/valhalla/valhalla/blob/master/src/thor/dijkstras.cc#L182-L184

but in the reverse we just take the cost:
https://github.com/valhalla/valhalla/blob/master/src/thor/dijkstras.cc#L370-L372

danpat:
Benchmark results using a full planet extract.  TL;DR - looks like any differences are under the noise floor, I'd say "no significant impact on performance".

### Before 17e9a3dffb0c1d42dee695698527e9edec2a8f85
```
------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
------------------------------------------------------------------------------------
BM_IsochroneUtrecht/1/repeats:10_mean          0.957 ms        0.956 ms           10
BM_IsochroneUtrecht/2/repeats:10_mean           1.22 ms         1.22 ms           10
BM_IsochroneUtrecht/4/repeats:10_mean           1.73 ms         1.73 ms           10
BM_IsochroneUtrecht/8/repeats:10_mean           3.37 ms         3.37 ms           10
BM_IsochroneUtrecht/16/repeats:10_mean          9.97 ms         9.96 ms           10
BM_IsochroneUtrecht/32/repeats:10_mean           888 ms          887 ms           10
BM_IsochroneUtrecht/64/repeats:10_mean          2289 ms         2288 ms           10
BM_IsochroneUtrecht/120/repeats:10_mean         6046 ms         6043 ms           10
```

### After 17e9a3dffb0c1d42dee695698527e9edec2a8f85
```
------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
------------------------------------------------------------------------------------
BM_IsochroneUtrecht/1/repeats:10_mean           1.05 ms         1.05 ms           10
BM_IsochroneUtrecht/2/repeats:10_mean           1.30 ms         1.30 ms           10
BM_IsochroneUtrecht/4/repeats:10_mean           1.89 ms         1.89 ms           10
BM_IsochroneUtrecht/8/repeats:10_mean           3.58 ms         3.57 ms           10
BM_IsochroneUtrecht/16/repeats:10_mean          9.55 ms         9.54 ms           10
BM_IsochroneUtrecht/32/repeats:10_mean           868 ms          868 ms           10
BM_IsochroneUtrecht/64/repeats:10_mean          2293 ms         2292 ms           10
BM_IsochroneUtrecht/120/repeats:10_mean         6074 ms         6071 ms           10
```

danpat:
Benchmark results for reachability (checks reachability of every edge in Utrecht):

### Before
```
BM_ReachUtrecht/repeats:10_mean         2713 ms         2696 ms           10
BM_ReachUtrecht/repeats:10_median       2663 ms         2661 ms           10
BM_ReachUtrecht/repeats:10_stddev        114 ms          103 ms           10
```

### After
```
BM_ReachUtrecht/repeats:10_mean         2696 ms         2695 ms           10
BM_ReachUtrecht/repeats:10_median       2689 ms         2688 ms           10
BM_ReachUtrecht/repeats:10_stddev       69.9 ms         69.2 ms           10
```

kevinkreiser:
@danpat can you bench something that is significantly larger than Utrecht? this may be an unpopular opinion but that graph is so small most of the deviation has to be in the setup/teardown or can you modify the bench to be able to take a different dataset and run longer routes on that? or maybe you could clarify a bit about what you meant by full planet extract, i didnt quite understand that.

kevinkreiser:
looks also like msvc doesnt like the syntax somewhere:

```
D:\a\1\s\src\thor\dijkstras.cc(172): error C2065: 'not': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(282): note: see reference to function template instantiation 'void valhalla::thor::Dijkstras::ExpandInner<valhalla::thor::ExpansionType::forward>(valhalla::baldr::GraphReader &,const valhalla::baldr::GraphId &,const valhalla::sif::BDEdgeLabel &,const uint32_t,const valhalla::baldr::DirectedEdge *,const bool,const valhalla::baldr::TimeInfo &)' being compiled
D:\a\1\s\src\thor\dijkstras.cc(172): error C2146: syntax error: missing ')' before identifier 'FORWARD'
D:\a\1\s\src\thor\dijkstras.cc(172): error C2059: syntax error: ')'
D:\a\1\s\src\thor\dijkstras.cc(173): error C2059: syntax error: ';'
D:\a\1\s\src\thor\dijkstras.cc(189): error C2065: 'opp_edge': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(189): error C2065: 't2': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(189): error C2065: 'oppedgeid': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(186): error C2789: 'allowed': an object of const-qualified type must be initialized
D:\a\1\s\src\thor\dijkstras.cc(186): note: see declaration of 'allowed'
D:\a\1\s\src\thor\dijkstras.cc(199): error C2065: 'opp_edge': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(199): error C2065: 't2': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(199): error C2065: 'oppedgeid': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(197): error C2789: 'allowed': an object of const-qualified type must be initialized
D:\a\1\s\src\thor\dijkstras.cc(197): note: see declaration of 'allowed'
D:\a\1\s\src\thor\dijkstras.cc(218): error C2065: 'opp_edge': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(222): error C2065: 'opp_edge': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(222): error C2065: 't2': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(245): error C2065: 't2': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(246): error C2065: 'oppedgeid': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(246): error C2065: 't2': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(250): warning C4267: 'initializing': conversion from 'size_t' to 'uint32_t', possible loss of data
D:\a\1\s\src\thor\dijkstras.cc(252): error C2065: 'oppedgeid': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(259): error C2065: 'opp_edge': undeclared identifier
D:\a\1\s\src\thor\dijkstras.cc(194): error C2044: illegal continue
D:\a\1\s\src\thor\dijkstras.cc(203): error C2044: illegal continue
D:\a\1\s\src\thor\dijkstras.cc(239): error C2044: illegal continue
D:\a\1\s\src\thor\dijkstras.cc(265): error C2059: syntax error: 'if'
D:\a\1\s\src\thor\dijkstras.cc(265): error C2143: syntax error: missing ';' before '{'
D:\a\1\s\src\thor\dijkstras.cc(265): error C2447: '{': missing function header (old-style formal list?)
D:\a\1\s\src\thor\dijkstras.cc(272): error C2059: syntax error: '}'
D:\a\1\s\src\thor\dijkstras.cc(272): error C2143: syntax error: missing ';' before '}'
D:\a\1\s\src\thor\dijkstras.cc(119): error C2143: syntax error: missing ';' before '{'
D:\a\1\s\src\thor\dijkstras.cc(119): error C2447: '{': missing function header (old-style formal list?)
```

kevinkreiser:
also a pleasant surprise that the standard deviation of reach is tightend

danpat:
> D:\a\1\s\src\thor\dijkstras.cc(172): error C2065: 'not': undeclared identifier

Dammit VC++ - I was trying to make the syntax more obvious, `!FORWARD` vs `not FORWARD` (the `not` operator is _supposed_ to be [perfectly valid C++](https://en.cppreference.com/w/cpp/language/operator_alternative)).

@kevinkreiser on the benchmark - yes, the isochrone numbers are from me running against a whole planet extract, not just utrecht (the utrecht isochrone maxes out at about 7 minutes, so you don't get useful numbers from the 120 minute benchmark on Utrecht).

For reach, I think utrecht is fine, as it tests reachability on every edge in utrecht.

danpat:
According to https://10xlearner.com/2020/11/04/how-to-use-alternative-operators-in-c-with-visual-studio/, one needs to add `/permissive-` to the VC++ compiler invocation.  I'm just going to go back to `!`...

valhalla/thor/dijkstras.h
dnesbitt61: Not sure why. Historically, the reverse direction required separating transition cost so that proper costing at the connection could be computed in FormPath. So it could have been omitted inadvertently from the reverse path when that was first changed.

kevinkreiser:
this looks good to me, all we need is a quick RAD (i feel pretty confident that it will pass given you didnt have to change the isochrones unit tests)

src/thor/dijkstras.cc
```diff
@@ -154,34 +159,71 @@ void Dijkstras::ExpandForward(GraphReader& graphreader,
...
+                                             offset_time.local_time, nodeinfo->timezone(),
+                                             restriction_idx);
+      if (!allowed || costing_->Restricted(directededge, pred, bdedgelabels_, tile, edgeid, true,
+                                           todo, offset_time.local_time, nodeinfo->timezone())) {
         continue;
       }
     } else {
-      if (!costing_->Allowed(directededge, pred, tile, edgeid, 0, 0, restriction_idx) ||
-          costing_->Restricted(directededge, pred, bdedgelabels_, tile, edgeid, true)) {
+      const bool allowed =
```
mandeepsandhu, line 197: Should we combine the `if(offset_time.valid) ...else` block (its a bit verbose now)?

```
const uint64_t current_time = offset_time.valid ? offset_time.local_time : 0;
const uint32_t tz_index = offset_time.valid ? nodeinfo->timezone() : 0;

const bool allowed =
          FORWARD ? costing_->Allowed(directededge, pred, tile, edgeid, current_time,
                                      tz_index, restriction_idx)
                  : costing_->AllowedReverse(directededge, pred, opp_edge, t2, oppedgeid,
                                             current_time, tz_index, restriction_idx);
if (!allowed || costing_->Restricted(directededge, pred, bdedgelabels_, tile, edgeid, true,
                                           todo, current_time, tz_index) {
...
```

mandeepsandhu (dismissed):
just 1 minor comment. otherwise lgtm thx for doing this! :+1:

danpat:
Using a full-planet extract, I ran this both against `master` and this branch:

```
./run_with_server.py --test-file ../test_requests/city_to_city/Australia_routes.txt  --concurrency 1
```

there were no differences in the output between the two branches.  This gives me some confidence that there has been no impact on reachability for normal routing.

valhalla/thor/dijkstras.h
danpat: Ok - I tried consolidating the logic here and just doing `+ transition_cost` instead of _just_ the `.cost` field.

I ran:

```
./run_with_server.py --test-file ../test_requests/city_to_city/Australia_routes.txt  --concurrency 1
```

and observed no different in the routes output, indicating no impact on reachability calculations.  I'm going to repeat this for reverse isochrones, and if it has no impact, I'm inclined to remove this difference and consolidate the forward/reverse behaviour here.

kevinkreiser:
Yep you should be able to do the same thing with some hacked up isochrones requests (just modify the json bodies in a python script or something) and if you set the format to raw the you can diff the results directly. Then id say we are 100% confident

danpat:
Ok - I tweaked `run_with_server` and added these changes:

```
--- run_with_server.py	2021-03-31 09:14:26.000000000 -0700
+++ run_isochrones_with_server.py	2021-04-01 15:26:15.000000000 -0700
@@ -18,6 +18,14 @@
       line = line[0:line.rfind('}') + 1]
       post_body = json.loads(line)
       post_body['id'] = str(line_number)
+
+      post_body["locations"] = [post_body["locations"][0]]
+      post_body["contours"] = [ { "time":15, "color":"ff0000" } ]
+
       yield post_body

 session = None
```

and ran a few thousand randomly selected routes against `master` and this branch.  There were no differences in the results.

valhalla/thor/dijkstras.h
danpat: Turns out we don't have reverse isochrones anyway :-)

My best guess based on testing and these comments is that the existence of this difference is an overlooked bug from a past age in the code, rather than a difference in forward/reverse traversal.

valhalla/thor/dijkstras.h
danpat: I benchmarked and tested the consolidated version of this - I see no difference in performance or behaviour.

src/thor/dijkstras.cc
danpat: The end goal is to merge this with *other* implementations of this same logic - I'd like to keep the code *mostly* the same for now to make it easier to diff against the other code.

mandeepsandhu (approved):

PR #2972, opened 2021-03-26
Map match fix for gps-point projecting exactly onto street intersection shape-point

ktatterso:
# Issue

Under the right conditions, if a gps-point's projection is exactly the same as a street-intersection shape-point, map-match will fail to match that point.

This issue is very much a "boundary case".  To achieve a gps-point projection onto a street-intersection shape-point is very rare!  Each point's lat/lon would have to be equal to 6-7 decimal places.  As a result, exposing/reproducing this issue in the real world can be difficult.

To expose the issue, the simplest approach is to have one of the gps-point's be the same as the street intersection's shape-point.  If the geometry and floating-point math allow, the gps-point projection will also be the same as the street intersection shape-point. This means the map-match logic will have achieved an exact node projection. Before my fix, one small part of the map-match logic was not prepared for this possibility.  (Fortunately, the downstream map-match logic handles node projection correctly.)

## Tasklist

 - [x] Add tests


test/gurka/test_match.cc
```diff
@@ -126,6 +126,56 @@ D--3--4--C--5--6--E)";
...
+      |
+      |
+      |
+      |
+      |
+      E)";
+
+  // Using specific geometry that sorta represents the above but is
+  // required to demonstrate the fix.
+  gurka::nodelayout layout; // = gurka::detail::map_to_coordinates(ascii_map, 10);
```
kevinkreiser, line 147: clever workaround!

kevinkreiser:
> map-match will fail to match that point.

> Before my fix, one small part of the map-match logic was not prepared for this possibility.

can you add a bit more detail about what goes wrong here? the code change seems to suggest that in master, we cant properly create a "real" (ie has an edge id) `MatchResult` for the input point in question? there is a bunch of handling of node snapping in there but the change seems to suggest that you are getting into the special case where we just end up using any edge that is on the candidate because it was a node to node trivial route. the other possiblity im thinking is that since the first input point is a node snap (or it is now that we added floating point fuzz) we cant go back any further to find an edge id and so we have a special case.

i guess my ask is can you walk us over the flow of the code in the failure scenario?

id also like to take the opportunity to once again lament that node routing in meili is so unfortunate. we need it because it drastically reduces the permutation space of routes that need to be computed but my god is it a pain in the butt to handle downstream :smile:

PR #2975, opened 2021-03-30
Adding roundabout_exit_shape_index to maneuver for roundabouts

kdiluca:
# Issue

This PR will allow us to set the roundabout heading of the line segment that join enter & exit points on a curve.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/proto/directions.proto
+++ b/proto/directions.proto
@@ -198,6 +198,7 @@ message DirectionsLeg {
     optional bool has_time_restrictions = 34;                // Whether edge has any time restrictions or not
     repeated GuidanceView guidance_views = 35;               // List of guidance views
     optional BssManeuverType bss_maneuver_type = 36;
+    optional uint32 roundabout_chord_heading = 37;           // The heading of the line segment that join enter & exit points on a curve
   }
 
   optional uint64 trip_id = 1;
--- a/proto/directions.proto
+++ b/proto/directions.proto
@@ -198,7 +198,6 @@ message DirectionsLeg {
     optional bool has_time_restrictions = 34;                // Whether edge has any time restrictions or not
     repeated GuidanceView guidance_views = 35;               // List of guidance views
     optional BssManeuverType bss_maneuver_type = 36;
-    optional uint32 roundabout_chord_heading = 37;           // The heading of the line segment that join enter & exit points on a curve
   }
 
   optional uint64 trip_id = 1;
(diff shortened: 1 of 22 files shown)


src/odin/maneuversbuilder.cc
```diff
@@ -2701,6 +2701,9 @@ bool ManeuversBuilder::AreRoundaboutsProcessable(const TripLeg_TravelMode travel
 }
 
 void ManeuversBuilder::ProcessRoundabouts(std::list<Maneuver>& maneuvers) {
+  std::vector<midgard::PointLL> shape =
```
dgearhart, line 2704: originally i was thinking about assigning in triplegbuilder so we do not have to decode shape again - but I do not think that is possible

proto/directions.proto
```diff
@@ -198,6 +198,7 @@ message DirectionsLeg {
     optional bool has_time_restrictions = 34;                // Whether edge has any time restrictions or not
     repeated GuidanceView guidance_views = 35;               // List of guidance views
     optional BssManeuverType bss_maneuver_type = 36;
+    optional uint32 roundabout_chord_heading = 37;           // The heading of the line segment that join enter & exit points on a curve
```
dgearhart, line 201: after reviewing - we can not store on the edge since the roundabout can have multiple edges

dgearhart, line 201: do not need to store in proto

src/odin/directionsbuilder.cc
```diff
@@ -189,6 +189,7 @@ void DirectionsBuilder::PopulateDirectionsLeg(const Options& options,
     trip_maneuver->set_turn_degree(maneuver.turn_degree());
     trip_maneuver->set_begin_shape_index(maneuver.begin_shape_index());
     trip_maneuver->set_end_shape_index(maneuver.end_shape_index());
+
```
dgearhart, line 192: this change to file is not needed

test/maneuversbuilder.cc
```diff
@@ -1927,6 +1928,7 @@ TEST(Maneuversbuilder, TestCombineRoundaboutManeuvers) {
   expected_maneuver2.set_roundabout_length(1.0);
   expected_maneuver2.set_roundabout_exit_length(2.0);
   expected_maneuver2.set_roundabout_exit_turn_degree(90);
+  expected_maneuver2.set_roundabout_exit_shape_index(0);
```
dgearhart, line 1931: zero does not make sense here

dgearhart:
just need to update the `TestCombineRoundaboutManeuvers` method to have shape indexes to validate the `roundabout_exit_shape_index` when an exit roundabout is collapsed

dgearhart (approved):
LGTM - when CI passes :ship:

PR #2977, opened 2021-03-31
Map match fix for when starting gps-point is exactly the same lat/lon as a street shape-point

ktatterso:
# Issue

The existing logic in meili map-matching `FindMatchResult()` was written to expect the following (here's the code comment):  _the very first state will have an initial dummy label with invalid edge/node ids_.  This isn't always true.

If the very first gps point exactly matches a street shape-point, then the first state will have an invalid edge-id and a valid node-id.

The fix is simple.  When you look at the diffs you'll see that the new terminating condition for state/label iteration is to simply check if the edge-id is null (instead of both edge-id and node-id).  Previously we would iterate one step too far and "lose" the edge-id.  Now we terminate the iteration at the right moment, keeping the edge-id in a valid state.  

Also, for both the forward/backward state iteration, the node-id is initialized in the same manner:  before the terminating condition for the edge-id being invalid.

## Tasklist

 - [x] Add tests


src/meili/map_matcher.cc
```diff
@@ -281,9 +281,12 @@ MatchResult FindMatchResult(const MapMatcher& mapmatcher,
       break;
     node_id = rbegin->nodeid();
     while (rbegin != rend && !prev_edge.Is_Valid()) {
-      // the very first state will have an initial dummy label with invalid edge/node ids
+      // the very first state will have an initial dummy label with invalid edge/node ids -
+      // except when the first state is a node, in which case the edge-id will be invalid
```
kevinkreiser, line 285: well that is unexpected for sure! this is part of the contract that `meili/routing.cc` provides to the rest of the code. did you happen to have a look there to see why it doesnt put the initial dummy label in the labelset. specifically this method: https://github.com/valhalla/valhalla/blob/master/src/meili/routing.cc#L176 is supposed to do what the comment here (and also over there) says. very strange find indeed!

PureW (dismissed):
LGTM, Nice find!

kevinkreiser (dismissed):
i really think we should at least check why there isnt an initial dummy label in the labelset. if there isnt that means the code is broken upstream (routing.cc) or maybe its a sign that the code thinks that there is a trivial node to node route right at the beginning for some reason and it doesnt have any labels or something (though that wouldnt match with your fix..)? anyway im not saying the fix isnt ok, it looks ok for this case but i am saying we really should understand why the upstream code isnt behaving as advertised. all over the code we expect there to be a dummy label at the beginning of the matching. its a hard requirement at this point.

ktatterso:
@PureW @kevinkreiser I was wondering about the initial state thing too... but there are comments in the code indicating that routes that begin with a node are possible and should be expected.  See this comment on 276:  "Normally the last label of the route from previous to current state has a valid edge id. But it wont if the destination (current state candidate) was a node. In this case it will have a valid node id".  And this comment on line 316:  `But its possible the origin (current state candidate) was a node which means the first label will only be a nodeid.`. 

Hence why I coded the logic to expect this scenario and didn't dig back further.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@
    * FIXED: Dont abort bidirectional a-star search if only one direction is exhausted [#2936](https://github.com/valhalla/valhalla/pull/2936)
    * FIXED: Fixed missing comma in the scripts/valhalla_build_config [#2963](https://github.com/valhalla/valhalla/pull/2963)
    * FIXED: Reverse and Multimodal Isochrones were returning forward results [#2967](https://github.com/valhalla/valhalla/pull/2967)
+   * FIXED: Map-match fix for first gps-point being exactly equal to street shape-point [#2977](https://github.com/valhalla/valhalla/pull/2977)
+
```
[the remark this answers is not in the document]
ktatterso: typo, thx

src/meili/map_matcher.cc
ktatterso: I can have a look, but also see my comment about this being something the previous implementor thought to expect but didn't (author: Malakas).

kevinkreiser:
@ktatterso good point! the comments are potentially a little bit imprecise for sure. specifically this one you pointed out:

> was a node which means the first label will only be a nodeid

That is refering to the label at any state, which indeed could be just a nodeid, but specific to the very first state, the very first label should be a dummy label iirc. but maybe there is something in routing.cc that somehow violates that or some other bit of code that pops that label off? anyway i just wanted to ask if you could take a quick look to see if there ever was a dummy label and it get removed or if there never is a dummy label in this specific case. sorry for the confusion! this kind of stuff is quite hard to communicate about over text :smile:

src/meili/map_matcher.cc
ktatterso: So I took a look at routing.cc, `set_origin`.  I also debugged it using my test case.  Based on the way the code is written, it seems very intentional that the origin state can either be 1) a dummy state (both node and edge-ids are invalid) or 2) a state with a valid node-id and invalid edge-id.  Maybe we can investigate this further when we spend some time on tech-debt, as I agree it would be much easier to understand and code against if the origin state was always a dummy state.

kevinkreiser:
ok i feel like my point is being missed here :smile: 

> I can have a look, but also see my comment about this being something the previous implementor thought to expect but didn't (author: Malakas).

yes @yuzheyan has some commits here. i paired with him to refactor the code a bit last summer. prior to that @ptpt and i wrote the first iteration of the code together. i'm well aware of the contracts of the different methods in meili. i'm not exactly sure how to prove it to you but trust me that i know how the code works (or is supposed to work haha).

> So I took a look at routing.cc, set_origin. I also debugged it using my test case. Based on the way the code is written, it seems very intentional that the origin state can either be 1) a dummy state (both node and edge-ids are invalid) or 2) a state with a valid node-id and invalid edge-id. Maybe we can investigate this further when we spend some time on tech-debt, as I agree it would be much easier to understand and code against if the origin state was always a dummy state.

Let me try to explain the code a bit.. because what you are saying seems to be missing the point of the dummy label. What im trying to do here is make sure we avoid a bandaid when we could just figure out the root of the problem. The dummy label is there to tell you that there is no path to the "state" that you are currently evaluating. In map matching we use the HMM approach in which each trace point becomes a series of  "state" (candidates) in a markov chain.

A state candiate will have a label set associated with it. this label set tells you how to get from that candidate in the current state to a candiate in the next state. State candidates whose label sets begin with a valid label (ie non-dummy) are those state candidates which have a path leading into them. Those that have a dummy label indicate that there is no path leading to that state candidate.

If you are seeing this happen on the first state candidate, then that seems impossible. Certainly there cannot be a route leading to the first gps point since there are no gps points prior to it. This point should always have a dummy label because there is no prior route and therefor there should be a `nullptr` passed in for the edgelabel that seeds the origin. Also note that if there is a break/discontinuity in the map match there will also be a dummy label (because again no route was found leading into this state candidate).

So looking at the code I see: https://github.com/valhalla/valhalla/blob/master/valhalla/meili/routing.h#L191 which has a comment in there that tries to spell this out. It basically says if the edge label (which is the last label of the previous route to this state candidate) is null (meaning there was no route to here, which there shouldnt be for the first state candidate) then we add a dummy label (ie to mark that this is the beginning of the route or will be the beginning of a route after a discontinuity/break in the map match).

You can also have a look at https://github.com/valhalla/valhalla/blob/master/src/meili/transition_cost_model.cc#L93-L95. This is the spot in the code before we try to do the 1 to many routing in routing.cc. Notice that we set the edgelabel to null when there isnt a previous state (which is definitely the case for the first point/state in the route) OR when the previous state didnt find a route to this state candidate (whic is the case when no path exists between two states).

So anyway, have I misunderstood what you are saying? I thought we were talking about the first state in the route. Reviewing the code seems pretty clear to me that the first one should always have a dummy label, even if its a node candidate rather than an edge candidate. Where am I misundersatnding you here?

mookerji:
@kevinkreiser terminology clarification: I initially thought that a dummy node in the Viterbi graph was actually the first and last virtual nodes (see below). Are the 'dummy' nodes here actually nodes we've pruned when searching through the graph since there isn't likely to be a Viterbi path/solution for the sequence?

image

kevinkreiser:
@mookerji no we are talking about dummy edge labels in this case. so each of the states (columns of circles) has 0-n candidates (rows). when one of those candidates is on the left hand side of a route being performed in routing.cc it may or may not have a dummy edge label as its first label in the label set. this is because there is no path to it, which is always the case for the first state's candidates and can be the case for subsequent states' candidates (if there is no path to them because of graph connectivity or the distances is too far or the time information is wrong or or .. you get the point).

kevinkreiser:
when i revert the fix in map_matcher.cc to master and i run the new unit test i see the following:

image

image

So indeed, it is missing the shape point at the beginning for `B`, the result is instead a map match that just goes from C to F.

And indeed, somehow the first label of the label chain for the first state does have a valid nodeid and an invalid edgeid...

image

Ah wierd... i see when we add the label we add the dummy label but then we overwrite it node id here: https://github.com/valhalla/valhalla/blob/master/valhalla/meili/routing.h#L134

Which we do for nodes: https://github.com/valhalla/valhalla/blob/master/valhalla/meili/routing.h#L200

But for edges we use an invalid id: https://github.com/valhalla/valhalla/blob/master/valhalla/meili/routing.h#L183

Ah i see this is because for nodes we using an invalid destation index and instead store the nodeid as the destination to look up in the `node_status_` member. Its basically serving double duty.

kevinkreiser:
i wonder if we can exploit the predecessor index here like we do in all other algorithms when we are traversing label sets. basically those algorithms all just loop until they hit `kInvalidLabel`.

There are two loops in `FindMatchResult` the first one needs to find the last edge of the path leading into the current state candidate and the second one needs to find the first edge leaving the current state candidate. That means the first loop should basically always stop in the first or second iteration. The second loop needs to go all the way back to the beginning of the  label set so it must iterate the whole path between the current state candidate and the next potentially stopping one short if the current state candidate is a node candidate. Perhaps we could update the code along those lines?

Just trying to get to the bottom of what is happening here!

ktatterso:
@kevinkreiser Ok I'm gonna need some time to consume all this.  thanks.

kevinkreiser:
I think this means we can stop the first loop as soon as we find a valid edge id (and dont hit a discontinuity) so that loop can be simplified to not even check for dummy labels.

The second loop which walks the path backwards from the next state i think can look like the PR @ktatterso made and closed previously where we just remember any valid edge id we see while walking the path backwards to potentially a node candidate.

More specifically i'm suggesting this, which ive tested an passes all the unit tests on my machine:

```diff
diff --git a/src/meili/map_matcher.cc b/src/meili/map_matcher.cc
index 06281d044..70b711346 100644
--- a/src/meili/map_matcher.cc
+++ b/src/meili/map_matcher.cc
@@ -281,10 +281,7 @@ MatchResult FindMatchResult(const MapMatcher& mapmatcher,
       break;
     node_id = rbegin->nodeid();
     while (rbegin != rend && !prev_edge.Is_Valid()) {
-      // the very first state will have an initial dummy label with invalid edge/node ids
       const auto& label = *rbegin;
-      if (!label.edgeid().Is_Valid() && !label.nodeid().Is_Valid())
-        break;
       prev_edge = label.edgeid();
       rbegin++;
     }
@@ -311,7 +308,7 @@ MatchResult FindMatchResult(const MapMatcher& mapmatcher,
     const auto& state = mapmatcher.state_container().state(state_id);
     // Normally we need to loop back to the first label of the route from current to the next state
     // and get its edge id. But its possible the origin (current state candidate) was a node which
-    // means the first label will only be a nodeid.  If we didnt find any edges after that (which is
+    // means the first label will only be a nodeid. If we didnt find any edges after that (which is
     // the case in a trivial node to node route) then we need to keep looking at more states into the
     // future or until we hit a discontinuity
     auto rbegin = state.RouteBegin(next_state), rend = state.RouteEnd();
@@ -320,10 +317,10 @@ MatchResult FindMatchResult(const MapMatcher& mapmatcher,
     while (rbegin != rend) {
       // the very first state will have an initial dummy label with invalid edge/node ids
       const auto& label = *rbegin;
-      if (!label.edgeid().Is_Valid() && !label.nodeid().Is_Valid())
-        break;
-      next_edge = label.edgeid();
-      node_id = label.nodeid();
+      if (label.edgeid().Is_Valid())
+        next_edge = label.edgeid();
+      if (label.nodeid().Is_Valid())
+        node_id = label.nodeid();
       rbegin++;
     }
   }
```

test/gurka/test_match.cc
```diff
@@ -126,6 +126,45 @@ D--3--4--C--5--6--E)";
...
+  const gurka::ways ways = {
+      {"BC", {{"highway", "primary"}}},
+      {"CF", {{"highway", "primary"}}},
+      {"FD", {{"highway", "primary"}}},
+  };
+
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/mapmatch_node_snapping");
+
+  auto result = gurka::do_action(valhalla::Options::trace_route, map, {"B", "1", "F"}, "auto",
+                                 {{"/trace_options/search_radius", "50"}}, {}, nullptr, "via");
```
kevinkreiser, line 156: nit: this confused me. the grid size in gurka is only 10 meters. by setting the search radius to 50 i would expect to see at least 2 candidates for your first input point `B` since `C` is also within 50 meters of `B`

kevinkreiser:
@ktatterso no worries at all, there is a ton to consume tbh, sorry for the long wind, but please let me know if any of it doesnt make sense or im talking non-sense here!

test/gurka/test_match.cc
ktatterso: Honestly I don't remember why I added the 50 m search radius, I'll take another look.

ktatterso:
@kevinkreiser Ok we've kinda come full circle (ish) with your code example above (effectively coming back to same style of fix in my initial PR).  And I'm a little confused as to what to do next.

You initial concern with this PR was that there is a "bigger issue" that this PR is merely band-aiding.  After debugging do you still believe there is a bigger issue?  I ask because your provided solution that "doesn't have to check for dummy labels" seems to simply say, "regardless of how we've implemented dummy labels this solution should work."

kevinkreiser:
Yep so I think after the code investigation we found 2 things.

1. the code in `FindMatchResult` had the wrong notion of what an initial (dummy) edge label should look like, in fact there are 2 styles, one where the destination index is set and one where the nodeid is set (to help the routing algorithm identify which is which. but critically both have a predecessor index of `kInvalidLabel` (as do all the routing algorithms, so not a surprise)
2. the code in `FindMatchResult` has no business caring about the "dumminess" of a label, it should do what its really trying to do, that is: for the first loop, find the last edge in the path to the current state and for the second loop find the first edge in the path leaving the current state. in other words both loops are bounded by the number of labels between the adjacent states, which is handled by the custom iterator (`RoutePathIterator`) for the labels that correctly bounds the iteration by looking for `kInvalidLabel` index on the `predecessor` of the label to tell it when it was the last label.

> do you still believe there is a bigger issue?

Nope. But I do believe that without debugging and showing the work no one could make a call as to whether or not the fix was sound. We needed the refresher of how it all works together before we can say with confidence we are doing the right thing in the code.

kevinkreiser:
Oh and also we should run a bunch of map matching requests through this code before merging. The unit tests do exorcise many cases but there are just so many cases that it warrents a bit of a "RAD" style test before merging.

kevinkreiser (approved):
@ktatterso i very much appreciate you wading through this with me. i hope that by getting into and figuring out all the nooks and cranies and presenting them as part of the PR we can make reviewing easier for all parties.

also i would highly recommend running large swathes of mapmatching requests through this just to be double sure. thanks again!

ktatterso:
@kevinkreiser I have ran a couple three dozen "real world" map-match requests manually and they look good.  This coming monday I plan to move these changes into staging for a bit to ensure we don't see anything unexpected.

test/gurka/test_match.cc
ktatterso: Yea I removed the 50 m radius, that didn't make sense.

ptpt:
@kevinkreiser @ktatterso I don't know how to fix the bug here yet, but it's likely to be related to misunderstanding of the labels used in meili routing, and it's all because of absence of docs here. To be honest it is very hard to understand the code without any docs -- I had at least 2 cups of coffee to catch up the details and write the following up. I hope it helpful to you.

tl;dr: check the invariants and "The dummy label" that might answer some of the questions in the thread.

------

## The Label
`Label` is the core structure used in meili's node-to-node routing algorithm. It serves two purposes: internally it's used to remember the search state, and externally it's used to reconstruct the route from the origin to any destination.

By node-to-node it means the most important information in a label is the nodeid, and the predecessor (that points to its previous label). However a single nodeid is not enough to reconstruct the route because there might be multiple edges between the same pair of nodes. Therefore we added edgeid to indicate which edge it comes from the previous node to the current node. So a label is more closely resemble an edge that ends at the node, than the node itself.

So far a label should the following shape:

* **nodeid**: the current node
* **edgeid**: which edge it comes from the previous node to the current node
* **predecessor**: the previous label that contains the previous node

However, in real world scenario, the origin and the destinations usually start/end at points along edges, instead of the routing algorithm's preferred nodes. We need to handle these cases. Meili's approach is treat these non-node origin/destinations as special "nodes" by introducing `dest` to distinguish the special "nodes" from the real nodes:

* **dest**: index of the destination

Note that dest=0 is a special destination reserved for the origin.

So a label with dest=1 and edgeid=AB indicates that the label ends at the first destination on edge AB. Since the label only represents part of the edge AB, so we introduce source and target. In this case the label has target=0.5, if the destination is at the middle of the edge, and source could be anywhere between [0, 0.5], depending on where the predecessor ended at. See the examples.

* **source**: the label's start location, as percent, along the edge
* **target**: the label's end location, as percent, along the edge

As you might see, the introduce of the special nodes makes the structure Label extremely complex as well as the routing implementation, as it has to handle two types of nodes separately. Here are some invariants that might be helpful to understand the code:

1. A label has either a valid node ID (real node) or a valid dest (special node), i.e. nodeid is valid XOR dest is valid. This holds for all labels, including dummy labels.
2. 0 <= source <= target <= 1
3. All labels have valid edgeids, except for the dummy label
4. Only the dummy label has invalid predecessor, while the other labels always have valid predecessor

## The dummy label

Dummy label is the label that ends at the origin with an invalid predecessor. We don't know and also don't care what edgeid/source/target it has. We only care its dest or nodeid. The dummy label is the first label that get pushed to the priority queue, and then poped out. It serves as root in the search tree.

When iterating the labels from an destination backwards to the origin with `RoutePathIterator`, the dummy label is always exposed at the end of the route. It's likely to be intentional because `RoutePathIterator` uses the dummy label to tell the caller if there is a route between them by checking the initial `rbegin == rend`. It's quite safe to skip the dummy label after you make sure there is a route between them. Check [`MergeRoute`](https://github.com/valhalla/valhalla/blob/47939fbf4b27d2a5ad262e2e521d3361779faedc/src/meili/match_route.cc#L88-L105).

## Examples
Here are 4 examples to help understand the internal label states given an input (graph/origin/destinations). In the input, letters are nodes, while numbers are origin/destinations where 0 is always the origin. The label table shows the final states of all labels. The search tree is constructed from the labels.

### Start/end on edges
Input:
```
A ----0----> B ---1---2---> C
```

Labels:
| Label Index | Destination Index | Node ID | Edge ID | Source | Target | Predecessor | Comment |
| ----------- |  ----------- | ------|----------- |  ----------- | ------|------|------|
| 0      | 0       | Invalid | Invalid | 0 | 0 | Invalid| Dummy label |
| 1      | Invalid       | B | AB | 0.5 | 1 | 0| 0 -> B |
| 2      | 1       | Invalid | BC | 0 | 0.3 | 1| B -> 1 |
| 3      | 2       | Invalid | BC | 0 | 0.6 | 1| B -> 2 |

The search tree constructed by the labels (the numbers are label indices):
```
      0
      |
      1
     / \
    2   3
```

### Start/end at the same point on edges
Input:
Origin 0 and destination 1 are at the same point on AB.
```
A ----0,1----> B ----2-----> C
```

Labels:
| Label Index | Destination Index | Node ID | Edge ID | Source | Target | Predecessor |  Comment |
| ----------- |  ----------- | ------|----------- |  ----------- | ------|------| ------|
| 0      | 0                | Invalid | Invalid | 0    | 0 | Invalid| Dummy label |
| 1      | 1        | Invalid          | AB | 0.5 | 0.5 | 0| 0 -> 1 (same point) |
| 2      | Invalid       | B          | AB | 0.5  | 1 | 0| 0 -> B |
| 3      | 2                | Invalid | BC | 0 | 0.4 | 2| B -> 2 |

The search tree constructed by the labels:
```
    0
   /  \
  1    2
         \
          3
```

### Start/end at nodes
Input:
Origin 0 and destination 1&2 are at nodes A, B, C respectively.
```
A,0 ------> B,1 ------> C,2
```

Labels:
| Label Index | Destination Index | Node ID | Edge ID | Source | Target | Predecessor | Comment |
| ----------- |  ----------- | ------|----------- |  ----------- | ------|------|-------|
| 0                 | Invalid                | A                  | Invalid | 0    | 0 | Invalid| Dummy label |
| 1                 | Invalid                 | B          | AB | 0 | 1 | 0| A -> B |
| 2                | Invalid                 | C          | BC | 0  | 1 | 1| B -> C |

The search tree constructed by the labels:
```
     0
     |
     1
     |
     2
```

### Start/end at the same nodes
Input:
Origin 0 and destination 1 are at the same node A.
```
A,0,1 ------> B ------> C,2
```

Labels:
| Label Index | Destination Index | Node ID | Edge ID | Source | Target | Predecessor |  Comment |
| ----------- |  ----------- | ------|----------- |  ----------- | ------|------|------|
| 0                 | Invalid                 | A                  | Invalid | 0    | 0 | Invalid| Dummy label |
| 1                  | Invalid                 | A                  | AB | 0 | 0 | 0| A -> A (same point) |
| 2                 | Invalid                 | B                  | AB | 0  | 1 | 0| A -> B |
| 3                 | Invalid                 | C                  | BC | 0  | 1 | 2| B -> C |


The search tree constructed by the labels:
```
     0
    /   \
   1     2
           \
            3
```

src/meili/map_matcher.cc
```diff
@@ -318,12 +315,11 @@ MatchResult FindMatchResult(const MapMatcher& mapmatcher,
     if (rbegin == rend)
       break;
     while (rbegin != rend) {
-      // the very first state will have an initial dummy label with invalid edge/node ids
       const auto& label = *rbegin;
-      if (!label.edgeid().Is_Valid() && !label.nodeid().Is_Valid())
-        break;
-      next_edge = label.edgeid();
-      node_id = label.nodeid();
+      if (label.nodeid().Is_Valid())
```
ptpt, line 319: I don't understand why it has to iterate the whole route here. Previously it just returned the rbegin's end location (rbegin->edgeid() and rbegin->target). But if it has to iterate the whole route then check how MergeRoute does:
```c++
  for (; std::next(rbegin) != rend; rbegin++) {
    // nodeid might be invalid if the label ends at an destination on the edge
    if (rbegin->nodeid().IsValid()) {node_id = rbegin->nodeid();}
    // edgeid is always valid
    assert rbegin->edgeid()->IsValid();
    next_edge = rbegin->edgeid();
  }
```

src/meili/map_matcher.cc
```diff
@@ -281,10 +281,7 @@ MatchResult FindMatchResult(const MapMatcher& mapmatcher,
       break;
     node_id = rbegin->nodeid();
     while (rbegin != rend && !prev_edge.Is_Valid()) {
-      // the very first state will have an initial dummy label with invalid edge/node ids
```
ptpt, line 284: This comment is actually true, but the loop doesn't look right. Try:
```c++
  for (; std::next(rbegin) != rend; rbegin++) {

// edgeid is always valid
    assert rbegin->edgeid()->IsValid();
    prev_edge = rbegin->edgeid();
  }
```

src/meili/map_matcher.cc
kevinkreiser: @ptpt the larger comment blocks above try to hint about why this is the case. the short answer is, because of node routing and if `interpolation_distance` is `0` you can get the case where you have situation where there are two adjacent points in the input which both snapped to the same graph node. in this case the routing will return a labelset with just one dummy node label representing the trivial node to node route where you route from one node to the same node haha. the problem with that is that this function needs to return an edge id for the part of the path that this match result matched to. we cant do that with a nodeid, so what we do is we iterate forward or backward in states so that we can find the actual path that connected to the node and use that edge id. its like the comments say above, its rare but we need to do it in special cases of trivial node to node routes

src/meili/map_matcher.cc
kevinkreiser: the edgeid will not be vaild when the route is a trivial node to node route as i mentioned below. its a very annoying case unfortunately :frowning_face:

kevinkreiser:
@ptpt we should move your comments to a md file in the docs folder :smile:

src/meili/map_matcher.cc
ptpt: @kevinkreiser Aha I see, that's indeed a unfortunate case. In this case there is no way to find the match edge, right? It is like another unfortunate case: map match a single point that snaps to a node -- all we could do is just take one of the edges that starting from the node as the match?

Let me go through `FindMatchResult` with this in mind first and see how it's handled and what I missed.

kevinkreiser: Yep, I believe the suggested change works for all cases, the node to same node case and it also handles traversing all the labels back to the original dummy label because inside of the `routepathiterator` we are looking for invalid predecessor index. we could detect the dummy label if we wanted to and stop one iteration earlier in the case of node routes but i think this one more iteration doesnt matter. i like the simplicity of the changes proposed here. though to your comment below i do think a `for` loop would be nicer than a `while` loop if others are so inclined

ktatterso:
I deployed these changes into our staging stacks for 24 hrs.  Our cloudwatch metrics showed no significant changes in any direction (good or bad).  Merging as soon as the pipeline finishes.

PR #2978, opened 2021-04-01
Allow expansion into a region when not_thru_pruning_ is false on 2nd pass

gknisely:
# Issue

This PR resolves an edge case where we may encounter only_restrictions with edges being marked as not_thru.  Basically the only way to get in this area is via this edge based on the restriction, but it is also marked as not_thru. To resolve this issue, we set the not_thru to false for all edges only if the global `not_thru_pruning_` is false. `not_thru_pruning_` is only set to false on the 2nd pass in route_action. See the gurka test `not_thru_pruning_` as it is based on the following real world example:

This [way/edge](https://www.openstreetmap.org/way/128573260#map=18/45.50834/-122.74235) is marked as a not_thru edge.

Screenshot from 2021-04-06 22-03-51

But the only way to get into this area is this way/edge due to the [only restriction](https://www.openstreetmap.org/relation/1838192).  

Screenshot from 2021-04-06 22-04-10

So basically in the forward or reverse expansion in bidirectional A* we stop expanding due to the edge being marked as not_thru.
I tried removing logic to not mark this edge as not_thru in Mjolnir; however, the [following edge](https://www.openstreetmap.org/way/136459355) is also marked as not_thru so we would just uturn on the [original edge ](https://www.openstreetmap.org/way/128573260#map=18/45.50834/-122.74235) and still not traverse into this area.      
   

## Tasklist

 - [x] Add tests

## Requirements / Relations


scripts/valhalla_build_config
```diff
@@ -355,7 +355,7 @@ help_text = {
     'service': {
       'proxy': 'IPC linux domain socket file location'
     },
-    'max_reserved_labels_count': 'Maximum capacity for edge labels reserved in path algorithm'
+    'max_reserved_labels_count': 'Maximum capacity for edge labels reserved in path algorithm',
```
mandeepsandhu, line 358: this was fixed in https://github.com/valhalla/valhalla/commit/4d9d106608b354d60f4d77b4e3501d6692bbe514

src/thor/bidirectional_astar.cc
```diff
@@ -324,11 +326,21 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
   float sortcost =
       newcost.cost + astarheuristic_forward_.Get(t2->get_node_ll(meta.edge->endnode()), dist);
 
+  // There is an edge case where we may encounter only_restrictions with edges being
+  // marked as not_thru.  Basically the only way to get in this area is via this edge
+  // based on the restriction, but it is also marked as not_thru. We set the not_thru
+  // to false here only if allow_restricted_thru is true and we are not restricted.
+  // This bool is only set to true on the 2nd pass in route_action. See the gurka
+  // test allow_restricted_thru
+  bool thru = (pred.not_thru_pruning() || !meta.edge->not_thru());
```
mandeepsandhu, line 335: minor nit: I think a var name of `not_thru_pruning` would be more appropriate since it reflects the `not_thru` pruning state of the current edge label (i.e whether we're culling-out not_thru edges or not)

src/thor/timedep_reverse.cc
```diff
@@ -37,6 +37,11 @@ void TimeDepReverse::Clear() {
   }
   edgelabels_rev_.clear();
   adjacencylist_rev_.clear();
+
+  // Set the ferry flag to false
+  has_ferry_ = false;
```
mandeepsandhu, line 42: I assume this isn't related to the not_thru issue? the `has_ferry` flag was simply missing from this implementation?

scripts/valhalla_build_config
gknisely: @mandeepsandhu thanks

src/thor/timedep_reverse.cc
gknisely: @mandeepsandhu correct.

gknisely: cc @danpat note the missing has_ferry

src/thor/bidirectional_astar.cc
```diff
@@ -549,11 +561,17 @@ inline bool BidirectionalAStar::ExpandReverseInner(GraphReader& graphreader,
   float sortcost =
       newcost.cost + astarheuristic_reverse_.Get(t2->get_node_ll(meta.edge->endnode()), dist);
 
+  // There is an edge case where we may encounter only_restrictions with edges being
+  // marked as not_thru.  Basically the only way to get in this area is via this edge
+  // based on the restriction, but it is also marked as not_thru. We set the not_thru
+  // to false here only if allow_restricted_thru is true.  This bool is only set to true
+  // on the 2nd pass in route_action.  See the gurka test allow_restricted_thru
+  bool thru = (allow_restricted_thru() ? false : (pred.not_thru_pruning() || !meta.edge->not_thru()));
```
mandeepsandhu, line 569: Why is the condition being checked different than the forward direction? Here we're switching off pruning without checking if the predecessor edge has restriction or not (i.e using `pred.restrictions()`).

test/gurka/test_bidir_search.cc
```diff
@@ -40,22 +40,15 @@ TEST(StandAlone, exhaust_reverse_search) {
...
-      {
-        try {
-          auto result = gurka::do_action(valhalla::Options::route, map, {"A", "F"}, "auto", {});
-        } catch (const std::exception& e) {
-          EXPECT_STREQ("No path could be found for input", e.what());
-          throw;
-        }
-      },
-      std::exception);
+  auto result = gurka::do_action(valhalla::Options::route, map, {"A", "F"}, "auto", {});
```
mandeepsandhu, line 46: hmmm, why is not failing to find a route anymore with your change? This is a simple `not_thru` case and the previous open edge does not have any restriction on it. I thought this case would still throw a no-route exception since there's no other restriction involved.

src/thor/bidirectional_astar.cc
gknisely: @mandeepsandhu unfortunately we can't.  I will explain when I fill in the details above.

test/gurka/test_bidir_search.cc
gknisely: @mandeepsandhu in the reverse search not_thru is not considered anymore on the 2nd pass

src/thor/bidirectional_astar.cc
dnesbitt61: I would simply have a not_thru_pruning option that is true on the first pass and false on the 2nd pass. No need (in my opinion) to check for restrictions - just disable all not_thru pruning on the 2nd pass. This was mostly added as a performance optimization for cycling and pedestrian (where hierarchies are not useful).

src/thor/bidirectional_astar.cc
dnesbitt61: I would call this not_thru_pruning and not confuse the issue with the restriction name.

src/thor/bidirectional_astar.cc
```diff
@@ -324,11 +326,17 @@ inline bool BidirectionalAStar::ExpandForwardInner(GraphReader& graphreader,
   float sortcost =
       newcost.cost + astarheuristic_forward_.Get(t2->get_node_ll(meta.edge->endnode()), dist);
 
+  // There is an edge case where we may encounter only_restrictions with edges being
+  // marked as not_thru.  Basically the only way to get in this area is via this edge
```
dnesbitt61, line 330: confusing description. The edge that is restricted is not marked as not_thru. I guess the "only_restriction" means the path can only go onto the not_thru edge, but Valhalla doesn't really have the notion of "only" - it just has edges that are restricted / not allowed.

src/thor/multimodal.cc
```diff
@@ -92,6 +92,8 @@ void MultiModalPathAlgorithm::Clear() {
 
   // Set the ferry flag to false
   has_ferry_ = false;
+  // Set not thru pruning to true
+  not_thru_pruning_ = true;
```
dnesbitt61, line 96: aren't multi-modal routes always forward search (not bidirectional) - and we don't use not_thru pruning except in bidirectional A*?

src/thor/timedep_forward.cc
```diff
@@ -43,6 +43,8 @@ void TimeDepForward::Clear() {
 
   // Set the ferry flag to false
   has_ferry_ = false;
+  // Set not thru pruning to true
+  not_thru_pruning_ = true;
```
dnesbitt61, line 47: same comment as multi-modal

valhalla/thor/pathalgorithm.h
```diff
@@ -89,6 +90,23 @@ class PathAlgorithm {
     return has_ferry_;
   }
 
+  /**
+   * Set the not_thru_pruning_
+   * @param pruning  set the not_thru_pruning_ to pruning value.
+   *                 only set on the second pass
```
dnesbitt61, line 96: this is set to true on first pass for bidirectional, so this comment line is incorrect

dnesbitt61: I would consider moving the comments on not_thru pruning from bidirectional_astar.cc into here.

valhalla/thor/pathalgorithm.h
```diff
@@ -106,6 +124,8 @@ class PathAlgorithm {
 
   bool has_ferry_; // Indicates whether the path has a ferry
 
+  bool not_thru_pruning_; // Indicates whether to allow access into a restricted no thru region.
```
dnesbitt61, line 127: take out the word restricted, change to "into a not-thru region"

src/thor/bidirectional_astar.cc
```diff
@@ -549,11 +557,17 @@ inline bool BidirectionalAStar::ExpandReverseInner(GraphReader& graphreader,
   float sortcost =
       newcost.cost + astarheuristic_reverse_.Get(t2->get_node_ll(meta.edge->endnode()), dist);
 
+  // There is an edge case where we may encounter only_restrictions with edges being
```
dnesbitt61, line 560: consider moving this lengthy comment into the PathAlgorithm method that sets not_thru pruning.

src/thor/astar_bss.cc
```diff
@@ -57,6 +57,8 @@ void AStarBSSAlgorithm::Clear() {
 
   // Set the ferry flag to false
   has_ferry_ = false;
+  // Set not thru pruning to true
+  not_thru_pruning_ = true;
```
dnesbitt61, line 61: I thought only bidirectional  A* uses not_thru_pruning?

dnesbitt61 (dismissed):

dnesbitt61 (dismissed):

dnesbitt61 (approved):

PR #2984, opened 2021-04-07
Parsenodes fix: check if index is out-of-bound first

merkispavel:
# Issue

if the code above results in `curren_way_node_index_ == npos` then  vector out-of-range error happens
```cpp
if (osmid > (*(*way_nodes_)[current_way_node_index_]).node.osmid_) {
      current_way_node_index_ =
          way_nodes_->find_first_of(OSMWayNode{{osmid}},
                                    [](const OSMWayNode& a, const OSMWayNode& b) {
                                      return a.node.osmid_ <= b.node.osmid_;
                                    },
                                    current_way_node_index_);
    }
```

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/mjolnir/pbfgraphparser.cc
+++ b/src/mjolnir/pbfgraphparser.cc
@@ -855,8 +855,8 @@ public:
     // if this nodes id is less than the waynode we are looking for then we know its a node we can
     // skip because it means there were no ways that we kept that referenced it. also we could run out
     // of waynodes to look for and in that case we are done as well
-    if (osmid < (*(*way_nodes_)[current_way_node_index_]).node.osmid_ ||
-        current_way_node_index_ >= way_nodes_->size()) {
+    if (current_way_node_index_ >= way_nodes_->size() ||
+        osmid < (*(*way_nodes_)[current_way_node_index_]).node.osmid_) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/mjolnir/pbfgraphparser.cc
```diff
@@ -855,8 +855,8 @@ struct graph_callback : public OSMPBF::Callback {
     // if this nodes id is less than the waynode we are looking for then we know its a node we can
     // skip because it means there were no ways that we kept that referenced it. also we could run out
     // of waynodes to look for and in that case we are done as well
-    if (osmid < (*(*way_nodes_)[current_way_node_index_]).node.osmid_ ||
-        current_way_node_index_ >= way_nodes_->size()) {
+    if (current_way_node_index_ >= way_nodes_->size() ||
```
danpat, line 858: Should this be `>`, not `>=` ?  If `current_way_node_index_` == `way_nodes_->size()`, then we'll access 1 past the end of the array, even with your fix.

merkispavel: if `current_way_node_index_  == way_nodes_->size()` then it immediately jump to `return;` without checking the second condition, nope?

danpat: 🤦 I had it backwards, nevermind!

src/mjolnir/pbfgraphparser.cc
```diff
@@ -855,8 +855,8 @@ struct graph_callback : public OSMPBF::Callback {
     // if this nodes id is less than the waynode we are looking for then we know its a node we can
     // skip because it means there were no ways that we kept that referenced it. also we could run out
     // of waynodes to look for and in that case we are done as well
-    if (osmid < (*(*way_nodes_)[current_way_node_index_]).node.osmid_ ||
-        current_way_node_index_ >= way_nodes_->size()) {
```
kevinkreiser, line 859: derp, good catch, sorry about that!

kevinkreiser (approved):

gknisely (dismissed):

Issue #2793, opened 2021-01-21
Avoid CI on unrelated changes


nilsnolde:
Happens quite often that we commit things that are wasting precious CI hours, but def don't deserve to be tested and some part of me starts weeping.. I'd propose to skip CI builds when changing .md and/or other unrelated files. I already had a look into it here: https://github.com/valhalla/valhalla/pull/2763#issuecomment-756804568

I'll do it if you agree.

nilsnolde:
Capture some research:
CircleCI: https://support.circleci.com/hc/en-us/articles/360015562253-Conditionally-end-a-running-job-gracefully
GH Actions: https://stackoverflow.com/questions/62968897/is-it-possible-to-not-run-github-action-for-readme-updates
Extrawurst Azure: https://stackoverflow.com/questions/62044055/is-it-possible-to-cancel-a-azure-devops-pipeline-job-programmatically :roll_eyes:

mandeepsandhu:
+1 from my side (you dont need to go through CI hoops if you've say, changed something in the docs or the README).

Do you have initial list of files dirs that don't require running CI?

kevinkreiser:
:ship:

nilsnolde:
> Do you have initial list of files dirs that don't require running CI
Changelog would be enough already;) Wildcard *.md I think to start with.. Glancing at the rest it's pretty much all used in the builds (with the exception of very few half-dead scripts), whiich is a great sign I think, very little orphans:)

mandeepsandhu:
> > Do you have initial list of files dirs that don't require running CI
> 
> Changelog would be enough already;) Wildcard *.md I think to start with.. Glancing at the rest it's pretty much all used in the builds (with the exception of very few half-dead scripts), whiich is a great sign I think, very little orphans:)

changelog & markdown files seem like a good start! :+1:

nilsnolde:
kinda there with circleci

PR #2986, opened 2021-04-08
Conditionally skip CI runs

mandeepsandhu:
Added support for gitignore like functionality for skipping CI runs.
Specify filename or patterns in .circleci/ciignore and if all staged
for comimt match these, then the commit message is prefix with a tag
to skip circle CI runs.

Currently, the ciignore rules contain all *.md files, the ciignore file
itself and run_route_scripts/*


mandeepsandhu:
The `skip ci` circleci feature has a gotcha as explained [here](https://circleci.com/docs/2.0/skip-build/)
```
This not only skips the marked commit, but also all other commits in the push.
```
This means if the last change before doing `git push` was a file from the ciignore list, the entire push's commits will not be built. A workaround for this could be to ask the user if a `skip ci` will be added to a commit (the user can opt not to add the tag).

For the most part, we're changing files that are not ignored, so the new flow will not be triggered and won't slowdown the regular workflow. However, a lot of times I'm only changing a changelog and would prefer to skip CI. In such cases I guess its okay to have an interactive prompt. What do you think?

nilsnolde:
Good start to "fix" #2793 , thanks! Def helps the main devs, so probably >80% of "unnecessary" CI runs, I doubt casual contributors will add the git hook. Eventually it'd be perfect to consolidate it all to github actions and use its convenient exclude or include patterns to handle the stuff CI-side. 

We'll also need to mark circle CI jobs as non-critical for a merge in case they're skipped client-side.

.githooks/commit-msg
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+import fnmatch
+import os.path
+import sys
+import subprocess
+
+CI_IGNORE_FILE='.circleci/ciignore'
+CI_IGNORE_TAG='[skip ci]'
+GIT_CMD=['git', 'diff', '--cached', '--name-only']
```
PureW, line 9: Perhaps this should compare to what's in master?

Wouldn't you be able to do the following otherwise:

1) Commit change to C++ code that fails tests with CI failing.
2) Commit change to readme, CI is now skipped and build becomes green?

mandeepsandhu: The idea here is to skip CI for intermediate commits when working on a PR. eg: you make code changes, push them up, then in a separate commit push just a changelog change. This will prevent running CI for just the changelog change. Also, if we make it ask the user, then they can decide if they want to skip CI or not.

If we diff with master, then we'll almost always run CI, since a PR changing just non-buildable files is quite rare, but changing them in a series of commit that make a PR might be more common.

Whats not very clear from the docs is what happens to in-progress runs when a push with `skip ci` is made. Say I do a code push, which triggers CI, and then before that run completes, I push a skippable change. Looks like the previous run still runs, but I can't see its progress in github UI. I need to do a bit more testing to see how it behaves.

mandeepsandhu:
Do not review. un-drafted it only to see if the merge button is enabled if the last CI was skipped but successful (looks like I need to update with master too )

.githooks/commit-msg
mandeepsandhu: Okay, so as per my tests, if a previous CI run is in progress and you push a commit with `[skip ci]` (or `[ci skip]`), the existing CI run continues on (and you can see its status against the commit sha) BUT, circle CI does not invoke a completion callback to github. And it seems like this prevents from the merge button from being activated (even though the previous run was successful) 

This makes using `skip ci` not very useful (it will work if we let the last CI run run fully and update status, before pushing the skip CI commit, but then thats hardly useful). :disappointed:

Let me see if there's a way to use parameters and use conditionals within CI config to skip CI runs (i.e circle jobs still run, but they dont execute certain workflows). This should keep github happy as the circle callback would still get invoked.

mandeepsandhu: Doh, the merge button is disabled because the PR is not approved :man_facepalming: Although the rest of my issues with skip-ci still remain.

gknisely (dismissed):

.githooks/commit-msg
mandeepsandhu: Looks like I was right after all, the status needs to be passing for the merge button to enable:
Screenshot from 2021-04-09 17-39-17

Back to square 1 :/

mandeepsandhu:
Okay, so this latest way of skipping CI seems to be working more reliably than the previous approach of using skip tag in commit.

There are 2 gotchas noted in the `TODO`s in `scripts/needs_ci_run`. I think both are doable, but will require making API calls to github (for fetching all commits in the last psuh) and circleci (to get build status of last run of the PR branch).

If we are okay with that ti begin with, we should consider merging this change.

CC @nilsnolde @PureW @kevinkreiser @danpat

PureW (approved):
LGTM

.circleci/config.yml
```diff
@@ -160,10 +170,15 @@ jobs:
...
-      - install_macos_dependencies
       - checkout
+      - run: |
+            if ! ./scripts/needs_ci_run; then
+                echo "Changes in last commit do not need CI. Skipping step"
+                circleci-agent step halt
+            fi
       - run: ./scripts/format.sh && ./scripts/error_on_dirty.sh
       - run: git submodule sync && git submodule update --init
+      - install_macos_dependencies
```
kevinkreiser, line 181: so all the dependencies needed for `needs_ci_run`, `format.sh` and `error_on_dirty.sh` must already been installed by default? i guess its just python, since format.sh gets its stuff from mason? i thought error on dirty used `parallel` but i havent checked!

mandeepsandhu: `needs_ci_run` uses vanilla python modules, and it seems that the OSX executor comes installed with it. `error_on_dirty.sh` is not using `parallel`. `format.sh` does have `mason` as a dependency, but it also installs it if not already installed. This change is strictly not need (the install step does add 2-3 mins which is not too bad I guess?). Initially I was checking to see if the skip check would work if I moved `install_dependencies` to a later stage. It did, so I left the change in the PR.

kevinkreiser: yep its all good, i was just refreshing my memory! wasnt asking for a change or anything like that. if we can save some time lets do it

kevinkreiser (approved):

nilsnolde:
Should actually run on azure too, thanks for using py, not bash! I'll try...

Issue #2973, opened 2021-03-29
Isochrone: isochrone visits edges incorrectly


merkispavel:
1. Post visiting strategy. Isochrone [overrides](https://github.com/valhalla/valhalla/blob/master/src/thor/isochrone.cc#L277) `ExpandingNode` method so once the node is popped from the queue we look at the edge `(prev_node, node)` and do isochrone specific logic. 
2. Stop expansion. The prev.secs() is [used](https://github.com/valhalla/valhalla/blob/master/src/thor/isochrone.cc#L301) to check whether this edge is interested to us.

(2) looks ok. However when it's combined with (1) - they do evil things.

### Example
Say we have contour 5 min.

Queue elements:
1. prev = 4 min, current = 8 min
2. prev = 6 min, current = 7 min

(2) node(i.e current node) has less ETA(~ less cost) so it is popped first. It triggers stop expansion because `prev.secs = 6min > contour = 5 min`. **BUT** there is another (4min, 8min) edge which is interested to us because there's a segment (4min, 5min) that lie within the contour

### Possible solutions
Change visiting strategy post -> pre. So we can visit the edge at the moment we add it to the queue.

### Impact
Underestimated isochrone region. There might be(the change is high) reachable regions that isochrone response don't include

kevinkreiser:
@merkispavel im pretty sure this is a known issue. @dnesbitt61 wrote the original implementation and there we added (and still do add) an extra bit of expansion (a buffer) to stop us from quitting too early. Your suggestion makes sense to me. We can add another pure virtual method (`EnquingEdge` or something) that is called when a edge is added to the queue and switch isochrones to use that method when updating the grid. This will cost more in one sense because we arent "drawing" only the popped edges but all pushed ones, but we'll be able to remove that buffer and quit when the queue is exhausted rather than when we hit a threshhold (because we'll prun paths that are outside of the contour)

merkispavel:
I spend some time thinking about it ^. We probably don't need another visitor method. We can prune edge based on its start node time/distance https://github.com/valhalla/valhalla/pull/2990

PR #2990, opened 2021-04-09
Isochrone: handle origin edges properly

merkispavel:
# Issue

The main cause of the issue is a long origin edge. See the diff for more details
image

- Added a fix to handle origin edges properly(see `Isochrone::UpdateIsoTile` method)
- Added a fix to handle long edges properly(see `Isochrone::ShouldExpand` method)). fixes https://github.com/valhalla/valhalla/issues/2973

Inputs: [planet.tar.zip](https://github.com/valhalla/valhalla/files/6285763/planet.tar.zip)
Request: `curl http://localhost:8002/isochrone --data '{"locations":[{"lat":39.803886,"lon":116.723650}],"costing":"pedestrian","contours":[{"time":10}]}'`

| Before  | After |
| ------------- | ---|
| b2  | a2 |

| Before  | After |
| ------------- | ---|
| b3 | a3 |

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/thor/isochrone.cc
```diff
@@ -200,76 +245,59 @@ void Isochrone::UpdateIsoTile(const EdgeLabel& pred,
   }
 
   // Get the time and distance at the end node of the predecessor
-  // TODO - do we need partial shape from origin location to end of edge?
```
merkispavel, line 203: It's your the hour of glory, 4-years-old-TODO

src/thor/isochrone.cc
```diff
@@ -298,10 +326,13 @@ ExpansionRecommendation Isochrone::ShouldExpand(baldr::GraphReader& /*graphreade
...
-      pred.path_distance() > max_meters_) {
-    return ExpansionRecommendation::stop_expansion;
-  }
+  float time =
+      pred.predecessor() == kInvalidLabel ? 0.f : bdedgelabels_[pred.predecessor()].cost().secs;
+  float distance =
+      pred.predecessor() == kInvalidLabel ? 0.f : bdedgelabels_[pred.predecessor()].path_distance();
+  // prune the edge if its start is above max contour
+  if (time > max_seconds_ && distance > max_meters_)
+    return ExpansionRecommendation::prune_expansion;
```
merkispavel, line 335: I wasn't sure if this PR is good place for this commit or I should create standalone one. However, both fixes are need for the illustrated issue(i.e long-origin-edge-issue)

mandeepsandhu:
@merkispavel what about long edges causes the issue? from some the **before** pics I can see that the the isochrone does not expand to certain road segments. Whats the root-cause for that?

merkispavel:
@mandeepsandhu My bad, I haven't said anything about that.
There's an example in https://github.com/valhalla/valhalla/issues/2973. Let me know if it makes long-edge-issue clearer

src/thor/isochrone.cc
merkispavel: https://github.com/valhalla/valhalla/issues/2973

src/thor/isochrone.cc
```diff
@@ -200,76 +245,59 @@ void Isochrone::UpdateIsoTile(const EdgeLabel& pred,
...
-      for (const auto& t : tiles) {
-        isotile_->SetIfLessThan(t.first, {secs1 * kMinPerSec, dist1 * kKmPerMeter});
-      }
+    if (pred.origin()) {
+      // interpolate ll0 for origin edge using edge_label.path_distance()
+      auto edge_info = tile->edgeinfo(edge);
+      const auto& shape = edge_info.shape();
+      const auto& ordered_shape =
+          edge->forward() ? shape : std::vector<midgard::PointLL>(shape.rbegin(), shape.rend());
+      auto origin_edge_shape = OriginEdgeShape(ordered_shape, pred.path_distance());
```
kevinkreiser, line 262: a reference to a temporary, does that work? maybe to be safe:

```suggestion
      if(!edge->forward):
        std::reverse(shape.begin(), shape.end());
      auto origin_edge_shape = OriginEdgeShape(shape, pred.path_distance());
```

src/thor/isochrone.cc
```diff
@@ -200,76 +245,59 @@ void Isochrone::UpdateIsoTile(const EdgeLabel& pred,
...
-  auto resampled = resample_polyline(shape, edge->length(), shape_interval_);
-  if (!edge->forward()) {
-    std::reverse(resampled.begin(), resampled.end());
+  std::vector<PointLL> resampled;
+  auto edge_info = tile->edgeinfo(edge);
+  const auto& shape = edge_info.shape();
+  if (pred.origin()) {
+    const auto& ordered_shape =
+        edge->forward() ? shape : std::vector<midgard::PointLL>(shape.rbegin(), shape.rend());
+    const auto origin_edge_shape = OriginEdgeShape(ordered_shape, pred.path_distance());
```
kevinkreiser, line 279: same comment as above

src/thor/isochrone.cc
```diff
@@ -200,76 +245,59 @@ void Isochrone::UpdateIsoTile(const EdgeLabel& pred,
...
   // the shape interval to get regular spacing. Use the faster resample method.
   // This does not use spherical interpolation - so it is not as accurate but
   // interpolation is over short distances so accuracy should be fine.
-  auto shape = tile->edgeinfo(edge).shape();
-  auto resampled = resample_polyline(shape, edge->length(), shape_interval_);
-  if (!edge->forward()) {
-    std::reverse(resampled.begin(), resampled.end());
+  std::vector<PointLL> resampled;
+  auto edge_info = tile->edgeinfo(edge);
+  const auto& shape = edge_info.shape();
```
kevinkreiser, line 275: i see you are taking a const reference here so you cant modify it below. in this case its fine to resample it first and then reverse it and then call just `OriginEdgeShape` right before you use it in the for loop.

but up above where the length is short maybe you can just use the snap point from the origin candidate somehow instead of accessing the shape at all

src/thor/isochrone.cc
merkispavel: this is the trick to avoid shape copying if reverse is not needed https://abseil.io/tips/107

src/thor/isochrone.cc
merkispavel: > but up above where the length is short maybe you can just use the snap point from the origin candidate somehow instead of accessing the shape at all

I thought about it. But I haven't found where you can get snap point from `EdgeLabel` struct. Wondering if assume that start and end point is "enough" for short edges and we can just interpolate snap point based on start--end edge rather than fetch whole shape

merkispavel: > i see you are taking a const reference here so you cant modify it below. in this case its fine to resample it first and then reverse it and then call just OriginEdgeShape right before you use it in the for loop.

I haven't thought about that! Thanks, applied

src/thor/isochrone.cc
kevinkreiser: Ah I see constifying it extended the lifetime. That makes sense I use this trick a lot for default parameter values from temporaries

src/thor/isochrone.cc
mandeepsandhu: Ah yes, that does explain the issue! Thanks :+1:

test/isochrone.cc
```diff
@@ -100,23 +100,23 @@ TEST(Isochrones, Basic) {
...
     const auto request =
         R"({"locations":[{"lat":52.078937,"lon":5.115321}],"costing":"auto","contours":[{"time":9}],"polygons":false,"denoise":0.2,"generalize":100})";
```
mandeepsandhu, line 105: I see we've updated the tests. Are these tests exercising the "long edge" issue? If yes, then :+1: , if no, then maybe add a test (gurka or otherwise)?

merkispavel: Gurka testing it is a good idea. I'll do it 👍

merkispavel: @mandeepsandhu gurka tests are added

merkispavel:
PR is ready for the second round of review. Feel free to merge it if it looks good

kevinkreiser (approved):
looks good to me!

PR #2991, opened 2021-04-09
Increase coverage jobs to 4

mandeepsandhu:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


PureW (approved):

merkispavel:
from @mandeepsandhu 
"This shaves off about 5 mins from the build time. unfortunately other targets cannot be bumped up as it causes the build machine to OOM-kill the compilation jobs (due to smaller sized instances), but coverage seems to run fine with higher count"

PR #2992, opened 2021-04-10
Avoid UB on converting NaN to int in annotations

averkhaturau:
# Issue

Speed annotation may fail with converting NaN to integer if the same geometry point in duplicated in a route. 

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/triplegbuilder.cc
+++ b/src/thor/triplegbuilder.cc
@@ -328,6 +328,9 @@ void SetShapeAttributes(const AttributesController& controller,
     }
     distance_total_pct = next_total;
     double time = distance / cut_itr->speed; // seconds
+    if (std::isnan(time)) {
+      time = 0.;
+    }
 
     // Set shape attributes time per shape point if requested
     if (controller.attributes.at(kShapeAttributesTime)) {
@@ -344,7 +347,11 @@ void SetShapeAttributes(const AttributesController& controller,
     // Set shape attributes speed per shape point if requested
     if (controller.attributes.at(kShapeAttributesSpeed)) {
       // convert speed to decimeters per sec and then round to an integer
-      leg.mutable_shape_attributes()->add_speed((distance * kDecimeterPerMeter / time) + 0.5);
+      double speed = (distance * kDecimeterPerMeter / time) + 0.5;
+      if (std::isnan(speed) || time == 0.) { // avoid NaN
+        speed = 0.;
+      }
+      leg.mutable_shape_attributes()->add_speed(speed);
     }
 
     // Set the maxspeed if requested
--- a/test/shape_attributes.cc
+++ b/test/shape_attributes.cc
@@ -80,6 +80,66 @@ TEST(ShapeAttributes, test_shape_attributes_included) {
   }
 }
 
+TEST(ShapeAttributes, test_shape_attributes_duplicated_point) {
+  tyr::actor_t actor(conf);
+
+  auto result_json = actor.trace_attributes(
+      R"({"shape":[
+        {"lat":52.09110,"lon":5.09806},
+        {"lat":52.09110,"lon":5.09806},
+        {"lat":52.09050,"lon":5.09769},
+        {"lat":52.09098,"lon":5.09679}
+      ],"costing":"auto","shape_match":"map_snap",
+      "filters":{"attributes":["edge.length","edge.speed","edge.begin_shape_index",
+      "edge.end_shape_index","shape","shape_attributes.length","shape_attributes.time","shape_attributes.speed"],
+      "action":"include"}})");
+
+  rapidjson::Document doc;
+  doc.Parse(result_json);
+  EXPECT_FALSE(doc.HasParseError()) << "Could not parse json response";
+
+  auto shape =
+      midgard::decode<std::vector<PointLL>>(rapidjson::Pointer("/shape").Get(doc)->GetString());
+  auto shape_attributes_time = rapidjson::Pointer("/shape_attributes/time").Get(doc)->GetArray();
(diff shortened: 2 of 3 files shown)


src/thor/triplegbuilder.cc
```diff
@@ -328,6 +328,9 @@ void SetShapeAttributes(const AttributesController& controller,
     }
     distance_total_pct = next_total;
     double time = distance / cut_itr->speed; // seconds
+    if (time != time) {                      // avoid NaN
```
SiarheiFedartsou, line 331: probably [`std::isnan`](https://en.cppreference.com/w/cpp/numeric/math/isnan) is more readable

SiarheiFedartsou (dismissed):

SiarheiFedartsou (dismissed):

kevinkreiser:
Why does the duplicate occur? I thought the cutting logic was robust to this which means it has to be duplicated shape in OSM itself? If so that is an easy gruka test to write to actually get test coverage here. Shouldn't we add that quickly before merging?

danpat (changes requested):
Can someone add a test for this case?  @kevinkreiser yeah, this is occuring when the underlying OSM data contains a zero-length, duplicate-coordinate node.

test/shape_attributes.cc
```diff
@@ -80,6 +80,66 @@ TEST(ShapeAttributes, test_shape_attributes_included) {
   }
 }
 
+TEST(ShapeAttributes, test_shape_attributes_duplicated_point) {
+  tyr::actor_t actor(conf);
+
+  auto result_json = actor.trace_attributes(
```
PureW, line 86: This test passes on master as well...

kevinkreiser:
i wonder if this line is a contributing factor:

`double speed = (edge->length() * (tgt_pct - src_pct)) / edge_seconds;`

maybe speed is 0 if at the beginning or end of a map match we use 0% of an edge or maybe edge_seconds is 0. @averkhaturau i assume you ran into this sometime. did you happen to debug it to see the root cause of the nan?

test/shape_attributes.cc
danpat: @PureW the test will pass, but is `ubsan` triggered?

averkhaturau: I'm not sure how to test UB without ubsan, I'll enable it on CI

averkhaturau:
> did you happen to debug it to see the root cause of the nan?

In short, if we have a duplicated coordinate in route geometry (we usually do in multileg routes), the speed is calculated as
`speed = length / time`
Both `length` and `time` are 0., so speed is set to `NaN`. Then it converts to integer in call of `set_speed(int)`, which is undefined behavior. Happily (or actually unhappily) clang converts NaN to integer 0 and nothing awful happens. ubsan catches it.

So, the reason is actually a duplicated coordinate.

kevinkreiser:
I do not understand your comment about multileg routes! TripLegBuilder only works on one leg at a time, so surely multileg routes cannot cause this problem. I'm just wondering where the point duplication comes in. Is it in the base OSM data like that or do we have a bug elsewhere that is erroneously repeating a coordinate (ie when cutting the shape for the first or last edge of the route, also done in triplegbuilder). its good to handle it because OSM could have duplicate adjacent points (and im certain we dont weed that out at tile building time), but my main question is what is causing duplicated points and do we need to fix something more in a separate PR

test/shape_attributes.cc
kevinkreiser: i would be fine to just undo the fix in triplegbuilder, turn on ubsan locally, show that the sanitizer finds the issue via the unit test and post it to this PR. we shouldnt change the behavior of CI in this PR, iirc when we had that enabled we either got randomly OOM killed or the build took forever

merkispavel:
> I do not understand your comment about multileg routes! TripLegBuilder only works on one leg at a time, so surely multileg routes cannot cause this problem. I'm just wondering where the point duplication comes in. Is it in the base OSM data like that or do we have a bug elsewhere that is erroneously repeating a coordinate (ie when cutting the shape for the first or last edge of the route, also done in triplegbuilder)? its good to handle it because OSM could have duplicate adjacent points (and im certain we dont weed that out at tile building time), but my main question is what is causing duplicated points and do we need to fix something more in a separate PR

@kevinkreiser as far as i see we do have request that reproduces the issue. I'll come back to you soon and let you know if we need one more fix somewhere else. But I believe this activity is not a blocker for this PR

test/shape_attributes.cc
averkhaturau: In my opinion, CI should test valhalla for UB, it would help to avoid UB issues in future, It's not the only issue found with ubsan.

averkhaturau:
This PR is a hot workaround of UB. Duplicated point is a separate issue.

kevinkreiser:
yeah im not trying to block the PR. let me reiterate my points more concisely:

> its good to handle it because OSM could have duplicate adjacent points
^^ we should ship this to protect us from NaN regardless of the cause

> what is causing duplicated points and do we need to fix something more in a separate PR
^^ lets investigate the cause of the duplicate adjacent shape points in a separate issue/PR

> we shouldnt change the behavior of CI in this PR
^^ revert the changes to CI

averkhaturau:
@kevinkreiser , I reverted CI changes. Should I revert a test as well?

merkispavel (approved):
Looks good

danpat (approved):

danpat:
Duplicated coordinates can _easily_ exist in the original OSM data - I don't think there's any need to look for another cause here.  A simple scan of OSM I'm sure will expose thousands, if not millions, of cases of this.  Valhalla needs to be robust against that, and as @averkhaturau pointed out, we've been lucky that it just so happens that NaN->0 for our serialization in most cases.

PR #2995, opened 2021-04-12
nit: Enables compiler warnings in part of mjolnir module 

merkispavel:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/mjolnir/bssbuilder.cc
```diff
@@ -580,7 +579,7 @@ void BssBuilder::Build(const boost::property_tree::ptree& pt, const std::string&
   auto local_level = TileHierarchy::levels().back().level;
 
   // Group the nodes by their tiles. In the next step, we will work on each tile only once
-  for (const auto& node : osm_nodes) {
```
merkispavel, line 583: sequence::iterator dereference returns `T` instead of `const T&`

src/mjolnir/graphvalidator.cc
```diff
@@ -630,8 +630,7 @@ void GraphValidator::Validate(const boost::property_tree::ptree& pt) {
       }
       sum += density;
     }
-    float average_density = sum / densities[level].size();
```
merkispavel, line 633: its usage is under LOG_DEBUG macro so I just moved whole expression

src/mjolnir/osmway.cc
```diff
@@ -140,7 +140,7 @@ OSMWay::GetNames(const std::string& ref, const UniqueNames& name_offset_map, uin
       tokens = GetTagTokens(name_offset_map.name(ref_index_));
     }
 
-    for (const auto& t : tokens) {
```
merkispavel, line 143: unused `t`

src/mjolnir/pbfadminparser.cc
```diff
@@ -36,9 +31,9 @@ struct admin_callback : public OSMPBF::Callback {
   }
   // Construct PBFAdminParser based on properties file and input PBF extract
   admin_callback(const boost::property_tree::ptree& pt, OSMAdminData& osmdata)
-      : shape_(pt.get<unsigned long>("id_table_size", kMaxOSMNodesHint)),
-        members_(pt.get<unsigned long>("id_table_size", kMaxOSMNodesHint)), osm_admin_data_(osmdata),
-        lua_(std::string(lua_admin_lua, lua_admin_lua + lua_admin_lua_len)) {
+      : lua_(std::string(lua_admin_lua, lua_admin_lua + lua_admin_lua_len)),
+        shape_(pt.get<unsigned long>("id_table_size", kMaxOSMNodesHint)),
+        members_(pt.get<unsigned long>("id_table_size", kMaxOSMNodesHint)), osm_admin_data_(osmdata) {
```
merkispavel, line 36: initialization order warning fix

src/mjolnir/graphvalidator.cc
merkispavel: if somebody is in love with variable we can do it the other way https://github.com/valhalla/valhalla/pull/2922#discussion_r591666498

src/mjolnir/CMakeLists.txt
```diff
@@ -25,12 +25,34 @@ if (UNIX)
...
+    osmway.cc
+    pbfadminparser.cc
+    restrictionbuilder.cc
+    servicedays.cc
+    shortcutbuilder.cc
+    timeparsing.cc
+    transitbuilder.cc
+    util.cc
+    validatetransit.cc
+    PROPERTIES COMPILE_FLAGS "-Werror -Wall -Werror=sign-compare")
```
merkispavel, line 55: I believe clang doesn't include `sign-compare` warning to the `-Wall` I had to add it manually to reproduce errors in CI.
I see 2 options here:
- `-Werror=sign-compare`(in the PR at the moment)
- `-Wno-error=sign-compare`, i.e don't treat this warning as error

kevinkreiser: CI isnt building with clang except for on osx though (unless that changed...?)

kevinkreiser (dismissed):

PureW (approved):
Nice, more code under compiler checks!

src/mjolnir/CMakeLists.txt
```diff
@@ -25,12 +25,33 @@ if (UNIX)
...
-    PROPERTIES COMPILE_FLAGS "-Werror -Wall")
+    osmway.cc
+    pbfadminparser.cc
+    restrictionbuilder.cc
+    shortcutbuilder.cc
+    timeparsing.cc
+    transitbuilder.cc
+    util.cc
+    validatetransit.cc
+    PROPERTIES COMPILE_FLAGS "-Werror -Wall -Wextra -Wno-error=missing-field-initializers")
```
merkispavel, line 54: 1. We have many lines like `OSMWay way{id}` that doesn't initialize all fields but the only one(id). That's why -Wno-error=missing-field-initializers is needed
2. gcc and clang have different set of -Wall so -Wextra is needed. E.g. clang's -Wextra contains `-Wsign-compare, -Wunused-parameter` cc @PureW

valhalla/mjolnir/osmpbfparser.h
```diff
@@ -34,10 +34,13 @@
 #include <fstream>
 #include <string>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
 // this describes the low-level blob storage
 #include <valhalla/proto/fileformat.pb.h>
 // this describes the high-level OSM objects
 #include <valhalla/proto/osmformat.pb.h>
+#pragma GCC diagnostic pop
```
merkispavel, line 43: So my local protobuf has plenty of unused parameters. I see at least to ways here
1. Disable specific warnings for thirparty/external headers. This option is implemented in the PR
2. Disable specific warnings across all project. e.g. https://github.com/mapbox/mapbox-gl-native/blob/b8edc2399b9640498ccbbbb5b8f058c63d070933/CMakeLists.txt#L61

merkispavel:
@kevinkreiser @PureW there were some broken CI initially so I had to add more changes to this PR
QA
1. decide what to do with warnings in external headers https://github.com/valhalla/valhalla/pull/2995#discussion_r613492306
2. FYI `-Wextra` flag is added https://github.com/valhalla/valhalla/pull/2995#discussion_r613489086

valhalla/baldr/datetime.h
```diff
@@ -19,7 +19,10 @@
 #include <valhalla/baldr/nodeinfo.h>
 #include <valhalla/midgard/constants.h>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
 #include <valhalla/proto/tripcommon.pb.h>
+#pragma GCC diagnostic pop
```
kevinkreiser, line 25: this is so unmanageable though. we'll have to litter this all over the code. maybe we could make a macro out of it or something:

`INCLUDE_PROTO valhalla/proto/tripcommon.pb.h`

OR maybe we should make a separate cmake target with all the generated protobuffer stuff in it and include just one header everywhere?

OR we could just disable this warning globally?

none of the options are particularly great

merkispavel: > this is so unmanageable though
agreed

I don't really like macro for a few reasons. I was thinking about upgrading cmake step. As I see all proto-headers live in `${BUILD_DIR}/src/valhalla/` directory. Maybe we could add
```
for header in ${BUILD_DIR}/src/valhalla/*.h:
  insert "#pragma GCC diagnostic push\n#pragma GCC diagnostic ignored "-Wunused-parameter" to the beginning of the header
  insert `#pragma GCC diagnostic pop` to the end of the header
```

I'll take a few days to think about best approach

merkispavel: `SYSTEM` option in https://cmake.org/cmake/help/latest/command/target_include_directories.html looks like the light at the end of the tunnel
`
If SYSTEM is specified, the compiler will be told the directories are meant as system include directories on some platforms (signalling this setting might achieve effects such as the compiler skipping warnings...
`

PureW: I looked at this too for a similar PR but couldn't quite get it to work when I tried. I'll follow this PR to see if you find a way.

kevinkreiser:
Yes treating "thirdparty" headers differently from our own is an intuitive solution to this problem, @merkispavel i hope we can get this to work!

src/mjolnir/CMakeLists.txt
```diff
@@ -72,6 +59,14 @@ set(sources
   util.cc
   validatetransit.cc)
 
+if (UNIX)
+  # Enables stricter compiler checks on a file-by-file basis
+  # which allows us to migrate piecemeal
+  set_source_files_properties(
+          ${sources}
+          PROPERTIES COMPILE_FLAGS "-Werror -Wall -Wextra -Wno-missing-field-initializers -Wno-unused-parameter")
```
PureW, line 67: Were there too many warnings about unused parameters? Since you enable `Wno-unused-parameter`?

valhalla/baldr/datetime.h
merkispavel: I did a few unsuccessful tries to disable warnings in external headers using `SYSTEM`  cmake option. Results:
The best I got is disabling warnings in all headers(external + internal/valhalla headers): I inserted `SYSTEM` right before `MODULE_INCLUDE_DIRECTORIES` https://github.com/valhalla/valhalla/blob/master/src/CMakeLists.txt#L100. However, we are in cpp world and some part of our codebase is templated(i.e lives in headers). So don't think this is an option for us
In a toy example the `SYSTEM` option works. So we could create a wrapper for each external module and disable warnings for its headers, e.g
```cmake
add_library(protobuf_wrapper INTERFACE)
target_link_libraries(protobuf_wrapper protobuf)
target_include_directories(protobuf SYSTEM INTERFACE $<TARGET_PROPERTY:protobuf,INTERFACE_INCLUDE_DIRECTORIES>)
....

target_link_libraries(valhalla-mjolnir 
  protobuf_wrapper
  valhalla-midgard
...)
// each library defines in its own cmakelists whether it's headers are system or not
```

However, in order to get this we need to refactor valhalla cmake configuration, i.e operate with cmake targets rather than use target properties like https://github.com/valhalla/valhalla/blob/master/src/CMakeLists.txt#L133. From my experiments when we extract `INTERFACE_INCLUDE_DIRECTORIES` property we loose `SYSTEM` property(if it was set)
```cmake 
add_library(lib_a INTERFACE)
# system is set
target_include_directories(lib_a SYSTEM INTERFACE ./headers/) 

add_library(lib_b INTERFACE)
target_link_libraries(lib_b INTERFACE lib_a)
// doesn't work for some reason. system is lost?
target_include_directories(lib_b INTERFACE $<TARGET_PROPERTY:lib_a,INTERFACE_INCLUDE_DIRECTORIES>)
```

merkispavel: > this is so unmanageable though. we'll have to litter this all over the code. maybe we could make a macro out of it or something:
> 
> `INCLUDE_PROTO valhalla/proto/tripcommon.pb.h`
> 
> OR maybe we should make a separate cmake target with all the generated protobuffer stuff in it and include just one header everywhere?
> 
> OR we could just disable this warning globally?
> 
> none of the options are particularly great

Finally, my choice is disabling unused-parameter for all sources. As it is the only warning that external headers have - its disabling solves the issue. I hope it's a short/mid term solution until cmake is refactored and we can apply `SYSTEM` cmake option

src/mjolnir/CMakeLists.txt
merkispavel: @PureW I commented in the other thread to make the history smooth, check this out pls https://github.com/valhalla/valhalla/pull/2995#discussion_r637993339

merkispavel:
@kevinkreiser @PureW PR is ready for review. The details are in the thread https://github.com/valhalla/valhalla/pull/2995#discussion_r637993339..
tldr:
- enable werror for all mjolnir sources
- disable `unused-parameter` warning for all source files to handle warnings in external headers as a short/mid term solution

PureW (approved):
Thanks for this, LGTM!

src/mjolnir/ferry_connections.cc
```diff
@@ -259,6 +261,7 @@ bool ShortFerry(const uint32_t node_index,
     }
   }
   if (short_edge) {
+    UNUSED(wayid);
     LOG_DEBUG("Skip short ferry: way_id = " + std::to_string(wayid));
   }
```
kevinkreiser, line 266:
```suggestion
#ifdef LOGGING_LEVEL_DEBUG
  if (short_edge) {
    LOG_DEBUG("Skip short ferry: way_id = " + std::to_string(wayid));
  }
#endif
```

merkispavel: there's a chain of the variables `wayid(local variable) -> way(local variable) -> ways(method parameter)`. Every one of them is used for that debug log.  So we have to mark at least one of them as `UNUSED` to calm the warning down

kevinkreiser: other than this one suggestion, this pr look good to me!

kevinkreiser: hmm i assume the warning is "assinged to but never read"? to hell with it then lets just remove the log?

merkispavel: sure! done

kevinkreiser (approved):

src/mjolnir/ferry_connections.cc
```diff
@@ -6,6 +6,8 @@
 #include "baldr/graphconstants.h"
 #include "midgard/util.h"
 
+#define UNUSED(x) (void)(x)
```
merkispavel, line 9: damn, I forgot to removed unused "unused" macro.. I'll clean it up in the next PR

PR #2996, opened 2021-04-13
Fix for non-unique route summary

ktatterso:
# Issue

Found an interesting case where two routes had the same summary, exposing a defect in the summary logic.  The test case covers the exact reason for the issue and proves the resolution, but here's the gist of the issue:

Two routes end up getting to the same destination via different paths.  The longest named segment in each path is different.  The flaw in the logic was we stopped "unique-fying" the route summary once we saw the longest named segment was different.  Later, since we desire route summaries to be comprised of at least two named segments (if possible) we append one more named segment to each - and as a result (in this case), the route summaries are no longer unique.

The fix is to set the desired minimum number of named segments per summary to two in the "unique-fying" step.

## Tasklist

 - [x] Add tests


test/gurka/test_route_summary.cc
```diff
@@ -18,9 +18,10 @@ using namespace valhalla::baldr;
 using namespace valhalla::gurka;
 using namespace valhalla::mjolnir;
 
-namespace {
+TEST(TestRouteSummary, GetSummary) {
```
ktatterso, line 21: Sorry this diff looks weird, I basically moved the map building step into the test.

test/gurka/test_route_summary.cc
```diff
@@ -79,15 +69,15 @@ TEST(TestRouteSummary, GetSummary) {
   EXPECT_EQ(result1.trip().routes_size(), 1);
   EXPECT_EQ(result1.trip().routes(0).legs_size(), 1);
   gurka::assert::osrm::expect_steps(result1, {"RT 1", "RT 4", "RT 5"});
-  gurka::assert::osrm::expect_summaries(result0, {"RT 1, RT 5"});
+  gurka::assert::osrm::expect_summaries(result1, {"RT 1, RT 5"});
```
ktatterso, line 72: There were a couple copy/paste errors in this test.

test/gurka/test_route_summary.cc
```diff
@@ -131,3 +121,98 @@ TEST(TestRouteSummary, GetSummary) {
...
+  std::unordered_map<std::string, std::string> config_map =
+      {{"mjolnir.data_processing.use_direction_on_ways", "true"},
+       {"mjolnir.admin", VALHALLA_SOURCE_DIR "test/data/netherlands_admin.sqlite"}};
+
+  std::string workdir = "test/data/gurka_test_dup_summary_fix";
+  valhalla::gurka::map map = gurka::buildtiles(node_layout, ways, {}, {}, workdir, config_map);
+
+  map.nodes = node_layout;
+
+  if (!filesystem::exists(workdir)) {
```
PureW, line 158: Unrelated to your change: Just curious, could/should this be handled by the gurka framework? I don't immediately see anything specific to your test that requires this which makes me wonder if all gurka tests need this?

kevinkreiser: gurka already does do this. i believe in the admin tests where we need to do stuff outside of gurka we needed to manually do stuf like this but calling `buildtiles` certainly does this as mjolnir will create the dir where it needs to put tiles etc

test/gurka/test_route_summary.cc
```diff
@@ -131,3 +121,98 @@ TEST(TestRouteSummary, GetSummary) {
...
+
+  // You can see above that all two results have the same summary: RT 1, RT 2. However, each
+  // chose different paths. The auto's longest named segment was RT 1 and the ped's longest
+  // named segment was RT 2. Because the longest legs were different, the summary logic
+  // stopped searching after the first named segment and returned the minimum number of
+  // named segments (2) which resulted in them both having the same summary: RT 1, RT 2.
+  // The route's can be made unique by having the auto choose RT 3 and the ped choose RT 4
+  // at the end of the route.
+
+  // Below we mash together the three above results into a single result. When we
```
kevinkreiser, line 185: looking below it seems like actually you wanted to have alternates that you needed to generate unique summaries for? you can get alternates by asking for it in the request `"alternates":1`. maybe the map above fails to find an alternate? anyway i think it would simplify and make the test a bit more understandable.

originally when i read the comment above you happened to use the word `leg` which made me think of multipoint route but that isnt what you meant, you really meant just a subsection (series of edges) of the route.

or have i confused what you meant?

src/tyr/route_serializer_osrm.cc
```diff
@@ -1616,7 +1616,7 @@ summarize_route_legs(const google::protobuf::RepeatedPtrField<DirectionsRoute>&
 
         // k is the number of named segments in the summary. It keeps going
         // up by 1 until route_i's summary is different than the route_j's.
-        size_t k = 1;
+        size_t k = std::min(num_named_segs_needed, num_comparable);
```
kevinkreiser, line 1619: you got to love these where the fix is one line but the test is like 100 lines :smile:

test/gurka/test_route_summary.cc
ktatterso: Hmm, that's a good suggestion about `"alternates":1`.  However, wrestling gurka into the configuration I desire can sometimes take more time than expected.  So if its alright I'd like to leave it as it is:  not really testing whether we _can_ generate alternates and produce unique summaries - but rather, just a test of unique summary generation.

And sorry, yes, I frequently catch myself typing comments with the term `leg` to just mean "some section of a route", realizing it is an important reserved valhalla term. I'll keep working on that. 😅

test/gurka/test_route_summary.cc
ktatterso: I will fix this, thanks.

test/gurka/test_route_summary.cc
kevinkreiser: Oh i agree we arent testing whether or not we can generate alternates we are testing whether or not alternates have unique names. I just think that your test can be simplified to something like:

```c++
  valhalla::Api result0 = gurka::do_action(valhalla::Options::route, map, {"A", "J"}, "auto", {{"alternates", "1"}});
  EXPECT_EQ(result.trip().routes_size(), 2);
  EXPECT_EQ(result.trip().routes(0).legs_size(), 1);
  EXPECT_EQ(result.trip().routes(1).legs_size(), 1);
  const std::string expected_route_summary0 = "RT 1, RT 2, RT 3";
  const std::string expected_route_summary1 = "RT 1, RT 2, RT 4";
  gurka::assert::osrm::expect_summaries(result, {expected_route_summary0, expected_route_summary1});
```

I find that a heck of a lot easier to understand then the the splicing etc

ktatterso: I'll give this a quick shot... if it isn't quick I'd like to punt and move on with other things.  Brb.

ktatterso: I took a quick shot at modifying the test to actually compute alternates.  Not sure why it didn't, I was still only getting one route.  I was able to remove the pedestrian-only attributes, fwiw.  I guess rather than spend more time simplifying this test (which sufficiently proves the fix), I'd like to move on with other things.  thanks.

kevinkreiser: no worries, thanks for taking a shot

PureW (approved):

PR #2997, opened 2021-04-13
Update readme for macbook arch64

CuriOusQuOkka:
# Issue

Building `valhalla` on MacBook with `arm64` fails. This is because not all dependencies support the `arm64` architecture (for example, `luajit`). 
The problem can be solved by emulating the `x86_64` behavior in the `Rosetta` terminal.

## Tasklist

 - [x] Update README.md. Document instructions for setting up `Rosetta` and building `valhalla`.


kevinkreiser (approved):
i cannot confirm this at the moment but since its all marked with the `rosetta` translation layer i think its safe to merge this for now

PR #2998, opened 2021-04-13
nit(rename): Renames the encoded live speed properties

PureW:
to make it clear that the raw values are usually not what you want.
They are not measured in kph, but have on several occasions been
used in that way due to the non-obvious naming.

This rename adds `encoded_` to the speed properties to make it
clearer that the value within may not be what you actually want.

There already are getter-functions defined which should be preferred
in most use-cases

### DIFF
--- a/valhalla/baldr/traffictile.h
+++ b/valhalla/baldr/traffictile.h
@@ -51,10 +51,11 @@ constexpr uint8_t UNKNOWN_CONGESTION_VAL = 0;
 constexpr uint8_t MAX_CONGESTION_VAL = 63;
 
 struct TrafficSpeed {
-  uint64_t overall_speed : 7; // 0-255kph in 2kph resolution (access with `get_overall_speed()`)
-  uint64_t speed1 : 7;        // 0-255kph in 2kph resolution (access with `get_speed(0)`)
-  uint64_t speed2 : 7;
-  uint64_t speed3 : 7;
+  uint64_t
+      overall_encoded_speed : 7; // 0-255kph in 2kph resolution (access with `get_overall_speed()`)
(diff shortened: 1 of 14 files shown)


valhalla/baldr/traffictile.h
```diff
@@ -130,15 +132,15 @@ struct TrafficSpeed {
...
+      : overall_encoded_speed{overall_encoded_speed}, encoded_speed1{s1}, encoded_speed2{s2},
+        encoded_speed3{s3}, breakpoint1{b1}, breakpoint2{b2}, congestion1{c1}, congestion2{c2},
+        congestion3{c3}, has_incidents{incidents}, spare{0} {
   }
 
   json::MapPtr json() const volatile {
     auto live_speed = json::map({});
     if (speed_valid()) {
-      live_speed->emplace("overall_speed", static_cast<uint64_t>(get_overall_speed()));
+      live_speed->emplace("overall_encoded_speed", static_cast<uint64_t>(get_overall_speed()));
```
danpat, line 143: This one isn't encoded, it's accessed via the getter

kevinkreiser: its also a breaking change to the locate API so we probably shouldnt change it

kevinkreiser:
@PureW was the pybind update on purpose?

PureW:
> @PureW was the pybind update on purpose?

Nope! Mistake, one sec.

PureW:
@kevinkreiser @danpat any last thoughts on this?

danpat:
@PureW just needs a changelog entry :-)

danpat (approved):

PR #3002, opened 2021-04-16
ci: Vendors the codecov script

PureW:
instead of downloading and executing an unverified script that may change
or be subverted to do malicious things.

Case in point being https://about.codecov.io/security-update/


kevinkreiser (dismissed):

kevinkreiser (approved):

kevinkreiser:
one caveat is that they might change how their api works and we might fail code coverage when that happens, but its fine to bit the bullet when it come and just update the script at that point

PureW:
> one caveat is that they might change how their api works and we might fail code coverage when that happens, but its fine to bit the bullet when it come and just update the script at that point

Yes, the way I think of it is two-fold: 

1) That's the cost for depending on an external service, right. Nothing is free.
2) If `codecov` designs a service that requires frequent changes to a script that runs on clients, then I start to question how they design all their systems :)

kevinkreiser:
yep i was just mentioning it in case it hadn't sprung to your mind. basically just saying "if it breaks in the future we know why and how to fix"

PR #3003, opened 2021-04-16
Fix run_with_server.py to run on non-Linux python platforms.

danpat:
# Issue

For some reason, `run_with_server.py` barfs on macOS with Python 3.9 installed from Homebrew.  If I run it, I get:

```
$ ./run_with_server.py --test-file r.txt  --concurrency 1
Placing 100 results in 20210415_224825_r
name 'session' is not defined
name 'session' is not defined
name 'session' is not defined
name 'session' is not defined
100%
```

and no saved results.

Looking at the code, it seems that on macOS, something is different about how `global` variables are inherited by the forked processes in the `multuprocessing.Pool`.  On `master`, `initialize(parsed_args)` is only called once in the primary process, but it seems that all the workers fully inherit all the global variables.  On macOS with Python 3.9, the global variables aren't present when the worker runs, and it barfs with the error printed above.

This PR makes a few slight modifications:

  1. Switches to "proper" use of the initializer/initargs parameters to `multiprocessing.Pool` - each worker performs its own initialization call (on `master`, I _think_ it's simply a mistake in invocation doing `initializer=initialize(parsed_args)` as `initialize` returns nothing).
  2. Moves the `multiprocessing.Value` into the master process, and passes in copies to the workers.
  3. Moves directory creation into the master process so it only needs to happen once.

With these changes, `run_with_server.py` works for me on macOS.  Can someone on Linux please give this a shot and confirm it's still working there?  Given the changes I had to make, I'm really not sure I understand how this was ever working on Linux - it seems to be some pretty fundamental difference in how `multiprocessing.Pool` is implemented.


kevinkreiser (approved):
thanks for doing this, as you can imagine it was difficult for me to look at considering i dont have a local mac to play with

danpat:
@kevinkreiser you're just going to have to trust me :-)  My main concern is that it's still working on Linux - can you confirm that?  It seems to work ok for me in a docker container, but given I don't really understand _why_ we're seeing these different behaviours, I don't trust anything any more.

kevinkreiser:
back when i first learned that it didnt work on mac i googled around and indeed found that the multiprocessing module doesnt work the same on macos. yep ill take this for a spin later tonight to confirm its still all good on linux

danpat:
Ah, finally found the documentation that explains it:

https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

On Linux, the default is to `fork()`, which naturally inherits everything that's been done up to that point (in this case, including initialization of globals).

On Windows/macOS, `multiprocessing.Pool` uses a thing it calls "spawn" - it launches a fresh python subprocess and somehow passes in the code/entrypoint for it to run - thus globals/etc don't exist in that context.

More info on why macOS uses "spawn" by default instead of fork in this bug report: https://bugs.python.org/issue33725

Learn something every day.

I believe the fixes I've made should work for both platforms, the original implementation only worked when `fork()` was done.

kevinkreiser:
yep that bug is the one i read back in the day!

PR #3005, opened 2021-04-16
build: allow None build type

PureTryOut:
This is used by at least Alpine Linux and probably in more places to
make sure downstream compiler options are used

As suggested in #2965


CMakeLists.txt
```diff
@@ -79,8 +79,10 @@ elseif(CMAKE_BUILD_TYPE STREQUAL RelWithDebInfo)
   message(STATUS "Configuring in release mode with debug flags")
 elseif(CMAKE_BUILD_TYPE STREQUAL MinRelSize)
   message(STATUS "Configuring in release mode with minimized size")
+elseif(CMAKE_BUIDL_TYPE STREQUAL None)
```
PureW, line 82: Spelling?

PureW (changes requested):

CMakeLists.txt
PureTryOut: Derp :see_no_evil:

PureW (approved):

PR #3008, opened 2021-04-16
add some info to status endpoint

### LINKED ISSUE #2736: /status endpoint for general info

nilsnolde:
fixes #2736 

First go at adding some useful info to the new `/status` endpoint.

@kevinkreiser what'ya think about the general layout so far? 

Also 2 areas for advice:
- bbox: I couldn't find an existing algo to get the bounds of existing tiles, but you mentioned that here: https://github.com/valhalla/valhalla/issues/2736#issuecomment-745446814. Can you point me there pls?
- also bbox: you were mentioning ops modes where the bbox might change. Can you also point me to the piece of code where I  can see some details? Would it be ok for you to dismiss bbox for that mode of operation, if that's possible. If not possible, re-calculate?


proto/api.proto
```diff
@@ -5,11 +5,13 @@ import public "options.proto"; // the request, filled out by loki
...
 import public "directions.proto"; // the directions, filled out by odin
 import public "info.proto"; // statistics about the request, filled out by loki/thor/odin
+import public "status.proto"; // info for status endpoint
 
 message Api {
   optional Options options = 1;
   optional Trip trip = 2;
   optional Directions directions = 3;
   optional Info info = 4;
+  optional Status status = 5;
```
nilsnolde, line 15: or add to `Info`? is that one even used?

src/tyr/actor.cc
```diff
@@ -313,7 +313,7 @@ actor_t::status(const std::string& request_str, const std::function<void()>* int
   pimpl->set_interrupts(interrupt);
   // parse the request
   Api request;
-  ParseApi(request_str, Options::centroid, request);
+  ParseApi(request_str, Options::status, request);
```
nilsnolde, line 316: I guess we could skip `ParseApi` here and set the `action` directly, but also doesn't hurt too much..

valhalla/baldr/graphreader.h
```diff
@@ -452,6 +452,13 @@ class GraphReader {
    */
   virtual bool DoesTileExist(const GraphId& graphid) const;
 
+  /**
+   * Test if traffic tiles exist.   *
+   */
+  bool DoTrafficTilesExist() {
+    return !tile_extract_->traffic_tiles.empty();
+  }
```
nilsnolde, line 460: open for other ways..

src/loki/status_action.cc
```diff
@@ -1,8 +1,23 @@
 #include "loki/worker.h"
+#include "proto/status.pb.h"
 
 namespace valhalla {
 namespace loki {
-void loki_worker_t::status(Api&) const {
+void loki_worker_t::status(Api& request) const {
+  baldr::graph_tile_ptr tile = nullptr;
+  for (const auto& tile_id : reader->GetTileSet(2)) {
```
kevinkreiser, line 8: i wouldnt limit to level 2 i dont think it adds anything extra it still doing a directory scan in tile_dir mode. there is the potential for that to be a huge disk crawling operation which isnt exactly the best idea. personally i think we should cache the answer to this, which reminds me of the connectivity map. maybe it makes sense to simply return information from that? i know you can disable the connectivity map but its only useful to do that when you are constantly changing what tiles you have in cache. speaking of such senarios, there is one pitfall here that i dont think we can overcome which is if you configure your graphreader to only hit http to get tiles. also there is no need to do the call to `DoesTileExist` the `GetTileSet` call ensures it exists if it returns it.

valhalla/baldr/graphreader.h
kevinkreiser: maybe a less perscriptive name like `HasLiveTraffic`

nilsnolde:
Thanks @kevinkreiser for the preliminary review. I adjusted to using `static` where reasonable and getting the bbox from the connectivity_map, good idea (didn't even know there was a geojson func)! The caveat is it's now a string in the `status.bbox`, not a JSON object. I can live with that.. more comments inline.. 

what do you think? I'll add some small tests once it's ok from your side.

src/loki/status_action.cc
```diff
@@ -1,8 +1,40 @@
...
 namespace loki {
-void loki_worker_t::status(Api&) const {
+void loki_worker_t::status(Api& request) const {
+  // get _some_ tile
+  static baldr::graph_tile_ptr tile = nullptr;
+  if (!tile) {
+    for (const auto& tile_id : reader->GetTileSet()) {
+      tile = reader->GetGraphTile(tile_id);
+      if (tile->id().level() < baldr::TileHierarchy::GetTransitLevel().level &&
+          tile->header()->nodecount() > 0) {
```
nilsnolde, line 18: I had some (bit older) tiles flying around which had tons of empty 2.level tiles reaching all the way to the antimeridian. Not sure if that was caused by an older valhalla bug which is fixed now or by slightly wrong Geofabrik PBFs. So I added this check, can't hurt I guess

src/tyr/serializers.cc
```diff
@@ -83,10 +83,29 @@ std::vector<std::string> openlr_edges(const TripLeg& leg) {
...
+  }
+
+  auto status_msg = baldr::json::map({{"version", std::to_string(VALHALLA_VERSION_MAJOR) + "." +
+                                                      std::to_string(VALHALLA_VERSION_MINOR) + "." +
+                                                      std::to_string(VALHALLA_VERSION_PATCH)},
+                                      {"has_tiles", request.status().has_tiles()},
+                                      {"has_admins", request.status().has_admins()},
+                                      {"has_timezones", request.status().has_timezones()},
+                                      {"has_live_traffic", request.status().has_live_traffic()},
+                                      {"bbox", request.status().bbox()},
```
nilsnolde, line 103: the bbox is a stringified JSON now.. would be much nicer to have it as JSON object, but that'd require to move a little code around in `connectivity_map.cc` to support just getting the coords of the different regions, put the coords as a `Ring` to the PBF in loki's `status_action.cc` and build the GeoJSON object here. While doing that we could also consider using boost:geometry in `connectivity_map.cc` to simplify the polgons a little, they're slightly overly verbose atm.

Inclined to leave it as is for now and visit again later?

test/loki_service.cc
```diff
@@ -102,7 +102,9 @@ const std::vector<http_request_t> valhalla_requests{
 };
 
 const std::vector<std::pair<uint16_t, std::string>> valhalla_responses{
-    {200, "{}"},
+    {200,
+     R"({"has_admins":false,"actions":["'\/locate'","'\/route'","'\/height'","'\/sources_to_targets'","'\/optimized_route'","'\/isochrone'","'\/trace_route'","'\/trace_attributes'","'\/transit_available'","'\/expansion'","'\/centroid'","'\/status'"],"has_live_traffic":false,"has_timezones":false,"has_tiles":false,"bbox":"{\"features\":[],\"type\":\"FeatureCollection\"}","version":")" VALHALLA_VERSION
+     R"("})"},
```
nilsnolde, line 107: should we take those out entirely and have a separate test file for the status action?

src/tyr/serializers.cc
nilsnolde: forget it.. using rapidjson instead to serialize the response is much easier to make `bbox` a JSON object. kinda lazy way putting a (de)serialize JSON inside a PBF (de)serializer instead of using pbf objects right away. other option could be to do the above suggestion, more code changes though.

nilsnolde:
Maybe remove all params for the config JSON pointing to paths for privacy reasons?

nilsnolde:
removed all paths from response except the log paths.

src/loki/status_action.cc
```diff
@@ -1,8 +1,43 @@
...
+  static std::string bbox = "";
+  if (bbox == "" && connectivity_map) {
+    bbox = connectivity_map->to_geojson(2);
+  }
+
+  auto* status = request.mutable_status();
+  status->set_bbox(bbox);
+  status->set_has_tiles(!tile ? false : true);
+  status->set_has_admins(!tile ? false : tile->header()->admincount() > 0);
+  status->set_has_timezones(!tile ? false : tile->node(0)->timezone() > 0);
```
nilsnolde, line 34: not sure if this is really right..

src/loki/status_action.cc
kevinkreiser: they probably werent empty they probably had bins in them. its totally possible to have a long ferry that doesnt have any graph nodes along it for a long long stretch over the ocean. at the antimeridian is weird but its not buggy to find a tile with no nodes or edges in it but just bins

src/loki/status_action.cc
kevinkreiser: yep it works just like admin actually except that the list of timezones is stored in code rather than in the tiles themselves.

src/loki/status_action.cc
nilsnolde: yeah I suspect faulty osm data here

nilsnolde:
Thanks for the few comments @kevinkreiser. Think for me it's complete for now (in case your were wondering)

kevinkreiser:
yep ill take a look over it this weekend (i hope)

kevinkreiser:
i havent forgotton about this, hopefully ill have looked at it today, sorry again for the wait!

nilsnolde:
no worries @kevinkreiser , it's really not urgent.

nilsnolde:
still not really urgent @kevinkreiser , just thought I'd try again in this pretty calm period:) I'll update the changelog after CI confirms this is still good

nilsnolde:
good grief.. the submodule updates are killing me these days..

test/loki_service.cc
```diff
@@ -22,6 +23,59 @@ using namespace prime_server;
...
+    pt.erase(path);
+  }
+  boost::property_tree::json_parser::write_json(config_stream, pt);
+
+  return config_stream.str();
+}
+
+// config for permanently running server
+auto const config = make_config();
+auto const config_str = write_config(config);
```
nilsnolde, line 77: just moved that whole block up from below (and added `config_str`) since I need the config as string in the response definition for `/status`

valhalla/baldr/rapidjson_utils.h
```diff
@@ -18,7 +18,8 @@
 #define RAPIDJSON_ASSERT(x)                                                                          \
   if (!(x))                                                                                          \
   throw std::logic_error(RAPIDJSON_STRINGIFY(x))
-// Because we now throw exceptions, we need to turn of RAPIDJSON_NOEXCEPT
+// Because we now throw exceptions, we need to turn off RAPIDJSON_NOEXCEPT
+#undef RAPIDJSON_HAS_CXX11_NOEXCEPT
```
nilsnolde, line 22: can't remember, but think this was to get rid of a warning

test/actor.cc
```diff
@@ -49,8 +49,12 @@ TEST(Actor, Basic) {
   auto transit = test::json_to_pt(transit_json);
   ASSERT_NE(transit_json.find(std::to_string(false)), std::string::npos);
 
+  actor.status("");
+  actor.cleanup();
   auto status_json = actor.status("");
-  ASSERT_EQ(status_json, "{}");
+  actor.cleanup();
+  auto status = test::json_to_pt(status_json);
+  ASSERT_NE(status_json.find("Polygon"), std::string::npos);
```
nilsnolde, line 57: just doing the same as the other actions

nilsnolde (commented):
accidentally reviewed my own pr..

nilsnolde:
ok, now though I hope..

I'm a bit confused about the submodule updates (also lacking experience, I rarely have to do this):

this PR was old so I had to pull in the updates from master incl the bench and robin-hood updates, so I did:
```
git merge upstream/master
git submodule update --init --recursive
```

`git status` was clear at that point, but when I pushed Github told me those libs would be updated to their previous commits, so really they weren't updated in this branch at all. Was I doing smth wrong or was the commit that updated those libs doing smth wrong?

test/loki_service.cc
```diff
@@ -379,46 +439,6 @@ std::list<std::pair<std::string, std::string>>
          */
     };
 
-boost::property_tree::ptree make_config(const std::vector<std::string>& whitelist = {
```
nilsnolde, line 382: moved to the beginning of the namespace

src/loki/status_action.cc
kevinkreiser: This was #3202

valhalla/baldr/rapidjson_utils.h
kevinkreiser: I would have expected you to get a redefinition below can you post what happens when you remove this change

src/loki/status_action.cc
```diff
@@ -1,8 +1,43 @@
...
+#include "proto/status.pb.h"
 
 namespace valhalla {
 namespace loki {
-void loki_worker_t::status(Api&) const {
+void loki_worker_t::status(Api& request) const {
+  // get _some_ tile
+  static baldr::graph_tile_ptr tile = nullptr;
+  if (!tile) {
+    for (const auto& tile_id : reader->GetTileSet()) {
```
kevinkreiser, line 15: This is the only thing I'm worried about in this PR. This operation is potentially slow when running in tile dir mode. I would say we could just cache the results but it's possible that the caller wants to know what the uncached results are when the router is configured in an on the fly mode where the tileset is mutable

valhalla/baldr/rapidjson_utils.h
nilsnolde: On gcc 11.1.0 I get this:
```
[build] [95/123  69% :: 257.071] Building CXX object src/tyr/CMakeFiles/valhalla-tyr.dir/serializers.cc.o
[build] In file included from ../src/tyr/serializers.cc:15:
[build] ../valhalla/baldr/rapidjson_utils.h:22: warning: "RAPIDJSON_HAS_CXX11_NOEXCEPT" redefined
[build]    22 | #define RAPIDJSON_HAS_CXX11_NOEXCEPT 0
[build]       | 
[build] In file included from ../third_party/rapidjson/include/rapidjson/allocators.h:18,
[build]                  from ../src/tyr/serializers.cc:5:
[build] ../third_party/rapidjson/include/rapidjson/rapidjson.h:562: note: this is the location of the previous definition
[build]   562 | #define RAPIDJSON_HAS_CXX11_NOEXCEPT 1
[build]       | 
[build] In file included from ../src/tyr/serializers.cc:15:
[build] ../valhalla/baldr/rapidjson_utils.h:24: warning: "RAPIDJSON_HAS_STDSTRING" redefined
[build]    24 | #define RAPIDJSON_HAS_STDSTRING 1
[build]       | 
[build] In file included from ../third_party/rapidjson/include/rapidjson/allocators.h:18,
[build]                  from ../src/tyr/serializers.cc:5:
[build] ../third_party/rapidjson/include/rapidjson/rapidjson.h:134: note: this is the location of the previous definition
[build]   134 | #define RAPIDJSON_HAS_STDSTRING 0 // no std::string support by default
```

Was just sick of seeing this I guess, but open to other ways of avoiding that warning

src/loki/status_action.cc
nilsnolde: IMO it's fine for it to be a slow operation as long as it's cached (which it is right now), but I'm very unclear on what to do about that dynamic operation mode of Valhalla, I have little experience with the different operation modes, no idea how to set that up even..

Is there a way you'd prefer? 

Can we detect it's running in dynamic mode and simply skip most attributes (and the tile lookup) to basically only give back the full `config`?

kevinkreiser: my point is that its not cached when you arent using a tar, it will go to the filesystem and check which files are there: https://github.com/valhalla/valhalla/blob/master/src/baldr/graphreader.cc#L820

to me it seems a bit mean to have an implementation detail like this dictate the way this method works unless, like i was saying above, that someone really wants to see how their tile set is changing since the last call to this. to me though that seems like a very odd use-case. these top level apis are for down stream consumers that dont need to know the details of what is going on underneath so i think we can safely forget this usecase.

maybe we should do something like `locate` where we make use of the `verbose` option and only provide all this extra info when the person askes for `verbose==true`. that way you can opt into a longer more taxing request? any thoughts about that?

nilsnolde: got ya, thanks for the explanation.

adding `verbose` param (default false I guess) to the endpoint is fine with me. I'll do it that way then. 

Anything else you'd like to have in terms of tests @kevinkreiser ?

nilsnolde: tileset specific information is now only returned if `verbose: true`

valhalla/baldr/rapidjson_utils.h
nilsnolde: any preference here?

nilsnolde:
to re-iterate, this was your main concern @kevinkreiser :

> maybe we should do something like locate where we make use of the verbose option and only provide all this extra info when the person askes for verbose==true. that way you can opt into a longer more taxing request? any thoughts about that?

I implemented it exactly like that, reviewing won't take much effort I hope

kevinkreiser:
Yep I just am having a busy weekend. Let me check over it tonight

test/actor.cc
```diff
@@ -49,8 +49,12 @@ TEST(Actor, Basic) {
   auto transit = test::json_to_pt(transit_json);
   ASSERT_NE(transit_json.find(std::to_string(false)), std::string::npos);
 
-  auto status_json = actor.status("");
-  ASSERT_EQ(status_json, "{}");
+  actor.status("");
```
kevinkreiser, line 52: Should probably assert that this one is empty?

src/tyr/actor.cc
kevinkreiser: We need it for the verbose flag now

src/bindings/python/python.cc
```diff
@@ -93,6 +93,9 @@ struct simplified_actor_t : public valhalla::tyr::actor_t {
   std::string centroid(const std::string& request_str) {
     return valhalla::tyr::actor_t::centroid(request_str, nullptr, nullptr);
   };
+  std::string status() {
+    return valhalla::tyr::actor_t::status("", nullptr, nullptr);
```
kevinkreiser, line 97: We need to pass the request string here for the verbose flag

CHANGELOG.md
```diff
@@ -49,6 +50,7 @@
    * ADDED: Allow going through accessible `barrier=bollard` and penalize routing through it, when the access is private [#3175](https://github.com/valhalla/valhalla/pull/3175)
    * ADDED: Add country code to incident metadata [#3169](https://github.com/valhalla/valhalla/pull/3169)
    * CHANGED: Use distance instead of time to check limited sharing criteria [#3183](https://github.com/valhalla/valhalla/pull/3183)
+   * ADDED: Added vehicle width and height as an option for auto (and derived: taxi, bus, hov) profile (https://github.com/valhalla/valhalla/pull/3179)
```
kevinkreiser, line 53: What's the story with this?

src/loki/status_action.cc
```diff
@@ -1,8 +1,46 @@
...
 namespace valhalla {
 namespace loki {
-void loki_worker_t::status(Api&) const {
+void loki_worker_t::status(Api& request) const {
+
+  auto* status = request.mutable_status();
+
+  std::stringstream ss;
+  boost::property_tree::json_parser::write_json(ss, config);
+  status->set_config(ss.str());
```
kevinkreiser, line 17: I also wonder of this is a security risk that should be disablable via config since it tells config details. And also very unfortunate its ptree here because ptree serialize a everything as string iirc. Nothing specific to do just philosophical musings

src/loki/status_action.cc
```diff
@@ -1,8 +1,46 @@
...
+        if (tile->id().level() < baldr::TileHierarchy::GetTransitLevel().level &&
+            tile->header()->nodecount() > 0) {
+          break;
+        }
+      }
+    }
+
+    static std::string bbox = "";
+    if (bbox == "" && connectivity_map) {
+      bbox = connectivity_map->to_geojson(2);
```
kevinkreiser, line 35: One day I want to make the map use the spatial index since it has even higher resolution, maybe not interest for this endpoint yet but showing an optional 'coverage' geojson might be pretty cool in the future

kevinkreiser:
Couple little things but otherwise looks good.

CHANGELOG.md
nilsnolde: hasty c/p I assume ;)

src/loki/status_action.cc
nilsnolde: yeah it does serialize everything to string unfortunately..

src/bindings/python/python.cc
nilsnolde: good catch!

src/tyr/actor.cc
nilsnolde: true..

test/actor.cc
nilsnolde: it'll actually return the config and version. but yeah, should be tested anyways

nilsnolde:
Thanks! I addressed your comments @kevinkreiser

src/loki/status_action.cc
kevinkreiser: this one is really bothering me... what do you think about the security risk posed by allowing returning the config directly? maybe we need to:

1. put everything behind the verbose flag, to allow status to be just an endpoint that returns 200 when the service is up
2. add a "service limit" for the status "service" that disables the ability to ask for verbose output

in this way someone who is very security conscientious could keep using `status` for health checks but not open up potentially sensitive information (im not really sure if this is true) to regular service users

nilsnolde: > Nothing specific to do just philosophical musings

I took that quite literal ;) well there's still a few paths for logs and maybe even internal connection ports exposed, which could be used for smth nasty.

ok fair enough. tbh, I wouldn't even have included the config, so for me it's also an option to simply not return it at all. if you wanna keep it, we can do what you suggest as well, your call.

nilsnolde: hm. yeah forget it, I'll do it anyways, I like just having a health endpoint not doing anything, we could use that for docker health statuses as well.

src/loki/worker.cc
```diff
@@ -242,6 +242,7 @@ loki_worker_t::loki_worker_t(const boost::property_tree::ptree& config,
   max_best_paths = config.get<unsigned int>("service_limits.trace.max_best_paths");
   max_best_paths_shape = config.get<size_t>("service_limits.trace.max_best_paths_shape");
   max_alternates = config.get<unsigned int>("service_limits.max_alternates");
+  allow_verbose = config.get<bool>("allow_verbose", false);
```
nilsnolde, line 245: IMO it's fine to allow for a non-existing key in the valhalla.json here, what do you think?

src/tyr/serializers.cc
```diff
@@ -83,10 +85,48 @@ std::vector<std::string> openlr_edges(const TripLeg& leg) {
...
+    for (const auto& path :
+         {"/mjolnir/tile_dir", "/mjolnir/tile_extract", "/mjolnir/admin", "/mjolnir/timezone",
+          "/mjolnir/transit_dir", "/additional_data/elevation"}) {
+      config_doc.RemoveMember(path);
+    }
+    status_doc.AddMember("config", config_doc, alloc);
+  }
+
+  // if not verbose, return empty string
+  return status_doc.ObjectEmpty() ? "" : rapidjson::to_string(status_doc);
```
nilsnolde, line 129: return empty string rather than empty object

src/loki/status_action.cc
nilsnolde: done @kevinkreiser !

src/loki/worker.cc
kevinkreiser: yep we have a few like that, shoudl the key be `service_limits.status.allow_verbose`?

src/tyr/serializers.cc
kevinkreiser: but why? all of our apis are json. i guess you are thinking what is the point of an empty object. my only rebuttal is that then the client can at least parse every single response as json and not have one special condition that changes the response format `""` isnt valid json

src/worker.cc
```diff
@@ -730,7 +730,7 @@ void from_json(rapidjson::Document& doc, Options& options) {
     options.set_height_precision(*height_precision);
   }
 
-  options.set_verbose(rapidjson::get(doc, "/verbose", false));
+  options.set_verbose(rapidjson::get<bool>(doc, "/verbose", false));
```
kevinkreiser, line 733: hmm are there any side effects of this? i forget exactly but i put a lot of handling for allowing conversions between different types (especially primitive ones). does verbose=1 still work? the reason i did this was to maintain api compatibility for back when we were using ptree to parse json. it is very loose in this way and will convert integers and bools freely

src/tyr/serializers.cc
```diff
@@ -1,6 +1,8 @@
+#include <boost/algorithm/string/replace.hpp>
 #include <boost/property_tree/ptree.hpp>
 #include <cstdint>
 #include <functional>
+#include <rapidjson/allocators.h>
```
kevinkreiser, line 5: this is the cause of the preprocessor redefintion. we must only ever include rapidjson stuff via rapidjson_utils.h. if allocators.h isnt in there, add it over there and the preprocessor stuff will go away

src/worker.cc
nilsnolde: hm.. that seems like a artifact on my side

src/loki/worker.cc
nilsnolde: jup

src/tyr/serializers.cc
nilsnolde: aaah makes sense!

src/tyr/serializers.cc
nilsnolde: good point

nilsnolde:
now maybe;) thanks for the review and the good catches!

kevinkreiser:
@nilsnolde you forgot to revert the non verbose unit test i think, judging by the CI logs

nilsnolde:
yeah just fixing that..

src/loki/status_action.cc
```diff
@@ -1,8 +1,48 @@
...
+    }
+    status->set_bbox(bbox);
```
kevinkreiser, line 34: hmm i think we mustnt be testing all the permutations. i would suggest:

```suggestion
  // bail if we wont be getting extra info
  if (!request.options().verbose() || !allow_verbose)
    return;

  // get _some_ tile
  static baldr::graph_tile_ptr tile = nullptr;
  if (!tile) {
    for (const auto& tile_id : reader->GetTileSet()) {
      tile = reader->GetGraphTile(tile_id);
      if (tile->id().level() < baldr::TileHierarchy::GetTransitLevel().level &&
          tile->header()->nodecount() > 0) {
        break;
      }
    }

    if (connectivity_map) {
       status->set_bbox(connectivity_map->to_geojson(2));
    }
```

src/loki/status_action.cc
```diff
@@ -1,8 +1,48 @@
...
+    status->set_has_tiles(!tile ? false : true);
+    status->set_has_admins(!tile ? false : tile->header()->admincount() > 0);
+    status->set_has_timezones(!tile ? false : tile->node(0)->timezone() > 0);
```
kevinkreiser, line 37:
```suggestion
    status->set_has_tiles(tile);
    status->set_has_admins(tile && tile->header()->admincount() > 0);
    status->set_has_timezones(tile && tile->node(0)->timezone() > 0);
```

src/loki/status_action.cc
```diff
@@ -1,8 +1,49 @@
...
+          tile->header()->nodecount() > 0) {
+        break;
+      }
+    }
+
+    if (connectivity_map) {
+      status->set_bbox(connectivity_map->to_geojson(2));
+    }
+
+    status->set_has_tiles(!!tile);
```
nilsnolde, line 36: didn't know how else to make a boolean out of the tile ptr :D

nilsnolde: my foggy head isn't really capable of asserting that's correct right now though..

nilsnolde:
if ok like this, feel free to merge @kevinkreiser . otherwise I'll do the remaining tasks next week, need a little break from jumping projects;)

kevinkreiser:
It looks like just some tiny stuff I might just patch and merge

nilsnolde:
Got some headspace today again, so if you let me know what needs to be done, I’ll quickly do it @kevinkreiser :)

docs/api/status/api-reference.md
```diff
@@ -0,0 +1,19 @@
+# Status service API reference
+
+By default the `/status` endpoint will simply return a HTTP status code of 200 without a response text, acting as a health endpoint for the HTTP API.
```
kevinkreiser, line 3: this is no longer true, it returns an empty json object

src/loki/status_action.cc
```diff
@@ -1,8 +1,49 @@
...
+  // get _some_ tile
+  static baldr::graph_tile_ptr tile = nullptr;
+  if (!tile) {
+    for (const auto& tile_id : reader->GetTileSet()) {
+      tile = reader->GetGraphTile(tile_id);
+      if (tile->id().level() < baldr::TileHierarchy::GetTransitLevel().level &&
+          tile->header()->nodecount() > 0) {
+        break;
+      }
+    }
```
kevinkreiser, line 30:
```suggestion
  for (const auto& tile_id : reader->GetTileSet()) {
    tile = reader->GetGraphTile(tile_id);
    if (tile->id().level() < baldr::TileHierarchy::GetTransitLevel().level &&
        tile->header()->nodecount() > 0) {
      break;
    }
```

src/loki/status_action.cc
kevinkreiser, line 36:
```suggestion
    status->set_has_tiles(tile);
```

src/tyr/serializers.cc
```diff
@@ -83,10 +84,47 @@ std::vector<std::string> openlr_edges(const TripLeg& leg) {
...
+  rapidjson::Document config_doc;
+  if (request.status().has_config()) {
+    config_doc.Parse(request.status().config());
+    for (const auto& path :
+         {"/mjolnir/tile_dir", "/mjolnir/tile_extract", "/mjolnir/admin", "/mjolnir/timezone",
+          "/mjolnir/transit_dir", "/additional_data/elevation"}) {
+      config_doc.RemoveMember(path);
+    }
+    status_doc.AddMember("config", config_doc, alloc);
+  }
```
kevinkreiser, line 125: what do you think about removing the config from this endpoint entirely for the time being? i'm already disappointed by ptree bastardizing the json types and this question about sensitive information is another thing to mull over. we can add config back later from my point of view. what do you think?

valhalla/baldr/rapidjson_utils.h
kevinkreiser: you put the allocators include in here but forgot to remove this `undef`

```suggestion
```

kevinkreiser:
@nilsnolde i made 3 small suggestions and then had a 4th larger suggestion to just forget about serializing out the config for now. after that then it looks done to me

src/tyr/serializers.cc
nilsnolde: I’m happy without and agree about the security concerns for public setups. Would be nice to eventually ditch ptree and replace with rapidjson for all Json related stuff, guess that’s a blast from the past when rapidjson wasn’t where it is today?

src/loki/status_action.cc
nilsnolde: One question here @kevinkreiser : you want to acquire a tile on _every_ request for the scenario where the tileset is changing, right?

nilsnolde: then there's no point for a `static tile` anymore I guess

src/loki/status_action.cc
nilsnolde: well, like I said, the compiler doesn't like that:
```
[build] /home/nilsnolde/dev/cpp/valhalla/src/loki/status_action.cc:35:25: error: cannot convert ‘valhalla::baldr::graph_tile_ptr’ {aka ‘boost::intrusive_ptr<const valhalla::baldr::GraphTile>’} to ‘bool’
[build]    35 |   status->set_has_tiles(tile);
[build]       |                         ^~~~
[build]       |                         |
[build]       |                         valhalla::baldr::graph_tile_ptr {aka boost::intrusive_ptr<const valhalla::baldr::GraphTile>}
```

`!!tile` did work, but no idea why.. any suggestion how to make both you and the compiler happy? :)

nilsnolde: `status->set_has_tiles(static_cast<bool>(tile));` seems to compile

src/loki/status_action.cc
kevinkreiser: oh crap i didnt even see that it was static! yeah we cant use static in this way anyway as its not threadsafe. you can make it threadsafe by using a static initializer though. to do that you can take this loop and throw it into an anonymous namespace above and then initialize the static local variable with that:

```c++
namespace {
  auto get_graphtile_with_nodes_and_edges(valhalla::baldr::GraphReader& reader) {
    graph_tile_ptr tile = nullptr;
    for (const auto& tile_id : reader->GetTileSet()) {
      tile = reader->GetGraphTile(tile_id);
      if (tile->id().level() < baldr::TileHierarchy::GetTransitLevel().level &&
          tile->header()->nodecount() > 0) {
        break;
      }
    }
    return tile;
  }
}

...

namespace valhalla {
namespace loki {
  void loki_worker_t::status(Api& request) const {
    ...
    static tile = get_graphtile_with_nodes_and_edges(reader);  
    ...
  }
}
}
...
```

nilsnolde: Haha thx good we talked about it then.. and yeah makes sense about thread safety

src/loki/status_action.cc
kevinkreiser: its boolean conversion operator must be marked as explicit

kevinkreiser: yeah it is: https://www.boost.org/doc/libs/1_76_0/libs/smart_ptr/doc/html/smart_ptr.html#intrusive_ptr

`explicit operator bool () const noexcept;`

docs/api/status/api-reference.md
nilsnolde: fixed

nilsnolde:
Ok I think that's it:
- removed config as part of response from all files (proto etc)
- broke out the tile getter into an anonymous namespace and kept the local `static` variable
- removed the `undef` in rapidjson_utils.h
- minimized impact on loki_service.cc test since config is not returned anymore

puh, turned out to be head scratcher here and there :sweat_smile: thanks for your patience on this @kevinkreiser , you didn't exactly ask for this stuff

kevinkreiser:
@nilsnolde all good, sorry it took so long

src/loki/status_action.cc
```diff
@@ -1,8 +1,50 @@
+#include <boost/property_tree/json_parser.hpp>
+#include <iostream>
+#include <sstream>
```
kevinkreiser, line 3: these are superfluous now

src/loki/status_action.cc
```diff
@@ -1,8 +1,50 @@
...
+
+  auto* status = request.mutable_status();
+
+  // only return more info if explicitly asked for (can be very expensive)
+  // bail if we wont be getting extra info
+  if (!request.options().verbose() || !allow_verbose)
+    return;
+
+  // get _some_ tile
+  static baldr::graph_tile_ptr tile = get_graphtile(reader);
```
kevinkreiser, line 36: pedantic but should be `const` i think

kevinkreiser (approved):
two tiny nit picks, looks good

nilsnolde:
thanks, all resolved, will merge when CI is done

PR #3009, opened 2021-04-16
Pass in thor config to PathAlgorithm classes used in valhalla_run_route.

dnesbitt61:
# Issue

I noted performance drop in a long bicycle route that I test with valhalla_run_route using the multi-run option. PR #2949 added max_reserved_labels_count to the config file and passes the config file into PathAlgorithm constructors. However, these were not passed in within valhalla_run_route. I still view this program as a valid test application and it needs to be updated.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


dnesbitt61:
@mandeepsandhu - Question on the Init method (bidirectional as an example):

  edgelabels_forward_.reserve(std::min(max_reserved_labels_count_, kInitialEdgeLabelCountBD));
  edgelabels_reverse_.reserve(std::min(max_reserved_labels_count_, kInitialEdgeLabelCountBD));

Is this intended or should this just reserve max_reserved_labels_count_?  If I want to set max_reserved_labels_count higher than kInitialEdgeLabelCountBD it will always reserve the lower value meaning I cannot set max_reserved_labels_count such that a reallocation is not needed if edge label counts exceed kInitialEdgeLabelCountBD.

mandeepsandhu:
> @mandeepsandhu - Question on the Init method (bidirectional as an example):
> 
> edgelabels_forward_.reserve(std::min(max_reserved_labels_count_, kInitialEdgeLabelCountBD));
> edgelabels_reverse_.reserve(std::min(max_reserved_labels_count_, kInitialEdgeLabelCountBD));
> 
> Is this intended or should this just reserve max_reserved_labels_count_? If I want to set max_reserved_labels_count higher than kInitialEdgeLabelCountBD it will always reserve the lower value meaning I cannot set max_reserved_labels_count such that a reallocation is not needed if edge label counts exceed kInitialEdgeLabelCountBD.

This change was done in https://github.com/valhalla/valhalla/pull/2719. I'll let the original authors clarify.
CC: @genadz @averkhaturau @kevinkreiser

kevinkreiser:
@dnesbitt61 that does seem strange, i suspect what we wanted to do was in the constructor we should use the constant in place of the config value if the config value is not present. @genadz @averkhaturau am i forgetting something?

dnesbitt61:
I suggest it should just be:
  edgelabels_forward_.reserve(max_reserved_labels_count_);
  edgelabels_reverse_.reserve(max_reserved_labels_count_);
This way I can set the reserved labels count in such a way that there shouldn't be any reallocation during path finding (assuming there is sufficient memory on the platform).

mandeepsandhu:
> I suggest it should just be:
> edgelabels_forward_.reserve(max_reserved_labels_count_);
> edgelabels_reverse_.reserve(max_reserved_labels_count_);
> This way I can set the reserved labels count in such a way that there shouldn't be any reallocation during path finding (assuming there is sufficient memory on the platform).

That looks right. Is `max_reserved_labels_count_` sanitized somewhere upstream of this callsite?

genadz:
> @dnesbitt61 that does seem strange, i suspect what we wanted to do was in the constructor we should use the constant in place of the config value if the config value is not present. @genadz @averkhaturau am i forgetting something?

agree, using `std::min(max_reserved_labels_count_, kInitialEdgeLabelCountBD)` looks ugly. the main purpose of this code was to prevent reserving more than `max_reserved_labels_count` . as far as we initialize `max_reserved_labels_count` with `kInitialEdgeLabelCountBD` in constructor, it's not necessary to call `std::min...` anymore

src/valhalla_run_route.cc
```diff
@@ -662,11 +662,10 @@ int main(int argc, char* argv[]) {
...
   // Get the route
-  TimeDepForward astar;
-  BidirectionalAStar bd;
-  MultiModalPathAlgorithm mm;
-  TimeDepForward timedep_forward;
-  TimeDepReverse timedep_reverse;
+  BidirectionalAStar bd(pt.get_child("thor"));
+  MultiModalPathAlgorithm mm(pt.get_child("thor"));
+  TimeDepForward timedep_forward(pt.get_child("thor"));
+  TimeDepReverse timedep_reverse(pt.get_child("thor"));
```
kevinkreiser, line 668: i believe we'll need to change these since the latest refactors/renames to the timedep classes

kevinkreiser (approved):

PR #3013, opened 2021-04-17
add python bindings to full mac os ci build

nilsnolde:
Let's see if this runs on the first try..

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 commands:
   install_macos_dependencies:
     steps:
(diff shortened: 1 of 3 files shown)


nilsnolde:
of course not..

looks like boost-python had bpo as dependency.. can't seem to find a brew way to install only bpo. I'll install whole boost, let me know if you'd prefer only boost-python, which seemed to work as well.

nilsnolde:
finally works, good to go..

kevinkreiser (approved):
Looks good

Issue #2612, opened 2020-09-23
French prononciation of route number over 99


st4141:
I'm using Valhalla in French ("fr-FR") and it seems that an English usage is used even in French.
In French, road's number over 99 should be be separated in two part
For example, road "D 993" is written "D 9 93" in verbal_xxx_instruction. It should stay "D 993".
I have no idea of how German, Italian, etc. pronounce such road names.

extract from the result of
_curl http://localhost:8002/route --data '{"locations":[{"lat":48.85856667,"lon":2.40277778, "type":"break", "name":"maison"},{"lat":48.87338333,"lon":2.29666667, "name":"Arret Arc de Triomphe", "type":"through"},{"lat":48.97155833,"lon":2.07861111, "type":"break", "name":"Systrans"}],"costing":"bus","directions_options":{"language":"fr-FR"}}' | jq '.'_

{
"travel_type": "bus",
"travel_mode": "drive",
"verbal_pre_transition_instruction": "Serrez à droite dans **D 9 93**.",
"verbal_transition_alert_instruction": "Serrez à droite dans **D 9 93**.",
"length": 1.928,
"instruction": "Serrez à droite dans **D 993**.",
"end_shape_index": 641,
"type": 9,
"time": 182.564,
"verbal_post_transition_instruction": "Continuez pendant 2 kilomètres.",
"street_names": [
"**D 993**"
],
"begin_shape_index": 533
},

PR #3015, opened 2021-04-19
Remove large number formatting for non-US countries

dgearhart:
# Issue

Fixes https://github.com/valhalla/valhalla/issues/2612

## Tasklist

 - [x] Add tests


test/gurka/test_instructions_obvious_maneuver.cc
```diff
@@ -135,13 +135,13 @@ TEST_F(InstructionsObviousManeuver, NotObviousBecauseSameNameIntersectingEdgeRam
   // Verify start maneuver prior to the continune maneuver
   gurka::assert::raw::expect_instructions_at_maneuver_index(result, maneuver_index,
                                                             "Drive east on US 322.", "",
-                                                            "Drive east on US 3 22.",
+                                                            "Drive east on US 322.",
```
danpaz, line 138: Are these tests updated because they're using netherlands admin?

dgearhart: @danpaz yes, exactly

danpaz: 🤔 I'm wondering if this formatting should really be based on the request language and not the country. In Canada and even in Europe I think it's desired to pronounce the digits separately, e.g. the example in one of the tests below `A120` seems like it'd be a regression if the instruction language is in English rather than French. @st4141 as you wrote the original issue https://github.com/valhalla/valhalla/issues/2612 do you have any thoughts on this?

dgearhart: @danpaz yes, I was thinking the same thing about the language in the future. However, right now I just wanted to fix the current issue.

st4141: @danpaz I guess the formatting should be based on the language of the request as you said, not on the country where the road is.

dgearhart: @danpaz also, tested the `A120` phrase and it sounded appropriate.

dgearhart: @danpaz @st4141 Currently, the formatter is based on location. I will submit an issue to enhance to use language. However, this PR fixes the number formatting for outside the US since that should have only applied to the US and is not English specific

danpaz: > @danpaz @st4141 Currently, the formatter is based on location. I will submit an issue to enhance to use language. However, this PR fixes the number formatting for outside the US since that should have only applied to the US.

👍 sounds good to tackle this in a separate issue.

mandeepsandhu: Is there an issue for using request language instead of country for the text formatter? Rest lgtm.

mandeepsandhu (approved):

PR #3020, opened 2021-04-20
Odin undefined behaviour: handle case when xedgeuse is not initialized

merkispavel:
# Issue

There real world cases when xedge is not initialized after `GetStraightestTraversableIntersectingEdgeTurnDegree` call, i.e if the list is intersecting edge_size = 0 https://github.com/valhalla/valhalla/blob/ce8ab9d82610cfbe90959b7fd3422ef7e4ed1851/src/odin/enhancedtrippath.cc#L1736

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
    * FIXED: Fix run_with_server.py to work on macOS [#3003](https://github.com/valhalla/valhalla/pull/3003)
    * FIXED: Removed unexpected maneuvers at sharp bends [#2968](https://github.com/valhalla/valhalla/pull/2968)
    * FIXED: Remove large number formatting for non-US countries [#3015](https://github.com/valhalla/valhalla/pull/3015)
+   * FIXED: Odin undefined behaviour: handle case when xedgeuse is not initialized [#3020](https://github.com/valhalla/valhalla/pull/3020)
 
 * **Enhancement**
    * Pedestrian crossing should be a separate TripLeg_Use [#2950](https://github.com/valhalla/valhalla/pull/2950)
--- a/src/odin/enhancedtrippath.cc
+++ b/src/odin/enhancedtrippath.cc
@@ -1728,7 +1728,7 @@ bool EnhancedTripLeg_Node::HasSpecifiedRoadClassXEdge(const RoadClass road_class
 uint32_t EnhancedTripLeg_Node::GetStraightestTraversableIntersectingEdgeTurnDegree(
     uint32_t from_heading,
     const TripLeg_TravelMode travel_mode,
-    TripLeg_Use* use) {
+    boost::optional<TripLeg_Use>* use) {
 
   uint32_t staightest_turn_degree = 180; // Initialize to reverse turn degree
   uint32_t staightest_delta = 180;       // Initialize to reverse delta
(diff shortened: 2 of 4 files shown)


mandeepsandhu (approved):

merkispavel:
@dgearhart I have not ideas how to test it as this variable isn't visible outside `IsPedestrianFork` function 😕

mandeepsandhu:
> @dgearhart I have not ideas how to test it as this variable isn't visible outside `IsPedestrianFork` function

Actually I have question - the way its written currently, nothing really sets `xedge_use` correct? `GetStraightestTraversableIntersectingEdgeTurnDegree()` will only set `xedge_use` if its not null, and since the caller - `IsPedestrianFork()`, never sets the pointer, it never gets used within it.

Previously, `xedge_use` had an undefined value, but `GetStraightestTraversableIntersectingEdgeTurnDegree()` could conditionally set it (eg if `xedge->has_use == true`). Now it seems it will never get set since its always initialized to `nullptr` when called in the context of `IsPedestrianFork`. Or am I missing somerthing?

mandeepsandhu (changes requested):
Left a comment, more for my clarification

merkispavel:
> since the caller - IsPedestrianFork(), never sets the pointer, it never gets used within it

The caller does set the "pointer"(i.e pointer to `boost::optional`).

```cpp
boost::optional<TripLeg_Use> xedge_use;
// xedgeuse is a valid variable so &xedgeuse != nullptr
GetStraightestTraversableIntersectingEdgeTurnDegree(...,..., &xedge_use);
```

Check out the difference with the following piece of code
```
boost::optional<TripLeg_Use>* xedge_use = nulltpr;
GetStraightestTraversableIntersectingEdgeTurnDegree(...,..., xedge_use); // xedgeuse is null obviously
```

mandeepsandhu:
> > since the caller - IsPedestrianFork(), never sets the pointer, it never gets used within it
> 
> The caller does set the "pointer"(i.e pointer to `boost::optional`).
> 
> ```c++
> boost::optional<TripLeg_Use> xedge_use;
> // xedgeuse is a valid variable so &xedgeuse != nullptr
> GetStraightestTraversableIntersectingEdgeTurnDegree(...,..., &xedge_use);
> ```
> 
> Check out the difference with the following piece of code
> 
> ```c++
> boost::optional<TripLeg_Use>* xedge_use = nulltpr;
> GetStraightestTraversableIntersectingEdgeTurnDegree(...,..., xedge_use); // xedgeuse is null obviously
> ```

Yup, makes sense. Thanks for the explanation! :+1:

mandeepsandhu (approved):

dgearhart:
@merkispavel I would at least run RAD using the branch

merkispavel:
> @merkispavel I would at least run RAD (master vs. branch) with the following files:
> 
> ```
> bicycle_routes.txt
> demo_routes.txt
> de_benchmark_routes.txt
> eu_roundabout_routes.txt
> pedestrian_roundabout_routes.txt
> pedestrian_routes.txt
> roundabout_routes.txt
> timed_access_routes.txt
> timed_conditional_routes.txt
> turn_lane_routes.txt
> ```

I RAD-tested the changes 🟢

PR #3025, opened 2021-04-21
CI Updates for Docker Builds

kevinkreiser:
This pr attempts to do 3 things:

1. fix the tagging mechanism for docker builds, basically the command wouldnt have worked if CI would have tried to run it
2. remove builds that we only ran on tag but no one cared about
3. modify the workflows so that tags will run only docker (previously they only ran the jobs that i removed in `2.` of this list)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,59 +35,11 @@ jobs:
               docker push valhalla/valhalla:run-latest
             elif [[ ! -z "${CIRCLE_TAG}" ]]; then
               echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+              docker tag valhalla/valhalla:build-latest valhalla/valhalla:build-${CIRCLE_TAG}
               docker push valhalla/valhalla:build-${CIRCLE_TAG}
+              docker tag valhalla/valhalla:run-latest valhalla/valhalla:run-${CIRCLE_TAG}
               docker push valhalla/valhalla:run-${CIRCLE_TAG}
             fi
-
-  build-release-binary-linux:
-    docker:
-      - image: valhalla/valhalla:build-latest
-    steps:
-      - checkout
-      - run: ./scripts/format.sh && ./scripts/error_on_dirty.sh
-      - run: git submodule sync && git submodule update --init
-      - restore_cache:
-          keys:
-            - ccache-release-linux-x86_64-{{ .Branch }}
-            - ccache-release-linux-x86_64
-      - run: mkdir build
-      - run: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -35,60 +35,12 @@ jobs:
               docker push valhalla/valhalla:run-latest
             elif [[ ! -z "${CIRCLE_TAG}" ]]; then
               echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+              docker tag valhalla/valhalla:build-latest valhalla/valhalla:build-${CIRCLE_TAG}
               docker push valhalla/valhalla:build-${CIRCLE_TAG}
+              docker tag valhalla/valhalla:run-latest valhalla/valhalla:run-${CIRCLE_TAG}
```
kevinkreiser, line 40: if you look above in the build step of this job you'll see regardless of master or tag we always tag the local image with `-latest`. if this was a github tag though that wouldnt have worked because here we have docker push with the git tag suffix on the image tag name.

.circleci/config.yml
```diff
@@ -35,60 +35,12 @@ jobs:
...
               docker push valhalla/valhalla:run-${CIRCLE_TAG}
             fi
 
-  build-release-binary-linux:
```
kevinkreiser, line 42: these were only run on git tags

.circleci/config.yml
```diff
@@ -198,73 +150,20 @@ jobs:
           paths:
             - ~/.ccache
 
-  build-release-binary-osx:
```
kevinkreiser, line 201: same, these were only run on git tags and no one cares about their success or failure

.circleci/config.yml
```diff
@@ -198,73 +150,20 @@ jobs:
...
-              only: /.*/
-            branches:
               ignore: /.*/
-      - build-debug-binary-linux:
+      - build-osx
           filters:
             tags:
-              only: /.*/
-            branches:
               ignore: /.*/
```
kevinkreiser, line 169: this part of the diff is hard to read but basically this is what i was seeing. when i tag the repo and it kicks off a build, because we had `filters.tags.only = .*` these were the only builds that got run on the tag. which means that when tagged the actual docker build was not run! see here:

image

it only ran those, which are useless to everyone. so ive changed it here so that of the builds that we care about (linux debug, linux release, docker and macos) tags will skip all of them execpt the docker build because the point of our tagging is basically to create runnable/usable artifacts rather than check that the code is working (presumably we do that by merging to master before tagging :smile: )

PureW (approved):
Makes sense to me

PR #3026, opened 2021-04-21
Avoid generating self-intersecting isochrone polygons that would otherwise result from ::Generalize()

ktatterso:
# Issue

When generating isochrones, customers can specify a numeric value, `generalize` or `gen_factor`, which simplifies the returned polygon.  This is particularly useful for reducing the size of the polygon, and hence, the returned payload.

The polyline/polygon simplification method employed by Valhalla is the Douglas-Peucker algorithm.  On its own, it has no facilities that would prevent it from generating a self-intersecting polygon during its simplification step.  Based on my studies, using a `gen_factor=55`, auto costing, and a 30 min contour - the probability of us returning a self-intersecting polygon is quite high, I'd estimate 80-90% (urban areas).

This is problematic for customers/users who take these isochrone polygons and input them to other GIS systems for analysis - only to find their analysis tools don't work with self-intersecting polygons.

My solution to avoid generating self-intersections all boils down to this question:  will simplifying this polyline result in a self-intersection?  To determine the answer to that question poses a spatial problem:  are there any polygon points in the "vicinity" of the polyline that would cause a self-intersection with the polyline I might simplify?

To answer the question "are there any points near this polyline I'm about to simplify?" I have created the `PointTileIndex` class.  This class can discretize the lat/lon space into whatever width is desired - in this case, `gen_factor` determines the width.  The `PointTileIndex` constructor will "bin" all the given points into the index.  Then, within the Douglas-Peucker algorithm, we can use the `PointTileIndex` to answer that very question:  "what points are near this polyline?"

Armed with a set of "points near my polyline",  we can very accurately ask, "will any of these nearby points cause a self-intersection if I simplify?"  Geometrically, you can think of line simplification as the geometric collapse of many triangles.  The triangles in question are formed by the start position, the end position, and each point along the polyline.  Hence, if a stray polygon point is found in any of the triangles we're about to collapse, we would cause a self-intersection if we simplified.
 
# Example 1

Location:  47.412426,9.334023
Costing:  Auto
Contour minutes:  30

Generalize:  0  **(no self-intersection)**

image

Generalize:  55  (without fix, self-intersects)

image

Generalize:  55  (after fix, no self-intersection)

image

# Example 2

Location:  34.061657,-118.268494
Costing:  Auto
Contour minutes:  30

Generalize:  0  **(no self-intersection)**

image

Generalize:  55  (without fix, self-intersects)

image

Generalize:  55  (after fix, no self-intersection)

image

# Performance

The `::tile()` method is O(n).  Asking for nearby points is k*O(1) where k is the number of nearby tiles.  Fortunately, most segments considered for simplification aren't very long, so k is small.

For example, the 30-min auto isochrone (`gen_factor=55`) for Orlando previously took 560 ms (not including serialization).  The `::Generalize()` step took 6 ms of that 560 ms.  After my changes, `::Generalize()` takes 22 ms, increasing the total time to 578 ms.

# Thoughts & other ideas

I tried using Boost's `boost::geometry::buffer()` method with a `distance_strategy=0.0`, thinking if I gave it a self-intersecting polygon it would fix it.  Unfortunately, it did not.  Given a self-intersecting polygon it would return an empty result.  In contrast, it worked fine for non-self-intersecting polygons.

I also wondered about other routines that might heal/repair/fix self-intersecting polygons.  In the end, I decided that this approach would  keep with the spirit of the original algorithm, only avoiding self-intersections as needed.

## Tasklist

 - [x] Add tests


src/midgard/point_tile_index.cc
```diff
@@ -0,0 +1,100 @@
...
+  tiled_space.reserve(polyline.size());
+  size_t index = 0;
+  for (auto iter = polyline.begin(); iter != polyline.end(); iter++, index++) {
+    this->points.emplace_back(*iter);
+    TileId pid = get_tile_id(*iter);
+    auto map_iter = tiled_space.find(pid);
+    if (map_iter == tiled_space.end())
+      tiled_space.insert(std::pair<TileId, std::unordered_set<size_t>>{pid, {index}});
+    else
+      map_iter->second.insert(index);
```
merkispavel, line 38:
```suggestion
    tiled_space[pid].insert(index)
```

src/midgard/point_tile_index.cc
```diff
@@ -0,0 +1,100 @@
...
+      auto iter = tiled_space.find(TileId{x, y});
+      if (iter != tiled_space.end()) {
+        near_pts.insert(iter->second.begin(), iter->second.end());
+      }
+    }
+  }
+}
+
+void PointTileIndex::get_points_near_segment(const LineSegment2<PointLL>& seg,
+                                             std::unordered_set<size_t>& near_pts) {
```
merkispavel, line 59: why dont we just return near points
```suggestion
std::unordered_set<size_t> PointTileIndex::get_points_near(const PointLL& pt)
std::unordered_set<size_t> PointTileIndex::get_points_near_segment(const LineSegment2<PointLL>& seg)
```

gknisely:
@ktatterso Can we get some before and after screen shots of the results.

src/midgard/polyline2.cc
```diff
@@ -42,17 +44,209 @@ typename coord_t::value_type Polyline2<coord_t>::Length(const container_t& pts)
...
+void DouglastPeuckerAvoidSelfIntersection(container_t& polyline,
+                                          typename coord_t::value_type epsilon_m,
+                                          const std::unordered_set<size_t>& exclusions) {
+  // Create a tile-space-index which allows us to perform a Douglas-Peucker style line
+  // simplification that avoids creating self-intersections within the polyline/polygon.
+  const PointLL& first_point = *polyline.begin();
+  double meters_per_deg = DistanceApproximator<PointLL>::MetersPerLngDegree(first_point.lat());
+  double epsilon_deg = epsilon_m / meters_per_deg;
+  PointTileIndex point_tile_index(epsilon_deg);
+  point_tile_index.tile<container_t>(polyline);
```
merkispavel, line 232: nit: doesn't compiler deduce container type?
```suggestion
  point_tile_index.tile(polyline);
```

src/midgard/polyline2.cc
```diff
@@ -42,17 +44,209 @@ typename coord_t::value_type Polyline2<coord_t>::Length(const container_t& pts)
...
+    if (pt != PointTileIndex::deleted_point) {
+      polyline.push_back(pt);
+    }
+  }
+}
+
+template <class coord_t, class container_t>
+void DouglasPeucker(container_t& polyline,
+                    typename coord_t::value_type epsilon,
+                    const std::unordered_set<size_t>& indices) {
```
merkispavel, line 249: what do you think about unifying parameters names(e.g use `exclusions` here as well)

merkispavel (changes requested):
First round of review

src/midgard/polyline2.cc
```diff
@@ -42,17 +44,209 @@ typename coord_t::value_type Polyline2<coord_t>::Length(const container_t& pts)
...
+
+    bool can_simplify = true;
+    for (size_t cidx = sidx + 1; (cidx < eidx) && can_simplify; cidx++) {
+      const PointLL& c = point_tile_index.points[cidx];
+
+      for (size_t point_idx : line_buffer_points) {
+        const PointLL& p = point_tile_index.points[point_idx];
+        if (triangle_contains(start, c, end, p)) {
+          can_simplify = false;
+#if 0
```
gknisely, line 179: So be `#ifdef LOGGING_LEVEL_TRACE` no?

valhalla/midgard/point_tile_index.h
```diff
@@ -0,0 +1,135 @@
...
+  uint32_t prevy(uint32_t y) {
+    return (y > 0) ? y - 1 : num_y_subdivisions - 1;
+  }
+  uint32_t nexty(uint32_t y) {
+    return (y < num_y_subdivisions - 1) ? y + 1 : 0;
+  }
+
+  struct TileId_hash_functor {
+    size_t operator()(const TileId& tid) const {
+#if 0
```
gknisely, line 64: What is the intent of this if block?

ktatterso:
> @ktatterso Can we get some before and after screen shots of the results.

I added some examples.  thanks.

test/isochrone.cc
```diff
@@ -134,26 +134,27 @@ TEST(Isochrones, Basic) {
...
+    printf("third case:\n");
```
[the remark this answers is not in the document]
ktatterso: Oops, thanks.

valhalla/midgard/polyline2.h
```diff
@@ -63,6 +64,15 @@ template <class coord_t> class Polyline2 {
...
   template <class container_t> static typename coord_t::value_type Length(const container_t& pts);
 
+  /**
+   * In an O(n^2) manner (only useful for debugging/testing), checks if there are any
+   * segments in the polyline that are intersecting. If so, returns those intersection
+   * points.
+   * @param   pts  Polyline vertices.
+   * @return  List of intersections found, if any.
+   */
+  std::list<coord_t> GetSelfIntersections();
```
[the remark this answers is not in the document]
ktatterso: Happy to make that change.  

Just fyi, my reason for selecting `list` was that the routine has no idea how many intersections it might return, so being unable to make a guess how much space should be `reserve`ed,  I decided to use `list` which unlike `vector` has low growth cost.

src/midgard/polyline2.cc
```diff
@@ -42,17 +44,209 @@ typename coord_t::value_type Polyline2<coord_t>::Length(const container_t& pts)
...
+    const coord_t& ib(points[i]);
+    for (size_t j = i + 2; j < points.size() - 1; j++) {
+      const coord_t& ja(points[j - 1]);
+      const coord_t& jb(points[j]);
+      LineSegment2<coord_t> segmenti(ia, ib);
+      LineSegment2<coord_t> segmentj(ja, jb);
+      coord_t intersection_point;
+      if (segmenti.Intersect(segmentj, intersection_point)) {
+        intersections.emplace_back(std::move(intersection_point));
+      }
```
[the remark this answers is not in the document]
ktatterso: Good catch, thanks.

valhalla/midgard/point_tile_index.h
ktatterso: Apparently someone in valhalla's past has thought about ways to hash-combine (x,y)'s as a means for well-distributed (performant) hashing.  I simply wanted to point out that I did try using it but found that it wasn't as performant as my approach.

Happy to remove it... I'll simply add a comment to the same effect.

src/midgard/polyline2.cc
ktatterso: Fair suggestion.  As to my thinking... this was the original routine, so I wanted to leave it unmodified.  Honestly tho, `indicies` just isn't very descriptive, so I am happy to rename it to `exclusions`.  I'll also give consideration to other variables that might need better names.

valhalla/midgard/point_tile_index.h
```diff
@@ -0,0 +1,135 @@
...
+  inline TileId get_tile_id(const PointLL& pt) {
+    TileId tid;
+    tid.x = std::lround(num_x_subdivisions * ((pt.lng() + max_x_range / 2.0) / max_x_range));
+    tid.y = std::lround(num_y_subdivisions * ((pt.lat() + max_y_range / 2.0) / max_y_range));
+    return tid;
+  }
+
+public:
+  // when points are deleted from our polyline they become this "special" value
+  static const PointLL deleted_point;
```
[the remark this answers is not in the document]
ktatterso: Sure.

src/midgard/point_tile_index.cc
```diff
@@ -0,0 +1,100 @@
...
+
+PointTileIndex::PointTileIndex(double tile_width_degrees) {
+  int level = 1;
+  num_x_subdivisions = 1;
+  double degrees_at_level = 180.0;
+  constexpr int max_level = 32;
+  while ((degrees_at_level > tile_width_degrees) && (level <= max_level)) {
+    level++;
+    degrees_at_level /= 2.0;
+    num_x_subdivisions = num_x_subdivisions << 1;
```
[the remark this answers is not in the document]
ktatterso: Sure.

valhalla/midgard/point_tile_index.h
```diff
@@ -0,0 +1,135 @@
...
+  // from your segment. The caller can decide what type of distance calculation
+  // to use to determine exact distances.
+  void get_points_near_segment(const LineSegment2<PointLL>& seg, std::unordered_set<size_t>& points);
+
+  // Removes a point from our tiled space given its index.
+  void remove_point(const size_t& idx);
+
+  // Remove a range of points by index from the tiled space. This includes
+  // the point from sidx (inclusive) to eidx (exclusive).
+  void remove_points(const size_t& sidx, const size_t& eidx) {
```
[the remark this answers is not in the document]
ktatterso: Sure.

valhalla/midgard/point_tile_index.h
```diff
@@ -0,0 +1,135 @@
...
+  void get_points_near(const PointLL& pt, std::unordered_set<size_t>& points);
+
+  // Get all the points roughly within the "tile_width_degrees" of the given
+  // segment. Some of the returned points could be as far as 2*tile_width_degrees
+  // from your segment. The caller can decide what type of distance calculation
+  // to use to determine exact distances.
+  void get_points_near_segment(const LineSegment2<PointLL>& seg, std::unordered_set<size_t>& points);
+
+  // Removes a point from our tiled space given its index.
+  void remove_point(const size_t& idx);
```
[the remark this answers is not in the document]
ktatterso: Sure.

valhalla/midgard/point_tile_index.h
```diff
@@ -0,0 +1,135 @@
...
+ * points are near me" - in effectively O(1) time (the speed of a hash
+ * lookup).
+ *
+ * Of important note is the fact that to determine the "points within an
+ * exact distance of a given point" you must round up all points in all
+ * surrounding tiles. You can subsequently compute the distance bewteen your
+ * point and each "nearby" point using a more accurate distance computation
+ * of your choosing.
+ */
+class PointTileIndex {
```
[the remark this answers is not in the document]
ktatterso: Ack!  That looks very similar to what I'm doing.  I swear I grepped around looking for something similar, somehow I missed this.  😞 .  Yes, this deserves some investigation.  Nice catch.

src/midgard/point_tile_index.cc
ktatterso: Nice one, thanks.

src/midgard/polyline2.cc
merkispavel: >  this was the original routine, so I wanted to leave it unmodified
Agreed, I prefer not to modify the code I don't touch. It's just the specificity of this case that you introduce similar functions. Of course, this is a "soft suggestion"(non-mandatory). I'm ok with both outcomes(unified/not unified)

src/midgard/point_tile_index.cc
ktatterso: Yea I think about that a lot.  C++ has evolved a lot and is making every effort (via optimizations like RVO, copy elison, move operators) to make returning complex objects an efficient and idiomatic part of the language.  I haven't checked lately, but some of those optimizations I mentioned are disabled in debug builds.  Hence, when returning complex objects, the debug build will fallback to copy-construction which would suffer performance.  Also, I've read that it is easy to "defeat" RVO by simply adding early `return`s to your function.  

I certainly have some more learning to do on these subjects, especially since compilers are making continuous improvements on these fronts.  However, I can say with (almost) absolute certainty that the manner in which I've written those methods is as efficient as can be written.  Nothing is left to hoping the compiler optimizer moves things efficiently - and debug builds suffer nothing.

All that said, I'm certainly willing to modify those routines as suggested.  Being correct is priority zero - being consistent is right behind.  Just wanted to share my thinking.  thanks.

src/midgard/polyline2.cc
ktatterso: Right... I had previously templated `tile()` with multiple things so the compiler needed a clue.  I later removed the extra templating and this was left behind.  thx.

valhalla/midgard/point_tile_index.h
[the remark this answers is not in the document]
kevinkreiser: yeah was going to say the same thing, thanks @merkispavel !

src/midgard/polyline2.cc
ktatterso: After thinking about it, I've decided that `#if 0` block was just for debugging, so I've removed it.  thx.

valhalla/midgard/point_tile_index.h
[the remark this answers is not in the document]
ktatterso: I was able to use the Tile class as the means for creating a tiled space.  Thanks.

valhalla/midgard/point_tile_index.h
```diff
@@ -0,0 +1,71 @@
...
+ * lookup).
+ *
+ * Of important note is the fact that to determine the "points within an
+ * exact distance of a given point" you must 1) use the get_points_near*()
+ * methods to round-up nearby points, then 2) compute the distance those
+ * nearby points using a more accurate distance computation of your choosing.
+ */
+class PointTileIndex {
+  // This guy can tell us which tile a points belongs to.
+  std::unique_ptr<Tiles<PointLL>> tiles;
```
ktatterso, line 28: Switched to using the `Tiles` class for creating a gridded space over the given points-of-interest.  Works well and performs equal to my previous solution.

src/midgard/point_tile_index.cc
```diff
@@ -0,0 +1,138 @@
...
+    return;
+  if (tile_width_degrees <= 0.0)
+    return;
+
+  // Determine the extents of the points we want to index. This determines
+  // the size of our Tile space.
+  double min_lat = 1000.0, max_lat = -1000.0;
+  double min_lng = 1000.0, max_lng = -1000.0;
+  for (auto iter = polyline.begin(); iter != polyline.end(); iter++) {
+    const PointLL& p = *iter;
```
merkispavel, line 25: nit: 
```suggestion
  for (const auto& p : polyline) {
```

src/midgard/point_tile_index.cc
```diff
@@ -0,0 +1,138 @@
...
+    if (p.lat() < min_lat)
+      min_lat = p.lat();
+    if (p.lat() > max_lat)
+      max_lat = p.lat();
+    if (p.lng() < min_lng)
+      min_lng = p.lng();
+    if (p.lng() > max_lng)
+      max_lng = p.lng();
```
merkispavel, line 33: nit
```suggestion
    min_lat = std::min(min_lat, p.lat());
    max_lat = std::min(max_lat, p.lat());
    min_lng = std::min(min_lng, p.lng());
    max_lng = std::max(max_lng, p.lng());
```

merkispavel (dismissed):
LGTM

src/midgard/CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
 file(GLOB headers ${VALHALLA_SOURCE_DIR}/valhalla/midgard/*.h)
 
-set_source_files_properties(
+if ((UNIX OR APPLE) AND ENABLE_SINGLE_FILES_WERROR)
```
merkispavel, line 3: I believe APPLE is unix https://cmake.org/cmake/help/v3.6/variable/UNIX.html

merkispavel (approved):
Re-approve

Issue #441, opened 2017-02-02
what():  blob-header-size is bigger than allowed 1010792557 > 65536


kevinkreiser:
(2016-10-17T07:37:23Z) @TommasoDb : originally created valhalla/mjolnir#547

(2016-10-17T12:15:52Z) @kevinkreiser : we only support osm pbf format at this time. take your `.osm` and convert it to pbf with `osmconvert`. you can get that in ubuntu via `sudo apt-get install osmctools`. once your file is in proper osm pbf format `valhalla_build_tiles` should like it a lot better :wink: 

(2016-10-18T08:54:45Z) @TommasoDb : Thank you very much, it works

PR #2987, opened 2021-04-08
Reduce code duplication in timedep forward/reverse

danpat:
# Issue

PR No. 3 in the series towards https://github.com/valhalla/valhalla/issues/2885.  This one tackles TimeDepForward/Reverse.

Main changes in this are:

  - Rename `TimeDep` to `UnidirectionalAStar` to be consistent with `BidirectionalAStar`
  - Templatize the behaviour and merge forward/reverse handling into a single `unidirectionalastar.cc` codebase.
  - Fix a bug in reverse routing that lead to incorrect routes, node counts and ETAs

## Tasklist

 - [x] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [x] Update the docs with any new request parameters or changes to behavior described
 - [x] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

Issue #2940, opened 2021-03-15
Incorrect path edges if location snaps to a node


genadz:
I noticed that in unidirectional astar algorithms we handle cases when location snaps exactly to the node not always correctly.

**TimedepForward**
A simple example:
```
A--------1----B
```
Let's suppose we build route from `A` to `1`. In this case we will not filter out edge `BA` (opposing to `AB`) and will get the path route `BA AB` instead of `AB`. Something wrong with logic that handles trivial paths: https://github.com/valhalla/valhalla/blob/master/src/thor/timedep_forward.cc#L448 .

**TimedepReverse**
I've noticed this https://github.com/valhalla/valhalla/blob/master/src/thor/timedep_reverse.cc#L549 - we skip **outbound** path edges for the origin (instead of inbound).

mandeepsandhu:
I found something similar in another test-case here: https://github.com/valhalla/valhalla/issues/2709 
@kevinkreiser explained a bit more about how node-snapping works in his [comment](https://github.com/valhalla/valhalla/issues/2709#issuecomment-755727783) on that issue. Maybe that sheds some light on the bug you're seeing.

mandeepsandhu:
Likely fixed by https://github.com/valhalla/valhalla/pull/2987. We should check if the issue is reproducible in master.

PR #3029, opened 2021-04-23
handling time dependent restrictions

CuriOusQuOkka:
# Issue

Valhalla missed conditional TR mapped according to previous tagging schema [#441](https://github.com/mapbox/customer-evaluation-issues/issues/441)

## Tasklist

 - [x] Add tests


test/gurka/test_time_dependent_restrictions.cc
```diff
@@ -0,0 +1,56 @@
...
+                          {{"mjolnir.timezone", VALHALLA_BUILD_DIR "test/data/tz.sqlite"}});
+  }
+};
+
+gurka::map TimeDependentRestrictions::map = {};
+
+TEST_F(TimeDependentRestrictions, ActiveHourRestriction) {
+  auto result =
+      gurka::do_action(valhalla::Options::route, map, {"A", "F"}, "auto",
+                       {{"/date_time/type", "1"}, {"/date_time/value", "2021-04-02T14:00"}});
```
mandeepsandhu, line 47: Can we test it for other `date_time` types (0,2 & 3)? These types invoke different routing algorithms so it'll be good to test the timed TRs here in case it breaks in one algo and not another.

You can use [parametrized tests](https://github.com/google/googletest/blob/master/docs/advanced.md#value-parameterized-tests) to pass each of the `date_time` types.

test/timeparsing.cc
```diff
@@ -401,6 +401,29 @@ TEST(TimeParsing, TestConditionalRestrictions) {
       TryConditionalRestrictions(conditions.at(x), 0, 1, 64, 3, 3, 0, 15, 0, 12, 1, 5, 17, 0);
     }
   }
+
```
mandeepsandhu, line 404: Do we want to test wrongly formatted restriction...eg: end day is set without a corresponding start day (`-Friday` or some such)?

test/gurka/test_time_dependent_restrictions.cc
danpat: If any of them are failing, they can be flagged with a `DISABLED_` prefix - at least then we'll get notice that they're not fixed yet, and we could add a CI step to try non-fatally running disabled tests :

https://google.github.io/googletest/advanced.html#temporarily-disabling-tests

test/timeparsing.cc
CuriOusQuOkka: As for me it sounds as incorrect input. In case only day_on is specified, '-' is not added, day_on can be a sequence of days separated by a comma or a range with '-' inside. When only day_off is specified it is harder to understand what it means: working hours are from Monday/Sunday to X, or each day except X....?

mandeepsandhu: You're right, in that the input is incorrect. My point was that we should test that such inputs are ignored (looking at the code, we only set the condition when day_start is set, so just having a day_end should not enable the TR). So adding a test with incorrect input and showing that it does not take effect would be sufficient. Does that make sense?

CuriOusQuOkka: Yeah, added regular and gurka tests.

test/gurka/test_time_dependent_restrictions.cc
```diff
@@ -0,0 +1,60 @@
...
+                                          {"hour_on", "07:00"},
+                                          {"hour_off", "19:00"}}}};
+    map =
+        gurka::buildtiles(layout, ways, {}, relations, "test/data/gurka_time_dependent_restrictions",
+                          {{"mjolnir.timezone", VALHALLA_BUILD_DIR "test/data/tz.sqlite"}});
+  }
+};
+gurka::map TimeDependentRestrictions::map = {};
+
+// TODO: Does not work for date_time_type=2, returns additional AB edge.
```
genadz, line 42: it's because of this https://github.com/valhalla/valhalla/issues/2940 ; try to put `start` and `end` locations on the edges
```
          D----B---1-A
          |    |
          |    |
          E----C---2-F
```

genadz (changes requested):

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1827,14 +1830,20 @@ struct graph_callback : public OSMPBF::Callback {
...
-            if (condition.empty()) {
-              condition = day_start + "-";
-              condition += day_end;
+            if (condition.empty() && !opening_hours.empty()) {
+              condition = opening_hours;
+            } else if (condition.empty()) {
+              if (!day_start.empty() && !day_end.empty()) {
+                condition = day_start + '-' + day_end;
+              } else if (!day_start.empty()) {
+                condition = day_start;
```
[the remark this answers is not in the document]
gknisely: @genadz we can't only have just a day_start.  if the restriction is only one day the day_start would be equal to the day_end.

test/timeparsing.cc
gknisely: @CuriOusQuOkka @mandeepsandhu if we have only the start day set then these restrictions should be tossed as the user entered the info incorrectly

test/gurka/test_time_dependent_restrictions.cc
mandeepsandhu: @genadz since `date_time: 2` uses timedep-reverse, I wonder if the extra edge is being returned due to the bug @danpat found [here](https://github.com/valhalla/valhalla/pull/2987#issuecomment-824448011).

@CuriOusQuOkka https://github.com/valhalla/valhalla/pull/2987 fixed a bug in timedep-reverse, which might fix the problem you're seeing. If so, you can re-enable `date_time: 2`.

src/mjolnir/pbfgraphparser.cc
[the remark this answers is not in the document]
mandeepsandhu: @gknisely from the code above it looks like just having `day_start` is also a valid option. If both start & end are mandatory, then maybe we should just have `if (!day_start.empty() && !day_end.empty()) { ...}` ?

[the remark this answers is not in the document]
gknisely: @mandeepsandhu correct.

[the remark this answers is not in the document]
CuriOusQuOkka: We support start date that stays alone. It can be a sequence (like Monday,Wednesday,Thursday), a range (Monday-Friday), or just one day.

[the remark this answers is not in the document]
CuriOusQuOkka: read the [comment](https://github.com/valhalla/valhalla/pull/3029/files#r620595224) form @gknisely. Ignore this case.

test/gurka/test_time_dependent_restrictions.cc
CuriOusQuOkka: Thanks @mandeepsandhu, it fixed the bug.

genadz: > @genadz since date_time: 2 uses timedep-reverse, I wonder if the extra edge is being returned due to the bug @danpat found here.

cool! so, I should close this issue #2940

test/gurka/test_time_dependent_tags.cc
```diff
@@ -0,0 +1,115 @@
...
+    if (x == 0) { //  Restriction is on.
+      gurka::assert::raw::expect_path(result, {"AB", "BD", "DE", "EC", "CF"});
+    } else { //  Restriction is off.
+      gurka::assert::raw::expect_path(result, {"AB", "BC", "CF"});
+    }
+  }
+}
+
+TEST_F(TimeDependentTags, IgnoreDayOffOnlyRestriction) {
+  tags day_off = {
```
genadz, line 98: if I understood correctly having only `day_on` or `hour_off/hour_on` also should be ignored. could you add all these cases ?

gknisely:
@CuriOusQuOkka I think we will be good after we test with https://github.com/valhalla/valhalla/blob/master/test_requests/timed_access_routes.txt and https://github.com/valhalla/valhalla/blob/master/test_requests/timed_conditional_routes.txt  Let me know if there are any diffs.

test/gurka/test_time_dependent_restrictions.cc
```diff
@@ -0,0 +1,58 @@
...
+  gurka::assert::raw::expect_path(result, {"AB", "BD", "DE", "EC", "CF"});
+}
+
+TEST_P(TimeDependentRestrictions, NoRestrictions) {
+  auto result = gurka::do_action(valhalla::Options::route, map, {"A", "F"}, "auto",
+                                 {{"/date_time/type", std::to_string(GetParam())},
+                                  {"/date_time/value", "2021-04-02T23:00"}});
+  gurka::assert::raw::expect_path(result, {"AB", "BC", "CF"});
+}
+
```
mandeepsandhu, line 55: For completeness, should we also add a testcase with date_time before start of restriction?

CuriOusQuOkka:
> @CuriOusQuOkka I think we will be good after we test with https://github.com/valhalla/valhalla/blob/master/test_requests/timed_access_routes.txt and https://github.com/valhalla/valhalla/blob/master/test_requests/timed_conditional_routes.txt Let me know if there are any diffs.

Rad tests have passed

genadz (approved):

PR #3031, opened 2021-04-24
[DO NOT MERGE. TESTING] Skip ci test

mandeepsandhu:
# Issue

What issue is this PR targeting? If there is no issue that addresses the problem, please open a corresponding issue and link it here.

## Tasklist

 - [ ] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Update the docs with any new request parameters or changes to behavior described
 - [ ] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #3014, opened 2021-04-17
Ignore irrelevant file changes on azure

nilsnolde:
azure is again pretty curious about design choices, e.g. wildcarding file patterns doesn't work..

also added the other CI configs to circle's `ciignore` list.

PR #3032, opened 2021-04-25
Make CI skip check more robust

mandeepsandhu:
Implements the 2 TODOs which make the CI check robust to erroneous
skips.

The script will now:
- not skip CI if previous CI was not successful
- will check all files since the last successful CI run for determining
  if the commits can skip CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,7 @@ commands:
   install_macos_dependencies:
     steps:
       - run: brew install protobuf cmake ccache libtool boost libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
+      - run: pip3 install requests
 
 jobs:
   build-docker-images:
@@ -123,6 +124,7 @@ jobs:
     executor: macos
     steps:
       - checkout
+      - install_macos_dependencies
       - run: |
             if ! ./scripts/needs_ci_run; then
                 echo "Changes in last commit do not need CI. Skipping step"
@@ -130,7 +132,6 @@ jobs:
             fi
       - run: ./scripts/format.sh && ./scripts/error_on_dirty.sh
       - run: git submodule sync && git submodule update --init
-      - install_macos_dependencies
       - restore_cache:
           keys:
             - ccache-release-macos-{{ .Branch }}
--- a/scripts/needs_ci_run
+++ b/scripts/needs_ci_run
@@ -75,7 +75,7 @@ def last_build_successful():
 
     if sorted(job_status.keys()) != sorted(JOBS):
         print(f'Unexpected job name fetched ({sorted(job_status)}). Expected {sorted(JOBS)}')
-        return False
+        return False, None
 
     # Return False if any of the jobs were not in 'success' state
     for status, gitsha in job_status.values():
--- a/scripts/needs_ci_run
+++ b/scripts/needs_ci_run
@@ -7,7 +7,6 @@ import requests
 
 from os import path, environ
 
-token = '8193cb473902c29f1a360547e9d4126b4df302ea'
 
 # jobs defined in .circleci/config.yml
 JOBS = ['build-docker-images',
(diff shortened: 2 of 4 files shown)


scripts/needs_ci_run
```diff
@@ -1,22 +1,105 @@
...
 import fnmatch
-import os.path
 import sys
 import subprocess
+import requests
+
+
+from os import path, environ
+
+token = '8193cb473902c29f1a360547e9d4126b4df302ea'
```
mandeepsandhu, line 10: Oos, I forgot to remove this! I'll do it in the next push

.circleci/config.yml
```diff
@@ -123,14 +124,14 @@ jobs:
     executor: macos
     steps:
       - checkout
+      - install_macos_dependencies
```
mandeepsandhu, line 127: Moved this back to its original place, since I needed to install python `requests` package and thought it'd be better to keep all installs at 1 place. This does add ~2+ mins before the skip CI script can run. But I guess that shouldn't be a big deal.

mandeepsandhu:
@kevinkreiser I think all the drawbacks of the previous attempt at determining if a change can skip CI have been addressed. Let me know if there are issues you see with this change.

CC: @PureW @nilsnolde

mandeepsandhu:
All tests were run on this branch: https://github.com/valhalla/valhalla/pull/3031

scripts/needs_ci_run
```diff
@@ -1,41 +1,126 @@
...
+        print('CIRCLE_WORKFLOW_ID env not set! Unable to do CI skip check')
+        sys.exit(0)
+    if not token:
+        print('CIRCLE_API_TOKEN env not set! Unable to do CI skip check')
+        sys.exit(0)
+
+    # Get twice the number of jobs configured. There would be len(JOBS) from this
+    # run, while the rest will be from the last run
+    cci_api_url = \
+        f'https://circleci.com/api/v1.1/project/github/valhalla/valhalla/tree/{branch}?circle-token={token}&shallow=true&limit={len(JOBS)*2}'
```
mandeepsandhu, line 53: [circle API doc](https://circleci.com/docs/api/#recent-builds-for-a-single-project)

nilsnolde:
What do you think integrating the „skip for draft PR“ for circle CI here? Ref https://github.com/valhalla/valhalla/pull/3014#issuecomment-821970848

mandeepsandhu:
> What do you think integrating the „skip for draft PR“ for circle CI here? Ref [#3014 (comment)](https://github.com/valhalla/valhalla/pull/3014#issuecomment-821970848)

Definitely a good idea to consider, although I'm not sure if everyone would want to skip CI for draft PRs (eg I've relied on CI for testing some of my changes on macos). I'm thinking this should be something that a user explicitly asks for. Which reminds me of the circle CIs `[skip ci]` tag that will cause it to not build such commits. Maybe users can use that instead?

nilsnolde:
It's a pretty subjective topic I guess. As for me I really like not to waste so many CI hours and I'm really bad at putting `[ci skip]` ;) I usually don't need/want CI before final review and if I do on rare occasions I find it ok to un-draft, do a small commit and let CI run, then draft again. It'd be best if CI ran automatically when I un-draft without committing again, but we only get that on platforms who support this draft PR stuff natively.

But I'm just a tiny speck here;) maybe someone else has an opinion on it.

nilsnolde (approved):

kevinkreiser:
lgtm

kevinkreiser (approved):

merkispavel (dismissed):
re-approve

mandeepsandhu:
> re-approve

Thanks. I've put the fix for osx in https://github.com/valhalla/valhalla/pull/3037. I'll update this PR once thats merged.

merkispavel (approved):
re-approve

PR #3038, opened 2021-04-26
Chunnel Routing

kevinkreiser:
Every now and then the Chunnel stops routing properly...

Recently (2 months ago) I made a planet build and the chunnel isn't the suggested route when going from the UK to France even if you put the origin and destination right at the chunnel entry and exit. Instead the route goes way south first and crosses on a ferry to France and then drives up north again to get back to the chunnel exit. 

@rzyc and I debugged this for a while and found out a couple of things. First we found that if we took an extract of the planet that has only the parts of the uk and france with the chunnel connections in them the route still fails first and then succeeds on the second pass in routing_action.cc

    osmium extract -f pbf -o chunnel.pbf planet-latest.osm.pbf -b 1.0787,50.89,1.9803,51.1277

then we started commenting out the 3 different relaxation types in the second pass. first the not_through and the destination_only stuff. the route still succeeded on the second pass. then we commented out the hierarchy relaxation and it failed the route.

this made me think that it had to do with reclassification. but ive added a unit test here that proves the reclassification does work for the chunnels tagging (route = shuttle_train).

after looking at the ferry reclassification for a little bit it seems that indeed the chunnels surrounding edges are not getting reclassified. what happen is that this function https://github.com/valhalla/valhalla/blob/master/src/mjolnir/ferry_connections.cc#L209 seems to think the chunnel is too short to be worth reclassifying for. if i disable this, routing works as expected (although the reclassified path it takes is kind of crappy (has a uturn)).

@dnesbitt61 what was the purpose of not reclassifying near short ferries? were we assuming there must be alternate ways around without taking the ferries themselves? it looks to me like maybe the chunnel is connected to a bunch of little edges in the queueing area on the british side and that might be triggering the shortferry logic to stop us from connecting up the chunnel.

@gknisely have you been suffering with the chunnel at all lately?


gknisely:
@kevinkreiser checking on a status for you.

SvetlanaBayarovich:
@kevinkreiser just curious about weights. If I use intermediate point to force ferry or chunnel I get chunnel route to be like 25% more expensive than the ferry one. Is it different for you? I'm checking regarding ferry connections reclassification now, but even if it's the case, difference in weight will still prevent us from adequate routing

SvetlanaBayarovich:
@kevinkreiser So I think there IS a problem with ferry connections. GB side pictures: current connection(s) with chunnel edges omitted, entering and exiting the chunnel. The ‘exiting’, is the only one that follows level 0 (red) edges. ‘Entering’ is a trash (and causes “search exhausted” stuff). To me it looks like the connection is built correctly only in one direction and not in reverse. Something similar happens on the France end as well.

Edges by level:
image
Entering the chunnel:
image
Exiting the chunnel:
image

SvetlanaBayarovich:
Also the France side:
Edges by level:
image
Entering:
image
Exiting:
image

kevinkreiser:
@SvetlanaBayarovich yep that is what we saw. we are pretty sure about the issue and actually @rzyc made a unit test to show why this is happening and the fix. i'll let him explain it, he should be pushing to this branch shortly

rzyc:
@SvetlanaBayarovich The issue was in the `ShortFerry` function in `ferry_connections.cc` as we originally suspected. According to the function's docstring, it is supposed to exclude ferries that are "just one edge and length < 2 km". However, the "just one edge" check was *not* actually performed and only the 2km check was. The Chunnel's entrance area consists of multiple short edges connected together, some of which are definitely < 2km. Therefore, `ShortFerry` thinks the Chunnel entrance edge is a short ferry and doesn't perform the reclassification around the entrance.

In this PR, I added the "just one edge" check by looping over all other connected edges (at both source and destination nodes) and seeing if any of them were also ferry edges. The function returns `false` immediately if it finds a connecting ferry edge. I also added a new unit test that checks for this issue, and made sure that this unit test was failing if we temporarily removed the "just one edge" check.

src/mjolnir/ferry_connections.cc
```diff
@@ -225,13 +225,27 @@ bool ShortFerry(const uint32_t node_index,
...
+      for (const auto& edge2 : bundle.node_edges) {
+        if (edge2.first.llindex_ != edge.first.llindex_ && edge2.first.attributes.driveable_ferry) {
+          return false;
+        }
+      }
+      for (const auto& edge2 : bundle2.node_edges) {
+        if (edge2.first.llindex_ != edge.first.llindex_ && edge2.first.attributes.driveable_ferry) {
+          return false;
+        }
+      }
```
rzyc, line 246: The "just one edge" check. Looping over all other connected edges at both source and destination nodes. Sees if any of them were also ferry edges, with a check on the latlong index to make sure we're not looking at the original ferry edge. Return `false` immediately if we find one, since we know that means this ferry consists of more than one edge.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Release Date: 2021-??-?? Valhalla 3.1.2
 * **Removed**
 * **Bug Fix**
+   * FIXED: Fix reclassification for "shorter" ferries and rail ferries (for Chunnel routing issues) [#3038](https://github.com/valhalla/valhalla/pull/3038)
```
kevinkreiser, line 4: can you move this to the bottom of the bug fix list? for whatever reason we've been adding them to the bottom of each of the respective lists so that the oldest is at the top. i guess so that when you read a releases change log its in chronological order from oldest to newest changes.

test/gurka/test_ferry_connections.cc
```diff
@@ -3,12 +3,13 @@
...
 using namespace valhalla;
 
-baldr::RoadClass
+std::pair<baldr::RoadClass, baldr::RoadClass>
 ReclassifyFerryConnectionEdge(std::map<std::string, std::string> const& way_description) {
   constexpr double gridsize_metres = 1000;
 
   const std::string ascii_map = R"(
           A--B--b--C----D--E
+          F--G--g--H----I--J
```
rzyc, line 12: Adding a parallel test case ensuring that the behavior for rail ferries is the same as for ferries. The only difference in these two test cases is that `HI` is a rail ferry and `CD` is a ferry.

test/gurka/test_ferry_connections.cc
```diff
@@ -26,15 +27,69 @@ ReclassifyFerryConnectionEdge(std::map<std::string, std::string> const& way_desc
...
+  auto Gg_edge = std::get<1>(gurka::findEdgeByNodes(graph_reader, layout, "G", "g"));
+  return {Bb_edge->classification(), Gg_edge->classification()};
+}
+
+TEST(Standalone, ShortFerry) {
+  const std::string ascii_map = R"(
+          A--B--C--F--D--E--G
+                   |
+                   N
+    )";
```
rzyc, line 59: Adding a test case here to cover the issue discovered. Both `CF` and `FD` are ferry edges, but they are both < 2km. This means that `ShortFerry` should return `false` since the ferry consists of > 1 edge. Thus, we should be reclassifying `BC` and `DE` and we check that in this unit test.

Also made sure that this unit test was failing if we temporarily removed this PR's changes to `ShortFerry`.

kevinkreiser: yeah the intuition here is we shouldnt assume the ferry is short just because it has 1 short edge in it. seems like somewhere along the chunnel another drivable edge intersects it and so makes the first edge short but then of course the whole rest of the chunnel is many km long. basically shortFerry was tricked. it still might be better to completely disable shortFerry, you can think of malicious cases where a short single edge ferry is the only way to get from one place to another but it likely doesnt happen in the wild that often (or maybe ever)

src/mjolnir/ferry_connections.cc
```diff
@@ -225,13 +225,27 @@ bool ShortFerry(const uint32_t node_index,
...
+
+      // If the end node has a non-ferry edge check the length of the edge
       if (bundle2.node.non_ferry_edge_) {
```
rzyc, line 249: The code previously had this if statement here as the "just one edge" check, but it wasn't sufficient. If we consider this case (used in the unit test described below):
```
          A--B--C--F--D--E--G
                   |
                   N
```
where `CF` and `FD` are ferry edges and `FN` is *not* a ferry edge. Suppose that `ShortFerry` is analyzing `CF` and `bundle2` is created at `F`. Then `bundle2.node.non_ferry_edge_` will be `true` since `FN` exists.

CHANGELOG.md
rzyc: Done

src/mjolnir/ferry_connections.cc
SvetlanaBayarovich: Haha, guess it was working in most cases because not typical for a 'traditional'  ferry to have a non-ferry edge somewhere in the middle

test/gurka/test_ferry_connections.cc
SvetlanaBayarovich: You may never know when it appears)

SvetlanaBayarovich:
@rzyc Wow that was a tricky one! And such an interesting intersection of different circumstances.

SvetlanaBayarovich (approved):
Looks reasonable to me

kevinkreiser:
We're going to do a round of RAD before we merge this to make sure no other unexpected changes occur.

rzyc:
Did a regression test with @kevinkreiser to check differences in routing before and after this PR. We did not notice any particularly drastic routing differences; most differences seem to be one-offs with just a couple slightly differing instructions ending up with similar cost values.

One interesting route change involving a ferry was this:

`master` branch:
image
After this fix:
image

Both routes have similar costs, but we noticed that after this fix with ferry reclassification, it elected to get on the ferry earlier in the route.

Since there aren't any drastic differences, we are looking to merge this PR on Monday.

gknisely (approved):

kevinkreiser:
@mandeep, looks like macos builds "just work" again, im going to maek them required again

mandeepsandhu:
> @mandeep, looks like macos builds "just work" again, im going to maek them required again

@kevinkreiser yeah. I got an email notification from circle saying they are updating all macos executors to use newer homebrew version. So looks like that did the trick. Glad its working without needing any change from us.

PR #3039, opened 2021-04-27
Evaluate restrictions with dow only

CuriOusQuOkka:
# Issue

Noticed, that restrictions with dow only, like Monday-Friday were not applied after routing to the edges. They are not applied to the edges when validating the routing path. 

In the example below, if I set conditional restriction `no_entry @ Mo-Fr` for `AB->BC` left turn
the restriction is ignored,  at any time we receive `AB->BC->CF` path from A to F instead of  `AB->BD->DE->EC->CF` for workdays.

          D----B----A
          |    |
          |    |
          E----C----F

## Tasklist

 - [x] Add tests
 - [x] Add gurka tests with day_on/day_off and no hour info [#3029](https://github.com/valhalla/valhalla/pull/3029)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@
    * FIXED: Odin undefined behaviour: handle case when xedgeuse is not initialized [#3020](https://github.com/valhalla/valhalla/pull/3020)
    * FIXED: Isochrone (::Generalize()) fix to avoid generating self-intersecting polygons [#3026](https://github.com/valhalla/valhalla/pull/3026)
    * FIXED: Handle day_on/day_off/hour_on/hour_off restrictions [#3029](https://github.com/valhalla/valhalla/pull/3029)
+   * FIXED: Apply conditional restrictions with dow only to the edges when routing [#3039](https://github.com/valhalla/valhalla/pull/3039)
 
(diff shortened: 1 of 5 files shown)


gknisely:
@CuriOusQuOkka I think we will be good after we test with https://github.com/valhalla/valhalla/blob/master/test_requests/timed_access_routes.txt and https://github.com/valhalla/valhalla/blob/master/test_requests/timed_conditional_routes.txt  Let me know if there are any diffs.

gknisely (dismissed):

CuriOusQuOkka:
> @CuriOusQuOkka I think we will be good after we test with https://github.com/valhalla/valhalla/blob/master/test_requests/timed_access_routes.txt and https://github.com/valhalla/valhalla/blob/master/test_requests/timed_conditional_routes.txt Let me know if there are any diffs.

the tests passed.

genadz:
what about gurka test ? I expected to see the example you described in the ticket in gurka tests

CuriOusQuOkka:
> what about gurka test ? I expected to see the example you described in the ticket in gurka tests

@gknisely Added.

gknisely (approved):

genadz (approved):

PR #3041, opened 2021-04-28
Fixed incorrect routing through `motor_vehicle:conditional=destination`.

maksimandrianov:
# Issue

Valhalla incorrect works when acts restriction by time limit for vehicles (period specified in the tag):

build route to roads which can only be visited by driving through the such roads (road with tag `motor_vehicle:conditional`=destination + other road = the only way)
on whole distance of road with tag `motor_vehicle:conditional`=destination. Works only on some parts of such roads.
When restriction by time limit doesn't act - routing correct.

What did you expect to happen?
This restriction (in the specified time period) should to avoid such roads from transit routing, but as destination and the only way they should be suggest.

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
    * FIXED: Handle day_on/day_off/hour_on/hour_off restrictions [#3029](https://github.com/valhalla/valhalla/pull/3029)
    * FIXED: Apply conditional restrictions with dow only to the edges when routing [#3039](https://github.com/valhalla/valhalla/pull/3039)
    * FIXED: Missing locking in incident handler needed to hang out to scop lock rather than let the temporary disolve [#3046](https://github.com/valhalla/valhalla/pull/3046)
+   * FIXED: Continuous lane guidance fix [#3054](https://github.com/valhalla/valhalla/pull/3054)
 
 * **Enhancement**
    * CHANGED: Refactor timedep forward/reverse to reduce code repetition [#2987](https://github.com/valhalla/valhalla/pull/2987)
--- a/src/odin/enhancedtrippath.cc
+++ b/src/odin/enhancedtrippath.cc
@@ -624,21 +624,23 @@ EnhancedTripLeg_Edge::ActivateTurnLanes(uint16_t turn_lane_direction,
     // and next maneuver is not a straight
     // Activate only specific matching turn lanes
     switch (next_maneuver_type) {
-      case DirectionsLeg_Maneuver_Type_kUturnLeft:
-      case DirectionsLeg_Maneuver_Type_kSharpLeft:
-      case DirectionsLeg_Maneuver_Type_kLeft:
       case DirectionsLeg_Maneuver_Type_kSlightLeft:
-      case DirectionsLeg_Maneuver_Type_kExitLeft:
+      case DirectionsLeg_Maneuver_Type_kLeft:
+      case DirectionsLeg_Maneuver_Type_kSharpLeft:
+      case DirectionsLeg_Maneuver_Type_kUturnLeft:
       case DirectionsLeg_Maneuver_Type_kRampLeft:
+      case DirectionsLeg_Maneuver_Type_kExitLeft:
+      case DirectionsLeg_Maneuver_Type_kStayLeft:
       case DirectionsLeg_Maneuver_Type_kDestinationLeft:
       case DirectionsLeg_Maneuver_Type_kMergeLeft:
         return ActivateTurnLanesFromLeft(turn_lane_direction, curr_maneuver_type, 1);
       case DirectionsLeg_Maneuver_Type_kSlightRight:
-      case DirectionsLeg_Maneuver_Type_kExitRight:
-      case DirectionsLeg_Maneuver_Type_kRampRight:
       case DirectionsLeg_Maneuver_Type_kRight:
       case DirectionsLeg_Maneuver_Type_kSharpRight:
       case DirectionsLeg_Maneuver_Type_kUturnRight:
+      case DirectionsLeg_Maneuver_Type_kRampRight:
+      case DirectionsLeg_Maneuver_Type_kExitRight:
+      case DirectionsLeg_Maneuver_Type_kStayRight:
       case DirectionsLeg_Maneuver_Type_kDestinationRight:
       case DirectionsLeg_Maneuver_Type_kMergeRight:
         return ActivateTurnLanesFromRight(turn_lane_direction, curr_maneuver_type, 1);
--- a/src/odin/maneuver.cc
+++ b/src/odin/maneuver.cc
@@ -170,23 +170,25 @@ bool Maneuver::IsMergeType() const {
 }
 
 bool Maneuver::IsLeftType() const {
-  return ((type_ == DirectionsLeg_Maneuver_Type_kUturnLeft ||
-           type_ == DirectionsLeg_Maneuver_Type_kSharpLeft ||
+  return ((type_ == DirectionsLeg_Maneuver_Type_kSlightLeft ||
            type_ == DirectionsLeg_Maneuver_Type_kLeft ||
-           type_ == DirectionsLeg_Maneuver_Type_kSlightLeft ||
-           type_ == DirectionsLeg_Maneuver_Type_kExitLeft ||
+           type_ == DirectionsLeg_Maneuver_Type_kSharpLeft ||
+           type_ == DirectionsLeg_Maneuver_Type_kUturnLeft ||
            type_ == DirectionsLeg_Maneuver_Type_kRampLeft ||
+           type_ == DirectionsLeg_Maneuver_Type_kExitLeft ||
+           type_ == DirectionsLeg_Maneuver_Type_kStayLeft ||
            type_ == DirectionsLeg_Maneuver_Type_kDestinationLeft ||
            type_ == DirectionsLeg_Maneuver_Type_kMergeLeft));
 }
 
 bool Maneuver::IsRightType() const {
   return ((type_ == DirectionsLeg_Maneuver_Type_kSlightRight ||
-           type_ == DirectionsLeg_Maneuver_Type_kExitRight ||
-           type_ == DirectionsLeg_Maneuver_Type_kRampRight ||
            type_ == DirectionsLeg_Maneuver_Type_kRight ||
            type_ == DirectionsLeg_Maneuver_Type_kSharpRight ||
            type_ == DirectionsLeg_Maneuver_Type_kUturnRight ||
+           type_ == DirectionsLeg_Maneuver_Type_kRampRight ||
+           type_ == DirectionsLeg_Maneuver_Type_kExitRight ||
+           type_ == DirectionsLeg_Maneuver_Type_kStayRight ||
            type_ == DirectionsLeg_Maneuver_Type_kDestinationRight ||
            type_ == DirectionsLeg_Maneuver_Type_kMergeRight));
 }
--- a/test/gurka/test_turn_lanes.cc
+++ b/test/gurka/test_turn_lanes.cc
@@ -298,7 +298,7 @@ TEST(Standalone, TurnLanesMultiLaneShort) {
                               });
 
   // makes the data left side driving.
-  map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_turn_lanes_4",
+  map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_turn_lanes_5",
                           {{"mjolnir.data_processing.use_admin_db", "true"}});
 
   // A -> F - takes leftmost right lane (left side driving)
@@ -696,7 +696,7 @@ TEST(Standalone, TurnLanesSerializedResponse) {
   };
 
   const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
-  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_turn_lanes_5");
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_turn_lanes_10");
 
   auto verify_lane_schema = [](const rapidjson::Value& lanes) -> void {
     // Verify common lane attributes:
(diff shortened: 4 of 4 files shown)


src/thor/unidirectional_astar.cc
```diff
@@ -210,8 +210,10 @@ inline bool UnidirectionalAStar<expansion_direction, FORWARD>::ExpandInner(
   // Skip shortcut edges for time dependent routes, if no access is allowed to this edge
   // (based on costing method)
   uint8_t restriction_idx = kInvalidRestriction;
+  bool is_dest =
+      destinations_percent_along_.find(meta.edge_id) != std::cend(destinations_percent_along_);
```
kevinkreiser, line 214: i assume this was the main case that you were worried about when you implemented this? time dependent unidirectional routing? does constantly hitting this unordered_map cost anything in performance? also why use `std::cend`, please just use `destinations_percent_along_.cend()`

src/thor/route_action.cc
```diff
@@ -399,6 +399,7 @@ std::vector<std::vector<thor::PathInfo>> thor_worker_t::get_path(PathAlgorithm*
     float expansion_within_factor = path_algorithm == &bidir_astar ? 2.f : 4.f;
     cost->RelaxHierarchyLimits(relax_factor, expansion_within_factor);
     cost->set_allow_destination_only(true);
+    cost->set_allow_conditional_destination_restrictions(true /* allow */);
```
kevinkreiser, line 402: i appreciate that the comment tells the person a little but more about what is going on but we dont follow this convention anywhere else in the code so i dont think we should start at this point

valhalla/sif/dynamiccost.h
```diff
@@ -171,6 +172,7 @@ class DynamicCost {
    * @return Returns true if access is allowed, false if not.
    */
   virtual bool Allowed(const baldr::DirectedEdge* edge,
+                       const bool is_dest,
```
kevinkreiser, line 175: why not add this as the last argument in the method?

gknisely (changes requested):
+1 on @kevinkreiser 's comments.  Also, I would like to see some real world before and after examples.

valhalla/sif/dynamiccost.h
maksimandrianov: This is going after `edge` because it is the property of this edge. It seems more readable if they are near each other. I think, it would be a little strange if I put it after timezone parameter.

src/thor/route_action.cc
maksimandrianov: I esteem valhalla code convention and I will fix it
but what does valhalla code convention say about methods, classes, structs, variables and enums naming? Why are there so different? :)

src/thor/unidirectional_astar.cc
maksimandrianov: Was important to support the same behaviour for all algorithms. There are some changes in recost_forward function for bidirectional routing.
I will fix std::cend -> .cend().

src/thor/route_action.cc
gknisely: @dgearhart don't we have a code conventions doc?

kevinkreiser: we do but it recommends googles style guidelines and they are garbage :stuck_out_tongue:

SvetlanaBayarovich:
Does it make sense to  add any RAD examples to, say,  `valhalla/test_requests/timed_conditional_routes.txt`

test/matrix.cc
```diff
@@ -37,6 +37,7 @@ class SimpleCost final : public DynamicCost {
   }
 
   bool Allowed(const DirectedEdge* edge,
+               const bool /*is_dest*/,
```
genadz, line 40: in some cases you skip `/*is_dest*/` but in some you add it. is it on purpose ?

gknisely:
> Does it make sense to add any RAD examples to, say, `valhalla/test_requests/timed_conditional_routes.txt`?

Yes.  +1

genadz:
@maksimandrianov @kevinkreiser @gknisely @merkispavel

Imho, PR looks good to me except for one thing: the logic is different from ordinar `destination_only` edges processing.
In order to use the same approach , we could use spare bits https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/directededge.h#L1124 https://github.com/valhalla/valhalla/blob/master/valhalla/sif/edgelabel.h#L446 in order to define new flag `conditional_dest_only_` . It will allow to implement the same logic as for `destination_only`

test/matrix.cc
maksimandrianov: I saw it in code below https://github.com/valhalla/valhalla/blob/0b0300b97096aec7b519d6b32dbb3fee0734ce95/test/matrix.cc#L45 and did the same. Where do I skip? This comments need for readable code, also it suppresses "unused variable" warning.  If a var be used, we can uncomment a var name. It is convenient.

genadz: I see, you followed local code style. ok

valhalla/sif/dynamiccost.h
```diff
@@ -828,6 +840,8 @@ class DynamicCost {
   // and bicycle generally allow access (with small penalties).
   bool allow_destination_only_;
 
+  bool allow_conditional_destination_restrictions_;
```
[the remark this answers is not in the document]
maksimandrianov: I did that for fine control, now we can disallow one of them separately. I can replace if we want to mix them together.

src/thor/astar_bss.cc
```diff
@@ -149,8 +149,10 @@ void AStarBSSAlgorithm::ExpandForward(GraphReader& graphreader,
     // directed edge), if no access is allowed to this edge (based on costing method),
     // or if a complex restriction exists.
     uint8_t has_time_restrictions = -1;
+    bool is_dest = destinations_.find(edgeid) != destinations_.cend();
```
stari4ek, line 152: what's perf penalty with additional map lookups added (here and the rest)?

valhalla/sif/dynamiccost.h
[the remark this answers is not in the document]
genadz: I can't come up with a case when we need them separate

maksimandrianov:
> @maksimandrianov @kevinkreiser @gknisely @merkispavel
> 
> Imho, PR looks good to me except for one thing: the logic is different from ordinar `destination_only` edges processing.
> In order to use the same approach , we could use spare bits https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/directededge.h#L1124 https://github.com/valhalla/valhalla/blob/master/valhalla/sif/edgelabel.h#L446 in order to define new flag `conditional_dest_only_` . It will allow to implement the same logic as for `destination_only`

I think that are a little different entities. In first case `dest_only` is a road property, it says that road is private, in this PR `dest_only` is a restriction type. Restriction info is stored separately and if we add `conditional_dest_only_ flag` to each edge there is data duplication. Because we already have information about restrictions. I think we can use this bytes for more useful things.

maksimandrianov:
I have tested performance using 2 ways:

maksimandrianov:
@kevinkreiser, @stari4ek I have tested performance using: 
* `time ./run_with_server.py --test-file ../test_requests/de_benchmark_routes.txt --url http://localhost:8002/route --concurrency 1`
* `valhalla-bench`

Both of them showed that performance is not changed.

src/thor/astar_bss.cc
maksimandrianov: I answered about performance below, I have tested using other tools. 
Hashtable lookups is O(1). 
You can think that next operations will be performed:
```
bucket = edgeid % hashtable_size
list = array[bucket]
return list.find(edgeid)
```
Size of lists -> 1 in good hastables.

valhalla/sif/dynamiccost.h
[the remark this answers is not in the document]
maksimandrianov: For example, we can to relax gradually. A path can be found after relaxing by one parameter and no need allow others.
But as you say, done

src/sif/autocost.cc
```diff
@@ -443,7 +446,7 @@ bool AutoCost::AllowedReverse(const baldr::DirectedEdge* edge,
     return false;
   }
 
-  return DynamicCost::EvaluateRestrictions(access_mask_, edge, tile, opp_edgeid, current_time,
+  return DynamicCost::EvaluateRestrictions(access_mask_, edge, false, tile, opp_edgeid, current_time,
```
SvetlanaBayarovich, line 449: Why is `is_dest` always false for `AllowedReverse` check?

test/gurka/test_conditional_restrictions.cc
```diff
@@ -174,3 +179,60 @@ TEST_F(ConditionalRestrictions, RestrictionPedestrian) {
       },
       std::exception);
 }
+
+TEST_F(ConditionalRestrictions, DestinationRestrictionOnLastEdgeIsValid) {
+  for (auto const& date_time_type : kDateTimeTypes) {
+    auto result = gurka::do_action(valhalla::Options::route, map, {"F", "I"}, "auto",
```
SvetlanaBayarovich, line 185: Checks for other costings (motorcycle, tuck, etc) as long as it is implemented for them?

src/sif/autocost.cc
SvetlanaBayarovich: What I mean is, if the restriction is near the end and the edges are (likely) processed with `AllowedReverse` during reverse flow, when/how will it be checked? Like we are going from A to K:
```
A---B---C---D---E---F---G---H---I---G---K
                                 ^^^
                             restr here
---->                                <---
fwd                               reverse
```

SvetlanaBayarovich (changes requested):
Couple of questions

gknisely:
@maksimandrianov can you show some examples that are fixed in OSM.  Just a few before and after screen shots showing the different routes.  Also add those tests to `valhalla/test_requests/timed_conditional_routes.txt`

src/sif/autocost.cc
maksimandrianov: `AllowedReverse` does not check last edge -> `is_dest` always false for `AllowedReverse `.
Check please an update in this PR for `recost_forward` function 
https://github.com/valhalla/valhalla/blob/9c14cf25103bd9ac36df2730bb1a5a842324eb56/src/sif/recost.cc#L97

src/thor/astar_bss.cc
kevinkreiser: the lookup is amortized constant time yes but that doesnt mean its free. it costs clock cycles and its in the tight loop of the algorithm. how did you validate that performance was unimpacted? can you share some numbers (especially for bidirection_astar)?

maksimandrianov:
```
// master
time ./run_with_server.py --test-file ../test_requests/de_benchmark_routes.txt --url http://localhost:8002/route --concurrency 1
Placing 2550 results in 20210503_135856_de_benchmark_routes
9%
14%
18%
24%
29%
34%
39%
44%
49%
55%
60%
65%
70%
75%
80%
85%
90%
95%
100%
./run_with_server.py --test-file ../test_requests/de_benchmark_routes.txt      4.08s user 0.91s system 5% cpu 1:39.65 total
// this PR
time ./run_with_server.py --test-file ../test_requests/de_benchmark_routes.txt --url http://localhost:8002/route --concurrency 1
Placing 2550 results in 20210503_140129_de_benchmark_routes
9%
14%
18%
24%
29%
34%
39%
44%
50%
55%
60%
65%
70%
75%
80%
85%
90%
95%
100%
./run_with_server.py --test-file ../test_requests/de_benchmark_routes.txt      4.09s user 0.91s system 5% cpu 1:39.65 total
```

valhalla/sif/dynamiccost.h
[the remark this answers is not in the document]
maksimandrianov: Rolled back to the first version, because `allow_destination_only_` is true by default.

test/gurka/test_conditional_restrictions.cc
maksimandrianov: Done

valhalla/sif/dynamiccost.h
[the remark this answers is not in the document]
genadz: I think by default `allow_destination_only_ = true` because of pedestrian and bicycle profiles. We set to `false` only for the first pass of bidirectional astar https://github.com/valhalla/valhalla/blob/master/src/thor/route_action.cc#L370

src/thor/bidirectional_astar.cc
```diff
@@ -206,8 +206,8 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
   const uint64_t localtime = time_info.valid ? time_info.local_time : 0;
   uint8_t restriction_idx = -1;
   if (FORWARD) {
-    if (!costing_->Allowed(meta.edge, pred, tile, meta.edge_id, localtime, time_info.timezone_index,
-                           restriction_idx) ||
+    if (!costing_->Allowed(meta.edge, false, pred, tile, meta.edge_id, localtime,
```
genadz, line 209: why always `false` ? If I remember correctly a destination edge can also be a connection edge

src/thor/dijkstras.cc
```diff
@@ -181,11 +181,12 @@ void Dijkstras::ExpandInner(baldr::GraphReader& graphreader,
     // Check if the edge is allowed or if a restriction occurs
     EdgeStatus* todo = nullptr;
     uint8_t restriction_idx = -1;
+    bool is_dest = false;
```
genadz, line 184: could add some comment here why `is_dest` always false ?

src/sif/autocost.cc
SvetlanaBayarovich: Got it now, ty

src/thor/bidirectional_astar.cc
genadz: from the other side it's a very specific case. even if we decline an edge on forward step we will use reverse search to make a connection. so, imho it can be always `false` but put some comment with explanation

maksimandrianov:
@gknisely 
Coordinates: -73.92015115260759,40.745479753957255;-73.9178346657051,40.74769327033357
Before:
Screenshot 2021-05-04 at 14 35 19

After:
Screenshot 2021-05-04 at 14 35 26

src/thor/dijkstras.cc
maksimandrianov: Done

src/thor/bidirectional_astar.cc
maksimandrianov: Added comments

maksimandrianov:
@gknisely, @SvetlanaBayarovich tests were added to `valhalla/test_requests/timed_conditional_routes.txt`

src/thor/bidirectional_astar.cc
```diff
@@ -206,8 +206,17 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
   const uint64_t localtime = time_info.valid ? time_info.local_time : 0;
   uint8_t restriction_idx = -1;
   if (FORWARD) {
-    if (!costing_->Allowed(meta.edge, pred, tile, meta.edge_id, localtime, time_info.timezone_index,
-                           restriction_idx) ||
+    // Why is is_dest false?
+    // We have to consider next cases:
+    //  1) At least one step of the reverse search was done -> the direct search never does not
+    //  process the destination edge.
```
genadz, line 212:
```suggestion
    //  1) At least one step of reverse search was done -> forward search will never reach the destination edge.
```

maksimandrianov: Thanks:)

valhalla/sif/dynamiccost.h
```diff
@@ -538,6 +542,8 @@ class DynamicCost {
             // We are in range at the time we are allowed at this edge
             if (access_type == baldr::AccessType::kTimedAllowed)
               return true;
+            else if (access_type == baldr::AccessType::kDestinationAllowed)
+              return allow_conditional_destination_restrictions_ || is_dest;
```
genadz, line 546: I'd rename it to `allow_conditional_destination_` - it directly says: we allow go through edges with `conditional=destination` tag.
Current name confuse me a little, looks like we allow conditional destination restrictions to use=)

valhalla/sif/dynamiccost.h
```diff
@@ -679,6 +685,12 @@ class DynamicCost {
    */
   virtual void SetAllowTransitConnections(const bool allow);
 
+  /**
+   * Sets the flag indicating whether edges with valid restriction conditional=destination are
+   * allowed.
+   */
+  void set_allow_conditional_destination_restrictions(const bool allow);
```
genadz, line 692: After some contemplation I came to the conclusion that it's useless. In this PR we handle `conditional=destination` as an ordinary time restriction. As far as we do not relax any restrictions in case of first run of path algorithm failed, I don't see any reason why we should do it for edges with `conditional=destination` tag. We will simplify code a lot if remove this parameter
@maksimandrianov @SvetlanaBayarovich WDYT ?

valhalla/sif/dynamiccost.h
maksimandrianov: It's better, done

valhalla/sif/dynamiccost.h
genadz: ok, looks like we should route through such roads in case of there are no other ways.

genadz (dismissed):

src/thor/dijkstras.cc
```diff
@@ -498,7 +501,7 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
     uint8_t restriction_idx = -1;
     if (directededge->IsTransitLine()) {
       // Check if transit costing allows this edge
-      if (!tc->Allowed(directededge, pred, tile, edgeid, 0, 0, restriction_idx)) {
+      if (!tc->Allowed(directededge, false, pred, tile, edgeid, 0, 0, restriction_idx)) {
```
SvetlanaBayarovich, line 504: nit: here `false` value is used directly, in the `ExpandInner` function above `is_dest` is declared. Does it make sense to implement it consistently?

src/thor/dijkstras.cc
```diff
@@ -181,11 +181,14 @@ void Dijkstras::ExpandInner(baldr::GraphReader& graphreader,
     // Check if the edge is allowed or if a restriction occurs
     EdgeStatus* todo = nullptr;
     uint8_t restriction_idx = -1;
+    // is_dest is false, because it is a traversal algorithm in this context, not a path search
+    // algorithm. In other words, destination edges are not defined for this Dijkstra's algorithm.
+    bool is_dest = false;
```
SvetlanaBayarovich, line 186:
```suggestion
    const bool is_dest = false;
```
If it's always false?

src/sif/dynamiccost.cc
```diff
@@ -74,8 +74,9 @@ DynamicCost::DynamicCost(const CostingOptions& options,
                          const TravelMode mode,
                          uint32_t access_mask,
                          bool penalize_uturns)
-    : pass_(0), allow_transit_connections_(false), allow_destination_only_(true), travel_mode_(mode),
-      access_mask_(access_mask), closure_factor_(kDefaultClosureFactor), flow_mask_(kDefaultFlowMask),
+    : pass_(0), allow_transit_connections_(false), allow_destination_only_(true),
+      allow_conditional_destination_(false), travel_mode_(mode), access_mask_(access_mask),
```
SvetlanaBayarovich, line 78: I'm a bit concerned that we have different defaults for regular and conditional 'destination' behavior. In what cases are the defaults used? (like I know `allow_destination_only_` is explicitly overridden in bidir a*)

SvetlanaBayarovich (commented):
Generally OK, just a couple of small comments

src/thor/dijkstras.cc
maksimandrianov: false -> is_dest

src/sif/dynamiccost.cc
maksimandrianov: I can say only about `allow_conditional_destination_`.  
Consider next:
1. The first attempt of finding the shortest path: it does not include edges in the middle of the path with active conditional destination restrictions -> the flag allow_conditional_destination_ is false
2. If first attempt is failed. (A path was not found) the second attempt of finding the shortest path is performed. We need relax this (set allow_conditional_destination_ to true) and we allow to go through edges with these active restrictions. 

I think the same as you about allow_destination_only_. This is overridden only for bidir a*

Issue #2976, opened 2021-03-30
Incorrect link reclassification


genadz:
There is a particular example. We have sequence of links: https://www.openstreetmap.org/way/699116538 https://www.openstreetmap.org/way/699116539 https://www.openstreetmap.org/way/25845125 . Despite of the fact that these links connect two trunk roads, valhalla downgrades them to the level 1 because of this tertiary road https://www.openstreetmap.org/way/128268520#map=17/59.31748/14.15687 . I think if a sequence of links leads to some major road we shouldn't downgrade all the sequence of links @dnesbitt61 @kevinkreiser

genadz:
moreover, we mark https://www.openstreetmap.org/way/25845125 as a `turn_channel` which is also suspicious

genadz:
> im not 100% sure what you mean by this but i agree about this case. if the links are connecting two pieces of motorway they can stay motorway links. the problem is when you have somethign like this: https://www.openstreetmap.org/way/9060368

my point is: when you have a sequence of connected links: `a -> b -> c` and `c` leads to a major road, we should mark all the sequence `a,b,c` with the highest class even if the link `b` can lead to some tertiary road

genadz:
@kevinkreiser @dnesbitt61 I tried just to **turn off links reclassification** logic and test performance locally. So, what I did:
1. downloaded Pennsylvania region from osm https://download.geofabrik.de/north-america/us/pennsylvania.html ;
2. generated 10000 random route requests in this region (the points in each request are located at a distance of more than 100 km).

I got the same performance results:

**master**
```
all items: 9569
percentiles	10%	50%	60%	70%	80%	90%	95%	99%	100%
response_time	0.021s	0.041s	0.045s	0.049s	0.056s	0.065s	0.076s	0.165s	0.229s
mean = 0.0437
```

**without links reclassification**
```
all items: 9569
percentiles	10%	50%	60%	70%	80%	90%	95%	99%	100%
response_time	0.021s	0.039s	0.043s	0.048s	0.054s	0.064s	0.076s	0.156s	0.220s
mean = 0.0426
```

so, it should be definitely tested on the whole planet but I don't see a reason why results will be different.

kevinkreiser:
@genadz how do the shortcut counts look, did they change?

genadz:
> @genadz how do the shortcut counts look, did they change?

yes, a little:
from **master** logs:
```
[...]

PR #3042, opened 2021-04-28
Use osm tags in links reclassification algorithm

genadz:
# Issue

This PR fixes the problem described in https://github.com/valhalla/valhalla/issues/2976 .

The main idea is to use `ref` and `destination:ref` tags in order to determine actual destination road for the link. It reduces false positive downgrades in cases when we have low-class roads attached in the middle of the link path.

## Tasklist

 - [x] Add tests

## Requirements / Relations


test/isochrone.cc
```diff
@@ -137,7 +137,7 @@ TEST(Isochrones, Basic) {
...
...2847,52.104463],[5.082407,52.106023],[5.079528,52.105730],[5.078321,52.106797],[5.077321,52.106618],[5.077321,52.104433],[5.078321,52.104674],[5.078969,52.103937],[5.077558,52.102937],[5.080371,52.099937],[5.079321,52.099417],[5.078321,52.100645],[5.077321,52.098424],[5.076321,52.099157],[5.071321,52.098792],[5.070321,52.097363],[5.070074,52.098690],[5.062808,52.105424],[5.058244,52.107860],[5.055535,52.110937],[5.059321,52.112586],[5.062321,52.109844],[5.063233,52.110024],[5.062950,52.111937],[5.057778,52.115937],[5.058321,52.117587],[5.062826,52.117937],[5.062321,52.119025],[5.060321,52.120095],[5.057536,52.119937],[5.059016,52.119632],[5.058321,52.118287],[5.056980,52.119596],[5.055928,52.119544],[5.054321,52.118176],[5.047308,52.116950],[5.049367,52.114983],[5.055683,52.115299],[5.055625,52.113633],[5.053596,52.111662],[5.051321,52.111620],[5.045321,52.114652],[5.043321,52.114448],[5.040321,52.116399],[5.038321,52.116250],[5.035837,52.118453],[5.037321,52.121383],[5.038692,52.121308],[5.039321,52.119998],[5.041321,52.120520],[5.038528,52.122144],[5.038476,52.123782],[5.040367,52.124891],[5.040321,52.125973]],"type":"LineString"},"type":"Feature"}],"type":"FeatureCollection"})";
```
genadz, line 140: fixed isochrone test for `auto` costing model

valhalla/mjolnir/linkclassification.h
```diff
@@ -15,6 +15,7 @@ void ReclassifyLinks(const std::string& ways_file,
                      const std::string& nodes_file,
                      const std::string& edges_file,
                      const std::string& way_nodes_file,
+                     const OSMData& osmdata,
```
genadz, line 18: it's used to extract `ref` and `destination:ref` tags

test/gurka/test_ramps_tc.cc
```diff
@@ -279,3 +279,148 @@ TEST(RampsTCs, test_tc_infer) {
...
+  check_edge_classification(graph_reader, layout, "G", "J", baldr::RoadClass::kSecondary);
+  // GH lead to 2 road, make sure that the best classification was chosen
+  check_edge_classification(graph_reader, layout, "G", "H", baldr::RoadClass::kTrunk);
+  // HM leads to only one tertiary road
+  check_edge_classification(graph_reader, layout, "H", "M", baldr::RoadClass::kTertiary);
+  // HE leads to only one trunk road
+  check_edge_classification(graph_reader, layout, "H", "E", baldr::RoadClass::kTrunk);
+}
+
+TEST(LinkReclassification, test_acyclic_graph) {
```
genadz, line 380: check that we correctly handle cases where graph nodes have several inbound edges and several outbound edges

test/gurka/test_ramps_tc.cc
```diff
@@ -279,3 +279,148 @@ TEST(RampsTCs, test_tc_infer) {
...
+  // check that ramp classification from M1 to M2 wasn't confused by the tertiary road
+  check_edge_classification(graph_reader, layout, "B", "J", baldr::RoadClass::kTrunk);
+  check_edge_classification(graph_reader, layout, "J", "F", baldr::RoadClass::kTrunk);
+
+  // check that ramp classification from M2 to M1 wasn't confused by the tertiary road
+  check_edge_classification(graph_reader, layout, "F", "K", baldr::RoadClass::kTrunk);
+  check_edge_classification(graph_reader, layout, "K", "C", baldr::RoadClass::kTrunk);
+}
+
+TEST(LinkReclassification, test_hierarchical_reclass) {
```
genadz, line 340: check that parent links have higher (or equal) classification than their children

test/gurka/test_ramps_tc.cc
```diff
@@ -279,3 +279,148 @@ TEST(RampsTCs, test_tc_infer) {
...
+                               const gurka::nodelayout& nodes,
+                               const std::string& b,
+                               const std::string& e,
+                               baldr::RoadClass rc) {
+  const auto edge = std::get<1>(gurka::findEdgeByNodes(graph_reader, nodes, b, e));
+  EXPECT_EQ(edge->classification(), rc);
+}
+} // namespace
+
+TEST(LinkReclassification, test_use_refs) {
```
genadz, line 294: check that `ref` and `destination:ref` tags are used correctly to detect destination roads. We should use destination road class instead of intermediate roads classes in order to determine final link classification

genadz:
I've tested these changes on the latest osm data for pennsylvania https://download.geofabrik.de/north-america/us/pennsylvania.html

Reclassification statistics:
|  | reclassified links | turn channels |
| --- | --- | --- |
| **master** | 12157 | 4421 |
| **branch** | 12034 | 4406 |

As we can see it's approximately the same; overall there is **149** differences for links reclassification and **17** for turn channels detection. I verified most of them, looks good to me.

Let me share some examples:

| | master | branch |
| --- | --- | --- |
| https://www.openstreetmap.org/way/197009560 reclass_3 | `residential` | `trunk` |
| https://www.openstreetmap.org/way/448588257 reclass_1 | `residential` | `primary` |
| https://www.openstreetmap.org/way/26404152 reclass_2 | `residential` | `trunk` |


### Run routes

1. Performance stayed the same (on 10000 random routes in pennsylvania).
2. Only **37** routes were changed (expected due to the small difference in links reclassification). For all of them **ETA decreased** or stayed the same.

There is a clear example how incorrect link classification may lead to bad route:
**master**
master_1
**branch**
use_refs_1

src/mjolnir/linkclassification.cc
```diff
@@ -9,6 +9,8 @@
 #include "baldr/graphid.h"
```
kevinkreiser, line 9: at a high level its looks pretty much like a rewrite of the reclassification (which isnt a bad thing) but i was wondering if you might be able to add some comments to this file to draw our attention to the flow of the algorithm. basically explain how this new version works?

test/gurka/test_ramps_tc.cc
```diff
@@ -279,3 +279,148 @@ TEST(RampsTCs, test_tc_infer) {
...
+                 \    |    K----------L
+     I------------J   |   /
+                   \  |  /
+                    \ | /
+                     \|/
+                      F
+                      |
+                      |
+                      E
+  )";
```
kevinkreiser, line 309: this diagram wins the award for **best ascii art in gurka test** :trophy:

genadz: haha :joy: Tank you!))

src/mjolnir/linkclassification.cc
```diff
@@ -134,272 +140,466 @@ nodelist_t FormExitNodes(sequence<Node>& nodes, sequence<Edge>& edges) {
...
-      }
+  static WayTags Parse(const OSMWay& way, const OSMData& osmdata) {
+    WayTags road_tags;
+
+    // Parse 'destination:ref' tag
+    if (way.destination_ref_index() != 0)
+      road_tags.dest_refs = GetTagTokens(osmdata.name_offset_map.name(way.destination_ref_index()));
+
+    // Parse 'ref' tag
+    if (way.ref_index() != 0)
```
gknisely, line 178: @genadz should we be looking at `int_ref` too?

src/mjolnir/linkclassification.cc
```diff
@@ -134,272 +140,466 @@ nodelist_t FormExitNodes(sequence<Node>& nodes, sequence<Edge>& edges) {
...
-          link_edge_indexes.push_back(parent->link_edge_index[i]);
-          break;
-        }
-      }
+  static WayTags Parse(const OSMWay& way, const OSMData& osmdata) {
+    WayTags road_tags;
+
+    // Parse 'destination:ref' tag
+    if (way.destination_ref_index() != 0)
+      road_tags.dest_refs = GetTagTokens(osmdata.name_offset_map.name(way.destination_ref_index()));
```
gknisely, line 175: @genadz should we consider looking at `destination:ref:to` too?

genadz: From osm wiki: "`destination:ref:to=*'s` reference road will never directly connect to the slip road indicated on the sign; it is provided as a means of navigational assistance to the driver."
 
so, I think it doesn't make sense, we won't reach a destination road following by links

src/mjolnir/linkclassification.cc
genadz: according to the https://taginfo.openstreetmap.org/tags/highway=motorway#combinations , `ref` tags is used in more than `94%` cases while `int_ref` is used only in `~23%`. For trunk and primary roads we have similar picture (`ref` tag is much more popular). So, I think `int_ref` tag won't affect classification too much.

and , in general , we definitely can use other tags (like `destination:street; destination:ref:lanes; destination:ref:forward/backward`), but all of these require additional testing round. I'd propose in this PR focus only on `destination:ref; ref` tags

src/mjolnir/linkclassification.cc
genadz: sure!

genadz: @kevinkreiser added some more detailed comments. feel free to ask if you need some more explanations)

src/mjolnir/linkclassification.cc
```diff
@@ -134,272 +140,486 @@ nodelist_t FormExitNodes(sequence<Node>& nodes, sequence<Edge>& edges) {
...
+ * This class builds acyclic link graph starting from some exit node. This node is the root node.
+ * Then we start recursively traversing the graph using only driveforward links. When we reach a
+ * node that doesn't contain nonlink edges we add this node to the graph and continue to exapnd
+ * (except when we've already visited the node). In case a node with nonlink edges is reached we
+ * check if the node has some destination road or not and based on this we determine should we stop
+ * or continue to expand. In order to avoid cycles there are two sets are used: 'processed' - contains
+ * nodes where all their children have been processed; 'in_progress' - contains nodes that have been
+ * visited but some of their children haven't been processed yet (in other words 'in progress' set
+ * contains all nodes on the path from the root node to the current node). So, when we reach a node
+ * from the 'processed' set we should and an edge to the graph (that goes from the current node to the
```
kevinkreiser, line 321:
```suggestion
 * from the 'processed' set we should add an edge to the graph (that goes from the current node to the
```

src/mjolnir/linkclassification.cc
```diff
@@ -134,272 +140,486 @@ nodelist_t FormExitNodes(sequence<Node>& nodes, sequence<Edge>& edges) {
...
+
+    graph_[to].parents.push_back(from);
+    graph_[to].parents_edges.push_back(edge_idx);
+  }
+};
+
+/*
+ * Reclassify links in the acyclic link graph. We maintain a queue of leaf nodes. On each step take
+ * some leaf node from the queue, build a link chain, determine the final road class for the whole
+ * chain (and set the new class), then virtually "remove" reclassified links from the grpaph and
```
kevinkreiser, line 449:
```suggestion
 * chain (and set the new class), then virtually "remove" reclassified links from the graph and
```

kevinkreiser (dismissed):
I reviewed linkclassification.cc and it looks ok to me, the only thing i would say is can you run the planet with this to be sure there aren't any edge cases (especially with the tree formation maybe?)? It would also be nice to run some cross country or longer routes just to validate that the worst case didnt get a bunch worse or something.

I think its just important to be sure it works for the planet before merging something crucial like that is all.

genadz:
> @genadz this still looks good to me, are you blocked on anything here, maybe re-running RAD since the last commit?

hey! you're right. I found one bad route when ran RAD tests for the first time. I fixed it and re-running again

genadz:
So, finally I finished with tests. I used the following test files from `test_requests` directory:
```
bear_vs_straight_routes.txt
de_benchmark_routes.txt
ferry_routes.txt
fork_bear_routes.txt
internal_edge_routes.txt
md_interchange_routes.txt
oh_interchange_routes.txt
pa_interchange_routes.txt
pedestrian_routes.txt
ramp_divebomb_routes.txt
ramp_fork_routes.txt
ramp_routes.txt
ramp_vs_turn_routes.txt
roundabout_routes.txt
small_end_ramp_forks.txt
turn_channel_routes.txt
```
and some other test files.

Most changes are minor (usually it's small difference in ETA). Other cases can be divided into 2 groups:
1. Different routes (with some notable ETA, like >1min).
2. Different instructions.

**Different routes**
Most of routes with different ETA were improved, and other have slight ETA increase (not more than 1min).
Example of improved route (save ~15min):
better_route

**Different instructions**
The same as for the previous section, in most case we got small improvements in guidance (thanks @dgearhart for help).
Example of **improved** guidance:
good_inst
(because the new algorithm reclassified these links [1](https://www.openstreetmap.org/way/12414746), [2](https://www.openstreetmap.org/way/12412946) and [3](https://www.openstreetmap.org/way/112326427) as `trunk_link` while the old one reclassified them as `residential`)

We also detected one **bad** example:
bad_inst
The new algorithm didn't downgrade these links [1](https://www.openstreetmap.org/way/554041398) and [2](https://www.openstreetmap.org/way/708078692) , but `master` reclassified them as `tertiary`. It affected guidance in a bad way.


As far as we definitely see positive dynamic we've decided to merge current branch and try to fix the last example in a new PR.

dgearhart (approved):
Some nice improvements - thanks @genadz

PR #3045, opened 2021-04-30
Pass the graphreader instance from Loki constructor to connectivity m…

mtehver:
Pass the graph reader instance from Loki constructor to connectivity map constructor. This allows using custom graphreader instances that are not constructed from property tree parameters.

Also fix a locking issue in incident_singleton.


kevinkreiser (approved):
Good catch on the missing instantiation for that scoped lock. Looks like we need to run scripts/format.sh for the linter though

mtehver:
@kevinkreiser Thanks for reminding that. I just ran the script and pushed the changes.

PR #3046, opened 2021-05-02
Nutiteq connectivity map construction

kevinkreiser:
fixes #3045 

#3045 needed linted but i wasnt able to manually do it to the existing branch so i pulled the commits into this branch and pr'd it separately


merkispavel (approved):
I thought about committing lint fix to the #3045. Now I see that I couldn't  do it because of the branch from the fork.
Thanks @kevinkreiser!

Issue #3048, opened 2021-05-03
Build unit tests fail with ENABLE_DATA_TOOLS=OFF


DenisPryt:
Mjolnit don't compile without data_tools [link](https://github.com/valhalla/valhalla/blob/7b11d09dded430f76ccad67fc0f32986699f10b7/src/CMakeLists.txt#L163)
```
if(ENABLE_DATA_TOOLS)
  add_subdirectory(mjolnir)
endif()
```
But tests depends on mjolnir [here](#include "mjolnir/graphtilebuilder.h")

PR #3050, opened 2021-05-03
Move dependency to mjolnir from valhalla_test to valhalla_gurka_test

DenisPryt:
# Issue

https://github.com/valhalla/valhalla/issues/3048

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -6,12 +6,11 @@ set_target_properties(gmock_main PROPERTIES FOLDER "Dependencies")
 
 # common things that the tests need
 add_library(valhalla_test
-  test.h test.cc gurka/gurka.h gurka/gurka.cc
+  test.h test.cc
   ${VALHALLA_SOURCE_DIR}/third_party/microtar/src/microtar.h
   ${VALHALLA_SOURCE_DIR}/third_party/microtar/src/microtar.c)
 target_include_directories(valhalla_test PUBLIC
   ${VALHALLA_SOURCE_DIR}/test
-  ${VALHALLA_SOURCE_DIR}/test/gurka
   ${VALHALLA_SOURCE_DIR}/third_party/protozero/include
   ${VALHALLA_SOURCE_DIR}/third_party/libosmium/include
   ${VALHALLA_SOURCE_DIR}/third_party/microtar/src)
--- a/test/gurka/gurka.h
+++ b/test/gurka/gurka.h
@@ -100,6 +100,47 @@ std::vector<std::vector<std::string>> get_paths(const valhalla::Api& result);
 } // namespace detail
 
 /**
+ * Generate a new GraphReader that doesn't re-use a previously
+ * statically initizalized tile_extract member variable.
+ *
+ * Useful if you need to reload a tile extract within the same
+ * process
+ */
+std::shared_ptr<valhalla::baldr::GraphReader>
+make_clean_graphreader(const boost::property_tree::ptree& mjolnir_conf);
+
+/*************************************************************/
+// Creates an empty traffic file
+//
+// To actually customize the traffic data, use `customize_live_traffic_data`
+//
+/*************************************************************/
+void build_live_traffic_data(const boost::property_tree::ptree& config,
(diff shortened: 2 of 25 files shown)


kevinkreiser:
@DenisPryt gurka is a very specific part of our unit testing framework. it doesnt make sense to move more generic stuff like tar building etc into it. it should remain in the general testing section. if you want to, you can move the gurka objects into the test target in cmake and combine them into one bigger testing target but please dont reorganize the code

test/gurka/gurka.h
```diff
@@ -99,6 +99,47 @@ void build_pbf(const nodelayout& node_locations,
...
+                                 const LiveTrafficCustomize& setter_cb);
+
+using HistoricalTrafficCustomize =
+    std::function<boost::optional<std::array<float, baldr::kBucketsPerWeek>>(baldr::DirectedEdge&)>;
+void customize_historical_traffic(const boost::property_tree::ptree& config,
+                                  const HistoricalTrafficCustomize& cb);
+
+using EdgesCustomize = std::function<void(const baldr::GraphId&, baldr::DirectedEdge&)>;
+void customize_edges(const boost::property_tree::ptree& config, const EdgesCustomize& setter_cb);
+
```
kevinkreiser, line 142: this stuff should go back to the file it came from and the rest of those changes reverted. if you want to you can combine gurka sources and test sources into the single test target

test/CMakeLists.txt
```diff
@@ -6,12 +6,11 @@ set_target_properties(gmock_main PROPERTIES FOLDER "Dependencies")
 
 # common things that the tests need
 add_library(valhalla_test
-  test.h test.cc gurka/gurka.h gurka/gurka.cc
```
kevinkreiser, line 9: i think all you really need to do here (no move any code around) is to add an `if` here that check is DATA_TOOLS is enabled and if it is then you can also add the gurka sources to this target.

then inside test.h and test.cc you can check add some `ifdef`s to disable the stuff that depends on mjolnir, rather than mixing all the stuff around.

specifically, not all tests that need mjolnir, need gurka, some of them dont use gurka at all, so forcing the mjolnir dependencies into gurka doesnt make sense

DenisPryt: Maybe would be better to make another utils file with this function, that depends on DATA_TOOLS=ON (mjolnir)?
I think `ifdefs` is not clear solution and can confusing

kevinkreiser: isnt it just as confusing as adding an `if` to cmake? no matter what we do, we have to conditionally include/exclude code. but yeah if you want you can make a new header and new cc file. maybe `test_data.h/cc` or something

DenisPryt: Yes, `if` in cmake confusing too. That's why I decided to add it to gurka because it looks like a place that builds with `DATA_TOOLS`

kevinkreiser: There is no way to get around it, there needs to be an `if` somewhere, my argument is that its not uncommon or confusing. Please do not move these functions into gurka they arent gurka-specific.

DenisPryt: In general, if there is at least one test that uses the functions from mjolnir the tests will breaks down.
Maybe separate in by new folder? Something like `test/mjolnir/test.cc`? (the same way you would do it with a gurka)?

kevinkreiser:
@DenisPryt are you going to open another pr to tackle the issue, i totally agree that we should fix it i just think its ok to break those compilation units out into their own files and conditionally compile them via cmake `if`s. does that sound ok to you or do you favor another approach?

DenisPryt:
@kevinkreiser it sounds good and I preparing new branch now. But I am not sure that I picked a good name for new file.

DenisPryt:
If you want I can move tests that depends on ENABLE_DATA_TOOLS to separate folder (like gurka). But mb it unnecessary.

kevinkreiser:
yeah no need to do that, we already have them in an `if` clause inside of the test/CMakeLists.txt

PR #3051, opened 2021-05-04
Reduce stop impact on parking aisles

SvetlanaBayarovich:
# Issue

Turn weights are very high when driving across parking lots. This sometimes leads to questionable/undesirable routes. For the case below, at some moment in prod
 **current** route has total turn weight equal to `27.25` of `~96` route weight (of which `7.8` is for parking lot). At the same time, **desired** route has total turn weight equal to `58.325` of `~107` route weight (of which `35.425` is for parking lot)
|current|desired|
|-|-|
|image|image|

### Proposed solution
The suggestion is to reduce stop impacts for edge transition if both edges have `Use::kParkingAisle`. As long as such edges are destination_only, no cutting corners through parking lots is possible even with decreased weights.

The example of how the turn weights changed for some data I currently have locally fro the **desired** route from above:
|current|altered|comment|
|-|-|-|
|"turn_weight": 7.8|"turn_weight": 2.6|reduced|
|"turn_weight": 10|"turn_weight": 5|reduced|
|"turn_weight": 1.95|"turn_weight": 0.65|reduced|
|"turn_weight": 2.925|"turn_weight": 0.975|reduced|
|"turn_weight": 3.9|"turn_weight": 1.3|reduced|
|"turn_weight": 0.75|"turn_weight": 0.75|slight turn + src edge has some measured speeds, not impacted|
|"turn_weight": 0.75|"turn_weight": 0.75|slight turn + src edge has some measured speeds, not impacted|
|"turn_weight": 2|"turn_weight": 2|not a parking|
|"turn_weight": 12.5|"turn_weight": 12.5|not a parking|

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/mjolnir/graphenhancer.cc
+++ b/src/mjolnir/graphenhancer.cc
@@ -1077,7 +1077,8 @@ uint32_t GetStopImpact(uint32_t from,
     }
   } else if (edges[from].use() == Use::kParkingAisle && edges[to].use() == Use::kParkingAisle) {
     // decrease stop impact inside parking lots
-    stop_impact /= 2;
+    if (stop_impact != 0)
+      stop_impact -= 1;
   }
   // add to the stop impact when transitioning from higher to lower class road and we are not on a TC
   // or ramp penalize lefts when driving on the right.
--- a/src/mjolnir/graphenhancer.cc
+++ b/src/mjolnir/graphenhancer.cc
@@ -939,7 +939,8 @@ bool IsCyclewayUturn(uint32_t from_index,
  * Higher stop impacts occur when the from and to edges are lower class
  * than the others. There is almost certainly a stop (stop sign, traffic
  * light) and longer waits are likely when a low class road crosses
- * a higher class road. Special cases occur for links (ramps/turn channels).
+ * a higher class road. Special cases occur for links (ramps/turn channels)
+ * and parking aisles.
  * @param  from  Index of the from directed edge.
  * @param  to    Index of the to directed edge.
  * @param  directededge   Directed edge builder - set values.
@@ -1074,6 +1075,9 @@ uint32_t GetStopImpact(uint32_t from,
     } else if (stop_impact != 0) { // make sure we do not subtract 1 from 0
       stop_impact -= 1;
     }
+  } else if (edges[from].use() == Use::kParkingAisle && edges[to].use() == Use::kParkingAisle) {
+    // decrease stop impact inside parking lots
+    stop_impact /= 2;
   }
   // add to the stop impact when transitioning from higher to lower class road and we are not on a TC
   // or ramp penalize lefts when driving on the right.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
    * ADDED: ci: Enable undefined sanitizer [#2999](https://github.com/valhalla/valhalla/pull/2999)
    * ADDED: Optionally pass preconstructed graphreader to connectivity map [#3046](https://github.com/valhalla/valhalla/pull/3046)
    * ADDED: Allow configuration-driven default speed assignment based on edge properties [#3055](https://github.com/valhalla/valhalla/pull/3055)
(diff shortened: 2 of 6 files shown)


kevinkreiser:
Intuitively it make sense to have a high stop impact in parking lots since you have to stop and look etc but i guess that means the algorithm tries to get out of the parking lot as fast as possible. seems like the stop impact is too high compared to regular roads stop impacts. so i guess im saying this makes sense :smile: 

If only CI would kick off an automatic planet wide RAD and you could come back to check it in a couple days. We can dream i guess haha

SvetlanaBayarovich:
@kevinkreiser Yes, you are exactly right. The 

> the algorithm tries to get out of the parking lot as fast as possible

and

> the stop impact is too high compared to regular roads stop impacts

Is exactly what I see.

As for the RAD, I'a about to create new tileset now and share some results as long as it's ready.
However I tested the equivalent change by dividing stop_impact by 2 in *cost.cc files (you can see it in the first commit, I left in intentionally an will squash later). And I could hardly find a couple of cases where the route differs, like it only choses a different gate to enter the parking lot
|route|osm|
|-|-|
|image|image|

SvetlanaBayarovich:
While doing RAD I came across a couple of changes. Of total 10k random_routes.txt only 16 changes were discovered. They fall under 3 categories:
1. Improvement

|||
|-|-|
|image|image|
|image|image|

2. No substantial change

|||
|-|-|
|image **Note** depends on traffic. Current prod shows such route sometimes as well |image Multi-level parking|
|image|image|

3. Questionable/undesirable change

|||
|-|-|
|image **But** Current prod with driving-traffic shows such route sometimes as well|image|
|image|image|

Currently testing a less aggressive strategy which does 
```
stop_impact -= 1;
```
instead of `stop_impact /= 2` to understand how it affects undesirable changes.

SvetlanaBayarovich:
With the less aggressive strategy even fewer diffs occurred. Some favorable changes still remain like 

||||
|-|-|-|
|image|image|image|

as well as some equivalent changes. As for questionable changes, there are only 2 cases left:

|1|2|
|-|-|
|image **But** Current prod with driving-traffic shows such route sometimes as well|image|

For case 1 I think it's a matter of opinion and sometimes traffic. 
For 2 it is a kind of corner case where all surrounding roads are private. In a regular case, however, no "cutting corners" will occur
[area link](https://sites.mapbox.com/id-internal/#background=PNOA-Spain-TMS&disable_features=boundaries&id=w154535638&map=18.30/41.53164/2.44750)
image

src/mjolnir/graphenhancer.cc
```diff
@@ -1074,6 +1075,10 @@ uint32_t GetStopImpact(uint32_t from,
     } else if (stop_impact != 0) { // make sure we do not subtract 1 from 0
       stop_impact -= 1;
     }
+  } else if (edges[from].use() == Use::kParkingAisle && edges[to].use() == Use::kParkingAisle) {
+    // decrease stop impact inside parking lots
+    if (stop_impact != 0)
+      stop_impact -= 1;
   }
```
averkhaturau, line 1082: Pattern `if (stop_impact != 0)  stop_impact -= 1;` is used often in the code. Maybe introduce a lambda like `auto decrease_stop_impact = [&stop_impact](){if (stop_impact != 0)  stop_impact -= 1;}`?

SvetlanaBayarovich: @averkhaturau How about 
```
auto decrement_if_not_zero = [](uint32_t& stop_impact){if (stop_impact != 0) stop_impact -= 1;};
```
I don't quite like the initial suggestion in terms of readability. Additionally I'd rather not capture local value by reference. I understand it's ok in this very case but still don't like such approach

averkhaturau: I'm ok with repeating the original code, imho it doesn't worth to make a general purpose solution for a trivial tasks.

averkhaturau (approved):
lgtm

PR #3052, opened 2021-05-04
move data tools specific test utlis to separate file

DenisPryt:
# Issue

https://github.com/valhalla/valhalla/issues/3048

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/bench/thor/isochrone.cc
+++ b/bench/thor/isochrone.cc
@@ -10,6 +10,7 @@
 #include "loki/worker.h"
 #include "thor/worker.h"
 
+#include "datatools/test.h"
 #include "test.h"
 
 using namespace valhalla;
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -5,16 +5,11 @@ set_target_properties(gmock PROPERTIES FOLDER "Dependencies")
 set_target_properties(gmock_main PROPERTIES FOLDER "Dependencies")
 
 # common things that the tests need
-add_library(valhalla_test
-  test.h test.cc gurka/gurka.h gurka/gurka.cc
-  ${VALHALLA_SOURCE_DIR}/third_party/microtar/src/microtar.h
-  ${VALHALLA_SOURCE_DIR}/third_party/microtar/src/microtar.c)
+add_library(valhalla_test test.h test.cc)
 target_include_directories(valhalla_test PUBLIC
   ${VALHALLA_SOURCE_DIR}/test
-  ${VALHALLA_SOURCE_DIR}/test/gurka
   ${VALHALLA_SOURCE_DIR}/third_party/protozero/include
-  ${VALHALLA_SOURCE_DIR}/third_party/libosmium/include
-  ${VALHALLA_SOURCE_DIR}/third_party/microtar/src)
+  ${VALHALLA_SOURCE_DIR}/third_party/libosmium/include)
 target_link_libraries(valhalla_test valhalla gtest gtest_main gmock pthread)
 
 if(ENABLE_DATA_TOOLS)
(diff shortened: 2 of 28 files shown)


test/CMakeLists.txt
```diff
@@ -5,16 +5,11 @@ set_target_properties(gmock PROPERTIES FOLDER "Dependencies")
...
-  ${VALHALLA_SOURCE_DIR}/third_party/microtar/src/microtar.h
-  ${VALHALLA_SOURCE_DIR}/third_party/microtar/src/microtar.c)
+add_library(valhalla_test test.h test.cc)
 target_include_directories(valhalla_test PUBLIC
   ${VALHALLA_SOURCE_DIR}/test
-  ${VALHALLA_SOURCE_DIR}/test/gurka
   ${VALHALLA_SOURCE_DIR}/third_party/protozero/include
-  ${VALHALLA_SOURCE_DIR}/third_party/libosmium/include
-  ${VALHALLA_SOURCE_DIR}/third_party/microtar/src)
+  ${VALHALLA_SOURCE_DIR}/third_party/libosmium/include)
```
kevinkreiser, line 12: libosmium is only used by gurka, so if we make a new gurka target this should go with it

test/gurka/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 if(ENABLE_DATA_TOOLS)
   file(GLOB_RECURSE TEST_FILES "${CMAKE_CURRENT_LIST_DIR}/test_*.cc")
 
+  add_library(valhalla_gurka_test STATIC gurka.h gurka.cc)
+  target_include_directories(valhalla_gurka_test PUBLIC ${VALHALLA_SOURCE_DIR}/test/gurka)
+  target_link_libraries(valhalla_gurka_test valhalla_test valhalla_datatools_test)
```
kevinkreiser, line 6: wont we have to also link this new target over in test/CMakeLists.txt for any of the data tools unit tests? Maybe none of them are using stuff from here yet but I suspect we should do it anyway so someone isnt hit by a gotcha later on

test/CMakeLists.txt
DenisPryt: Thanks!

test/gurka/CMakeLists.txt
DenisPryt: Yes, you're right.  I will do this

bench/thor/isochrone.cc
```diff
@@ -10,6 +10,7 @@
 #include "loki/worker.h"
 #include "thor/worker.h"
 
+#include "datatools/test.h"
```
kevinkreiser, line 13: why would this be needed? the only thing it seems to use is `test::make_config` that one should not be in the  datatools target as it has nothing to do with datatools. actually let me take a closer look at what you moved into there...

test/datatools/test.cc
```diff
@@ -0,0 +1,249 @@
...
+    edges.reserve(tile.header()->directededgecount());
+
+    GraphId edgeid = tile_id;
+    for (size_t j = 0; j < tile.header()->directededgecount(); ++j, ++edgeid) {
+      edges.push_back(tile.directededge(j));
+      setter_cb(edgeid, edges.back());
+    }
+    tile.Update(nodes, edges);
+  }
+}
```
kevinkreiser, line 247: these two functions look like the only two things that wouldnt work if a person build the repository with datatools turned off. which even more makes me think it would be fine to just ifdef them rather than make these huge changes.

did i miss something here???

kevinkreiser (changes requested):
i suggest something a heck of a lot more simple: https://github.com/valhalla/valhalla/pull/3053

test/datatools/test.cc
DenisPryt: Hm, maybe you are right. I need to use `add_compile_definitions` to add define, right?

DenisPryt:
Same issue PR: https://github.com/valhalla/valhalla/pull/3053

PR #3053, opened 2021-05-04
fixes running unit tests when building with -DENABLE_DATA_TOOLS=Off

### LINKED ISSUE #3048: Build unit tests fail with ENABLE_DATA_TOOLS=OFF

kevinkreiser:
Fixes #3048

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,6 +218,7 @@ endif()
 
 ## Mjolnir and associated executables
 if(ENABLE_DATA_TOOLS)
+  add_compile_definitions(DATA_TOOLS)
   find_package(Boost 1.51 REQUIRED COMPONENTS program_options)
   find_package(SQLite3 REQUIRED)
   find_package(SpatiaLite REQUIRED)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -5,8 +5,12 @@ set_target_properties(gmock PROPERTIES FOLDER "Dependencies")
 set_target_properties(gmock_main PROPERTIES FOLDER "Dependencies")
 
 # common things that the tests need
+set(TEST_SRCS test.h test.cc)
+if(ENABLE_DATA_TOOLS)
+  list(APPEND TEST_SRCS gurka/gurka.h gurka/gurka.cc)
+endif()
 add_library(valhalla_test
-  test.h test.cc gurka/gurka.h gurka/gurka.cc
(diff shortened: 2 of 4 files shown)


kevinkreiser:
cc @DenisPryt

CMakeLists.txt
```diff
@@ -218,6 +218,7 @@ endif()
 
 ## Mjolnir and associated executables
 if(ENABLE_DATA_TOOLS)
+  add_compile_definitions(DATA_TOOLS)
```
kevinkreiser, line 221: this gives us a preprocessor define which we can use in code to disable the parts of the unit test utils that rely on mjolnir (which is disabled when you set DATA_TOOLS=Off)

test/CMakeLists.txt
```diff
@@ -5,8 +5,12 @@ set_target_properties(gmock PROPERTIES FOLDER "Dependencies")
 set_target_properties(gmock_main PROPERTIES FOLDER "Dependencies")
 
 # common things that the tests need
+set(TEST_SRCS test.h test.cc)
+if(ENABLE_DATA_TOOLS)
+  list(APPEND TEST_SRCS gurka/gurka.h gurka/gurka.cc)
+endif()
```
kevinkreiser, line 11: no gurka either unless you enable data_tools (gurka builds tiles of course which is what mjolnir does)

DenisPryt:
ok, I will rework it with add_compile_definitions

kevinkreiser:
@DenisPryt there's no need to rework this PR is already complete

test/test.h
```diff
@@ -84,12 +84,14 @@ using LiveTrafficCustomize = std::function<void(valhalla::baldr::GraphReader&,
 void customize_live_traffic_data(const boost::property_tree::ptree& config,
                                  const LiveTrafficCustomize& setter_cb);
 
+#ifdef DATA_TOOLS
```
kevinkreiser, line 87: these two functions rely on GraphTileBuilder from mjolnir which is disabled when data_tools are disabled

Issue #3021, opened 2021-04-20
Statistically Improving Default Speed Assignments


kevinkreiser:
# Whats the problem?

There are two fairly important considerations when building a routing engine. Path selection and estimated time of arrival for traveling that path. This issue is focused on the latter though in reality they are not independent problems.

If you take Valhalla, build a routing tileset from the OSM planet and run routes on that resulting tileset then you will notice an important deficiency. Currently, routing in certain locales is pretty inaccurate when it comes to the estimated time the route is going to take. I would say that it's especially deficient in urban and heavily trafficked locales. Basically the ETAs we provide out of the box are more akin to driving at night with no one else on the roads. This is unfortunate for 2 reasons.

1. Most people are driving during the day
2. Most people are concentrated in more heavily populated areas

Both of these mean that its statistically more likely that we'll get the kind of routing request that fairs poorly with our current default ETA (which does well at night or in rural areas with less road congestion).

# Why does this happen?

OSM is an amazing dataset without which Valhalla wouldn't exist (let alone countless other companies or divisions of companies). For the purposes of display and basic routing, the dataset has amazing value. But there are some problems. Mainly, we have to make some assumptions about the speed a motorvehicle will travel on a given edge in the graph. We use a few factors to make this assumption:

1. Road class (FRC from the highway tag)
2. Road use (FoW from various tags, roundabout, *_link, turn channel, driveway, track, ...)
2. Road density (urban vs rural)
3. Speed limit (if available)

The general flow of how a default speed gets assigned is:

1. `graph.lua` assigns a default speed to the way via the highway tag (FRC). for some special cases of low class road uses  (driveways and tracks) it sets even lower speeds
2. `mjolnir::PBFGraphParser` checks for other speed tags like speed limit, assumed speed, truck speed etc and stores those on the way as well
3. `mjolnir::GraphBuilder` builds the initial set of tiles and when it does so it sets the initial speed for an edge in the graph. in this case it prefers to use any tagged speed over the default speed we computed in graph.lua
4. `mjolnir::GraphEnhancer` then determines the road density around a given edge and uses this and other derived uses (turn channel etc) to further modify speed assigned to the edge (basically to reduce the speeds of urban edges and turn channels).

# What can we do about it?

## Make it configurable

[...]

PR #3055, opened 2021-05-05
control default speed assignment via config

kevinkreiser:
This PR adds initial support for what was proposed in #3021 of note is the change in proposed file format. We'll document that over in the other repo where the code that generates this output can be found.

~At the moment this PR is still in draft because the unit tests fail. I'm having trouble getting the enhancer to think the gurka map is dense enough to be considered as urban.~

### DIFF
--- a/src/mjolnir/graphenhancer.cc
+++ b/src/mjolnir/graphenhancer.cc
@@ -31,9 +31,11 @@
 #include "baldr/graphid.h"
 #include "baldr/graphreader.h"
 #include "baldr/graphtile.h"
+#include "baldr/rapidjson_utils.h"
 #include "baldr/streetnames.h"
 #include "baldr/streetnames_factory.h"
 #include "baldr/tilehierarchy.h"
+#include "filesystem.h"
 #include "midgard/aabb2.h"
 #include "midgard/constants.h"
 #include "midgard/distanceapproximator.h"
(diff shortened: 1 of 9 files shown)


kevinkreiser:
strange that the changelog CI passed... i didnt add a changelog entry yet...

src/mjolnir/graphenhancer.cc
```diff
@@ -94,58 +96,232 @@ struct enhancer_stats {
...
+   */
+  void UpdateSpeed(DirectedEdge& directededge,
+                   const uint32_t density,
+                   const uint32_t* urban_rc_speed,
+                   bool infer_turn_channels,
+                   const std::string& country_state_code) {
+
+    // If we have config loaded we'll use that
+    if (!tables.empty() && FromConfig(directededge, density, country_state_code))
+      return;
```
kevinkreiser, line 283: i highly recommend viewing this diff with whitespace hidden. if you dont do that it will look like i really modified a lot more than i did. what really happened here is i made this anonymous namespace function `UpdateSpeed` into a member function of `SpeedAssigner` other than indentation (and so also line wrapping) I didnt modify this function except to call the `FromConfig` method to first see if the config-driven speed assignment should be used.

src/mjolnir/graphenhancer.cc
```diff
@@ -1360,16 +1538,21 @@ void enhance(const boost::property_tree::ptree& pt,
...
+  uint32_t urban_rc_speed[] = {
+      89, // 55 MPH - motorway
+      73, // 45 MPH - trunk
+      57, // 35 MPH - primary
+      49, // 30 MPH - secondary
+      40, // 25 MPH - tertiary
+      35, // 22 MPH - unclassified
+      30, // 20 MPH - residential
+      20, // 13 MPH - service/other
+  };
```
kevinkreiser, line 1555: i did make this change because i thought it was easier to read

src/mjolnir/graphenhancer.cc
```diff
@@ -1516,27 +1699,33 @@ void enhance(const boost::property_tree::ptree& pt,
...
           lock.lock();
           endnodetile = reader.GetGraphTile(directededge.endnode());
           lock.unlock();
           end_admin_index = endnodetile->node(directededge.endnode().id())->admin_index();
-          end_node_code = endnodetile->admin(end_admin_index)->country_iso();
+          admin = endnodetile->admin(end_admin_index);
         }
+        end_node_code = admin->country_iso();
+        if (!end_node_code.empty())
+          country_state_code += admin->state_iso();
```
kevinkreiser, line 1720: here i got a concatination of country and state codes for lookup in the config

src/mjolnir/graphenhancer.cc
```diff
@@ -1360,16 +1538,21 @@ void enhance(const boost::property_tree::ptree& pt,
   // Local Graphreader
   GraphReader reader(hierarchy_properties);
 
+  // Config driven speed assignment
+  auto speeds_config = pt.get_optional<std::string>("default_speeds");
+  SpeedAssigner speed_assigner(speeds_config);
```
kevinkreiser, line 1543: here we potentially parse the speed config

src/mjolnir/graphenhancer.cc
```diff
@@ -1629,11 +1818,13 @@ void enhance(const boost::property_tree::ptree& pt,
           directededge.set_use(Use::kFootway);
         }
 
-        // Update speed.
-        UpdateSpeed(directededge, density, urban_rc_speed, infer_turn_channels);
+        // Speed assignment
+        speed_assigner.UpdateSpeed(directededge, density, urban_rc_speed, infer_turn_channels,
+                                   country_state_code);
```
kevinkreiser, line 1823: here we call the anonymous namespace struct to update the speed

src/mjolnir/graphenhancer.cc
```diff
@@ -94,58 +96,232 @@ struct enhancer_stats {
...
-      // If no tagged speed set ramp speed to slightly lower than speed
-      // for roads of this classification
-      RoadClass rc = directededge.classification();
-      if ((rc == RoadClass::kMotorway) || (rc == RoadClass::kTrunk) || (rc == RoadClass::kPrimary)) {
-        speed = (density > 8) ? static_cast<uint32_t>((speed * kRampDensityFactor) + 0.5f)
-                              : static_cast<uint32_t>((speed * kRampFactor) + 0.5f);
-      } else {
-        speed = static_cast<uint32_t>((speed * kRampFactor) + 0.5f);
+struct SpeedAssigner {
+  struct SpeedTable {
```
kevinkreiser, line 132: this object closely mirrors the json but is easier to use than a rapidjson object so we parse the json into it

src/mjolnir/graphenhancer.cc
```diff
@@ -94,58 +96,232 @@ struct enhancer_stats {
...
+      if (!doc.IsArray())
+        throw std::runtime_error("must be a json array");
+
+      // loop over each country/state pair
+      for (const auto& cs : doc.GetArray()) {
+        std::string code = cs["iso3166-1"].GetString();
+        code += cs["iso3166-2"].GetString();
+        tables.emplace(std::move(code), std::array<SpeedTable, 2>{
+                                            SpeedTable(cs["urban"]),
+                                            SpeedTable(cs["rural"]),
```
kevinkreiser, line 186: at the moment we have only urban and rural speed tables per each geographic region (country/state combo)

src/mjolnir/graphenhancer.cc
```diff
@@ -94,58 +96,234 @@ struct enhancer_stats {
...
+                                        });
+      }
+    } // something went wrong with parsing or opening the file
+    catch (const std::exception& e) {
+      LOG_WARN(std::string("Could not load default speeds: ") + e.what());
+      tables.clear();
+    } // something else was thrown
+    catch (...) {
+      LOG_WARN("Could not load default speeds from config");
+      tables.clear();
```
kevinkreiser, line 196: if anything goes wrong when parsing the config its completely disabled

kevinkreiser: maybe this should be parsed when the tile building starts even before the enhancer stage and fail the whole build?

src/mjolnir/graphenhancer.cc
```diff
@@ -94,58 +96,234 @@ struct enhancer_stats {
...
+    }
   }
 
-  // If speed is assigned from an OSM max_speed tag we only update it based
-  // on surface type.
-  if (directededge.speed_type() == SpeedType::kTagged) {
-    // Reduce speed on rough pavements. TODO - do we want to increase
-    // more on worse surface types?
-    if (directededge.surface() >= Surface::kPavedRough) {
+  bool FromConfig(DirectedEdge& directededge,
```
kevinkreiser, line 200: this is the main function that either uses a speed from config or aborts and relies on the previous method. anywhere where the method returns `false` it is signalling that it couldnt assign a speed for this edge. ive commented the sections below to make it pretty clear. if its a kind of edge we know we cant find a speed for either becuase its a special edge that we dont handle or because the geography is not in the config then we abort immediately and rely on the conventional speed setting. once it gets past that point it figures out whether it needs rural or urban speeds. finally it picks which speed to assign.

ill add a javadoc to this function

src/mjolnir/graphenhancer.cc
```diff
@@ -94,58 +96,249 @@ struct enhancer_stats {
...
-    } else if ((use == Use::kRamp) && (directededge.speed_type() != SpeedType::kTagged)) {
-      // If no tagged speed set ramp speed to slightly lower than speed
-      // for roads of this classification
-      RoadClass rc = directededge.classification();
-      if ((rc == RoadClass::kMotorway) || (rc == RoadClass::kTrunk) || (rc == RoadClass::kPrimary)) {
-        speed = (density > 8) ? static_cast<uint32_t>((speed * kRampDensityFactor) + 0.5f)
-                              : static_cast<uint32_t>((speed * kRampFactor) + 0.5f);
-      } else {
-        speed = static_cast<uint32_t>((speed * kRampFactor) + 0.5f);
+struct SpeedAssigner {
```
kevinkreiser, line 134: anyone have opinions on moving this new class out into a private header?

gknisely:
@kevinkreiser what are the resulting RAD diffs with this PR

kevinkreiser:
I wasn't planning on doing RAD for this but I can do a small area if you prefer. Please see my PR comments, especially this one: https://github.com/valhalla/valhalla/pull/3055#discussion_r627008146

The new code is disabled completely and the old code is called as normal so long as you dont pass a speeds config file in your valhalla config json.

src/mjolnir/graphenhancer.cc
```diff
@@ -94,58 +96,249 @@ struct enhancer_stats {
...
+    "link_turning": [15,16,17,18,19,20],
+    "roundabout": [21,22,23,24,25,26,27,28],
+    "driveway": 29,
+    "alley": 30,
+    "parking_aisle": 31,
+    "drive-through": 32
+  },
+  "rural": {
+    "way": [33,34,35,36,37,38,39,40],
+    "link_exiting": [41,42,43,44,45,46],
```
gknisely, line 117: @kevinkreiser we don't have any TC or links less than kTertiary

kevinkreiser: when i scanned my latest planet i did find a very small number of `residential_link`s but i think you're right the number is so low its not worth it. thanks for pointing that out :+1:

gknisely: @kevinkreiser hmmm confused a bit.  based on this https://github.com/valhalla/valhalla/blob/master/src/mjolnir/linkclassification.cc#L231 the lowest classification for *_link would be `kTertiary`

kevinkreiser: there were 10 links in my planet file that had a class of residential. tbh im not surprised that reclassification didnt touch them. they are already a low class road which means the reclassifier wouldnt want to touch them anyway since they are already low classed.

but in any case ive removed this from the configuration as you suggested. also its important to note that im not trying to build this config to match valhallas schema, im trying to build it to match OSM in general so that any routing project can use it. the final home for this format will be determined over here: https://github.com/OpenStreetMapSpeeds

gknisely (approved):
If small RAD testing looked good, then merge.  Thanks.

rzyc (dismissed):

src/mjolnir/speed_assigner.h
```diff
@@ -0,0 +1,353 @@
...
+   */
+  bool UpdateSpeed(DirectedEdge& directededge,
+                   const uint32_t density,
+                   bool infer_turn_channels,
+                   const std::string& country_code,
+                   const std::string& state_code) const {
+
+    // If we have config loaded we'll use that
+    if (!tables.empty() && FromConfig(directededge, density, country_code, state_code)) {
+      directededge.set_speed_type(SpeedType::kClassified);
```
kevinkreiser, line 258: @rzyc i added this since you last reviewed realizing that it would be set improperly if the speed was a tagged speed

nilsnolde:
@kevinkreiser just had a little time going over this nice addition! Looks really good, thanks for that. 

Do I read the code correctly, that, provided a valid JSON exists, the speed enhancer doesn't care whether the edge was tagged or not and updates its speed regardless, i.e. here:
https://github.com/valhalla/valhalla/blob/edf42ea46f1c8104f168ae84460a0ec849e976fd/src/mjolnir/speed_assigner.h#L257-L260

If so, why would you prefer to change the tagged speed to a default speed from the JSON? My assumption would be the tagged speed is (hopefully) closer to the ground truth than a country/state-wide default speed.

nilsnolde:
Maybe it could/should be adjustable in the config by a boolean whether one wants to override tagged speeds with the JSON values? 

And if tagged speeds won't be overridden by default, how about preparing a default JSON following the ["official"](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#Default_speed_limits_by_country) OSM recommendations with all values being the same for urban/suburban/rural?

That also reminds me: IMO it should be possible to have missing keys or at least have some keys populated with `0` values which will signal to the algo "don't mind me, use `UpdateSpeed` to determine my speed". If I'm reading it right, currently it would actually determine 0 edge speed if a JSON value would be 0.

kevinkreiser:
@nilsnolde 

> My assumption would be the tagged speed is (hopefully) closer to the ground truth than a country/state-wide default speed

I think not actually! The vast vast majority of tagged speeds are maxspeed tags. Speed limit is not a measured speed it is a suggested speed. I would agree that in areas with low traffic volumes most people either drive at or above the speed limit and in these scenarios it is a good proxy for actual travel speed. I suspect that by using a measured speed that in those areas where this is the case we'll probably not much change from what the tagged speed is. Where this will make a difference is in cities or areas with high road congestion (hopefully). I did a manual tuning and tested an area in philly and got much better times out of it. Basically places where the speed limit is aspirational :smile: 

However you question does resurface a thought that I was having and I think i commented somewhere. We could, instead of putting these values into `DirectedEdge::speed_` we could put them into `DirectedEdge::constrained_flow_`. Normally that is reserved for historical traffic but the whole point of this work is if you dont have real historical traffic it gets you something better than the default speed assignments. Even better why not put all the default speed asignments into `constrained_flow_` too and migrate `speed_` to just be the speed limit if populated. I think that is why i marked that issue linked above as `v4` because it would be a breaking tile spec change.

Regarding the other questions about when and with what to override they are all good points. I'm trying to add this support iteratively meaning I wanted to get something working and merged and come back to do further enhancements and protections which you've listed:

1. We could add a flag that says dont override tagged speeds as is done for the default speed rules
2. Ship the official OSM default speeds as a possible default
3. Add fall back for unknown values to previous speed method or to adjacent geograph or something else...
4. Handle invalid values in the json

nilsnolde:
> The vast vast majority of tagged speeds are maxspeed tags

makes sense:) That does hit close to being able to distinguish `maxspeed` tags from other speed tags, in which case I'd prefer to keep the tagged speed rather than taking a default. Maybe those tags will be more abundant in the future by automatic data collection from all those rolling 2 ton computers..

I'll help out a little on adding those things, starting with 1. which I can already see clients whining about :sweat_smile:

Issue #3064, opened 2021-05-07
Possible null pointer dereference


DenisPryt:
We have places with no check tile pointer for nullptr. This could cause a crash if we don't have some tiles in memory.

PR #3065, opened 2021-05-07
Fix possible null pointer dereferencing

DenisPryt:
# Issue

https://github.com/valhalla/valhalla/issues/3064

## Tasklist

 - [ ] Add tests

## Requirements / Relations


kevinkreiser:
Rather than blanketing the code with this could you maybe figure out which one you are seeing in practice and fixt that one? A lot of these will be unnecessary.

For example. In the Origin function of bidirectional a* the ede id's come from loki using the same graph reader. It is possible to write code to clear the tile cache between loki and Thor but it's harder to understand why someone would do that.

So yeah please can you focus the pr to a specific case you were running into? Or maybe you are using the library in another project and the cache can be cleared at any time by another thread?

A bit more explanation of what is going on here would be great!

DenisPryt:
@kevinkreiser Sorry for the incomplete description. This fixes for MapBox project. And yes, in our case the cache can be cleared from another thread.

kevinkreiser:
thats a lot of branches and print statements. we should clean up print statements and we should check what impacts all of these branches have on the code in terms of performance

DenisPryt:
@kevinkreiser `printf` statements were there before my changes and used inside `#ifdef LOGGING_LEVEL_TRACE` 
I removed LOG statement that I added.
If `if`s will be a problem we can use hints for branch prediction cpu unit. Without this `if`s we can just crash

src/thor/bidirectional_astar.cc
```diff
@@ -1016,6 +1040,9 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
       const BDEdgeLabel& edgelabel = edgelabels_forward_[edgelabel_index];
 
       const auto* edge = graphreader.directededge(edgelabel.edgeid(), tile);
+      if (edge == nullptr) {
```
kevinkreiser, line 1043: this just seems crazy! for this to happen it means the algorithm found the edge (meaning it had the tile at some point) then the tile disappeared but couldnt be gotten again. should we really support such  behavior!? i understand the cache could be cleared at any time by another thread but in that case shouldnt `GetGraphTile` block until it can go get the damned tile again? why would you write a tile cache implementation that has such behavior? i suppose if its network based the connection can be intermittent but still this is crazy.

`continue` is also absolutely not the thing you want to do here. that continue just skips the edge in the path which is totally wrong. it means you'll return whatever part of the route is retrievable from cache. the route could be skipping all over the map! if this were to happen you should just:

```c++
throw std::runtime_error("tile cache is connected to tile source via a piece of wet spaghetti and is randomly able and unable to load tiles. please ask the implementer to think a bit more clearly about how he designs a cache");
```

the crux of the issue is that you cant distinguish between a tile you cant get and a tile that doesnt exist

kevinkreiser (changes requested):

src/thor/bidirectional_astar.cc
```diff
@@ -1016,6 +1040,10 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
...
+        throw std::runtime_error(
+            "tile cache is connected to tile source via a piece of wet spaghetti and is randomly able and unable to load tiles. please ask the implementer to think a bit more clearly about how he designs a cache");
```
kevinkreiser, line 1045: i didnt mean you should literally use that string as the exception... my point is broader, instead of blindly continuing when you cant find a tile you need to t hink first, should have i been able to get this tile? is the algorithm at a point where it has clearly seen this tile before but now for some reason cannot get the tile again. in these cases we need to throw immediately and stop the computation.

DenisPryt: :D 
Sorry about that, I will think about a normal message.

src/thor/bidirectional_astar.cc
SiarheiFedartsou: Hi @kevinkreiser 

>  in that case shouldnt GetGraphTile block until it can go get the damned tile again

Do you mean that we should prevent any cache evictions until route building is finished?

kevinkreiser: @SiarheiFedartsou that is the assumption that the code has been working on since it was written. if you can do that, that would be ideal because it removes the need for all of these additional checks.

if you absolutely must purge cache while the computation is running then i find it unreasonable that the cache would have been able to get a tile before but cant get it back into cache now. in practical cases most caches will use disk as a place to put the tile when its not in memory. it seems extremely unlikely that you wouldnt be able to read the tile back off of the disk if some other thread cleared it from memory

kevinkreiser: but i think at the end of the day, this is the important part:

> the crux of the issue is that you cant distinguish between a tile you cant get and a tile that doesnt exist

up until now the interface of the graphreader was such that `nullptr` means that the tile DOES NOT EXIST

it seems to me that your use-case is now opening up another possibility of the tile CANNOT BE LOADED

if your tilecache knows that its the latter, then it should throw right away and let that bubble all the way up to the algorithm where the algorithm will be stopped.

basically we change graphreaders interface to be:

nullptr == tile does not exist in this tile set
exception == tile cache failure to get tile

and this has the added benefit of keep the code the way it is and just adding throwing to the tile cache

SiarheiFedartsou: > it seems extremely unlikely that you wouldnt be able to read the tile back off of the disk if some other thread cleared it from memory

Yeah, but it is possible that we could remove it from disk completely, i.e. we will have no chance to get it back(since we will have to load it from the network)

kevinkreiser: sure, but my point is if that happens and you cant get it back from network then you should throw in the tile cache. the problem isnt one you can gracefully handle. if you lose the ability to get tiles you need to just stop doing routing computations because the results of those will be wrong anyway.

DenisPryt: In this case you are right, and we should throw exception, but we can't do it from cache because we want to build route even some tile doesn't unavailable.

kevinkreiser:
## Algorithms Should Not Be Robust to Data Disappearing On-the-fly

We'll start by clearing up one thing. None of the algorithms are robust to randomly removing tiles out from underneath them. The `trim` and `clear` operations on `GraphReader` were meant to reclaim memory between top level methods on the `actor_t` interface. Like this:

```
actor.route(graphreader)
graphreader.clear()
actor.mapmatch()
graphreader.clear()
// and so on ...
```

Yes the code is robust to a tile not existing or not being on the device that is fine but its not robust to it being there for part of the computation and going away the next. And actually it SHOULNT be anyway. Think about what would happen in a couple scenarios and you'll realize why its not worth it to try to make the code "gracefully" handle this.

1. loki finds candidate edges but the tiles those edges are in are purged by another thread, thor starts up with those candidate edges and the route is immediately failed
2. thor does part of an expansion in one tile moves on expanding in another tile, the first tile gets removed from cache, now we start popping edges from the queue that are in that tile, we have to skip them, we wasted CPU expanding in that tile. whats worse is we can some non-deterministic routing depending on the state of the cache at any one time, do you think the user wants a route that randomly differs from minute to minute?
3. thor finds a path but before it goes to serialize that into a trip path object you yank the tiles from it, you wasted that time findign the route only to have to abort and fail the route because you cant do anything with it since you no longer have the data

As you can see, we can just keep playing little scenarios like this out over and over and every time it doesnt make sense to continue our work.

We need `clear` and `trim` to only happen when routing isn't already in progress in one of the threads. Basically we just need extra synchronization around cache cleanup and we dont need to try to "do something" inside the algorithms themselves.

## Your Particular Problem: Routing on Regional Extracts

I'm pretty sure the scenario you are worrying about is the following:

You configure graphreader to cache tiles in memory, but the in memory cache is backed up by disk and the disk only has a subset of tiles from the planet on it so its backed up by a network call. Your cache hierarchy is like this:

    memory -> disk -> network

To screw us even worse we have multiple threads sharing synchronized access to this tile cache and they all have the power to `clear` it whenever they want to keep your application within some reasonable memory/disk limits.

Now I get that you still want to route with the data you have and you want to be able to clear some of it to stay within the boundaries of what the configuration demands. The router already supports that so long as you understand you can only shrink back down to your limit if as route isnt already in progress. For the reasons described above it doesnt make sense to have the algorithms constantly checking if the data they just saw still exists (for the same routing calculation).

## Potential Solution

I believe you are using this library in another application where you have the individual workers (loki, thor, odin) and they share a thread synchronized tile cache via the graphreader. I also believe when you do a route or a map match etc you probably call `trim`/`clear` afterwards. I suggest you remove that call to trim the cache and instead spawn another thread whose sole purpose is to call `trim`/`clear` on the graphreader when its over its limits. I suggest that you add a shared lock with the other threads that do routing and map matching and that when they are doing routing or map matching they acquire the lock so that the `trim`ming thread cannot clear any cache while a route/map match is in progress. In this way we dont have to make the algorithms robust for pretty much no gain and you can still make sure your application stays within the limits.

kevinkreiser (changes requested):
I still dont think this is the right way to go. It is pointless to make the router robust to "disappearing" tiles. We should be robust to running on regional tile sets (ie nullptrs are allowed) but tiles cant be there in one iteration of the loop and not be there in the next.

Please see my comments here: https://github.com/valhalla/valhalla/pull/3065#issuecomment-842340207

kevinkreiser:
if we assume that we need the router to deal with tiles blinking into and out of existence then we need to look at each one of these changes and determine can the algorithm even continue past this point if it cant get the tile. If it can we continue if it cant we throw and abort the computation.

For continuing some common cases would be:

1. we pop an edgeid out of the queue but the tile that holds it is now gone. we can `continue` the algorithm but the results might be pretty weird
2. ??

For aborting the computation common cases would be:
1. loki finds edge candidates calls thor, thor cant get the tiles with the candidates, so we `throw`
2. thor finds a path but while recovering it by following the edge labels it finds an edge in a tile it cant get, so we `throw`
3. thor finds a path sends it off to triplegbuilder but when its iterating over it it cant get a tile, so we `throw`

@mskurydin @DenisPryt @SiarheiFedartsou it seems to me the only "recoverable" one above is when a tile goes away and we basically cut all search paths (expansions) that are in it out and continue. another other failure should `throw` and abort the computation immediately. i assume that if this happens you'll want to know that it did and retry, this time without disappearing tiles presumably?

if that is the case i kind of wonder if we should just always throw when we detect a tile disappeared out from under us and then retry. like maybe we make a method on graphreader that is called `GetGraphTileOrThrow` and we call it anywhere where we couldn't tolerate a tile being missing. basically all the places in the algorithm that we should never see an edge_id but not be able to get the tile that goes with it.

anyway i'll go through all of these and mark each type `continue` or `throw` if they are mishandled

src/thor/bidirectional_astar.cc
```diff
@@ -895,15 +897,22 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader,
     }
     // Get the directed edge
     graph_tile_ptr tile = graphreader.GetGraphTile(edgeid);
+    if (tile == nullptr) {
+      continue;
+    }
     const DirectedEdge* directededge = tile->directededge(edgeid);
 
     // Get the opposing directed edge, continue if we cannot get it
-    GraphId opp_edge_id = graphreader.GetOpposingEdgeId(edgeid);
```
kevinkreiser, line 901: instead of the other changes below i would simplify this to this right at the beginning:

```c++
graph_tile_ptr opp_tile = nullptr;
const DirectedEdge* opp_dir_edge = nullptr;
auto opp_edge_id = graphreader.GetOpposingEdgeId(edgeid, opp_dir_edge, opp_tile);
if (!opp_dir_edge)  {
  continue;
}
```

then you dont need any more other checks and continues, just use the variables if it gets passed this first check

src/thor/bidirectional_astar.cc
```diff
@@ -1016,6 +1040,10 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
       const BDEdgeLabel& edgelabel = edgelabels_forward_[edgelabel_index];
 
       const auto* edge = graphreader.directededge(edgelabel.edgeid(), tile);
+      if (edge == nullptr) {
+        throw std::runtime_error("BidirectionalAStar::FormPath can't find edge by id " +
```
kevinkreiser, line 1044: i think a more exact message would be in order here.

`"BidirectionalAStar::FormPath failed, tile no longer available " + std::to_string(edgelabel.edgeid().Tile_Base())`

src/thor/bidirectional_astar.cc
```diff
@@ -1041,6 +1069,10 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
       const BDEdgeLabel& edgelabel = edgelabels_reverse_[edgelabel_index];
       const DirectedEdge* opp_edge = nullptr;
       GraphId opp_edge_id = graphreader.GetOpposingEdgeId(edgelabel.edgeid(), opp_edge, tile);
+      if (opp_edge == nullptr) {
+        LOG_INFO("graphreader.GetOpposingEdgeId(edgelabel.edgeid(), opp_edge, tile) is nullptr");
+        continue;
```
kevinkreiser, line 1074:
```suggestion
        throw std::runtime_error("BidirectionalAStar::FormPath failed, tile no longer available");
```

src/thor/bidirectional_astar.cc
```diff
@@ -1181,7 +1213,6 @@ bool IsBridgingEdgeRestricted(GraphReader& graphreader,
     const auto* edge = tile->directededge(edgeid);
     if (edge == nullptr) {
       throw std::logic_error("Edge pointer was null in IsBridgingEdgeRestricted");
-      return false;
     }
```
kevinkreiser, line 1216: delete this block completely its redundant. the call to `GraphTile::DirectedEdge` already throws if the edgeid doesnt make sense

src/thor/dijkstras.cc
```diff
@@ -214,6 +214,9 @@ void Dijkstras::ExpandInner(baldr::GraphReader& graphreader,
                 costing_->EdgeCost(directededge, tile, offset_time.second_of_week, flow_sources) +
                 transition_cost;
     } else {
+      if (opp_pred_edge == nullptr) {
+        continue;
+      }
```
kevinkreiser, line 219: there is no point to this, you protected against it on line 361 below

```suggestion
```

src/thor/dijkstras.cc
```diff
@@ -354,7 +357,10 @@ void Dijkstras::Compute(google::protobuf::RepeatedPtrField<valhalla::Location>&
 
     const baldr::DirectedEdge* opp_pred_edge = nullptr;
     if (expansion_direction == ExpansionType::reverse) {
-      opp_pred_edge = graphreader.GetGraphTile(pred.opp_edgeid())->directededge(pred.opp_edgeid());
+      const auto opp_edge_tile = graphreader.GetGraphTile(pred.opp_edgeid());
+      if (opp_edge_tile != nullptr) {
+        opp_pred_edge = opp_edge_tile->directededge(pred.opp_edgeid());
+      }
```
kevinkreiser, line 363: we should fail here as soon as we know we cant get the tile, not wait until later

```suggestion
      opp_pred_edge = graphreader.GetOpposingEdge(pred.opp_edgeid());
      if (opp_edge_tile != nullptr) {
        continue;
      }
```

src/thor/dijkstras.cc
```diff
@@ -858,7 +870,12 @@ void Dijkstras::SetDestinationLocations(
       if (!opp_edge_id.Is_Valid()) {
         continue;
       }
+
       const DirectedEdge* opp_dir_edge = opp_tile->directededge(opp_edge_id);
+      graph_tile_ptr opp_edge_tile = graphreader.GetGraphTile(opp_edge_id);
+      if (opp_edge_tile == nullptr) {
+        continue;
+      }
```
kevinkreiser, line 878: again simplify all this by getting it all in one call:

```c++
graph_tile_ptr opp_tile = tile;
const DirectedEdge* opp_dir_edge = nullptr;
auto opp_edge_id = graphreader.GetOpposingEdge(edgeid, opp_dir_edge, opp_tile);
if (opp_dir_edge == nullptr) {
  continue;
}
```

src/thor/route_action.cc
```diff
@@ -175,6 +175,9 @@ std::string thor_worker_t::expansion(Api& request) {
                                 baldr::GraphId edgeid, const char* status, bool full_shape = false) {
     // full shape might be overkill but meh, its trace
     auto tile = reader.GetGraphTile(edgeid);
+    if (tile == nullptr) {
+      return;
```
kevinkreiser, line 179: i'm not so sure about what to do wit this one. the point of the expansion API is to visualize the way the algorithm traverses the graph. if it gets here that means it did in the algorithm but we cant track it properly because the tile disappeared... its this kind of thing that makes me think this is just a terrible idea in general.

i get the idea of "best attempt" but the scenario in which that is an acceptable outcome is very limited.

anyway i think we should at least `LOG_ERROR` here that the expansion result will be incomplete

src/thor/triplegbuilder.cc
```diff
@@ -1138,8 +1138,15 @@ void TripLegBuilder::Build(
   // Get the first nodes graph id by using the end node of the first edge to get the tile with the
   // opposing edge then use the opposing index to get the opposing edge, and its end node is the
   // begin node of the original edge
-  auto* first_edge = graphreader.GetGraphTile(path_begin->edgeid)->directededge(path_begin->edgeid);
+  auto begin_tile = graphreader.GetGraphTile(path_begin->edgeid);
+  if (begin_tile == nullptr) {
+    return; // or throw?
```
kevinkreiser, line 1143: `throw` everything in triplegbuilder has to be a throw. at this point in the computation you have the route. if you cant get the tiles along it then you cant proceed to serialize a route.

src/thor/triplegbuilder.cc
```diff
@@ -1297,7 +1310,7 @@ void TripLegBuilder::Build(
     float trim_end_pct = is_last_edge ? end_pct : 1;
 
     // Process the shape for edges where a route discontinuity occurs
-    uint32_t begin_index = (is_first_edge) ? 0 : trip_shape.size() - 1;
+    uint32_t begin_index = (is_first_edge || trip_shape.size() == 0) ? 0 : trip_shape.size() - 1;
```
kevinkreiser, line 1313: this seems unrelated please remove

kevinkreiser (changes requested):
i just went over the whole PR. please make the requested changes. after that we need to do a performance check in sever mode just to make sure we dont have any regressions there (hopefully branch prediction will help us there)

src/thor/dijkstras.cc
DenisPryt: @kevinkreiser I think on line 361 we can pass `opp_pred_edge` as `nullptr`. Or am I wrong?
```
const baldr::DirectedEdge* opp_pred_edge = nullptr;
if (expansion_direction == ExpansionType::reverse) {
  const auto opp_edge_tile = graphreader.GetGraphTile(pred.opp_edgeid());
  if (opp_edge_tile != nullptr) {
    opp_pred_edge = opp_edge_tile->directededge(pred.opp_edgeid());
  }
}
```

kevinkreiser: sorry im saying on line 361 you should not let it get to this point. basically you may as well stop it sooner than let it get to here

src/thor/dijkstras.cc
kevinkreiser: https://github.com/valhalla/valhalla/pull/3065#discussion_r637948351 refers to this suggestion

src/thor/dijkstras.cc
DenisPryt: Thanks for the explanation :)

DenisPryt:
@kevinkreiser should I add new exception like `tile_no_longer_available_error` or just use `std::runtime_error` ?

kevinkreiser:
@DenisPryt it would be a nice-to-have but i wouldnt say it is required. its your choice!

src/thor/bidirectional_astar.cc
DenisPryt: @kevinkreiser 
We have an issue inside `for (auto edge_itr = path_begin; edge_itr != path_end; ++edge_itr, ++edge_index) {`
We can `continue` loop. And I think in case we skip first edge we can't use `is_first_edge` flag. And `std::prev` for iterator at all. 
Maybe I should make `throw` instead of `continue` inside this loop?
Or will be better to make `prev_edge_itr` variable and use it instead of `is_first_edge` flag and `std::prev(edge_itr)`?

DenisPryt: About `trip_shape.size() == 0`. It can happen if we skip all iterations here
```
for (auto edge_itr = path_begin; edge_itr != path_end; ++edge_itr, ++edge_index) {
  const GraphId& edge = edge_itr->edgeid;
  graphtile = graphreader.GetGraphTile(edge, graphtile);
  if (graphtile == nullptr) {
    continue;
  }
```

src/thor/triplegbuilder.cc
DenisPryt: @kevinkreiser
We have an issue inside for (auto edge_itr = path_begin; edge_itr != path_end; ++edge_itr, ++edge_index) {
We can continue loop. And I think in case we skip first edge we can't use is_first_edge flag. And std::prev for iterator at all.
Maybe I should make throw instead of continue inside this loop?
Or will be better to make prev_edge_itr variable and use it instead of is_first_edge flag and std::prev(edge_itr)?

DenisPryt: About `trip_shape.size() == 0`. It can happen if we skip all iterations here

```
for (auto edge_itr = path_begin; edge_itr != path_end; ++edge_itr, ++edge_index) {
  const GraphId& edge = edge_itr->edgeid;
  graphtile = graphreader.GetGraphTile(edge, graphtile);
  if (graphtile == nullptr) {
    continue;
  }
  ```

File: src/thor/triplegbuilder.cc [hunk 13]
```diff
@@ -1213,6 +1221,9 @@ void TripLegBuilder::Build(
   for (auto edge_itr = path_begin; edge_itr != path_end; ++edge_itr, ++edge_index) {
     const GraphId& edge = edge_itr->edgeid;
     graphtile = graphreader.GetGraphTile(edge, graphtile);
+    if (graphtile == nullptr) {
+      continue;
```
kevinkreiser, line 1225: `throw`

src/thor/triplegbuilder.cc
```diff
@@ -1221,6 +1232,9 @@ void TripLegBuilder::Build(
     // Set node attributes - only set if they are true since they are optional
     graph_tile_ptr start_tile = graphtile;
     graphreader.GetGraphTile(startnode, start_tile);
+    if (start_tile == nullptr) {
+      continue;
```
kevinkreiser, line 1236: `throw`

src/thor/triplegbuilder.cc
```diff
@@ -1452,9 +1466,11 @@ void TripLegBuilder::Build(
...
   if (controller.attributes.at(kNodeAdminIndex)) {
     auto last_tile = graphreader.GetGraphTile(startnode);
-    node->set_admin_index(
-        GetAdminIndex(last_tile->admininfo(last_tile->node(startnode)->admin_index()), admin_info_map,
-                      admin_info_list));
+    if (last_tile != nullptr) {
+      node->set_admin_index(
+          GetAdminIndex(last_tile->admininfo(last_tile->node(startnode)->admin_index()),
+                        admin_info_map, admin_info_list));
+    }
```
kevinkreiser, line 1473:
```suggestion
    if (last_tile == nullptr) {
      throw;
    }
    node->set_admin_index(
        GetAdminIndex(last_tile->admininfo(last_tile->node(startnode)->admin_index()),
                      admin_info_map, admin_info_list));
```

src/thor/bidirectional_astar.cc
```diff
@@ -1023,7 +1053,9 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
          edgelabel_index = edgelabels_forward_[edgelabel_index].predecessor()) {
       const BDEdgeLabel& edgelabel = edgelabels_forward_[edgelabel_index];
 
-      const auto* edge = graphreader.directededge(edgelabel.edgeid(), tile);
+      const auto* edge = get_directed_edge_or_throw("BidirectionalAStar::FormPath failed",
+                                                    graphreader, edgelabel.edgeid(), tile);
```
kevinkreiser, line 1057: i think  its better without the helper function

```suggestion
      const auto* edge = graphreader.directededge(edgelabel.edgeid(), tile);
      if (edge == nullptr) {
        throw tile_no_longer_available_error_t("BidirectionalAStar::FormPath failed", edgelabel.edgeid());
      }
```

valhalla/thor/tile_no_longer_available_error.h
```diff
@@ -0,0 +1,18 @@
...
+#include "baldr/graphid.h"
+#include <string>
+
+namespace valhalla {
+namespace thor {
+
+struct tile_no_longer_available_error_t : public std::runtime_error {
+  explicit tile_no_longer_available_error_t(const std::string& errormessage);
+  tile_no_longer_available_error_t(std::string prefix, baldr::GraphId edgeid);
+};
```
kevinkreiser, line 13: move this into the graphreader header inside of the baldr namespace. also i suggest a shorter name like maybe `tile_gone_t`

src/thor/bidirectional_astar.cc
DenisPryt: Yes, I agree. It was an experiment that I forgot to remove

DenisPryt:
@kevinkreiser so now we need to make performance tests, right? How can I do it? Just make run-benchmarks?

kevinkreiser:
@DenisPryt different people do it in different ways. Perhaps @genadz could show you the way he and i have been doing it lately. or if you can wait until the evening i can run the check on my end. @danpat did have a way of running the algorithm benchmarks with larger datasets (like a server would) but i cant remember if its documented or not.

DenisPryt:
@kevinkreiser I've been downloading tiles for benchmarks since this morning. If it's not difficult, can you try to run a benchmark?

kevinkreiser:
Yep I can. I'll do it tonight

src/thor/triplegbuilder.cc
```diff
@@ -1138,8 +1138,15 @@ void TripLegBuilder::Build(
   // Get the first nodes graph id by using the end node of the first edge to get the tile with the
   // opposing edge then use the opposing index to get the opposing edge, and its end node is the
   // begin node of the original edge
-  auto* first_edge = graphreader.GetGraphTile(path_begin->edgeid)->directededge(path_begin->edgeid);
+  auto begin_tile = graphreader.GetGraphTile(path_begin->edgeid);
+  if (begin_tile == nullptr) {
+    throw tile_gone_error_t("TripLegBuilder::Build failed", path_begin->edgeid);
+  }
+  auto* first_edge = begin_tile->directededge(path_begin->edgeid);
```
kevinkreiser, line 1145: nit: should be `const auto*`

kevinkreiser (approved):
other than the small nitpick, please at a changelog entry to record that you made this enhancement

i checked the performance and there was no impact as we had hoped. i am certain there are many many more of these types of fixes througout the code base that we will want to do (in loki and in other parts of the algorithms in thor) but we can get them in another pull request.

i do think that in the future we should consider adding to `GraphReader` a method which throws immediately if the tile is not found `RequireGraphTile` that returns a tile just like `GetGraphTile` does but throws an exception if its not there

DenisPryt:
@kevinkreiser done

kevinkreiser:
image

please try not to do this ^^ it makes the review process harder. we only allow squash merges on the repository anyway so rewriting history serves no purpose.

kevinkreiser (approved):

DenisPryt:
> image
> 
> please try not to do this ^^ it makes the review process harder. we only allow squash merges on the repository anyway so rewriting history serves no purpose.

Of course. Sorry for that.

PR #3067, opened 2021-05-08
Use std::shared_ptr in case if ENABLE_THREAD_SAFE_TILE_REF_COUNT is ON

SiarheiFedartsou:
# Issue

During implementation of https://github.com/valhalla/valhalla/pull/2730 we learnt that using of `std::shared_ptr`(which has such ability) has certain performance impact, that's why we ended up using `boost::intrusive_ptr`. Then we added `ENABLE_THREAD_SAFE_TILE_REF_COUNT` to work around issues with non thread-safe reference counter in multithreading environment. Since `ENABLE_THREAD_SAFE_TILE_REF_COUNT` is supposed to be used in the cases when performance is not top-1 priority and sometimes it is very useful to be able to get weak pointer to tile I propose to use `std::shared_ptr` as tile pointer in case of `ENABLE_THREAD_SAFE_TILE_REF_COUNT`(instead of `boost::intrusive_ptr` with thread-safe counter).

## Tasklist

 - [ ] Add tests

## Requirements / Relations


averkhaturau (dismissed):
lgtm

mandeepsandhu:
Is getting a weak pointer from a shared ptr the main motivation for this change?

kevinkreiser (approved):
Looks good to me. I wish we could have just used shared_ptr regardless but the perf penalty was too much.

SiarheiFedartsou:
> Is getting a weak pointer from a shared ptr the main motivation for this change?

Yes. Do you have alternative ideas how it can be solved?

mandeepsandhu:
> > Is getting a weak pointer from a shared ptr the main motivation for this change?
> 
> Yes. Do you have alternative ideas how it can be solved?

No, I think your solution is good! I was wondering if we still need to have `intrusive_ptr` for the non-thread safe usecase?

Minor optional nit: we could rename `ENABLE_THREAD_SAFE_TILE_REF_COUNT` to something like `ENABLE_THREAD_SAFE_TILE_PTR` (since we're not explicitly managing a ref counter). I'm fine leaving it as-is too.

kevinkreiser:
@mandeepsandhu the only reason we have it at all is performance issues. iirc it was about a 6% performance reduction

mandeepsandhu:
> @mandeepsandhu the only reason we have it at all is performance issues. iirc it was about a 6% performance reduction

6% reduction compared to using `shared_ptr` in single or multi-threaded context?

kevinkreiser:
24 thread load test using mem mapped tar with no locking needed (every thread has its own graph reader)

danpat:
@kevinkreiser it's usually strongly encouraged to use const references to avoid triggering shared_ptr internal mutexes - might we be better served refactoring to follow that pattern across the board?  In what cases do we need to pass a non-const reference or copy?

SiarheiFedartsou:
> Minor optional nit: we could rename `ENABLE_THREAD_SAFE_TILE_REF_COUNT` to something like `ENABLE_THREAD_SAFE_TILE_PTR` (since we're not explicitly managing a ref counter). I'm fine leaving it as-is too.

The proposed name is definitely better, but I don't want to break backward compatibility(some Valhalla users use this flag already) and probably I'll leave it as is.

averkhaturau (dismissed):

kevinkreiser (approved):

averkhaturau (approved):

PR #3078, opened 2021-05-18
Don't penalize gates that have allowed access

CuriOusQuOkka:
# Issue

Gates that have specified access information should not be penalised. 
- In case any access info was specified, consider the data correct. Don't penalise the node when access is allowed. 
- Otherwise the gates can be either closed or opened, add penalty for such nodes.

**The solution takes 1 spare bit from NodeInfo and OSMNode.** But the collected info can also be used for other node types like bollard, sump buster and etc.

## Tasklist

 - [x] Add tests

## Requirements / Relations


valhalla/sif/dynamiccost.h
```diff
@@ -1001,7 +1014,7 @@ class DynamicCost {
     // Cases with both time and penalty: country crossing, ferry, rail_ferry, gate, toll booth
     sif::Cost c;
     c += country_crossing_cost_ * (node->type() == baldr::NodeType::kBorderControl);
-    c += gate_cost_ * (node->type() == baldr::NodeType::kGate);
+    c += gate_cost_ * (node->type() == baldr::NodeType::kGate) * (!InitiallyAllowed(node));
```
genadz, line 1017: why do you use `Allowed` method instead of simple `access_indication` flag ?
```suggestion
    c += gate_cost_ * (node->type() == baldr::NodeType::kGate) * (!node->access_indication());
```

valhalla/sif/dynamiccost.h
```diff
@@ -217,6 +217,19 @@ class DynamicCost {
     return (node->access() & access_mask_) || ignore_access_;
   }
 
+  /**
+   * Checks if access information was initially specified in the data.
+   * If the node didn't have any access specifications then it can be restricted.
+   * @param   node  Pointer to node information.
+   * @return  Returns true if access was specified for the node and access is allowed.
+   */
+  virtual bool InitiallyAllowed(const baldr::NodeInfo* node) const {
```
genadz, line 226: what is the purpose of this method ? how is supposed to be used ?

genadz (changes requested):

valhalla/sif/dynamiccost.h
kevinkreiser: i'll add that we must be very careful when considering expanding the scope of the costing interface and we should consider existing semantics and algorithms that make use of those and if perhaps they already could handle the functionality we are looking to achieve

from the title and a bit of looking at the diff i see that we are basically trying keep track of whether or not access was "tagged" or "inferred" for barrier nodes. i guess the idea is that if its tagged as access yes then we dont need to penalize it. if that is the case it seems to me all we need to do is:

1. make the changes to nodeinfo to have the extra bit that says the access is tagged (i would consider naming it as such as well since we have the same concept for speeds on ways being "tagged")
2. modify base_transition_cost to check that flag on the nodeinfo and not penalize if its there

CuriOusQuOkka: The main idea was to penalise the gates with no access information and additionally the gates that have restricted access. But since the fine does not affect the restricted nodes I agree that it is clearer to use the flag directly.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
    * CHANGED: Use std::shared_ptr in case if ENABLE_THREAD_SAFE_TILE_REF_COUNT is ON. [#3067](https://github.com/valhalla/valhalla/pull/3067)
    * CHANGED: Reduce stop impact when driving in parking lots [#3051](https://github.com/valhalla/valhalla/pull/3051)
    * ADDED: Added another through route test [#3074](https://github.com/valhalla/valhalla/pull/3074)
+   * ADDED: Do not penalise gates that have allowed access [#3078](https://github.com/valhalla/valhalla/pull/3078)
```
genadz, line 25:
```suggestion
   * ADDED: Do not penalize gates that have allowed access [#3078](https://github.com/valhalla/valhalla/pull/3078)
```

genadz (dismissed):

kevinkreiser (approved):
looks good to me as well! thank you!

CHANGELOG.md
CuriOusQuOkka: fixed

genadz (approved):

gknisely (approved):

PR #3082, opened 2021-05-19
Increased kDefaultServicePenalty.

maksimandrianov:
# Issue

What's wrong?
Was found case in which, at first, driver received a 3 reroutes (as such U-turn and shortcuts) through service roads to back on the original route, before was suggested correct reroute.
For correct routing driver shouldn't receive such reroutes. This issue can be especially acute in USA due to specific road network.
This issue can't be fixed in the OSM data, because US driver rules allow make left turn to service roads in most this place.

Where were you routing from/to?
Long Island, NY, USA.

What did you expect to happen?
The driver should be suggest U-turn on the main road or reroute through higher class roads then service roads

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * **Bug Fix**
    * FIXED: Fix heading on small edge [#3114](https://github.com/valhalla/valhalla/pull/3114)
    * FIXED: Added support for `access=psv`, which disables routing on these nodes and edges unless the mode is taxi or bus [#3107](https://github.com/valhalla/valhalla/pull/3107)
+   * FIXED: Fixed U-turns through service roads [#3082](https://github.com/valhalla/valhalla/pull/3082)
 
 * **Enhancement**
 
--- a/src/sif/autocost.cc
+++ b/src/sif/autocost.cc
@@ -39,7 +39,7 @@ constexpr float kDefaultFerryCost = 300.0f;              // Seconds
 constexpr float kDefaultRailFerryCost = 300.0f;          // Seconds
 constexpr float kDefaultCountryCrossingCost = 600.0f;    // Seconds
 constexpr float kDefaultCountryCrossingPenalty = 0.0f;   // Seconds
-constexpr float kDefaultServicePenalty = 15.0f;          // Seconds
+constexpr float kDefaultServicePenalty = 75.0f;          // Seconds
 
 // Other options
 constexpr float kDefaultUseFerry = 0.5f;     // Default preference of using a ferry 0-1
--- a/test/gurka/test_service_turn.cc
+++ b/test/gurka/test_service_turn.cc
@@ -1,4 +1,5 @@
 #include "gurka.h"
+#include "test.h"
 #include <gtest/gtest.h>
 
 using namespace valhalla;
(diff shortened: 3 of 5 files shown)


maksimandrianov:
It's fixed these cases, but we need tolerate increased ETA:

| Before | After |
|-|-|
|||


| Before | After |
|-|-|
|||

maksimandrianov:
I can experiment with penalty values, that impact on U-turns, but it's more complex and can touch more cases in RAD testing. What do you think which way is better?

maksimandrianov:
Also it opened a new issue, that @SvetlanaBayarovich already had mentioned "Dead-end roundabout labels are sometimes placed incorrectly" (incorrect marks Use::kCuldesac).

maksimandrianov:
Some questions from the research:
1. **Could we penalize these sort of turns?** (cases № 2 and 4)
Imagine two alternative roads. First road is represented as one osm way with sharp corners, second - straight osm way:
photo_2021-05-24_15-34-08
Currently Valhalla engine prefers going through first road. It does not penalize these sort of turns. 
 
2. **Could we fix these cases automatically by replacing tag value "service" to "trunk_link"?** (cases № 5 and 12)
Sometimes trunk link roads are marked as service roads. For example, https://www.openstreetmap.org/way/568013065 is the service road, https://www.openstreetmap.org/way/716003464 is the trunk link. But as you can see both are trunk link:
Screenshot 2021-05-24 at 15 58 58
Screenshot 2021-05-24 at 15 59 52

kevinkreiser:
@maksimandrianov those ways are correctly tagged as service, a trunk link should be a ramp onto or off of a trunk road these small "internal" edges are correctly marked as service in my opinion. @gknisely what do you reckon?

https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtrunk_link

also please dont use google maps here. osm.org has aerial imagery from various providers. its on by default in edit mode.

maksimandrianov:
> @maksimandrianov those ways are correctly tagged as service, a trunk link should be a ramp onto or off of a trunk road these small "internal" edges are correctly marked as service in my opinion. @gknisely what do you reckon?
> 
> https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtrunk_link
> 
> also please dont use google maps here. osm.org has aerial imagery from various providers. its on by default in edit mode.

But also it is not really "service" road, as can you see in aerial imagery this road quite wide and it is used for do u-turns.

gknisely:
@maksimandrianov users mark these roads as service roads all the time; therefore, you can not penalize these like normal service roads (e.g., https://www.openstreetmap.org/way/568013065)  I am assume these are also marked as internal.  Can you test out not penalizing roads that are service and internal?

maksimandrianov:
> @maksimandrianov users mark these roads as service roads all the time; therefore, you can not penalize these like normal service roads (e.g., https://www.openstreetmap.org/way/568013065) I assume these are also marked as internal. Can you test out not penalizing roads that are service and internal?

I have tested out. RAD testing showы better results when the engine is not penalizing roads that are service and internal. Thanks!

gknisely:
@maksimandrianov looks good but can you add a gurka test where an edge is service and internal and one where the edge is service and not internal.

maksimandrianov:
> @maksimandrianov looks good but can you add a gurka test where an edge is service and internal and one where the edge is service and not internal.

Done!

gknisely (changes requested):
@maksimandrianov tests are failing and you are out of date with master

maksimandrianov:
> @maksimandrianov tests are failing and you are out of date with master

Fixed

gknisely (approved):

PR #3084, opened 2021-05-19
Change treatment of pencil point u-turns to ignore situations where two unnamed roads form the pencil-point

kshehadeh:
Fixed a problem where the pencil-point u-turn logic is triggered even when the turn transitions between two unnamed roads.  It now checks to see if the connected edges are named before checking for name consistency.

# Issue

There are situations where a pencil point u-turn is identified when it shouldn't be.  In these cases, a u-turn is formed from two unnamed roads but the two roads are treated as the same road.  They shouldn't be.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/valhalla/sif/dynamiccost.h
+++ b/valhalla/sif/dynamiccost.h
@@ -624,7 +624,7 @@ public:
         seconds += kTCUnfavorableUturn;
       // Did we make a pencil point uturn?
       else if (edge->turntype(idx) == baldr::Turn::Type::kSharpLeft && edge->edge_to_right(idx) &&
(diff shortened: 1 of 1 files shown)


gknisely (changes requested):
@kshehadeh looks good but we should add a gurka test to this [one](https://github.com/valhalla/valhalla/blob/master/test/gurka/test_avoid_pencilpoint_uturns.cc) and you need to update the changelog.

kevinkreiser:
@kshehadeh nice catch!

gknisely (approved):

mandeepsandhu (approved):

PR #3085, opened 2021-05-19
Allow_destination_only fix

ktatterso:
# Issue

Given a route with waypoints, we may switch between bidirectional A* and unidirectional A*.  (We mostly use bidir A* unless the conditions aren't favorable.)

Also, given a route with waypoints, we use a single `thor_worker_t` for determining the route from each waypoint to the next.  Note that within the `thor_worker_t` resides the `costing`.  Within the `costing` is a boolean, `allow_destination_only`.  Note that `allow_destination_only` has a default value of `true`.

Within `thor_worker_t::get_path()`, if the path algorithm is bidir A* - we immediately set `allow_destination_only=false`.  As a consequence, if the next waypoint-to-waypoint route uses unidirectional A*, `allow_destination_only` is still `false`.  I don't think this is correct.  I think the intent is that routing algorithms other than bidir A* should have `allow_destination_only` as `true`.  Bidir A* is the exception, where it first tries to find a route with `allow_destination_only=false`, and if that fails, it enters fallback logic that sets `allow_destination_only=true`.

I'm open to suggestions as to how to write a unit test for this... perhaps I could mock up a gurka test...

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/thor/route_action.cc
```diff
@@ -363,12 +363,14 @@ std::vector<std::vector<thor::PathInfo>> thor_worker_t::get_path(PathAlgorithm*
...
+  // Find the path.
   valhalla::sif::cost_ptr_t cost = mode_costing[static_cast<uint32_t>(mode)];
-  if (path_algorithm == &bidir_astar) {
-    cost->set_allow_destination_only(false);
-  }
+
+  // If bidirectional A* disable use of destination-only edges on the
+  // first pass. If there is a failure, we allow them on the second pass.
+  // Other path algorithms can use destination-only edges on the first pass.
+  cost->set_allow_destination_only(path_algorithm == &bidir_astar ? false : true);
```
kevinkreiser, line 372: good catch! for multi point routes it would carry forward the "false" setting if the first leg were bidira* and the second were regular a*.

as for a unit test i think you can exploit what makes it chose the algorithm type. specifically unidirectional is chosen for trivial routes (a route with a single edge). i guess in this case the edge must be marked as destination only as well. im not sure if a single edge route will fail due to this in a* but that is where i would start hacking in gurka. it sounds possible but id have to hack around at it before i could with confidence say it will be straight forward :smile:

ktatterso: Thanks for the suggestions, I'll give em a go.

ktatterso: I can create a gurka test geometry that successfully chooses bidir A* for the first leg and unidir A* for the second leg.  I can also specify that the second leg is "destination only".  However, `thor_worker_t::get_path()` will succeed (find a route) with and without my fix.  I can't seem to expose the relevant aspects of proving the fix, e.g., 1) that `cost->destination_only==false` for bidir A* and is `true` for unidir A*, or 2) that we avoid the fallback logic that allows destination_only.  I'll keep digging and thinking...

kevinkreiser: i have another suggestion! one of the other cases that regular a* handles is the i-have-to-go-around-the-block case. what i mean by that is say you have a 3 point route (A->B->C). We put A on some edge in the graph that is at least a few edges away from B. That gets us bidira* for the first leg. But then we put C also on the same edge B is on BUT that edge is a oneway edge and we put B "after" C on the oneway edge. So that when you continue the route from B you cant use the same edge, you have to go around the block to get back to the earlier part of the edge where C is located. perhaps this is the edge case your fix will prevent from failing?

test/gurka/test_route.cc
```diff
@@ -821,3 +821,42 @@ TEST(MultipointRoute, WithPointsOnDeadends) {
     gurka::assert::raw::expect_path(result, {"AB", "BC", "BC"});
   }
 }
+
```
ktatterso, line 824: Here's my shot at a test case.  It passes with and without my fix.  Without my fix the unidir A* leg enters the fallback logic and allows destination-only.  With my fix the unidir A* leg allows destination-only on the first pass.  I debugged everything I could think of (outside of doing something crazy like returning a new statistic) to author a case to prove my fix.

mandeepsandhu: IIUC, without your fix, the second leg (B->C) succeeds in the first pass (since it has `allow_destination_only=true` from the previous bidir astar run), but with your fix, the unidirectional astar succeeds only on the second try, correct? The number of passes is unfortunately not emitted out in the response, else we can use it to verify it in the test.

Can we make a map which has a different route if it uses a destination only edge somewhere in the middle? That way the 2 paths returned with  without the fix would be different?

Eg: in your current map, what if e introduce a destination-only edge somewhere on QP? Without your fix, it will use QP to reach C, but with your fix it should avoid it and take the edge further away (ZY).

ktatterso: @mandeepsandhu You sort of have the gist of it.  Let me clarify a bit.

Old behavior:
Leg 1:  bidir A*.  Sets `allow_destination_only=false`.  Succeeds on first pass.
Leg 2:  unidir A*.  (Because leg 1 used bidir A*, `allow_destination_only` is still `false`.). Unidir A* fails to find a route on first pass.  Enters pass 2 logic, sets `allow_destination_only=true` and succeeds in second pass.

New behavior:
Leg 1:  bidir A*.  Sets `allow_destination_only=false`.  Succeeds on first pass.
Leg 2:  unidir A*.  Sets `allow_destination_only=true`.  Succeeds on first pass.

In both cases the result is successful and produce the same result.

I'm gonna ponder your suggestion a bit now...

ktatterso: @mandeepsandhu Regarding your suggestion to have a destination edge along QP...

Having a destination edge in the middle of the route is supported but has a high transition cost so is avoided.

Even if I modify the cost of the intermediate destination edge transition cost to be 0.0, the same route is returned with/without my fix.  As mentioned above, without my fix it succeeds by entering the pass 2 logic, with my fix it succeeds in pass 1.

Good thoughts though, thanks.

ktatterso:
As to how to proceed, I see two options:
1) Recognize this fix as correct and proceed without a test.
2) Keep the simple test I wrote even though it doesn't really prove the fix.
3) Introduce a new statistic, "thor_worker_t/fallbacks", which counts the number of times we enter "pass 2" logic.  With the fix, the "thor_worker_t/fallbacks" value will be 0, without the fix it will be 1.

kevinkreiser:
@ktatterso maybe `2` is sufficient but id say disable the test because it doesnt do anything yet. there might be time in the future to do it in a different way and reenable it

kevinkreiser (approved):

test/gurka/test_route.cc
```diff
@@ -821,3 +821,88 @@ TEST(MultipointRoute, WithPointsOnDeadends) {
...
+  EXPECT_EQ(api.trip().routes(0).legs(0).node(0).edge().destination_only(), false);
+  EXPECT_EQ(api.trip().routes(0).legs(1).node(0).edge().destination_only(), true);
+
+  // These "expected_path_edge_sizes" are from the perspective of each node (7, 8, 9).
+  // Without the fix, the path-edge sizes are {2, 1, 2}. The third int is 2 because unidir A*
+  // would previously enter fallback logic to solve the second leg (8->9) and consequently
+  // add a "filtered edge". We prove we aren't entering fallback logic by asserting that the
+  // third int is 1.
+  //
+  // Honestly, this is a slightly convoluted way to test that "non-bidir A* algorithms allow
```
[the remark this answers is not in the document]
ktatterso: Right on.

kevinkreiser:
once the last build finishes ill merge this. i want to make a release shortly

PR #3090, opened 2021-05-22
Fix avoid shortcuts

### LINKED ISSUE #3087: Avoid polygon custom costs
I am trying to enforce absolute zone restrictions in my routing. Valhalla is perfect and seems to be, as far as I know, the only open source toolbox that provides this. But with some experimentation I figured that the routes still go through the 'avoid_polygon' when the cost of not doing so is high enough. It seems as if behind the scenes there is a cost associated with going through these polygons. Is there a way to specify this cost to enforce absolute restricting polygons?

nilsnolde:
fixes #3087 

Shortcuts were not included in the avoid edges set for avoid_polygons, since the rough `Allowed()` method excludes shortcuts:
https://github.com/valhalla/valhalla/blob/6d9f04ba581c0776c9b9e17ff9653d5c0a758bd8/valhalla/sif/dynamiccost.h#L233-L244

Skipping this check for shortcuts I ended up with 1562 avoid edges instead of 1476 edges before this change (though rather rural area in Austria).


test/gurka/test_avoids.cc
```diff
@@ -214,6 +216,44 @@ TEST_P(AvoidTest, TestAvoid2Polygons) {
...
+  auto node_b = avoid_map.nodes.at("B");
+  auto node_c = avoid_map.nodes.at("C");
+  auto node_e = avoid_map.nodes.at("E");
+  auto dx = std::abs(node_c.lng() - node_b.lng());
+  auto dy = std::abs(node_b.lat() - node_e.lat());
+
+  // create a small polygon on the shortcut: should fail if shortcuts are also avoided
+  //                x-x
+  //      A------B--|-|-C--F
+  //                x-x
```
nilsnolde, line 229: this does not actually work right now: it would succeed even if the shortcut was not in the avoid edges set. Any advice for building the map to make sure there's a shortcut I intersect with the tiny polygon?

kevinkreiser:
My quick thoughts on this are that loki::search should manually reject shortcuts and we should remove the shortcut check from the allowed method

nilsnolde:
right wanted to add actually if that check is even well placed in the Allowed method. I'll check who else uses it for what purpose.

kevinkreiser:
Should be only for reach and search. Both should get explicit no shortcut checks

nilsnolde:
Ok done

src/loki/reach.cc
```diff
@@ -68,7 +68,8 @@ directed_reach Reach::operator()(const DirectedEdge* edge,
   // seed the expansion with a place to start expanding from
   Clear();
   graph_tile_ptr tile, start_tile = reader.GetGraphTile(edge_id);
-  if ((tile = start_tile) && costing->Allowed(edge, tile, sif::kDisallowSimpleRestriction))
+  if ((tile = start_tile) && costing->Allowed(edge, tile, sif::kDisallowSimpleRestriction) &&
+      !edge->is_shortcut())
```
kevinkreiser, line 72: for all of these i would suggest always checking the shortcut before calling the allowed method. basically always do the cheaper thing first as it is more efficient to fail earlier.

```suggestion
  if ((tile = start_tile) && costing->Allowed(edge, tile, sif::kDisallowSimpleRestriction) &&
      !edge->is_shortcut())
```

kevinkreiser:
looks like the benchmarks are throwing `std::logic_error` somewhere.. i havent checked why yet but its possible we are forgetting something about how `Allowed` is used :smile: hopefully not though, maybe the benchmark stuff is abusing it

nilsnolde:
I got some of it, meili was using it as well in the candidate search.. still not passing, but almost there

nilsnolde:
That should be it. Benchmarks were hanging for a few mins, had to kill them, so :crossed_fingers:

src/loki/polygon_search.cc
```diff
@@ -112,9 +112,9 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Options_Ring>&
         baldr::GraphId opp_id;
 
         // bail if we wouldnt be allowed on this edge anyway (or its opposing)
-        if (!costing->Allowed(edge, tile) &&
+        if ((!costing->Allowed(edge, tile) || edge->is_shortcut()) &&
             (!(opp_id = reader.GetOpposingEdgeId(edge_id, opp_edge, opp_tile)).Is_Valid() ||
-             !costing->Allowed(opp_edge, opp_tile))) {
+             !costing->Allowed(opp_edge, opp_tile) || opp_edge->is_shortcut())) {
```
nilsnolde, line 117: temp, only to get the test to fail, which it doesn't right now, see below

test/gurka/test_avoids.cc
```diff
@@ -214,6 +217,50 @@ TEST_P(AvoidTest, TestAvoid2Polygons) {
...
+                   {node_c.lng() - 0.1 * dx, node_c.lat() + 0.01 * dy},
+                   {node_c.lng() - 0.1 * dx, node_c.lat() - 0.01 * dy},
+                   {node_b.lng() + 0.1 * dx, node_b.lat() - 0.01 * dy},
+                   {node_b.lng() + 0.1 * dx, node_b.lat() + 0.01 * dy}});
+
+  // build request manually for now
+  auto lls = {avoid_map.nodes["A"], avoid_map.nodes["G"]};
+
+  // TODO: remove before commit
+  std::cout << gurka::dump_geojson_graph(avoid_map) << std::endl;
```
nilsnolde, line 246: so, this test should print "it worked" right now but doesn't. Actually it's not a good test tbh, it wouldn't really fail if there were no shortcuts in avoid edges set.

Anyways, the intent is that we check that shortcuts are being avoided as well. Somehow the request from "A" to "G" doesn't use the shortcut (BCF) apparently. Anyone sees the problem here with the map or request?

kevinkreiser: crap i should have said something earlier. we only start using shortcuts once the route starts doing hierarchy cully that will never happen with short routes like this. i have another idea for a test though, much simpler!

simply call the `do_action` api in gurka but hold on to the API object that it returns and inspect that for your test. all you have to do is make sure that the edge id for the shortcut is in the list of avoid polygon edges. no need to check that the routing avoided it, the regression we found was that the edge wasnt making it into the set of edges to exclude in the first place not whether the routing algorithm was actually excluding them or not. does that make sense?

test/gurka/test_avoids.cc
```diff
@@ -214,6 +217,50 @@ TEST_P(AvoidTest, TestAvoid2Polygons) {
...
+  auto node_b = avoid_map.nodes.at("B");
+  auto node_c = avoid_map.nodes.at("C");
+  auto node_e = avoid_map.nodes.at("E");
+  auto dx = std::abs(node_c.lng() - node_b.lng());
+  auto dy = std::abs(node_b.lat() - node_e.lat());
+
+  // create a small polygon on the shortcut: should fail if shortcuts are also avoided
+  //                x-x
+  //      A------B--|-|-C--F--G
+  //                x-x
```
kevinkreiser, line 230: you can literally make those x's into osm nodes and use their lat lons directly from gurka to make the polygon rather than relying on hand done math below.

```suggestion
  //                h-i
  //      A------B--|-|-C--F--G
  //                j-k
```

test/gurka/test_avoids.cc
```diff
@@ -214,6 +217,50 @@ TEST_P(AvoidTest, TestAvoid2Polygons) {
...
+  rapidjson::Document doc;
+  doc.SetObject();
+  auto& allocator = doc.GetAllocator();
+  auto value = get_avoid_polys(rings, allocator);
+  auto type = "avoid_polygons";
+  auto req = build_local_req(doc, allocator, lls, GetParam(), value, type);
+
+  // make sure the right exception is thrown
+  try {
+    gurka::do_action(Options::route, avoid_map, req);
```
kevinkreiser, line 257: something approximately like:
```suggestion
    auto api = gurka::do_action(Options::route, avoid_map, req);
    size_t shortcut_count = 0;
    const auto& avoid_edges = api.options().costing_options(costing::type::auto).avoid_edges();
    for (const auto& avoid : avoid_edges){
      shortcut_count += graphreader.directededge(GraphId(avoid.id()))->shortcut();
    }
    ASSERT_TRUE(shortcut_count == 2);
    ASSERT_TRUE(avoid_edges.Size() == 4);
```

src/loki/search.cc
```diff
@@ -491,11 +491,11 @@ struct bin_handler_t {
 
       // if this edge is filtered
       const auto* edge = tile->directededge(edge_id);
-      if (!costing->Allowed(edge, tile)) {
+      if (edge->is_shortcut() || !costing->Allowed(edge, tile)) {
         // then we try its opposing edge
         edge_id = reader.GetOpposingEdgeId(edge_id, edge, tile);
         // but if we couldnt get it or its filtered too then we move on
-        if (!edge_id.Is_Valid() || !costing->Allowed(edge, tile))
+        if (edge->is_shortcut() || !edge_id.Is_Valid() || !costing->Allowed(edge, tile))
```
kevinkreiser, line 498: edge will be null if the id is invalid so we have to check the id first
```suggestion
        if (!edge_id.Is_Valid() || edge->is_shortcut() || !costing->Allowed(edge, tile))
```

test/gurka/test_avoids.cc
nilsnolde: makes a lot more sense than what I was trying to do :smile:

nilsnolde: I had to do an actual unit test here. since the route is supposed to fail there's no API object to inspect.

test/gurka/test_avoids.cc
nilsnolde: made the whole test suite so much more legible!

nilsnolde:
Integrated your suggestions, the test is much more concise now! I bet you like the diff numbers :D

test/gurka/test_avoids.cc
```diff
@@ -214,28 +186,50 @@ TEST_P(AvoidTest, TestAvoid2Polygons) {
...
-  std::vector<vm::PointLL> points{{lon, node_a.lat()}};
+  size_t found_shortcuts = 0;
+  for (const auto& edge_id : avoid_edges) {
+    if (reader.GetGraphTile(edge_id)->directededge(edge_id)) {
+      found_shortcuts++;
+      break;
+    }
+  }
+
+  ASSERT_EQ(found_shortcuts, 1);
```
kevinkreiser, line 219: maybe i forgot a detail here, dont we add the edge and its opposing edge? why isnt the count 2?

valhalla/sif/dynamiccost.h
```diff
@@ -240,7 +240,7 @@ class DynamicCost {
         ((disallow_mask & kDisallowStartRestriction) && edge->start_restriction()) ||
         ((disallow_mask & kDisallowEndRestriction) && edge->end_restriction()) ||
         ((disallow_mask & kDisallowSimpleRestriction) && edge->restrictions());
-    return !edge->is_shortcut() && accessible && !assumed_restricted;
```
kevinkreiser, line 243: you know what might have been more elegant, adding a new option to the disallow mask for shortcuts. what do you think about that @nilsnolde? if you dont have the time but you like the idea i could take a whack at it too. let me know

test/gurka/test_avoids.cc
nilsnolde: Ah thanks I wondered the same thing! It's not the avoid polygon algo, the graph has simply 1 shortcut (I printed it out before).. I figured it might be bcs the other direction is pretty pointless for a shortcut and the graphbuilder is smart enough to omit it (i.e. going from E -> C one could never use that shortcut since search locations will never project on it?!)?

kevinkreiser: that cant be possible! every edge has to have an opposing edge otherwise the graph is broken. the detail is that we only ever put one of the twin edges into the bins. we should, for both loki::search avoid edges and edges_in_ring avoided edges, always also return the opposing edge ids because if we dont then the algorithm will randomly be allowed to use one of the side of the intersecting roads and not the other.

nilsnolde: it's the same as here no?

https://github.com/valhalla/valhalla/blob/d15ae9f69968a2c2008c1b3d0cddd15783f4bef0/test/gurka/test_shortcut.cc#L260-L291

kevinkreiser: and it seems we correctly do! https://github.com/valhalla/valhalla/blob/master/src/loki/polygon_search.cc#L132-L133

valhalla/sif/dynamiccost.h
kevinkreiser: @nilsnolde that is a big feature request, we should write up an issue for it imho and start a larger discussion before we jfdi, is that ok with you? i can write one up, but i feel like others have asked for this before too...

test/gurka/test_avoids.cc
nilsnolde: what am I missing here then?

valhalla/sif/dynamiccost.h
nilsnolde: was hoping for less big. sure, that'd be nice @kevinkreiser

test/gurka/test_avoids.cc
kevinkreiser: oh sorry i didnt notice you were looking at the bins in this test, i got confused about which test it was... apologies!

valhalla/sif/dynamiccost.h
kevinkreiser: sadly "one does not simply" swap out the tilecache in a multithreaded process where its shared across threads.. thats the main eventuality that we'd have to decide how we feel would be ok to handle it

test/gurka/test_avoids.cc
nilsnolde: uff.. copy/paste fail.. there were never any shortcuts after I re-arranged the map (plus some other stupid things..).. added another node to make the shortcut.

valhalla/sif/dynamiccost.h
nilsnolde: in any case, I switched to bit masking for shortcuts as well

nilsnolde:
test fixed (2 shortcuts now) and switched to bit masking the shortcuts for the Allowed method

kevinkreiser (approved):
this looks great. i have 2 things im thinking about:

1. i wonder if we should default the mask to be `kDisalllowShortcut`. then the only change we need to make are when we want shortcuts or other special behavior. having a non empty default though seems somewhat confusing to me so probably best to leave it the way you have it
2. it would be nice if loki::search for avoid(exclude)_locations could work the same way as avoid_polygons in that it would just allow shortcuts to be found and not have that hack in loki/worker.cc where it recovers the shortcuts that go with the edges. that is non-trivial though because of the way loki::search finds only one edge pair if radius==0. we'd have to do some other work to be able to remove that shortcut logic from the worker code

PR #3093, opened 2021-05-24
Replace avoid_* with exclude_*

### LINKED ISSUE #3092: Replace `avoid_*` with `exclude_*`
The `avoid_polygons/locations` are somewhat confusingly named suggesting penalization at work, while it's really exclusion (see e.g. https://github.com/valhalla/valhalla/issues/3087#issue-897125339)

Would be worthwhile to **alias** those params with `exclude_*` equivalents.

RhnSharma:
Fixes #3092 
Hi @nilsnolde, how does this look?
Let me know if this needs any changes.
Thanks


danpat (changes requested):
The problem with this PR is that it's a breaking API change - folks who are using this feature will need to change their code in order to update to any build of Valhalla that includes this PR.

What I think this PR should do is _add_ `exclude_` that triggers the existing code.  Ideally, the `avoid_` parameter would trigger proper weighted _avoid_ behaviour, but think we could live with avoid and exclude doing the same thing for the interim until somebody gets around to implementing proper avoid behaviour.

I do not think this PR can be merged as-is.

nilsnolde:
Nice you want to take this on @RhnSharma ! I should've left some more notes on the issue, @danpat is right: the important point of the issue was actually to **alias** `avoid_polygons` with `exclude_polygons`. 

It's usually ok to rename things in the config JSON, of course it's also ok to rename the variables. Where it gets tricky:
- we'll need to make sure that the API request can still be successful with the old name as well as the new one. We did smth similar when we deprecated `auto_shorter` here: https://github.com/valhalla/valhalla/pull/2555/files#diff-dda1eacb6be34d7946468f09844f48890c1266509f634990ec27263271bd6f48
- I'm not exactly sure if it's ok to rename the proto objects.. valhalla is also used as a library and that would be a breaking change for that use case. I'd rather not rename those..

kevinkreiser:
@RhnSharma @nilsnolde yes it is ok to rename the proto objects the protobuffers dont care about the names they just change the name in the generated code, the binary format stays the same so its not a breakign change in protobuf. 

As @danpat and @nilsnolde pointed out you need to allow the code in src/worker.cc to parse either `avoid_polygons` or `exclude_polygons`. this way existing code that calls this API will still work.

proto/options.proto
```diff
@@ -130,7 +130,7 @@ message CostingOptions {
   // these are not specified directly by the user but they get filled in as the request is parsed and fulfilled
   optional Costing costing = 90;
   optional string name = 91;
-  repeated AvoidEdge avoid_edges = 92;                       // Avoid edges for any costing - derived from avoid_locations
+  repeated AvoidEdge avoid_edges = 92;                       // Avoid edges for any costing - derived from exclude_locations
```
kevinkreiser, line 133: you can rename this actual variable as well

src/worker.cc
```diff
@@ -376,8 +376,8 @@ void parse_locations(const rapidjson::Document& doc,
     locations = options.mutable_sources();
   } else if (node == "targets") {
     locations = options.mutable_targets();
-  } else if (node == "avoid_locations") {
-    locations = options.mutable_avoid_locations();
+  } else if (node == "exclude_locations") {
+    locations = options.mutable_exclude_locations();
```
kevinkreiser, line 380:
```suggestion
  } else if (node == "exclude_locations" || node == "avoid_locations") {
    locations = options.mutable_exclude_locations();
```

src/worker.cc
```diff
@@ -932,12 +932,12 @@ void from_json(rapidjson::Document& doc, Options& options) {
   parse_locations(doc, options, "targets", 132, ignore_closures);
 
   // get the avoids in there
-  parse_locations(doc, options, "avoid_locations", 133, ignore_closures);
+  parse_locations(doc, options, "exclude_locations", 133, ignore_closures);
```
kevinkreiser, line 935:
```suggestion
if (doc.HasMember("avoid_locations")
  parse_locations(doc, options, "avoid_locations", 133, ignore_closures);
else
  parse_locations(doc, options, "exclude_locations", 133, ignore_closures);
```

src/worker.cc
```diff
@@ -932,12 +932,12 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
 
   // get the avoid polygons in there
-  auto rings_req = rapidjson::get_child_optional(doc, "/avoid_polygons");
+  auto rings_req = rapidjson::get_child_optional(doc, "/exclude_polygons");
```
kevinkreiser, line 938:
```suggestion
  auto rings_req = rapidjson::get_child_optional(doc, doc.HasMember("avoid_polygons") ? "/avoid_polygons" : "/exclude_polygons");
```

RhnSharma:
@kevinkreiser Thanks for suggesting the changes. I made the changes, please take a look at it and let me know if you need any changes.

docs/api/turn-by-turn/api-reference.md
```diff
@@ -26,194 +26,199 @@ A location must include a latitude and longitude in decimal degrees. The coordin
...
...ptional filters to exclude candidate edges based on their attribution. The following exclusion filters are supported: <ul><li>`exclude_tunnel` (boolean, defaults to `false`): whether to exclude roads marked as tunnels</li><li>`exclude_bridge` (boolean, defaults to `false`): whether to exclude roads marked as bridges</li><li>`exclude_ramp` (boolean, defaults to `false`): whether to exclude link roads marked as ramps, note that some turn channels are also marked as ramps</li><li>`exclude_closures` (boolean, defaults to `true`): whether to exclude roads considered closed due to live traffic closure. **Note:** This option cannot be set if `costing_options.<costing>.ignore_closures` is also specified. An error is returned if both options are specified. **Note 2:** Ignoring closures at destination and source locations does NOT work for date_time type `0/1` & `2` respectively</li><li>`min_road_class` (string, defaults to `"service_other"`): lowest road class allowed</li><li>`max_road_class` (string, defaults to `"motorway"`): highest road class allowed</li></ul>Road classes from highest to lowest are: motorway, trunk, primary, secondary, tertiary, unclassified, residential, service_other.
```
danpat, line 48: Can you revert the changes to the documentation you've made?  None of these edits seem related to the purpose of the PR.

nilsnolde:
> they just change the name in the generated code

@kevinkreiser that's sorta what I mean though: if someone wrote code using valhalla's public PBF interface to create avoid locations/polygons, changing the names would break their code right? Not a huge deal but a bigger deal than breaking the config JSON IMHO

danpat:
@nilsnolde changing field names in protobuf is considered "binary compatible" - as long as the IDs remained the same and the meaning of the fields is unchanged, then client code generated with old protobuf interfaces should still work.

However these changes are not _source_ compatible - clients calling old generated interface functions will not compile against interfaces generated with these new protobuf changes.  It is not well defined what Valhalla considers its compatibility interface to be, but it is not unusual to exclude source-level compatibility for these kinds of reasons.

nilsnolde:
> @nilsnolde changing field names in protobuf is considered "binary compatible" - as long as the IDs remained the same and the meaning of the fields is unchanged, then client code generated with old protobuf interfaces will still be able to generate/parse protobuf binary objects successfully without anything breaking.

@danpat ah didn't know that about protobuf, thanks! that's super considerate behaivor :)

docs/api/turn-by-turn/api-reference.md
RhnSharma: Sure @danpat

kevinkreiser:
im not sure if we've documented this but this is the general conventions we follow are:

**major version change**: data incompatibilty, changes to the tiles that makes them not work with previous version soft the software
**minor version change**: this one is very poorly defined.. sometimes we've done this when we had either larger data changes (eg adding live traffic tiles) or when we add new top level APIs (eg isochrone)
**patch version change**: pretty much every thing else

We've never maintained source compatibility between any versions of the software as its just impractical and would require way way way too much planning and foresight to be feasible.

kevinkreiser:
@RhnSharma there hsould be some docs changes, specifically you should rename `avoid_locations` and `avoid_polygons` here: https://github.com/valhalla/valhalla/blob/master/docs/api/turn-by-turn/api-reference.md#other-request-options to encourage people to use the new more accurate names

kevinkreiser:
also i was thinking about this a bit more and beginning to wonder if maybe it should stay `avoid_` to allow us to add a penalty factor down the road. So that a person can soft avoid edges all the way up to hard avoidance. for that we'd add a new parameter to go with the avoids, the locations can easily be updated for that but the polygons might be harder. anyway renaming in this way doesnt technically preclude it but it would make the api more confusing if we ever did decide to do soft user provided avoids. just a thought to ask ourselves here

danpat:
>  maybe it should stay avoid_ to allow us to add a penalty factor down the road

There are use-cases for both strict "exclude" (e.g. hazmat, and you want a Noroute response if constraints can't be met) and soft "avoid" (avoid tolls if possible, but take them if costing says it's the best), so ultimately, I think we want both (and this applies to all features of this nature, including `use_highways` and `use_tolls`).

My vote here would be to do what @nilsnolde originally suggested - add `exclude_` as an _alias_ for the same feature so no APIs break, then later on, fix the `avoid_` implementation so that it's a soft avoidance, rather than the current strict one.

kevinkreiser:
@danpat i favored the alias too (it was my suggestion) but i just wanted to throw out the question here and talk myself in a circle one more time ;)

anyway this pr is looking close, whats missing i believe is the other change to the costing options proto didnt get reflected in costing where the avoid edges are used. @RhnSharma

nilsnolde:
Makes me think: what do you think about a `warnings` string array in the response generally? Warnings could be like deprecated things, e.g. using `avoid_polygons` instead of `exclude_polygons` or maybe smth like `private roads considered` when `destination` roads have to be enabled in a second pass. Anything that's not worth throwing an error but may be worth notifying the user about. Pelias has that since a few years and I find it pretty handy for an HTTP API.

kevinkreiser:
@nilsnolde yeah ive thought about it too. its nice also for those errors in the request that we smooth over without aborting the request too. eg if you set radius on a location above the max allowed we clip it and continue with the request. it would be good to tell the user we arent perfectly adhering to their request

RhnSharma:
@kevinkreiser @nilsnolde I made the changes you suggested. Please take a look at it and let me know if you need any changes.
Thanks

nilsnolde (changes requested):
@RhnSharma you're missing a few spots here where you didn't change the name to `exclude_*`, e.g. look at the CI logs. Before you push next time, can you pls make sure:
- you can build the project successfully with default cmake options
- you can build `run-gurka_avoids`, e.g. `make -C build run-gurka_avoids`

When CI is happier (i.e. at least the build step runs through) we can review further.

bench/meili/config.json
```diff
@@ -29,9 +29,7 @@
     }
   },
   "skadi": {
-    "actons": [
-      "height"
-    ],
+    "actons": ["height"],
```
nilsnolde, line 32: unrelated change.. actually the key should be `actions`. not your fault probably, but could you change it to `actions`?

src/loki/worker.cc
```diff
@@ -84,9 +84,9 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
...
   } catch (const std::runtime_error&) { throw valhalla_exception_t{125, "'" + costing_str + "'"}; }
 
-  if (options.avoid_polygons_size()) {
+  if (options.exclude_polygons_size()) {
     const auto edges =
-        edges_in_rings(options.avoid_polygons(), *reader, costing, max_avoid_polygons_length);
+        edges_in_rings(options.exclude_polygons(), *reader, costing, max_exclude_polygons_length);
     auto* co = options.mutable_costing_options(options.costing());
     for (const auto& edge_id : edges) {
       auto* avoid = co->add_avoid_edges();
```
nilsnolde, line 92: this is what CI was tripping over

bench/meili/config.json
RhnSharma: Sure @nilsnolde

nilsnolde: actually it seems like the whole `skadi` member is deprecated, it's not being generated in the config anymore.. you can take it out completely or not even change it, up to you I think

nilsnolde:
@RhnSharma are you still planning to finish or should I carry this over the review line? In any case, thanks for the effort already put in!

nilsnolde:
I quickly finished this. thanks a lot @RhnSharma , was just tiny stuff left

nilsnolde (approved):

kevinkreiser (approved):

nilsnolde:
btw @danduk82 &@ismailsunni, this should be relevant for you too.

ismailsunni:
Thanks @nilsnolde

PR #3101, opened 2021-05-25
Taginfo updates

gknisely:
# Issue

Updated PR template to included taginfo.json.  Made sure taginfo.json is up 2 date.


kevinkreiser (dismissed):
sadly i am very skeptical that the PR template will keep us on the happy path of updating this but the only other option is a CI check/warning that says "ah ha! you updated lua or graphparser but i see no updates in taginfo.json"

gknisely:
> sadly i am very skeptical that the PR template will keep us on the happy path of updating this but the only other option is a CI check/warning that says "ah ha! you updated lua or graphparser but i see no updates in taginfo.json"

@kevinkreiser I thought the same, but hope this will be enough :pray:

mandeepsandhu (approved):

PR #3106, opened 2021-05-26
nit: Enables compiler warnings in thor module for single source files

merkispavel:
# Issue

- major: replaced tuple with struct `MapMatchResult` https://github.com/valhalla/valhalla/pull/3106/files#r639994553
- minor: removed a few unused variables & function

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/valhalla/meili/match_result.h
+++ b/valhalla/meili/match_result.h
@@ -66,6 +66,12 @@ struct MatchResult {
   }
 };
 
+struct MapMatchResult {
+  float confidence_score = 0.f;
+  float raw_score = 0.f;
+  std::vector<MatchResult> match_results;
+};
+
 struct EdgeSegment {
   EdgeSegment(baldr::GraphId the_edgeid,
               double the_source = 0.f,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,8 @@
    * ADDED: Adds incident-length to metadata proto [#3083](https://github.com/valhalla/valhalla/pull/3083)
    * ADDED: Do not penalize gates that have allowed access [#3078](https://github.com/valhalla/valhalla/pull/3078)
    * ADDED: Added missing k/v pairs to taginfo.json.  Updated PR template. [#3101](https://github.com/valhalla/valhalla/pull/3101)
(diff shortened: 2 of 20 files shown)


valhalla/meili/match_result.h
```diff
@@ -66,6 +66,12 @@ struct MatchResult {
   }
 };
 
+struct MapMatchResult {
+  float confidence_score = 0.f;
+  float raw_score = 0.f;
+  std::vector<MatchResult> match_results;
+};
+
```
merkispavel, line 74: The most important change is replacing tuple with a new struct `MapMatchResult`. Warnings as errors pointed out that a few constants(i.e `kRawScoreIndex, kRawScoreIndex, kMatchResultsIndex) were unused. + These constants a declared in multiple files. So replacing tuple with a struct with named members looks like a good way to go

merkispavel: Q: What's the right file/module to declare this struct? I think of moving it to the thor module but not sure what's the right file for this purpose

kevinkreiser: yeah its a bit of a mess to be honest. in the first implementation where thor exposed things from meili to the outward facing api there was a lot of duplication of objects. recently we reduced some of that but there is still this...

i think its a bit confusing to name this object like this because we already have these: https://github.com/valhalla/valhalla/blob/master/valhalla/meili/match_result.h#L138-L141

maybe we just add the addtional score to that existing matchresult object and use that?

mandeepsandhu:
@merkispavel are these changes being tested on Linux instances? valhalla currently does not build with `-Werror` on Linux (the mjolnir module still has warnings in some files which need to be fixed before we can enable `Werror` on it).

Our current setup disables `-Werror` on CI and thats why it didn't show up in and of the CI worflows (I think this flag does not get set on macos as well).

mandeepsandhu:
> @merkispavel are these changes being tested on Linux instances? valhalla currently does not build with `-Werror` on Linux (the mjolnir module still has warnings in some files which need to be fixed before we can enable `Werror` on it).
> 
> Our current setup disables `-Werror` on CI and thats why it didn't show up in and of the CI worflows (I think this flag does not get set on macos as well).

Okay, so the reason the warnings were showing up for me was because I had set `LOGGING_LEVEL_ERROR` and some of the unused variables were only used with INFO log level, and thats why it showed up for me. Since default is INFO, it doesn't show up in CI.

PR #3111, opened 2021-05-27
Fix linux build failures with -Werror enabled in mjolnir

mandeepsandhu:
Some files in mjolnir still have warnings which cause build failures
when -Werror is enabled.

Removed affected files from having -Werror enabled.

### DIFF
--- a/src/mjolnir/CMakeLists.txt
+++ b/src/mjolnir/CMakeLists.txt
@@ -26,10 +26,8 @@ set(sources
   ${CMAKE_CURRENT_BINARY_DIR}/admin_lua_proc.h
   admin.cc
   adminbuilder.cc
-  bssbuilder.cc
   complexrestrictionbuilder.cc
   countryaccess.cc
-  dataquality.cc
   directededgebuilder.cc
   graphtilebuilder.cc
   edgeinfobuilder.cc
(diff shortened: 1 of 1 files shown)


ktatterso (approved):

merkispavel:
LGTM as a hot fix
@mandeepsandhu do you have ubuntu 20 locally(or other version but different from the CI one which is 18.04)?
Q: Isn't it a problem that ci doesn't cover environment that is used for development? I know there's infinite number of environments, just a question to the audience

merkispavel (commented):
Sorry for that @mandeepsandhu. I tested those changes locally on mac os only(+ ci checks)
LGTM https://github.com/valhalla/valhalla/pull/3111#issuecomment-849587157

kevinkreiser (approved):

mandeepsandhu:
> LGTM as a hot fix
> @mandeepsandhu do you have ubuntu 20 locally(or other version but different from the CI one which is 18.04)?
> Q: Isn't it a problem that ci doesn't cover environment that is used for development? I know there's infinite number of environments, just a question to the audience

@merkispavel yeah I run on ubuntu 20.04 with gcc 9.3.0. I'm surprised that this didn't showup in CI since the per-file Werror should've tripped the build. Anyway I'll merge this for now.

Issue #3102, opened 2021-05-26
Refactor costing options to handle common stuff in a one place


genadz:
It's easy to notice that different costing options have a lot of common parameters (for example, look at this section https://github.com/valhalla/valhalla/blob/master/src/sif/autocost.cc#L31-L42). Moreover, most of the common parameters have the same default values. Despite of the fact, we have to:
- parse the same costing option for each costing type;
- declare default value for each costing option even if it's the same.

So, each time when you try to introduce a new parameter you should copy-paste the same code for each costing option. It's not very convenient and should be refactored.

kevinkreiser:
agree, i have been wanting to do this forever and just havent had the time. when i add a new thing, i try to put it just inside of dynamiccost and let the derived class use it from there. im sure there are some gotchas but id love to put all the parsing into the parent class and when we need to somehow have a way for the child class to customize. at the moment we are using some static functions though so we might want to make them non static so that we could get that specialization from the child class if we ever need to. but maybe its a problem we actually dont need to worry about yet, i need to refresh my memory

genadz:
> agree, i have been wanting to do this forever and just havent had the time. when i add a new thing, i try to put it just inside of dynamiccost and let the derived class use it from there. im sure there are some gotchas but id love to put all the parsing into the parent class and when we need to somehow have a way for the child class to customize. at the moment we are using some static functions though so we might want to make them non static so that we could get that specialization from the child class if we ever need to. but maybe its a problem we actually dont need to worry about yet, i need to refresh my memory

As for me it's two different tasks:
1) move costing options parsing to one place;
2) move common parameters from all derived classes to the base one (`DynamicCost`).

This issue is more about the 1st point. Here I want to generalize parsing and setting default values.

genadz:
rough code snippet
```c++
// dynamiccost.h

struct SharedCostingOptions {
  SharedCostingOptions();
  // define shared options
  float opt1;
  float opt2;
  ...
};

const SharedCostingOptions kDefaultSharedOptions{};

void ParseSharedCostOptions(const rapidjson::Value& obj, CostingOptions* pbf_costing_options,
                            const SharedCostingOptions& defaults);

// dynamiccost.cc

SharedCostingOptions::SharedCostingOptions() {
  // set default values
  opt1 = default_opt_1;
  opt2 = default_opt_2;
  ...
}

void ParseSharedCostOptions(const rapidjson::Value& obj, CostingOptions* pbf_costing_options,
[...]

PR #3116, opened 2021-05-28
Added forgotten penalties for kLivingStreet and kTrack.

maksimandrianov:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * **Bug Fix**
    * FIXED: Fix heading on small edge [#3114](https://github.com/valhalla/valhalla/pull/3114)
    * FIXED: Added support for `access=psv`, which disables routing on these nodes and edges unless the mode is taxi or bus [#3107](https://github.com/valhalla/valhalla/pull/3107)
+   * FIXED: Added forgotten penalties for kLivingStreet and kTrack for pedestrian costing model [#3116](https://github.com/valhalla/valhalla/pull/3116)
 
 * **Enhancement**
 
--- a/src/sif/pedestriancost.cc
+++ b/src/sif/pedestriancost.cc
@@ -484,6 +484,10 @@ public:
     c.cost +=
         maneuver_penalty_ * (!edge->link() && edge->use() != Use::kEgressConnection &&
                              edge->use() != Use::kPlatformConnection && !edge->name_consistency(idx));
+    c.cost += living_street_penalty_ *
+              (edge->use() == baldr::Use::kLivingStreet && pred->use() != baldr::Use::kLivingStreet);
+    c.cost +=
+        track_penalty_ * (edge->use() == baldr::Use::kTrack && pred->use() != baldr::Use::kTrack);
     c.cost += service_penalty_ *
               (edge->use() == baldr::Use::kServiceRoad && pred->use() != baldr::Use::kServiceRoad);
 


src/sif/pedestriancost.cc
```diff
@@ -484,6 +517,10 @@ class PedestrianCost : public DynamicCost {
     c.cost +=
         maneuver_penalty_ * (!edge->link() && edge->use() != Use::kEgressConnection &&
                              edge->use() != Use::kPlatformConnection && !edge->name_consistency(idx));
+    c.cost += living_street_penalty_ *
+              (edge->use() == baldr::Use::kLivingStreet && pred->use() != baldr::Use::kLivingStreet);
+    c.cost +=
+        track_penalty_ * (edge->use() == baldr::Use::kTrack && pred->use() != baldr::Use::kTrack);
```
merkispavel, line 523: As I see this is the only change around logic itself and everything else is renaming, right? Maybe we can leave naming as it is and apply only these cost-specific changes?

maksimandrianov: 1. Fixed incorrect method overriding. (35925a3bef82018e39157959485f26f77cf88bf6)
@merkispavel There is the incorrect function overriding!
Example:
```
DynamicCost *ptr = new PedestrianCost(...)
ptr->base_transition_cost(...)  
```
DynamicCost::base_transition_cost was called before these changes, and PedestrianCost::base_transition_cost after.

2. Renamed base_transition_cost -> transition_cost. (174a4ac541dc10db0930dd74ad96a1e19a71d687)
@mandeepsandhu may be we should add prefix `base_` in each base class method? It's always the common something applicable for all derived classes:)
Why is PedestrianCost::base_transition_cost named as base_transition_cost?

Why shouldn't these changes be merged into the master?

SvetlanaBayarovich: Hmmm, to me 'base' in base_transition_cost sounded more like 'common', something that we start with. And having it as a function in some base class is only a way to implement it.

Also, with the naming suggested in this pr we get methods named `Cost::TransitionCost` and `Cost::transition_cost` at the same time, which may be more confusing

merkispavel: > Also, with the naming suggested in this pr we get methods named Cost::TransitionCost and Cost::transition_cost at the same time, which may be more confusing

agreed, not the best combo 😕

merkispavel: > Fixed incorrect method overriding. (35925a3)
@merkispavel There is the incorrect function overriding!
Example:
DynamicCost *ptr = new PedestrianCost(...)
ptr->base_transition_cost(...) 

As far as I see `base_transition_cost` is a **protected** method so I'm not sure the listed example is possible

kevinkreiser: @dnesbitt61 i believe you are the original author here. can you let us know your thoughts?

dnesbitt61: I likely only ever intended this to be called from within the derived class's TransitionCost methods.

kevinkreiser (changes requested):
my assumption all this time was the same as @SvetlanaBayarovich 's that we called it `base` because its a starting point and other costings add to afterwards. i suggest that we revert this part of the PR and focus on the missing penalties and adding tests for those

dnesbitt61 (commented):
I would leave the naming as is. Each derived TransitionCost method can call base_transition_cost to get all transition costs that are common to all costing methods - then add transition costs that are different for the derived class.

I hesitate to put too many things into the base transition cost though as this will lead to decreased performance. For example (and this isn't part of this PR), why is bikeshare cost part of the base_transition_cost? This only applies to a very limited set of costing classes.

maksimandrianov:
Folks, thanks for comments. Commits with refactoring were dropped.

merkispavel (approved):
Thank you!

genadz (approved):

kevinkreiser:
@maksimandrianov thank you for removing the unrelated changes. now i see that what we are really working on is adding support for penalizing livingstreet and track edges, which was present in vehicular modes of travel, but not in pedestrian. my problem is that this is very confusing to me. i would think that most pedestrians would, if they cant use a pedestrian only edge, prefer to use a living street or a track edge above other types of edges. because of that, it does not make sense for them to have a default penalty applied to these types of edges. i agree that they should be able to avoid them if they really wanted but the default being the same as vehicular modes of travel makes no sense to me at all. to me the default penalty for these types of edge uses for pedestrian would be correctly set at `0`

@genadz @merkispavel what am i missing here?

merkispavel:
my bad @kevinkreiser. I think I auto-approved without diving deeply into the logic of the change: just shallow argument "we do this for autocost, let's do the same for pedestrian" did work for me. I'll take one more round of thinking ^ 🤔 
bringing @SvetlanaBayarovich here as she should be better to know the motivation

kevinkreiser:
@merkispavel no worries, i just think that by default pedestrian probably should favor living and tracks or at least not penalize them compared to other higher class roads. i understand the motivation to have the ability to control this behavior for pedestrian like we do for the other costings though so if we want to add that that makes sense to me. @genadz worked on this stuff too so i expect he will have an opinion here too!

danpat:
It seems to me that the _math_ for calculating the cost belongs in DynamicCost and is common to all costings.  But the factors should default to different values for different costings.

kevinkreiser:
@danpat yep that is something that @genadz said he was going to work on at some point. he made a recent issue that encompassed some of that at least in my opinion: https://github.com/valhalla/valhalla/issues/3102

SvetlanaBayarovich:
@kevinkreiser Currently the math for penalties is https://github.com/valhalla/valhalla/blob/ae987bf50a26ac506b3883d1fcaec860072bfd8b/src/sif/dynamiccost.cc#L268-L269

So if we favor living streets (factor is >= 0), the penalty becomes 0. And for pedestrians the default value is just like that https://github.com/valhalla/valhalla/blob/ae987bf50a26ac506b3883d1fcaec860072bfd8b/src/sif/pedestriancost.cc#L64

SvetlanaBayarovich:
To sum up, default `use` values are https://github.com/valhalla/valhalla/search?q=%22float+kDefaultUseLivingStreets%22

SvetlanaBayarovich:
Sorry, closed by mistake

SvetlanaBayarovich (approved):

SvetlanaBayarovich (approved):

kevinkreiser (dismissed):

kevinkreiser:
@genadz @SvetlanaBayarovich thanks for refreshing my memory on this. i for that that the per-cost defaults came in through those range functors in the parsing and were overridden.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
    * FIXED: Fix heading on small edge [#3114](https://github.com/valhalla/valhalla/pull/3114)
    * FIXED: Added support for `access=psv`, which disables routing on these nodes and edges unless the mode is taxi or bus [#3107](https://github.com/valhalla/valhalla/pull/3107)
    * FIXED: Fixed U-turns through service roads [#3082](https://github.com/valhalla/valhalla/pull/3082)
+   * FIXED: Added forgotten penalties for kLivingStreet and kTrack for pedestrian costing model [#3116](https://github.com/valhalla/valhalla/pull/3116)
```
kevinkreiser, line 7: one thing this pr didnt do is check what happens in RAD now that these costs are added. for tracks i suppose it made no difference because the value was set to .5 by default but the default of .6 for living streets should have some impact in some areas.

kevinkreiser: i think its ok to merge anyway though as this was clearly an oversight. if the pedestrian cost needs tuning we can do that in a separate pr. anyone else have a different opinion or some time to check RAD?

kevinkreiser (approved):

maksimandrianov:
RAD tests passed successfully

CHANGELOG.md
SvetlanaBayarovich: @kevinkreiser .6 for living streets wasn't changed in this PR. It's been this way since the day I introduced this feature.

kevinkreiser: right i get that but it wasnt being used before in pedestrian transition costs which means we do have a potential for pedestrian paths to change right?

SvetlanaBayarovich: Uhm, sorry, I'm slow, not getting what you are concerned about... .6 means zero penalty, so no change was expected now regarding changes in transition cost

kevinkreiser: no you arent slow, i see it now. the penalty is set to 0 in the ternary in dynamiccost.cc when its >= to .5. for some reason i was looking at the factor rather than the penalty. sorry about that!

Issue #3071, opened 2021-05-11
Unzipped cache in skadi


molind:
I've tried to build tiles using compressed elevation data and found that 
unzipped_cache inside skadi sample used without any locks between different threads.

And because it caches only 1 tile it's terribly slow in `add_elevation` task. But lack of speed is not the worse issue. Pointer from `unzipped_cache` passed to multiple inflate functions at the same time. https://github.com/valhalla/valhalla/blob/master/src/skadi/sample.cc#L151-L156

 https://github.com/valhalla/valhalla/blob/master/src/baldr/compression_utils.cc#L102-L111 there I've added a log function to check what happens.
```
      // several fatal errors to worry about
      code = inflate(&stream, flush);
      switch (code) {
        case Z_STREAM_ERROR:
        case Z_NEED_DICT:
        case Z_DATA_ERROR:
        case Z_MEM_ERROR:
          inflateEnd(&stream);
          LOG_ERROR(std::string("inflate Error: ") + std::to_string(code) + ", " + stream.msg);
          return false;
      }
```
It prints "inflate Error: -3, incorrect data check".

Same issue happens while reading data from "cache" without any sync primitives. After unzipped_cache.first check unzipped_cache.second could be easily changed by other thread.
https://github.com/valhalla/valhalla/blob/master/src/skadi/sample.cc#L139-L142

P.S. I'd suggest to use lz4, instead of gzip. Lz4 provides ~ the same compression ratio while decompression speed is 4-5 times faster.

molind:
When separate buffer is used by each thread it works without errors. I know it doesn't return anything just makes sure it could inflate elevation tiles.

```diff
diff --git a/src/skadi/sample.cc b/src/skadi/sample.cc
index 641c0d8..6350b3f 100644
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -147,25 +147,34 @@ const int16_t* sample::source(uint16_t index) const {
     s.avail_in = static_cast<unsigned int>(mapped.second.size());
   };
 
+  std::vector<int16_t> out_buffer;
+  out_buffer.resize(HGT_PIXELS);
+
   // for setting where to write the uncompressed data to
-  auto dst_func = [this](z_stream& s) -> int {
-    unzipped_cache.second.resize(HGT_PIXELS);
-    s.next_out = static_cast<Byte*>(static_cast<void*>(unzipped_cache.second.data()));
[...]

PR #3120, opened 2021-06-01
Cache for gzipped elevation tiles

molind:
# Issue

As discussed in #3071 current implementation don't work correctly with gzipped elevation tiles. This is basic implementation of cache with parallel tile decoding support. I've used unpacked tile limit = 50. But it could be tuned from config or any other way you want.

Please review and let me know what should I improve.


valhalla/skadi/sample.h
```diff
@@ -56,30 +63,81 @@ class sample {
...
+      return rv;
+    }
+
+    bool unpack(const char* unpacked);
+    static boost::optional<std::pair<uint16_t, sample::format_t>>
+    parse_hgt_name(const std::string& name);
+  };
 
   // using memory maps
-  mutable std::vector<std::pair<format_t, midgard::mem_map<char>>> mapped_cache;
```
kevinkreiser, line 77: so this is the main question to me. completely removing memory mapping will have some performance impact. i dont care as much about peformance impact for offline processing like that we have in `valhalla_build_tiles` but when we run the service and somehow does a `/height` request it needs to be super fast (as it was before). with a naive caching scheme we will not be able to beat what the OS is doing with memory mapping but we'd have to test it to find out. indeed i know that memory mapping many small files can be slower than just reading them from disk and in that scenario its faster to have an in memory cache like we do for graph tiles but there we are talking about 10s of kilobytes and here we are talking about tiles that are 25 megabytes in size. anyway we need to consider how this change affects the service. it could be for the better (ie we load up the 300GB of compressed elevation data and now its fast enough to be usable but still not need a 1.6TB disk). i just know from 3 years ago when i implemented lz4 it ended up not being fast enough performance wise though i will say i dont remember what my caching looked like (though i guess git history would tell us).

molind:
It runs right the same way as earlier version. `mapped_cache` is now called just `cache`. 
https://github.com/valhalla/valhalla/blob/191760c52b1c9dfb03a8bbb87f7ba2cc971a923f/valhalla/skadi/sample.h#L137
To store all required information `cache_item_t` was added.
https://github.com/valhalla/valhalla/blob/191760c52b1c9dfb03a8bbb87f7ba2cc971a923f/valhalla/skadi/sample.h#L97-L134
I've tried to keep workflow identical to current one without any useless thread synchronisation when all data stored in RAW files. `cache` filled in `sample` constructor.
https://github.com/valhalla/valhalla/blob/191760c52b1c9dfb03a8bbb87f7ba2cc971a923f/src/skadi/sample.cc#L218-L228
And `cache_item_t::init` maps the file. No matter it's RAW or GZIP.
https://github.com/valhalla/valhalla/blob/191760c52b1c9dfb03a8bbb87f7ba2cc971a923f/src/skadi/sample.cc#L85-L93
When coordinate requested in `get` or `get_all` function they call `source` function.
https://github.com/valhalla/valhalla/blob/191760c52b1c9dfb03a8bbb87f7ba2cc971a923f/src/skadi/sample.cc#L240-L260
It works the same way as earlier version. Checks tile inside the cache. If there is no mapped data inside `cache_item` it tries to lazy load hgt file. This part don't use any locks and it should work at the same speeds as version in `master`.
Magic happens later in source function:
https://github.com/valhalla/valhalla/blob/191760c52b1c9dfb03a8bbb87f7ba2cc971a923f/src/skadi/sample.cc#L262-L304
It checks current tile in `pending_tiles`  map. It stores future objects of `tile_data` class. Main goal of `tile_data` is to increment and decrement reference counter for `unpacked` buffers. We don't want to unpack everything at once. And it tries to keep count of `unpacked` buffers under 50 (`UNPACKED_TILES_COUNT `).
If there is no pending job it checks unpacked data inside the cache and returns it. If there is no unpacked data in `cache_item_t` it allocates memory and releases the lock before unpack. Because of this different gzipped tiles could be unpacked in parallel. If requests for same tile called from different threads, they'll wait in future.get();

valhalla/skadi/sample.h
```diff
@@ -56,30 +63,81 @@ class sample {
...
-
-  // TODO: make an LRU
-  using unzipped_t = std::pair<int16_t, std::vector<int16_t>>;
-  mutable unzipped_t unzipped_cache;
-
+  std::vector<cache_item_t> cache;
+  std::set<uint16_t> reusable;
+  std::map<uint16_t, std::shared_future<tile_data>> pending_tiles;
```
kevinkreiser, line 139: why not use the `unordered_` variants of these data structures as they are almost always faster at the cost of higher memory usage

molind: Good point. I’ll change it.

kevinkreiser:
@molind i have the data downloaded and should be able to test it out over the weekend. thanks for your patience

molind:
I’ve generated routing data for a whole planet this week with packed elevation data. Everything worked fine. Waiting for your test results.

roadrouting:
Sorry for the off-topic comment, but I've been meaning to ask: roughly how long should a full planet build with elevation data take?

Our initial experiments have been taking 3-4 days on a very overprovisioned machine. From what we've seen in issues and comments, this feels way too long, and we suspect there is a slow disk causing problems.

dnesbitt61:
Last planet build I did (with elevation) was about 36 hours. Make sure you use an SSD - I remember trying on a laptop with a hard drive and it was agonizingly slow.

kevinkreiser:
in my experience on amazon infra (and on my personal machine) it takes about 24 hours. these are machines with SSDs and 24-32 threads and something like 32-64GB of ram.

kevinkreiser:
@molind im sorry again i just havent found the time... i just wanted to let you konw i didnt forget, i will review it!

test/sample.cc
```diff
@@ -110,11 +110,10 @@ struct testable_sample_t : public skadi::sample {
       // first row is 1's second is 2's
       for (size_t i = 0; i < 3601 * 2; ++i)
         s.push_back(((i / 3601 + 1) & 0xFF) << 8);
-      for (size_t i = 0; i < 3601 * 4; ++i)
+      for (size_t i = 0; i < 3601 * (3601 - 2); ++i)
```
kevinkreiser, line 113: ok i see we have to make the full thing or else it will error now

src/skadi/sample.cc
```diff
@@ -241,11 +200,155 @@ template <class coord_t> double sample::get(const coord_t& coord) const {
   return value / adjust;
 }
 
-template <class coords_t> std::vector<double> sample::get_all(const coords_t& coords) const {
+::valhalla::skadi::sample::sample(const std::string& data_source) : data_source(data_source) {
```
kevinkreiser, line 203: i know it was like this before but do you mind removing the namespacing here since its not needed

valhalla/skadi/sample.h
```diff
@@ -56,30 +62,81 @@ class sample {
...
-  mutable std::vector<std::pair<format_t, midgard::mem_map<char>>> mapped_cache;
-
-  // TODO: make an LRU
-  using unzipped_t = std::pair<int16_t, std::vector<int16_t>>;
-  mutable unzipped_t unzipped_cache;
-
+  std::vector<cache_item_t> cache;
+  std::unordered_set<uint16_t> reusable;
+  std::unordered_map<uint16_t, std::shared_future<tile_data>> pending_tiles;
+  std::recursive_mutex mutex;
```
kevinkreiser, line 139: @molind what do you think about using the pimpl idiom here to encapsulate all of the cache, basically:

```c++
// in sample.h we simple forward declare a generic cache inside the sample class
class sample {
 ...
 protected:
  class cache_t;
  std::unique_ptr<cache_t> cache;
};

//then in sample.cc we move all of the details about anything the cache is doing or needs
struct tile_data_t {
  ...
};
struct cache_item_t {
  ...
};
struct cache_t {
  ...
};
```

I would like to basically keep the public interface to the sample class as free of details as possible. i know its a little bit more work but i think its better for the end user of the class. maybe i have totally overlooked a detail about why we cant do this but i think the top level methods of sample:: are just the constructor and the couple methods to get samples from the data.

let me know what you think!

src/skadi/sample.cc
```diff
@@ -241,11 +200,155 @@ template <class coord_t> double sample::get(const coord_t& coord) const {
...
+  }
+
+  // if we dont have anything maybe its lazy loaded
+  auto& item = cache[index];
+  if (item.get_data() == nullptr) {
+    auto f = data_source + get_hgt_file_name(index);
+    item.init(f, format_t::RAW);
+  }
+
+  if (item.get_format() == format_t::UNKNOWN) {
```
kevinkreiser, line 253:
```suggestion
  // it wasnt in cache and when we tried to load it the file was of unknown type
  if (item.get_format() == format_t::UNKNOWN) {
```

src/skadi/sample.cc
```diff
@@ -241,11 +200,155 @@ template <class coord_t> double sample::get(const coord_t& coord) const {
...
+    return tile_data(this, index, false, nullptr);
+  }
+
+  // we have it raw or we dont
+  if (item.get_format() == format_t::RAW) {
+    return tile_data(this, index, false, (const int16_t*)item.get_data());
+  }
+
+  mutex.lock();
```
kevinkreiser, line 262:
```suggestion
  // we were able to load it but the format wasnt RAW, which only leaves the GZIP format
  // now we've got to schedule or wait for some gunzipping to get the raw bytes into cache
  mutex.lock();
```

kevinkreiser:
@molind ive added a couple trivial suggestions and one bigger refactor type suggestion. please have a look and let me know what you think. other than those things, this pr looks perfectly fine to me thank you for clearing that up!

molind:
Great. I'll change it according to the code review.

BTW, Generated routing data for planet on EX52-NVME from Hetzner and it took just 2 hours to add elevation info.  
```
2021/07/21 15:02:53.166361 [INFO] Adding elevation to 181525 tiles with 12 threads...
2021/07/21 17:11:14.106379 [INFO] Finished
```

molind:
@kevinkreiser Refactored as you suggested.

src/skadi/sample.cc
```diff
@@ -10,13 +10,15 @@
...
-
 #include "baldr/compression_utils.h"
 #include "filesystem.h"
 #include "midgard/logging.h"
 #include "midgard/pointll.h"
 
+namespace valhalla {
+namespace skadi {}
+} // namespace valhalla
+
```
kevinkreiser, line 21:
```suggestion
```

src/skadi/sample.cc
```diff
@@ -80,103 +68,342 @@ uint64_t file_size(const std::string& file_name) {
...
+#endif
+  tile_data(cache_t* c, uint16_t index, bool reusable, const int16_t* data)
+      : c(c), data(data), index(index), reusable(reusable) {
+    if (reusable) {
+      c->increment_usages(index);
+    }
+  }
+
+  tile_data(const tile_data& other) {
+    this->operator=(other);
```
kevinkreiser, line 205:
```suggestion
    *this = other;
```

src/skadi/sample.cc
```diff
@@ -80,103 +68,342 @@ uint64_t file_size(const std::string& file_name) {
...
+    // if we were missing some we need to adjust by that
+    return value / adjust;
+  }
+};
+
+tile_data cache_t::source(uint16_t index) {
   // bail if its out of bounds
   if (index >= TILE_COUNT) {
-    return nullptr;
+    return tile_data(this, index, false, nullptr);
```
kevinkreiser, line 298: why not make a default constructor, instead of just having one for windows, that initializes the object in this way? in the case that its an invalid entry, why n ot set the cache parent to null and also set the index to some crazy value, so that every bit of the object is telling the caller this object is invalid? apologies in advance if there is an obvious reason that i didnt see

kevinkreiser:
refactor looks great, i like how narrow the api in the header file remains even t hough  you added all this functionality. i had a couple of nitpicks and questions, but nothign major!

src/skadi/sample.cc
molind: Makes sense. There is a bool operator, to check tile state. No need to check for crazy values. :)

molind:
@kevinkreiser Done.

valhalla/skadi/sample.h
```diff
@@ -2,23 +2,31 @@
 #define __VALHALLA_SAMPLE_H__
 
 #include <cstdint>
+#include <functional>
+#include <future>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
```
kevinkreiser, line 11: nitpick: i think most of these changes can move to the cc file

molind: Done

kevinkreiser (approved):
thank you so much for bearing with me during the review process

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
    * REMOVED: Unused overloads of `to_response` function [#3167](https://github.com/valhalla/valhalla/pull/3167)
 
 * **Bug Fix**
+   * FIXED: Cache for gzipped elevation tiles [#3120](https://github.com/valhalla/valhalla/pull/3120)
```
kevinkreiser, line 40: this is in the wrong section, since we've release since you prd this. i will fix it manually after merge

src/skadi/sample.cc
```diff
@@ -80,103 +71,347 @@ uint64_t file_size(const std::string& file_name) {
...
+class tile_data;
+
+struct cache_t {
+  // Cached tiles
+  std::vector<cache_item_t> cache;
+  // Set of reusable tile indexes
+  std::unordered_set<uint16_t> reusable;
+  // Map of pending tiles. No matter how many requests received, only one inflate job per tile
+  // started.
+  std::unordered_map<uint16_t, std::shared_future<tile_data>> pending_tiles;
```
PureW, line 173: Flagging here that this fails to compile in newest gcc and clang:

Gcc
```
/usr/include/c++/11.1.0/bits/unordered_map.h:105:18:   required from ‘class std::unordered_map<short unsigned int, std::shared_future<valhalla::skadi::tile_data> >’
/home/anders/Documents/Programming/mapbox/valhalla/src/skadi/sample.cc:173:63:   required from here
/usr/include/c++/11.1.0/future:906:21: error: static assertion failed: result type must be destructible
  906 |       static_assert(is_destructible<_Res>{},
      |                     ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/11.1.0/future:906:21: note: ‘std::is_destructible<valhalla::skadi::tile_data>()’ evaluates to false
/usr/include/c++/11.1.0/future: In instantiation of ‘class std::future<valhalla::skadi::tile_data>’:
/home/anders/Documents/Programming/mapbox/valhalla/src/skadi/sample.cc:329:23:   required from here
/usr/include/c++/11.1.0/future:769:21: error: static assertion failed: result type must be destructible
  769 |       static_assert(is_destructible<_Res>{},
      |                     ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/11.1.0/future:769:21: note: ‘std::is_destructible<valhalla::skadi::tile_data>()’ evaluates to false
/usr/include/c++/11.1.0/future: In instantiation of ‘class std::promise<valhalla::skadi::tile_data>’:
/home/anders/Documents/Programming/mapbox/valhalla/src/skadi/sample.cc:342:27:   required from here
/usr/include/c++/11.1.0/future:1065:21: error: static assertion failed: result type must be destructible
 1065 |       static_assert(is_destructible<_Res>{},
      |                     ^~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/11.1.0/future:1065:21: note: ‘std::is_destructible<valhalla::skadi::tile_data>()’ evaluates to false
[106/180] Building CXX object src/thor/CMakeFiles/valhalla-thor.dir/astar_bss.cc.o


# version
g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++,d --with-isl --with-linker-hash-style=gnu --with-system-zlib --enable-__cxa_atexit --enable-cet=auto --enable-checking=release --enable-clocale=gnu --enable-default-pie --enable-default-ssp --enable-gnu-indirect-function --enable-gnu-unique-object --enable-install-libiberty --enable-linker-build-id --enable-lto --enable-multilib --enable-plugin --enable-shared --enable-threads=posix --disable-libssp --disable-libstdcxx-pch --disable-libunwind-exceptions --disable-werror gdc_include_dir=/usr/include/dlang/gdc
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.1.0 (GCC) 
```

Clang
```
/home/anders/Documents/Programming/mapbox/valhalla/src/skadi/sample.cc:173:63: note: in instantiation of template class 'std::unordered
_map<unsigned short, std::shared_future<valhalla::skadi::tile_data>>' requested here                                                   
  std::unordered_map<uint16_t, std::shared_future<tile_data>> pending_tiles;                                                           
                                                              ^                                                                        In file included from /home/anders/Documents/Programming/mapbox/valhalla/src/skadi/sample.cc:6:                                        /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../include/c++/11.1.0/future:769:21: error: cannot initialize object parameter of type 'const std::integral_constant<bool, false>' with an expression of type 'is_destructible<valhalla::skadi::tile_data>'               static_assert(is_destructible<_Res>{},                                                                                                               ^~~~~~~~~~~~~~~~~~~~~~~                                                                                            
/home/anders/Documents/Programming/mapbox/valhalla/src/skadi/sample.cc:329:23: note: in instantiation of template class 'std::future<va
lhalla::skadi::tile_data>' requested here                                                                                              
    auto future = it->second;                                                                                                          
                      ^                                                                                                                
In file included from /home/anders/Documents/Programming/mapbox/valhalla/src/skadi/sample.cc:6:
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../include/c++/11.1.0/future:1065:21: error: cannot initialize object paramet
er of type 'const std::integral_constant<bool, false>' with an expression of type 'is_destructible<valhalla::skadi::tile_data>'
      static_assert(is_destructible<_Res>{},
                    ^~~~~~~~~~~~~~~~~~~~~~~
/home/anders/Documents/Programming/mapbox/valhalla/src/skadi/sample.cc:342:27: note: in instantiation of template class 'std::promise<v
alhalla::skadi::tile_data>' requested here
  std::promise<tile_data> promise;
                          ^
3 errors generated.

# Version

clang++ --version
clang version 12.0.1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

molind: It depends on STL version on your machine. It builds fine with libc++ 9, but won't build on libc++ 11.1 :(

nilsnolde: Oh, seems like I didn’t build master since the merge.. can smth be done about that? If I can’t build Valhalla anymore on arch linux, I’d have a semi-serious problem

PureW: Ohh another arch user! I have a semi-serious problem too :)

kevinkreiser: lets make an issue out of it and who ever gets to it first can take a stab at it

kevinkreiser: https://github.com/valhalla/valhalla/issues/3316

molind: gcc with libstdc++ 11.1.0 throws this error if `tile_data` forward declared before cache_t  
https://godbolt.org/z/8sYhf8Wqb
while clang 12.0.1 with libc++ works fine. `-stdlib=libc++` 
https://godbolt.org/z/vPYTdb3aW

PR #3124, opened 2021-06-01
add optional headers argument to python server testing script

kevinkreiser:
nothing other than what is said in the title


run_route_scripts/run_with_server.py
```diff
@@ -96,6 +97,9 @@ def make_request(post_body):
     shutil.rmtree(parsed_args.output_dir)
   os.mkdir(parsed_args.output_dir)
 
+  # setup http headers
+  parsed_args.headers = { k: v for k, v in [ h.split(': ') for hs in parsed_args.headers for h in hs] }
```
kevinkreiser, line 101: so the cool thing about `action='append'` above is that it lets you specify the arg more than once. so you can call the program like:

    ./run_with_server.py --headers "x-my-custom-header: foo" --headers "x-my-other-custom-header: bar" ...

And what comes out in python is:

    [["x-my-custom-header: foo"], ["x-my-other-custom-header: bar"]]

In the end the `requests` module wants it in a dict so i unroll the lists of lists, split on the separator and do a dict comprehension to turn it into the required dict

run_route_scripts/run_with_server.py
```diff
@@ -85,6 +85,7 @@ def make_request(post_body):
   parser.add_argument('--output-dir', type=str, help='The directory in which to place the result of each request')
   parser.add_argument('--concurrency', type=int, help='The number of processes to use to make requests', default=multiprocessing.cpu_count())
   parser.add_argument('--format', type=str, help='Supports csv, json, raw and null output formats', default='csv')
+  parser.add_argument('--headers', type=str, help='Additional http headers to send with the requests', action='append', nargs='*')
```
mandeepsandhu, line 88: Maybe add help text with an example of the expected format - `header_name: header_value` or some such?

kevinkreiser: sure

kevinkreiser: done

mandeepsandhu (approved):
lgtm

PR #3129, opened 2021-06-07
Fixed incorrect dead-end roundabout labels.

maksimandrianov:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


maksimandrianov:
We cannot mark roundabouts as `kCuldesac` correctly because we have no information about all ways. We need to do a second pass through processed ways and try to figure out the number of roads adjacent to a roundabout and mark roundabouts as `kCuldesac` correctly.  
 I see two ways to fix incorrect `kCuldesac` labels:
1. Put this logic in `ParseWays`. Result of this stage is the file with ways. I think it would be great if after this stage all ways will be correct. It means this file is consistent. BUT it needs an explicit second pass.  I prefer this way.
2. Put this logic in `BuildTileSet`. We can fix incorrect marks `kCuldesac`  in `BuildTileSet`. This way allows using `collect_node_edges` function (https://github.com/valhalla/valhalla/blob/master/src/mjolnir/node_expander.cc#L6). BUT the file with ways after `ParseWays` will not be consistent.

@gknisely , @SvetlanaBayarovich , @kevinkreiser which way is better?

maksimandrianov:
There are examples in gurka tests in this PR

merkispavel:
> 1. Put this logic in ParseWays. Result of this stage is the file with ways. I think it would be great if after this stage all ways will be correct. It means this file is consistent. BUT it needs an explicit second pass. I prefer this way.

`Parseways` for the whole planet takes 2-3 hours. It is better to avoid the second pass

maksimandrianov:
> > 1. Put this logic in ParseWays. Result of this stage is the file with ways. I think it would be great if after this stage all ways will be correct. It means this file is consistent. BUT it needs an explicit second pass. I prefer this way.
> 
> `Parseways` for the whole planet takes 2-3 hours. It is better to avoid the second pass

This second pass is simple. We do not need to process the same as in the first pass all. It is faster.  Filtering and tags processing already done.

kevinkreiser:
as you know this stuff is complicated and if the person reviewing it isnt working on this code in the moment its difficult to them to understand what is going on. so just to make this easier for everyone to reason about could you explain a few things a bit more:

1. because the code isnt linted the builds fail and we cant tell which one of these tests is su pposed to be failing in master, maybe you can explain a bit more about what is the current state of things and where we need to get to.
2. from the title of this PR i would guess that you are saying we have roundabouts that are INCORRECTLY set to cul-de-sac use. if that is the case, please explain what is currently marking F->C as a deadend/cul-de-sac?
3. do you propose to remove some naive code that is doing this incorrectly for all deadends and replace it with a final pass that does it correctly basically?

thank you!

maksimandrianov:
> as you know this stuff is complicated and if the person reviewing it isnt working on this code in the moment its difficult for them to understand what is going on. so just to make this easier for everyone to reason about could you explain a few things a bit more:
> 
> 1. because the code isnt linted the builds fail and we cant tell which one of these tests is supposed to be failing in master, maybe you can explain a bit more about what is the current state of things and where we need to get to.
> 2. from the title of this PR i would guess that you are saying we have roundabouts that are INCORRECTLY set to cul-de-sac use. if that is the case, please explain what is currently marking the edge F->C as a deadend/cul-de-sac in the first new unit test?
> 3. do you propose to remove some naive code that is doing this incorrectly for all deadends and replace it with a final pass that does it correctly basically?
> 4. i forget the details but we stopped using the deadend marker on edges in the main algorithms because a deadend is subjective (based on the costing and costing options you use). what problem are deadend roundabouts causing or not causing?
> 
> thank you!

@SvetlanaBayarovich First, who has investigated this problem:
***
`kCouldesac` use. Sometimes  this `use` is placed incorrectly, like in the picture below (orange edges are the ones that are marked as coul-de-sac:
image
As dead-end edges are excluded from ferry connection search this roundabout-deadend-imposter actually breaks the connection.
This is actually not the only case of incorrect usage of this label. Below are some examples from Palma de Mallorca. A cheap fix would be to include deadend edges in connection search as well OR [for this particular case] fix the map for this way to have roundabout tag. But a proper fix is certainly to revise dead end labelling logic.
 
|1|2|3|
|--|--|--|
|image|image|image|
***
1-2. Yes! Test `use_kCuldesac_false` is failing now.  The roundabout is marked as `kCuldesac`. But you can see that is incorrect on the ascii map.
3. Naive code that do it: https://github.com/valhalla/valhalla/blob/master/src/mjolnir/pbfgraphparser.cc#L1481-L1486. I do not want to replace this code. I want to add some code that will check and clarify a type of roundabout after finishing the processing of ways. 
4. Some problems appeared after increasing service road penalties.  And the engine can prefer an unwanted route, because `kCuldesac` marks add a big penalty:

|image2| |
|--|--|
|image5|image3|

Thanks @SvetlanaBayarovich for the screenshots!
@kevinkreiser If I was short or not clear, let me know.

test/gurka/test_loops.cc
```diff
@@ -161,3 +162,43 @@ TEST(loops, please_god_why) {
   auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_please_god_why", build_config);
   check_opposing(map, 12);
 }
+
+TEST(loops, use_kCuldesac_false) {
```
maksimandrianov, line 166: This test is failing now.

test/gurka/test_loops.cc
```diff
@@ -161,3 +162,43 @@ TEST(loops, please_god_why) {
...
+  const std::string ascii_map = R"(
```
maksimandrianov, line 167: The map looks like https://user-images.githubusercontent.com/50918075/110827032-b9e66c00-82a6-11eb-9673-6f418163a19b.png

SvetlanaBayarovich:
@maksimandrianov Could you fix linting so that the test failure becomes self-explanatory?

maksimandrianov:
> @maksimandrianov Could you fix linting so that the test failure becomes self-explanatory?

Done

kevinkreiser:
@maksimandrianov thanks for the details that really helps, i had no idea you were worried about ferry connections!

the code in the way_callback that you linked is strange... the first part of it is: `!way_.roundabout()`. that is unintuitive what am i missing? it seems to say mark it as a culdusac if "its a low class road, a loop, and NOT a roundabout". are these roundabouts that just arent tagged as roundabouts in osm?

in any case you're right that these shouldnt be marked this way though i still think the main algorithms dont care if they are marked as cul-du-sac apparently the ferry connections stuff does. my question to you would be, what is the logic you will replace the naive logic with? i understand you dont want to change it, but surely you should delete it and add your second pass somewhere else in code? if thats the case, what wil this code look like? i guess it checks for loops that dont also have other nodes that are used twice?

i feel like to fix these we could, when parsing nodes, notice that more than one node occured in the list (duplicates mean connections) and if a duplicate appeared where a way was marked as a cul-du-sac, we unmark it. i forget if we are already looking at the way when we are parsing the node though, iirc we are not. the other option i think would be to do it in the `ConstructEdges` method where i thikn we are looking at both at the same time. i would hate to add a whole new phase to the data creation just for this if we can do it in one of the other places we scan over these data structures

SvetlanaBayarovich:
@kevinkreiser Frankly speaking, I don't feel like it's that disastrous for ferries, just that unfortunate cases may happen sometimes. But it adds to some undesirable situations with service roads.

> it seems to say mark it as a culdusac if "its a low class road, a loop, and NOT a roundabout". are these roundabouts that just arent tagged as roundabouts in osm?

Hmm, not quite. Not every non-dead-end loop is a circle and should be tagged as `roundabout`, like below. It's about the way people draw ways in osm 
|||
|-|-|
|image|image|

Similarly, we might miss an actual dead end if its loop consist of 2+ ways for some reason. However, I don't think it may negatively impact routing now so I wouldn't worry

maksimandrianov:
@kevinkreiser, @SvetlanaBayarovich PTAL

kevinkreiser:
@maksimandrianov we should definitely add a changelog entry for this

src/mjolnir/pbfgraphparser.cc
```diff
@@ -46,6 +46,94 @@ int get_number(const std::string& tag, const std::string& value) { // NOLINT
...
+      if (meta.is_culdesac()) {
+        auto way_it = osm_way_seq.at(meta.get_way_index());
+        auto way = *way_it;
+        way.set_use(Use::kCuldesac);
+        way_it = way;
+        ++number_of_culdesac;
+      }
+    }
+
+    LOG_INFO("Added " + std::to_string(number_of_culdesac) + " culdesac roundabouts from " +
```
genadz, line 129: why is it `INFO` level , not `DEBUG` ?

src/mjolnir/pbfgraphparser.cc
```diff
@@ -46,6 +46,94 @@ int get_number(const std::string& tag, const std::string& value) { // NOLINT
   return num;
 }
 
+// This class helps to set "culdesac" labels to loop roads correctly.
+class culdesac_processor {
```
genadz, line 50: @maksimandrianov could you add more detailed explanation (as comments) how this class works ?

src/mjolnir/pbfgraphparser.cc
maksimandrianov: I cannot imagine how it can be useful for debug
It is printed one time, I think INFO level fits for this message and frequency

src/mjolnir/pbfgraphparser.cc
maksimandrianov: yes, done)

genadz: thanks

src/mjolnir/pbfgraphparser.cc
genadz: ok

genadz (dismissed):

genadz (approved):

PR #3143, opened 2021-06-11
Bicycle costing improvements

### LINKED ISSUE #2935: Questionable use of speed in bicycle costing
Bicycle costing uses a bicycle speed that is derived from several factors: costing option- cycling speed (or default based on bicycle type), weighted grade, and surface type. This speed is used to determine the time along an edge and base cost (unless shortest distance is specified).

The speed on the edge is used to penalize edges that are high speed. By using time dependent speed or traffic based speed it seems that this will apply smaller penalties to roads that have their speed reduced by traffic - which is not what a cyclist would want!

I believe bicycle costing should revert to using just the speed assigned in the DirectedEdge as the "default" speed of an edge (based on road class or speed limit). This also will have positive impact on performance of bicycle routing (which is generally less performant than other costing types).

dnesbitt61:
Remove use of tile->GetSpeed from bicycle costing. Speed is used to penalize higher classification roads that are high speed. We do not want these penalties reduced when heavy traffic slows the speed on an edge.

Fixes #2935

Also, added sidepath_factor so it is computed once rather than each time use_sidepath is present on an edge.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


gknisely (changes requested):
@dnesbitt61 can you show some before and after screenshots

dnesbitt61:
I don't have any screenshots. I don't have any traffic data. The basic premise is that if we use speed from traffic data it will reduce penalties on high class roads that have a lot of traffic (slower speed on the road reduces the roadway_stress due to high speed by drivers). I don't think roads with heavy traffic (but slower speeds) are less stressful than light traffic with higher speed cars.

dnesbitt61:
Added several other improvements - specifically added some class members that are precomputed so that less work is required in EdgeCost and TransitionCost methods. These changes result in 3-4% performance improvement on 2 long routes that were tested (100x in valhalla_run_route to just time the path finding algorithm). No changes to these routes occurred. In addition, I printed out the EdgeLabel counts and they did not change. Even the slightest change to costing will change the EdgeLabel counts so I think this is a valid way to determine that no routes will change.

Note - my testing is all done without any traffic data. In fact, the original premise of this PR is that traffic data should NOT be used for bicycle routes in the current bicycle costing structure as using traffic could start to favor roads with high traffic volume (due to less penalty on high speed roads).

kevinkreiser:
this looks good to me, @gknisely what say you?

dnesbitt61:
Added several other improvements - specifically added some class members that are precomputed so that less work is required in EdgeCost and TransitionCost methods. These changes result in 3-4% performance improvement on 2 long routes that were tested (100x in valhalla_run_route to just time the path finding algorithm). No changes to these routes occurred. In addition, I printed out the EdgeLabel counts and they did not change. Even the slightest change to costing will change the EdgeLabel counts so I think this is a valid way to determine that no routes will change.

Note - my testing is all done without any traffic data. In fact, the original premise of this PR is that traffic data should NOT be used for bicycle routes in the current bicycle costing structure as using traffic could start to favor roads with high traffic volume (due to less penalty on high speed roads).

kevinkreiser:
@gknisely @dnesbitt61 any movement here?

dnesbitt61:
up to @gknisely

gknisely (approved):

PR #3144, opened 2021-06-14
Penalize private gates

CuriOusQuOkka:
# Issue

Apply penalty to the gates that are marked as private. The penalty is set higher than the penalty for regular gates.

## Tasklist

 - [x] Add tests

## Requirements / Relations

 Continue this task https://github.com/valhalla/valhalla/commit/048813359b0701bea902ffa8b2749d62d390c84f


test/parse_request.cc
```diff
@@ -104,6 +110,8 @@ constexpr float kDefaultBicycle_DrivewayPenalty = 300.0f;      // Seconds
 constexpr float kDefaultBicycle_AlleyPenalty = 60.0f;          // Seconds
 constexpr float kDefaultBicycle_GateCost = 30.0f;              // Seconds
 constexpr float kDefaultBicycle_GatePenalty = 600.0f;          // Seconds
+constexpr float kDefaultBicycle_PrivateAccessCost = 5.0f;      // Seconds
+constexpr float kDefaultBicycle_PrivateAccessPenalty = 750.0f; // Seconds
```
gknisely, line 114: @CuriOusQuOkka why is kDefaultBicycle_PrivateAccessPenalty 750 for bike but 450 for the other costing

CuriOusQuOkka: The values are temporary now, I'm working on tuning them. But in the beginning it was set so high, to have `kDefaultBicycle_GatePenalty < kDefaultBicycle_PrivateAccessPenalty`. As we assume that private gates should be penalized more than gates with no access information.

dnesbitt61: I generally use a lower gate cost these days for bicycles so perhaps it can be lowered.

CuriOusQuOkka: Good point, I'll try it out.

genadz (approved):

test/parse_request.cc
CuriOusQuOkka: I've set default `gate_penalty` to 300 for bikes, tested routes didn't change. `private_access_penalty=450` looks good for bikes.

gknisely:
@CuriOusQuOkka can we see some before and after screen shots of what is changing?  I assume this was RAD tested?

CuriOusQuOkka:
> @CuriOusQuOkka can we see some before and after screen shots of what is changing? I assume this was RAD tested?

Made a huge RAD testing for this task and no problems were caught. Here are examples of what've changed
1. Used regular gate instead of private, 
image

2. Do not go through private gate. Didn’t work earlier because access was given (private), 
image

3. Took the route with 1 private gate instead of 2, image

gknisely (approved):
Looks good to me.  @dnesbitt61 thoughts

dnesbitt61 (approved):
Does private_access_penalty only apply to gates? If so maybe the name should include gates within it? I just wonder if there will be other places where transitioning onto private access might be penalized.
Otherwise I approve.

CuriOusQuOkka:
> Does private_access_penalty only apply to gates? If so maybe the name should include gates within it? I just wonder if there will be other places where transitioning onto private access might be penalized.
> Otherwise I approve.

@dnesbitt61 It is used for gates now, but we plan to include this penalty for bollards soon.

PR #3150, opened 2021-06-16
Exposing rest-area names in passive maneuvers

ktatterso:
# Issue

We currently support returning `{"service", "rest_area"}` for in our routing response.  This PR adds support for also returning the name of the rest-area (if one exists).

The gist of the changes:
- When building "intersecting edges" in trip-leg-builder, add sign information if requested. (The rest-area name is on the sign.)
- To accomplish the previous step required adding a `Sign` to the `TripLeg_IntersectingEdge` protobuf object.  (Which is then passed from thor to odin.)
- When creating/serializing the routing response (in odin, route_serializer_osrm.cc) the passive maneuver for the rest-area, add the rest-area name the the response.

I sort-of apologize in advance for some refactoring that makes this look like a lot of changes.  I simply felt that some of the routines in triplegbuilder_utils.h should instead live in triplegbuilder.cc.

I will add comments to the diffs that will help steer the reviewer(s) to the code changes of interest. 

# Testing

This was straightforward.  There were existing gurka tests with a rest-area name.  All I did was add an assertion that we returned the rest-area name.

RAD tests passed.

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/thor/triplegbuilder.cc
```diff
@@ -462,6 +462,287 @@ void SetHeadings(TripLeg_Edge* trip_edge,
   }
 }
 
+// Walk the edge_signs, add sign information onto the trip_sign, honoring which to
+// add per the attributes-controller.
+void AddSignInfo(const AttributesController& controller,
```
ktatterso, line 467: Broke this out into its own routine since it now called from two places.

src/thor/triplegbuilder.cc
```diff
@@ -462,6 +462,287 @@ void SetHeadings(TripLeg_Edge* trip_edge,
...
+  }
+
+  // Set the sign info for the intersecting edge if requested
+  if (controller.attributes.at(kNodeIntersectingEdgeSignInfo)) {
+    if (intersecting_de->sign()) {
+      // Need the index of the intersecting-edge to access its sign info from the tile
+      GraphId endnode = intersecting_de->endnode();
+      valhalla::baldr::graph_tile_ptr t2 = graphreader.GetGraphTile(endnode);
+      const DirectedEdge* zeroth_de = t2->directededge(0);
+      size_t idx = intersecting_de - zeroth_de;
```
ktatterso, line 644: Is there a better way to get the index/offset of a directed-edge?  This works fine but seems a little hokey.

src/thor/triplegbuilder.cc
```diff
@@ -462,6 +462,287 @@ void SetHeadings(TripLeg_Edge* trip_edge,
...
+      std::vector<SignInfo> edge_signs = t2->GetSigns(idx);
+      if (!edge_signs.empty()) {
+        valhalla::Sign* sign = intersecting_edge->mutable_sign();
```
ktatterso, line 647: Here's the part where we add a `Sign` to the intersecting_edge and populate it.

src/thor/triplegbuilder.cc
```diff
@@ -536,80 +817,8 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
     // Add the edge signs
     std::vector<SignInfo> edge_signs = graphtile->GetSigns(idx);
     if (!edge_signs.empty()) {
-      valhalla::Sign* trip_sign = trip_edge->mutable_sign();
```
ktatterso, line 539: Made this into its own routine, `AddSignInfo()`.

src/thor/triplegbuilder.cc
kevinkreiser: you shouldnt use the end node to get the tile and reference that with the intersecting edge. the intersecting edge is in the same tile as its begin node. its end node may be in another tile.

src/thor/triplegbuilder_utils.h
```diff
@@ -334,181 +334,4 @@ struct MultimodalBuilder {
   }
 };
 
-/**
```
ktatterso, line 337: Moved these into triplegbuilder.cc.

src/tyr/route_serializer_osrm.cc
```diff
@@ -470,254 +470,6 @@ struct IntersectionEdges {
   }
 };
 
-// Add intersections along a step/maneuver.
```
ktatterso, line 473: Moved this lower in the file.

src/thor/triplegbuilder.cc
kevinkreiser: you can probably just use the tile that was passed in and do the same thing you were doing. nothign wrong with pointer arithmetic.

```c++
auto edge_index =  intersecting_de - graphtile->directededge(0);
```

src/tyr/route_serializer_osrm.cc
```diff
@@ -1307,6 +1059,263 @@ names_and_refs(const valhalla::DirectionsLeg::Maneuver& maneuver) {
...
+    // Add rest_stop when passing by a rest_area or service_area
+    if (i > 0 && !arrive_maneuver) {
+      auto rest_stop = json::map({});
+      for (uint32_t m = 0; m < node->intersecting_edge_size(); m++) {
+        auto intersecting_edge = node->GetIntersectingEdge(m);
+        bool routeable = intersecting_edge->IsTraversableOutbound(curr_edge->travel_mode());
+
+        if (routeable && intersecting_edge->use() == TripLeg_Use_kRestAreaUse) {
+          rest_stop->emplace("type", std::string("rest_area"));
+          if (intersecting_edge->has_sign()) {
```
ktatterso, line 1138: If the `intersecting_edge` has a `Sign`, we extract the sign's `guide_destinations` text, which is the name of the rest-area.

test/gurka/test_use.cc
```diff
@@ -90,6 +90,7 @@ TEST_F(Use, test_passing_rest_area) {
   EXPECT_FALSE(steps[step_idx]["intersections"][0].HasMember("rest_stop"));
   EXPECT_TRUE(steps[step_idx]["intersections"][1].HasMember("rest_stop"));
   EXPECT_STREQ(steps[step_idx]["intersections"][1]["rest_stop"]["type"].GetString(), "rest_area");
+  EXPECT_STREQ(steps[step_idx]["intersections"][1]["rest_stop"]["name"].GetString(), "Bear Peak");
```
ktatterso, line 93: The one line of test code that proves things work.  (I also thought I'd give the rest-area a name better than "Rest Area".)

src/thor/triplegbuilder.cc
ktatterso: @kevinkreiser I think I've fixed this let me know thx.

PR #3152, opened 2021-06-16
nit: Tweaks avoid-test to only run once

PureW:
since it is not using the parametrization of the test-suite

### DIFF
--- a/test/gurka/test_avoids.cc
+++ b/test/gurka/test_avoids.cc
@@ -187,7 +187,7 @@ TEST_P(AvoidTest, TestAvoid2Polygons) {
   };
 }
 
-TEST_P(AvoidTest, TestAvoidShortcuts) {
+TEST_F(AvoidTest, TestAvoidShortcutsTruck) {
   valhalla::Options options;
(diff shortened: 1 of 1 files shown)


test/gurka/test_avoids.cc
```diff
@@ -187,7 +187,7 @@ TEST_P(AvoidTest, TestAvoid2Polygons) {
   };
 }
 
-TEST_P(AvoidTest, TestAvoidShortcuts) {
+TEST_F(AvoidTest, TestAvoidShortcutsTruck) {
```
PureW, line 190: @nilsnolde Tagged you on the PR as you showed up in `git blame`. I'm tweaking the test to `TEST_F` so it doesn't run for each of the parameters at the end of the file. This test is hard-coded to use truck costing and does not refer to the `GetParameter()` function.

nilsnolde: yep thanks, looks like c/p fail

nilsnolde (approved):
sounds good

PR #3169, opened 2021-06-24
Add country code to incident metadata

mandeepsandhu:
Added country code (iso 2 & 3 char) to incident metadata.


run_route_scripts/run_with_server.py
```diff
@@ -85,7 +85,8 @@ def make_request(post_body):
   parser.add_argument('--output-dir', type=str, help='The directory in which to place the result of each request')
   parser.add_argument('--concurrency', type=int, help='The number of processes to use to make requests', default=multiprocessing.cpu_count())
   parser.add_argument('--format', type=str, help='Supports csv, json, raw and null output formats', default='csv')
-  parser.add_argument('--headers', type=str, help='Additional http headers to send with the requests. Follows the http header spec, eg. some-header-name: some-header-value', action='append', nargs='*')
+  parser.add_argument('--headers', type=str, help='Additional http headers to send with the requests. Follows the http\
```
mandeepsandhu, line 88: This accidentally crept in. This script is currently broken in master. But I remember a PR from either @kevinkreiser or @merkispavel that fixed it (can't find the PR right now though :/ )

src/thor/triplegbuilder.cc
```diff
@@ -102,7 +119,9 @@ void AssignAdmins(const AttributesController& controller,
 void UpdateIncident(const std::shared_ptr<const valhalla::IncidentsTile>& incidents_tile,
                     TripLeg& leg,
                     const valhalla::IncidentsTile::Location* incident_location,
-                    uint32_t index) {
+                    uint32_t index,
+                    const graph_tile_ptr& tile,
+                    const valhalla::baldr::DirectedEdge* de) {
```
mandeepsandhu, line 124: This is the opposing edge of the edge on which the incident is present. I'm using it to get the start node of the incident edge.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1534,9 +1490,9 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
       auto admin_map = json::map({});
       if (admin.has_country_code()) {
         admin_map->emplace("iso_3166_1", admin.country_code());
-        auto country_iso3 = iso2_to_iso3.find(admin.country_code());
-        if (country_iso3 != iso2_to_iso3.end()) {
-          admin_map->emplace("iso_3166_1_alpha3", country_iso3->second);
+        auto country_iso3 = valhalla::tyr::get_iso_3166_1_alpha3(admin.country_code());
```
mandeepsandhu, line 1493: Moved the lookup to a util function.

src/tyr/serializers.cc
```diff
@@ -223,9 +223,13 @@ void serializeIncidentProperties(rapidjson::Writer<rapidjson::StringBuffer>& wri
     writer.String(std::string(valhalla::incidentTypeToString(incident_metadata.type())));
   }
   if (!incident_metadata.iso_3166_1_alpha2().empty()) {
-    writer.Key(key_prefix + "iso_3166_1_alpha2");
+    writer.Key(key_prefix + "iso_3166_1");
```
mandeepsandhu, line 226: Renamed this to be consistent with attribute names in the `admins` object of the serialized response.

test/gurka/test_route_incidents.cc
```diff
@@ -29,20 +29,35 @@ class IncidentsTest : public ::testing::Test {
 
     // connect the ways via the nodes
     const gurka::ways ways = {
-        {"AB", {{"highway", "tertiary"}}},  {"BC", {{"highway", "service"}}},
-        {"DEF", {{"highway", "primary"}}},  {"GHI", {{"highway", "primary"}}},
-        {"ADG", {{"highway", "motorway"}}}, {"BE", {{"highway", "secondary"}}},
-        {"EH", {{"highway", "tertiary"}}},
+        {"AB", {{"highway", "tertiary"}, {"driving_side", "left"}}},
```
mandeepsandhu, line 32: Had to add driving side since we're no longer using admin db and instead manually setting certain admin attributes.

src/thor/triplegbuilder.cc
```diff
@@ -92,6 +93,22 @@ void AssignAdmins(const AttributesController& controller,
   }
 }
 
+// Helper function to get the iso country code from an edge
+inline std::string country_code_from_edge(const graph_tile_ptr& tile,
+                                          const valhalla::baldr::DirectedEdge* de) {
+  return tile->admininfo(tile->node(de->endnode())->admin_index()).country_iso();
```
PureW, line 99: Does this need a null-pointer check of `de` or could a reference be passed instead of the pointer?

src/thor/triplegbuilder.cc
```diff
@@ -92,6 +93,22 @@ void AssignAdmins(const AttributesController& controller,
...
+}
+
+inline const DirectedEdge*
+get_opposite_de(const DirectedEdge* edge, GraphReader& reader, const graph_tile_ptr& tile) {
```
PureW, line 103: Hmm I thought we had a helper function like this elsewhere.

Same question here though regarding null-pointer check or passing reference.

src/thor/triplegbuilder.cc
PureW: I don't quite understand this comment. Why is this the opposing edge?

src/tyr/route_serializer_osrm.cc
```diff
@@ -1784,7 +1741,8 @@ TEST(RouteSerializerOsrm, testserializeIncidents) {
         {
           "id": "18446744073709551615",
           "type": "weather",
-          "iso_3166_1_alpha2": "AU",
```
PureW, line 1787: Hmm just a note that this changes existing api.

src/tyr/serializers.cc
PureW: This is a potentially api-breaking change though.

mandeepsandhu: Yes. I'm open to reverting it to its original name. However it'd be nice to have a consistent name for the ISO 2-char country code everywhere in the response. @kevinkreiser do you know of existing consumers of valhalla API that are relying on this name?

src/thor/triplegbuilder.cc
mandeepsandhu: Consider the following case:
```
         i   i
... A----+---+--->B ...
```
Incident is between the 2 `i`'s. Node A is in country C1 & B in country C2. In triplegbuilder when consuming edge A->B the incident needs to be associated with the country code of A (since we haven't crossed the border into country C2 of B, CMIIW here). Therefore, we get the opposing edge B->A and use its end node (which is A) to get the country code (since there's no API to get the start node given a directed edge).

The whole premise of doing this is that an incident needs to be associated with the country of the node just prior to the start of the incident.

PureW: I see. Could the variable be named `opposite_edge` instead of `de` though?

run_route_scripts/run_with_server.py
kevinkreiser: its over in @genadz 's pr

src/thor/triplegbuilder.cc
kevinkreiser: yes we have this as a method on graphreader i would suggest removing it and using that

src/thor/triplegbuilder.cc
kevinkreiser: in a lot of places we use the end node of the edge to denote which admin its in since we dont split the edges where admin boundaries occur begin node is as good as end node. i would suggest simplifying by just using the end node of the edge that the incident is on as its simpler to get and will be just as in correct in the extreme edge cases where it would occur

src/tyr/serializers.cc
kevinkreiser: i dont know anyone else that uses it but it is in a release already so its kind of evil. why not just leave it and have the alpha3 one have the suffix?

valhalla/tyr/util.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <string>
+
+namespace valhalla {
+namespace tyr {
+
+// Returns the 3-char equivalent of the 2-char country code (iso_3166_1_alpha2) or an empty string
+// if the 2-char code is unknown
+std::string get_iso_3166_1_alpha3(const std::string& iso_3166_1_alpha2);
```
kevinkreiser, line 10: this seems like an unfortunate expansion of the public api, can we just throw this into the compilation unit where it is used? or if we want to use it in more than one place can we make it a private header (ie move it to `src/tyr`)?

src/thor/triplegbuilder.cc
```diff
@@ -120,6 +139,12 @@ void UpdateIncident(const std::shared_ptr<const valhalla::IncidentsTile>& incide
     const auto& meta = valhalla::baldr::getIncidentMetadata(incidents_tile, *incident_location);
     *new_incident->mutable_metadata() = meta;
 
+    // Set iso country code (2 & 3 char codes) on the new incident obj created for this leg
+    std::string country_code_iso_2 = country_code_from_edge(tile, de);
+    std::string country_code_iso_3 = valhalla::tyr::get_iso_3166_1_alpha3(country_code_iso_2);
```
kevinkreiser, line 144: we shouldnt be calling tyr from thor. this makes me think actually that this util funciton should be moved to like `baldr` or `midgard` or something

valhalla/tyr/util.h
kevinkreiser: i see you use it in triple g builder which makes me think th is should just go into like baldr/graphconstants or something

run_route_scripts/run_with_server.py
mandeepsandhu: ah ok :facepalm:

src/tyr/serializers.cc
mandeepsandhu: I'll do that. When looking at the code initially, I found that we used a different name for iso2 code in the "admins" object.

valhalla/tyr/util.h
mandeepsandhu: yeah its needed in 2 places thor::triplegbuilder and tyr when serializing the response. I looked at midgard initially but it seemed specific to logging (?). Is putting it within baldr a good option?

mandeepsandhu:
@kevinkreiser @PureW this is ready for review. gist of changes:
- moved the iso 2->3 conversion function to baldr/admin (i felt admin was the appropriate place as the iso lookup is kind of related)
- use endnode for incident CC
- updated tests

PureW (approved):
LGTM!

PureW (approved):

PR #3183, opened 2021-06-29
Use distance instead of time to check limited sharing criteria

genadz:
# Issue

This PR changes limited sharing criteria that is used to find good alternatives.
Right now to check if alternative candidate is different enough we calculate how much time (sum of duration on the shared edges) this alternative candidate shares with already chosen routes. In some cases it leads to a situation when alternative is geometrically too similar to the main route. Our goal is to avoid proposing such alternatives that are not very "alternative". If we use distance instead of time we will guarantee that at least 25% of route geometry will be different from the main route.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/valhalla/thor/pathinfo.h
+++ b/valhalla/thor/pathinfo.h
@@ -18,6 +18,7 @@ struct PathInfo {
                           // of the edge
   uint32_t trip_id;       // Trip Id (0 if not a transit edge).
   baldr::GraphId edgeid;  // Directed edge Id
+  float path_distance;    // Distance (in meters) from the start to the edge
   uint8_t restriction_index;    // Record which restriction
   sif::Cost transition_cost;    // Turn cost at the beginning of the edge
   bool start_node_is_recovered; // Indicates if the start node of the edge is an inner node
@@ -29,11 +30,13 @@ struct PathInfo {
            const sif::Cost c,
            const baldr::GraphId& edge,
            const uint32_t tripid,
+           const float path_distance,
            const uint8_t restriction_idx = baldr::kInvalidRestriction,
            const sif::Cost tc = {},
            bool start_node_is_recovered = false)
-      : mode(m), elapsed_cost(c), trip_id(tripid), edgeid(edge), restriction_index(restriction_idx),
-        transition_cost(tc), start_node_is_recovered(start_node_is_recovered) {
+      : mode(m), elapsed_cost(c), trip_id(tripid), edgeid(edge), path_distance(path_distance),
+        restriction_index(restriction_idx), transition_cost(tc),
+        start_node_is_recovered(start_node_is_recovered) {
   }
 
   // Stream output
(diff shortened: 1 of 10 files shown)


genadz:
No difference in performance.

valhalla/thor/pathinfo.h
```diff
@@ -29,19 +30,22 @@ struct PathInfo {
            const sif::Cost c,
            const baldr::GraphId& edge,
            const uint32_t tripid,
+           const float path_distance,
```
genadz, line 33: added this field in order not to calculate edge lengths twice.

genadz:
How I tested the influence of this PR ? - I generated 1000 random routes in Europe and 1000 in US (origin and destination points match with the graph nodes). Some statistics:

**Europe**
Number of routes with at least one alternative: **master** = 783, **branch** = 748.
Only in **master** = 36, only in **branch** = 1, differences = 49, same alternatives =  698.

**US**
Number of routes with at least one alternative: **master** = 841, **branch** = 823.
Only in **master** = 18, only in **branch** = 0, differences = 57, same alternatives =  766.

We can see, that the number of alternatives slightly **decreased** (cases when alternative route geometrically is very similar to the main route). Also there are several routes that have **different** alternatives for **master** and **branch** (when first alternative passed limited sharing criteria based on time but was declined by limited sharing criteria based on distance).

Let me share some random examples:

**in Europe**

`{"locations":[{"lon":12.264997,"lat":49.603917},{"lon":8.634936,"lat":52.797862}],"costing":"auto","alternates":2}`
| before | after |
| --- | --- |
| Screenshot from 2021-06-29 15-02-43 | Screenshot from 2021-06-29 15-02-49 |

`{"locations":[{"lon":19.915883,"lat":44.010619},{"lon":21.454978,"lat":41.988423}],"costing":"auto","alternates":2}`
| before | after |
| --- | --- |
| Screenshot from 2021-06-29 15-05-02 | Screenshot from 2021-06-29 15-05-13 |

`{"locations":[{"lon":18.314892,"lat":49.86183},{"lon":21.376047,"lat":50.058683}],"costing":"auto","alternates":2}`
| before | after |
| --- | --- |
| Screenshot from 2021-06-29 15-08-31 | Screenshot from 2021-06-29 15-08-27 |


**in US**

`{"locations":[{"lon":-121.538765,"lat":37.766828},{"lon":-122.388933,"lat":37.76652}],"costing":"auto","alternates":1}`
| before | after |
| --- | --- |
| Screenshot from 2021-06-29 15-12-55 |  Screenshot from 2021-06-29 15-12-59 |

`{"locations":[{"lon":-79.980707,"lat":43.873853},{"lon":-76.619195,"lat":40.336108}],"costing":"auto","alternates":1}`
| before | after |
| --- | --- |
| Screenshot from 2021-06-29 15-18-45 | Screenshot from 2021-06-29 15-18-39 |

`{"locations":[{"lon":-96.687353,"lat":33.104741},{"lon":-93.721444,"lat":32.506684}],"costing":"auto","alternates":2}`
| before | after |
| --- | --- |
| Screenshot from 2021-06-29 15-21-37 | Screenshot from 2021-06-29 15-21-40 |

`{"locations":[{"lon":-83.763403,"lat":31.191534},{"lon":-84.440056,"lat":33.978027}],"costing":"auto","alternates":2}`
| before | after |
| --- | --- |
| Screenshot from 2021-06-29 15-24-59 | Screenshot from 2021-06-29 15-24-54 |


As for me, these look like improvement. We decline very similar alternative candidates and in some cases propose alternatives with a small sharing value.

kevinkreiser (approved):
this is more intuitive to me anyway since its unaffected by mode of travel and changing traffic conditions, thanks for the cleanup!

mandeepsandhu:
I like the shape of the new alternatives! Great work @genadz !

genadz:
> I like the shape of the new alternatives! Great work @genadz !

Thanks!

PR #3185, opened 2021-06-29
fix zero length paths

CuriOusQuOkka:
# Issue

Currently when start and end points are mapped to the same location, we return final maneuvers with `time=0s, distance=100m`. It means that the **speed** should be **infinite**. 

Distance should be not modified, do not override it with a bigger number.

## Requirements / Relations

This behaviour was added in the commit https://github.com/valhalla/valhalla/commit/663c0fcda1a9981bd874a6326821979b945e867f but it is not clear, which problem was originally solved using the threshold.


CuriOusQuOkka:
@dnesbitt61 Could you please give me a bit of context, what was the problem that you wanted to solve here https://github.com/valhalla/valhalla/commit/663c0fcda1a9981bd874a6326821979b945e867f? maybe some examples?

dnesbitt61:
That is 1m. I do not remember the context of why that was added.

CuriOusQuOkka:
> this looks ok to me as long as there are no parts of the code that rely on this anymore. i did a quick scan and didnt see any butits a large code base. seems like its only used in odin and in serialization and those uses looked safe to me

I had information that this meters were added to make bearing parameter work. Checked conner cases here, they look good.

PR #3187, opened 2021-06-30
try again to update googletest

### LINKED ISSUE #3186: googletest not actually updated
PR #3166 should've updated googletest from commit [d166e09483845b9b6a658dccc3d3dbb293676b62](https://github.com/google/googletest/commit/d166e09483845b9b6a658dccc3d3dbb293676b62) to 
[e2239ee6043f73722e7aa812a459f54a28552929](https://github.com/google/googletest/commit/e2239ee6043f73722e7aa812a459f54a28552929), but seems it didn't: in https://github.com/valhalla/valhalla/tree/master/third_party there's still the old commit referenced.

cc @PureW

nilsnolde:
fixes #3186 (hopefully)

at least the third_party module on this branch shows the right commit now, so should've done the job


PureW (approved):

kevinkreiser:
i didnt check git blame but its fairly common for folks to accidentally not update thier submodules when they merge master into their branches and then just commit up the revert of the submodule version. sometimes reviewers dont notice the submodule version changes in the diff because they are so small compared to the rest of the diff as well. we could do some automation in CI to call out submodule changes

PureW:
Googletest was upgraded earlier but then downgraded again in https://github.com/valhalla/valhalla/pull/3168

nilsnolde:
Oh right I missed that..

PR #3189, opened 2021-06-30
Via Waypoints

kdiluca:
# Description

This PR introduces a new via_waypoints array on the leg in the osrm route serializer that describes where a particular waypoint from the root-level array matches to the route.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/tyr/serializers.cc
```diff
@@ -209,6 +234,42 @@ json::ArrayPtr waypoints(const valhalla::Trip& trip) {
   return waypoints;
 }
 
+json::ArrayPtr via_waypoints(valhalla::Options options, const std::vector<PointLL>& shape) {
```
PureW, line 237: Could you add a doc-string describing what this function does? And also for the `via_waypoint` above?

PureW:
@kdiluca it seems like you've commited back the old thirdparty modules that were updated in master

src/tyr/serializers.cc
```diff
@@ -179,6 +179,34 @@ waypoint(const valhalla::Location& location, bool is_tracepoint, bool is_optimiz
   return waypoint;
 }
 
+/*
+ * This function serializes the via_waypoints object on the leg.
+ */
+valhalla::baldr::json::MapPtr via_waypoint(const valhalla::Location& location,
```
PureW, line 185: Would it be possible to add some unittests of this and the `via_waypoints` below in this file to exercise them some more?

src/tyr/serializers.cc
```diff
@@ -179,6 +179,34 @@ waypoint(const valhalla::Location& location, bool is_tracepoint, bool is_optimiz
...
+                                           const uint64_t geometry_idx) {
+
+  // Create a via waypoint to add to the array
+  auto via_waypoint = json::map({});
+  // Add distance in meters from the input location to the nearest
+  // point on the road used in the route
+  // TODO: since distance was normalized in thor - need to recalculate here
+  //       in the future we shall have store separately from score
```
PureW, line 193: What does this comment mean?

src/tyr/serializers.cc
```diff
@@ -179,6 +179,34 @@ waypoint(const valhalla::Location& location, bool is_tracepoint, bool is_optimiz
...
+  via_waypoint
+      ->emplace("distance_from_leg_start",
+                json::fixed_t{to_ll(location.ll()).Distance(to_ll(location.path_edges(0).ll())), 3});
+
+  if (location.shape_index() == numeric_limits<uint32_t>::max()) {
+    // when waypoint is neither a break nor leg's starting/ending
+    // point (shape_index is uint32_t max), we assign null to its waypoint_index
+    via_waypoint->emplace("waypoint_index", static_cast<std::nullptr_t>(nullptr));
```
PureW, line 201: Why do we need to add it at all if it's not a via waypoint?

src/tyr/serializers.cc
```diff
@@ -179,6 +179,34 @@ waypoint(const valhalla::Location& location, bool is_tracepoint, bool is_optimiz
...
+  } else {
+    via_waypoint->emplace("waypoint_index", static_cast<uint64_t>(location.shape_index()));
```
PureW, line 203: Is `shape_index` correct here? We want `waypoint_index` to be the index of the input location we're referring to?

src/tyr/serializers.cc
```diff
@@ -209,6 +237,47 @@ json::ArrayPtr waypoints(const valhalla::Trip& trip) {
...
+json::ArrayPtr via_waypoints(valhalla::Options options, const std::vector<PointLL>& shape) {
+  // Create a vector of indexes based on the number of locations.
+  auto locs = *options.mutable_locations();
+  uint32_t i = 0;
+  std::vector<uint32_t> indexes;
+  for (const auto& loc : locs) {
+    indexes.push_back(i++);
+  }
+  auto via_waypoints = json::array({});
+  for (const auto& index : indexes) {
```
PureW, line 254: Nit: Could `index` and `indexes` be more verbosively named, maybe something like `location_index`. There is a lot of indexing going on I feel like.

src/tyr/serializers.cc
```diff
@@ -209,6 +237,47 @@ json::ArrayPtr waypoints(const valhalla::Trip& trip) {
...
+    // Only create via_waypoints object if the locations are via or through types
+    if (locs.Get(index).type() == 1 || locs.Get(index).type() == 2) {
```
PureW, line 256: Could we use the protobuf enum here instead of magic numbers? As in `..type() == Location::Type::kBreak`?

src/tyr/serializers.cc
```diff
@@ -209,6 +237,47 @@ json::ArrayPtr waypoints(const valhalla::Trip& trip) {
...
+      locs.Mutable(index)->set_shape_index(index);
+      int geometry_idx = 0;
+
+      for (const auto& ll : shape) {
+        geometry_idx++;
+
+        if (index == 0 || index == (locs.size() - 1)) {
```
PureW, line 263: This conditional might as well live up on line 256 together with the check for kThrough and kVia?

src/tyr/serializers.cc
```diff
@@ -209,6 +237,47 @@ json::ArrayPtr waypoints(const valhalla::Trip& trip) {
...
+          continue;
+        }
+        if (valhalla::midgard::equal<double>(static_cast<double>(
+                                                 locs.Get(index).path_edges(0).ll().lng()),
+                                             static_cast<double>(ll.first), .001) &&
+            valhalla::midgard::equal<double>(static_cast<double>(
+                                                 locs.Get(index).path_edges(0).ll().lat()),
+                                             static_cast<double>(ll.second), .001)) {
```
PureW, line 271: A stylistic nit. This conditional is quite hard for me to read. What do you think about breaking it down into two `bool`s that are constructed before the if-statement?

src/tyr/serializers.cc
kdiluca: Removed comment as this doesn't relate to via_waypoints

src/tyr/serializers.cc
kdiluca: We don't since I'm already doing a check for these, I have removed the extra one.

src/tyr/serializers.cc
kdiluca: As we discussed in our meeting, this looks like it was mis-named in the proto so I have renamed to waypoint_index in the proto to fix any more confusion.

src/tyr/serializers.cc
kdiluca: Yes agree

src/tyr/serializers.cc
kdiluca: fixed

src/tyr/serializers.cc
kdiluca: moved.

src/tyr/serializers.cc
kdiluca: refactored

src/tyr/serializers.cc
```diff
@@ -173,12 +172,31 @@ waypoint(const valhalla::Location& location, bool is_tracepoint, bool is_optimiz
...
+ */
+valhalla::baldr::json::MapPtr serialize_via_waypoint(const valhalla::Location& origin,
+                                                     const valhalla::Location& location,
+                                                     const uint64_t geometry_idx) {
+
+  // Create a via waypoint to add to the array
+  auto via_waypoint = json::map({});
+  // Add distance in meters from the input location to the silent waypoint
+  via_waypoint->emplace("distance_from_leg_start",
+                        json::fixed_t{to_ll(origin.ll()).Distance(to_ll(location.path_edges(0).ll())),
```
PureW, line 192: `PointLL::Distance` measerus the straight line distance which is not what we want here. We want the distance along the route-line.

src/tyr/serializers.cc
```diff
@@ -209,6 +224,60 @@ json::ArrayPtr waypoints(const valhalla::Trip& trip) {
...
+    }
+
+    // Only create via_waypoints object if the locations are via or through types
+    if (locs.Get(location_index).type() == valhalla::Location::kVia ||
+        locs.Get(location_index).type() == valhalla::Location::kThrough) {
+      locs.Mutable(location_index)->set_waypoint_index(location_index);
+
+      double distance_from_leg_start = 0;
+      uint32_t geometry_idx = 0;
+      for (const auto& ll : shape) {
```
PureW, line 253: You could also adjust this second loop to not re-calculate the done segments over and over, and instead continue where the last waypoint was. It could matter in the case where there are a large number of via waypoints to not have to recalculate number of coordinates times number of waypoints.

PureW (dismissed):
This looks good, nice work! My only last thought is if we can [bring down the computational complexity a little](https://github.com/valhalla/valhalla/pull/3189/files#r6687183930)

kdiluca:
Will be storing the via/through locations earlier in the pipeline when calling astar instead of matching on shape for phase II.

PureW (approved):
Approving with the intention of a followup PR to make logic more robust to tricky edge-cases.

PureW (approved):
Approving with the intention of a followup PR to make logic more robust to tricky edge-cases.

PR #3191, opened 2021-07-02
Optionally Send Statsd Metrics for Running Service

kevinkreiser:
This PR seeks to add statsd integration to the library. The first step is adding it to the running service however we should also add this to the data processing (mjolnir binaries) as well.

It makes use of a header only statsd client that I've been working on with @vthiery. ~Since we target windows builds here as well and there is no reason the code shouldnt work with windows too I've vendored my own personal fork of the library. The reason for that is because windows support isnt yet merged into the original project. As soon as I get CI working for windows in the upstream project I'll update the submodule to point to the real repository instead of my fork. Basically waiting on me fixing this: https://github.com/vthiery/cpp-statsd-client/pull/33~ this is merged and now this PR points at the canonical repository rather than my fork

This adds support for publishing the following stats:

1. all valhalla_execption_t errors/warnings
2. latencies of top level apis at each stage of the request pipeline (eg how much time to thor spend in /route)
3. "ok" stats for when a request is returned succesfully
4. worker starts (helpful for knowing when you are crashing)

The format of the stats keys is:

`action.worker.log_level.metric_name`

Which should make for some pretty easy to write querries for debugging production issues.

There are 2 future work items we might want to tackle:

1. white/blacklisting for certain metrics, in pracice the number of unique metrics yhou see in a running system should be pretty small but technically there are about 90 possible.
2. breaking out serialization latency stats from odin latency stats. for the moment they are rolled into one

cc @mskurydin i did change the odin worker api a little bit which might need changes on your end, its very mechanical but this is just a heads up! if its unbearable let me know and we can revert that and find another way to accomplish what i was after


kevinkreiser:
This is ready for review but I think it could use some discussion. The format I followed for the metrics key was:

    user_defined_prefix.action.worker.metric

where `action` is one of `route` `locate` `isochrone` etc and worker is one of `loki` `thor` `odin`. i did not include the http code in the metric name because some stats naturally wouldnt have one. for example the timings dont have an http code because they arent concerned with the return value. or if we added number of iterations in thor for example it wouldnt have one. but what this means is that downstream when you look at these metrics you will have to know their names. i was tempted to add http code between `action` and `worker` so downstream metrics querys could be like `valhalla.route.5*` to find any http 500s. does anyone have opinions on this?

src/worker.cc
```diff
@@ -1233,12 +1089,93 @@ worker_t::result_t to_response(const std::string& data,
...
+
+  // before we are done with the request, if this was not an error we log it was ok
+  if (api.info().has_error() && api.info().error()) {
+    auto worker = typeid(*this) == typeid(loki::loki_worker_t)
+                      ? ".loki."
+                      : (typeid(*this) == typeid(thor::thor_worker_t) ? ".thor." : ".odin.");
+    const auto& action = Options_Action_Enum_Name(api.options().action());
+
+    statsd_client->count(action + worker + "ok", 1, 1.f, statsd_client->tags);
+  }
```
nilsnolde, line 1161: isn't that the wrong way around? you're setting "ok" _if_ there was an error

src/odin/worker.cc
```diff
@@ -73,29 +74,36 @@ odin_worker_t::work(const std::list<zmq::message_t>& job,
...
   } catch (const std::exception& e) {
-    return jsonify_error({299, std::string(e.what())}, info, request);
+    result = jsonify_error({299, std::string(e.what())}, info, request);
   }
+
+  // keep track of the metrics (for now this should always happen)
+  if (!result.intermediate)
+    enqueue_statistics(request);
+
+  // keep track of the metrics if the request is going back to the client (this should be the case)
```
nilsnolde, line 105: misplaced comment?

src/loki/worker.cc
```diff
@@ -323,16 +319,22 @@ loki_worker_t::work(const std::list<zmq::message_t>& job,
...
     LOG_ERROR("400::" + std::string(e.what()) + " request_id=" + std::to_string(info.id));
-    return jsonify_error({199, std::string(e.what())}, info, request);
+    result = jsonify_error({199, std::string(e.what())}, info, request);
   }
+
+  // keep track of the metrics
+  if (!result.intermediate)
+    enqueue_statistics(request);
+
+  // keep track of the metrics if the request is going back to the client
```
nilsnolde, line 336: misplaced comment?

nilsnolde (dismissed):
Ui, lots of changes:) Can't say much about statsd unfortunately, only thing I ever used was Tyk + Tyk pump -> Graylog. But sure looks interesting as fully opensource..

Anyways, left a few comments, and: think you updated googletest accidentally.

kevinkreiser:
@nilsnolde yeah i wanted to go over the diff and weed out all of those little things, sorry for making you a guinea pig there. i will clean them up and then ill go back through and throw in comments. at this point only the structure ready for review :smile:

nilsnolde:
didn't have much to say on structure, so figured at least I can take a look at some of the logic. looks pretty sane to put in the service_worker though

src/loki/height_action.cc
```diff
@@ -71,7 +71,7 @@ std::vector<PointLL> loki_worker_t::init_height(Api& request) {
 */
 std::string loki_worker_t::height(Api& request) {
   // time this whole method and save that statistic
-  auto _ = measure_scope_time(request, "loki_worker_t::height");
+  auto _ = measure_scope_time(request);
```
kevinkreiser, line 74: there will be a lot of changes that look just like this. i thought it was stupid to have to have the caller know the action and the worker. instead i can figure this out programatically from within the `measure_scope_time` function

src/loki/worker.cc
```diff
@@ -149,7 +149,7 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
 
 loki_worker_t::loki_worker_t(const boost::property_tree::ptree& config,
                              const std::shared_ptr<baldr::GraphReader>& graph_reader)
-    : config(config), reader(graph_reader),
+    : service_worker_t(config), config(config), reader(graph_reader),
```
kevinkreiser, line 152: now that the base class does something we actually have to pass config to it

src/loki/worker.cc
```diff
@@ -245,6 +245,7 @@ loki_worker_t::loki_worker_t(const boost::property_tree::ptree& config,
 }
 
 void loki_worker_t::cleanup() {
+  service_worker_t::cleanup();
```
kevinkreiser, line 248: same thing, the base class actually has stuff to clean up, in this case it ships all the queued metrics to the stats daemon if so configured

src/loki/worker.cc
```diff
@@ -261,10 +262,11 @@ loki_worker_t::work(const std::list<zmq::message_t>& job,
                     void* request_info,
                     const std::function<void()>& interrupt_function) {
 
-  // grab the request info
+  // grab the request info and make sure to record any metrics before we are done
   auto& info = *static_cast<prime_server::http_request_info_t*>(request_info);
   LOG_INFO("Got Loki Request " + std::to_string(info.id));
   Api request;
+  prime_server::worker_t::result_t result{true, {}, ""};
```
kevinkreiser, line 269: i refactored every worker method to look like this. basically i make the result upfront and then overwrite below. this was to allow the same object to carry the final state of the request so that at the bottom of the function i could..

src/loki/worker.cc
```diff
@@ -323,16 +319,22 @@ loki_worker_t::work(const std::list<zmq::message_t>& job,
...
+    result = jsonify_error(e, info, request);
   } catch (const std::exception& e) {
     LOG_ERROR("400::" + std::string(e.what()) + " request_id=" + std::to_string(info.id));
-    return jsonify_error({199, std::string(e.what())}, info, request);
+    result = jsonify_error({199, std::string(e.what())}, info, request);
   }
+
+  // keep track of the metrics
+  if (!result.intermediate)
+    enqueue_statistics(request);
```
kevinkreiser, line 334: ...make this call to suck any statistics out to the queue to be pushed to statsd on `cleanup`. note that we only do this when the result isnt intermediate (which means that it will be returned to the caller at this stage rather than go on to a deeper stage of the pipeline)

src/mjolnir/bssbuilder.cc
```diff
@@ -95,23 +95,6 @@ struct BSSConnection {
...
-  Finally& operator=(const Finally&) = delete;
-  Finally& operator=(Finally&&) = delete;
-  ~Finally() {
-    t();
-  };
-};
-
-template <typename T> Finally<T> make_finally(T t) {
-  return Finally<T>{t};
-};
```
kevinkreiser, line 113: this RAII based approach to implementing something like javas finally was exactly the same one i had previously implemented for the scoped_timer. since @xlqian 's version here is more generic, i decided to move it to midgard and then use it to replace the scoped_timer. great minds think alike!

src/odin/worker.cc
```diff
@@ -27,23 +27,23 @@ using namespace valhalla::baldr;
 namespace valhalla {
 namespace odin {
 
-odin_worker_t::odin_worker_t(const boost::property_tree::ptree&) {
+odin_worker_t::odin_worker_t(const boost::property_tree::ptree& config) : service_worker_t(config) {
```
kevinkreiser, line 30: these changes are the same as the other workers

src/thor/worker.cc
```diff
@@ -53,8 +53,7 @@ std::string serialize_to_pbf(Api& request) {
   std::string buf;
```
kevinkreiser, line 53: these changes are teh same as the other workers

src/odin/worker.cc
```diff
@@ -27,23 +27,23 @@ using namespace valhalla::baldr;
...
 }
 
 odin_worker_t::~odin_worker_t() {
 }
 
-void odin_worker_t::cleanup() {
```
kevinkreiser, line 36: the parent one gets called instead

nilsnolde:
Btw, how much overhead does this approx add for a typical routing request when it's purging the statsd request(s) at the end? Or is that somehow async and it's not a blocking POST request? I have little feeling how many stats would be typically collected now.

src/tyr/actor.cc
```diff
@@ -62,10 +62,8 @@ actor_t::route(const std::string& request_str, const std::function<void()>* inte
   pimpl->loki_worker.route(request);
   // route between the locations in the graph to find the best path
   pimpl->thor_worker.route(request);
-  // get some directions back from them
-  pimpl->odin_worker.narrate(request);
-  // serialize them out to json string
-  auto bytes = tyr::serializeDirections(request);
+  // get some directions back from them and serialize
+  auto bytes = pimpl->odin_worker.narrate(request);
```
kevinkreiser, line 66: i decided to collapse down the serialization into the `narrate` method in the odin worker. i did this mainly so  that i could count that time of doing serialization as part of the latency metric. if we'd rather track the odin maneuver gen and narration stuff separately from serialization i can undo this and move some of the bits around to accomplish it

test/parse_request.cc
```diff
@@ -2354,7 +2354,9 @@ TEST(ParseRequest, test_use_living_streets) {
   costing_with_defaults.emplace_back(Costing::pedestrian, kDefaultPedestrian_UseLivingStreets);
   costing_with_defaults.emplace_back(Costing::bicycle, kDefaultBicycle_UseLivingStreets);
 
-  for (const auto& [costing, default_value] : costing_with_defaults) {
```
kevinkreiser, line 2357: the changes here are to fix compiler warnings about c++17 that i was sick of seeing

src/worker.cc
```diff
@@ -1,297 +1,140 @@
...
+constexpr const char* OSRM_NO_ROUTE = R"({"code":"NoRoute","message":"Impossible route between points"})";
+constexpr const char* OSRM_NO_SEGMENT = R"({"code":"NoSegment","message":"One of the supplied input coordinates could not snap to street segment."})";
+constexpr const char* OSRM_SHUTDOWN = R"({"code":"ServiceUnavailable","message":"The service is shutting down."})";
+constexpr const char* OSRM_SERVER_ERROR = R"({"code":"InvalidUrl","message":"Failed to serialize route."})";
+constexpr const char* OSRM_DISTANCE_EXCEEDED = R"({"code":"DistanceExceeded","message":"Path distance exceeds the max distance limit."})";
+constexpr const char* OSRM_PERIMETER_EXCEEDED = R"({"code":"PerimeterExceeded","message":"Perimeter of avoid polygons exceeds the max limit."})";
+constexpr const char* OSRM_BREAKAGE_EXCEEDED = R"({"code":"BreakageDistanceExceeded","message":"All coordinates are too far away from each other"})";
+
+using ve = valhalla_exception_t;
+const std::unordered_map<unsigned, valhalla::valhalla_exception_t> error_codes{
```
kevinkreiser, line 50: so we had a bunch of parallel maps and it was a real mental burdon to keep them synchronized not to mention look up which ones went with eachother. so instead of maintaining that, ive revamped the `valhalla_exception_t` class to store all the different types of info directly and declare all of those predefined (ie non custom) ones here. i turned off clang format because it just brutally destroys the formatting and makes it again very h ard to read. 1 line, 1 error and sanity remains.

src/worker.cc
```diff
@@ -1079,15 +922,14 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
-  message += (extra ? ":" + *extra : "");
-  auto http_code_iter = ERROR_TO_STATUS.find(code);
-  http_code = (http_code_iter == ERROR_TO_STATUS.cend() ? 0 : http_code_iter->second);
-  auto http_message_iter = HTTP_STATUS_CODES.find(http_code);
-  http_message = (http_message_iter == HTTP_STATUS_CODES.cend() ? "" : http_message_iter->second);
+valhalla_exception_t::valhalla_exception_t(unsigned code, const std::string& extra)
+    : std::runtime_error("") {
+  auto code_itr = error_codes.find(code);
+  if (code_itr != error_codes.cend()) {
+    *this = code_itr->second;
```
kevinkreiser, line 929: so now if you specify just the valhalla error code we look up a predefined exception from above and return that.

src/worker.cc
```diff
@@ -1097,18 +939,14 @@ void ParseApi(const std::string& request, Options::Action action, valhalla::Api&
...
+std::string jsonify_error(const valhalla_exception_t& exception, Api& request) {
   // get the http status
   std::stringstream body;
 
   // overwrite with osrm error response
   if (request.options().format() == Options::osrm) {
-    auto found = OSRM_ERRORS_CODES.find(exception.code);
-    if (found == OSRM_ERRORS_CODES.cend()) {
-      found = OSRM_ERRORS_CODES.find(199);
-    }
```
kevinkreiser, line 1109: this saves us having to do multiple looks ups

src/worker.cc
```diff
@@ -1120,6 +958,24 @@ std::string jsonify_error(const valhalla_exception_t& exception, const Api& requ
...
+
+  auto* err_stat = request.mutable_info()->mutable_statistics()->Add();
+  err_stat->set_key(action + worker + exception.statsd_key);
+  err_stat->set_value(1);
+  err_stat->set_type(count);
+
+  auto* stat = request.mutable_info()->mutable_statistics()->Add();
+  stat->set_key(action + worker + "http_" + std::to_string(exception.http_code));
+  stat->set_value(1);
+  stat->set_type(count);
```
kevinkreiser, line 977: here we add both the specific error metric and we send another basic metric with the http code in it. i guess this is a way to solve the question asked in: https://github.com/valhalla/valhalla/pull/3191#issuecomment-875931103

but it does double the amount of reporting we do. should we combine them into one metric?

src/worker.cc
```diff
@@ -1120,6 +958,24 @@ std::string jsonify_error(const valhalla_exception_t& exception, const Api& requ
     body << (request.options().has_jsonp() ? request.options().jsonp() + "(" : "") << *json_error
          << (request.options().has_jsonp() ? ")" : "");
   }
+
+  // write a few stats about the error
+  request.mutable_info()->set_error(true);
```
kevinkreiser, line 963: notice we also mark the request as having errored out so that later on we can send a stat that says the request was ok

src/worker.cc
```diff
@@ -1233,12 +1089,93 @@ worker_t::result_t to_response(const std::string& data,
...
+                             conf.get<uint64_t>("statsd.batch_size", 500),
+                             0) {
+    auto added_tags = conf.get_child_optional("statsd.tags");
+    if (added_tags) {
+      for (const auto& tag : *added_tags) {
+        tags.push_back(tag.second.data());
+      }
+    }
+  }
+  std::vector<std::string> tags;
```
kevinkreiser, line 1110: at the moment the only point of this pimpl class is to store the tags from config. in the future we could add more custom stuff as we need it for other logging purposes

src/worker.cc
```diff
@@ -1233,12 +1089,93 @@ worker_t::result_t to_response(const std::string& data,
...
+    : interrupt(nullptr), statsd_client(new statsd_client_t(config)) {
 }
 service_worker_t::~service_worker_t() {
 }
 void service_worker_t::set_interrupt(const std::function<void()>* interrupt_function) {
   interrupt = interrupt_function;
 }
+void service_worker_t::cleanup() {
+  // sends metrics to statsd server over udp
+  statsd_client->flush();
```
kevinkreiser, line 1123: this is what actually sends the metrics to statsd. notice that its in the cleanup method so it doesnt block the request flow its specifically done when the worker is not actively processing a request

src/worker.cc
```diff
@@ -1233,12 +1089,93 @@ worker_t::result_t to_response(const std::string& data,
...
+}
+void service_worker_t::enqueue_statistics(Api& api) const {
```
kevinkreiser, line 1125: this just translates all of the pbf flavored stats into statsd messages

src/worker.cc
kevinkreiser: yeah im missing a `!`

src/worker.cc
```diff
@@ -1233,12 +1089,93 @@ worker_t::result_t to_response(const std::string& data,
...
+                      ? ".loki."
+                      : (typeid(*this) == typeid(thor::thor_worker_t) ? ".thor." : ".odin.");
+    const auto& action = Options_Action_Enum_Name(api.options().action());
+
+    auto* stat = api.mutable_info()->mutable_statistics()->Add();
+    stat->set_key(action + worker + "latency_ms");
+    stat->set_value(e);
+    stat->set_type(timing);
+  });
+}
```
kevinkreiser, line 1179: ported the `scoped_timer` stuff to use the `Finally` stuff. i made it a member function so it could use `this` to do type checkign to figure out what kind of worker it was, which precluded it from being called from within serialization (hence the change to odin_worker::narrate)

scripts/valhalla_build_config
```diff
@@ -534,6 +541,13 @@ help_text = {
     'max_timedep_distance': 'Maximum b-line distance between locations to allow a time-dependent route',
     'max_alternates': 'Maximum number of alternate routes to allow in a request',
     'max_exclude_polygons_length': 'Maximum total perimeter of all exclude_polygons in meters'
+  },
+  'metrics': {
+    'host': 'The statsd host address',
+    'port': 'The statsd port',
+    'prefix': 'The statsd prefix to use for each metric',
+    'batch_size': 'Approximate maximum size in bytes of each batch of stats to send to statsd',
```
PureW, line 549: How does udp vs tcp factor in here? If we're shipping to a udp socket, I imagine you dont want to go above the MTU size (1476 byte?).

src/tyr/actor.cc
PureW: Hmm, it'd be interesting to know serialization cost separately, especially if we're exploring alternate response formats like pbf in the future. Haven't thought too much this change yet though nor have strong opinions on the change.

src/worker.cc
```diff
@@ -1120,6 +958,24 @@ std::string jsonify_error(const valhalla_exception_t& exception, const Api& requ
...
+
+  // write a few stats about the error
+  request.mutable_info()->set_error(true);
+  auto worker = exception.code < 200 || (exception.code >= 300 && exception.code < 400)
+                    ? ".loki."
+                    : (exception.code >= 400 && exception.code <= 500 ? ".thor." : ".odin.");
+  const auto& action = Options_Action_Enum_Name(request.options().action());
+
+  auto* err_stat = request.mutable_info()->mutable_statistics()->Add();
+  err_stat->set_key(action + worker + exception.statsd_key);
```
PureW, line 970: I don't know how generic this is, but at least Cloudwatch warns about creating high-cardinality metrics due to cost. It seems like this creates a different metric per exception which _may_ be costly in certain logging backends.

src/worker.cc
PureW: > but it does double the amount of reporting we do. should we combine them into one metric?

I'd lean towards combining them. Related to my comment at https://github.com/valhalla/valhalla/pull/3191/files?diff=unified&w=1#r666469902

src/worker.cc
PureW: And this will result in multiple sends if the total batched payload is larger than the configured batch-size?

scripts/valhalla_build_config
kevinkreiser: it only supports udp. max size for ethernet mtu is about 1500 as you suggest. i am setting the default to 1024 to stay below that thought i highly doubt that we'll ever send so many metrics in one go that we'd exceed that

src/worker.cc
kevinkreiser: it does however i expect the number of exceptional requests to remain relatively low. if that turns out not to be the case, what we can do is make use of the "frequency" option which only sends a percentage of the total metrics through to the backend. that should be doable pretty easily so i can just add it right now if you want!

src/tyr/actor.cc
kevinkreiser: i thought the same but didnt want to bother any more with crazy refactorings. now that i have it working i can take a second look at this

src/worker.cc
kevinkreiser: me too. let me think about the best way to do that and propose a change

src/worker.cc
kevinkreiser: correct, more information is available in the library documentation: https://github.com/vthiery/cpp-statsd-client#batching

src/worker.cc
PureW: Not fully clear on how the frequency param would work. The cost I'm thinking of is per metric itself, and not per "publish" or "flush".

For example, if we end up pushing a total of 1000 different statsd keys, regardless of it's 1 or 100k samples pushed, that'd cost $300 per month in AWS as they charge $0.30 per metric and month

https://aws.amazon.com/cloudwatch/pricing/

I haven't counted the number of possible keys in this PR but it's something to be aware of I think. With that cost-calculation in mind, maybe something like a "blacklist" rather than a frequency param would be helpful, where the blacklist could be used for disabling certain groups of metrics that aren't of interest?

kevinkreiser: ah yeah i didnt go check the docs on aws, i had figured it was by total number of metrics rather than by total number of unique keys. but yeah we could blacklist keys or we could prefix the exceptions with `error_`  and cut off everythign after the `error` part and collapse all exceptions into one key. it kind of neuters your ability to tell anything about what is going on though

kevinkreiser: again i didnt read the aws docs (because they always give me headaches) but i think realistically only a very small percentage of these metrics will come out of a running service. you can grep your logs to see what they are. iirc its about 5 to 10 unique ones

nilsnolde (approved):
:+1:

kevinkreiser:
I had 2 things I wanted to do yet. I checked that it was working and that seems to be the case. I'll post a screen shot of the statsd console logging backend. The stuff left to do is:

1. Run perf test to make sure it's not costly. I assume it's not plus it's optional
2. Unit tests. I wrote a mock server just for this

kevinkreiser:
first question is answered, there was zero impact on performance with it enabled vs master with it non-existant. now ill add unit tests

kevinkreiser:
Ok i have a unit test in there now that tracks all of the error and warns work properly. I also added a metric for worker start (which is useful to see if things are crashing)

The only TODO left is breaking out serialization into its own metric. I'll see if I can figure a decent way to do that

kevinkreiser:
@nilsnolde i n ever gave you a direct answer on this:

> Btw, how much overhead does this approx add for a typical routing request when it's purging the statsd request(s) at the end? Or is that somehow async and it's not a blocking POST request? I have little feeling how many stats would be typically collected now.

The way i've implemented it is that we only send the stats to statsd after the response has been returned to the requester. so no latency is added on the individual request but rather between requests. if you have a huge request backlog (because you are underprovisioned) it will most certainly cost you something.

i did run my usual 14k routes performance test at 24 threads and there was no measurable difference in performance, t hough holy hell we dropped about 25% performance over the last 6 months or so due to algorithmic changes (this test skews toward longer routes though so maybe this is unfair as a comparison).

nilsnolde (approved):
almost makes me want to set up a statsd instance;)

kevinkreiser:
> almost makes me want to set up a statsd instance;)

just for kicks you can test it out really easy locally. you just need to clone statsd and then run with the console logging backend: 

```
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/kevinkreiser/statsd$ cat config.js 
{
  port: 8125,
  backends: ["./backends/console"],
  debug: true
}
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/kevinkreiser/statsd$ node stats.js config.js 
15 Jul 16:04:14 - [805237] reading config file: config.js
15 Jul 16:04:14 - DEBUG: Loading server: ./servers/udp
15 Jul 16:04:14 - server is up INFO
15 Jul 16:04:14 - DEBUG: Loading backend: ./backends/console
Flushing stats at  Thu Jul 15 2021 16:04:24 GMT-0400 (GMT-04:00)
{ counters:
   { 'statsd.bad_lines_seen': 0,
     'statsd.packets_received': 0,
     'statsd.metrics_received': 0 },
  timers: {},
  gauges: {},
  timer_data: {},
  counter_rates:
   { 'statsd.bad_lines_seen': 0,
     'statsd.packets_received': 0,
     'statsd.metrics_received': 0 },
  sets: {},
  pctThreshold: [ 90 ] }
```

PR #3194, opened 2021-07-06
Fix avoid polygons again

### LINKED ISSUE #3182: Avoid polygons
The algorithm can't calculate a path, however when I take a polygon it can calculate the path here, everything normal if this polygon affected the path, I leave the example images here: 

Image with Polygons:
notwork

Image without Polygons:
work

And my request:

{
	"directions_type": "maneuvers",
	"avoid_polygons": [
		[
			[-8.899850490212819, 39.13892207569951],
			[-8.899855854630864, 39.13892415606849],
			[-8.899920227647314, 39.13891167385339],
			[-8.900242092729133, 39.14053226293175],
			[-8.900164308667662, 39.14054474485951],
			[-8.900000693917733, 39.139745897023865],
			[-8.899829032540767, 39.138930397175216],
			[-8.899850490212819, 39.13892207569951]
		],
		[
			[-8.90089085203469, 39.13890553686528],
			[-8.900890852034703, 39.138823362220464],
			[-8.901337439835736, 39.138843125751414],
			[-8.901637847245356, 39.13839480429094],
			[-8.90171026688879, 39.13841248755808],
			[-8.901405836165658, 39.13892113963512],
			[-8.900937790692456, 39.1389044966805],
			[-8.90089085203469, 39.13890553686528]
		]
	],
	"costing": "auto",
	"locations": [{
		"name": "startPlace",
		"lon": -9.1485128705079,
		"type": "break_through",
		"lat": 39.188744520260286
	}, {
		"min_road_class": "residential",
		"name": "firstStopPoint",
		"lon": -8.90767222222222,
		"type": "break",
		"lat": 39.1243722222222,
		"preferred_side": "opposite"
	}, {
		"name": "enPlace",
		"lon": -9.1485128705079,
		"type": "break_through",
		"lat": 39.188744520260286
	}],
	"language": "pt-PT",
	"costing_options": {
		"auto": {
			"service_penalty": 0,
			"use_highways": 1,
			"use_tolls": 1,
			"top_speed": 120,
			"use_tracks": 0,
			"use_living_streets": 0,
			"shortest": true
		}
	}
}

Is there any way to overcome this?
Thanks

nilsnolde:
Fixes #3182 

The issue was a little subtle: we register `std::vector<PointLL>` as a `bg::polygon` so the edge's `shape` was interpreted as a polygon geometry and then used to intersect with the avoid_polygons, which I guess is sorta undefined behavior (at least for me).

I also added a `to_geojson()` for debugging as you suggested @kevinkreiser, as a trace log, very helpful..

~~Could be done a bit different, by registering another structure with bg, e.g. `std::deque<PointLL>`, to avoid casting every edge to a linestring. With a few experiments I didn't see any significant performance deviation from master though, but let me know if you'd prefer that.~~ that only works if we register std::vector<PointLL> as linestring, of course bg::intersects needs two geometry types. For polygons we wouldn't need to register our own geometry type, we could use bg::model::polygon, but the Tiles::Intersect method somehow won't take that type as an argument.

### DIFF
--- a/src/loki/polygon_search.cc
+++ b/src/loki/polygon_search.cc
@@ -37,6 +37,7 @@ ring_bg_t PBFToRing(const valhalla::Options::Ring& ring_pbf) {
   for (const auto& coord : ring_pbf.coords()) {
     new_ring.push_back({coord.lng(), coord.lat()});
   }
+  // fixes windedness & closes open rings
   bg::correct(new_ring);
   return new_ring;
 }
@@ -91,7 +92,7 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Options_Ring>&
   for (const auto& ring_pbf : rings_pbf) {
     rings_bg.push_back(PBFToRing(ring_pbf));
     const ring_bg_t ring_bg = rings_bg.back();
-    rings_length += bg::perimeter(ring_bg);
+    rings_length += bg::perimeter(ring_bg, Haversine());
   }
   if (rings_length > max_length) {
     throw valhalla_exception_t(167, std::to_string(max_length));
--- a/src/loki/polygon_search.cc
+++ b/src/loki/polygon_search.cc
@@ -2,8 +2,10 @@
 #include <boost/geometry/geometries/register/point.hpp>
 #include <boost/geometry/geometries/register/ring.hpp>
 
+#include <valhalla/baldr/json.h>
 #include <valhalla/loki/polygon_search.h>
 #include <valhalla/midgard/constants.h>
+#include <valhalla/midgard/logging.h>
 #include <valhalla/midgard/pointll.h>
 #include <valhalla/worker.h>
 
@@ -23,6 +25,7 @@ namespace {
 // register a few boost.geometry types
 using line_bg_t = bg::model::linestring<vm::PointLL>;
 using ring_bg_t = std::vector<vm::PointLL>;
+using namespace vb::json;
 
 // map of tile for map of bin ids & their ring ids
 // TODO: simplify the logic behind this a little
@@ -47,6 +50,40 @@ double GetRingLength(const ring_bg_t& ring) {
   auto length = bg::length(line, Haversine());
   return length;
 }
+
+#ifdef LOGGING_LEVEL_TRACE
+// serializes an edge to geojson
+std::string to_geojson(const std::unordered_set<vb::GraphId>& edge_ids, vb::GraphReader& reader) {
+  auto features = array({});
+  for (const auto& edge_id : edge_ids) {
+    auto tile = reader.GetGraphTile(edge_id);
+    auto edge = tile->directededge(edge_id);
+    auto shape = tile->edgeinfo(edge).shape();
+    if (!edge->forward()) {
+      std::reverse(shape.begin(), shape.end());
+    }
+
+    auto coords = array({});
+    for (const auto& p : shape) {
+      coords->emplace_back(array({fixed_t{p.lng(), 6}, fixed_t{p.lat(), 6}}));
+    }
+    features->emplace_back(
+        map({{"type", std::string("Feature")},
+             {"properties",
+              map({{"shortcut", edge->is_shortcut() ? std::string("True") : std::string("False")},
+                   {"edge_id", edge_id.value}})},
+             {"geometry", map({{"type", std::string("LineString")}, {"coordinates", coords}})}}));
+  }
+
+  auto collection =
+      vb::json::map({{"type", std::string("FeatureCollection")}, {"features", features}});
+
+  std::stringstream ss;
+  ss << *collection;
+
+  return ss.str();
+}
+#endif // LOGGING_LEVEL_TRACE
 } // namespace
 
 namespace valhalla {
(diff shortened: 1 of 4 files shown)


src/loki/polygon_search.cc
```diff
@@ -121,9 +158,11 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Options_Ring>&
         // TODO: some logic to set percent_along for origin/destination edges
         // careful: polygon can intersect a single edge multiple times
         auto edge_info = tile->edgeinfo(edge);
+        // vector<PointLL> is reserved for bg as ring, need to cast to linestring
+        line_bg_t edge_line{edge_info.shape().begin(), edge_info.shape().end()};
         bool intersects = false;
         for (const auto& ring_loc : bin.second) {
-          intersects = bg::intersects(rings_bg[ring_loc], edge_info.shape());
+          intersects = bg::intersects(rings_bg[ring_loc], edge_line);
```
nilsnolde, line 165: that's the actual change

nilsnolde:
@PureW maybe this also fixes the clang issues over at https://github.com/valhalla/valhalla/pull/2750#issuecomment-871672316 ?

src/loki/polygon_search.cc
```diff
@@ -41,16 +37,10 @@ ring_bg_t PBFToRing(const valhalla::Options::Ring& ring_pbf) {
   for (const auto& coord : ring_pbf.coords()) {
     new_ring.push_back({coord.lng(), coord.lat()});
   }
+  bg::correct(new_ring);
```
kevinkreiser, line 40: whats that do? fix winding or self intersection?

src/loki/polygon_search.cc
```diff
@@ -101,7 +91,7 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Options_Ring>&
   for (const auto& ring_pbf : rings_pbf) {
     rings_bg.push_back(PBFToRing(ring_pbf));
     const ring_bg_t ring_bg = rings_bg.back();
-    rings_length += GetRingLength(ring_bg);
+    rings_length += bg::perimeter(ring_bg);
```
kevinkreiser, line 94: how does this know to use spherical coordinates and to return meters?

kevinkreiser: oh probably because of `geographic`?

src/loki/polygon_search.cc
```diff
@@ -123,7 +150,8 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Options_Ring>&
         auto edge_info = tile->edgeinfo(edge);
         bool intersects = false;
         for (const auto& ring_loc : bin.second) {
-          intersects = bg::intersects(rings_bg[ring_loc], edge_info.shape());
+          intersects = bg::intersects(rings_bg[ring_loc],
+                                      line_bg_t(edge_info.shape().begin(), edge_info.shape().end()));
```
kevinkreiser, line 154: > avoid_polygons intersected edges as polygons instead of linestrings

so previously it was treating the edge shape as a polygon rather than a line? treating it as a line should be slightly cheaper but i dont understand how that would make the difference for having an intersection and not. maybe it has to do with those edge shapes making self intersecting rings and the self intersecting parts not counting for intersection?

PureW:
> @PureW maybe this also fixes the clang issues over at [#2750 (comment)](https://github.com/valhalla/valhalla/pull/2750#issuecomment-871672316) ?

Initially getting one error
```
../src/loki/polygon_search.cc:31:19: error: unused variable 'Haversine' [-Werror,-Wunused-const-variable]
static const auto Haversine = [] {
                  ^
1 error generated.
[17/81] Building CXX object src/meili/CMakeFiles/valhalla-meili.dir/map_matcher.cc.o
```

Fixing that, I still get the error in the unittest:

```
[ RUN      ] AvoidTest.TestAvoidShortcutsTruck                                                                                         ../test/gurka/test_avoids.cc:214: Failure                        
  avoid_edges.size()                                                                                                                   
    Which is: 0                                                                                                                        
  4                                                                                                                                    [  FAILED  ] AvoidTest.TestAvoidShortcutsTruck (5 ms)
```

src/loki/polygon_search.cc
nilsnolde: jep, our rings are expected as clockwise so it fixes counterclockwise rings. also closes open rings if the rings are typed that way (that we do ourselves actually while parsing the request, could remove that tiny bit now). 

I don't remember why, but I had `correct()` in an early version before merging the initial PR and I took it out. couldn't find any commit or so to figure out why I took it out. but since it compiles and tests pass, I'll leave it in (and remove the bit where we close the polygons in the request).

src/loki/polygon_search.cc
nilsnolde: oops, I should've specified the `Haversine` function above, forgot that. But apparently it does some deductions from the underlying point type, no idea what that exactly defaults too though. So rather be explicit and put the Haversine strategy there.

src/loki/polygon_search.cc
nilsnolde: so the `ei.shape` has `std::vector<PointLL>` type which we register with bg as ring type so that's how it's treated in the bg::intersects call. But of course as an open ring which is the "undefined behavior" part for me.

nilsnolde:
@kevinkreiser thanks for the review, I added haversine perimeter calculation and removed our own ring closing logic when parsing the request.

@PureW the unused-variable warning is gone, forgot to add that function as a strategy.. No idea what to make of that failing test, maybe I'll try clang one of these days

src/loki/polygon_search.cc
nilsnolde: When I printed out the avoided edges from a polygon covering only one way I got back like 20 edges before this change. No idea what bg does internally in a situation like that. Now it only prints those 2 DE the polygon really intersects.

kevinkreiser:
this looks good to me. are we saying that compiling with clang fails the tests though?

nilsnolde:
> clang fails the tests

seems so @kevinkreiser . though that's not new with this PR, @PureW reported it a couple of weeks back in that initial PR. Must've somehow been introduced with clang 12.0, no idea what's so compiler-specific about that test.. I'll open another issue about it.

PureW:
> this looks good to me. are we saying that compiling with clang fails the tests though?

Yea, [I'm seeing this](https://github.com/valhalla/valhalla/pull/2750#issuecomment-871672316) I haven't dug into why this is though.

kevinkreiser (approved):

PR #3198, opened 2021-07-08
add dynamic to kRoadClassFactor

huggymann:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/sif/motorscootercost.cc
+++ b/src/sif/motorscootercost.cc
@@ -68,7 +68,7 @@ constexpr float kDestinationOnlyFactor = 0.2f;
 // roads. These penalties are modulated by the road factor - further
 // avoiding higher class roads for those with low propensity for using
 // primary roads.
-constexpr float kRoadClassFactor[] = {
+float kRoadClassFactor[] = {
     1.0f,  // Motorway
     0.5f,  // Trunk
     0.2f,  // Primary
@@ -361,6 +361,25 @@ MotorScooterCost::MotorScooterCost(const CostingOptions& costing_options)
   // start to increase the differences.
   float use_primary = costing_options.use_primary();
   road_factor_ = (use_primary >= 0.5f) ? 1.5f - use_primary : 3.0f - use_primary * 5.0f;
+  
+  kRoadClassFactor[1] = 1.0f - use_primary * 0.2f;
+  kRoadClassFactor[2] = 1.0f - use_primary * 0.8f;
+  kRoadClassFactor[3] = 0.4f + (1 - use_primary) * 0.2f;
+  kRoadClassFactor[4] = 0.0f + use_primary * 0.5f;
+  kRoadClassFactor[5] = 0.1f + use_primary * 0.5f;
+  kRoadClassFactor[6] = 0.15f + use_primary * 0.85f;
+  kRoadClassFactor[7] = 0.2f + use_primary * 0.8f;
+
+/*
+    1.0f,  // Motorway       0
+    0.95f,  // Trunk         1
+    0.8f,  // Primary        2
+    0.4f,  // Secondary      3
+    0.05f, // Tertiary       4
+    0.05f, // Unclassified   5
+    0.0f,  // Residential    6
+    0.2f   // Service, other 7
+*/
 }
 
 // Check if access is allowed on the specified edge.
@@ -435,7 +454,7 @@ Cost MotorScooterCost::EdgeCost(const baldr::DirectedEdge* edge,
 
   float speed_penalty = (speed > top_speed_) ? (speed - top_speed_) * 0.05f : 0.0f;
   float factor = 1.0f + (density_factor_[edge->density()] - 0.85f) +
-                 (road_factor_ * kRoadClassFactor[static_cast<uint32_t>(edge->classification())]) +
+                 2* (road_factor_ * kRoadClassFactor[static_cast<uint32_t>(edge->classification())]) +
                  grade_penalty_[static_cast<uint32_t>(edge->weighted_grade())] + speed_penalty;
 
   if (edge->destonly()) {


huggymann:
is it clear what i want to achieve in my code? I am sure this is not the best way to be programmed. 

What are the next steps here?

kevinkreiser:
@huggymann i just need to find time to look at it and go over our comments in the other issue. it just takes time to fit it in. there are lots of other prs and things we are looking at simultaneously so it can sometimes take a while if something is in a very raw form

huggymann:
Thanks man! :)

src/sif/motorscootercost.cc
```diff
@@ -435,7 +454,7 @@ Cost MotorScooterCost::EdgeCost(const baldr::DirectedEdge* edge,
 
   float speed_penalty = (speed > top_speed_) ? (speed - top_speed_) * 0.05f : 0.0f;
   float factor = 1.0f + (density_factor_[edge->density()] - 0.85f) +
-                 (road_factor_ * kRoadClassFactor[static_cast<uint32_t>(edge->classification())]) +
+                 2* (road_factor_ * kRoadClassFactor[static_cast<uint32_t>(edge->classification())]) +
```
danpat, line 457: Could you move this `2` into a named constant so readers can have some hint as to why we're doubling here?

src/sif/motorscootercost.cc
```diff
@@ -68,7 +68,7 @@ constexpr float kDestinationOnlyFactor = 0.2f;
 // roads. These penalties are modulated by the road factor - further
 // avoiding higher class roads for those with low propensity for using
 // primary roads.
-constexpr float kRoadClassFactor[] = {
+float kRoadClassFactor[] = {
```
danpat, line 71: I'm not a fan of the runtime modification of this lookup table.  A better approach that would let us keep `constexpr` here, would be to create _two_ tables.

It looks like only `use_primary` affects the result, so we could have `kRoadClassFactorWhenPrimary` and `kRoadClassFactorOtherwise`, or something of that nature.

That would avoid the per-request modification of the kRoadClassFactor lookup table.

src/sif/motorscootercost.cc
huggymann: let me think of a good name. Can you do changes to the code as well?

src/sif/motorscootercost.cc
huggymann: Thank you very much for your feedback. I am not the programmer expert and am more "digging" through the code. Sorry! 

as i mentioned above, every feedback is VERY welcome. Can you do some changes as well to the code as well?

huggymann:
Could anyone please help me on this? i am not a programmer expert but still, would like this to be solved.

nilsnolde:
personally I'm semi-interested in this costing model, but I could look into it for you as part of a commercial request. sorry folks, I hate doing it like this (unfortunately no public email on your account @huggymann ), but you can contact me on enquiry at gis-ops dot com. I can't promise any timeline for this to get merged, but at least it could make the PR more reviewable and less work for the main maintainers who will need to approve eventually.

kevinkreiser:
@huggymann apologies but i havent had much time lately. when something is more or less clear to me then i usually just answer immediately but for this issue it is not the case. for this issue i will need to modify the code to solve the issue of optionally disabling hierarchy culling, and that will take a lot more time investment on my part, which as i said, i dont have a lot of at the moment. i appreciate the digging you have done to date but i'm not able to commit to a timeline of when i can look deeper at this.

when there are 'free' hours at night i will make a PR for optionally disabling the hierarchy limits. that should unblock this pr functionally anyway

huggymann:
@nilsnolde Thank you for the offer but I guess the open source/community based approach does not fit into payed development here. I assumed, that I would make it myself with my little C programming skills, but as Kevin sayed, a little more need to be touched here. 

@kevinkreiser Thank you for the explanation. I did not know how much effort this would be before i started this change and i hoped that a lot of poeple would be here waiting for new challenges ;) 
I tried many times now to understand the routines of hierarchy culling. I am willed to dig into it myself as well, but i feel like i need more explanation. For example I couldnt really get what happens with this "factor" here
https://github.com/valhalla/valhalla/blob/f0d402e9fb449795b6cb8a0cac0b7e786f4a999d/src/sif/motorscootercost.cc#L456
And why the routing engine is not taking smaller roads even if i priorize them in the factor because of the hierarchy culling. If you are willed, please give me some higher level explanation and i will dig into more myself.

kevinkreiser:
@huggymann id be happy to talk you through making hierarchy culling optional. the main thing to understand about it is that we do it as an optimization for runtime performance. the algorithm that we use basically searches from the origin and destination at the same time and finds a meeting point in the middle. a part of the search tree gets further away from those start/end points it looks at less and less of the road network. the less we look at the faster the algorithm performs but of course then the more of the road network that we can't use.

this heuristic works pretty well for motorized vehicles interested in doing longer trips but i could see medium sized motor vehicles like motorcycles and scooters more interested in less highway routing in general even on longer trips. i was proposing allowing to disable it all together but maybe we should just have different limits for different costing models (maybe we already do, id have to check..). anyway i should write up a proposal in the other issue that you opened. once we get this one out of the way you can more easily reason about the values  you want to tweak in this pr.

i should also mention this is just my hypothesis, i havent verified any of it yet so i could be prescribing the wrong solution here. thats the main problem i need a decent amount of time to really try it out and then to reason about what is going on.

huggymann:
Hello @kevinkreiser , 

I have been working on automating the build process of valhalla on my side to easily make and test changes. that is bringing the delay in between. 

I guess I am understanding the routine correctly, I am just struggeling with the fact that i cannot disable the hierarchy limits anymore. Thus, the changes i make in routing will not have any effect. 

From other issue i have this message: 

> @huggymann you have to "fork" the repository into your own github account, make your changes on the fork and then pull request the fork's branch back to this repository.

>     i changed again to return false all the time but with your new changes it doesnt work

> thats right, we made some changes to that recently. i think it would be best to allow the requester to disable hierarchy culling via a request parameter. i can help make this option a possibility once you get your fork PR'd

You already approved the fact that setting the line to false will not disable it. I also enlarged the distances to nearly infinite, no change. Could you please let me know how can I disable it for testing?

kevinkreiser:
@huggymann yes sorry for the wait. disabling heirarchy culling is something i can help with tonight or so i can make another branch with the changes for illustration. i will link it here. if its working for you then i can turn that into a pr and we can figure out a way to make it a configuration option or request option

huggymann:
Oh wow. That would be awesome ! :) I spent so many hours building/testing/trying. I thought i am too stupid.
I will be desperately waiting :D

huggymann:
> 
> 
> @huggymann yes sorry for the wait. disabling heirarchy culling is something i can help with tonight or so i can make another branch with the changes for illustration. i will link it here. if its working for you then i can turn that into a pr and we can figure out a way to make it a configuration option or request option

Is that a lot of changes to be made to disable it? Is there anything i can help you with?

I just thought that hierarchy culling does not necessarily need to be disabled by user as an option. But maybe by the code-routines for example in pedestrian bike or scooter, where the distances wont be that far as with a car.

kevinkreiser:
@huggymann it is already done in pedestrian and bike, but not in scooter if i am remembering correctly. perhaps you can look at how those costings achieve that to try an implmentation or even just a hack in yoru branch to get what you are looking for:

https://github.com/valhalla/valhalla/blob/master/src/sif/pedestriancost.cc#L551-L554

PR #3199, opened 2021-07-08
Fix Mac Builds

### LINKED ISSUE #3036: Update homebrew to avoid downloads from bintray (which is shutting down)
[bintray is shutting down](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/). `homebrew` has already [moved off of it](https://brew.sh/2021/04/12/homebrew-3.1.0/).

However, our osx executor is using an older version, which is causing failures of osx builds.

We should update homebrew before using it and also upgrade to the [newest xcode](https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions) version available from circle.

kevinkreiser:
fixes #3036 

for some reason homebrew doesnt like the way the cli build tools look on mac and it borks the install of prime server from the brew formula. so i say forget  homebrew and will just install it manually. long term we should probably submodule prime_server but that does mean we'll have a bit of an odd situation when we want the binary targets that prime_server creates. so i think ill hold off on that for a little while..

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,7 @@ commands:
     steps:
       - run: brew install protobuf cmake ccache libtool boost libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
       - run: pip3 install requests
+      - run: git clone https://github.com/kevinkreiser/prime_server --recurse-submodules && cd prime_server && ./autogen.sh && ./configure && make -j4 && make install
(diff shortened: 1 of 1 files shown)


PureW (approved):

kevinkreiser:
just fyi i also re-enabled osx builds as a required check, they were temporarily disabled while the build was broken

PR #3204, opened 2021-07-10
Fix Fast Route Rejection

### LINKED ISSUE #3202: Fast Fail Impossible Routes Seems Broken
Recently I tried to do this:
image

This should result in an immediate 400 and a message about how the locations are in disconnected regions however i saw this:
image

This request should never make it to thor so long as loki is using the connectivity map so i checked my config and:
image

It looks like it should be!

kevinkreiser:
fixes #3202 

@dnesbitt61 was right switching to double fixed all instances of this. i took all of the shapes that i could find that exposed the issue from an older planet, added them to the unit test, they failed in that they were thought to intersect hundreds of tiles. i changed to double precision and then they all passed.

~i still want to add a protection in the algorithm against this so we dont have to worry about it in the future.~ ive done that

as for an explanation, here is the first unit test case:

image

in the debugger it looks like this:

image

so you can see that x1 ends up just slightly smaller than x0 and y0 and y1 are equal, which means the the line is traveling only left/west. looking at the loop that does the "rasterization" you can see that it will stop when the pixel we iterate ends at the last discretized (floored) coordinate of the line. 

floor of y and y1 are already equal, which makes perfect sense because we are just moving in the x direction (the line is horizontal). so the loop should stop when we go far enough to the left to hit that last line coordinate, ie when floor of x and x1 are equal. given that they are so close together (much closer than a tiles distance) that should happen in one iteration.

stepping further we see something very unfortunate. if you look at the calculations for triangle areas (tx and ty)  you'll see that the right hand sides of both equations cancel out (because we arent moving in the y direction at all). this leaves two triangle areas with different signs because they are formed on different sides of the line but they have the same area. then we hit the if block that says should we move in the x or the y. since they are the same the tie breaker is the else clause which just happens to pick the y direction, which is wrong in this case.

now lets see what happens when we go to double precision:

image

look at y0 and y1. you will see that now we have the slightest upward slope. this leads to the triangle areas not being equal and us not having to handle that case. basically single precision float quantized the y values to be perfectly horizontal which lead to us hitting the unhandled case of what to do when the areas are equal.

i think its pretty clear by now we need to handle the case when the triangle areas are equal. now there are cases where it doesnt matter which one we choose. thinking of a line that is perfectly diagonal, going in the x vs the y would create equal areas but we dont actually care who wins as long as its consistent. however in the case of purely horizontal or purely vertical lines we need to move in the direction of the line only. 

i will add tests that show this breaking for double precision and then add code to handle the equal area case.

its actually slightly more complex than this, turns out for spherical coords we dont need to add code to handle the equal area case because we wont ever send a perfectly horizontal line that covers more than 2 subdivisions to the bresenham algorithm, for more info see this comment: https://github.com/valhalla/valhalla/pull/3204#discussion_r667413227

### DIFF
--- a/src/midgard/tiles.cc
+++ b/src/midgard/tiles.cc
@@ -17,20 +17,20 @@ namespace {
 // at each step it decides to either move in the x or y direction based on which pixels midpoint
 // forms a smaller triangle with the line. to avoid edge cases we allow set_pixel to make the
 // the loop bail if we leave the valid drawing region
-void bresenham_line(float x0,
-                    float y0,
-                    float x1,
-                    float y1,
+void bresenham_line(double x0,
+                    double y0,
+                    double x1,
+                    double y1,
                     const std::function<bool(int32_t, int32_t)>& set_pixel) {
   // this one for sure
   bool outside = set_pixel(std::floor(x0), std::floor(y0));
   // steps in the proper direction and constants for shoelace formula
-  float sx = x0 < x1 ? 1 : -1, dx = x1 - x0, x = std::floor(x0) + .5f;
-  float sy = y0 < y1 ? 1 : -1, dy = y1 - y0, y = std::floor(y0) + .5f;
+  double sx = x0 < x1 ? 1 : -1, dx = x1 - x0, x = std::floor(x0) + .5f;
+  double sy = y0 < y1 ? 1 : -1, dy = y1 - y0, y = std::floor(y0) + .5f;
   // keep going until we make it to the ending pixel
   while (std::floor(x) != std::floor(x1) || std::floor(y) != std::floor(y1)) {
-    float tx = std::abs(dx * (y - y0) - dy * ((x + sx) - x0));
-    float ty = std::abs(dx * ((y + sy) - y0) - dy * (x - x0));
+    double tx = std::abs(dx * (y - y0) - dy * ((x + sx) - x0));
+    double ty = std::abs(dx * ((y + sy) - y0) - dy * (x - x0));
     // less error moving in the x
     if (tx < ty) {
       x += sx;
--- a/src/midgard/tiles.cc
+++ b/src/midgard/tiles.cc
@@ -32,10 +32,9 @@ void bresenham_line(double x0,
     double tx = std::abs(dx * (y - y0) - dy * ((x + sx) - x0));
     double ty = std::abs(dx * ((y + sy) - y0) - dy * (x - x0));
     // less error moving in the x
-    if (tx < ty) {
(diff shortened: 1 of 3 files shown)


PureW (dismissed):
👍 any idea more specifically what breaks with the lower precision?

kevinkreiser:
@PureW i updated the description above to answer that question

gknisely:
@kevinkreiser don't forget the changelog

src/midgard/tiles.cc
```diff
@@ -17,25 +17,24 @@ namespace {
...
+  double sy = y0 < y1 ? 1 : -1, dy = y1 - y0, y = std::floor(y0) + .5f;
   // keep going until we make it to the ending pixel
   while (std::floor(x) != std::floor(x1) || std::floor(y) != std::floor(y1)) {
-    float tx = std::abs(dx * (y - y0) - dy * ((x + sx) - x0));
-    float ty = std::abs(dx * ((y + sy) - y0) - dy * (x - x0));
+    double tx = std::abs(dx * (y - y0) - dy * ((x + sx) - x0));
+    double ty = std::abs(dx * ((y + sy) - y0) - dy * (x - x0));
     // less error moving in the x
-    if (tx < ty) {
+    if (tx < ty || (tx == ty && y0 == y1)) {
```
kevinkreiser, line 35: this is the extra layer of protection here. i put a bunch of comments on the two sets of new test cases i added that explains what is going on. TLDR; moving to double precision makes it so that this line here isnt needed for spherical coordinates (which is really the only thing we care about). this is because i have code below that before calling this function first resamples (adds more coordinates) along the geodesic between two polar coordinates in an edge shape. in practice, when doing this it makes it so that this bresenham method will never be called on a fully horizontal set of points. however i kept the fix because for planer (cartesian) coordinates we do not resample (since lines in the plane dont curve).

the root problem was mixing double and float. in the (`Intersect`) method that calls this method, we check for some easy cases, if two points are in the same bin we mark the bin and exit, if they are adjacent we mark the two and exit. the shapes we found were mostly horiztonal but ever so slightly had differences in the y coordinate. so when you looked at them with double precision they would, for example when looking at a quad of 4 bins, start in the bottom left bin and end in the upper right. this is not one of the simple cases, so it needs to come to bresenham to be rasterized. however when it gets here and we do the math in single precision float it thinks that the line is moving only horizontally which causes this tie breaker where it defaults to moving in the y direction.

kevinkreiser:
This is ready for merge, just waiting on a ship. please have a read of my comments on the pr but also in the test file.

dnesbitt61 (approved):

PR #3205, opened 2021-07-12
Fix improper iterator usage in ManeuversBuilder

SiarheiFedartsou:
# Issue
I found it with a help of `-D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC`: https://kristerw.blogspot.com/2018/03/detecting-incorrect-c-stl-usage.html
There is similar flag in [libc++](https://libcxx.llvm.org//DesignDocs/DebugMode.html).

Was trying to enable it on CI, but it was too difficult for me to implement since I'm limited in the time.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/odin/maneuversbuilder.cc
+++ b/src/odin/maneuversbuilder.cc
@@ -564,6 +564,7 @@ void ManeuversBuilder::Combine(std::list<Maneuver>& maneuvers) {
         // Update with no combine
         prev_man = curr_man;
         curr_man = next_man;
+        assert(next_man != maneuvers.end());
         ++next_man;
       }
 
@@ -2940,6 +2941,7 @@ void ManeuversBuilder::ProcessRoundabouts(std::list<Maneuver>& maneuvers) {
     // on to the next maneuver...
     prev_man = curr_man;
     curr_man = next_man;
+    assert(next_man != maneuvers.end());
     ++next_man;
   }
 }
@@ -3255,7 +3257,9 @@ void ManeuversBuilder::ProcessTurnLanes(std::list<Maneuver>& maneuvers) {
     // on to the next maneuver...
     prev_man = curr_man;
     curr_man = next_man;
-    ++next_man;
+    if (next_man != maneuvers.end()) {
+      ++next_man;
+    }
   }
 }
 
@@ -3631,6 +3635,7 @@ void ManeuversBuilder::CollapseMergeManeuvers(std::list<Maneuver>& maneuvers) {
     }
     // on to the next maneuver...
     curr_man = next_man;
+    assert(next_man != maneuvers.end());
     ++next_man;
   }
 }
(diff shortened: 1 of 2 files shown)


src/odin/maneuversbuilder.cc
```diff
@@ -3255,7 +3257,9 @@ void ManeuversBuilder::ProcessTurnLanes(std::list<Maneuver>& maneuvers) {
     // on to the next maneuver...
     prev_man = curr_man;
     curr_man = next_man;
-    ++next_man;
+    if (next_man != maneuvers.end()) {
```
SiarheiFedartsou, line 3260: loop condition is `while (curr_man != maneuvers.end())` here, i.e. there is a case when `next_man` points to end of the list.

dgearhart (changes requested):
If there is an issue we need to handle gracefully - therefore, we cannot use an `assert`
What is the test case that causes this issue?

PureW:
> If there is an issue we need to handle gracefully - therefore, we cannot use an `assert`
> What is the test case that causes this issue?

This is a debug assert @dgearhart It is only active on debug builds such as when testing. On release builds the only difference from master would be the change at https://github.com/valhalla/valhalla/pull/3205/files#r667782407

src/odin/maneuversbuilder.cc
kevinkreiser: i think it would help to explain what is really going on here. i dont think its undefined behavior to increment an iterator past the end but it is undefined behavior to deference it. we should note here that there is a some code here that checks `next_man` before using it, but it checks for equality `if (next_man != maneuvers.end())`. of course that check will fail when we iterate past the end and we'll dereference it and be screwed. maybe we should just change the other code to: `if (next_man < maneuvers.end())`?

mandeepsandhu: Yup, this fix looks correct since the loop is only checking validity of `curr_man` and not `next_man`.

mandeepsandhu (approved):

SiarheiFedartsou:
> If there is an issue we need to handle gracefully - therefore, we cannot use an `assert`
> What is the test case that causes this issue?

@dgearhart It will be difficult to write test case for this issue. We are incrementing std::list::end() here and AFAIR it is an undefined behaviour. I was able to find it with a help of `-D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC` flags([see](https://kristerw.blogspot.com/2018/03/detecting-incorrect-c-stl-usage.html)). And as @PureW mentioned assert won't have any impact on release build, it would only inform us if we unexpectedly start incrementing end iterator in other similar cases.

kevinkreiser (approved):

dgearhart (commented):
conflict with changelog

kevinkreiser (approved):

mloskot:
> I found it with a help of -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC

FYI, the very same issue could have been found with default Debug build using MSVC. I've just patched my local 3.1.2

Issue #1703, opened 2019-02-01
Serve Vector Tiles from Valhalla Data


kevinkreiser:
We should have more tools for visually inspecting the valhalla data and ways to visually understand how sif/odin interact with it.

We currently have a tool that lets us see most of this data in json form however, reading enourmous amounts of properties in plain text is quite tedious and error prone.

What we propose is a vector tile server that will generate [mapbox vector tile spec encoded](https://github.com/mapbox/vector-tile-spec/tree/master/2.1) tiles. These are the defacto-standard when it comes to visualizing osm and derived datasets such as valhalla. This should allow for the creation of a visualization tool and accompanying styles to aide in debugging or just understanding how libvalhalla makes the decisions that it does when routing.

Guo-astro:
@kevinkreiser  Hi! Really interested in this, where is a good start point?

kevinkreiser:
i have an unpushed branch that has the request logic for this but none of the tile cutting logic yet (just the pbf definitions). i reckon i should write a more detailed plan here so that others can take up the work. give me a bit and ping back if i dont get back to it shortly!

ianthetechie:
@kevinkreiser I'd be curious at having a look into implementing this. As I was building the LSV profile (and unrelatedly investigating random complaints over the years haha), I've long wished for something like this to get better visibility into how the graph looks in specific areas / why paths weren't taken.

(The expansion endpoint is useful, but in very different ways.)

kevinkreiser:
yeah the locate endpoint also gives a slice of this data but not in the same slippy map way that we are used to. to me, serializing all the properties in mvt would be the best bet for doing this. there are some serious problems though:

1. at high zoom numbers everything is fine: its easy to create the geoms and properties for individual edges and everyone is happy
2. at low zooms life is hard: they start to become intractable around z12 and under or so and the only remedy is, as we do in the graph algorithm, cull the search space 
3. so we need to have a configuration to specify how that culling happens there are a few things to do:
  1. drop the least importance properties
  2. merge features with matching properties as much as possible
  3. drop features that are too small to be visible
  4. generalize the geometry of the remaining features appropriate to the target zoom level
4. all of that is just to make it usable on the client side (ie not too much data) we still have the problem that it will take a while to compute the tiles
5. so we need to add caching to disk and probably an optional pre-rendering (seeding) step that makes the low zoom tiles before we start a service (could build them when we build the routing tiles)
[...]

PR #3206, opened 2021-07-12
Tree view for Dijkstra

### LINKED ISSUE #1963: Street Network View with Valhalla
Hi,

I would like to know if there is a way to calculate a Street Network View (see image) by using Valhalla. 

image

Reading the documentation it seems that has not been implemented, but I can see that Valhalla uses many algorithms and maybe it is possible to have also this functionality.

Thank you

nilsnolde:
fixes #1963 

So far the `/expansion` endpoint is good for debugging the routing algos. I modified that action to also track Dijkstra's expansion to return all settled edges with their duration/distance. This PR is adds a request parameter `expansion_type` to distinguish between route debugging and isochrone requests, which will also lead to slightly different `properties` keys. We could also infer the type from the request format I guess, I personally like explicit better though.

This is still heavily draft, I just hacked it in yesterday to see how it performs (of course terrible;)). That's the things I'd like to sort out before finishing this up:
- [x] general structure ok for you? (i.e. hijacking `/expansion` for this)
- [ ] add to default `actions` in valhalla.json? I'd almost continue to not expose this endpoint by default (mostly analytical purpose) or we'd need another round of service limit keys only for expansion (which right now is only bound by the route and isochrone service limits)
- [x] output format: so far the expander is storing everything in `MultiLineString` and adding index-based property arrays (here e.g. `durations` & `distances`) for each sub-geometry. Save for pbf or other binary output that's probably the most "compact" one gets with ASCII stuff. I did a few experiments on the city of Vienna, Austria for 60 mins and `auto`:
      - 123 MB response in 18 secs (**no simplify, no encoding**)
      - 59 MB response in 17 secs (**simplify with `generalize` request param (or 21.4 default: ~ zoom level 15), no encoding**)
      - 30 MB response in 14 secs (**simplify with `generalize` request param (or 21.4 default: ~ zoom level 15), precision-5-encoded**)

~~I guess in the future this could use mapbox vector tiles spec when pbf output is real, IMO the second option (simplify but don't encode) would be OKish for now.~~ forget about that for now..

Opinions?


valhalla/thor/pathalgorithm.h
```diff
@@ -124,8 +124,8 @@ class PathAlgorithm {
    * @param  expansion_callback  the functor to call back when the algorithm makes progress
    *                             on a given edge
    */
-  using expansion_callback_t =
-      std::function<void(baldr::GraphReader&, const char*, baldr::GraphId, const char*, bool)>;
+  using expansion_callback_t = std::function<
+      void(baldr::GraphReader&, baldr::GraphId, const char*, const char*, float, float)>;
```
nilsnolde, line 128: just shuffling the args around a bit

src/tyr/actor.cc
```diff
@@ -267,7 +267,12 @@ actor_t::expansion(const std::string& request_str, const std::function<void()>*
   Api request;
   ParseApi(request_str, Options::expansion, request);
   // check the request and locate the locations in the graph
-  pimpl->loki_worker.route(request);
+  if (request.options().expansion_type() ==
+      Options::ExpansionType::Options_ExpansionType_expand_route) {
+    pimpl->loki_worker.route(request);
+  } else {
+    pimpl->loki_worker.isochrones(request);
+  }
```
nilsnolde, line 275: or you prefer to have a `expansion_action.cc` in loki?

src/proto_conversions.cc
```diff
@@ -351,6 +351,15 @@ bool RoadClass_Enum_Parse(const std::string& rc_name, valhalla::RoadClass* rc) {
   return true;
 }
 
+std::string ExpansionType_Enum_Name(Options::ExpansionType et) {
```
maksimandrianov, line 354: const std::string & ?

src/proto_conversions.cc
```diff
@@ -351,6 +351,15 @@ bool RoadClass_Enum_Parse(const std::string& rc_name, valhalla::RoadClass* rc) {
...
+  static const std::string empty;
+  static const std::unordered_map<Options::ExpansionType, std::string>
+      types{{Options::ExpansionType::Options_ExpansionType_expand_route, "route"},
+            {Options::ExpansionType::Options_ExpansionType_expand_isochrone, "isochrone"}};
+  auto i = types.find(et);
```
maksimandrianov, line 359: const?

kevinkreiser:
> I guess in the future this could use mapbox vector tiles spec when pbf output is real 

we would not use dijkstras to return tiled output so i think of this as a separate effort. we would make the tiles directly from the existing graph tiles. there is no point in doing a graph traversal to build something visual unless it is the graph traversal that you care about.

so while this expansion work is visual and is related to #1703, its very much the result of a graph algorithm. to me they are not the same. when we get around to doing proper vector tile rendering like is mentioned in #1703, i expect it more to be a "transcoding" operation of graph tiles to MVT rather than anything specific to graph algorithms. i also expect it to be way way faster because of it not being a graph algorithm :wink:

nilsnolde:
Sure, it's not really related to #1703, here we really are interested in the traversal output. MVT format would still be beneficial and if it's only for response size and some serialization seconds. This feature is so niche though (for me), it doesn't really justify digging into MVT output.. 

Let me know what you think about the raised points in the PR description, I'll be holding off with tests and rigorous checks (e.g. @maksimandrianov 's comments) until then.

kevinkreiser:
> This PR is adds a request parameter expansion_type to distinguish between route debugging and isochrone requests

maybe we should just have a secondary `action` type passed in. like i want to do the `expansion` for `route` or for `isochrone` or for `centroid` or for ... now of course some secondary `action`s wont be supported like `locate` or `status` etc

kevinkreiser:
> it doesn't really justify digging into MVT output

i completely agree, i was just pointing out the different use-case for MVT. to me i think when we do MVT we will do it so that people could get away with just calling valhalla both for routing and map rendering. then way way down the line we could do some rudimentary geocoding and become a one stop map platform... check back in 2035 to see if we made it that far :smile:

danpat:
> MVT format would still be beneficial and if it's only for response size and some serialization seconds.

@nilsnolde also note that MVTs are spatially indexed (tiles have a a fixed spatial bounds) - what exact tile would you be generating in response to a single request to this endpoint?

To support direct MVT access from a client to make a slippy map, you'd need a way for independent tile requests to clip  the needed result.  That would mean either running the whole algorithm repeatedly for each tile to snip out its spatial extent of the results, or a way to run the expansion first, then request independent tiles on that as a second step to avoid repeating the calculation for each tile.

I don't think that is worth the lift compared to simply retrieving geojson and adding it as a layer to your map.

nilsnolde:
> maybe we should just have a secondary action type passed in.

Not sure I understand, can you elaborate a little (passed in where?) or give a quick example? 

Honestly, I don't really like this whole logic there right now in `thor/expansion_action.cc` so I'm _very_ open for alternative ways, but kinda lack the creativity atm..

nilsnolde:
jep, you got me @danpat , no idea about MVT specs ;) graphhopper had a similar discussion a few years back for the exact same feature, they were thinking about using MVT, but settled on streamable CSV (which would be hardly an improvement compared to geojson MultiLineString in terms of response size/performance (?), but would actually be easier for client-side post-processing).

kevinkreiser:
>  can you elaborate a little

yeah i was just thinking someone could specify what endpoint they should be running expansion on:

```
# regular route expansion
/expansion?json={"action":"route", ...
# isochrone expansion
/expansion?json={"action":"isochrone", ...
# centroid expansion
/expansion?json={"action":"centroid", ...
```
Then in code you can re-use the action name parsing like we do for the normal url and if you get back an action that we dont support expansion for you can return an "unsupported" error.

Just a thought, i didnt think it through in depth yet but since "expansion" is like a meta thing of a regular request it seemed to me we needed a way to relate the two. I mean maybe its as simple as not creating a new "expansion_type" but rather just re-using the protobuf "action" enum?

nilsnolde:
> I mean maybe its as simple as not creating a new "expansion_type" but rather just re-using the protobuf "action" enum?

Thx, I see. And makes perfect sense, I'll adjust :)

nilsnolde:
I adjusted to using `action` to infer isochrone or route @kevinkreiser .

And I just wrote a test for this and there it hit me that there's a problem: loops where the Dijkstra traverses the same edge twice in both directions will be tracked twice, but IMO should only be tracked once for the least-cost (/-time/-distance?) direction to solve analytical questions such as #1963. I didn't test enough (i.e. looked at real-world output) to get a feeling how often that occurs, but I assume _very_ often in areas with mostly bidirectional access?

I'm unsure what's the best solution here.. First idea: maintain a map of visited edges in the tracker callback to look up if the opposing edge was already visited before? Question is what happens then:
  1. don't track the edge since it's apparently a higher cost than its opposing (not entirely sure how/if _sort_ cost is different from `Cost.cost`, so I might've mixed things up in that statement)
  2. include the directed edge with the least _duration_ of both, discard the other
  3. include the directed edge with the least _distance_ of both, discard the other
  4. make it request parameter to do any of 1. - 3.?

On top of that I also have little feeling how much performance penalty those checks would bring in, though it's fairly quickly implemented and tested.

Before I do that, maybe someone has a better idea? Or this is a non-issue? Don't have a lawn to mow, but I'll contemplate it again during a run;)

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,152 @@
...
+                std::to_string(edgeid.Tile_Base()));
+      return;
+    }
+    const auto* edge = tile->directededge(edgeid);
+    auto shape = tile->edgeinfo(edge).shape();
+
+    // TODO: remove debug stuff
+    std::cout << std::setprecision(4) << std::setw(6) << tile->edgeinfo(edge).GetNames()[0]
+              << std::setw(10) << (edge->forward() ? ": forward" : ": reverse") << std::setw(8)
+              << duration << std::setw(8) << distance << std::setw(8) << cost << std::endl;
```
nilsnolde, line 70: will print a table like (here part of the included test output):

```
  Name       dir     dur    dist    cost
    AD : forward    0.96      20   0.816
    AB : forward    0.96      20   0.816
    AB : reverse   29.42      40   1.632  -- higher cost than next one?
    AD : reverse   29.42      40   1.632  -- higher cost than next one?
    AE : forward    1.44      30   1.224
    AC : forward    1.44      30   1.224
    ...
```

Bit surprised that smth with higher cost and u-turn (`AB`&`AD` reverse) is popped off the queue before smth with lower cost (`AE`&`AC` forward).. so sort cost is smth else than cost.cost?

src/worker.cc
```diff
@@ -959,6 +961,18 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+  auto exp_action_str = rapidjson::get_optional<std::string>(doc, "/action");
+  Options::Action exp_action = Options::isochrone;
+  if (exp_action_str) {
+    static const std::vector<Options::Action> valid_types{Options::route, Options::isochrone};
+    if (!Options_Action_Enum_Parse(*exp_action_str, &exp_action) ||
+        !std::count(valid_types.begin(), valid_types.end(), exp_action)) {
+      throw valhalla_exception_t(144, *exp_action_str);
+    }
+  }
+  options.set_expansion_action(exp_action);
```
nilsnolde, line 974: Sets default to `isochrone` if `action` was omitted from `/expansion` request. Rather throw in that case?

test/gurka/gurka.h
```diff
@@ -161,15 +161,15 @@ valhalla::Api do_action(const valhalla::Options::Action& action,
                         const map& map,
                         const std::string& request_json,
                         std::shared_ptr<valhalla::baldr::GraphReader> reader = {},
-                        std::string* json = nullptr);
+                        std::string& json = *std::unique_ptr<std::string>(new std::string()));
```
nilsnolde, line 164: couldn't make it work with a string pointer, segfaulted for me when passing a string pointer..

test/gurka/test_expansion.cc
```diff
@@ -0,0 +1,59 @@
...
+    const gurka::ways ways = {
+        {"AB", {{"highway", "primary"}, {"name", "AB"}}},
+        {"AD", {{"highway", "primary"}, {"name", "AD"}}},
+        {"AE", {{"highway", "primary"}, {"name", "AE"}}},
+        {"AC", {{"highway", "primary"}, {"name", "AC"}}},
+        {"CF", {{"highway", "primary"}, {"name", "CF"}}},
+        {"FG", {{"highway", "primary"}, {"name", "FG"}}},
+        {"GH", {{"highway", "primary"}, {"name", "GH"}}},
+        {"HC", {{"highway", "primary"}, {"name", "HC"}}},
+    };
```
nilsnolde, line 31: that's where it hit me. this is the full expansion table:

```c++
  Name       dir     dur    dist    cost
    AD : forward    0.96      20   0.816
    AB : forward    0.96      20   0.816
    AB : reverse   29.42      40   1.632  -- higher cost than next one
    AD : reverse   29.42      40   1.632  -- higher cost than next one
    AE : forward    1.44      30   1.224
    AC : forward    1.44      30   1.224
    AE : reverse   30.38      60   2.448
    HC : reverse   2.898      60   8.948
    CF : forward   4.505      50   10.04
    GH : reverse   3.858      80   14.76
    FG : forward   5.945      80   16.26
    FG : reverse   5.298     110   20.99
    GH : forward   6.905     100   22.08
    CF : reverse   6.258     130    26.8
    HC : forward   8.345     130    28.3
    AC : reverse   9.803     160   36.03
```

clear for you probably, I still found it interesting to exactly track the expansion and the cost:)

valhalla/thor/worker.h
```diff
@@ -58,12 +58,13 @@ class thor_worker_t : public service_worker_t {
   void route(Api& request);
   std::string matrix(Api& request);
   void optimized_route(Api& request);
-  std::string isochrones(Api& request);
+  std::string isochrones(Api& request, bool only_expand = false);
```
nilsnolde, line 61: to skip building geometry & serializing the response

src/tyr/actor.cc
```diff
@@ -267,7 +267,12 @@ actor_t::expansion(const std::string& request_str, const std::function<void()>*
   Api request;
   ParseApi(request_str, Options::expansion, request);
   // check the request and locate the locations in the graph
-  pimpl->loki_worker.route(request);
+  if (request.options().has_expansion_action() &&
```
nilsnolde, line 270: the `has_expansion_action()` will be removed since there's a default (for now)

src/thor/isochrone_action.cc
```diff
@@ -8,7 +8,7 @@ using namespace valhalla::midgard;
 namespace valhalla {
 namespace thor {
 
-std::string thor_worker_t::isochrones(Api& request) {
+std::string thor_worker_t::isochrones(Api& request, bool only_expand) {
```
nilsnolde, line 11: I thought it's nicer to have it as argument rather than implicitly skip when the `action` is `expansion`

nilsnolde:
Sorry, making it a bit hard to follow here right now I think.. 

So I thought again about this issue of returning “duplicates”: it’s a bit painful but of course this could be solved client-side as well instead of Valhalla. It’s a whole lot more expensive though to weed out (costlier) opposites from the full expansion response compared to letting the tracker function do it. On the other hand the full expansion tree is really cool too as it gives you a full directed network and you can do with it what you want, the linked issue request being just one possible application.

test/gurka/gurka.h
nilsnolde: made it work..

kevinkreiser:
@nilsnolde regarding the traversal in the opposing direction i think you're right that the best choice would be while the algorithm is marking stuff keep track if we've seen an opposing edge already and dont mark again. i havent really thought about what that will do to the code though. i guess it just means the expansion logic will need to keep some state, if that is the case that should be fine so long as we can make it optional to do this extra work of tracking opposing edges. probably we need to add a new flag for the expansion endpoint that enables this feature. good luck on naming that parameter though :smile:

nilsnolde:
Ok cool, happy your're fine with that @kevinkreiser ! Then while we're at a new parameter, we could do as I proposed [here](https://github.com/valhalla/valhalla/pull/3206#issuecomment-879231870):


> 1. don't track the edge since it's apparently a higher cost than its opposing (not entirely sure how/if sort cost is different from Cost.cost, so I might've mixed things up in that statement)
> 2. include the directed edge with the least duration of both, discard the other
> 3. include the directed edge with the least distance of both, discard the other
> 4. make it request parameter to do any of 1. - 3.?

And 5.: don't track opposites.

HahaHow's `expand_strategy` or `collect_strategy`? Then e.g. `none` (default, i.e. skip opposite tracking), `cost`, `time`, `distance` as possible values?

Or just a flag and have `cost` determine it :smile: would be easier anyways..

nilsnolde:
I'll do a flag for now, the other proposal is too nit-picky for my taste.. can always change that in the future (wouldn't consider this endpoint tooop critical for backwards-incompatible changes..)

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,144 @@
...
+  // default generalization to ~ zoom level 15
+  float gen_factor = options.has_generalize() ? options.generalize() : 21.4f;
+
+  // default the expansion geojson so its easy to add to as we go
+  Document dom;
+  dom.SetObject();
+  FillDom(dom, exp_action);
+
+  // a lambda that the path algorithm can call to add stuff to the dom
+  // route and isochrone produce different GeoJSON properties
```
kevinkreiser, line 49: does it have to lead to different properties though? we track the same values in edge labels regardless of the algorithm type. i didnt go back and look at bidirea* and all the places it fires the expansion callback but if we have the label then we could also pass the details that isochrones passes. i have to say im not a huge fan of the outputs differing between different algorithms. im totally not opposed to changing the format now though, as you mentioned elsewhere this api is not even in the docs yet so its ok to make changes to it (we can consider it beta haha).

maybe you could fill in the section in the docs that you made with the proposed output so it would be easier to see where different actions would differ? i had hoped that at the worst one algorithm might have less or more information than the other but the properties in the geojson wouldnt be wholly different. maybe that is the case, i didnt try it out yet.

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,144 @@
...
+      GetValueByPointer(dom, "/features/0/properties/durations")
+          ->GetArray()
+          .PushBack(Value{}.SetInt(static_cast<u_int64_t>(duration)), a);
+      GetValueByPointer(dom, "/features/0/properties/distances")
+          ->GetArray()
+          .PushBack(Value{}.SetInt(distance), a);
+      GetValueByPointer(dom, "/features/0/properties/costs")
+          ->GetArray()
+          .PushBack(Value{}.SetInt(cost), a);
+    }
```
kevinkreiser, line 100: i see. yeah to me i feel like we could merge these together. statuses still exist in dijkstras but we could omit them for isochrones, and having the edge label info for all the other algorithms would be very helpful imho

src/thor/isochrone_action.cc
kevinkreiser: i actually prefer the other way around. here is my reasoning: i really want the top level apis to be as simple as possible and i want all of the info to process a request to be in the request object and not elsewhere. basically my main concern is about library api design.

src/thor/expansion_action.cc
nilsnolde: yeah, I'm just refactoring the whole thing (and some logic change), so this will look/feel better and be more consistent. I'll remove the draft when I'm ready for review:)

src/thor/isochrone_action.cc
```diff
@@ -37,6 +37,10 @@ std::string thor_worker_t::isochrones(Api& request) {
                                                                         : ExpansionType::forward;
   auto grid = isochrone_gen.Expand(expansion_type, request, *reader, mode_costing, mode);
 
+  // e.g. in case of /expansion request
+  if (only_expand)
+    return "";
```
kevinkreiser, line 42: i see why you wanted to pass in the flag, because you can do so from the expansion action in the worker, the root cause of having to do this is that we dont store the intermediate information of the result of the isochrone in the protobuf. if we did we would return from here having modified the protobuf and then just send the protobuf to a serialization function separately. if that were the case we wouldnt have to return a blank string here we would just return after the expansion and then the serializer would call generate contours and do the serialization of those.

its not worth the refactoring in my opinion to do this now but i think to fit the pattern we have elsewhere and avoid any trickery it might be worth doing the refactoring at some point so every action works the same, basically: look in the protobuf to see what to do, add the work you did to the protobuf and return.

src/thor/isochrone_action.cc
nilsnolde: yeah, based on previous reviews I was literally 50:50 which way you'd prefer;) will change

src/worker.cc
kevinkreiser: yeah i think it should be explicit especially if the output will be potentially different depending on the action. the client should know what they are asking for at some point

test/gurka/gurka.cc
```diff
@@ -52,21 +52,11 @@ std::vector<midgard::PointLL> to_lls(const nodelayout& nodes,
...
- * @param costing        which costing name to use, defaults to auto
- * @param options        overrides parts of the request, supports rapidjson pointer semantics
- * @param stop_type      break, through, via, break_through
- * @return json string
- */
 std::string build_valhalla_request(const std::string& location_type,
                                    const std::vector<midgard::PointLL>& waypoints,
-                                   const std::string& costing = "auto",
-                                   const std::unordered_map<std::string, std::string>& options = {},
-                                   const std::string& stop_type = "break") {
```
kevinkreiser, line 69: why remove the doc string and why remove the default parameters? this one isnt obvious to me :smile:

kevinkreiser: oh i didnt scroll down, you moved it to the header

src/thor/isochrone_action.cc
nilsnolde: > we dont store the intermediate information of the result of the isochrone in the protobuf

yep would be much nicer..

this was also where I wasn't sure which one you prefer:

1. pass in the flag to the isochrone_action (advantage: other (future) callers of the action could ask for `only_expand`)
2. inspect the `options.action()` right here and return if it's `expansion` ("only" advantage: cleaner/more consistent in thor actions ctors)

I prefer 1. but can understand if you want 2. and eventually refactor to not need this at all as you mentioned.

test/gurka/gurka.cc
nilsnolde: look into the header ;) needed to access it (or wanted to to save some lines..)

proto/options.proto
```diff
@@ -221,4 +221,6 @@ message Options {
   optional bool linear_references = 45;                                   // Include linear references for graph edges returned in certain responses.
   repeated CostingOptions recostings = 46;                                // Costing options to use to recost a path after it has been found
   repeated Ring exclude_polygons = 47;                                    // Rings/polygons to exclude entire areas during path finding
+  optional Action expansion_action = 48;                                  // Meta action for /expansion endpoint
+  optional bool skip_opposites = 49;                                  // Whether to return opposite edges encountered during expansion
```
nilsnolde, line 225: settled on `skip_opposites` as api param. whatya think?

src/thor/bidirectional_astar.cc
```diff
@@ -318,8 +318,8 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
 
   // setting this edge as reached
   if (expansion_callback_) {
-    expansion_callback_(graphreader, "bidirectional_astar", FORWARD ? meta.edge_id : opp_edge_id, "r",
-                        false);
+    expansion_callback_(graphreader, FORWARD ? meta.edge_id : opp_edge_id, "bidirectional_astar", "r",
+                        0, 0, 0);
```
nilsnolde, line 322: or you want duration, distance, cost for bdira* as well?

src/thor/expansion_action.cc
nilsnolde: > i have to say im not a huge fan of the outputs differing between different algorithms.

me neither really.. but those arrays are not cheap in terms of output size (calculated it yesterday during jogging, think was roughly smth like 20% of the `geometry` size), so I'd prefer even more to not have `edge_ids` (only good for debugging isnt it?) & `statuses` (pointless, we're only tracking settled edges so far) in the isochrone output.

I let you decide if you want those additional attributes (dist, dur, cost) for the routing output. 

That being said, I think the `isochrone` action has more meaning than `route`, so it's more user-facing. Consequently I think it's fine if that's returning smth entirely different. 

We could also at least have the same keys in both responses, but then some empty arrays? On the other hand it's easier to add attributes down the road rather than removing any.

What do you think?

nilsnolde: right now it's:
- isochrones: `durations`, `distances`, `costs`
- route: `statuses`, `edge_ids`

I can include the isochrone ones to `route` as well?

docs/api/expansion/api-reference.md
```diff
@@ -0,0 +1,38 @@
...
+
+| Property   | Description                           |
+| :--------- | :------------------------------------ |
+| `distances`   | JSON array of the accumulated distance in meters for each edge in order of graph traversal. | 
+| `durations`   | JSON array of the accumulated duration in seconds for each edge in order of graph traversal. | 
+| `costs`   | JSON array of the accumulated edge cost for each edge in order of graph traversal. | 
+| `edge_ids`   | JSON array of the internal edge IDs for each edge in order of graph traversal. | 
+| `statuses`   | JSON array of the edge states at the  for each edge in order of graph traversal. |
```
nilsnolde, line 28: not settled yet entirely. right now the output keys are:
- isochrones: `durations`, `distances`, `costs`
- route: `statuses`, `edge_ids`

I can include the isochrone ones to `route` as well?

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,183 @@
...
+  dom.SetObject();
+  // set algorithm to Dijkstra, will be overwritten by other algos
+  SetValueByPointer(dom, "/properties/algorithm", "unidirectional_dijkstra");
+  SetValueByPointer(dom, "/type", "FeatureCollection");
+  SetValueByPointer(dom, "/features/0/type", "Feature");
+  SetValueByPointer(dom, "/features/0/geometry/type", "MultiLineString");
+  SetValueByPointer(dom, "/features/0/geometry/coordinates", Value(kArrayType));
+  for (int prop = PropType::kEdgeId; prop != LAST; prop++) {
+    rapidjson::Pointer(kPropPaths[static_cast<PropType>(prop)]).Set(dom, Value(kArrayType));
+  }
```
nilsnolde, line 60: I found this _fairly_ elegant, but not sure if you like that usage of enums..

src/thor/bidirectional_astar.cc
kevinkreiser: why not, its incredibly useful debug info right?

src/thor/expansion_action.cc
kevinkreiser: yeah i feel like why not all of them in both? maybe the api lets you specify which bits of information you care about getting back?

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,183 @@
...
+// these need to represent array types
+enum PropType {
+  kEdgeId,
+  kStatus,
+  kDist,
+  kTime,
+  kCost,
+  // insert new keys before this
+  LAST
+};
```
kevinkreiser, line 23: we could make this part of the protobuf request and let people select a repeated list of these to turn on the ones they do and dont want, and protobuf handles that `LAST` value for you too

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,183 @@
...
+  measure_scope_time(request, "thor_worker_t::expansion");
+
+  // get the request params
+  auto options = request.options();
+  auto exp_action = options.expansion_action();
+  bool skip_opps = options.skip_opposites();
+  std::unordered_set<baldr::GraphId> opp_edges;
+
+  // default generalization to ~ zoom level 15
+  float gen_factor = options.has_generalize() ? options.generalize() : 21.4f;
```
nilsnolde, line 47: you'd need to visualize this in qgis or other desktop software, at least my browser hopelessly crashes even for the tiniest isochrone in geojson.io :sweat_smile: which reminds me that we're always adding 10 mins/kms to each dijkstra computation, so there's actually no way to get less than 10 mins/kms expansion (fine for me, but good to keep in mind)

A few stats/pics for different epsilon from Vienna 1 km (aka 11 km) isodistance:
- 10 m epsilon: 112184 total **vertices**, 3.1 MB response
- 21.4 m epsilon: 114660 total **vertices**, 3.1 MB response

the 10 m default has relatively very few more vertices, no difference in response size and simply looks better (the yellow circled road is fine IMO). the pic (red: 20 m, green: 10 m):
image

test/gurka/test_expansion.cc
```diff
@@ -0,0 +1,132 @@
...
+  }
+  const auto req =
+      gurka::detail::build_valhalla_request(std::string("locations"), lls, "auto", options);
+  std::string res;
+  auto api = gurka::do_action(Options::expansion, expansion_map, req, nullptr, &res);
+
+  rapidjson::Document res_doc;
+  res_doc.Parse(res.c_str());
+
+  check_props(res_doc, 23, {"statuses", "edge_ids"});
```
nilsnolde, line 117: it's expected that bidir a* branches don't circle back on the opposites right (i.e. both `skip_opposites` true & false give 23 directed edges)? not entirely sure about that..

test/thor_worker.cc
```diff
@@ -98,8 +98,7 @@ TEST(ThorWorker, test_linear_references) {
   std::vector<std::string> requests = {
       R"({"costing":"auto","linear_references":true,"locations":[
           {"lat":52.09110,"lon":5.09806},
-          {"lat":52.09098,"lon":5.09679}],
-          "action":"include"})",
+          {"lat":52.09098,"lon":5.09679}]})",
```
nilsnolde, line 101: that seemed redundant and actually crashed now that we do have an `action` parameter

src/thor/expansion_action.cc
nilsnolde: I'll get back to this in a bit

src/thor/expansion_action.cc
nilsnolde: same, will get back in the general section

nilsnolde: yeah damn it, I'll do that now, sounds actually awesome! I'm really cautious about adding params to the requests, but definitely down for it if you are!

docs/api/expansion/api-reference.md
nilsnolde: introduced new request parameter to choose response geojson property fields

docs/api/expansion/api-reference.md
```diff
@@ -0,0 +1,38 @@
...
+**Note**, for even moderately long routes or isochrones the `/expansion` action can produce gigantic GeoJSON responses of 100s of MB.
+
+## Inputs of the Expansion service
+
+Since this service wraps other services, the request format mostly follows the ones of the [route](../turn-by-turn/api-reference.md#inputs-of-a-route) and [isochrone](../isochrone/api-reference.md#inputs-of-the-isochrone-service). Additionally, it accepts the following parameters:
+
+| Parameter          | Description                           |
+| :---------         | :------------------------------------ |
+| `action`           | The service whose expansion should be tracked. One of `route` or `isochrone`.  | 
+| `skip_opposites`   | If set to `true` the output won't contain an edge's opposing edge. Opposing edges can be thought of as both directions of one road segment. Of the two, we discard the directional edge with higher cost and keep the one with less cost. |
```
nilsnolde, line 16: still misses new param, will do tmrw

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,176 @@
...
+
+  // get the request params
+  auto options = request.options();
+  auto exp_action = options.expansion_action();
+  bool skip_opps = options.skip_opposites();
+  auto exp_props = options.expansion_props();
+  std::unordered_set<baldr::GraphId> opp_edges;
+
+  // default generalization to ~ zoom level 15
+  float gen_factor = options.has_generalize() ? options.generalize() : 10.f;
```
nilsnolde, line 38: this generalization works fine.. had an earlier comment with some some stats and a pic for 10 and 20 m epsilon, 10m wasn't that much different in size but much better geometries. I guess the difference would be bigger in autobahn areas

src/thor/expansion_action.cc
nilsnolde: implemented this

src/thor/isochrone_action.cc
nilsnolde: I went with 2.

src/worker.cc
```diff
@@ -959,6 +965,55 @@ void from_json(rapidjson::Document& doc, Options& options) {
     options.set_resample_distance(*resample_distance);
   }
 
+  // expansion action, default to isochrone
```
nilsnolde, line 968: invalid comment

src/worker.cc
```diff
@@ -959,6 +965,55 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+      }
+      exp_props_pbf->Add(exp_prop);
+    }
+  }
+  // default to at least return durations & distances
+  if (!exp_props_pbf->size()) {
+    for (const auto& prop : {Options_ExpansionProps_durations, Options_ExpansionProps_distances}) {
+      exp_props_pbf->Add(prop);
+    }
+  }
```
nilsnolde, line 1000: default yes/no? I feel we should at least give back `distances` and `durations` as they're most likely to make sense to the user.

test/gurka/test_expansion.cc
nilsnolde: must've been smth off, now I get 16 edges instead of 23, sounds more reasonable

nilsnolde:
@kevinkreiser I'm happy now, let me know if you're too when you find the time.

I implemented a request parameter to specify the `property` types in the request as you suggested, much nicer.

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@
    * ADDED: Allow going through accessible `barrier=bollard` and penalize routing through it, when the access is private [#3175](https://github.com/valhalla/valhalla/pull/3175)
    * ADDED: Add country code to incident metadata [#3169](https://github.com/valhalla/valhalla/pull/3169)
    * CHANGED: Use distance instead of time to check limited sharing criteria [#3183](https://github.com/valhalla/valhalla/pull/3183)
+   * ADDED: Added vehicle width and height as an option for auto (and derived: taxi, bus, hov) profile (https://github.com/valhalla/valhalla/pull/3179) 
+   * ADDED: isochrone action for /expansion endpoint to track dijkstra expansion [#3215](https://github.com/valhalla/valhalla/pull/3179)
```
nilsnolde, line 43: will move before merge

src/loki/worker.cc
```diff
@@ -320,6 +319,14 @@ loki_worker_t::work(const std::list<zmq::message_t>& job,
...
         result.messages.emplace_back(request.SerializeAsString());
         break;
+      case Options::expansion:
+        if (options.expansion_action() == Options::route) {
+          route(request);
+        } else {
+          isochrones(request);
+        }
+        result.messages.emplace_back(request.SerializeAsString());
+        break;
```
nilsnolde, line 329: we catch in the worker if an unsupported action was requested.

src/proto_conversions.cc
```diff
@@ -122,6 +122,17 @@ bool Options_Action_Enum_Parse(const std::string& action, Options::Action* a) {
...
+bool Options_ExpansionAction_Enum_Parse(const std::string& action, Options::Action* a) {
+  static const std::unordered_map<std::string, Options::Action> actions{{"route", Options::route},
+                                                                        {"isochrone",
+                                                                         Options::isochrone}};
+  auto i = actions.find(action);
+  if (i == actions.cend())
+    return false;
+  *a = i->second;
+  return true;
+}
```
nilsnolde, line 134: could also add a param to `Options_Action_Enum_Parse` above, but this feels more natural

src/thor/bidirectional_astar.cc
```diff
@@ -880,7 +884,10 @@ void BidirectionalAStar::SetOrigin(GraphReader& graphreader,
 
     // setting this edge as reached
     if (expansion_callback_) {
-      expansion_callback_(graphreader, "bidirectional_astar", edgeid, "r", false);
+      EdgeStatusInfo fwd_edge_status = edgestatus_forward_.Get(edgeid);
+      auto fwd_pred = edgelabels_forward_[fwd_edge_status.index()];
+      expansion_callback_(graphreader, edgeid, "bidirectional_astar", "r", fwd_pred.cost().secs,
+                          fwd_pred.path_distance(), fwd_pred.cost().cost);
```
nilsnolde, line 890: check some of those I added to bidira*, not 100% if they're all correct

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,188 @@
...
+  SetValueByPointer(dom, "/properties/algorithm", "unidirectional_dijkstra");
+  SetValueByPointer(dom, "/type", "FeatureCollection");
+  SetValueByPointer(dom, "/features/0/type", "Feature");
+  SetValueByPointer(dom, "/features/0/geometry/type", "MultiLineString");
+  SetValueByPointer(dom, "/features/0/geometry/coordinates", Value(kArrayType));
+  SetValueByPointer(dom, "/features/0/properties", Value(kArrayType));
+  for (const auto& prop : exp_props_pbf) {
+    rapidjson::Pointer(kPropPaths[static_cast<Options_ExpansionProps>(prop)])
+        .Set(dom, Value(kArrayType));
+    exp_props_set.insert(static_cast<Options_ExpansionProps>(prop));
```
nilsnolde, line 55: need a container which implements some find/count or smth, PBF repeated fields seem to behave more like a vector

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    // no properties asked for, don't collect any
+    if (!exp_props_set.size()) {
+      return;
+    }
+
+    // make the properties
+    if (algorithm)
+      SetValueByPointer(dom, "/properties/algorithm", algorithm);
```
nilsnolde, line 108: we default the algorithm attribute to dijkstra, so isochrones don't have to set it, but others will

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,188 @@
...
+    isochrone_gen.set_track_expansion(nullptr);
+  }
+
+  // remove the arrays which weren't used
+  for (const auto& prop : exp_props_set) {
+    const auto& member = kPropPaths[prop];
+    if (!Pointer(member).Get(dom)->GetArray().Size()) {
+      Pointer(member).Erase(dom);
+    }
+  }
```
nilsnolde, line 181: forgot to remove this!

src/worker.cc
```diff
@@ -802,6 +805,36 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+  auto exp_action_str = rapidjson::get_optional<std::string>(doc, "/action");
+  Options::Action exp_action = Options::isochrone;
+  if (exp_action_str) {
+    if (!Options_ExpansionAction_Enum_Parse(*exp_action_str, &exp_action)) {
+      throw valhalla_exception_t(144, *exp_action_str);
+    }
+    options.set_expansion_action(exp_action);
+  } else if (options.action() == Options_Action_expansion) {
+    throw valhalla_exception_t(115, std::string("action"));
+  }
```
nilsnolde, line 818: thow if:
- no "action" for expansion request
- unsupported action key

src/worker.cc
```diff
@@ -802,6 +805,36 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+  auto* exp_props_pbf = options.mutable_expansion_props();
+  Options::ExpansionProps exp_prop;
+  if (exp_props_req && exp_props_req->IsArray()) {
+    for (const auto& prop : exp_props_req->GetArray()) {
+      if (!valhalla::Options_ExpansionProps_Enum_Parse(std::string(prop.GetString()), &exp_prop)) {
+        throw valhalla_exception_t(168, std::string(prop.GetString()));
+      }
+      exp_props_pbf->Add(exp_prop);
+    }
+  }
```
nilsnolde, line 831: only throw on unsupported property keys; allow for no keys at all (ie empty `properties` array in the response, which doesn't seem to be explicitly forbidden according to spec, geojson.io is happy with `properties` array)

test/gurka/gurka.cc
```diff
@@ -86,6 +76,24 @@ std::string build_valhalla_request(const std::string& location_type,
...
+      std::getline(ss, substr, ',');
+      props.push_back(substr);
+    }
+    for (const auto& prop : props) {
+      rapidjson::Value s(prop.c_str(), allocator);
+      props_arr.PushBack(s, allocator);
+    }
+    doc.AddMember("expansion_props", props_arr, allocator);
+  }
+
```
nilsnolde, line 96: forgot to remove this

nilsnolde:
I confused another type which win doesn’t like (good thing).. will amend after final review

nilsnolde:
Ok, think I squashed the windows bug..

src/worker.cc
```diff
@@ -58,6 +58,7 @@ const std::unordered_map<unsigned, valhalla::valhalla_exception_t> error_codes{
     {112, {112, "Insufficiently specified required parameter 'locations' or 'sources & targets'", 400, HTTP_400, OSRM_INVALID_OPTIONS, "matrix_locations_parse_failed"}},
     {113, {113, "Insufficiently specified required parameter 'contours'", 400, HTTP_400, OSRM_INVALID_OPTIONS, "contours_parse_failed"}},
     {114, {114, "Insufficiently specified required parameter 'shape' or 'encoded_polyline'", 400, HTTP_400, OSRM_INVALID_OPTIONS, "shape_parse_failed"}},
+    {115, {115, "Insufficiently specified required parameter 'action'", 400, HTTP_400, OSRM_INVALID_OPTIONS, "action_parse_failed"}},
```
kevinkreiser, line 61: what did you think of this refactor? i hoped by putting all this crap together it would be much easier to do all the things you needed for a new service level error

src/worker.cc
```diff
@@ -802,6 +805,36 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+    for (const auto& prop : exp_props_req->GetArray()) {
+      if (!valhalla::Options_ExpansionProps_Enum_Parse(std::string(prop.GetString()), &exp_prop)) {
+        throw valhalla_exception_t(168, std::string(prop.GetString()));
+      }
+      exp_props_pbf->Add(exp_prop);
+    }
+  }
+
+  // should the expansion track opposites?
+  auto exp_skip_opps = rapidjson::get_optional<bool>(doc, "/skip_opposites");
```
kevinkreiser, line 834: i kind of forgot about this. can you explain what this does again? it somehow doesnt track opposing edges?

test/thor_worker.cc
kevinkreiser: haha yeah im not sure how that got there. it was probably a copy past fail from a request with a `filter` in it which have actions of include and excluding filters

kevinkreiser: should crash though, it should throw and then be an unhandled exception which crashes?

valhalla/thor/worker.h
```diff
@@ -64,6 +64,7 @@ class thor_worker_t : public service_worker_t {
   std::string expansion(Api& request);
   void centroid(Api& request);
   void status(Api& request) const;
+  void expansion(Api& request) const;
```
kevinkreiser, line 67: this is wierd, do we actually need two expansion fuctions or did you just forget to remove the old signature?

kevinkreiser: oh no it actually looks like this one is superfluous?

src/worker.cc
nilsnolde: if you're asking how I like the "new" way of defining valhalla_exceptions: love it:) I guess your comment is unrelated to the actual change?

src/worker.cc
nilsnolde: right: the problem here was that dijkstra will traverse and settle _lots_ of opposing edges which have (of course) higher cost than their previously settled counterparts. in the context of data analysis that's pretty undesirable, hence this option. you could weed out the opposites client-side, but only at super high costs.

test/thor_worker.cc
nilsnolde: not sure if this is a comment where you'd like to see smth changed or just a general notion :sweat_smile: but yeah, the exception will have code 115 and would be unhandled in this case

valhalla/thor/worker.h
nilsnolde: huh, wonder how that got in there.. yeah, very superfluous!

src/worker.cc
```diff
@@ -701,8 +704,7 @@ void from_json(rapidjson::Document& doc, Options& options) {
   }
 
   // Option to use timestamps when computing elapsed time for matched routes
-  options.set_use_timestamps(
-      rapidjson::get_optional<bool>(doc, "/use_timestamps").get_value_or(false));
+  options.set_use_timestamps(rapidjson::get<bool>(doc, "/use_timestamps", false));
```
nilsnolde, line 707: you'll see a bunch of those: I made all the boolean options one-liners for purely cosmetic reasons. I also did check that none had a `has_xx` call somewhere which could've messed things up.

src/thor/bidirectional_astar.cc
```diff
@@ -1003,7 +1007,10 @@ void BidirectionalAStar::SetOrigin(GraphReader& graphreader,
 
     // setting this edge as reached
     if (expansion_callback_) {
-      expansion_callback_(graphreader, "bidirectional_astar", edgeid, "r", false);
+      EdgeStatusInfo fwd_edge_status = edgestatus_forward_.Get(edgeid);
+      auto fwd_pred = edgelabels_forward_[fwd_edge_status.index()];
```
nilsnolde, line 1011: get rid of this, everything's up there..

test/gurka/test_expansion.cc
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  // common method can't deal with arrays of strings
+  std::string build_local_req(rapidjson::Document& doc,
+                              rapidjson::MemoryPoolAllocator<>& allocator,
+                              const std::vector<std::string>& waypoints,
+                              const std::string& costing,
+                              const std::string& action,
+                              bool skip_opps,
+                              const std::vector<std::string> props) {
```
nilsnolde, line 41: see how to remove: https://github.com/gis-ops/valhalla/blob/b37eb477b5d024868d31dfb6612fbe47bb88fad2/test/gurka/test_route.cc#L403

test/loki_service.cc
```diff
@@ -514,6 +514,10 @@ TEST(LokiService, test_actions_whitelist) {
   for (auto action = Options::Action_MIN; action < Options::Action_ARRAYSIZE;
        action = static_cast<Options::Action>(static_cast<int>(action) + 1)) {
 
+    // the json parser would throw a "insufficiently specidfied 'action'" (115) error for /expansion
+    if (action == Options_Action_expansion)
+      continue;
```
nilsnolde, line 519: add test to check this

CHANGELOG.md
```diff
@@ -40,6 +41,7 @@
 
 * **Enhancement**
    * CHANGED: Favor turn channels more [#3222](https://github.com/valhalla/valhalla/pull/3222)
+   * CHANGED: Rename `valhalla::midgard::logging::LogLevel` enumerators to avoid clash with common macros [#3236](https://github.com/valhalla/valhalla/pull/3236)
```
nilsnolde, line 44: random changelog entry duplication happens far too often to me..

test/gurka/gurka.cc
```diff
@@ -52,16 +52,6 @@ std::vector<midgard::PointLL> to_lls(const nodelayout& nodes,
...
-/**
- * build a valhalla json request body
- *
- * @param location_type  locations or shape
- * @param waypoints      sequence of pointlls representing the locations
- * @param costing        which costing name to use, defaults to auto
- * @param options        overrides parts of the request, supports rapidjson pointer semantics
- * @param stop_type      break, through, via, break_through
- * @return json string
- */
```
nilsnolde, line 64: this was an accident, but we could leave it this way. the header has the same description, no point in duplicating docstrings

nilsnolde:
@kevinkreiser I brushed up the PR according to our discussion on monday, to recap:

- using the built-in `gurka::valhalla_build_request` instead of a homegrown, with rapidjson's slash notation
- increased coverage in `loki_service` test
- some other minor stuff I forgot about

only changelog needs some cleaning, then it'd be good to ship for me.

test/gurka/gurka.cc
nilsnolde: .. forget it.. the whole "problem" is that I switched the definition to the header.. i'll revert while fixing the test bug failing CI

test/gurka/test_expansion.cc
```diff
@@ -0,0 +1,149 @@
...
+    } else {
+      req += R"("expansion_props":[)";
+    }
+    for (uint8_t i = 0; i < props.size(); i++) {
+      req += (boost::format(R"("%s")") % props[i]).str();
+      if (i != props.size() - 1) {
+        req += ",";
+      }
+    }
+    req += "]}";
```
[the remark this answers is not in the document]
kevinkreiser: sorry is the main problem you need to set booleans? that is completely fine to do it just pass `"0"` or `"1"` as the value part of the regular gurka request building stuff. valhalla will happily convert numbers to booleans. if this isnt what you're refering to then just ignore me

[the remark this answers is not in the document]
nilsnolde: :facepalm: yep, didn't think about that, thanks @kevinkreiser !

src/thor/dijkstras.cc
```diff
@@ -366,10 +366,16 @@ void Dijkstras::Compute(google::protobuf::RepeatedPtrField<valhalla::Location>&
...
     if (cb_decision != ExpansionRecommendation::prune_expansion) {
-      // Expand from the end node in forward direction.
+      // Expand from the end node in expansion_direction.
       ExpandInner<expansion_direction>(graphreader, pred.endnode(), pred, predindex, opp_pred_edge,
                                        false, time_infos.front());
     }
+
+    if (expansion_callback_) {
+      // "algorithm" arg defaults already to dijkstra
+      expansion_callback_(graphreader, pred.edgeid(), nullptr, "s", pred.cost().secs,
```
kevinkreiser, line 376: i see the comment but still, i couldnt see where it defaults in the code. i think it would also be odd to use a default, do you mind  if we just force the caller to specify the algorithm?

src/thor/expansion_action.cc
```diff
@@ -0,0 +1,168 @@
...
+    coords->GetArray().PushBack(Value(kArrayType), a);
+    auto& linestring = (*coords)[coords->Size() - 1];
+    for (const auto& p : shape) {
+      linestring.GetArray().PushBack(Value(kArrayType), a);
+      auto point = linestring[linestring.Size() - 1].GetArray();
+      point.PushBack(p.first, a);
+      point.PushBack(p.second, a);
+    }
+
+    // no properties asked for, don't collect any
```
kevinkreiser, line 94: so previously you couldn't select the properties, you just always got the hardcoded ones. i think we should retain that behavior if not set? otherwise if none were asked for why are we even running the request? maybe im missing something but if none are asked for we should have returned a 400 in loki at the beginning OR we should have defaulted to the ones that already existed before you improved the API

nilsnolde: you could just want the geometries for some maps? one property adds quite a lot of stuff to the response, I thought this was the most natural way to "support" that

src/thor/dijkstras.cc
nilsnolde: it's odd I agree. I'll change that

src/thor/expansion_action.cc
kevinkreiser: ok cool, so that is a good case for no additional properties other than geom. thanks for pointing it out. i guess i'll note that this is an API change along with getting full geom rather than just nodes in the graph. thats probably ok since the API was previously undocumented.

docs/api/expansion/api-reference.md
```diff
@@ -0,0 +1,51 @@
+# Expansion service API reference
```
kevinkreiser, line 1: we should slap a big fat "BETA" on this and note that its subject to change. the reason i say this is because the first iteration is reasonably different than this new iteration (at least the API is breaking). so i think until we feel comfortable we might mention that its beta and that the API could break until we remove beta from the label. what do you think @nilsnolde ?

src/worker.cc
```diff
@@ -818,14 +820,40 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+    if (!Options_ExpansionAction_Enum_Parse(*exp_action_str, &exp_action)) {
+      throw valhalla_exception_t(144, *exp_action_str);
+    }
+    options.set_expansion_action(exp_action);
+  } else if (options.action() == Options_Action_expansion) {
+    throw valhalla_exception_t(115, std::string("action"));
+  }
+
+  // expansion response properties
+  auto exp_props_req = rapidjson::get_child_optional(doc, "/expansion_props");
```
kevinkreiser, line 836: nit: could we spell out properties and also in the protobuf definition above?

docs/api/expansion/api-reference.md
nilsnolde: yep agreed!

nilsnolde:
fixed the minor stuff and resolved changelog conflict

kevinkreiser (approved):

src/thor/dijkstras.cc
```diff
@@ -364,10 +364,16 @@ void Dijkstras::Compute(google::protobuf::RepeatedPtrField<valhalla::Location>&
...
     if (cb_decision != ExpansionRecommendation::prune_expansion) {
-      // Expand from the end node in forward direction.
+      // Expand from the end node in expansion_direction.
       ExpandInner<expansion_direction>(graphreader, pred.endnode(), pred, predindex, opp_pred_edge,
                                        false, time_infos.front());
     }
+
+    if (expansion_callback_) {
+      // "algorithm" arg defaults already to dijkstra
+      expansion_callback_(graphreader, pred.edgeid(), "unidirectional_dijkstra", "s",
```
kevinkreiser, line 374: can you change this to just `dijkstras` i think a test above checks for this as well

nilsnolde: changed name to dijkstras :+1:

kevinkreiser (approved):

CHANGELOG.md
```diff
@@ -59,7 +60,7 @@
    * FIXED: Better hash function of GraphId [#3332](https://github.com/valhalla/valhalla/pull/3332)
 
 * **Enhancement**
-   * CHANGED: Favor turn channels more [#3222](https://github.com/valhalla/valhalla/pull/3222)
+   * CHANGED: Favor turn channels more [#3222](https://github.com/valhalla/valhalla/pull/3222)
```
kevinkreiser, line 63: i accidental white space but i cant stand to make ci build again

missinglink: sup sup, GH actions has a setting now where you can tell it to skip runs if it's only something trivial like docs changes, based on filepath pattern matching.

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-paths

nilsnolde: It’s stupid circleci we use here mostly.. there’s been attempts to trick circle to skip stuff but IIRC wouldn’t work in all git scenarios. there’s been a branch (by Kevin I think) switching to actions. we do like PRs in that direction though, esp by semi-retired, experienced devs 😆(ain’t looking at no one in particular)

kevinkreiser: @missinglink :wave:  yep we did actually add a couple hacks to our CI flow to detect when the changes were trivial so i think im actually talking crap above.... but @nilsnolde is right we'd like to get over to actions if we can. i did start a branch but never took it to the end, also that was back when actions just came out and it was the wild west a little bit..

missinglink: I can have a look, can you please give me access to push dev branches to this repo?
It's much easier to test if I can push up my own dev branch and open a DRAFT PR if that's cool wit youz?

```bash
/code/valhalla/valhalla ghactions ❯ git push origin ghactions:missinglink/ghactions
ERROR: Permission to valhalla/valhalla.git denied to missinglink.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

kevinkreiser: if you really want to sure i can add you. i will warn you though, it is significantly convoluted!

missinglink: so is every C project, ooo... 🔥 haha, I'll give it a crack, it can't be as bad as compiling spatialite as a loadable extension can it?

looks like most of the work is done already in the Docker scripts {famous last words}

nilsnolde: haha lovely! thanks a ton for even attempting! obviously let us know if there's questions around the convolutions

PR #3208, opened 2021-07-13
Fix for Sign retrieval

ktatterso:
I didn't realize `graphreader.GetBeginNodeId()` might modify the given `graph_tile_ptr`. This modified approach for retrieving Sign's solves the problem without that concern.

# Issue

There were some routing issues.  These coordinates exposed the issue:  -116.768599,36.9116110;-116.72504902777372,36.220713408337346101001

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/triplegbuilder.cc
+++ b/src/thor/triplegbuilder.cc
@@ -579,7 +579,7 @@ void AddSignInfo(const AttributesController& controller,
  */
 void AddTripIntersectingEdge(const AttributesController& controller,
                              valhalla::baldr::GraphReader& graphreader,
-                             graph_tile_ptr& graphtile,
+                             const graph_tile_ptr& graphtile,
                              const DirectedEdge* directededge,
                              const DirectedEdge* prev_de,
                              uint32_t local_edge_index,
@@ -659,10 +659,15 @@ void AddTripIntersectingEdge(const AttributesController& controller,
   // Set the sign info for the intersecting edge if requested
   if (controller.attributes.at(kNodeIntersectingEdgeSignInfo)) {
     if (intersecting_de->sign()) {
-      GraphId beginnode = graphreader.GetBeginNodeId(intersecting_de, graphtile);
-      valhalla::baldr::graph_tile_ptr t2 = graphreader.GetGraphTile(beginnode);
-      size_t edge_idx = intersecting_de - t2->directededge(0);
-      std::vector<SignInfo> edge_signs = t2->GetSigns(edge_idx);
+      // To retrieve the sign information from the intersecting_de I need
+      // to access the graph-tile for the "begin node" of the intersecting_de.
+      // GraphReader::GetGraphTile(DirectedEdge *) returns the graph-tile for
+      // the "end node". Hence, I need to obtain the opposing-directed-edge
+      // and then ask for it's graph tile. (Sigh)
+      GraphId opp_intersecting_de = graphtile->GetOpposingEdgeId(intersecting_de);
+      graph_tile_ptr opp_tile = graphreader.GetGraphTile(opp_intersecting_de);
+      size_t edge_idx = intersecting_de - opp_tile->directededge(0);
+      std::vector<SignInfo> edge_signs = opp_tile->GetSigns(edge_idx);
       if (!edge_signs.empty()) {
         valhalla::TripSign* sign = intersecting_edge->mutable_sign();
         AddSignInfo(controller, edge_signs, sign);
@@ -684,7 +689,7 @@ void AddTripIntersectingEdge(const AttributesController& controller,
  * @param trip_node                pbf node in the pbf structure we are building
  */
 void AddIntersectingEdges(const AttributesController& controller,
-                          graph_tile_ptr& start_tile,
+                          const graph_tile_ptr& start_tile,
                           const NodeInfo* node,
                           const DirectedEdge* directededge,
                           const DirectedEdge* prev_de,


src/thor/triplegbuilder.cc
```diff
@@ -659,10 +659,15 @@ void AddTripIntersectingEdge(const AttributesController& controller,
...
-      GraphId beginnode = graphreader.GetBeginNodeId(intersecting_de, graphtile);
-      valhalla::baldr::graph_tile_ptr t2 = graphreader.GetGraphTile(beginnode);
-      size_t edge_idx = intersecting_de - t2->directededge(0);
-      std::vector<SignInfo> edge_signs = t2->GetSigns(edge_idx);
+      // To retrieve the sign information from the intersecting_de I need
+      // to access the graph-tile for the "begin node" of the intersecting_de.
+      // GraphReader::GetGraphTile(DirectedEdge *) returns the graph-tile for
+      // the "end node". Hence, I need to obtain the opposing-directed-edge
+      // and then ask for it's graph tile. (Sigh)
+      GraphId opp_intersecting_de = graphtile->GetOpposingEdgeId(intersecting_de);
```
kevinkreiser, line 667: I believe we have another flavor of this function that gets you all the information in one call

kevinkreiser: https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/graphreader.h#L567

kevinkreiser: actually scratch that..

src/thor/triplegbuilder.cc
```diff
@@ -659,10 +659,15 @@ void AddTripIntersectingEdge(const AttributesController& controller,
...
+      graph_tile_ptr opp_tile = graphreader.GetGraphTile(opp_intersecting_de);
+      size_t edge_idx = intersecting_de - opp_tile->directededge(0);
+      std::vector<SignInfo> edge_signs = opp_tile->GetSigns(edge_idx);
```
kevinkreiser, line 670: the point of those functions which modify the tile is to give you the tile which contains the edge you are looking for. its highly likely that you will get the same tile so, we start with the same tile to speed up the look up, if it turns out not to be the same tile the function will update the tile to point to the right one.

then the graph id contains the sequential index in the `id` field, so you dont have to do pointer arithmetic you just use the `id` field directly:

```suggestion
      graph_tile_ptr opp_tile = graphtile;
      GraphId opp_intersecting_de = graphtile->GetOpposingEdgeId(intersecting_de, opp_tile);
      std::vector<SignInfo> edge_signs = opp_tile->GetSigns(opp_intersecting_de.id());
```

kevinkreiser (changes requested):
ive added a suggestion which i think is more inline with the rest of the code base. the operation you are trying to accomplish there is very common and my suggestion follows that common pattern

src/thor/triplegbuilder.cc
dgearhart: @kevinkreiser thanks for the explanation

ktatterso: @kevinkreiser `GraphReader::GetOpposingEdgeId()` takes a `GraphId` as its first argument.  

First question:  given a `DirectedEdge *` how do I get its `GraphId`?

Second and sort of related, I think the last line should be:
`std::vector<SignInfo> edge_signs = opp_tile->GetSigns(intersecting_de_id??);`

But again, I need the `intersecting_de`'s GraphId before I can call `.id()`....

thanks

mandeepsandhu: > @kevinkreiser `GraphReader::GetOpposingEdgeId()` takes a `GraphId` as its first argument.

@ktatterso there's an [overload](https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/graphtile.h#L254) of `GetOpposingEdgeId` that accepts a directed edge.

ktatterso: @mandeepsandhu I was just trying to follow @kevinkreiser 's suggestions.  He suggests using the `GraphId` of the `DirectedEdge * intersecting_de` - but I don't see a way to get that?  

The logic I wrote uses the method you suggest, `GraphId GraphReader::GetOpposingEdgeId(const GraphId& edgeid, graph_tile_ptr& opp_tile)`

kevinkreiser: ah yeah sorry i thought for some reason it was a graphid already. i didnt look but i guess the calling function is iterating over directed edges leaving the node which means it probably doesnt mess with edge ids either. in that case we'll need to do pointer arithmetic to get the id or update the calling function to iterate the id as well

we should probably just add a helper function to `DirectedEdge::GraphId` and `NodeInfo::GraphId` and throw the pointer arithmetic in there so anyone can do this operation more easily.

kevinkreiser: ive updated the suggestion to just use the id directly but im considering pr'ing helper functions to do the pointer arithmetic, does anyone else think those helpers would be useful?

ktatterso: Gotcha.  Well, in lieu of writing those methods, I wonder if we can move forward with my changes?  Even if those methods were written I don't think the solution would substantially different/shorter.  Although possibly a bit clearer since there wouldn't be any pointer math.

dgearhart: @kevinkreiser Now I am confused: what about the other method: https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/graphreader.h#L599-L613
image

kevinkreiser: actually im completely confused by the comment in this PR:

>       // To retrieve the sign information from the intersecting_de I need
>       // to access the graph-tile for the "begin node" of the intersecting_de.
>       // GraphReader::GetGraphTile(DirectedEdge *) returns the graph-tile for
>       // the "end node". Hence, I need to obtain the opposing-directed-edge
>       // and then ask for it's graph tile. (Sigh)

this doesnt make sense, if you look at the calling code, https://github.com/valhalla/valhalla/blob/master/src/thor/triplegbuilder.cc#L720-L735

it passes `start_tile` and `intersecting_de` which came from `start_tile`. the comment is talking about the begin node of `intersecting_de`. its begin node is already in `start_tile` which in th is method is called `graphtile`. so you already have the right graph tile for `intersecting_de`. that makes me t hink the code should just be::

```suggestion
      std::vector<SignInfo> edge_signs = graphtile->GetSigns(intersecting_de - graphtile->directededge(0));
```

dgearhart: > @kevinkreiser Now I am confused: what about the other method: https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/graphreader.h#L599-L613

nm - I was thrown off from the one comment

ktatterso: @kevinkreiser My understanding of how Sign data is stored is entirely based on "peeking" inside the original data and following through to how we store it in the Sign array in the tile. 

When you say, "this doesn't make sense", I agree.  And it's entirely possible that I misunderstand things.  However, the sign data I seek can be found using my approach - whereas it is not found using your suggested approach.

It seems to me - and again, I could be mistaken - that the Sign data for an edge is stored with respect to the edge's begin-node.  Hence the gyration to figure out the tile of the intersecting_de's begin-node.

kevinkreiser: The edges begin node is in the same exact tile that the edge is in, by definition. Getting the tile by getting its begin node would be just getting the same tile over again. If you are seeing this work then there is some else wrong. I'm going to resolve this comment and then ill make a line by line commentary to see if we can get this explained :smile:

src/thor/triplegbuilder.cc
kevinkreiser, line 667: this call call is actually very dangerous and i kind of think we should remove it! it assumes that the `endnode` of `intersecting_de` is in the tile you are calling it on. if `intersecting_de` is an edge that ends in another tile, this will get the wrong node and get the wrong opposing edge id: https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/graphtile.h#L250

if the edge doesnt leave the tile it would work. also nitpick naming this with "_de" is confusing because its a graphid not a "de"

src/thor/triplegbuilder.cc
```diff
@@ -659,10 +659,15 @@ void AddTripIntersectingEdge(const AttributesController& controller,
...
-      valhalla::baldr::graph_tile_ptr t2 = graphreader.GetGraphTile(beginnode);
-      size_t edge_idx = intersecting_de - t2->directededge(0);
-      std::vector<SignInfo> edge_signs = t2->GetSigns(edge_idx);
+      // To retrieve the sign information from the intersecting_de I need
+      // to access the graph-tile for the "begin node" of the intersecting_de.
+      // GraphReader::GetGraphTile(DirectedEdge *) returns the graph-tile for
+      // the "end node". Hence, I need to obtain the opposing-directed-edge
+      // and then ask for it's graph tile. (Sigh)
+      GraphId opp_intersecting_de = graphtile->GetOpposingEdgeId(intersecting_de);
+      graph_tile_ptr opp_tile = graphreader.GetGraphTile(opp_intersecting_de);
```
kevinkreiser, line 668: nothing wrong here, assuming we really wanted to get the tile  in which the opposing edge lives this would be just fine

src/thor/triplegbuilder.cc
```diff
@@ -659,10 +659,15 @@ void AddTripIntersectingEdge(const AttributesController& controller,
...
+      size_t edge_idx = intersecting_de - opp_tile->directededge(0);
```
kevinkreiser, line 669: here we have another problem. if `intersecting_de` is not in the same tile as its opposing edge then this pointer arithmetic would be completely off because `intersecting_de` doesnt actually occur inside of `opp_tile`s directected edges array

src/thor/triplegbuilder.cc
kevinkreiser, line 670: this is totally fine although if we have the graphid of the edge we can just call `edge_id.id()` to get the index into the directed edges.

kevinkreiser:
> There were some routing issues. These coordinates exposed the issue: -116.768599,36.9116110;-116.72504902777372,36.220713408337346101001

Is there any chance you know more specifically where it was happening along that route. It was OSM data right?

ktatterso:
@kevinkreiser Hey... ugh.  So I do believe I found the issue.  I was passing the wrong graph-tile-ptr to `AddTripIntersectingEdge`.  We call it from two places and it was a copy/paste error.

The code was "working" with my changes because when I asked for the opp-intersecting-de the edge returned was hierarchy level = 1, which then retrieved the correct tile inadvertently.

src/thor/triplegbuilder.cc
```diff
@@ -755,7 +753,7 @@ void AddIntersectingEdges(const AttributesController& controller,
           continue;
         }
 
-        AddTripIntersectingEdge(controller, graphreader, start_tile, directededge, prev_de,
+        AddTripIntersectingEdge(controller, graphreader, endtile, directededge, prev_de,
```
ktatterso, line 756: Was incorrectly passing `start_tile` before... need to pass `endtile`.

ktatterso:
Thanks to @kevinkreiser for questioning and @dgearhart and @gknisely for helping debug my error.

kevinkreiser (approved):

PR #3215, opened 2021-07-16
Avoid install of shared library if BUILD_SHARED_LIBS=OFF

mloskot:
# Issue

This is a small correction to `src/CMakeLists.txt` that ensures `make install` for build configued with `BUILD_SHARED_LIBS=OFF` does not fail with error `libvalhalla.so` or `valhalla.dll` is not found.

If static-only build is configured, then no shared library (e.g. valhalla.dll) will be available to deploy.

## Tasklist

 - [ ] Add tests


src/CMakeLists.txt
```diff
@@ -251,7 +250,13 @@ if(BUILD_SHARED_LIBS)
   install(FILES ${VALHALLA_SOURCE_DIR}/COPYING ${VALHALLA_SOURCE_DIR}/CHANGELOG.md
     DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/doc/libvalhalla0"
     COMPONENT shared)
-else()
```
mloskot, line 254: This whole `else` clause was confusing to me: why create so-symlink if `BUILD_SHARED_LIBS=OFF` at all?
```
if(UNIX)
    add_custom_target(${LIBVALHALLA_SO_LINK} ALL
...
```

kevinkreiser: good point :smile:

mloskot: I'm glad I'm not alone with the confusion ;P

kevinkreiser (approved):

PR #1120, opened 2018-02-28
Add #define NOGDI for MSVC to avoid WinSock macros clashing with enum/function names

mloskot:
This aims to prevent a couple of issues while compiling with Visual C++ on Windows:
  * `winsock2.h` includes `wingdi.h` with `#define ERROR`
    which clash with `midgard::logging::LogLevel:ERROR`
  * `winsock2.h` includes `ws2tcpip.h` with `#define GetNameInfo`
    which clash with `baldr::EdgeInfo::GetNameInfo`

PR #3216, opened 2021-07-16
WIP: Avoid LogLevel clash with Win32 common macros with unfortunate names

mloskot:
# Issue

Further refinement to PR #1120

By the way, [push_macro and pop_macro are also supported by GCC](https://gcc.gnu.org/onlinedocs/gcc/Push_002fPop-Macro-Pragmas.html), so I hope they are also available for MinGW.

## Tasklist

 - [ ] Add tests

### DIFF
--- a/valhalla/midgard/logging.h
+++ b/valhalla/midgard/logging.h
@@ -1,8 +1,13 @@
 #ifndef VALHALLA_MIDGARD_LOGGING_H_
 #define VALHALLA_MIDGARD_LOGGING_H_
 
-#if defined(_WIN32) && defined(ERROR)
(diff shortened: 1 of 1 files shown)


mloskot:
Dropping the solution as suboptimal

kevinkreiser:
We could just rename them all universally to like VALHALLA_ERROR VALHALLA_WARN etc?

mloskot:
@kevinkreiser That would be a PITA saver. Shall I submit PR?

kevinkreiser:
Yeah I don't see why not we definitely don't want to clash with other common macros

PR #3218, opened 2021-07-19
remove bg::correct to stay compatible with older boost.geometry versions

### LINKED ISSUE #3217: Remove "bg::correct" again
ref https://github.com/valhalla/valhalla/pull/3194#discussion_r665346192

now I know why I initially removed the `bg::correct`: it's not available on older boost versions. I use debian 9 (i.e. boost 1.62) to build the python bindings, which fails right now..

guess we'll have to revert that change and make it more explicit in the docs that clockwise winding is needed for `exclude_polygons` to work..

nilsnolde:
fixes #3217 

~~eft a TODO to fix winding for counterclockwise rings~~ done:)


src/loki/polygon_search.cc
```diff
@@ -32,13 +32,23 @@ static const auto Haversine = [] {
   return bg::strategy::distance::haversine<float>(vm::kRadEarthMeters);
 };
 
+void correct_ring(ring_bg_t& ring) {
```
nilsnolde, line 35: reminds me what I always wanted to ask: what's the convention in this repo for function names? pascalcase for public API, snake for internal? couldn't really make sense of it yet but was also never complained about :)

nilsnolde:
windedness is now corrected for as well @kevinkreiser

nilsnolde:
good to go from my side:)

src/loki/polygon_search.cc
kevinkreiser: once upon a time we made the huge mistake of using googles coding guidelines. they basically look like a java developer figured c++ was close enough to use the same style. i assume their motivation was minimal friction between moving around projects for developers inside of google.

i personally hated this style and whenever someone else hadn't already had tons of established google-style code i used my preferred STL-like underscores and lower case. basically a lot of it is my fault and i've caught a lot of flack for it over the years :stuck_out_tongue_winking_eye:

kevinkreiser (approved):

PR #3035, opened 2021-04-26
compatibility for older boost::optional

kevinkreiser:
`boost::optional::has_value` didn't exist in older versions of boost. but casting to boolean always did. i know its less expressive but frankly i'm pretty sure anyone who knows anything about optional knows it works "like a pointer"

PR #3219, opened 2021-07-19
replace boost::optional::has_value with more compatible is_initialized

nilsnolde:
also discovered during building with boost 1.62. has_value() was introduce in 1.68: https://www.boost.org/doc/libs/1_69_0/libs/optional/doc/html/boost_optional/relnotes.html#boost_optional.relnotes.boost_release_1_68

`has_value()` is just an alias for `is_initialized()` which has more compatibility: https://github.com/boostorg/optional/commit/5182f7f30fad87023ae5e5a4e1f660fb513fb469#diff-39c347130088a3259b9ba81c9aa6f72ed620fb23ff0cc52484c6088dc8ed6661R1336

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
    * FIXED: Fix improper iterator usage in ManeuversBuilder [#3205](https://github.com/valhalla/valhalla/pull/3205)
    * FIXED: Modified approach for retrieving signs from a directed edge #3166 [#3208](https://github.com/valhalla/valhalla/pull/3208)
    * FIXED: Improve turn channel classification: detect slip lanes [#3196](https://github.com/valhalla/valhalla/pull/3196)
+   * FIXED: Compatibility with older boost::optional versions [#3219](https://github.com/valhalla/valhalla/pull/3219)
 
 * **Enhancement**
    * CHANGED: Refactor base costing options parsing to handle more common stuff in a one place [#3125](https://github.com/valhalla/valhalla/pull/3125)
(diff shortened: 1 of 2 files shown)


nilsnolde:
btw, I know it's slightly annoying to have those backwards compat PRs for boost here and there, but (for me at least) less annoying than building boost myself in the case where we'd upgrade the min version. My plan is to build the python packages at least once a month, so I can guarantee we'll be compatible down to 1.62 for the time being.

merkispavel:
> btw, I know it's slightly annoying to have those backwards compat PRs for boost here and there, but (for me at least) less annoying than building boost myself

ah, it's second when i forget about boost compat. Sorry for that and thanks for the fix! https://github.com/valhalla/valhalla/pull/3035

merkispavel (approved):

kevinkreiser:
we can make this easier on ourselves by having a base docker image with boost built from source to the lowest version we support. we used to cheat by setting the version to the one that came with the ubuntu release that we were targetting but we've since moved to 20.04 and of course that has newer boost (also in ci) than the previous target. it sounds like at least @nilsnolde is targetting an older version of boost on some platforms which means we cant just up the boost version to the one that comes with 20.04.

nilsnolde:
how low you'd like to go with boost? 1.51 seems a little radical ;) (if it's even possible without quite some changes..). 

anyone here using this project who relies on boost < v1.62?

kevinkreiser:
i guess i was advocating for adhering to what cmake says is required, which is boost 1.51, which as i was aluding to is probably from 16.04 or maybe even 14.04 (trusty?). its possible that that doesnt even work at this point though so maybe whatever 18.04 has?

nilsnolde:
18.04. is on 1.65: https://pkgs.org/search/?q=Boost

Would you be ok with 1.62 as min version? I can amend the docker file with a custom boost. Not now, but at some point :)

kevinkreiser:
@nilsnolde let me start an issue about this and ill put my thoughts there

PR #3225, opened 2021-07-21
Update valhalla_build_config

kevinkreiser:
In a previous pr we added support for statsd #3191 

statsd supports a notion called `tags` with which you can tag a metric. there can be many usecases for these and support for tags is built into the client we use. support for tags in valhalla comes via a list of tags in the config file. the problem though is that if you didnt make your config file by hand its pretty much impossible to use. basically i had an oversight though in how `valhalla_build_config` populates that list. i let python use the default string to list conversion which takes every char in the string and makes it an element in the list. so if you did `--statsd-tags "['foo','bar']"` the json that came out would be:

```json
  "statsd": {
    "tags": [
      "[",
      "'",
      "f",
      "o",
      "o",
      "'",
      ",",
      "'",
      "b",
      "a",
      "r",
      "'",
      "]"
    ]
  },
```

This pr fixes that by using eval to let you specify the exact python syntax for a list that you want and uses that verbatim


scripts/valhalla_build_config
```diff
@@ -579,7 +582,12 @@ def add_leaf_args(path, tree, leaves, parser):
       h = h[k]
     t = comma_separated_list if type(tree) == type([]) else (type(tree) if type(tree) != type(optional()) else tree.type)
     arg = '--' + path.replace('_', '-').replace('\0', '-')
-    parser.add_argument(arg, type = t if t != type(bool()) else parse_bool, help=h, default=tree)
+    if t == type(bool()):
+      t = parse_bool
+    elif t == type(list()):
+      t = parse_list
```
kevinkreiser, line 588: moved the ternary up here to a normal if block so i could add the special case for lists. i suppose i could have used a dict as well, meh

scripts/valhalla_build_config
```diff
@@ -563,6 +563,9 @@ def parse_bool(arg):
   else:
     raise argparse.ArgumentTypeError('Boolean value expected.')
 
+def parse_list(arg):
+  return eval(arg)
```
kevinkreiser, line 567: you pass the python literal as the argument and we take it verbatim

scripts/valhalla_build_config
nilsnolde, line 588: if you wanna get rid of a few lines (untested), you can inline those functions:

```suggestion
    if t == type(bool()):
      t = lambda arg: bool(strtobool(arg))  # from distutils.util import strtobool
    elif t == type(list()):
      t = lambda arg: eval(arg)
```

nilsnolde (approved):

scripts/valhalla_build_config
keshavnandan: Is it possible to add a unit test for this change just in case?

keshavnandan (dismissed):

scripts/valhalla_build_config
kevinkreiser: i like this suggestion, let me make the changes

nilsnolde: testing the config seems like a good idea. would need a few more changes though, no @keshavnandan ? e.g. `valhalla_build_config` can't be imported by a test module currently, missing the `__init__.py` in both the `valhalla` root and the `scripts` dir. IMO `valhalla_build_config` should be copied to `build/src/bindings/python/valhalla` during the build, then the python bindings could import, use and unit test it.

kevinkreiser: at the moment we test it to make sure it works (ie returns non-zero). adding a unit test in this case would mean creating a whole new suite since we dont have one at all. im not against it, its just a decent chunk of work depending on how we do it.

regarding copying it to bindings, i humbly disagree. its not a binding to the c++ library, its just a helper script. that being said though, we've for a long time wanted to make a config object within the c++ library and get away from the unstructured ptree. if we did that work we could of course expose this to the python bindings and indeed have the same code shared all the way from the guts of the library, to the c++ unit tests, and into the python bindings (we'd probably make a c++ executable to generate configs to take the place of the python script though).

in the mean time, how about a cheaper test that is inline within the script? i can add another arg to the program that exorcises all of the args and all of their values and if something doesnt checkout it can `exit(1)` or `raise`. this program is obviously important but i feel like its not clear what the best way forward is yet.

scripts/valhalla_build_config
```diff
@@ -551,18 +552,6 @@ help_text = {
   }
 }
 
-#used to convert comma separated list of strings into a list
-def comma_separated_list(arg):
```
kevinkreiser, line 555: so we already suppored lists via comma separated strings, we just didnt support it for `optional`

scripts/valhalla_build_config
```diff
@@ -577,9 +566,19 @@ def add_leaf_args(path, tree, leaves, parser):
     h = help_text
     for k in keys:
       h = h[k]
-    t = comma_separated_list if type(tree) == type([]) else (type(tree) if type(tree) != type(optional()) else tree.type)
+
+    # its either required and is the right type or optional and has a type to use
+    # lists are supported as comma separated and bools support a bunch of conventions
+    # the rest of the primatives (strings and numbers) parse automatically
+    t = type(tree) if type(tree) != type(optional()) else tree.type
```
kevinkreiser, line 573: this is the real trick here. previously optional was only supported for string and number types. now we first resolve optional to the type and then we add the special parsing functions. this allows `optional(list)` to be su pported now

scripts/valhalla_build_config
nilsnolde: >  its not a binding to the c++ library, its just a helper script.

true. guess I'm more thinking of the bindings as a python package which is supposed to be imported in other projects/scripts. in that context it really makes a lot of sense to include the config script in the distribution, the package could create its config at runtime without relying on separate config files or system calls. depends a little how you look at it I guess..

reg inline tests, I think that's a painless sorta patchy good enough option :smile:

mandeepsandhu (dismissed):

nilsnolde (approved):

nilsnolde:
this can be merged right @kevinkreiser ?

kevinkreiser:
@nilsnolde yep just forgot about it

nilsnolde:
Think you accidentally closed instead of merged the PR @kevinkreiser 😄

kevinkreiser:
sorry, trying to burn down my inbox as quickly as possible. soon all the buttons look the same!

PR #3227, opened 2021-07-22
Get rid of typeid in statistics-related code

SiarheiFedartsou:
# Issue

typeid requires RTTI, which is not acceptable for some projects using Valhalla. So this PR refactors it in order to not use RTTI.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -1153,12 +1153,9 @@ void service_worker_t::enqueue_statistics(Api& api) const {
 
   // before we are done with the request, if this was not an error we log it was ok
   if (!api.info().error()) {
-    auto worker = typeid(*this) == typeid(loki::loki_worker_t)
-                      ? ".loki"
-                      : (typeid(*this) == typeid(thor::thor_worker_t) ? ".thor" : ".odin");
     const auto& action = Options_Action_Enum_Name(api.options().action());
 
-    statsd_client->count(action + ".info" + worker + ".ok", 1, 1.f, statsd_client->tags);
+    statsd_client->count(action + ".info." + service_name() + ".ok", 1, 1.f, statsd_client->tags);
   }
 }
 midgard::Finally<std::function<void()>> service_worker_t::measure_scope_time(Api& api) const {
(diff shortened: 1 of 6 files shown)


src/worker.cc
```diff
@@ -1160,8 +1172,10 @@ void service_worker_t::enqueue_statistics(Api& api) const {
 
     statsd_client->count(action + ".info" + worker + ".ok", 1, 1.f, statsd_client->tags);
   }
+#endif // ENABLE_STATSD_CLIENT
 }
 midgard::Finally<std::function<void()>> service_worker_t::measure_scope_time(Api& api) const {
+#ifdef ENABLE_STATSD_CLIENT
```
kevinkreiser, line 1178: i would suggest a different fix here.. while i dont disagree that disabling the statsd client is a useful thing to have. the main thin you are trying to accomplis here is to remove the need of using `typeid`. indeed this function below actually doesnt make any calls to statsd at all yet it is disabled. in fact this function adds statistics to the protobuf which we later emit through statsd but i think these stats being in protobuf is still useful regardless of sending to statsd. thought i cant blame you for thinking of them as tightly coupled.

also, statsd is disabled so long as you dont configure it. if your statsd host is either a bogus one or a blank string (the default) the client is disabled. we still make a function call but it returns immediately.

anyway i propose a different approach than this one. i propose you simple refactor the workers to know their own names.

basically add a pure virtual method to `std::string service_worker_t::name() const = 0` this will force you to implement that m ethod in all the service workers and then you can call that in the fucntions below and not have to use `typeid` at all. i think this will be simpler and not require the `ifdef` swiss cheese

kevinkreiser (changes requested):
i think we should take a different approach: https://github.com/valhalla/valhalla/pull/3227/files#r675002791

SiarheiFedartsou:
> i think we should take a different approach: https://github.com/valhalla/valhalla/pull/3227/files#r675002791

Hi @kevinkreiser 
agree with your concerns, I refactored PR to introduce `service_worker_t::service_name()` method. Could you take a look one more time?

kevinkreiser (approved):

kevinkreiser:
@SiarheiFedartsou i dont know how its possible but it seems like master failed because of the tests and it seems your build did not. i wonder if the build shortcuts are screwed up somehow and it didnt actually build your code?

kevinkreiser:
yeah, your code didnt build: 
image


cc @mandeepsandhu it looks like some trick happened here that made CI think @SiarheiFedartsou didnt change anything that needed building

kevinkreiser:
pushed a hot fix to master

mandeepsandhu:
> yeah, your code didnt build:
> image
> 
> cc @mandeepsandhu it looks like some trick happened here that made CI think @SiarheiFedartsou didnt change anything that needed building

@kevinkreiser indeed. The `needs_ci_run` runs `git diff HEAD..{commit of last successful CI}` and its possible that between [568da88](https://github.com/valhalla/valhalla/commit/568da88dc16ba8fbc46d45f4260db363eb6ca5ae) & [9cfff20](https://github.com/valhalla/valhalla/pull/3227/commits/9cfff20ac9da05658fc14c65563a83dfe441ed5c) it didn't get any files (I see that the last push was a force push, probably because it was rebased?). This tricked the script into thinking [nothing had changed](https://github.com/valhalla/valhalla/blob/master/scripts/needs_ci_run#L116-L123). I think the script should be more conservative and not skip if the list of changed files is empty (only skip if we get something and we know that something is ignore-able).

kevinkreiser:
@mandeepsandhu yeah frankly i wish we could disable force pushing, since we always squash merge there is no point to cleaning up the history unless its really important to make well-organized commits for the sake of review (which is pretty rare imho)

mandeepsandhu:
> @mandeepsandhu yeah frankly i wish we could disable force pushing, since we always squash merge there is no point to cleaning up the history unless its really important to make well-organized commits for the sake of review (which is pretty rare imho)

yeah, agreed.

I opened https://github.com/valhalla/valhalla/issues/3244 for fixing this issue in the skip ci script.

PR #3229, opened 2021-07-22
Only log errors on azure for win builds, no warnings

nilsnolde:
was driving me crazy debugging #3206 with those ridiculously verbose msbuild logs.. now it's only logging errors, no more warnings. guess that's ok, we're (read I'm ;)) not really interested optimizing win builds.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -118,7 +118,7 @@ jobs:
           for /f "delims=" %%x in ('.\vswhere.exe -latest -property InstallationPath') do set VSPATH=%%x
           popd
           call "%VSPATH%\VC\Auxiliary\Build\vcvars64.bat"
-          cmake --build %BUILD_DIR% --config %BUILD_CFG%
+          cmake --build %BUILD_DIR% --config %BUILD_CFG% -- /clp:ErrorsOnly
         displayName: 'Run CMake to build'
 
       - script: |
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -118,7 +118,7 @@ jobs:
           for /f "delims=" %%x in ('.\vswhere.exe -latest -property InstallationPath') do set VSPATH=%%x
           popd
           call "%VSPATH%\VC\Auxiliary\Build\vcvars64.bat"
-          cmake --build %BUILD_DIR% --config %BUILD_CFG% -- -clp:ErrorsOnly
+          cmd /C "cmake --build %BUILD_DIR% --config %BUILD_CFG% -- /clp:ErrorsOnly"
         displayName: 'Run CMake to build'
 
       - script: |
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -118,7 +118,7 @@ jobs:
           for /f "delims=" %%x in ('.\vswhere.exe -latest -property InstallationPath') do set VSPATH=%%x
           popd
           call "%VSPATH%\VC\Auxiliary\Build\vcvars64.bat"
-          cmake --build %BUILD_DIR% --config %BUILD_CFG% -- /clp:ErrorsOnly
+          cmake --build %BUILD_DIR% --config %BUILD_CFG% -- -clp:ErrorsOnly
         displayName: 'Run CMake to build'
 
       - script: |


nilsnolde:
no changelog is fine right?

mloskot (dismissed):

nilsnolde:
thx @mloskot azure is failing though;)

I don't understand though.. 🙄 `ninja: error: unknown target '/clp:ErrorsOnly` and the cmake build command is `cmake --build %BUILD_DIR% --config %BUILD_CFG% -- /clp:ErrorsOnly`. thinking that after the `--` cmake treats those as compiler arguments (apparently it's a generator argument). Locally it works like that with the visual studio generator.. any idea by any chance?

ninja is more trouble on ci than it's worth..

nilsnolde:
oh no, it's treating is as a target.. why..

mloskot:
Try `cmake --build %BUILD_DIR% --config %BUILD_CFG% -- -clp:ErrorsOnly` perhaps

kevinkreiser:
nice, once its working feel free to merge, no changelog needed

nilsnolde:
uff, shelving it for now.. `-` instead of `/` didn't work, also wrapping the whole thing in quotes with `cmd /C "xx"` didn't work.. not sure if it's a cmake bug from an older version or what.. it did work locally on a fairly recent cmake (3.19 or so)

mloskot:
I have [asked on #cmake channel at cpplang.slack.com](https://cpplang.slack.com/archives/C5Y2GDECX/p1627025852337700). I'll forward any answers.

mloskot:
Let's scratch my https://github.com/valhalla/valhalla/pull/3229#issuecomment-885459690 above

@nilsnolde Following an enlightening moment, I think I know what's the problem:
The `-clp`, or [-consoleLoggerParameters](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference#switches) for long version, is MSBuild switch, and Ninja does not recognise it, and assumes it's a target name.

If the build switched CMake to use MSBuild generator instead of Ninja, then the `-clp:ErrorsOnly' should work.

nilsnolde:
argh.. yeah.. I was entirely on the wrong track.. confused generators with compilers.. now I'm even more confused.. is a compiler directive needed here (in theory)? ninja doesn't seem to care about log levels at all

mloskot:
In [`cmake --build <dir> [<options>] [-- <build-tool-options>]`](https://cmake.org/cmake/help/latest/manual/cmake.1.html) the `build-tool-options` are "options to the native tool" where the tool is the build system for which CMake generated the configuration files ("Unix Makefiles", Ninja, "Visual Studio...", etc.).

So, "the tool" is not a compiler, but build system.
So, the switches after ` -- ` are passed to build system, not to compiler.

If Ninja does not offer a switch to control logging of errors or warnings, then the only way to disable the warnings is to pass compiler flags to the CL.EXE, specifically, the [/W0 to disable all warnings](https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level).

Question is, how to reliably overwrite the default `/W...` flags with `/W0`, because

> [Command line options like CMAKE_C_FLAGS and CMAKE_CXX_FLAGS always append and never overwrite.](https://stackoverflow.com/a/62888014/151641)

The canonical way is to use toolchain file where `CMAKE_CXX_FLAGS` can be _initialised_.

Alternatively, Valhalla's `CMakeLists.txt` could have CI-dedicated setting:

```cmake
if(DEFINED ENV{CI} OR DEFINED ENV{GITHUB_ACTION} OR DEFINED ENV{SYSTEM_TEAMFOUNDATIONCOLLECTIONURI}) 
  target_compile_options(target PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/W0>)
endif()
```

For CMake earlier than 3.15, due to 'bug' https://gitlab.kitware.com/cmake/cmake/-/issues/18317
it may be necessary to find-and-replace `/W3` with `/W0` in the flags, see https://stackoverflow.com/a/3818084/151641

nilsnolde:
> So, the switches after -- are passed to build system, not to compiler.

yeah, that's what I was confusing.

thanks @mloskot for the detailed guide, really appreciate it. I'll try the cmake ci-dedicated one

PR #3233, opened 2021-07-25
update base docker image before building

nilsnolde:
test if circle ci is updating the base image regularly. if this pulls a new images then they don't and we should do it ourselves I think.

yeah, seems it's not at the latest:

```
20.04: Pulling from library/ubuntu

Digest: sha256:b3e2e47d016c08b3396b5ebe06ab0b711c34e7f37b98c9d37abe794b71cea0a2
Status: Downloaded newer image for ubuntu:20.04
```


.circleci/config.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       - run:
           name: Build Valhalla Docker Images
           command: |
+            docker pull ubuntu:20.04
             docker build -f ./docker/Dockerfile-build --tag valhalla/valhalla:build-latest ./docker
```
ImreSamu, line 28: imho:  the `docker build --pull -f ./docker/Dockerfile-build --tag valhalla/valhalla:build-latest ./docker` is better/safer 
see: https://docs.docker.com/engine/reference/commandline/build/
 * `--pull |   | Always attempt to pull a newer version of the image`

nilsnolde: jup agreed, didn't know that flag.. it doesn't specifically say "pull base image" but anything else would be weird;)

ImreSamu: :-)  not critical suggestions ... 
in the `docker-postgis` repo - this is already used: 
https://github.com/postgis/docker-postgis/blob/dabd4d3824cf32e633cf2a6f07515e5e13bc3774/Makefile#L79

nilsnolde: done

kevinkreiser (approved):

nilsnolde:
hmpf, need another approval after merging master (even though I resolved the conflict in github's UI..)

ImreSamu (approved):

nilsnolde:
should we get this tiny one merged too?

kevinkreiser (approved):

PR #3234, opened 2021-07-26
pedestrian cost to take hill grade into account

farfromrefug:
The one thing i am not sure of are the `kGradeBasedSpeedFactor` and `kAvoidHillsStrength`. I kind of guessed it.I am not sure if you have a better way of "computing" those.


nilsnolde (commented):
just a preliminary review, others might have more to say about this. 

also you'll need to run the `format.sh` script etc (according to [guidelines](https://github.com/valhalla/valhalla#contributing)), and lastly we'd want some tests making sure this won't break in the future.

src/sif/pedestriancost.cc
```diff
@@ -815,6 +881,12 @@ void ParsePedestrianCostOptions(const rapidjson::Document& doc,
     pbf_costing_options->set_bike_share_penalty(
         kBSSPenaltyRange(rapidjson::get_optional<uint32_t>(*json_costing_options, "/bss_rent_penalty")
                              .get_value_or(kDefaultBssPenalty)));
+      
+      
+    // use_hills
+    pbf_costing_options->set_use_hills(
+        kUseHillsRange(rapidjson::get_optional<float>(*json_costing_options, "/use_hills")
```
kevinkreiser, line 888: @farfromrefug we've been putting off doing this for forever, thanks for contributing this!

kevinkreiser:
@dnesbitt61 or @dgearhart i'm wondering if either of you would have a bit of free time to take a look at this as well

farfromrefug:
@kevinkreiser i ll look at all the comments next week, pretty much overbooked right now

src/sif/pedestriancost.cc
```diff
@@ -620,14 +683,17 @@ Cost PedestrianCost::EdgeCost(const baldr::DirectedEdge* edge,
   }
 
   float sec =
-      edge->length() * speedfactor_ * kSacScaleSpeedFactor[static_cast<uint8_t>(edge->sac_scale())];
+      edge->length() * speedfactor_ * kSacScaleSpeedFactor[static_cast<uint8_t>(edge->sac_scale())] / kGradeBasedSpeedFactor[edge->weighted_grade()];
```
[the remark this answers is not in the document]
dnesbitt61: I would avoid division here and precompute these as 1/xxx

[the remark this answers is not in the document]
farfromrefug: @kevinkreiser It is divided to work like`kSacScaleSpeedFactor` to slow down when > 1. But `kGradeBasedSpeedFactor`  is "inversed" in the factors.
AS @dnesbitt61 asked i will inverse them

farfromrefug:
@kevinkreiser think i changed them all. The remaining things are the factors to see if we all think they are correct. That web site https://www.myopencountry.com/average-hiking-speed/ talked about a 30% decrease in speed per 5.5degrees.

src/sif/pedestriancost.cc
```diff
@@ -151,6 +156,54 @@ BaseCostingOptionsConfig GetBaseCostOptsConfig() {
 
 const BaseCostingOptionsConfig kBaseCostOptsConfig = GetBaseCostOptsConfig();
 
+// Speed adjustment factors based on weighted grade. Comments here show an
+// example of speed changes based on "grade", using a base speed of 5 KPH
+// on flat roads. 
+// Tobler seems a bit too "fast" uphill so we use
```
kevinkreiser, line 162: this sentance looks like it trails off, what are we using for uphill?

src/sif/pedestriancost.cc
```diff
@@ -151,6 +156,54 @@ BaseCostingOptionsConfig GetBaseCostOptsConfig() {
...
+    1.69f,  // 13%   - 2.11
+    1.82f   // 15%   - 1.97
+};
+
+// Avoid hills "strength". How much do we want to avoid a hill. Combines
+// with the usehills factor (1.0 - usehills = avoidhills factor) to create
+// a weighting penalty per weighted grade factor. This indicates how strongly
+// edges with the specified grade are weighted. Note that speed also is
+// influenced by grade, so these weights help further avoid hills.
+constexpr float kAvoidHillsStrength[] = {
```
kevinkreiser, line 188: i can see us wanting to change the shape of this curve depending on the mode of travel in the future like for wheelchair users for example. we can save that for another day though, just wanted to mention it

src/sif/pedestriancost.cc
```diff
@@ -620,14 +683,14 @@ Cost PedestrianCost::EdgeCost(const baldr::DirectedEdge* edge,
   }
 
   float sec =
-      edge->length() * speedfactor_ * kSacScaleSpeedFactor[static_cast<uint8_t>(edge->sac_scale())];
+      edge->length() * speedfactor_ * kSacScaleSpeedFactor[static_cast<uint8_t>(edge->sac_scale())] * kGradeBasedSpeedFactor[static_cast<uint8_t>(edge->weighted_grade())];
```
kevinkreiser, line 686: this is much more intuitive to me now. i still wonder t hough about the interplay between sac_scale and grade. we will pretty much have grade coverage for the whole dataset but sac scale will be sparsely populated. what i think that will mean is that any place that actually tagged sac scale will get penalized while anything untagged will get preference. does anyone else have thoughts/worries about that? basically im worried that having both is unfair to sac scale tagged edges because they are rare and we'll end up accidentally avoiding them just because hardly anything is tagged with that (though i guess this is already the case)

farfromrefug:
@kevinkreiser lint done! is that enough for you on my side?

kevinkreiser:
@farfromrefug would you mind also adding one row in the costing options table in the docs? you can copy paste the one from the bike costing.

aside from that you are off the hook. on our side we just need to do what we call RAD testing to see how pedestrian routes may have changed due to the code changes. if those look ok im good to merge!

dgearhart:
@kevinkreiser i will take a look in the morning

farfromrefug:
@kevinkreiser done hope it is good.

kevinkreiser:
@farfromrefug looks great to me

@dgearhart :pray: thank you!

dgearhart:
@farfromrefug I am [RAD](https://github.com/valhalla/valhalla/blob/master/run_route_scripts/README.md) testing (master vs. pedestrian_hill branch) with the following two files:

- [pedestrian_routes.txt](https://github.com/valhalla/valhalla/blob/master/test_requests/pedestrian_routes.txt)
- [pedestrian_roundabout_routes.txt](https://github.com/valhalla/valhalla/blob/master/test_requests/pedestrian_roundabout_routes.txt)

I did not expect to see many diffs - however, I am seeing 28 diffs out of 71 for the `pedestrian_routes.txt` test file and 3 diffs out of 18 for the `pedestrian_roundabout_routes.txt]` test file.

Can someone confirm they are seeing the same and explain the diffs?

/cc @kevinkreiser @gknisely

farfromrefug:
@dgearhart were are those routes? Are there hills there ? Maybe it comes from the changes in cost/speed for very low grade routes ?

kevinkreiser:
i definitely think there will be diffs, exactly because of the reason @farfromrefug mentioned, we modulate the speed based on the grade, which i personally think is a huge improvement. the question would be, are the diffs clearly bad?

dgearhart:
@kevinkreiser @farfromrefug Concerned when I see diffs like:
[pedestrian_roundabout_routes.txt#2](https://github.com/valhalla/valhalla/blob/master/test_requests/pedestrian_roundabout_routes.txt#L2)
image

farfromrefug:
@dgearhart any way we can see both on a map? But yes the uturn seems not normal!

dgearhart:
@farfromrefug 
## Master
image

## Branch
image

@farfromrefug this one looks like a data issue with the roundabout - I looking at the OSM data

dgearhart (approved):
I updated data for the roundabout u-turn.
The other diffs seem ok

kevinkreiser:
oh i see it must have been some tiny little shape stuff or something? anyway great news, thanks very much for the review @dgearhart

src/sif/pedestriancost.cc
hungerburg: Indeed, using sac_scale to penalize would be wrong. Unless plain "path" without much specifics would penalize even more.

PR #3238, opened 2021-07-27
Reject alternatives with too long detours

genadz:
# Issue

A motivation of this PR is to reject bad alternatives that have unreasonable long detours comparing to the optimal route.

Code is simple enough: find a different segment for two routes and compare their costs. In case cost of alternative segment is much bigger than cost of the optimal segment -> reject such alternative.

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/alternates.cc
+++ b/src/thor/alternates.cc
@@ -15,7 +15,13 @@ using namespace valhalla::midgard;
  */
 namespace {
 // Defaults thresholds
+// TODO: global stretch parameter should depend on the optimal route cost/duration.
+// For an optimal route that takes 10min it's okay to have an alternative that takes 15min. But
+// it's unreasonable to propose an alternative that takes 15 hours if the optimal one takes 10 hours.
 float kAtMostLonger = 1.25f; // stretch threshold
+// Alternative route shouldn't contain unreasonable detours. We should skip an alternative
+// if it has a detour longer than 2 x cost of the corresponding path in the optimal route.
+float kAtMostLongerDetour = 2.f;
 float kAtMostShared = 0.75f; // sharing threshold
 // float kAtLeastOptimal = 0.2f; // local optimality threshold
 } // namespace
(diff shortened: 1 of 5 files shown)


test/gurka/test_alternates.cc
```diff
@@ -75,4 +75,32 @@ TEST(Alternates, test_long_route) {
...
+TEST(Alternates, test_too_long_detour) {
+  const std::string ascii_map = R"(
+       A------------B-C------------D
+                    | |
+                    | |
+                    | |
+                    | |
+                    | |
+                    E-F
+    )";
```
genadz, line 89: typical example of an alternative that should be rejected by these changes

src/thor/alternates.cc
```diff
@@ -15,7 +15,13 @@ using namespace valhalla::midgard;
  */
 namespace {
 // Defaults thresholds
+// TODO: global stretch parameter should depend on the optimal route cost/duration.
```
genadz, line 18: I'm sure that we miss some good alternatives for short routes because of one global threshold. Added as "TODO", but it should be transformed into an issue/PR

genadz:
1. **No changes in performance**.
2. This change **affects ~1-2%** of routes. Let me share several examples:

#### No good alternatives were found

| before | after |
| --- | --- |
| 1_before | 1_after |
| 2_before | 2_after |
| 3_before | 3_after |

#### Other alternatives were proposed

| before | after |
| --- | --- |
| 4_before | 4_after |
| 5_before | 5_after |
| 6_before | 6_after |

src/thor/alternates.cc
genadz: and, imho, we should also refactor `max_sharing` coefficient in the following way:
1. a little bit increase `max_sharing` coefficient for shorter routes. I think value `0.6` for routes < 20km is too strict, it should be ~0.7.
2. replace piecewise constant function with a smoother one (cubic?) to ensure that for two close routes `max_sharing` constant values also will be close. It's a little bit strange that for 9.9-km route we will use `max_sharing` constant equals to `0.5` but for 10-km route this constant will be `0.6`.

kevinkreiser: both of these suggestions make way more sense to me as well :+1:

src/thor/alternates.cc
```diff
@@ -56,7 +62,74 @@ void filter_alternates_by_stretch(std::vector<CandidateConnection>& connections)
...
+
+  return {{idx_1_first, idx_1_last}, {idx_2_first, idx_2_last}};
+}
+
+// Check if the candidate path contains unreasonable long detours comparing to the optimal path.
+bool validate_alternate_by_stretch(const std::vector<PathInfo>& optimal_path,
+                                   const std::vector<PathInfo>& candidate_path) {
+  const auto diff_segment = find_diff_segment(optimal_path, candidate_path);
+  const auto& idxs_opt = diff_segment.first;
+  const auto& idxs_alt = diff_segment.second;
```
kevinkreiser, line 110: this naming is difficult for my brain. `idxs_opt` is the segment of the optimal path which is different and `idxs_alt` is the segment of the candidate path that is different. what about like `unique_optimal_segment` and `unique_candidate_segment`? maybe also `diff_segment` would be like `unique_segments`? i dont know just a suggestion, though pretty long names...

genadz: I like that, thank you !

src/thor/alternates.cc
```diff
@@ -56,7 +62,74 @@ void filter_alternates_by_stretch(std::vector<CandidateConnection>& connections)
...
+
+  if (idxs_opt.first == optimal_path.size()) {
```
kevinkreiser, line 112: we could use a comment here i think, its hard to understand that this is really the second clause of the comment in the next `if` block. basically its saying, there was no start to a unique part in the optimal path, as all its edges were found in the alternate path. then the second `if` is really saying there was a unique part in the alternate path past the end of the optimal, which is just wasteful so reject it

kevinkreiser: by the way is there any way to make that happen in gurka? im guessing its pretty difficult. you have to have two loki::search candidates on both locations so that the search paths can go the "wrong" way and then meet after the "right" way finds the optimal route

kevinkreiser (approved):
this looks great to me. if you could comment a little more the logic in the main function and if you can figure out how to test those pathological cases that would be great but this is good either way

src/thor/alternates.cc
genadz: I've met such corner case when tested this PR; but it's difficult to reproduce

kevinkreiser: no worries, just wondered if you tried

kevinkreiser (approved):

PR #3240, opened 2021-07-28
Unpaved roads

maksimandrianov:
# Issue

Added the new option `exclude_unpaved`. This value indicates the willingness to take unpaved roads. If `exclude_unpaved` is set to 1, unpaved roads are excluded from the process of generating the route path, otherwise included.

## Tasklist

 - [x] Add tests

## Requirements / Relations


maksimandrianov:
The light demo of using the use_unpaved_roads option:

| Use unpaved roads  | Do not use unpaved roads | OSM |
| ------------- | ------------- | ------------- |
| Screenshot 2021-08-04 at 15 46 06 | Screenshot 2021-08-04 at 15 45 53 | [Unpaved OSM way](https://www.openstreetmap.org/way/474096633)<br/> Screenshot 2021-08-04 at 16 16 53 | 
| Screenshot 2021-08-04 at 15 46 32 | Screenshot 2021-08-04 at 15 46 23 | [Unpaved OSM way](https://www.openstreetmap.org/way/17021717)<br/> Screenshot 2021-08-04 at 16 27 56 |
| Screenshot 2021-08-04 at 15 47 06 |  Screenshot 2021-08-04 at 15 46 40 | [Unpaved OSM way](https://www.openstreetmap.org/way/17055444)<br/> Screenshot 2021-08-04 at 16 32 24 |

kevinkreiser:
@maksimandrianov can you please fill out the description of the pr just for book-keeping purposes. just a brief description of the changes. also dont forget to add it to the costing documentation in `/docs`

maksimandrianov:
There are two implementations of the exclusion unpaved roads from a route in this PR:
* Do not allow to use unpaved roads in the beginning and end of a route. (The commit -  https://github.com/valhalla/valhalla/pull/3240/commits/b88ba5eec7c01868b217f3fef6ec932e78ab4827)
* Allow to use unpaved roads in the beginning and end of a route. (This branch)

Examples:
| Do not allow | Allow | OSM |
| ----- | ----- | ----- |
| No route found | Screenshot 2021-08-04 at 15 47 43 | [Unpaved OSM way](https://www.openstreetmap.org/way/474096633) <br/> [Unpaved OSM way](https://www.openstreetmap.org/way/17016265) <br/> [Unpaved OSM way](https://www.openstreetmap.org/way/17025967) <br/> Screenshot 2021-08-04 at 17 17 41 |
| No route found | Screenshot 2021-08-04 at 15 47 55 | [Unpaved OSM way](https://www.openstreetmap.org/way/516664652) <br/> Screenshot 2021-08-04 at 17 23 51 |

merkispavel:
> There are two implementations of the exclusion unpaved roads from a route in this PR:
-Do not allow the use of unpaved roads in the beginning and end of a route. (The commit - b88ba5e)
-Allow the use of unpaved roads in the beginning and end of a route. (This branch)

origin and destination points are usually set intentionally so if one of them gets snapped to the unpaved road - why shouldn't we use it. The second (up-to-date) implementation definitely makes sense to me 👍 
Though there's a small pitfall. In this case we don't say "get out of the unpaved road asap" so we can make completely-unpaved route in theory. It doesn't seem like a blocker for this PR imho, just wanted to share

gknisely:
@kevinkreiser @dnesbitt61 don't we usually have a range(0 to 1) for use_* options?  Not sure if the naming (use_unpaved_roads) is correct here.  Thoughts?  cc @maksimandrianov

kevinkreiser:
@gknisely yes exactly, use_* options are in the range of 0-1, meaning they are not hard avoids but a soft avoid on a finite scale.

if you want to exclude certain types of roads from the start or end of the route, as in its not allowed to use those as candidates, then you should look into how loki does excludes. that is done in a different part of the request using the search filter. here is where it gets applied: https://github.com/valhalla/valhalla/blob/master/src/loki/search.cc#L38-L42

gknisely:
> @gknisely yes exactly, use_* options are in the range of 0-1, meaning they are not hard avoids but a soft avoid on a finite scale.
> 
> if you want to exclude certain types of roads from the start or end of the route, as in its not allowed to use those as candidates, then you should look into how loki does excludes. that is done in a different part of the request using the search filter. here is where it gets applied: https://github.com/valhalla/valhalla/blob/master/src/loki/search.cc#L38-L42

@maksimandrianov I think that the above is a better implementation.  Thoughts ^

kevinkreiser:
I agree that 0 should not be a hard exclusion of unpaved roads in the path but that they should be severely penalized. in the cases where unpaved roads are the only way between two locations we need to allow for that and this is the strategy that the project has followed since we started allowing tweaking costing in this way.

we can do hard exclusion on the search candidates as i pointed out above but have yet to add hard exclusion in the path finding  phase. not that we couldnt do that, but we need to do a bit of a design process there to make sure it works for all cases where we would want to do that. id argue that is not so useful and not so common that a person will tolerate absolutely no unpaved roads or no highways or not tolls for example. if that is the only way to go, they'd rather have A route than none or than one that makes them drive 2 hours out of their way

genadz:
@merkispavel
> Though there's a small pitfall. In this case we don't say "get out of the unpaved road asap" so we can make completely-unpaved route in theory. It doesn't seem like a blocker for this PR imho, just wanted to share

To avoid that we can penalize unpaved roads more than ordinary roads.

genadz:
> I agree that 0 should not be a hard exclusion of unpaved roads in the path but that they should be severely penalized. in the cases where unpaved roads are the only way between two locations we need to allow for that and this is the strategy that the project has followed since we started allowing tweaking costing in this way.
> 
> we can do hard exclusion on the search candidates as i pointed out above but have yet to add hard exclusion in the path finding phase. not that we couldnt do that, but we need to do a bit of a design process there to make sure it works for all cases where we would want to do that. id argue that is not so useful and not so common that a person will tolerate absolutely no unpaved roads or no highways or not tolls for example. if that is the only way to go, they'd rather have A route than none or than one that makes them drive 2 hours out of their way

@kevinkreiser I think in this case hard exclusion is more preferable and clear way. There are several reasons for that I see:
- It's expected and completely clear behavior to return a route without unpaved roads in case I set `exclude_unpaved=true`. I will be very surprise if I get a route with unpaved edges. From my perspective it's worse to get a route with unpaved roads in case there is a route without unpaved roads rather than return "No route" in case the only way to go is to use unpaved edges.
- Choosing correct penalty it's very difficult task. This value should be aligned with other many constants that we already use for other options. Moreover, it should be aligned with other options that we will introduce in the future. So, we can't guarantee for 100% that `avoid_unpaved` will work correctly with all combinations of options that we have for now (like `use_tracks`, `use_tolls` and other that have big penalties).
- I do believe that there are very few cases when unpaved road is the only way to got from origin to destination. As for me it's okay to return "No route", a user takes a responsibility for setting parameters. But, if we really want to cover such cases and return a route, we can do that on the second pass of the path finding algorithm.

maksimandrianov:
> @gknisely yes exactly, use_* options are in the range of 0-1, meaning they are not hard avoids but a soft avoid on a finite scale.
> 
> if you want to exclude certain types of roads from the start or end of the route, as in its not allowed to use those as candidates, then you should look into how loki does excludes. that is done in a different part of the request using the search filter. here is where it gets applied: https://github.com/valhalla/valhalla/blob/master/src/loki/search.cc#L38-L42

But we want to see the hard exclusion here. I cannot imagine how the level of avoidance of unpaved roads can help drivers.  
I have used the dest_only example, when implemented this.

maksimandrianov:
> @kevinkreiser @dnesbitt61 don't we usually have a range(0 to 1) for use_* options? Not sure if the naming (use_unpaved_roads) is correct here. Thoughts? cc @maksimandrianov

If it is a problem I can rename this variable.  `exclude_unpaved_roads`? Or do you have any suggestion of a better readable name?

genadz:
> > @kevinkreiser @dnesbitt61 don't we usually have a range(0 to 1) for use_* options? Not sure if the naming (use_unpaved_roads) is correct here. Thoughts? cc @maksimandrianov
> 
> If it is a problem I can rename this variable. `exclude_unpaved_roads`? Or do you have any suggestion of a better readable name?

what about `exclude_unpaved` ? to be aligned with other parameters

kevinkreiser:
if you want to start a new type of option that does hard exclusion that is fine. but the `use_*` options are soft exclusions so we cannot use this nomenclature. since exclusions are not on a floating scale why not just support a list of them  similar to `search_filter`. maybe even just call it `filter`. then you can do like: `"filter":["exclude_unpaved", "exclude_bridge", "exclude_motorway"]`. at least in this way the the API is a bit consistent.

due note that this is very tricky. @mandeepsandhu and i worked a lot on how to combine both the search_filter for closures and the costing option ignore_closures. basically if you want the routing algorithm to include or exclude something but you dont do the same in the edge candidate search then you can fail the route very easily.

to me it makes sense to propose a design for things as big as hard exclusions before we implement them so we all have a shared understanding up front. @maksimandrianov @genadz filter with the excludes listed makes sense to me but my question is where does it live. the obvious choice would be in costing as then we dont have to implement it in each algorithm. then the next question is how do we protect against users allowing unpaved search candidates and then immediateliy failing the route once the thor gets only unpaved edge candidates and needs to exclude them. for closures we did special logic to let the user start and end on them if they wanted to because that might be where they actually are but after that they didnt want to go through any closures. the same could be said for unpaved roads.

if this is confusing, good, thats about what it should be :smile: let me know if you want to have a voice chat about it to clear it up

genadz:
@kevinkreiser 
> for closures we did special logic to let the user start and end on them if they wanted to because that might be where they actually are but after that they didnt want to go through any closures. the same could be said for unpaved roads.

this is exactly what this PR proposes

> -Allow the use of unpaved roads in the beginning and end of a route. (This branch)

maksimandrianov:
> if you want to start a new type of option that does hard exclusion that is fine. but the `use_*` options are soft exclusions so we cannot use this nomenclature. since exclusions are not on a floating scale why not just support a list of them similar to `search_filter`. maybe even just call it `filter`. then you can do like: `"filter":["exclude_unpaved", "exclude_bridge", "exclude_motorway"]`. at least in this way the the API is a bit consistent.
> 
> due note that this is very tricky. @mandeepsandhu and i worked a lot on how to combine both the search_filter for closures and the costing option ignore_closures. basically if you want the routing algorithm to include or exclude something but you dont do the same in the edge candidate search then you can fail the route very easily.
> 
> to me it makes sense to propose a design for things as big as hard exclusions before we implement them so we all have a shared understanding up front. @maksimandrianov @genadz filter with the excludes listed makes sense to me but my question is where does it live. the obvious choice would be in costing as then we dont have to implement it in each algorithm. then the next question is how do we protect against users allowing unpaved search candidates and then immediateliy failing the route once the thor gets only unpaved edge candidates and needs to exclude them. for closures we did special logic to let the user start and end on them if they wanted to because that might be where they actually are but after that they didnt want to go through any closures. the same could be said for unpaved roads.
> 
> if this is confusing, good, thats about what it should be 😄 let me know if you want to have a voice chat about it to clear it up

I dont say that I understand you.. Could you get a gurka test that shows me when it will fail please?

maksimandrianov:
> > if you want to start a new type of option that does hard exclusion that is fine. but the `use_*` options are soft exclusions so we cannot use this nomenclature. since exclusions are not on a floating scale why not just support a list of them similar to `search_filter`. maybe even just call it `filter`. then you can do like: `"filter":["exclude_unpaved", "exclude_bridge", "exclude_motorway"]`. at least in this way the the API is a bit consistent.
> > due note that this is very tricky. @mandeepsandhu and i worked a lot on how to combine both the search_filter for closures and the costing option ignore_closures. basically if you want the routing algorithm to include or exclude something but you dont do the same in the edge candidate search then you can fail the route very easily.
> > to me it makes sense to propose a design for things as big as hard exclusions before we implement them so we all have a shared understanding up front. @maksimandrianov @genadz filter with the excludes listed makes sense to me but my question is where does it live. the obvious choice would be in costing as then we dont have to implement it in each algorithm. then the next question is how do we protect against users allowing unpaved search candidates and then immediateliy failing the route once the thor gets only unpaved edge candidates and needs to exclude them. for closures we did special logic to let the user start and end on them if they wanted to because that might be where they actually are but after that they didnt want to go through any closures. the same could be said for unpaved roads.
> > if this is confusing, good, thats about what it should be 😄 let me know if you want to have a voice chat about it to clear it up
> 
> I dont say that I understand you.. Could you get a gurka test that shows me when it will fail please? Do you read gurka tests in this PR?

kevinkreiser:
@genadz @maksimandrianov its not about the implementation. the implementation looks pretty good to me, i even like the change in semantics. let me try to explain again:

`_roads` is redundant and `use_` implies that its an "avoid". it makes no sense to have the meaning of one of them be different from all of the rest. look at all the other `use_` options, they are "avoids" not "excludes". basically 0 means maximum penalty and 1 means minimum penalty. these are the semantics of the `use_` options. maybe you want to say that it was an oversight or a poor design decision but it remains that the library should, in so far as it can, remain consistent so that people are not suprised by how an option works.

in summary:
1. i dont think its ok to have different semantics for options of the same type. all `use_` options should work the same
2. i actually like the change in semantics because it allows both avoid and exclude in the same parameter (less options for more features is good)
3. we could change all `use_` options to work like this but that would be a breaking change. existing requests would start hard excluding stuff if set to 0

i think the best way around this is probably through the deprecation of the `use_` options in favor of another name.  this is a tactic for migrating existing users while still being able to change the api. in this case we would:

1. mark all the `use_` options as deprecated in the documentation and add replacement options. need to think of a name
2. parse out the `use_` options in the request as we already do, but if the person set it to 0, we could map it to `.0001` or something. this would signal to give it the max penalty or very close but not hard exclude.
2. if you wanted to use hard exclude youd need to use the new name for this parameter
3. we implement this change as @maksimandrianov has done for the other `use_` options (the hills one is tricky because its not binary though)

maksimandrianov:
> 3\. we could change all `use_` options to work like this but that would be a breaking change. existing requests would start hard excluding stuff if set to 0

Unfortunately, if `use_*` set to 0, it does not mea hard exclusion - https://github.com/valhalla/valhalla/blame/master/docs/api/turn-by-turn/api-reference.md#L105

I will fix naming, thank you for the detailed explanation!

genadz:
> 1. i dont think its ok to have different semantics for options of the same type. all use_ options should work the same

completely agree, we shouldn't put different meanings for two `use_*` options

kevinkreiser:
@maksimandrianov @genadz yeah i think if we change the name to something else, then we could also migrate the other use_ options to match that naming and depricate as i mentioned in the last part of my previous comment

src/sif/autocost.cc
```diff
@@ -412,7 +412,8 @@ bool AutoCost::Allowed(const baldr::DirectedEdge* edge,
       ((pred.restrictions() & (1 << edge->localedgeidx())) && !ignore_restrictions_) ||
       edge->surface() == Surface::kImpassable || IsUserAvoidEdge(edgeid) ||
       (!allow_destination_only_ && !pred.destonly() && edge->destonly()) ||
-      (pred.closure_pruning() && IsClosed(edge, tile))) {
+      (pred.closure_pruning() && IsClosed(edge, tile)) ||
+      (!use_unpaved_roads_ && !pred.unpaved() && edge->unpaved())) {
```
kevinkreiser, line 416: i like this trick of using a bit to specify whether the predecessor was or wasnt unpaved instead of keeping track of the pruning state. i think its easier to reason about than setting and unsetting the pruning. we should switch to doing this for closures. cc @mandeepsandhu

dnesbitt61:
The bicycle costing option avoid_bad_surfaces does a hard exclusion when set to a value of 1. Not sure if I like the naming of this option, but it does mix penalties and exclusion.

kevinkreiser:
> The bicycle costing option avoid_bad_surfaces does a hard exclusion when set to a value of 1. 

@dnesbitt61 i had the same thought, use the term `avoid_` reverse the range (since it has the opposite meaning) and then deprecate all of the `use_` options. and i had the same thought that i didnt love the naming but i have a feeling there is no naming that i would love :smile:

maksimandrianov:
> > The bicycle costing option avoid_bad_surfaces does a hard exclusion when set to a value of 1.
> 
> @dnesbitt61 i had the same thought, use the term `avoid_`, reverse the range (since it has the opposite meaning) and then deprecate all of the `use_` options. and i had the same thought that i didnt love the naming but i have a feeling there is no naming that i would love 😄

May be do we leave `use_*` option naming as it is? I agree with @kevinkreiser, because `avoid` and `use`  are similar in clarity/readability terms.

maksimandrianov:
@kevinkreiser, @maksimandrianov naming was fixed, PTAL

docs/api/turn-by-turn/api-reference.md
```diff
@@ -122,6 +122,7 @@ The following options are available for `auto`, `bus`, `hov`, `taxi`, and `truck
 | :-------------------------- | :----------- |
 | `height` | The height of the vehicle (in meters). |
 | `width` | The width of the vehicle (in meters). |
+| `exclude_unpaved` | This value indicates the willingness to take unpaved roads. If `exclude_unpaved` is set to 1, unpaved roads are excluded from the process of generating the route path, otherwise included. |
```
kevinkreiser, line 125: i think clearer language around this option being binary would be helpful

```suggestion
| `exclude_unpaved` | This value indicates the whether or not the path should include unpaved roads. If `exclude_unpaved` is set to 1, unpaved roads not allowed to be used in the process of generating the route path, otherwise they are allowed. |
```

src/mjolnir/pbfgraphparser.cc
```diff
@@ -653,14 +653,18 @@ struct graph_callback : public OSMPBF::Callback {
       } else if (value.find("paved") != std::string::npos ||
```
kevinkreiser, line 653: @gknisely do these look good to you?

kevinkreiser (approved):
other than the change to the docs wording, this looks good to me. i would like @gknisely 's opinion on the tagging values though

kevinkreiser:
@maksimandrianov please add an entry to the changelog. a new request parameter certainly deserves an entry in the changelog

src/mjolnir/pbfgraphparser.cc
gknisely: @kevinkreiser checking

gknisely: @kevinkreiser they look good.

docs/api/turn-by-turn/api-reference.md
```diff
@@ -122,6 +122,7 @@ The following options are available for `auto`, `bus`, `hov`, `taxi`, and `truck
 | :-------------------------- | :----------- |
 | `height` | The height of the vehicle (in meters). |
 | `width` | The width of the vehicle (in meters). |
+| `exclude_unpaved` | This value indicates the whether or not the path may include unpaved roads. If `exclude_unpaved` is set to 1, unpaved roads are not allowed to be used in the process of generating the route path, otherwise they are allowed. |
```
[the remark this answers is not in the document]
gknisely: I agree.  @maksimandrianov can you make the change log more descriptive too?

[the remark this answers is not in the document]
maksimandrianov: Yes, I made

docs/api/turn-by-turn/api-reference.md
```diff
@@ -122,6 +122,7 @@ The following options are available for `auto`, `bus`, `hov`, `taxi`, and `truck
 | :-------------------------- | :----------- |
 | `height` | The height of the vehicle (in meters). |
 | `width` | The width of the vehicle (in meters). |
+| `exclude_unpaved` | This value indicates the whether or not the path may include unpaved roads. If `exclude_unpaved` is set to 1, unpaved roads are not allowed to be used in the process of generating the route path, otherwise they are allowed. If `exclude_unpaved` is set to 1 it is allowed to start and end with unpaved roads, but is not allowed to have them in the middle of the route path. |
```
genadz, line 125: there is some contradiction between second and third sentences. @maksimandrianov Could you rewrite it ?

maksimandrianov: done

gknisely (approved):

PR #3245, opened 2021-08-04
Add osm ids for nodes refactor

Sleepingwell:
#2083 

This is the first part of addressing issue #2083. Here we add OSM ids for nodes only.

This still requires documentation, testing and potentially other changes. **Its primary purpose is to provide visibility of a set of proposed changes and is probably not ready to be merged as stands**.

Tests are not added here. **I was hoping that someone with more experience with this code base could provide advice on how to implement a test for a change like this** (hopefully that would save me time in going through the existing tests trying to figure out the most appropriate idiom/pattern).

There are two commits in the request.

- The first commit is rather messy and simply sets up the plumbing for the changes; it does not make any functional changes. There is a significant amount of plumbing and someone with more experience than I might be able to identify some which is not required, or if there is a cleaner way of doing this.

- The second commit makes the functional changes.

## Still to do

- Probably need to add another configuration setting to specify whether OSM ids should be written to the outputs. At present, there is a single configuration setting that determines whether they should be built into the tile.

- I have only included OSM ids in the output of *trace_attributes*. This is the only service I have used in Valhalla and hence the only path through the code base I am somewhat familiar with. Hopefully it is enough to only include OSM ids here at first and add other services later. I would need help with working through those other services due to my inexperience with this code base.

- tests, documentation, changelog, ...

## Notes to Reviewers

- I write the OSM ids into the tile 'at the end of the fixed size vectors'. I think this is OK, since the locations in the tile of the stuff that comes after this are specified by `offsets'... but I might be wrong, or you might prefer it at the end of the tile for some other reason.

- There appears to be inconsistency with naming methods on some classes; some are camel case and some are snake case. I have used snake case but I'm not sure if there is reason for these different conventions or if it is just an consequence organic evolution.

- Please check if you like my mixed use of assertions, warnings and runtime errors. I think these are are safe in release mode and will blow up appropriately in debug. In particular, **Please check the logic and comment in the block starting on line 330 of src/mjolnir/graphtilebuilder.cpp**.

## Tasklist

It is not clear whether you want ticks on the things still to do, or the tasks which are done. I assume the latter. As noted above, I don't anticipate that, at this stage, this is ready for merge. Will complete these tasks if the reviewers think this has legs.

 - [] Add tests
 - [] Add #fixes with the issue number that this PR addresses
 - [] Update the docs with any new request parameters or changes to behavior described
 - [] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

NA


src/baldr/graphtile.cc
```diff
@@ -317,6 +317,11 @@ void GraphTile::Initialize(const GraphId& graphid) {
   admins_ = reinterpret_cast<Admin*>(ptr);
   ptr += header_->admincount() * sizeof(Admin);
 
+  if (header_->has_osmids()) {
```
kevinkreiser, line 320: you mentioned this in the description but yeah i think this is a breaking change. take for example the scenario where you use this new code to make tiles and then you send them to a device that is running the old code. when it goes to load the new tiles and they have this extra data, then the pointer arithmetic here will set the edgebins pointer to where you are writing the osm ids. because of this you pretty much have to write this at the end of the tile

scripts/valhalla_build_config
```diff
@@ -36,6 +36,7 @@ config = {
     'include_bicycle': True,
     'include_pedestrian': True,
     'include_driving': True,
+    'include_osmids': False,
```
kevinkreiser, line 39: nit: we need to include the word "node" in here. we already include the osm way ids for all the ways but i think its not descriptive enough to just say "osmid"

src/mjolnir/elevationbuilder.cc
```diff
@@ -64,7 +65,7 @@ void add_elevation(const boost::property_tree::ptree& pt,
     lock.unlock();
 
     // Get the tile. Serialize the entire tile?
-    GraphTileBuilder tilebuilder(graphreader.tile_dir(), tile_id, true);
+    GraphTileBuilder tilebuilder(graphreader.tile_dir(), tile_id, true, include_osmids);
```
kevinkreiser, line 68: in all the cases like these where we are reloading an already existing tile, we shouldnt need to add parameters to tell if the ids are included. the graphtileheader that was written with the tile that we are reloading already knows that so we can remove this unneeded change. i think there are several more. the only place where we will need it is likely in `graphbuilder.cc` where we make a new blank one. though maybe in the hierarchy builder it wont be so straight forward. i didnt check it

src/mjolnir/graphbuilder.cc
```diff
@@ -983,6 +985,11 @@ void BuildTileSet(const std::string& ways_file,
         // Set drive on right flag
         graphtile.nodes().back().set_drive_on_right(dor);
 
+        // set the osmid for the node
+        if (include_osmids) {
+          graphtile.set_last_node_osmid(node.osmid_);
```
kevinkreiser, line 990: maybe its confusing, but the very act of adding ids to the vector in the tile builder object should be enough to signal that we are including node ids. then we dont need that boolean flag in the constructor. what do you think?

valhalla/baldr/graphtileheader.h
```diff
@@ -614,7 +630,7 @@ class GraphTileHeader {
   uint64_t nodecount_ : 21;             // Number of nodes
   uint64_t directededgecount_ : 21;     // Number of directed edges
   uint64_t predictedspeeds_count_ : 21; // Number of predictive speed records
-  uint64_t spare1_ : 1;
+  uint64_t has_osmids_ : 1;             // (bool) Whether the tile has OSM ids
```
kevinkreiser, line 633: same nitpick, we should be specific that its node ids, since we already have the way ids

valhalla/mjolnir/graphtilebuilder.h
```diff
@@ -256,6 +257,17 @@ class GraphTileBuilder : public baldr::GraphTile {
                        bool& added,
                        bool diff_names = false);
 
+  /**
+   * Set the OSM id for the most recently added node.
+   */
+  void set_last_node_osmid(const uint64_t id) {
```
kevinkreiser, line 263: why not something just like the the nodes and edges:

`std::vector<uint64_t>& osm_node_ids();`

kevinkreiser:
@Sleepingwell this is looking good thanks for all of the effort here. i left a few comments so far, let me know what you think

src/baldr/graphtile.cc
Sleepingwell: Everything below this data is indexed using an offset. The offsets in the new tiles (created using new code) all include 'the jump' over where the OSM ids are written. If access to the data beyond where the OSM ids is only ever done using the offsets (in the old code), then I think this will still work and is hence not a breaking change.

Apologies if I'm not understanding, but happy to move the data to the end of the tile in any case (just means that we have to access it via 'an offset', which complicates things).

scripts/valhalla_build_config
Sleepingwell: Agreed.

src/mjolnir/elevationbuilder.cc
Sleepingwell: Yeah, the only trick is that what if I want to build tiles that do not contain OSM ids 'on top of' tiles that do? Check the logic in the implementation of the constructor to see what I mean.

Yes... this is in *a lot* of places.

valhalla/baldr/graphtileheader.h
Sleepingwell: Agreed.

valhalla/mjolnir/graphtilebuilder.h
Sleepingwell: Happy to do that, especially given that it is consistent with nodes and edges. I guess this feels to me like it hides more implementation detail and the existence of a method with this name tells us something.

src/baldr/graphtile.cc
kevinkreiser: but have a look at the next snippet of code. you can see that `edge_bins_` is set to `ptr` which is always the first byte after the end of the admins. now after your change it might be the first byte after the node ids. so i agree with what your saying except that it isnt the case with the edge bins. concretely, someone who is running a version of valhalla before your code change has the following code:


```c++
...

  // Set a pointer to the admininstrative information list
  admins_ = reinterpret_cast<Admin*>(ptr);
  ptr += header_->admincount() * sizeof(Admin);

  // Set a pointer to the edge bin list
  edge_bins_ = reinterpret_cast<GraphId*>(ptr);

...
```

now imagine we merge your pr and then we create new tile data with it and enable osm node ids. then we give that person with old code, the new data. their old code will point the edge bins at the osm node ids. that is a breaking change. i think maybe you read in the graphtile header about the edgebin offsets and got confused. the edge bins are essentially an array within an array. so the "offsets" in the header arent with respect to the graphtile data, they are with respect to this `edge_bins_` pointer, which would be screwed up if we keep the code as you have it here.

in order to remove this burden of thought i think its always just easier to put new stuff on the end but technically you are right, there is a way to stay compatible and not put it on the end of the tile. its just not at this particular spot, it'd need to be after the edge bins which are also not fixed length and so at that point why not just put it on the end :smile: i mean i agree with you, itd be nice to keep all the fixed length stuff together but sadly the fools that designed it did not foresee this little snag :wink:

src/mjolnir/elevationbuilder.cc
kevinkreiser: i dont understand what you mean by "build on top of" existing tiles. we have two modes of graphtilebuilder essentially, a fresh one with nothing in it and one that initialized from an existing tile. in the former, if you add stuff to the node ids vector then we assume we will be storing node ids (unless, as you put checks in the code, someone doesnt put the right number of these to match the number of graph nodes). in the latter, i we just do whatever we see in the tile, if it had ids we deserialize them and if it didnt we have an empty vector of ids (and the header should be set to false).

i dont think there is a case where you start without osm node ids in the tile and then add them later. at the very least we dont support that as it can only happen between the parsing phases of the build and the graphbuilder phase of the build where we have those `OSMNode` still present on the disk to copy the ids out of.

i looked at the constructor but it didnt dawn on me what you were trying to point me at

src/baldr/graphtile.cc
Sleepingwell: Agreed... I had not paid attention enough to notice enough that this was not something like

```c++
complex_restriction_forward_ = tile_ptr + header_->complex_restriction_forward_offset();
```

like the rest of the ptr arithmetic.

Just out of interest, do you think this would make it possible to put it after the edge bins?

I'll move it to the end of the tile.

src/mjolnir/elevationbuilder.cc
Sleepingwell: It is probably me that is not understanding the building process. What I *think* is going on is that we build a set of tiles at some resolution, the refine (or is it coarsen?) them. All my point is, and I don't know why one might want to do it, is that someone might want to include OSM ids at the 'initial' resolution but not at the others, so they would not copy the OSM ids as they copy the nodes.

In the constructor of GraphTileBuilder, the builder only gets to have `has_osmids_` set to true if `include_osmids` is true *and* the tile from which it is initialised has OSM ids (probably should have just said that rather than saying 'go look at...', sorry).

This was just a 'what if' idea that reflects my inexperience and is clearly not a useful option. I think your proposition of inferring intent by just setting it ids is much cleaner.

valhalla/baldr/graphtileheader.h
```diff
@@ -687,6 +711,9 @@ class GraphTileHeader {
   // GraphTile data size in bytes
   uint32_t tile_size_;
 
+  // Offset to the beggining of the OSM ids for nodes data.
+  uint32_t osmids_for_nodes_offset_;
```
Sleepingwell, line 715: I guess one could get this from difference between the lane connectivity offset from the end of tile offset and save yourself a slot, but since you can only do this for some subset of offsets, that feels kinda hacky.

src/mjolnir/elevationbuilder.cc
kevinkreiser: while it is possible to do what you are saying, its quite strange and definitely not a normal case. so our whole tile build is configuration driven. the tile build does have phases and you can stop at one phase (process terminates gracefully) and then restart it at the next phase and continue. to do what you are saying one would have to run the tile building with the ids turned on. wait for it to stop. edit their config to turn it off, and continue the build. in practice i think there is very little use for it but you are right that it is a possible "feature". i think for the sanity of us though it'd be better off if we stick to what was in the config at the beginning bake it into the header from the start and keep on rolling with it.

Sleepingwell: Agreed... it is all gone now and *much* cleaner.

valhalla/baldr/graphtileheader.h
Sleepingwell: This is no longer required.

valhalla/mjolnir/graphtilebuilder.h
Sleepingwell: Changed.

src/mjolnir/graphbuilder.cc
Sleepingwell: Done.

src/mjolnir/elevationbuilder.cc
Sleepingwell: Constructor now untouched (simply adding OSM ids signals their presence).

scripts/valhalla_build_config
Sleepingwell: Changed to `include_osmids_for_nodes`.

src/baldr/graphtile.cc
Sleepingwell: Done.

valhalla/baldr/graphtileheader.h
```diff
@@ -232,6 +232,14 @@ class GraphTileHeader {
...
   }
 
+  /**
+   * Does the tile contain OSM ids?
+   * @return \c true if it does and \c false otherwise.
+   */
+  bool has_osmids_for_nodes() const {
+    return osmids_for_nodes_offset_ != 0;
+  }
+
```
Sleepingwell, line 242: Are you happy with using the offset like this? In `GraphTileBuilder`, if there are no OSM ids I set the offset to zero. In `GraphTile` I have an `assert(has_osmids_for_nodes())` before indexing the array of OSM ids.

- Should that be a runtime error rather than an assert?
- Might I be causing problem or simply not be being explicit enough by using an offset in a logical condition like this?

valhalla/baldr/graphtile.h
```diff
@@ -181,6 +181,41 @@ class GraphTile {
...
+    assert(has_osmids_for_nodes());
+    if (idx < header_->nodecount()) {
+      return osmids_for_nodes_[idx];
+    }
+    throw std::runtime_error(
+        std::string(__FILE__) + ":" + std::to_string(__LINE__) +
+        " GraphTile NodeInfo index out of bounds: " + std::to_string(header_->graphid().tileid()) +
+        "," + std::to_string(header_->graphid().level()) + "," + std::to_string(idx) +
+        " nodecount= " + std::to_string(header_->nodecount()));
+  }
```
kevinkreiser, line 209: i think this bit is too strong tbh. it assumes the caller knows how the data was made (with or without node ids) or that they conditionally call the method. i'd prefer to return a sentinal value that the caller can check to see if the node had or didnt have an id. i know its all or nothing since it doesnt support sparse data but the assertion is just too much

```suggestion
  uint64_t osmid_for_node(const size_t idx) const {
    if (idx > header_->nodecount()) {
      throw std::runtime_error(
        std::string(__FILE__) + ":" + std::to_string(__LINE__) +
        " GraphTile NodeInfo index out of bounds: " + std::to_string(header_->graphid().tileid()) +
        "," + std::to_string(header_->graphid().level()) + "," + std::to_string(idx) +
        " nodecount= " + std::to_string(header_->nodecount()));
    }
    return header_->has_osmids_for_nodes() ? osmids_for_nodes_[idx] : kInvalidNodeId;
  }
```

valhalla/baldr/graphtileheader.h
kevinkreiser: i think this works perfectly fine. i dont think we need an assertion though. if anything checking for validity should be when we set the pointer in graphtiles initialization, we could check that the number of ids matches what the number of nodes in the tile is `(end_of_tile - ids_offset)/sizeof(uint64_t) == header->nodecount()` or something like that

valhalla/baldr/graphtileheader.h
kevinkreiser: nah this is all good. if we ever run out of slots we can always get creative and add a secondary header or something haha. but since it wouldnt be at the beginning of the tile we'll have to call it a middler or something haha

kevinkreiser:
@Sleepingwell this basically looks done to me, save for some tiny little stuff. sorry i didnt look at it for so long. do you mind if i set it to ready for review, or are you waiting on something else yet?

valhalla/baldr/graphtileheader.h
Sleepingwell: Do we want to add that check for validity?

Sleepingwell:
@kevinkreiser No worries (for taking some time to look at it).

As far as whether this is ready to set to ready to review, I think there are a few things left to do:

- I should probably update the documentation on the outputs of *trace_attributes* to describe the change.
- Should I add a setting allowing the user to specify whether or not to include the OSM ids in the output of *trace_attributes*?
- Should I add a test(s) for this? If so, could you give me a hint on how (e.g. point me to an existing test that I might mimic the structure/pattern of)?

Let me know what you think.

kevinkreiser:
@Sleepingwell all 3 excellent points and definitely the next steps. let me see if i can describe what needs to be done:

1. updating the docs should be straight forward see: https://github.com/valhalla/valhalla/blob/master/docs/api/map-matching/api-reference.md#attribute-filters-trace_attributes-only, specifically it seem slike we should add a section `node.node_id` to match `edge.way_id`
2. this follows directly from the documentation actually, you should be able to find the code for AttributesController, it is what eventually sets a flag in the protobuf that stores the request options. you wont actually need to add an option for this directly, just add it in the attributes controller and whatever strings the users sends to us we store generically here: https://github.com/valhalla/valhalla/blob/master/src/worker.cc#L899. then later on the attributes controller checks if any of them are valid. you add a new one to the controller as mentioned above and then you need to add code to triplegbuilder to set those ids on the nodes in the leg. you'll need a place to store that, which you can add to the protobuf here: https://github.com/valhalla/valhalla/blob/master/proto/trip.proto#L266 do note that some nodes still wont have ids, as these are the nodes which we "invent" on the fly to match the starting and ending point of your route/mapmatch
3. for tests i would simply expand these: https://github.com/valhalla/valhalla/blob/master/test/gurka/test_trace_attributes.cc.  you'll need to be a bit tricky to add the request options i think... check around other gurka tests to see what the best way might be to set these options. i think the key difficulty is that its an array, i notice also that the `test_traffic.cc` file has some examples of this as well

Sleepingwell:
@kevinkreiser Have not had time to get to this sorry, might be another week before I do... apologies.

Nuren-Shams:
Following this. Currently need to pipe snapped pings through Nominatim's PostgreSQL backend to get OSM ID's through reverse geocoding.

kevinkreiser:
> Should we add the OSM id for the node at the start of an edge?

this isnt a bad idea but we would need some way to also denote that it may not actually apply to this exact location (lat lon) of the node. 

> Do you know if there are there instances in

1. in bss and in transit we add nodes to the graph, when we do this we need to take care to keep the ids that we have so far and give invalid ids to those that are added.
2. hierarchy just adds node transitions, but when we do that we should, in the new hiearachy, copy the ids along. basically what happens is we move nodes out of some tiles into new tiles. i forget if this pr already handles that but its somethign we must do
3. shortcuts just creates new edges, when we do that we do copy nodes to build the new tiles so  when that copy happens we have to take the ids too if they are there this may happen automatically in tilebuilder but i think it doesnt.
4. by `restrings` i assume you mean restrictions? i dont th ink there is anything to worry about there
5. elevation can also be safely ignored

devShahriar:
@Sleepingwell  @kevinkreiser so to retrieve the osm nodeid what field i need to add in my request payload ? is it ready yet ?

kevinkreiser:
@devShahriar as you can see we havent merged the pr so it depends what you mean by "ready". you can definitely checkout @Sleepingwell 's branch and try it yourself. read the code a bit to see how to use it.

kevinkreiser:
also @Sleepingwell it looks like you need to lint the code (format it) by running `scripts/format.sh`

devShahriar:
> @devShahriar as you can see we havent merged the pr so it depends what you mean by "ready". you can definitely checkout @Sleepingwell 's branch and try it yourself. read the code a bit to see how to use it.

Yeah definitely i will take a look .
And waiting for a merge :) . This nodeid will solve a lot problem for us .

Sleepingwell:
Sorry people, will get back to this soon. I was presuming we'd squash this rambling history before merging. I will be sure to lint before that.

kevinkreiser:
@Sleepingwell we only allow squash merges to the repo anyway so it will all become one commit in the end anyway. unless you really want to there is really no gain to rebasing ahead of time

Sleepingwell:
@kevinkreiser I think we're getting close. A few more points and questions:

- By default, this includes osmids in the outputs (if they exist in the tile). Is this as you want it? I am wondering because of your comment on the potential need for 'trickiness' when implementing the tests. Also, [in the `trace_attributes` documentation](https://valhalla.readthedocs.io/en/latest/api/map-matching/api-reference/#attribute-filters-trace_attributes-only), it mentions that _If no filters are used, all attributes are enabled and returned in the `trace_attributes` response._, but when I look at [the example for `trace_attributes`](https://valhalla.readthedocs.io/en/latest/api/map-matching/api-reference/#example-trace_attributes-requests), it appears one can specify that the intention is to include:
    ```json
    {
        ...,
        "filters":{
            "attributes":["edge.names","edge.id", "edge.weighted_grade","edge.speed"],
                "action":"include"
        }
    }
    ```

- Commit 17a78ed4e27f8cbe252522a3d05c30f2195eace9 includes changes not related to this request and which I had intended to submit by a later requests. It relates to missing documentation for another attribute. I can remove it if you like. Basically, the [documentation on what end\_node can contain](https://valhalla.readthedocs.io/en/latest/api/map-matching/api-reference/#end-node-items) appears to be out of date: my outputs contain `transition_time` but that is not in the docs.

- You might want to have a look at what I've done in [src/mjolnir/bssbuilder.cc](https://github.com/valhalla/valhalla/pull/3245/files?file-filters%5B%5D=#diff-f4f84d77c4efe536c0426eb2213b4cfd0b91d539f53911b99b4e1a321c365d2d) and, in particular, in [src/mjolnir/transitbuilder.cc](https://github.com/valhalla/valhalla/pull/3245/files?file-filters%5B%5D=.cc#diff-f4f84d77c4efe536c0426eb2213b4cfd0b91d539f53911b99b4e1a321c365d2d). I found it hard to get my head around what is going in these two files and I don't really know how to check that what I have done is working (I don't know what transit tiles are or where to get them from). In the latter case, what I've added faux osmids for all nodes in the transit tile.

- With respect to discussion in previous comments, I'm not sure how one might go about denoting that the OSM id for the node may not actually apply to the exact location corresponding the the start of the edge. Two possibilities are:
    1. Don't include the OSM id if it is not (i.e. avoid the potential inconsistency), and
    2. Add another field with a (boolean) flag.

    In either case, with my level of familiarity with the code, I think I'd determine whether it is at the start or not by comparing the lng/lat at the actual node with the lng/lat in the output. I imagine that somewhere in the matching code this is determined by other means. How would you proceed?

kevinkreiser:
i havent gotten a chance to look deeply at this yet, i just wanted to apologize for the wait

Sleepingwell:
@kevinkreiser: just a gentle reminder that this is here... would be good to get this out before I completely forget what I've done!

kevinkreiser:
Yep I totally agree. Let me put this on my schedule for Monday morning

xlqian:
Hello guys, first of all, Happy new year :)

We're very interested in this PR and we just want to have an idea when this PR is gonna be merged. Do you intend to merge that recently? 

Thanks

kevinkreiser:
Yep I think we just need to do a review and do some data builds with it

Sleepingwell:
No worries. I squashed onto master and resolved in what looked like the
natural way.

Sleepingwell:
@kevinkreiser Hi Kevin. Are you still interested in this commit. The project I was working on winds up at EOFY and I won't get time to think about this again on company coin, so I'd like to deal with it soon if we can.

kevinkreiser:
@Sleepingwell I am but the only thing that is bothering me about this is the next logical step and how we would support that, ie having all osm node ids stored in the dataset.

i have been noodling over an alternative approach which would piggy back off of the shape storage and store the IDs in an optional 3rd component of the shape data. we could indeed varint/offset encode them and we would probably be able to get away with only storing a single bit in the tile header that says whether or not this data was included in the tile. then when we make EdgeInfo objects, we simply pass this bit/flag along to that objects constructor and when its unpacking shape it will also need to unpack the node ids... this design has the drawback that indeed, the shape unpacking now has to always deal with a 3rd component so that definitely will cost more to get at the shape if you had ids in your data but never needed to use them. though i guess in that case, why the heck did you build the data with the ids in them haha

anyway im sorry this is taking so long and im waffling here but i'm wondering if the benefits to a simpler design as proposed above that make it more pallatable. specifically the benefits i can think of are:

1. you get all the ids if you need them (that will be the first complaint of this current approach in this PR)
2. the changes are quite a bit more streamlined, essentially just following around all of the shape functions during data building to care about a 3rd component
3. we'll get varint/offset encoding compression for free which could make the data not so big and maybe a no brainer to include?
4. we dont have to implement 2 different schemes to get data for graph nodes and for nodes along the shape of the edges, there's just one implementation

Sleepingwell:
One of the longer term goals of the project we were working on was to have a set of identifiers that would remain constant through time (i.e., between versions of the OSM data)... a feature which seems to have broad appeal and seems to have left _at least a few_ orphan implementations in its wake. These would clearly not be OSM ids, but for the prototype we chose a specific version of the OSM data so the OSM ids sufficed.

I always envisaged that we would eventually use some sort of matching of intersections based on characteristics such as location, number of connecting roads, ... , and potentially using the OSM ids as hints in the matching process (though I never thought it through in detail as the number of orphans suggested that one should not to take the task lightly). I thought one might use the geohash as the id. I can't remember if OSM ids occur at 'non-intersections' (which _might_ complicate the matching), but we only cared about intersections.

I never formed an opinion on whether those ids lived alongside the OSM ids or replaced them, but we would probably not use the OSM ids if they existed, though one can imagine that a (potentially temporal) correspondence might have uses.

Would that future goal affect the architecture?

kevinkreiser:
i dont think so! so long as we could fit such an id into a 64bit wide integer then we'd be fine. actually with varint we could technically support wider integers. then the code could either use the id that came with the osm node or call into some other code to generate an id following whatever algorithm we felt like implementing. if however, you are only ever interested in ids for intersections in the graph (graph nodes) then what i proposed above would be more data than you would strictly need. there are ways to workaround that though. we could have like 2 bits in the tileheader dedicated to telling what kind of node id data we had: `{none:0, full: 1, sparse: 2}`. finally if we wanted to have our own special schema for assigning ids we could codify that as well with another couple bits for an enum. the question would mainly be about how to represent the sparse data vs the full data, in other words an id stored at just the graph nodes vs stored at every shape point in the graph not just those where two edges intersect. the good news is that if we know the data is sparse by virtue of the aforementioned enum we know how many 3rd components to decode from the shape string to get the ids. in the case that its sparse, there is only ever 2 and everythjing else is lat lon information, in the case of full it would be interleaved and in the case of none it would just be 2 columns of encoded data for lat and lon.

now im wondering if maybe we should also split out the columns instead of interleaving them. i realize this requires the reader to completely understand how the polyline encoding algorithm works but basically the data looks like:

    varint_lat_0_bytes, varint_lon_0_bytes, varint_lat_1_bytes, varint_lon_1_bytes, ... , varint_lat_n_bytes, varint_lon_n_bytes
    
for full id support im suggesting the encoded data would look like:

    varint_lat_0_bytes, varint_lon_0_bytes, varint_id_0_bytes, varint_lat_1_bytes, varint_lon_1_bytes, varint_id_0_bytes, ... , varint_lat_n_bytes, varint_lon_n_bytes, varint_id_0_bytes
    
and for sparse it would be just 2 ids total, like:

    varint_lat_0_bytes, varint_lon_0_bytes, varint_id_0_bytes, varint_lat_1_bytes, varint_lon_1_bytes, varint_id_0_bytes, ... , varint_lat_n_bytes, varint_lon_n_bytes
    
perhaps we could just stick the ids at the end of the encoded string though... this would mean that regular shape decoding would be unaffected but that getting out just the ids would be annoying because you'd need to also get to the end of the shape. oh and actually that brings up a great point... if we interleaved them as above it would be a non backwards compatible change... so that idea is out actually. let me think a bit about how we could attach them to the shape though.. i feel like having a configurable way to bind them to the shape would be the best option becauase we wouldnt need an additional offset into a variable sized structure

src/mjolnir/graphtilebuilder.cc
```diff
@@ -362,9 +368,24 @@ void GraphTileBuilder::StoreTileData() {
...
-    // Set the end offset
-    header_builder_.set_end_offset(header_builder_.lane_connectivity_offset() +
-                                   (lane_connectivity_builder_.size() * sizeof(LaneConnectivity)));
+    if (has_osmids_for_nodes()) {
+      if (osmids_for_nodes_.size() != nodes_builder_.size()) {
+        throw std::runtime_error(std::string("length of osmids does not equal length of nodes: ") +
+                                 std::to_string(osmids_for_nodes_.size()) +
+                                 " != " + std::to_string(nodes_builder_.size()));
+      }
+      header_builder_.set_osmids_for_nodes_offset(header_builder_.lane_connectivity_offset() +
```
tastyeggs, line 377: We're missing one line from line 1129 below:

```
header.set_osmids_for_nodes_offset(header.osmids_for_nodes_offset() + shift);
```

Sleepingwell: Thank you!

jverstichel:
Hi, I am new to the Valhalla project, which looks awesome btw, and I am also quite interested in this functionality. What is still needed to complete this pull request?

tastyeggs:
@jverstichel if it helps, my team has been running this branch in production for a few months, and it is quite stable.

kevinkreiser:
yeah im sorry we still didnt get to this. i would just hate to get this ready to merge and then have to do another project to support all node ids (even those that arent on graph nodes).

does anyone have any design ideas of how we could support both modes of node id fidelity? or perhaps supporting all node ids is good enough for those that only need them for the graph nodes?

one option that springs to mind is:

1. store a custom tag in edgeinfo (we have support for this now) which is simple an index into an array of nodeids that we add to the tile
2. the array can either have full node ids for all shape in the graph or just pairs (begin and end graph nodes osmids) this can be configurable in mjolnir
3. the custom tag (its an enum) in edgeinfo can be used to differentiate between full ids or graph node ids or no ids (if the tag doesnt exist)
4. we can then add helper functions to edgeinfo to return the node ids for a given edge
5. bonus points, we could even layout the nodeides in the array so that adjacent edges that share nodes simply overlap by their index and we dont actually store the nodeid twice, shortcuts are a pain in the butt but if we wanted to those wouldnt even have to duplicate the ids

long story short, im sorry that we've been blocking on support for this for such a long time. it wasnt the intent. does anyone have any opinions about the idea above?

@Sleepingwell once again many apologies!

jverstichel:
@kevinkreiser thank you for the swift reply! As I am still busy getting familiar with the project, I do not have real design opinions yet.

For us, the main idea is that we would like to have the osmid of the start and end nodes for all edges, including those for the first edge (first node osmid is covered in the PR's implementation even if you start halfway the edge) and the last edge (last node osmid is not covered in the PR's implementation if you end halfway the edge).

Storing the desired information in the EdgeInfo does sound sensible for that. But perhaps the desired behavior is also possible with a small change to the current PR code?

Other than that we may consider using 'start_node_id' instead of 'node_id' on the edge in the serialization?

jverstichel:
@kevinkreiser small update; I've made some code changes to also include the end node of the last edge in the response. I've also brought the code changes up-to-date with the latest code on master (it was quite a bit behind).
This was of course on a branch I made on my local repository.

What is the best way forward? Do I create a new pull request for the updated code (as I do not own the branch this pull request is on)? Is it ok if I push to a branch on this repo, or do you prefer I request the PR from a fork? Or should I wait with the PR until we are clear on the best path forward (e.g. using the new EdgeInfo possibilities)?

kevinkreiser:
you could either:

1. pr it to @Sleepingwell 's branch so that this pr gets updated when he merges it in OR
2. open a new pr marking this pr as superseded

i would love to have time to take a crack at the suggesation i made above but its  increasingly difficult to find the time.

to me we could use most of the work you all have already done, just change around a bit how it gets referenced (ie from edgeinfo rather than as a parallel array to graph nodes) so maybe it wont be   all that much of a refactor...

Sleepingwell:
@jverstichel I am unlikely to be able to contribute further to this. The project where we were using Valhalla finished and I have not looked at this for a very long time so getting back up to speed would take time I don't currently have. Another option for (in addition to those given by Kevin above) could be to make you a collaborator on [CVTS/valhalla](https://github.com/CVTS/valhalla) so you can keep this pull request alive. Let me know if you want that.

In our work, we did require all the OSM ids and hacked them in, though in a way that broke backward binary compatibility, so it was never going to be a pull request here, though I had intended to make a follow up pull request with a reworked version once I had cut me teeth on the current one. You can find the changes in the last 3 commits of [this branch](https://github.com/CVTS/valhalla/tree/edge-and-node-osmids). That is a couple of years out of date and is likely to be quite conflicted, but *may* contain useful tidbits or context.

Good Luck!

jverstichel:
@Sleepingwell my apologies for the very delayed reply, I got buried under other work. If you could make me a collaborator on the CVTS/valhalla, I should be able to get the PR up-to-date with the current version and also apply the changes I made.

Many thanks in advance!

Sleepingwell:
Done.

Cheers,
Simon

jverstichel:
@kevinkreiser I've updated the PR so now it is up-to-date with master and the osm id's of both the start and end nodes are available. Is this PR now ready for review? Or are there other steps that need to be taken first?

VlaamsVerkeerscentrum:
We are using this branch in our application (developed by @jverstichel ) for a few years and would like to get this into the master. 
Our new developer has reworked our application to be up to date with the latest version of the main branch and we would like to know the next steps necessary to get this request reviewed?

kevinkreiser:
@VlaamsVerkeerscentrum he can open a new pr and talk about it there but the main thing i would like to know is whether we are storing all the nodes (even the ones that arent intersections in the graph) or are we just storing the topological nodes (graph nodes)? i think we should do the former and i think the data has to be optional. if we meet those two bars, then im sold. my detailed opinion is fairly well intact on this older comment: https://github.com/valhalla/valhalla/pull/3245#issuecomment-1452230047

l-brabant:
@kevinkreiser I have created a new PR (including a merge with master as it was behind). It includes osmid of the start and end nodes for all edges (optional, by setting flag bool include_osmids_for_nodes to true (default false). Please let me know if further changes are required!

kevinkreiser:
@l-brabant sorry maybe i wasnt very clear, im advocating that we store also the osm nodes that arent graph nodes. since thats a bit more work than is done here perhaps i can imagine a way that we can support the internal osm nodes as well and we can just put it as a TODO for the future. specifically i guess we could add to the edge info those nodes which arent at the ends of a given edge and fulfill this requirement. if others agree then i think we can proceed.

kinkard:
> i guess we could add to the edge info those nodes which arent at the ends of a given edge and fulfill this requirement

Then it would probably be easier to store OSM Node IDs *only* in the edge info to have a single source of this information

kevinkreiser:
@kinkard yeah i was saying that back in https://github.com/valhalla/valhalla/pull/3245#issuecomment-1452230047 but ive since softened on it a bit simply because i dont have the time to do the implementation. for me its hard to tell people: no we should do it this way, but then i never have time to actually show what i mean and do it.

PR #3251, opened 2021-08-05
Added stop and give_way/yield signs to the data and traffic signal fixes.

gknisely:
# Issue

This PR is the initial logic that is needed so that we can call out stop and give_way/yield signs in the narrative.  For example, "Turn right at the stop sign to continue on xyz Road." 

I tried to use the logic that already existed for traffic signals; however, I found a bug with how they are marked.  Therefore, I fixed the issue and added tests for traffic signals.

## Tasklist

 - [x] Add tests

## Requirements / Relations


valhalla/baldr/directededge.h
```diff
@@ -1118,10 +1152,12 @@ class DirectedEdge {
   uint64_t tunnel_ : 1;         // Is this edge part of a tunnel
   uint64_t bridge_ : 1;         // Is this edge part of a bridge?
   uint64_t traffic_signal_ : 1; // Traffic signal at end of the directed edge
+  uint64_t stop_sign_ : 1;      // Stop sign at end of the directed edge
+  uint64_t yield_sign_ : 1;     // Yield/give way sign at end of the directed edge
```
kevinkreiser, line 1156: these have to go at the end where the spare was. by putting them here you break backwards compatibility. old code loading new tiles would get the value of stop sign for seasonal and the value of yield sign for deadend and bss connection would be seasonal. we cant shift the values down inside the 64bit integer

kevinkreiser (changes requested):
i didnt thoroughly review the rest but you have to at least make the change in directededge bitfields

valhalla/baldr/directededge.h
gknisely: @kevinkreiser my bad.  I will fix

gknisely: @kevinkreiser updated.

dnesbitt61 (approved):
I don't understand the whole "minor" logic but trust you tested all of this thoroughly

gknisely:
> I don't understand the whole "minor" logic but trust you tested all of this thoroughly

@dnesbitt61 Let me explain.  If there is a k/v pair `stop=minor` or `give_way=minor` we are supposed to assign the stop sign or give_way to only the edges that have a lower road class at the intersection.  If minor does not exist, then we assign to all intersecting edges.  

Yes tested.  Even created tests for traffic_signals too as they were missing.  Next step will be stop impact updates an guidance changes.

PR #3253, opened 2021-08-08
Add RelaxHierarchyLimits with pre-defined algorithm-based factors

mloskot:
# Issue

Single place of definition of the hard-coded factors which apparently
are re-used in at least two locations.

This also fixes a inconsistency or bug/regression that was introduced around  two commits
- https://github.com/valhalla/valhalla/commit/1b150f7153a41de673acf88a5c22ac7709e91c1a
- https://github.com/valhalla/valhalla/commit/9b646089cf6db833ee712e22e988885ac3930487
which pass a boolean to RelaxHierarchyLimits,  instead of float-point factor.
which was later cultivated unnoticed in PR https://github.com/valhalla/valhalla/pull/2719

Maintaining the factors in single place should help prevent such situations.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

- https://github.com/valhalla/valhalla/pull/2719


src/valhalla_run_route.cc
```diff
@@ -154,8 +154,7 @@ const valhalla::TripLeg* PathTest(GraphReader& reader,
       LOG_INFO("Try again with relaxed hierarchy limits");
       cost->set_pass(1);
       pathalgorithm->Clear();
-      const float expansion_within_factor = (using_astar) ? 4.0f : 2.0f;
```
kevinkreiser, line 157: keeping this binary up to date with the server has been a constant point of failure for the project. i have always advocated for stripping this binary down to just calling the same bits of code that the server does rather than having its own slightly different implementation.

i am certain that, with enough care, we can get there but in the mean time thank you for cleaning this up

src/sif/dynamiccost.cc
```diff
@@ -260,6 +260,13 @@ void DynamicCost::RelaxHierarchyLimits(const float factor, const float expansion
   }
 }
 
+void DynamicCost::RelaxHierarchyLimits(const bool using_bidirectional) {
+    // since bidirectional A* does about half the expansion we can do half the relaxation here
+    const float relax_factor = using_bidirectional ? 8.f : 16.f;
+    const float expansion_within_factor = using_bidirectional ? 4.0f : 2.0f;
```
kevinkreiser, line 266: this one seems backwards. shouldnt it be:

`const float expansion_within_factor = using_bidirectional ? 2.0f : 4.0f;`

src/sif/dynamiccost.cc
```diff
@@ -260,6 +260,13 @@ void DynamicCost::RelaxHierarchyLimits(const float factor, const float expansion
   }
 }
 
+void DynamicCost::RelaxHierarchyLimits(const bool using_bidirectional) {
```
kevinkreiser, line 263: maybe we should make the other version of this method with the two floats, private, so that we dont make this mistake again, though im not sure there arent other places that call this

src/sif/dynamiccost.cc
```diff
@@ -260,6 +260,13 @@ void DynamicCost::RelaxHierarchyLimits(const float factor, const float expansion
...
+  // since bidirectional A* does about half the expansion we can do half the relaxation here
+  const float relax_factor = using_bidirectional ? 8.f : 16.f;
+  const float expansion_within_factor = using_bidirectional ? 4.0f : 2.0f;
```
kevinkreiser, line 266: the linting outdated my p revious comment but this ternary is backwards, it should be a smaller value for bidir and a larger for unidirectional searches

src/sif/dynamiccost.cc
mloskot: Thanks for catching this!
I must have dumbly copied it from the other place 🤦 
https://github.com/valhalla/valhalla/blob/a2fa483893a6fd9fa129a6353ae21ab5de9f3112/src/valhalla_run_route.cc#L157

It's been fixed now

src/sif/dynamiccost.cc
mloskot: Yes, thanks for catching this.
I must have dumbly copied it from the other place where the condition is a bit different 🤦 
https://github.com/valhalla/valhalla/blob/a2fa483893a6fd9fa129a6353ae21ab5de9f3112/src/valhalla_run_route.cc#L157

It's been fixed now.

src/sif/dynamiccost.cc
mloskot: or just remove it completely keeping single:
```cpp

void DynamicCost::RelaxHierarchyLimits(const bool using_bidirectional) {
  // since bidirectional A* does about half the expansion we can do half the relaxation here
  const float relax_factor = using_bidirectional ? 8.f : 16.f;
  const float expansion_within_factor = using_bidirectional ? 2.0f : 4.0f;
  
  for (auto& hierarchy : hierarchy_limits_) {
    hierarchy.Relax(factor, expansion_within_factor);
  }
}
```

What you say?

src/valhalla_run_route.cc
mloskot: Well, with the advent of `valhalla_service.cc`, we could simply remove the `valhalla_run_route.cc` completely. Unless there are strong reasons to not to.

src/sif/dynamiccost.cc
kevinkreiser: ship it

src/valhalla_run_route.cc
kevinkreiser: that is what i would like to do too but... there are currently some uses of `valhalla_run_route` in the scripting that we use to do regression testing. if we can port the formatting and functionality out of this binary to satisfy those use cases or change those downstream processes to use the other tools then we could do this, but today is sadly not yet that day :smile:

src/sif/dynamiccost.cc
kevinkreiser: i like your suggestion of combining the two into one with just the single boolean parameter. ill wait for you to make that change before putting a ship on it

src/valhalla_run_route.cc
mloskot: I see. I wish I had more time, but this year is hectic here too :-)

src/sif/dynamiccost.cc
mloskot: I've just pushed the updating commit

test/astar.cc
```diff
@@ -1639,7 +1639,7 @@ TEST(BiDiAstar, test_recost_path) {
   vt::BidirectionalAStar astar;
 
   // hack hierarchy limits to allow to go through the shortcut
-  mode_costing[int(travel_mode)]->RelaxHierarchyLimits(0.f, 0.f);
+  mode_costing[int(travel_mode)]->RelaxHierarchyLimits(true);
```
mloskot, line 1642: @kevinkreiser I'm not sure how to cope with this hack.

mloskot: The change to the hack above leads to the test failure 
https://app.circleci.com/pipelines/github/valhalla/valhalla/7514/workflows/08681a6b-9258-4295-b132-2a7439addc8a/jobs/46371

```
021/08/09 09:21:15.482272 [TRACE] /root/project/src/thor/bidirectional_astar.cc:1012: CONNECTIONS FOUND 2
[[0.031441,0.000000],[0.085340,0.000000]],
[[0.022458,-0.008983],[0.031441,0.000000]],
2021/08/09 09:21:15.482290 [DEBUG] path_cost::898
2021/08/09 09:21:15.482303 [DEBUG] FormPath path_iterations::6,8
/root/project/test/astar.cc:1653: Failure
  path.size()
    Which is: 5
  6
Final path has wrong number of edges
[  FAILED  ] BiDiAstar.test_recost_path (1354 ms)
[----------] 1 test from BiDiAstar (1354 ms total)

[==========] 36 tests from 3 test suites ran. (4212 ms total)
[  PASSED  ] 35 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] BiDiAstar.test_recost_path

 1 FAILED TEST
```

kevinkreiser: it doesnt look easy unfortunately. seems like you'd have to clone auto costing into this compilation unit, make its hierarchy limits accessible, and then register its create function with the costfactory. then you'd have to cast it to its cloned class name and call the accessible hierarchy limits function you added. its doable, but a giant mess tbh. what makes it a pain is that the costing classes are all pimpls... if we could include the source file with the definition of autocost into this test as a header (ie if it were a private header) then you could just derive from autocost, and add a new function called hackHierarchyLimits, register that costing type and move on... i think

let me know if that helps in anyway, in the meantime ill give it some more thought. sorryy for the annoyance

mloskot: > if we could include the source file with the definition of autocost into this test as a header (ie if it were a private header) then you could just derive from autocost, and add a new function called hackHierarchyLimits, register that costing type and move on...

Actually, I've tried something like that and it works.

Then, I've realised the hierarchy limits can be read-write accessed too, so I'm now trying simpler solution, see https://github.com/valhalla/valhalla/pull/3253/commits/f136f226070218738925a95700f05843de07ae43

test/astar.cc
```diff
@@ -1639,7 +1639,13 @@ TEST(BiDiAstar, test_recost_path) {
   vt::BidirectionalAStar astar;
 
   // hack hierarchy limits to allow to go through the shortcut
-  mode_costing[int(travel_mode)]->RelaxHierarchyLimits(0.f, 0.f);
+  {
+    auto& hierarchy_limits =
+        mode_costing[int(travel_mode)]->GetHierarchyLimits(); // access mutable limits
```
kevinkreiser, line 1644: ha i didnt even notice they were returned by reference! good eyes!

mloskot:
@kevinkreiser Shall I be concerned about [this failure of lint-build-debug](https://app.circleci.com/pipelines/github/valhalla/valhalla/7532/workflows/74ce933e-c801-4875-b733-6dbf3bd3c1a4/jobs/46445)?

```
Suppressed 236 warnings (236 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
diff --git a/src/valhalla_run_route.cc b/src/valhalla_run_route.cc
index 9f7503f97..334b0e286 100644
--- a/src/valhalla_run_route.cc
+++ b/src/valhalla_run_route.cc
@@ -131,7 +131,6 @@ const valhalla::TripLeg* PathTest(GraphReader& reader,
                                   PathStatistics& data,
                                   bool multi_run,
                                   uint32_t iterations,
-                                  bool using_astar,
                                   bool using_bd,
                                   bool match_test,
                                   const std::string& routetype,
@@ -716,7 +715,7 @@ int main(int argc, char* argv[]) {
     const valhalla::TripLeg* trip_path = nullptr;
     try {
       trip_path = PathTest(reader, origin, dest, pathalgorithm, mode_costing, mode, data, multi_run,
-                           iterations, using_astar, using_bd, match_test, routetype, request);
+                           iterations, using_bd, match_test, routetype, request);
     } catch (std::runtime_error& rte) { LOG_ERROR("trip_path not found"); }
 
     // If successful get directions
Tidy introduced changes
```

kevinkreiser:
@mloskot  you should indeed make the changes that it proposes. its basically saying, you dont make use of this parameter at all so why not remove it completely from the function

mloskot:
@kevinkreiser 
>  even if you didnt modify that part of the code clang tidy will trigger on it.

I see. It makes sense now. Fixed in https://github.com/valhalla/valhalla/pull/3253/commits/d854cce9f5cdcb6d400287d1083e8963a713bb2f

kevinkreiser (approved):

Issue #2884, opened 2021-02-23
Interrupt bidirectional search as soon as we found all alternates


genadz:
According to the comment https://github.com/valhalla/valhalla/pull/2868#issuecomment-781402924 we could save several percent of iterations if we stopped bidirectional search as soon as all alternates were found. In order to do that, we should refactor the code as proposed here https://github.com/valhalla/valhalla/pull/2868#issuecomment-783018463.
So, this refactoring, for example, could allow us to increase [`cost_extend_rel`](https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L28) coefficient in order to find more alternates.
@kevinkreiser

genadz:
actually, @kevinkreiser @dnesbitt61 do you know why we use `sortcost` instead of `route cost` when set cost threshold in bidirectional astar https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L792 ? is it only about performance ?

I tested it locally (replaced `sortcost` with `route cost`) and got ~20% performance decrease. It's a lot, but, to be honest, I expected much more decrease. The same time, current heuristic can miss optimal routes that lie far from the straight line `Source - Target`.

dnesbitt61:
I haven't followed logic for alternates. However, expansion is based on lowest sort cost, not real cost so I would think that it is proper to use sort cost. As long as the A* heuristic underestimates the true cost from any source to the target then this should be valid.

kevinkreiser:
> current heuristic can miss optimal routes

what do you mean when you say optimal here? i guess you mean the final real cost of a route. as dave said so long as we pick an admissible (underestimates the cost to the goal) heuristic what you are saying shouldnt happen. however there are two details there that can cause this heuristic to OVERestimate the cost to the goal.

1. we use a max speed and b-line distance to estimate the cheapest cost you possible reach the goal at. that max speed is possible to be less than the speeds reported by historical or real time traffic. i forget if we cap the speeds coming out of that in the end so that we dont overestimate the a* heuristic
2. if we have cost factors that have the affect of making parts of the graph cheaper in cost units that the a* heuristic is unaware of then that heuristic may still be overestimating the cost
3. something i didnt think of yet?

If anything with costing or heuristic estimation is happing to make the heuristic overestimate the cost to the goal wrt to the actual cost then what you are seeing could happen.

genadz:
[...]

Issue #2979, opened 2021-04-02
Long shortcuts may lead to long detours


genadz:
I bumped into the problem when I was working on another issue https://github.com/valhalla/valhalla/issues/2976 . I used osm data of Pennsylvania region : https://download.geofabrik.de/north-america/us/pennsylvania.html .

There are some examples of bad routes:
```
{"locations":[{"lon":-79.38572171614598,"lat":39.88228625865044},{"lon":-77.38753900842603,"lat":39.86557883977756}],"costing":"auto"}
```
![Screenshot from 2021-04-02 16-32-29](https://user-images.githubusercontent.com/9436623/113419952-21d53180-93d1-11eb-93bf-97c5d6c2baf7.png)

```
{"locations":[{"lon":-80.34434464274464,"lat":41.24558383530254},{"lon":-77.20773916063406,"lat":41.59117578530729}],"costing":"auto"}
```
![Screenshot from 2021-04-02 16-34-19](https://user-images.githubusercontent.com/9436623/113420116-64970980-93d1-11eb-83f1-639f1060f97b.png)

I debugged them and found the reason: in the place where long detour is happening forward/reverse search uses a shortcut edge that is very long. The same time the opposing search comes to a some edge that is superseded by the shortcut. As far as we don't use superseded edges we don't know if we're on a shortcut from the opposing search or no. Then the search traverse superseded edges , reach one of the shortcut ends but can't reach another one because of static threshold https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L25 .

Similar problem has been already described in the comment https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L21-L24 . The obvious fix (use max_edge_cost instead of static threshold) causes significant performance drop ( > **15%** on the whole planet).

PR #3257, opened 2021-08-10
Reach-based pruning for bidirectional astar

genadz:
# Issue

This PR fixes the "bug" described here https://github.com/valhalla/valhalla/issues/2928 (incorrect stop criterion in the bidirectional astar algorithm). 

There are following changes were made.
- Fixed **cost threshold** in bidirectional astar: use full route cost instead of sortcost.
- Implemented **reach-based pruning** technique described in the article from the comment https://github.com/valhalla/valhalla/issues/2884#issuecomment-785277797 . The main idea is to estimate lower bound cost for the path that goes through the current edge using min sortcost in the queue of the opposing search.
- Added logic that **exhausts hierarchy limits** more or less **simultaneously** for both directions (if we reached the limit in the forward/reverse direction - wait until the opposing direction also will expand this hierarchy level). The main intention of this logic is to provide (almost) valid conditions for the reach-based pruning.
- "**Synchronize**" shortcuts usage. Apply the same behavior for the same shortcut in both directions, i.e., if the shortcut was skipped by the forward/reverse search - it will be also skipped by the opposing search. And, if the shortcut was used by the forward/reverse search - it will be also used by the opposing search. This logic also helps to avoid false positive prunings.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -156,20 +156,46 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
                                             uint32_t& shortcuts,
                                             const graph_tile_ptr& tile,
                                             const baldr::TimeInfo& time_info) {
+  // Skip if this is a regular edge superseded by a shortcut.
+  if (shortcuts & meta.edge->superseded()) {
+    return false;
+  }
+
+  graph_tile_ptr t2 = nullptr;
+  baldr::GraphId opp_edge_id;
+  const auto get_opp_edge_data = [&t2, &opp_edge_id, &graphreader, &meta, &tile]() {
+    // Get end node tile, opposing edge Id, and opposing directed edge.
+    t2 = meta.edge->leaves_tile() ? graphreader.GetGraphTile(meta.edge->endnode()) : tile;
+    if (t2 == nullptr) {
+      return false;
+    }
+
+    opp_edge_id = t2->GetOpposingEdgeId(meta.edge);
+    return true;
+  };
+
   constexpr bool FORWARD = expansion_direction == ExpansionType::forward;
   auto& hierarchy_limits = FORWARD ? hierarchy_limits_forward_ : hierarchy_limits_reverse_;
   // Skip shortcut edges until we have stopped expanding on the next level. Use regular
   // edges while still expanding on the next level since we can still transition down to
-  // that level. If using a shortcut, set the shortcuts mask. Skip if this is a regular
-  // edge superseded by a shortcut.
+  // that level. If using a shortcut, set the shortcuts mask.
   if (meta.edge->is_shortcut()) {
-    if (hierarchy_limits[meta.edge_id.level() + 1].StopExpanding(pred.distance())) {
+    if (!get_opp_edge_data())
+      return false;
+
+    const auto& opp_edgestatus = FORWARD ? edgestatus_reverse_ : edgestatus_forward_;
+    const auto opp_edge_set = opp_edgestatus.Get(opp_edge_id).set();
+    // Synchronize shortcuts for both directions. If this shortcut has been already
+    // encountered on the opposing search we should do the same now: skip or traverse.
+    if ((opp_edge_set != EdgeSet::kSkipped &&
+         hierarchy_limits[meta.edge_id.level() + 1].StopExpanding(pred.distance())) ||
+        opp_edge_set == EdgeSet::kPermanent || opp_edge_set == EdgeSet::kTemporary) {
       shortcuts |= meta.edge->shortcut();
     } else {
+      // Mark this edge as "skipped".
+      *meta.edge_status = {EdgeSet::kSkipped, 0};
       return false;
     }
-  } else if (shortcuts & meta.edge->superseded()) {
-    return false;
   }
 
   // Skip this edge if edge is permanently labeled (best path already found
@@ -179,8 +205,6 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
     return true; // This is an edge we _could_ have expanded, so return true
   }
 
-  graph_tile_ptr t2 = nullptr;
-  baldr::GraphId opp_edge_id;
   const baldr::DirectedEdge* opp_edge = nullptr;
 
   if (!FORWARD) {
@@ -191,13 +215,10 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
       return false;
     }
 
-    // Get end node tile, opposing edge Id, and opposing directed edge.
-    t2 = meta.edge->leaves_tile() ? graphreader.GetGraphTile(meta.edge->endnode()) : tile;
-    if (t2 == nullptr) {
+    if (t2 == nullptr && !get_opp_edge_data()) {
       return false;
     }
 
-    opp_edge_id = t2->GetOpposingEdgeId(meta.edge);
     opp_edge = t2->directededge(opp_edge_id);
   }
 
@@ -262,13 +283,8 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
   }
 
   // Get end node tile (skip if tile is not found) and opposing edge Id
-  if (FORWARD) {
-    t2 = meta.edge->leaves_tile() ? graphreader.GetGraphTile(meta.edge->endnode()) : tile;
-    if (t2 == nullptr) {
-      return false;
-    }
-    opp_edge_id = t2->GetOpposingEdgeId(meta.edge);
-  }
+  if (t2 == nullptr && !get_opp_edge_data())
+    return false;
 
   // Find the sort cost (with A* heuristic) using the lat,lng at the
   // end node of the directed edge.
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -617,11 +617,31 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
       return {};
     }
 
+    // Exhaust hierarchy limits simultaneously in both directions. As soon as forward/reverse
+    // search exhausts limits on the particular level - it should stop and wait until reverse/forward
+    // search exhausts limits on the same hierarchy level. This logic ensures local optimality near
+    // the origin and destination and provides valid conditions for the reach-based pruning.
+    bool force_forward = false;
+    bool force_reverse = false;
+    assert(hierarchy_limits_forward_.size() == hierarchy_limits_forward_.size());
+    for (size_t level = TileHierarchy::levels().size() - 1; level > 0; --level) {
+      if (hierarchy_limits_reverse_[level].StopExpanding(rev_pred.distance()) &&
+          !hierarchy_limits_forward_[level].StopExpanding(fwd_pred.distance())) {
+        force_forward = true;
+        break;
+      } else if (hierarchy_limits_forward_[level].StopExpanding(fwd_pred.distance()) &&
+                 !hierarchy_limits_reverse_[level].StopExpanding(rev_pred.distance())) {
+        force_reverse = true;
+        break;
+      }
+    }
+
     // Expand from the search direction with lower sort cost
     // Note: If one direction is exhausted, we force search in the remaining
     // direction
     if (!forward_exhausted &&
-        ((fwd_pred.sortcost() + cost_diff_) < rev_pred.sortcost() || reverse_exhausted)) {
+        ((!force_reverse && (fwd_pred.sortcost() + cost_diff_) < rev_pred.sortcost()) ||
+         force_forward || reverse_exhausted)) {
       // Expand forward - set to get next edge from forward adj. list on the next pass
       expand_forward = true;
       expand_reverse = false;
@@ -636,6 +656,24 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
       if ((fwd_pred.not_thru() && fwd_pred.not_thru_pruning()) ||
           hierarchy_limits_forward_[fwd_pred.endnode().level()].StopExpanding(fwd_pred.distance())) {
         continue;
+      }
+
+      // Check if this branch can be pruned. It's implementation of the reach-based pruning technique
+      // for bidirectional astar: https://repub.eur.nl/pub/16100/ei2009-10.pdf .
+      if (cost_threshold_ != std::numeric_limits<float>::max() &&
+          fwd_pred.predecessor() != kInvalidLabel) {
+        const auto& fwd_pred_pred = edgelabels_forward_[fwd_pred.predecessor()];
+        const auto pred_tile = graphreader.GetGraphTile(fwd_pred_pred.endnode());
+        if (pred_tile != nullptr) {
+          // Estimate lower bound cost for the shortest path that goes through the current edge.
+          float route_lower_bound =
+              fwd_pred_pred.cost().cost + fwd_pred.transition_cost().cost + rev_pred.sortcost() -
+              astarheuristic_reverse_.Get(pred_tile->get_node_ll(fwd_pred_pred.endnode()));
+          // Prune this edge if estimated lower bound cost exceeds the cost threshold.
+          if (route_lower_bound > cost_threshold_) {
+            continue;
+          }
+        }
       }
 
       // Expand from the end node in forward direction.
@@ -655,6 +693,24 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
       if ((rev_pred.not_thru() && rev_pred.not_thru_pruning()) ||
           hierarchy_limits_reverse_[rev_pred.endnode().level()].StopExpanding(rev_pred.distance())) {
         continue;
+      }
+
+      // Check if this branch can be pruned. It's implementation of the reach-based pruning technique
+      // for bidirectional astar: https://repub.eur.nl/pub/16100/ei2009-10.pdf .
+      if (cost_threshold_ != std::numeric_limits<float>::max() &&
+          rev_pred.predecessor() != kInvalidLabel) {
+        const auto& rev_pred_pred = edgelabels_reverse_[rev_pred.predecessor()];
+        const auto pred_tile = graphreader.GetGraphTile(rev_pred_pred.endnode());
+        if (pred_tile != nullptr) {
+          // Estimate lower bound cost for the shortest path that goes through the current edge.
+          float route_lower_bound =
+              rev_pred_pred.cost().cost + rev_pred.transition_cost().cost + fwd_pred.sortcost() -
+              astarheuristic_forward_.Get(pred_tile->get_node_ll(rev_pred_pred.endnode()));
+          // Prune this edge if estimated lower bound cost exceeds the cost threshold.
+          if (route_lower_bound > cost_threshold_) {
+            continue;
+          }
+        }
       }
 
       // Get the opposing predecessor directed edge. Need to make sure we get
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -714,23 +714,22 @@ bool BidirectionalAStar::SetForwardConnection(GraphReader& graphreader, const BD
     c = pred.cost().cost + oppcost + opp_pred.transition_cost().cost;
   }
 
+  // Set thresholds to extend search
+  if (cost_threshold_ == std::numeric_limits<float>::max() || c < best_connections_.front().cost) {
+    if (desired_paths_count_ == 1) {
+      cost_threshold_ = c + kThresholdDelta;
+    } else {
+      cost_threshold_ = (1.f + kAlternativeCostExtend) * c + kThresholdDelta;
+      iterations_threshold_ =
+          edgelabels_forward_.size() + edgelabels_reverse_.size() + kAlternativeIterationsDelta;
+    }
+  }
+
   // Keep the best ones at the front all others to the back
   best_connections_.emplace_back(CandidateConnection{pred.edgeid(), oppedge, c});
 
   if (c < best_connections_.front().cost)
     std::swap(best_connections_.front(), best_connections_.back());
-
-  // Set thresholds to extend search
-  if (cost_threshold_ == std::numeric_limits<float>::max()) {
-    float sortcost = std::max(pred.sortcost() + cost_diff_, opp_pred.sortcost());
-    if (desired_paths_count_ == 1) {
-      cost_threshold_ = sortcost + kThresholdDelta;
-    } else {
-      cost_threshold_ = sortcost + std::max(kAlternativeCostExtend * sortcost, kThresholdDelta);
-      iterations_threshold_ =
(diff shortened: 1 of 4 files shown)


src/thor/bidirectional_astar.cc
```diff
@@ -156,20 +156,46 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
...
                                             const graph_tile_ptr& tile,
                                             const baldr::TimeInfo& time_info) {
+  // Skip if this is a regular edge superseded by a shortcut.
+  if (shortcuts & meta.edge->superseded()) {
+    return false;
+  }
+
+  graph_tile_ptr t2 = nullptr;
+  baldr::GraphId opp_edge_id;
+  const auto get_opp_edge_data = [&t2, &opp_edge_id, &graphreader, &meta, &tile]() {
```
genadz, line 166: we call this logic in several places, so moved out this logic into a separate lambda function

src/thor/bidirectional_astar.cc
```diff
@@ -156,20 +156,46 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
...
+    if (!get_opp_edge_data())
+      return false;
+
+    auto& opp_edgestatus = FORWARD ? edgestatus_reverse_ : edgestatus_forward_;
+    const auto opp_edge_set = opp_edgestatus.Get(opp_edge_id).set();
+    // Synchronize shortcuts for both directions. If this shortcut has been already
+    // encountered on the opposing search we should do the same now: skip or traverse.
+    if ((opp_edge_set != EdgeSet::kSkipped &&
+         hierarchy_limits[meta.edge_id.level() + 1].StopExpanding(pred.distance())) ||
+        opp_edge_set == EdgeSet::kPermanent || opp_edge_set == EdgeSet::kTemporary) {
```
genadz, line 192: choose to skip or use the shortcut based on the opposing search results

src/thor/bidirectional_astar.cc
```diff
@@ -531,9 +547,14 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
...
         // reverse search tree. Do not expand further past this edge since it will just
-        // result in other connections.
-        if (edgestatus_reverse_.Get(fwd_pred.opp_edgeid()).set() == EdgeSet::kPermanent) {
-          if (SetForwardConnection(graphreader, fwd_pred)) {
+        // result in other connections. Handle special edge case when we encountered the
+        // destination edge that wasn't still pulled out of the queue.
+        const auto opp_status = edgestatus_reverse_.Get(fwd_pred.opp_edgeid());
+        if (opp_status.set() == EdgeSet::kPermanent ||
+            (opp_status.set() == EdgeSet::kTemporary &&
+             edgelabels_reverse_[opp_status.index()].predecessor() == kInvalidLabel)) {
```
genadz, line 555: we can pass through the destination edge (in case we have several snapping edges) and prune all branches earlier than pull out this edge from the queue. In order to prevent loosing such paths , we can add this connection even if the opposing edge marked as "temporary"

src/thor/bidirectional_astar.cc
```diff
@@ -617,11 +643,31 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
...
+    // search exhausts limits on the particular level - it should stop and wait until reverse/forward
+    // search exhausts limits on the same hierarchy level. This logic ensures local optimality near
+    // the origin and destination and provides valid conditions for the reach-based pruning.
+    bool force_forward = false;
+    bool force_reverse = false;
+    assert(hierarchy_limits_forward_.size() == hierarchy_limits_forward_.size());
+    for (size_t level = TileHierarchy::levels().size() - 1; level > 0; --level) {
+      if (hierarchy_limits_reverse_[level].StopExpanding(rev_pred.distance()) &&
+          !hierarchy_limits_forward_[level].StopExpanding(fwd_pred.distance())) {
+        force_forward = true;
```
genadz, line 656: force forward pass in case the reverse search exhausted hierarchy limits for this level

src/thor/bidirectional_astar.cc
```diff
@@ -617,11 +643,31 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
...
+        break;
+      } else if (hierarchy_limits_forward_[level].StopExpanding(fwd_pred.distance()) &&
+                 !hierarchy_limits_reverse_[level].StopExpanding(rev_pred.distance())) {
+        force_reverse = true;
```
genadz, line 660: force reverse pass in case the forward search exhausted hierarchy limits for this level

src/thor/bidirectional_astar.cc
```diff
@@ -638,6 +684,24 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
...
+      // for bidirectional astar: https://repub.eur.nl/pub/16100/ei2009-10.pdf .
+      if (cost_threshold_ != std::numeric_limits<float>::max() &&
+          fwd_pred.predecessor() != kInvalidLabel) {
+        const auto& fwd_pred_pred = edgelabels_forward_[fwd_pred.predecessor()];
+        if (auto pred_tile = graphreader.GetGraphTile(fwd_pred_pred.endnode());
+            pred_tile != nullptr) {
+          // Estimate lower bound cost for the shortest path that goes through the current edge.
+          float route_lower_bound =
+              fwd_pred_pred.cost().cost + fwd_pred.transition_cost().cost + rev_pred.sortcost() -
+              astarheuristic_reverse_.Get(pred_tile->get_node_ll(fwd_pred_pred.endnode()));
```
genadz, line 697: Estimate lower bound cost for the route that goes through the current edge. The technique from the article was adopted to the edges (in the article used vertices). Let me describe the formula for vertices. Let's suppose the current vertex we pulled out of the forward queue is `v` (`v'` - a vertex with min sortcost in the reverse queue), so the are following equations are true:
```
cost_f(v) + sortcost_r(v') - h_r(v) <= cost_f(v) + sortcost_r(v) - h_r(v) = cost_f(v) + cost_r(v) + h_r(v) - h_r(v) =
cost_f(v)  + cost_r(v) = cost(shortest path through v)

where
cost, sortcost, h - costing function, sortcost, heuristic function;
*_f, *_r - indicates forward and reverse directions respectively.
```
these equations say that `cost_f(v) + sortcost_r(v') - h_r(v) ` is a lower bound cost for the route through `v`. Using this logic we can replace vertices by edges and get approximately the same formula for our case.

src/thor/bidirectional_astar.cc
```diff
@@ -714,24 +796,23 @@ bool BidirectionalAStar::SetForwardConnection(GraphReader& graphreader, const BD
...
-  if (c < best_connections_.front().cost)
-    std::swap(best_connections_.front(), best_connections_.back());
-
   // Set thresholds to extend search
-  if (cost_threshold_ == std::numeric_limits<float>::max()) {
-    float sortcost = std::max(pred.sortcost() + cost_diff_, opp_pred.sortcost());
+  if (cost_threshold_ == std::numeric_limits<float>::max() || c < best_connections_.front().cost) {
     if (desired_paths_count_ == 1) {
-      cost_threshold_ = sortcost + kThresholdDelta;
+      cost_threshold_ = c + kThresholdDelta;
```
genadz, line 802: correct cost threshold. it guarantees that the optimal route will not be missed

valhalla/thor/edgestatus.h
```diff
@@ -16,9 +16,11 @@ enum class EdgeSet : uint8_t {
                          // reset due to encountering a complex restriction:
                          // https://github.com/valhalla/valhalla/issues/2103
   kPermanent = 1,        // Permanent - shortest path to this edge has been found
-  kTemporary = 2         // Temporary - edge has been encountered but there could
+  kTemporary = 2,        // Temporary - edge has been encountered but there could
                          //   still be a shorter path to this edge. This edge will
                          //   be "adjacent" to an edge that is permanently labeled.
+  kSkipped = 3           // Skipped - edge has been encountered but was thrown out
```
genadz, line 22: new edge status for shortcuts that were encountered but were not used

kevinkreiser:
i have to ask it :smile: whats the performance change, is it similar to the comment you posted before when you first started this approach?

also thank you so much for the extensive description it goes a long way to helping make the changes reviewable. i will take an in depth look at the end of the day

genadz:
## Comparing

### Shortest path

#### Performance
passed requests with `response_time` field: 13631
**master**
```
5%	10%	50%	90%	95%	99%	100%	
0.0068	0.0098	0.0625	0.2281	0.4471	1.5088	3.1784	
mean: 0.1261
```
**branch**
```
5%	10%	50%	90%	95%	99%	100%	
0.0065	0.0095	0.0644	0.2288	0.5087	1.7072	3.3404	
mean: 0.1363
```
**branch-with-no-pruning**
```
5%	10%	50%	90%	95%	99%	100%	
0.0070	0.0101	0.0765	0.3554	0.8619	2.1825	3.9858	
mean: 0.1851
```
I measured timings for the _master_ branch, __current_ branch_ and _current branch with disabled pruning_ to showcase the influence of pruning technique on performance. As we can see, the average response time increased only by **8%** for the branch and by **46%** for the branch with disabled pruning.

#### Routes quality
Number of routes to analyze: 13797

This branch affected 687 (out of 13797) routes, it's about **5%**. Final cost of the optimal path was improved in 581 cases and degraded in 106 cases (in terms of ETA, the numbers are the following: 532 <-> 156). So, we can say that in **85%** cases changes were positive.

It's important to notice, that I got only 10 routes (out of 13797) that were different for the branch with pruning and without pruning. So, in **99.9%** cases we can be sure that the pruning technique didn't cut a branch with more optimal route. The reasonable question may be here: why we actually got a difference ? - after analyzing differences I found out that the main reason is this one https://github.com/valhalla/valhalla/issues/2979 . We can't completely "synchronize" shortcuts usage for bidirectional search, there may be a cases when the shortcut was expanded in forward direction but we came into the middle of the shortcut using reverse direction. To solve this we probably should recover shortcuts while traversing trees, but it's too expensive now.

valhalla/thor/edgestatus.h
kevinkreiser: need to add this to the expansion callback for visualization haha

genadz:
> i have to ask it  whats the performance change, is it similar to the comment you posted before when you first started this approach?
> 
> also thank you so much for the extensive description it goes a long way to helping make the changes reviewable. i will take an in depth look at the end of the day

Thank you! I posted results for the "shortest path" case, also will add results for alternatives.

dnesbitt61:
I tried some sort of reach based pruning a couple of years ago and saw some performance gains for bicycle routes but there were too many changes to routes for me to feel comfortable.

I tried this branch for my long, test bicycle route and performance decreased by 21%. I would advocate creating a simpler, bidirectional A* algorithm for use where no hierarchies exist (that is how I create my data for bicycling) or are needed (e.g. bicycle and pedestrian routes).

src/thor/bidirectional_astar.cc
```diff
@@ -780,24 +861,23 @@ bool BidirectionalAStar::SetReverseConnection(GraphReader& graphreader, const BD
...
           edgelabels_forward_.size() + edgelabels_reverse_.size() + kAlternativeIterationsDelta;
     }
   }
 
+  // Keep the best ones at the front all others to the back
+  best_connections_.emplace_back(CandidateConnection{fwd_edge_id, rev_pred.edgeid(), c});
+
+  if (c < best_connections_.front().cost)
+    std::swap(best_connections_.front(), best_connections_.back());
+
```
kevinkreiser, line 880: any significance to moving this down here?

kevinkreiser: oh yes i see it, in the first if second part of the or clause, you want ot check the current best connection against this one

src/thor/bidirectional_astar.cc
```diff
@@ -780,24 +861,23 @@ bool BidirectionalAStar::SetReverseConnection(GraphReader& graphreader, const BD
...
   // Set thresholds to extend search
-  if (cost_threshold_ == std::numeric_limits<float>::max()) {
-    float sortcost = std::max(rev_pred.sortcost(), fwd_pred.sortcost() + cost_diff_);
+  if (cost_threshold_ == std::numeric_limits<float>::max() || c < best_connections_.front().cost) {
     if (desired_paths_count_ == 1) {
-      cost_threshold_ = sortcost + kThresholdDelta;
+      cost_threshold_ = c + kThresholdDelta;
     } else {
-      cost_threshold_ = sortcost + std::max(kAlternativeCostExtend * sortcost, kThresholdDelta);
+      cost_threshold_ = (1.f + kAlternativeCostExtend) * c + kThresholdDelta;
```
kevinkreiser, line 869: why not just set the constant to 1.1 and explain its a multiplier in the comment for it?

kevinkreiser: also i notice the semantic difference of always adding the thresholdDelta rather than using it only if its greater than the sortcost. maybe this is worth a description? i guess its because now we dont double the sortcost we use the actual cost of both sides of the route added together, scale that by 1.1 to do the extension and then we add some fluff (delta) just in case the route is super short and the scaling doesnt do much. anyway a comment to that effect would make this less magic. i know we have some above but i feel like they dont capture the full picture

src/thor/bidirectional_astar.cc
```diff
@@ -617,11 +643,31 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
       return {};
     }
 
+    // Exhaust hierarchy limits simultaneously in both directions. As soon as forward/reverse
```
kevinkreiser, line 646: @dnesbitt61 i am wondering about your comment on performance for routes that dont do hierarchy culling. where do you think that the slow down is coming from? the way this code reads to me is that we are trying to balance the searches, but wouldnt this never happen for bike and ped routes? since stop expanding would never return true? i'm just wondering where the heck the performance drop is coming, maybe its all just the updated heuristic using the actual cost. did you notice if it was more iterations?

src/thor/bidirectional_astar.cc
```diff
@@ -531,9 +547,14 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
...
-        if (edgestatus_reverse_.Get(fwd_pred.opp_edgeid()).set() == EdgeSet::kPermanent) {
-          if (SetForwardConnection(graphreader, fwd_pred)) {
+        // result in other connections. Handle special edge case when we encountered the
+        // destination edge that wasn't still pulled out of the queue.
+        const auto opp_status = edgestatus_reverse_.Get(fwd_pred.opp_edgeid());
+        if (opp_status.set() == EdgeSet::kPermanent ||
+            (opp_status.set() == EdgeSet::kTemporary &&
+             edgelabels_reverse_[opp_status.index()].predecessor() == kInvalidLabel)) {
+          if (SetForwardConnection(graphreader, fwd_pred) &&
+              opp_status.set() == EdgeSet::kPermanent) {
```
kevinkreiser, line 557: i am wondering what this added check does its not clear to me why we check for permanent here. is this the part you are refering to when you say

> wasn't still pulled out of the queue

src/thor/bidirectional_astar.cc
```diff
@@ -156,20 +156,46 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
...
-  // that level. If using a shortcut, set the shortcuts mask. Skip if this is a regular
-  // edge superseded by a shortcut.
+  // that level. If using a shortcut, set the shortcuts mask.
   if (meta.edge->is_shortcut()) {
-    if (hierarchy_limits[meta.edge_id.level() + 1].StopExpanding(pred.distance())) {
+    if (!get_opp_edge_data())
+      return false;
+
+    const auto& opp_edgestatus = FORWARD ? edgestatus_reverse_ : edgestatus_forward_;
+    const auto opp_edge_set = opp_edgestatus.Get(opp_edge_id).set();
```
kevinkreiser, line 187: i guess this hash lookup isnt 100% free :smile: maybe for modes of transport that dont use shortcuts we can disable this to help bike and ped performance

what do you think @dnesbitt61 and @genadz ?

kevinkreiser:
honestly that wasnt too bad to review, i had a couple small questions but it made sense to me mostly. the performance numbers for the auto routes look fine to me. the basic gist is that the 99% percentile of an alreay skewed request set towards very long routes, shows a pretty big performance drop. honestly i dont care too much about p99 but i guess my quesiton is, is @dnesbitt61 route a p99 route or a p50 route. is bike routing particularly impacted or is it just as bad as auto and its that we are focused on the 99th percentile (i feel like we shouldnt focus on that tbh). i mean i understand we want to be performant on very long routes, but if those constitute 1% of the traffic that would come to a running service it doesnt make sense to focus so much effort on make them efficient at the cost of getting worse routes

src/thor/bidirectional_astar.cc
genadz: on my local machine I noticed that iterations increased by ~6%

src/thor/bidirectional_astar.cc
genadz: make sense

src/thor/bidirectional_astar.cc
genadz: this just says that if the opposing edge marked as "temporary" we should keep searching

src/thor/bidirectional_astar.cc
genadz: we can also take into account the fact if we use hierarchies or not

src/thor/bidirectional_astar.cc
genadz: looks like we can skip `kThresholdDelta` in case of `ignore_hierarchy_limits=true` (no hierarchy prunings, no shortcuts). But it should be tested on some big bicycle/pedestrian dataset that I don't have right now. It can be done in a separate PR

src/thor/bidirectional_astar.cc
genadz: yeah, for short routes scale `1.1` is not enough to find alternatives. Ok, will add a comment about that

valhalla/thor/edgestatus.h
genadz: ok

genadz:
@dnesbitt61 
> I tried some sort of reach based pruning a couple of years ago and saw some performance gains for bicycle routes but there were too many changes to routes for me to feel comfortable.

That's why I did some stuff about shortcuts "synchronization" and balancing hierarchies during trees expansion. It eliminated false positive prunings.

> I tried this branch for my long, test bicycle route and performance decreased by 21%. I would advocate creating a simpler, bidirectional A* algorithm for use where no hierarchies exist (that is how I create my data for bicycling) or are needed (e.g. bicycle and pedestrian routes).

As I mentioned above, on my local machine I noticed increase by ~6% in number of iterations. But, the same time this test
```
./valhalla_run_route --config ../../data/config.json --multi-run 100 -j '{"locations":[{"lat":39.64225,"lon":-76.10536,"type":"break"},{"lat":38.899677,"lon":-77.050916,"type":"break"}],"costing":"bicycle","costing_options":{"bicycle":{"bicycle_type":"Road","use_roads":"0.0","cycling_speed":"25.0","use_hills":"0.5"}}}' | grep GetBestPath
```
that you proposed some time ago showed greater performance drop. With last changes (using `ignore_hierarchy_limits` flag) it's about ~10%.

genadz:
@kevinkreiser 
> another track to take with that is we could relax some of the optimality when we know the route is going to be very long. one could argue that on a very long route optimality of somethign that is already a combination of imperfect measurements isnt that important when we are talking about costs that differ by 1% or something over thousands of miles. i'm not sure if implementing this wouldnt make the code spaghetti but it is a possibility. the problem here is that we cant really just measure performance in the typical way to see what we need to optimize. here the algorithm itself causes more iterations, we're probably already optimized on the cpu cycles that an iteration takes so the only thing we can do is intelligently remove iterations by culling search space.

interesting thoughts. Probably we can do something similar to what alternatives routes search does: after we found a route - limit number of additional iterations. should it be implemented in this PR ?

genadz:
@kevinkreiser 
> honestly that wasnt too bad to review,

Thank you! I will consider this a compliment)))

src/thor/bidirectional_astar.cc
dnesbitt61: It is more iterations on an already long search: about 6% increase on forward direction and 10% on reverse. Bicycle and pedestrian routes need a way to decrease the search space/# of iterations and I was hoping this might since the title has the word **pruning**!

genadz: as I already mentioned this PR actually fixes a "bug". If it wasn't, we would discuss iterations decrease now

src/thor/bidirectional_astar.cc
genadz: hm, on my local planet I didn't notice any difference on `pedestrian_roundabout_routes.txt`, `bicycle_routes.txt`, and `pedestrian_routes.txt` from `test_requests` folder

valhalla/thor/edgestatus.h
genadz: so, need help here

kevinkreiser: i mentioned it before i saw how it was used, no worries just ignore my comment :smile:

genadz: ok)

kevinkreiser:
I saw you did some of the optimizations where certain costings dont do hierarchy culling, did you mean that neither of them had any effect? thats too bad. from my perspective this is ready to be shipped. @dnesbitt61 do you think we should further investigate ped and bike performance before merging?

genadz:
@kevinkreiser 
> I saw you did some of the optimizations where certain costings dont do hierarchy culling, did you mean that neither of them had any effect? 

not really, here https://github.com/valhalla/valhalla/pull/3257#issuecomment-896670638 I mentioned that it's still slower than master by ~10% (on my local machine and planet) using David's test:
```
./valhalla_run_route --config ../../data/config.json --multi-run 100 -j '{"locations":[{"lat":39.64225,"lon":-76.10536,"type":"break"},{"lat":38.899677,"lon":-77.050916,"type":"break"}],"costing":"bicycle","costing_options":{"bicycle":{"bicycle_type":"Road","use_roads":"0.0","cycling_speed":"25.0","use_hills":"0.5"}}}' | grep GetBestPath

```

kevinkreiser:
@dnesbitt61 `"use_roads": 0.0` ouch!

dnesbitt61:
maybe it should be "prefer_roads": 0.0 - meaning the bicycle route should try to stay on cycleways and bicycle lanes (basically a safer route if possible).

kevinkreiser:
@dnesbitt61 yep that bit is clear i was just pointing it out because i assume it puts a lot of penalties which would i think make a worst case for expansion (big costing numbers)

kevinkreiser (approved):
this looks good to me but i'd like @dnesbitt61 to weigh in on whether we need to do performance optimization for bike and ped now or we can do it in a separate pr

dnesbitt61 (approved):

Issue #2524, opened 2020-08-12
Path Cost Computed in Reverse Differs from Forward


kevinkreiser:
https://github.com/valhalla/valhalla/pull/2425/files#r469288577

In the recosting PR we found out that the same path found by timedep_forward and timedep_reverse yielded different costs. from recollection i think it was the turn costs. this points to a descrepancy either in the logic between `DynamicCost::TransitionCost` vs `DynamicCost::TransitionCostReverse` or the way its being used in the forward and backward searches. It should be noted that comparisions between bidirectional astar and timedep_forward have not shown any differences (though its possible those routes were not long enough to do so)

merkispavel:
I found this issue while working on the same bug in bidir A*:
> So the issue is the `TransitionCostReverse` returns different cost from the `TransitionCost` during recosting.
Looks like the cause is hidden there https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L237:L238. We pass `pred.has_measured_speed()` but we should use `current_edge.has_mesagured_speed()` to be consistent with how we calculate forward transition cost https://github.com/valhalla/valhalla/blob/master/src/sif/autocost.cc#L567

Looks like it's exactly what you described @kevinkreiser. And this is across all algorithms I guess ...

When you have long routes - this bug doesn't impact a lot: cost of the intersections is usually smaller than edge costs. However when you request small routes - you can find that the routes are "swapped" after recosting, i.e cost of the first route is greater than cost of the alternative one. That's annoying

kevinkreiser:
Ah yes! someone else mentioned this recently but its already slipping my mind. This was the set of unit tests, the recosting ones, where I had to disable some of them because I couldnt find the source of this bug. Are you looking at fixing this? By the way where is that quote from an older PR or another issue??

kevinkreiser:
these are the tests that show it: https://github.com/valhalla/valhalla/blob/master/test/gurka/test_recost.cc#L132-L135

merkispavel:
> Are you looking at fixing this?
Yes, I'm working on the fix

The quote is from the internal issue. We've discovered it pretty recently(like a week ago)

kevinkreiser:
rock on, thanks!

PR #3260, opened 2021-08-11
Fix TransitionReverseCost usage

merkispavel:
# Issue

https://github.com/valhalla/valhalla/issues/2524
https://github.com/valhalla/valhalla/issues/2524#issuecomment-889152247

The original issue is that in some cases main route cost can be greater than alternative cost. This happens because of the bug in the `TransitionCostReverse` usage, i.e we pass parameters of the wrong edge. So what happens is
1. Main route gets less cost before recosting(i.e when connections are found)
2. After recosting main route gets **greater** cost because `TransitionCost` function is used now for cost calculation
image

The picture after the fix
Screenshot 2021-08-11 at 10 28 01

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -230,17 +230,26 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
     }
   }
 
-  // Get cost. Separate out transition cost.
-  sif::Cost transition_cost =
-      FORWARD ? costing_->TransitionCost(meta.edge, nodeinfo, pred)
-              : costing_->TransitionCostReverse(meta.edge->localedgeidx(), nodeinfo, opp_edge,
-                                                opp_pred_edge, pred.has_measured_speed(),
-                                                pred.internal_turn());
+  // Get cost
   uint8_t flow_sources;
   sif::Cost newcost =
-      pred.cost() + transition_cost +
-      (FORWARD ? costing_->EdgeCost(meta.edge, tile, time_info.second_of_week, flow_sources)
-               : costing_->EdgeCost(opp_edge, t2, time_info.second_of_week, flow_sources));
+      pred.cost() + (FORWARD
+                         ? costing_->EdgeCost(meta.edge, tile, time_info.second_of_week, flow_sources)
+                         : costing_->EdgeCost(opp_edge, t2, time_info.second_of_week, flow_sources));
+
+  // Separate out transition cost.
+  InternalTurn internal_turn;
+  sif::Cost transition_cost;
+  if (FORWARD) {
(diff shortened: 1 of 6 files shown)


src/thor/bidirectional_astar.cc
```diff
@@ -230,17 +230,26 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
...
+                         ? costing_->EdgeCost(meta.edge, tile, time_info.second_of_week, flow_sources)
+                         : costing_->EdgeCost(opp_edge, t2, time_info.second_of_week, flow_sources));
+
+  // Separate out transition cost.
+  InternalTurn internal_turn;
+  sif::Cost transition_cost;
+  if (FORWARD) {
+    transition_cost = costing_->TransitionCost(meta.edge, nodeinfo, pred);
+  } else {
+    internal_turn = costing_->TurnType(meta.edge->localedgeidx(), nodeinfo, opp_edge, opp_pred_edge);
```
merkispavel, line 246: `TurnType` method doesn't look like a trivial operation so I decided to memoize its result and reuse later when adding new label

test/gurka/test_recost.cc
```diff
@@ -7,6 +7,57 @@ using namespace valhalla;
 
 const std::unordered_map<std::string, std::string> build_config{{"mjolnir.shortcuts", "false"}};
 
+TEST(recosting, forward_vs_reverse) {
```
merkispavel, line 10: I tried to write a gurka test for bidir-a* for a few hours but it's too tough, sorry

merkispavel:
I RAD tested the changes. Though I didn't expect the diff to be big enough - it is(7%).
- There are good examples where dive bombs got fixed with this PR(see pictures below)
- However, there are cases in the opposite direction

I believe whatever diff we have - we should merge this fix as it makes the routes consistent with costs
image
image
image
image

And a few bad cases from the diff
image
image
image

genadz:
@merkispavel
1. do you have a chance to measure performance diff ?
2. according to the penultimate example, looks like we allow or forbid the turn depending on the expansion order. Did you investigate this example in more detail ?

test/gurka/test_recost.cc
merkispavel: correct. it's red on master and green with after the fix

genadz (approved):

merkispavel:
> 1. do you have a chance to measure performance diff ?
I'm not sure results from mac os will be sane. I had problems with getting stable benchmark results in the past with OS X

> 2. according to the penultimate example, looks like we allow or forbid the turn depending on the expansion order. Did you investigate this example in more detail ?
I looked into a few divebombs ^ a bit deeply. There's a chance that hierarchy limits are the cause: whenever origin point is at the low level road - divebomb is there

PR #3132, opened 2021-06-08
[WIP] Phonemes

gknisely:
# Issue

What issue is this PR targeting? If there is no issue that addresses the problem, please open a corresponding issue and link it here.

## Tasklist

 - [ ] Add tests
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Update the docs with any new request parameters or changes to behavior described
 - [ ] Update the [changelog](CHANGELOG.md)
 - [ ] If you made changes to the lua files, update the [taginfo](taginfo.json) too.

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #3261, opened 2021-08-11
Add Z-level field to EdgeInfo

### LINKED ISSUE #3267: tunnels and bridges unit tests
over in https://github.com/valhalla/valhalla/pull/3264 we finally added tests for tagged values support but unfortunately it was done in the old style. when doing data tests it is highly encouraged to use the gurka style. we should convert this test to gurka and when doing the new phoneme and layer tag support we should do the same. cc @SiarheiFedartsou

SiarheiFedartsou:
# Issue

It will allow us to use this field in `loki` in order to better choose edge from where to start routing in case of multi-level roads. User of Valhalla can send current Z-level and it will give us a hint which level we should choose.

Z-level is obtained from this OSM tag https://wiki.openstreetmap.org/wiki/Key:layer

also closes #3267 via removing extra test
## Tasklist

 - [ ] Add tests

## Requirements / Relations


valhalla/baldr/graphconstants.h
```diff
@@ -201,6 +201,10 @@ constexpr uint32_t kMaxAddedTime = 255;
 // Elevation constants
 constexpr float kNoElevationData = 32768.0f;
 
+// Z-level range
+constexpr int8_t kMinEdgeZLevel = -6;
```
SiarheiFedartsou, line 205: based on https://taginfo.openstreetmap.org/keys/layer#values

SiarheiFedartsou: From https://wiki.openstreetmap.org/wiki/Key:layer:
> Values between -5 and 5 are generally used; 0 should not be used explicitly.

Should we use [-5;5] range instead?

valhalla/mjolnir/osmway.h
```diff
@@ -1727,6 +1741,9 @@ struct OSMWay {
   // Truck speed in kilometers per hour
   uint8_t truck_speed_;
 
+  // Z-level of the way relatively to other levels
+  int8_t z_level_;
```
kevinkreiser, line 1745: you should rename the spare below and use that instead

SiarheiFedartsou: okay, will do. Could you elaborate here? I cannot get why do we have this spare field. Do we serialize ways somewhere?

valhalla/baldr/directededge.h
```diff
@@ -1157,7 +1172,8 @@ class DirectedEdge {
   uint64_t bss_connection_ : 1; // Does this lead to(come out from) a bike share station?
   uint64_t stop_sign_ : 1;      // Stop sign at end of the directed edge
   uint64_t yield_sign_ : 1;     // Yield/give way sign at end of the directed edge
-  uint64_t spare4_ : 7;
+  uint64_t z_level_ : 4;        // Z-level of the edge
```
kevinkreiser, line 1175: i dont think this should go in the directededge, it makes more sense in edgeinfo. first because its a property of the edge that doesnt change based on the direction of travel and second because we wont be using it to change the costing function for the edge. its a "display" property, those belong in edgeinfo. now the next problem is that edgeinfo is already chaulk full and its harder to modify hahaha but i do believe it belongs there

SiarheiFedartsou:
> with respect to rendering order it makes sense, but when we are talking about topology i would expect to put it on the nodes. sadly, im going to have to guess that osm doesnt populate this very well and so rendering order (tagging for the map) is the best we can get, is this a correct assumption?

Yes. More than it: one more feature I have in mind is that we can attach z-levels to the route geometry in response and it will allow clients to better understand how they should render routes taking into account bridges, tunnels etc.

valhalla/baldr/directededge.h
SiarheiFedartsou: Well :) 
What do you think if we add [1 bit here](https://github.com/valhalla/valhalla/blob/79c404631c475607411252b78de1ee97a0467ed0/valhalla/baldr/edgeinfo.h#L207) indicating if we have smth like `EdgeInfoInnerExt` written after `EdgeInfoInner`? `EdgeInfoInnerExt` could look smth like:
```
struct EdgeInfoInnerExt {
   uint64_t z_level: 4;
   uint64_t z_level: 60;
}
```
?

SiarheiFedartsou: Or we could come up with some more sophisticated variable-size encoding...

SiarheiFedartsou: Probably we could utilize these things here https://en.wikipedia.org/wiki/Elias_delta_coding or https://en.wikipedia.org/wiki/Elias_gamma_coding. It would allow us to not waste bits when they are not actually needed for the cost of additional bits for each value 🤔 
In other words at the moment we have to have this `EdgeInfoInnerExt` of fixed size and z-level is always 4 bit. For example with delta coding we could encode our [-6;8] range with 1-8 bits depending on actual value and do not waste bits if they are unused.
WDYT?

valhalla/baldr/graphconstants.h
PureW: Could the doc-string contain that wiki link? I'm sure that would be helpful for anyone wondering what unit Z-level is.

kevinkreiser:
since this information is very sparse i think this is a perfect candidate for the tag/value support that edgeinfo already offers actually. you pick an enumeration to represent your tag, in this case the layer tag, and then you can store your value in a single byte. it does mean we will have to get the tag values for whatever edge we care about but we do that anyway when we get the street names.

an example of this pattern is here: https://github.com/valhalla/valhalla/pull/2608/files

unfortunately there is a lot more in that pr because it also is adding support for this concept for the first time, but the concept is pretty straight forward, you add a new entry to that enum in graphconstants.h (we shouldnt call it TaggedName, it should be TaggedValue since its not just about names it can be any series of bytes). then you add the parsing code as you already have done, for you i would suggest you cast the value of your integer to a char and store a single char offset by your minimum value that you expect to see, this will keep the char unsigned. then in the code that reads it back out in edgeinfo you just undo the operation, cast it back to int and offset it back using the minimum value.

what do you think about that?

SiarheiFedartsou:
@kevinkreiser seems it would solve the issue, trying to implement

> we shouldnt call it TaggedName, it should be TaggedValue since its not just about names it can be any series of bytes

Should we rename it then?

kevinkreiser:
@SiarheiFedartsou i think we should yes

SiarheiFedartsou:
@kevinkreiser 
I updated PR. Could you please check if I got your idea about tagged names right and if so I'll finish PR(will add some more tests and do this renaming stuff)

src/baldr/edgeinfo.cc
```diff
@@ -195,5 +208,15 @@ json::MapPtr EdgeInfo::json() const {
...
 
+const std::unordered_map<TaggedName, std::string>& EdgeInfo::GetTags() const {
+  if (tagCache_.empty()) {
+    auto tags = GetTaggedNamesAndTypes();
+    for (const auto& tag : tags) {
+      tagCache_.emplace(static_cast<TaggedName>(tag.second), tag.first);
+    }
+  }
+  return tagCache_;
+}
```
kevinkreiser, line 219: i like this approach much much better than what was done above in `GetTaggedNamesAndTypes`. i think we should move this caching into that function and indeed return a map of `Tag -> Value` like this does rather than `Value -> Tag` like the above method does. i know it will cause some work downstream but its clearly the right way to go. maybe as a first step just do the caching up there and then as a second step, in one isolated commit flip the key and the value in the returned unordered_map. what do you think?

src/mjolnir/osmway.cc
```diff
@@ -173,16 +181,20 @@ OSMWay::GetNames(const std::string& ref, const UniqueNames& name_offset_map, uin
...
     // tunnel names
-    tokens = GetTagTokens(name_offset_map.name(tunnel_name_index_));
+    auto tokens = GetTagTokens(name_offset_map.name(tunnel_name_index_));
     for (const auto& t : tokens) {
       names.emplace_back(std::to_string(static_cast<uint8_t>(TaggedName::kTunnel)) + t);
     }
   }
 
+  if (z_level_ != 0) {
+    names.emplace_back(std::to_string(static_cast<uint8_t>(TaggedName::kZLevel)) +
```
kevinkreiser, line 194: this is on an other PR too #3132 . i think the original implementation must have screwed this up but i didnt go check the logic yet. we werent supposed to `to_string` the enum value, that doesnt make any sense, because that wont fit in a single byte! we want to cast it to a char and use that... if this is the case i think we royally screwed up this whole idea in the original implementation..

kevinkreiser: and yeah its screwed up: https://github.com/valhalla/valhalla/pull/2608/files#r687122174

kevinkreiser: issue is here, *looks* like we can trivially fix it: https://github.com/valhalla/valhalla/issues/3262

src/baldr/edgeinfo.cc
SiarheiFedartsou: Well, I just realized we cannot use unordered_map here, because tags can have multiple values(tunnels is an example if I properly get their parsing code), but I think we can go with multimap here...

kevinkreiser: yep exactly

gknisely:
@SiarheiFedartsou make sure you update the [taginfo](taginfo.json) too.

SiarheiFedartsou:
> @SiarheiFedartsou make sure you update the [taginfo](https://github.com/valhalla/valhalla/blob/sf-z-level/taginfo.json) too.

Okay, description can be in free form, right?

SiarheiFedartsou:
@kevinkreiser 
any ideas how I could break isochrone here:
```
[ RUN      ] Isochrones.Basic
/Users/distiller/project/test/isochrone.cc:96: Failure
  actual_geom.Size()
    Which is: 269
  expected_geom.Size()
    Which is: 271
/Users/distiller/project/test/isochrone.cc:96: Failure
  actual_geom.Size()
    Which is: 269
  expected_geom.Size()
    Which is: 271
```
?

test/isochrone.cc
```diff
@@ -148,15 +148,15 @@ TEST(Isochrones, Basic) {
...
...147924,52.093540],[5.148321,52.094324],[5.149321,52.094187],[5.149858,52.095400],[5.151321,52.095321],[5.151894,52.097364],[5.154121,52.097937],[5.153491,52.098107],[5.153321,52.098841],[5.152974,52.098284],[5.151978,52.098280],[5.150908,52.098524],[5.150321,52.099513],[5.148321,52.099072],[5.146321,52.099355],[5.145730,52.099937],[5.146519,52.100937],[5.146321,52.101629],[5.145019,52.101239],[5.144912,52.099937],[5.144321,52.099476],[5.142828,52.100937],[5.143321,52.101667],[5.144485,52.101937],[5.142656,52.102937],[5.140321,52.102566],[5.140037,52.102221],[5.137321,52.102151],[5.136321,52.103072],[5.135894,52.102364],[5.134059,52.101937],[5.133781,52.099937],[5.134321,52.099377],[5.135597,52.099213],[5.135579,52.096937],[5.134662,52.096596],[5.134362,52.095896],[5.133518,52.095937],[5.132492,52.096108],[5.132153,52.096769],[5.130321,52.097087],[5.126772,52.099388],[5.124938,52.099320],[5.124251,52.098867],[5.124359,52.102937],[5.120321,52.103258],[5.119321,52.102698],[5.116321,52.105645],[5.115321,52.104355],[5.114321,52.104836],[5.112321,52.104914],[5.111321,52.104485],[5.110655,52.105271],[5.108321,52.106148]]],"type":"Polygon"},"type":"Feature"}],"type":"FeatureCollection"})";
```
SiarheiFedartsou, line 151: Have no clue why it has changed. But it seems to be trivial. Also I see other PRs unrelated to isochrone which changed it.

Screenshot 2021-08-13 at 12 54 48
Screenshot 2021-08-13 at 12 55 16
[merged.geojson.zip](https://github.com/valhalla/valhalla/files/6981528/merged.geojson.zip)

@kevinkreiser any ideas?

SiarheiFedartsou: [merged.geojson.zip](https://github.com/valhalla/valhalla/files/6981556/merged.geojson.zip)

test/isochrone.cc
```diff
@@ -148,15 +148,15 @@ TEST(Isochrones, Basic) {
...
...148321,52.099072],[5.146321,52.099355],[5.145730,52.099937],[5.146519,52.100937],[5.146321,52.101629],[5.145019,52.101239],[5.144912,52.099937],[5.144321,52.099476],[5.142828,52.100937],[5.143321,52.101667],[5.144485,52.101937],[5.142656,52.102937],[5.140321,52.102566],[5.140037,52.102221],[5.137321,52.102151],[5.136321,52.103072],[5.135894,52.102364],[5.134059,52.101937],[5.133781,52.099937],[5.134321,52.099377],[5.135597,52.099213],[5.135579,52.096937],[5.134662,52.096596],[5.134362,52.095896],[5.133518,52.095937],[5.132492,52.096108],[5.132153,52.096769],[5.130321,52.097087],[5.126772,52.099388],[5.124938,52.099320],[5.124251,52.098867],[5.124359,52.102937],[5.120321,52.103258],[5.119321,52.102698],[5.116321,52.105645],[5.115321,52.104355],[5.114321,52.104836],[5.112321,52.104914],[5.111321,52.104485],[5.110655,52.105271],[5.108321,52.106148]],"type":"LineString"},"type":"Feature"},{"geometry":{"coordinates":[[5.115328,52.078940]],"type":"MultiPoint"},"properties":{"location_index":0,"type":"snapped"},"type":"Feature"},{"geometry":{"coordinates":[5.115321,52.078937],"type":"Point"},"properties":{"location_index":0,"type":"input"},"type":"Feature"}],"type":"FeatureCollection"})";
```
SiarheiFedartsou, line 159: the same as above 
[merged.geojson.zip](https://github.com/valhalla/valhalla/files/6981559/merged.geojson.zip)

valhalla/thor/attributes_controller.h
```diff
@@ -73,7 +73,7 @@ const std::string kEdgeTruckRoute = "edge.truck_route";
 const std::string kEdgeDefaultSpeed = "edge.default_speed";
 const std::string kEdgeDestinationOnly = "edge.destination_only";
 const std::string kEdgeIsUrban = "edge.is_urban";
-const std::string kEdgeTaggedNames = "edge.tagged_names";
+const std::string kEdgeTaggedValues = "edge.tagged_values";
```
kevinkreiser, line 76: sadly this is a breaking api change. at best we need to support both names for this concept, ie keep both of these strings and if either is found then we turn on flag for this attribute

test/gurka/test_tagged_values.cc
```diff
@@ -137,7 +173,7 @@ TEST_F(TaggedNames, test_taking_tunnel) {
   }
 }
 
-TEST_F(TaggedNames, test_bypass_tunnel) {
+TEST_F(TaggedValues, test_bypass_tunnel) {
```
kevinkreiser, line 176: so these are the places where i would have expected the tunnel and bridge tests. i saw you added more tests in your other pr. can you review these tunnle and bridge tests here and see if they already cover what you did in your new test and if they already cover it please remove the new one? for data style tests like this we sh oudl always use gurka, #3267

SiarheiFedartsou: So, yeah, here we have tests which actually test the same thing. Removing one I added in that PR.

valhalla/thor/attributes_controller.h
SiarheiFedartsou: Could you tell how does it actually work? I couldn't find usages of this const/value and just renamed for consistency.

SiarheiFedartsou: found it https://github.com/valhalla/valhalla/pull/3261/files#diff-fd61a9dafa6119b300d3fcc49f59456388c00d6851fb8cc325141be44daf58c4R819

kevinkreiser: yeah sorry, that string there is part of the public json request api, you have to continue to support in the request when someone says they want "edge.tagged_names". the variable name doesnt matter but the string matchign in request parsing does. you should keep your new string as well and we'll just treat the old string as deprecated but supported. basically this needs to still support the old string as well as the new one: https://github.com/valhalla/valhalla/blob/master/src/worker.cc#L888-L895

SiarheiFedartsou: done it this way:
https://github.com/valhalla/valhalla/pull/3261/files#diff-dda1eacb6be34d7946468f09844f48890c1266509f634990ec27263271bd6f48R893-R899

wdyt?

SiarheiFedartsou: https://github.com/valhalla/valhalla/blob/79b7c59521817c32b1012af5886a691cbdf1736e/src/worker.cc#L894-L899

kevinkreiser:
interesting i missed the original point of this issue, which was about edge candidate search. i would suggest that in this pr we only ship the data and that we make the changes to loki in a separate pr, is that ok with you @SiarheiFedartsou ?

SiarheiFedartsou:
@kevinkreiser exactly, I already have a local branch with changes related to Loki - I hope will make a PR next week.

kevinkreiser:
@SiarheiFedartsou sounds good. my opinion is its not clear how to prioritize an edge that is farther away from the input location but has the layer that was passed in. how far away from the input location should the edge be before it doesnt matter that the layer matches

kevinkreiser:
also, ill wait for the PR but, we should break that filtering out into another function so that we can do other similar fuzzy matching like name matching for example. ive always wanted to add that but still lingering is how to consider a matching name that is much further away from the input location than other matching edges. i believe we will need to discuss design before we can ship anything so as to not paint ourselves into a corner functionality-wise

SiarheiFedartsou:
> @SiarheiFedartsou sounds good. my opinion is its not clear how to prioritize an edge that is farther away from the input location but has the layer that was passed in. how far away from the input location should the edge be before it doesnt matter that the layer matches

Well, my initial implementation is just a POC which filters edges out in the same way as we do with heading in loki(if z index does not match - just filter it out). Also I thought about smth like scoring model which would take into account all criteria's at the same time(distance from location, difference in z level, bearing etc) and then we would sort candidates by this score... But it means much more radical changes to loki.
Current use case why I'm working on it: our on-board online map matcher may know current z level of user's position on graph(because it uses the same tiles for map matching as we use for routing on backend) and in case if we go off-route on 2 level road we may provide z level to Valhalla to build route from proper level of road. I think for this use case such approach would work in 99.9% cases(0.1% is for the case when tile versions used on backend and onboard are different and z level values were changed somehow in new version, but it most likely mean new road level was built :) ). 
Anyway I understand your concerns and will try to think how can we handle it too, but I think even such simple approach is really useful too(of course, with ability to turn it off). What do you think? I'll create draft PR with implementation a bit later - will be easier to discuss the idea there.

kevinkreiser:
on first thought i think a very specific use-case such as this and an implementation to optionally filter like heading for just this use-case would be quite a small change and less risky. basically the user knows they will get crazy results if they are lying about the layer they send us. but there is a problem that comes to mind. the heading filter works as a final pass filter after results have been found. for this to work for your use case you must provide a radius so that you will get more than one edge/opp edge pair. other than that small detail i think making it work like heading filter for the time being is a great/safe approach that doesnt require any additional design concerns :+1: 

when i first wrote loki i had imagined scoring based on all kinds of different properties but it turned out to be quite a difficult thing to balance them all, kind of like costing in general it would need the same kind of tweaking and testing to set the default coefficients of properties contributions to the "cost" of a candidate edge. frankly i never wanted to put in the tedious work for it so everythign has been distance based for scoring and then binary filtering. which is simple but less powerful for sure

SiarheiFedartsou:
@kevinkreiser 
please see draft of implementation here https://github.com/valhalla/valhalla/pull/3270 
I tested it and it seems to work as expected, but still need to write a couple of unit tests to be more confident.

> the heading filter works as a final pass filter after results have been found. for this to work for your use case you must provide a radius so that you will get more than one edge/opp edge pair, you want to make sure to get all the levels near by.

Tbh it is not clear enough for me, could you please elaborate here? As I can see with heading filter [we do the following](https://github.com/valhalla/valhalla/blob/c9ddd4bfa631b7dc44939bc35bde8017159712c2/src/loki/search.cc#L309):
1. if candidate should be filtered add it to `filtered` array
2. if candidate shouldn't be filterer leave it in `correlated` array
3. `filtered` array then can be used in the case if we failed to find a route with `correlated` locations(tbh it is just my guess how it works, not quite sure here)
In #3270 I simply inserted `z_level_filter` nearby `heading_filter`(probably worth introducing `filter` function or smth like this) and it seems to work.

So I'm not quite sure where should we use radiuses here...

SiarheiFedartsou:
@kevinkreiser 
I got your note about radiuses. What do you think would be the best option here? Just force users to pass radius in case if they want to use z-levels?

kevinkreiser:
> Just force users to pass radius in case if they want to use z-levels?

yes. anything else would be unintuitive because of the way the rest of the request options work.

src/baldr/edgeinfo.cc
```diff
@@ -122,28 +122,37 @@ EdgeInfo::GetNamesAndTypes(bool include_tagged_names) const {
...
-        if (name.size() > 1) {
-          try {
-            auto num = static_cast<uint8_t>(name.at(0));
-            name_type_pairs.push_back({name.substr(1), num});
-          } catch (const std::logic_error&) { LOG_DEBUG("logic_error thrown for name: " + name); }
+const std::multimap<TaggedValue, std::string>& EdgeInfo::GetTags() const {
+  // we could check `tag_cache_.empty()` here, but many edges contain no tags
+  // and it would mean we traverse all names on each `GetTags` call
+  // for such edges
+  if (!tag_cache_ready_) {
```
kevinkreiser, line 129: you could still use `tag_cache_` by setting the reservation/capacity to 0 on initialization and then resetting it here to at least 1 and then checking that as a sign for having tried to get the tag/values or not. it would be one less variable to track but its probably more confusing than just having the ready_ variable

src/mjolnir/pbfgraphparser.cc
```diff
@@ -188,6 +188,11 @@ struct graph_callback : public OSMPBF::Callback {
       }
     };
 
+    tag_handlers_["layer"] = [this]() {
+      auto z_level = static_cast<int8_t>(std::stoi(tag_.second));
```
kevinkreiser, line 192: do we have a try catch around all of the handlers when they get called? i didnt go check, but if we don't then we need to worry about that here. when someone puts `layer=chicken` we need to ignore it

test/graphparser.cc
```diff
@@ -331,6 +334,16 @@ void Baltimore(const std::string& config_file) {
   sequence<OSMWay> ways(ways_file, false);
   ways.sort(way_predicate);
 
+  // check z-level
```
kevinkreiser, line 337: did we get lucky and some of the tests just happened to have this tag in the data we were using?

test/graphtilebuilder.cc
```diff
@@ -110,7 +110,10 @@ TEST(GraphTileBuilder, TestDuplicateEdgeInfo) {
   // add edge info for node 0 to node 1
   bool added = false;
   test.AddEdgeInfo(0, GraphId(0, 2, 0), GraphId(0, 2, 1), 1234, 555, 0, 120,
-                   std::list<PointLL>{{0, 0}, {1, 1}}, {"einzelweg"}, {"1xyz tunnel"}, 0, added);
+                   std::list<PointLL>{{0, 0}, {1, 1}}, {"einzelweg"},
+                   {"1xyz tunnel",
+                    std::string(1, static_cast<char>(TaggedValue::kZLevel)) + char(-1)},
```
kevinkreiser, line 115: honestly i would shy away from updating these t ypes of tests and instead focus on gurka. the gurka tests do the entire process, not just pieces of hte process like these and they are less error prone. if it seems like gurka wont let you express the thing you want to express lets give gurka that functionality and let these old school tests alone if we can

test/isochrone.cc
kevinkreiser: yeah this is worrying to me. there is nothing in this pr that should have changed the output, it makes me think that the changes to edgeinfo somehow affect the shapes in the tiles and so affect the final output. if you disable `layer` tag handler in graphparser (so that layer tags/values to show up in the tiles), does fix this test back to the previous geometry?

test/graphparser.cc
SiarheiFedartsou: yes, it is turned out to be quite spread

test/graphtilebuilder.cc
SiarheiFedartsou: So probably tests which we have in `test_tagged_values` is enough?

kevinkreiser (changes requested):
All of the code changes look good to me, I have no changes to suggest but the change to the isochrones test values is very unexpected. I think we need to disable parts of this pr until we find what is causing that.

It would also be nice if @dgearhart and @gknisely could weigh in on the interface change. I think its much more obvious this way and its a trivial change but still they should weigh in.

test/graphtilebuilder.cc
kevinkreiser: yes, i would feel free to  revert this. i hope that one day gurka will have so much in it we can delete these tests completely

kevinkreiser: actually that would be a nice experiment. disable the mjolnir old style tests and see how much coverage drops (or even better where it drops)

src/mjolnir/pbfgraphparser.cc
SiarheiFedartsou: It seems we don't https://github.com/valhalla/valhalla/blob/c9ddd4bfa631b7dc44939bc35bde8017159712c2/src/mjolnir/pbfgraphparser.cc#L1229

But I see many other places where we use `std::stoi` in tag handlers. Should we add try/catch [here](https://github.com/valhalla/valhalla/blob/c9ddd4bfa631b7dc44939bc35bde8017159712c2/src/mjolnir/pbfgraphparser.cc#L1229)?

src/baldr/edgeinfo.cc
SiarheiFedartsou: Since we are using `std::multimap` here it means we don't have "capacity" there. We probably could use `std::unordered_multimap` here(I had some reason to use ordered version here btw), which has `reserve` but it doesn't actually have `capacity` 😄 More than it I think reservation would mean additional dynamic memory allocation which we can avoid here. WDYT?

src/mjolnir/pbfgraphparser.cc
kevinkreiser: i swore that we did a bunch of work to avoid unhandled exceptions when doing exactly this. now im wondering where the hell that code is!? is it possible that this stuff gets thrown out in lua if its not a number? to me it makes sense to put a try/catch there where yoou are suggesting. @gknisely @merkispavel where are we handling garbage tag values these days in the parser?

src/baldr/edgeinfo.cc
kevinkreiser: ah i missed that detail. i think using the ready_ variable is the easiest to understand so i think we should stick with that. i assumed that we used the `unordered` version of this datastructure so i didnt even check it. in this case, there are likely to be, if any, only a tiny number of tag/value pairs so i dont think it matters performance-wise. at most you might have tunnel with a layer or bridge with a layer but probably no more than 2 on any real way in osm

src/mjolnir/pbfgraphparser.cc
SiarheiFedartsou: Added try-catch https://github.com/valhalla/valhalla/pull/3261/files#diff-5fcdff37eb551f814e2a44f24f9c6bcc37856bb8dadbbcb59e93d173e46c1132R1234

valhalla/thor/attributes_controller.h
gknisely: @SiarheiFedartsou why are we renaming this.  @dgearhart are you ok with this?

valhalla/baldr/graphconstants.h
```diff
@@ -358,7 +358,8 @@ inline std::string to_string(Use u) {
   return i->second;
 }
 
-enum class TaggedName : uint8_t { // must start at 1 due to nulls
+enum class TaggedValue : uint8_t { // must start at 1 due to nulls
+  kZLevel = 1,
```
gknisely, line 362: Why is this kZLevel = 1?  Shouldn't it be kZLevel = static_cast<uint8_t>('3'),

valhalla/thor/attributes_controller.h
dgearhart: @gknisely I am assuming it is because z-level is not a name - I am okay with the more generic `tagged_values`

valhalla/baldr/graphconstants.h
gknisely: @SiarheiFedartsou @kevinkreiser ^

kevinkreiser: @gknisely no, it should be the integer `1`, just like bridge and tunnel should have been integers in the first place. this was the bug fix we recently did. remember that strings are just collections of bytes they dont have to be readible characters. this allows us the possibility to keep lots of different tag/value combinations. this change aligns perfectly with the fix in #3264 for more explanation please read the issue i wrote up and the comment i made on the original tunnel pr

kevinkreiser:
@SiarheiFedartsou now that @dgearhart said hes good with the name changes etc i think this is ready to merge except that we need to get to the bottom of that isochrones test value change. let me know if you need help figuring that out. @merkispavel might also be someone who could help figure it out

valhalla/thor/attributes_controller.h
kevinkreiser: the original idea was always supposed to be tags and values so that we could put any data in there we liked. this is more true to the original idea imho

test/isochrone.cc
SiarheiFedartsou: Well, I located the issue to a couple of modified lines and found that if I comment this line the issue with isochrone has gone https://github.com/valhalla/valhalla/blob/dfd7d0f9b205e682b109536b8583c685b3bca223/src/mjolnir/osmway.cc#L192

any ideas?

kevinkreiser: ok so indeed, if we add the actual tag/value to the data it changes this... does it matter what value you store there? like what if you hardcode the value on the next line to be a 1 and then hardcode it to be a 2 and try it. are the results always different depending on the value?

src/mjolnir/graphenhancer.cc
```diff
@@ -1620,7 +1620,7 @@ void enhance(const boost::property_tree::ptree& pt,
         }
 
         // Update the named flag
-        auto names = tilebuilder->edgeinfo(&directededge).GetNamesAndTypes(true);
+        auto names = tilebuilder->edgeinfo(&directededge).GetNamesAndTypes();
```
SiarheiFedartsou, line 1623: this `true` seems to be the reason of the issue with isochrones
I don't think we should consider edges with z-level to be "named", but I'm not quite confident if we should do it for edges which have tunnel or bridge names. Probably yes. So I think we can have 2 options here:
1. Simply use `false`(i.e. edge is named only if there is street name)
2. Change API of `EdgeInfo` a little bit to have convenient method to know if we have anything there which allows to call this node "named"(probably simply have `bool named() const` method there too).
WDYT? @kevinkreiser

test/isochrone.cc
SiarheiFedartsou: I found the reason, please see https://github.com/valhalla/valhalla/pull/3261/files#r689778738

src/mjolnir/graphenhancer.cc
SiarheiFedartsou: > if we have anything there which allows to call this node "named"

i.e. if we have tunnel name or street name there inside

but most likely if we have tunnel name we would have a street name too 🤔

kevinkreiser: good find! ok so i agree, that no, zlevel has nothing to do with a name and that yes, bridge and tunnel are named. 

i'm not sure i like any of the suggestions though unless im not understanding them correctly. the point of that function call is to get all names for an edge. previously we only had bridge and tunnel as special tag values and they are indeed names. now we are finally realizing the point of the tag value feature and adding a non name. what i would suggest is to modify that function to return only names. we can make a `const unordered_set<TaggedValue> name_tags{ kBridge, kTunnel};` which we can use in that function to look up which tagged values should be included in this return.

kevinkreiser: https://github.com/valhalla/valhalla/pull/3261/files#diff-9be7fe3b13039175d64773b5e79235508b5875285151c60f376bef02f8efbfe1R106

we could add to this `if` a check if the tag enum is present in the `name_tags` set and only if it is do we push it back

dgearhart: @gknisely does this make sense with the phoneme work?

SiarheiFedartsou:
@kevinkreiser 
I fixed the issue with name/non-name tags via introduction of `IsNameTag` function:
https://github.com/valhalla/valhalla/blob/c2a979416f887c4d0d836c9a9dd86200aa73c7bd/src/baldr/edgeinfo.cc#L11

Could you please review when you have a chance?

src/baldr/edgeinfo.cc
```diff
@@ -70,17 +77,21 @@ NameInfo EdgeInfo::GetNameInfo(uint8_t index) const {
...
+    if ((only_tagged_values && !ni->tagged_) || (!only_tagged_values && ni->tagged_)) {
       continue;
     }
     if (ni->name_offset_ < names_list_length_) {
-      names.push_back(names_list_ + ni->name_offset_);
+      std::string name = names_list_ + ni->name_offset_;
+      if (ni->tagged_ && !IsNameTag(name[0])) {
+        continue;
+      }
+      names.emplace_back(std::move(name));
```
SiarheiFedartsou, line 94: this function returns "tagged" names with tag, is it by design? I was going to fix it in this PR, but we even have tests which check this behaviour

@kevinkreiser wdyt?

valhalla/baldr/graphconstants.h
gknisely: @kevinkreiser got it.  i missed that we were writing out as strings

src/baldr/edgeinfo.cc
kevinkreiser: i would say leave it unchanged in this PR. seems like callers of this function must check what tags they see. which makes me think that maybe the fix for the encoding might have broken those downstream callers of this... we should check who calls this and make an issue to fix it if it doesn't look robust to handling the first char. cc @dgearhart

scripts/needs_ci_run
```diff
@@ -99,14 +103,20 @@ print(f'last build successful?, gitsha: {status}, {last_pr_gitsha}')
 if not status:
     sys.exit(0)
 
+# if commit does not exist anymore, dont skip
+if subprocess.run(['git', 'cat-file', '-t', 'last_pr_gitsha'],
+                  capture_output=True,
+                  text=True).stdout != 'commit':
+    sys.exit(0)
```
kevinkreiser, line 110: wow you decided to also fix this as well? i remember recently mentioning this to @mandeepsandhu on a certain other PR but i forget if we made an issue for it

SiarheiFedartsou: I faced this issue when doing some force pushes and fixed it :)

src/baldr/edgeinfo.cc
```diff
@@ -7,6 +7,14 @@ using namespace valhalla::baldr;
 
 namespace {
 
+// should return true for any tags which we should consider "named"
+bool IsNameTag(char ch) {
+  return true;
+  // static const std::unordered_set<TaggedValue> kNameTags = {TaggedValue::kBridge,
+  //                                                           TaggedValue::kTunnel};
+  // return kNameTags.count(static_cast<TaggedValue>(static_cast<uint8_t>(ch))) > 0;
```
kevinkreiser, line 15: this looked good to me, why comment it out?

src/baldr/edgeinfo.cc
SiarheiFedartsou: agree, also I splitted `GetNames(only_tagged_values)` into 2 separate functions `GetNames()` and `GetTaggedValues()` - seems to be less confusing.

src/baldr/edgeinfo.cc
SiarheiFedartsou: oops, sorry, I was experimenting locally and accidentally pushed it
will fix soon

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1226,7 +1231,12 @@ struct graph_callback : public OSMPBF::Callback {
       tag_ = kv;
       const auto it = tag_handlers_.find(tag_.first);
       if (it != tag_handlers_.end()) {
-        it->second();
+        try {
+          it->second();
+        } catch (const std::exception& ex) {
+          LOG_ERROR("Error during parsing of `" + tag_.first + "` tag: " + std::string{ex.what()});
```
kevinkreiser, line 1237: we should log the wayid and the value so that when this happens we can make the parsing robust to it or eventually support new tagging patterns. i would switch it to a WARN rather than an ERROR though.

kevinkreiser:
@SiarheiFedartsou at this point the code looks good to me. the last build though had the unit tests both failing and segfaulting. ping me back when those are working and ill put a ship on this pr

SiarheiFedartsou:
> @SiarheiFedartsou at this point the code looks good to me. the last build though had the unit tests both failing and segfaulting. ping me back when those are working and ill put a ship on this pr

@kevinkreiser tests are green now(at least on one job, but I don't expect they would be be red on other jobs).

kevinkreiser (approved):

src/baldr/edgeinfo.cc
liyinxiao: @SiarheiFedartsou  Hi, I am trying to understand this PR around z-level and why we cannot use unordered_map - could you help explain why "tags can have multiple values"? I couldn't find any such examples (Tag:key=tunnel) that you have described.. thanks a lot!

SiarheiFedartsou: @liyinxiao I meant tunnels in Valhalla tiles may have multiple names and we store all these names under the same tag name. "Tags" here is not directly mirrored by tags from OSM - it is what was generated by Mjolnir from OSM data.

liyinxiao: @SiarheiFedartsou  Very helpful, thanks!

Issue #3262, opened 2021-08-11
Fix Tagged Value Support in EdgeInfo


kevinkreiser:
back in #2608 we added the ability to put arbitrary way tag and value pairs into the tileset. unfortunately, as discribed in this comment https://github.com/valhalla/valhalla/pull/2608/files#r687122174, we made a huge mistake. we used `to_string` to turn the enum value into a string for use in the tiles text list. this is very unfortunate, see the enum supports 256 values, which may be more than we will ever need, but we should have written that number into just one byte of the string, by converting the number to a character we limit the total number of these supported extra tags to just 9 (a single digit excluding 0 for the moment). thats a huge bummer but i do think we can patch our way back to glory...

what i think we can do is:

1. renumber tunnel and bride to be the integer equivalent of the characters `1` and `2`, this lets old code continue to parse them as a character
2. change the parsing code in edgeinfo to actually not convert them from char to int, but just cast them to it directly, new code will also parse them fine this way

after that i think we can get back the full usage of all the possible values for tags that we want to support. whew crisis averted...

SiarheiFedartsou:
It seems I cannot finish https://github.com/valhalla/valhalla/pull/3261 without fixing this issue? @kevinkreiser what do you think if I try to fix it in that PR? I think we can go this way:
> renumber tunnel and bride to be the integer equivalent of the characters 1 and 2, this lets old code continue to parse them as a character

kevinkreiser:
Can you do that in another pr, I wanted to last night but didn't find the time. If you can wait a few hours I can do then as well. The important thing is testing old code with new data and new code with old data to verify the change

SiarheiFedartsou:
Waiting for CI :) https://github.com/valhalla/valhalla/issues/3262

PR #3264, opened 2021-08-12
Fix Tagged Value Support in EdgeInfo

SiarheiFedartsou:
# Issue

fixes https://github.com/valhalla/valhalla/issues/3262

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/mjolnir/osmway.cc
```diff
@@ -173,13 +173,13 @@ OSMWay::GetNames(const std::string& ref, const UniqueNames& name_offset_map, uin
 std::vector<std::string> OSMWay::GetTaggedNames(const UniqueNames& name_offset_map) const {
 
   std::vector<std::string> names;
-  std::vector<std::string> tokens;
 
+  auto encode_tag = [](TaggedName tag) { return std::string(1, static_cast<char>(tag)); };
```
kevinkreiser, line 177: im not sure if we should use `char` i think it should be like `std::string::CharT`, i think its platform dependent i guess is my point

kevinkreiser (approved):
this looks good to me, can you check the two combinations of:

old data, new code
new data, old code

SiarheiFedartsou:
> this looks good to me, can you check the two combinations of:
> 
> old data, new code
> new data, old code

Do you know the fastest way to do it? Maybe we have something convenient in tests? I can do everything manually, of course, but just trying to save some time 😄

gknisely:
> yep i would just build tiles manually. @gknisely or @dgearhart can you help with this by pointing to a place that has this data but also mentioning where we see it in the response? basically find a place with a bridge or tunnel name and download an extract for there. then you can make the data with old code and make the data with new code. then you can test the data with old code and test the data with new code and make sure the bridge/tunnel names come out

You can extract data around this tunnel  https://www.openstreetmap.org/way/4394511.  It should be returned via an OSRM response.  https://github.com/valhalla/valhalla/blob/master/src/tyr/route_serializer_osrm.cc#L578

SiarheiFedartsou:
> > yep i would just build tiles manually. @gknisely or @dgearhart can you help with this by pointing to a place that has this data but also mentioning where we see it in the response? basically find a place with a bridge or tunnel name and download an extract for there. then you can make the data with old code and make the data with new code. then you can test the data with old code and test the data with new code and make sure the bridge/tunnel names come out
> 
> You can extract data around this tunnel https://www.openstreetmap.org/way/4394511. It should be returned via an OSRM response. https://github.com/valhalla/valhalla/blob/master/src/tyr/route_serializer_osrm.cc#L578

Well, I extracted osm-file via https://www.openstreetmap.org/export#map=16/51.1933/6.7963. How do I convert it to pbf which mjolnir can work with?

gknisely:
> Well, I extracted osm-file via https://www.openstreetmap.org/export#map=16/51.1933/6.7963. How do I convert it to pbf which mjolnir can work with?

@SiarheiFedartsou you should be able to follow the steps here https://github.com/valhalla/valhalla#running

SiarheiFedartsou:
> > Well, I extracted osm-file via https://www.openstreetmap.org/export#map=16/51.1933/6.7963. How do I convert it to pbf which mjolnir can work with?
> 
> @SiarheiFedartsou you should be able to follow the steps here https://github.com/valhalla/valhalla#running
> 
> Convert the data from osm to pbf via `osmconvert ./your osm file -o=output.pbf`

Thanks, it works.

I started from comparison of tile sets generate by new and old code. It is 100% the same:
```
➜  valhalla git:(sf-fix-3262) ✗ find -s tiles_master -type f -name *.gph -exec md5sum {} \;
b61988c8aaa658368f58fb10fad5cb1a  tiles_master/0/003/196.gph
b77dc19d938d72f842d1ba8d494c4a63  tiles_master/1/050/946.gph
017297c49f9b58d5fb16d81771c1bd02  tiles_master/2/000/812/907.gph
➜  valhalla git:(sf-fix-3262) ✗ find -s tiles_fix -type f -name *.gph -exec md5sum {} \;
b61988c8aaa658368f58fb10fad5cb1a  tiles_fix/0/003/196.gph
b77dc19d938d72f842d1ba8d494c4a63  tiles_fix/1/050/946.gph
017297c49f9b58d5fb16d81771c1bd02  tiles_fix/2/000/812/907.gph
```
Will finish test tomorrow.

kevinkreiser:
excellent thought! now all you have to do is prove that the new code works, which, i guess it does, we probably already have units tests for that from the original pr. if that is the case id say this is already ready

SiarheiFedartsou:
> excellent thought! now all you have to do is prove that the new code works, which, i guess it does, we probably already have units tests for that from the original pr. if that is the case id say this is already ready

I couldn't find end-to-end test for tunnel names in that PR, that's why I added test which requests route near tunnel and checks if we actually respond with tunnel name field:
https://github.com/valhalla/valhalla/blob/bdc2283e240fa9132de9ae2d974421dd63b8ac47/test/thor_worker.cc#L21

It seems we can be confident everything would work here as expected.

valhalla/baldr/graphconstants.h
```diff
@@ -359,8 +359,10 @@ inline std::string to_string(Use u) {
 }
 
 enum class TaggedName : uint8_t { // must start at 1 due to nulls
-  kTunnel = 1,
-  kBridge = 2
+  // we used to have bug when we encoded 1 and 2 as their ASCII codes, but not actual 1 and 2 values
+  // see https://github.com/valhalla/valhalla/issues/3262
+  kTunnel = static_cast<uint8_t>('1'),
+  kBridge = static_cast<uint8_t>('2'),
```
averkhaturau, line 365: looks like it duplicates declaration proto/tripcommon.proto

averkhaturau (approved):

mskurydin (approved):

SiarheiFedartsou:
@kevinkreiser 
I asked @mskurydin and @averkhaturau to approve to be able to merge it now to continue work on #3261, because your approve was removed due to commit with added test. I'm merging now, but if you find something important here - let me know and I'll fix asap.

test/CMakeLists.txt
```diff
@@ -236,6 +236,15 @@ add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/test/data/nh_ar_tiles/2/000/765/07
...
 
+add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/test/data/universitatstunnel/2/000/812/907.gph
+  COMMAND ${CMAKE_COMMAND} -E make_directory test/data/universitatstunnel/
+  COMMAND ${CMAKE_BINARY_DIR}/valhalla_build_tiles
+      --inline-config '{"mjolnir":{"id_table_size":1000,"tile_dir":"test/data/universitatstunnel","timezone":"test/data/tz.sqlite","hierarchy":true,"shortcuts":true,"concurrency":1,"logging":{"type":""}}}'
+      ${VALHALLA_SOURCE_DIR}/test/data/universitatstunnel.osm.pbf
+  COMMENT "Building Universitatstunnel Tiles..."
+  WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+add_custom_target(de_universitatstunnel_tiles DEPENDS ${CMAKE_BINARY_DIR}/test/data/universitatstunnel/2/000/812/907.gph)
+set_target_properties(de_universitatstunnel_tiles PROPERTIES FOLDER "Tests")
```
kevinkreiser, line 247: unfortunately this wasnt the right way to test this, it should have been done with a gurka test, we'll want to port this. @gknisely do you or anyone else have spare cycles?

PR #3270, opened 2021-08-15
Add preferred_layer as a parameter of loki requests

SiarheiFedartsou:
# Issue
The idea is to give clients ability to provide current Z-level(if they know it) and use it as a hint for loki where it should snap location - it is the only way to have proper snapping in in the case of multilevel roads.

Some discussions can be found at https://github.com/valhalla/valhalla/pull/3261

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -95,6 +95,7 @@ message Location {
   optional uint32 waypoint_index = 31;
   optional SearchFilter search_filter = 32;
   optional uint32 street_side_max_distance = 33;
+  optional int32 preferred_z_level = 34;
 }
 
 message TransitEgressInfo {
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -95,7 +95,7 @@ message Location {
   optional uint32 waypoint_index = 31;
   optional SearchFilter search_filter = 32;
   optional uint32 street_side_max_distance = 33;
-  optional int32 preferred_z_level = 34;
+  optional int32 preferred_layer = 34;
 }
 
 message TransitEgressInfo {
@@ -156,7 +156,7 @@ enum RoadClass {
 
 message TaggedValue {
   enum Type {
-    kZLevel = 1;
+    kLayer = 1;
     // we used to have bug when we encoded 1 and 2 as their ASCII codes, but not actual 1 and 2 values
     // see https://github.com/valhalla/valhalla/issues/3262
     kTunnel = 49; // static_cast<uint8_t>('1')
--- a/src/loki/search.cc
+++ b/src/loki/search.cc
@@ -304,7 +304,8 @@ struct bin_handler_t {
         // get some info about this edge and the opposing
         GraphId id = tile->id();
         id.set_id(node->edge_index() + (edge - start_edge));
-        const auto& info = *candidate.edge_info;
+        auto info = tile->edgeinfo(edge);
+
         // calculate the heading of the snapped point to the shape for use in heading filter
         size_t index = edge->forward() ? 0 : info.shape().size() - 2;
         float angle =
--- a/src/loki/search.cc
+++ b/src/loki/search.cc
@@ -81,6 +81,15 @@ bool heading_filter(const Location& location, float angle) {
          location.heading_tolerance_;
 }
 
+bool z_level_filter(const Location& location, int8_t z_level) {
+  // no z-level - we do not filter
+  if (!location.preferred_z_level_) {
+    return false;
+  }
+
+  return *location.preferred_z_level_ != z_level;
+}
+
 PathLocation::SideOfStreet flip_side(const PathLocation::SideOfStreet side) {
   if (side != PathLocation::SideOfStreet::NONE) {
     return side == PathLocation::SideOfStreet::LEFT ? PathLocation::SideOfStreet::RIGHT
@@ -295,18 +304,19 @@ struct bin_handler_t {
         // get some info about this edge and the opposing
         GraphId id = tile->id();
         id.set_id(node->edge_index() + (edge - start_edge));
-        auto info = tile->edgeinfo(edge);
+        const auto& info = *candidate.edge_info;
         // calculate the heading of the snapped point to the shape for use in heading filter
         size_t index = edge->forward() ? 0 : info.shape().size() - 2;
         float angle =
             tangent_angle(index, candidate.point, info.shape(),
                           GetOffsetForHeading(edge->classification(), edge->use()), edge->forward());
+        auto z_level = info.z_level();
         // do we want this edge
         if (costing->Allowed(edge, tile, kDisallowShortcut)) {
           auto reach = get_reach(id, edge);
           PathLocation::PathEdge
               path_edge{id, 0, node_ll, distance, PathLocation::NONE, reach.outbound, reach.inbound};
-          if (heading_filter(location, angle)) {
+          if (heading_filter(location, angle) || z_level_filter(location, z_level)) {
             filtered.emplace_back(std::move(path_edge));
           } else if (correlated_edges.insert(path_edge.id).second) {
             correlated.edges.push_back(std::move(path_edge));
@@ -330,7 +340,8 @@ struct bin_handler_t {
                                            reach.outbound,
                                            reach.inbound};
           // angle is 180 degrees opposite direction of the one above
-          if (heading_filter(location, std::fmod(angle + 180.f, 360.f))) {
+          if (heading_filter(location, std::fmod(angle + 180.f, 360.f)) ||
+              z_level_filter(location, z_level)) {
             filtered.emplace_back(std::move(path_edge));
           } else if (correlated_edges.insert(path_edge.id).second) {
             correlated.edges.push_back(std::move(path_edge));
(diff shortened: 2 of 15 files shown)


proto/tripcommon.proto
```diff
@@ -95,6 +95,8 @@ message Location {
   optional uint32 waypoint_index = 31;
   optional SearchFilter search_filter = 32;
   optional uint32 street_side_max_distance = 33;
+
```
kevinkreiser, line 98: nit: why the space?

src/loki/search.cc
```diff
@@ -301,12 +311,13 @@ struct bin_handler_t {
         float angle =
             tangent_angle(index, candidate.point, info.shape(),
                           GetOffsetForHeading(edge->classification(), edge->use()), edge->forward());
+        auto z_level = candidate.edge_info->z_level();
```
kevinkreiser, line 314: why does this use `candidate.edge_info` and not `info` as is used above? should everything be using `candidate.edge_info`?

kevinkreiser (dismissed):
these look good to me! lets get the other data pr merged to master first then point this pr at master and merge it separately. please add another entry to the change log for the request feature adding in this pr

thanks for all of this awesome work!

src/loki/search.cc
SiarheiFedartsou: Yeah, I think we should be using `candidate.edge_info` everywhere - it allows to get rid of additional call of `GraphTile::edgeinfo` we have here.

SiarheiFedartsou:
@kevinkreiser 
I'm done with this PR - I'm setting "ready for review". I only want to ask small question about this "radius" parameter. I was experimenting with different parameters passed to loki. I set `radius` to 1 and asked Valhalla to build route from point in the middle between A and E(i.e. there is definitely no any nodes/edges near by) and G in this gurka test(basically the same setup as we have in tests in this PR):
```
 constexpr double gridsize_metres = 100;

    ascii_map = R"(
                          A
                          H
                          |
                          |
                          B-------E        
                          |       |
                          |       |
                          I       |
                          C       |
                          |       |
                          |       |
                          D-------F
                          |
                          |
                          G
    )";

    const gurka::ways ways = {{"AB", {{"highway", "motorway"}}},
                              {"BE", {{"highway", "motorway"}}},
                              {"EF", {{"highway", "motorway"}}},
                              {"FD", {{"highway", "motorway"}}},
                              {"HI", {{"highway", "motorway"}, {"layer", "-1"}}},
                              {"ID", {{"highway", "motorway"}, {"layer", "-1"}}},
                              {"DG", {{"highway", "motorway"}}}

    };
```
My expectation was that since we set this radius to 1 Valhalla would say smth like "cannot find edge to snap to, failed", but it was able to snap somewhere and build route. This behaviour is that I'm totally okay with. But just wondering why does it work? Do we maybe have writeup somewhere on how actually this `radius` parameter works?

SiarheiFedartsou:
And one more small question: do I understand it correctly that heading/heading_tolerance have no sense too without this radius parameter?

SiarheiFedartsou:
@kevinkreiser 
One more question: what is the difference between `radius` and `search_cutoff`?

src/loki/search.cc
SiarheiFedartsou: Well, actually it should be `GraphTile::edgeinfo` - we iterate here edge by edge and `candidate.edge_info` would be wrong here - please, forget my previous comment 😄

kevinkreiser:
@SiarheiFedartsou loki wants to return you a result. radius says i will give you candidates within the radius, if there are none then i will give you the single closest thing up to the search cutoff, if the nearest thing is further than that i will give nothing, its a cutoff. the result it returns are both directions of a directed edge, meaning it and its opposing edge count as "one" result (but at two path edges). for the heading filter this usually means that you will filter out of the directions of the edge but depending on tolerance different things can happen. the reason why this is slightly different that zlevel is because every edge has a heading :smile: but not every edge has a zlevel set

kevinkreiser:
i will review this at the end of the day

SiarheiFedartsou:
> @SiarheiFedartsou loki wants to return you a result. radius says i will give you candidates within the radius, if there are none then i will give you the single closest thing up to the search cutoff, if the nearest thing is further than that i will give nothing, its a cutoff. the result it returns are both directions of a directed edge, meaning it and its opposing edge count as "one" result (but at two path edges). for the heading filter this usually means that you will filter out of the directions of the edge but depending on tolerance different things can happen. the reason why this is slightly different that zlevel is because every edge has a heading 😄 but not every edge has a zlevel set

Thanks, got it. So at the moment in our production code we have `radius` hardcoded to be always 0. If I set `radius` to some relatively small value(like 10 meters) what side effects can I expect in comparison to behaviour with `radius = 0`? I'm thinking about our original use case(build of new route in case if we know current Z-level) and it seems we could just utilize this `preferred_z_level` from this PR and set `radius` to some non-zero value, but I'm a bit scary that it may introduce some unwanted side effects... What do you think?

kevinkreiser:
the way that we penalize edge candidates is based on distance. this means that in most circumstances it will still work as if you put radius=0, unless the layer selection filters the closest one to the filtered list. so i think a small radius is a good way to go

kevinkreiser:
@SiarheiFedartsou i will review this shortly, sorry for the delay here

proto/tripcommon.proto
```diff
@@ -95,6 +95,7 @@ message Location {
   optional uint32 waypoint_index = 31;
   optional SearchFilter search_filter = 32;
   optional uint32 street_side_max_distance = 33;
+  optional int32 preferred_z_level = 34;
```
kevinkreiser, line 98: any chance we could call this `preferred_layer` so that it closely matches the osm tagging name?

kevinkreiser: i guess we probably should have named it layer in edgeinfo and TaggedValue as well then?

test/gurka/test_multi_level_loki.cc
```diff
@@ -0,0 +1,133 @@
+#include "gurka.h"
+#include "rapidjson/document.h"
```
kevinkreiser, line 2: dont include this, include baldrs rapidjson utils header

test/gurka/test_multi_level_loki.cc
```diff
@@ -0,0 +1,133 @@
...
+  midgard::PointLL ll;
+  boost::optional<int8_t> preferred_z_level;
+};
+
+std::string ToString(const rapidjson::Document& doc) {
+  rapidjson::StringBuffer sb;
+  rapidjson::Writer<rapidjson::StringBuffer> writer(sb);
+  doc.Accept(writer);
+  return sb.GetString();
+}
```
kevinkreiser, line 25: this exists in the header i suggested above

test/gurka/test_multi_level_loki.cc
```diff
@@ -0,0 +1,133 @@
...
+
+std::vector<std::string> GetStepNames(const rapidjson::Document& doc) {
+  std::vector<std::string> names;
+  auto steps = doc["routes"][0]["legs"][0]["steps"].GetArray();
+  names.reserve(steps.Size());
+  for (const auto& step : steps) {
+    names.emplace_back(step["name"].GetString());
+  }
+  return names;
+}
```
kevinkreiser, line 35: i swore we had a gurka "expect" helper function for this..

test/gurka/test_multi_level_loki.cc
```diff
@@ -0,0 +1,133 @@
...
+  for (const auto& waypoint : waypoints) {
+    rapidjson::Value p(rapidjson::kObjectType);
+    p.AddMember("lon", waypoint.ll.lng(), allocator);
+    p.AddMember("lat", waypoint.ll.lat(), allocator);
+    if (waypoint.preferred_z_level) {
+      p.AddMember("preferred_z_level", *waypoint.preferred_z_level, allocator);
+      // `preferred_z_level` is only working correctly if radius is provided,
+      // because Z-level filtering is performed as last step in loki and if we don't provide
+      // radius we would have only single candidate on the last step in this case
+      p.AddMember("radius", 1, allocator);
```
kevinkreiser, line 51: we support doing this already in the basic gurka functions, you can specify options  by their json paths in an unordered_map. it looks like this:

https://github.com/valhalla/valhalla/blob/master/test/gurka/test_via_waypoints.cc#L77-L81

it uses slashes to do the hierarchical json location, using integers to specify the indices of the array elements

proto/tripcommon.proto
SiarheiFedartsou: It is actually `z_level` there:
https://github.com/valhalla/valhalla/blob/9500f40838f30a2cd8ddc81c19a9ffdc6b744591/valhalla/baldr/edgeinfo.h#L195

We can do as you wish:
1. Rename all of them to `layer`(I don't think it should be even called as breaking change since we merged it 2 days ago)
2. Rename only this field
3. Leave it as is

WDYT?

SiarheiFedartsou: I addressed all other review notes, let me know your decision here and I'll finish it.

kevinkreiser: i would love to rename all of them, option 1

SiarheiFedartsou: Okay, will do 👍

SiarheiFedartsou: Done, please take a look one more time @kevinkreiser

kevinkreiser (approved):
this looks good to me once the builds pass. i saw that windows failed with an error that we see from time to time about it cant find the right compiler. we can ignore this

PR #3273, opened 2021-08-16
Support routing onto HOV lanes

ktatterso:
# Definitions

An HOV2 lane requires the vehicle to have two occupants. 
An HOV3 lane requires the vehicle to have three occupants.
A HOT lane is an HOV lane that optionally allows the driver to pay a toll fee if they don't have the required number of occupants (high occupancy toll).

When I say "HOV lane" I mean a separately digitized lane that is "HOV only".  I am not referring to roadways that are singly digitized and have mixed-usage, e.g., the leftmost lane is HOV but the rightmost lanes are non-HOV.

Here is an example of an HOT3 lane:  https://www.openstreetmap.org/way/815145148#map=19/39.97138/-105.20052

# Issue

Prior to these changes, only the `HOVCost` profile could route onto HOV (high occupancy vehicle) lanes.

With these changes, the caller can specify the types of HOV lanes they'd like to consider for routing.  These boolean options will now be honored for the `AutoCost` profile and ignored by other profiles:
`use_hov2`
`use_hov3`
`use_hot`

# Research

I've spent a fair number of hours going between [Overpass-Turbo](https://overpass-turbo.eu/), [OpenStreetMap](https://www.openstreetmap.org/), [TagInfo](https://taginfo.openstreetmap.org/), and Google Street view.  Here are my conclusions:

HOV lanes are largely found in the US with only a few examples outside.  Within OSM, the most correct way to tag an HOV lane is with these two tags:  `hov=designated` and `hov:minimum=n`, where `n` is either 2 or 3.

A fair number of HOV lanes are tagged `oneway=alternating` or `oneway=reversible`.  For example:  https://www.openstreetmap.org/way/66339393#map=18/39.80735/-104.98314 .  I live near this area and this HOV road is usually open to traffic going into Denver in the morning and changes direction in the evening for traffic leaving Denver.  However, it can randomly be closed for unknown reasons.  For this project, I've decided to avoid all HOV roads that are access conditional, time conditional, and/or alternating direction.

In Europe I see a single HOV2 roadway:  https://www.openstreetmap.org/way/198495331.  There are many other usages of the `hov=designated` tag, but the `hov:minimum` tag is not present.  These fall into two categories:  [bus lanes](https://www.openstreetmap.org/way/406633433#map=19/40.62607/-4.00435) and [parking areas](https://www.openstreetmap.org/way/399190171).

There are a few ways that are tagged `hov:minimum=4`, e.g., https://www.openstreetmap.org/way/882722563.  The preceding and following ways are tagged `hov:minimum=2`, so I believe the tagging is incorrect.

I also found a series of ways tagged `hov:minimum=6` in the Vancouver area (but do not have `hov=designated`).  This appears to be a [bus lane](https://www.openstreetmap.org/way/191677518) on the rightmost lane.

In conclusion, the vast majority of HOV lanes for auto's are HOV2/HOV3 and are in the US.  Some of these are also tolled (with `toll=yes`) making them HOT lanes.

# Concept

The guiding principle is we do not want to route onto an HOV lane unless the user elects to do so.  This is because driving on an HOV lane without the correct number of occupants is illegal.

Also, HOT lanes are seen as conceptually separate from 100% tolled roadways.  The thinking is that a pragmatic driver would allow for tolled roads (the default), but would avoid HOT lanes by default.

# Design/Implementation

~~From the data side, we need to read/store enough information to identify non-HOV, HOV2, and HOV3 ways.  Three states requires 2-bits, so you'll see 2-bits being moved through `OSMWay` and into `DirectedEdge`.~~

Update:  After much discussion and experimentation, I was able to modify the implementation to only consume one-bit within the `DirectedEdge`.  Here's how it works: 

We can determine if a way/edge is hov-only by examining the access-mask.  An hov-only edge will have the `kHOVAccess` bit set, but the `kAutoAccess` bit will not be set.  There is a new method that does this determination:   `DirectedEdge::is_hov_only()`.

If `DirectedEdge::is_hov_only()` returns `true`, the coder can call `DirectedEdge::hov_type()` to determine if the edge is HOV2 or HOV3.   

The trick that allows auto's to route onto HOV is to construct the `AutoCost` as follows:
`AutoCost(const CostingOptions& costing_options, uint32_t access_mask = (kAutoAccess | kHOVAccess));`

Note the `| kHOVAccess`.  This means the auto-costing will consider HOV edges.  `AutoCost::EvaluateHOVAllowed()` makes the final determination.

From the input side, we simply capture the `use_hot`, `use_hov2`, and `use_hov3` booleans (if present).  These make their way into `DynamicCost`.

## Tasklist

 - [x] Add tests


kevinkreiser:
The main thing i wante to metioned here was that we already encode HOVness, or at least we support doing so, via the access mask on an edge. It does not differentiate between 2 and 3+ passengers but we do seem to already support this (i remembered seeing it in the code and noticed it in the access mask in graphconstants). it doesnt make sense to me to duplicate this information. can we somehow make use of the access mask that already exists? cc @dnesbitt61

dnesbitt61:
The access flag allows the general case of HOV access. It does not distinguish between types of HOV access though, so adding info to DirectedEdge to distinguish HOV types seems fine. Deprecating HOV costing and adding as a parameter for auto costing also makes sense to me. It seems to me the HOV access flag would not be required anymore once fully deprecated.

I think that usually motorcycles are allowed on HOV. Do any of these changes impact motorcycle costing?

kevinkreiser:
i guess one concern is the number of spare bits remaining in directed edge. its already quite low. there is a spare bit in the access mask which we could use to mean hov3. or we could use just 1 spare bit of directed edge to mean hov3 and use it in conjunction with the mask bit. it seems wasteful to take 2 bits when we have some of the information already stored in 1 bit of the access mask.

ktatterso:
@dnesbitt61 You are correct that some HOV lanes do allow motorcycles.  I think buses might eventually be considered.  And emergency vehicles.  I scoped this project to the auto-costing since it is the most-used costing - and I wanted to think very carefully about the rules for autos.  However, I think this framework could work for any costing, since the model is based on the user "asserting" which types of HOV lanes they would like to use for routing.

gknisely:
@dnesbitt61 @kevinkreiser so are we fine with using the 2 bits in the DE or not.  I am feeling like there is not a concern here, but not sure.

kevinkreiser:
@gknisely if we use 2 bits in the directed edge then we should rename the HOV access mask to `RESERVED` or something and also remove the bits in the code that set it. we shouldnt build redudancy into the code base. the other option that doesnt require taking spare bits from the directed edge is to use the spare mask value for for HOV3 and the already used one to mean HOV2.

if we go the route of deprecating the access mask HOV then we essentially used 3 bits to accomplish this task because depracating it means we are never allowed to use it again. if we use the single spare access mask bit then it only cost us 1 extra bit for this feature. or if we use the access mask HOV in conjunction with just 1 bit from the directed edge then again it only cost us 1 bit for this feature. the reason i think its relevant is because we only have 7 bits left in total. if we have the option to use only 1 bit to ship this feature instead of 3 then i think we should take that option. the reason i think this is imporant is because i think we want to avoid having to make a new tile version for as long as possible. its true we wouldnt have to, we could use the extended directed edge struct but that will add huge bloat to the data.

my personal suggestion would be to keep the HOV mask meaning that some level of HOV is supported and then use 1 bit on the directed edge to differentiate between HOV2 and HOV3+. its both frugal with the bits and also semantically pretty clear.

gknisely:
> @gknisely if we use 2 bits in the directed edge then we should rename the HOV access mask to `RESERVED` or something and also remove the bits in the code that set it. we shouldnt build redudancy into the code base. the other option that doesnt require taking spare bits from the directed edge is to use the spare mask value for for HOV3 and the already used one to mean HOV2.
> 
> if we go the route of deprecating the access mask HOV then we essentially used 3 bits to accomplish this task because depracating it means we are never allowed to use it again. if we use the single spare access mask bit then it only cost us 1 extra bit for this feature. or if we use the access mask HOV in conjunction with just 1 bit from the directed edge then again it only cost us 1 bit for this feature. the reason i think its relevant is because we only have 7 bits left in total. if we have the option to use only 1 bit to ship this feature instead of 3 then i think we should take that option. the reason i think this is imporant is because i think we want to avoid having to make a new tile version for as long as possible. its true we wouldnt have to, we could use the extended directed edge struct but that will add huge bloat to the data.
> 
> my personal suggestion would be to keep the HOV mask meaning that some level of HOV is supported and then use 1 bit on the directed edge to differentiate between HOV2 and HOV3+. its both frugal with the bits and also semantically pretty clear.

@ktatterso bit twiddling time?  ^

dnesbitt61:
Isn't there a 3rd value for HOT (toll access for HOV lanes)? This would argue for using 2 bits in DirectedEdge and deprecating the HOV access flag (change to RESERVED so it isn't immediately re-used). Also, HOV is really a special condition for automobile travel. Access flags to me should represent different modes of travel.

kevinkreiser:
HOT == toll (which already exists) + HOV which also already exists

kevinkreiser:
seems like most folks would be cool with fully deprecating the access mask for HOV and just use 2 bits from the spare in the directed edge? maybe we can thumbs up this comment to show support for that idea?

kevinkreiser:
seems like most folks would be cool with fully deprecating the access mask for HOV and just use 2 bits from the spare in the directed edge? maybe we can thumbs up this comment to show support for that idea?

ktatterso:
@kevinkreiser @dnesbitt61 Thanks for your thoughts.  Regarding HOT lanes, as KK said, these are just HOV lanes that are also tolled.  We already have a toll boolean flag on DirectedEdge, so I didn't need to add anything to handle HOT lanes.

Regarding KK's suggestion to use one less bit, I'm fine with that.  I'm also fine using two bits.  I understand and respect the value in conserving bits.

kevinkreiser:
@ktatterso welcome to the twiddlers club!

ktatterso:
Alright, I'm going to adapt things to use the access-mask to indicate (notHOV, HOV) - and save a bit.  Stay tuned.

ktatterso:
@kevinkreiser @dnesbitt61  I'm coding up what's been discussed... using the access-mask to indicate (notHOV, HOV) and using a single bit on the `DirectedEdge` to indicate (HOV2=0, HOV3=1).  It will work but it feels... kinda greasy.  The default state of a DE is "all zeroes" (via a `memset`).  As I've coded this, the casual observer would think, "wait, why do all DE's default to HOV2?".  Of course I can add comments in various places to explain the design... but still.  (With my 2-bit approach, the default state is `kNotHOV=0`, a sensible default.)

Also, I think that using the `kHOVAccess` access-mask bit as a boolean to mean (notHOV, HOV) breaks the philosophical design intent of the access-mask.

In conclusion, we will have saved a bit taking this approach - but we might leave future readers scratching/shaking their heads.

Happy to proceed in either direction.  Thoughts?

kevinkreiser:
maybe i am missing something but i was thinking it could be mostly fixed with proper naming, more concretely i would propose the following:

1. change the api to specify how many passengers (must be 1 or more, no autonomous vehicles at this point haha)
2. rename the access mask bit to HOV2
3. name the single bit on directededge HOV3

then the code can be as simple as in costing you do:

```c++
bool autocost::allowed(...) const {
  ...
  // either you make the hov requirements via passenger count OR you are cool with paying a toll to use it
  bool hov_access = passengers_ >= edge->HOVPassengers() || (costing->allow_hov_tolls() && edge->toll());
  ...
}
```

and then in directed edge you add this helper method:

```c++
uint8_t directededge::HOVPassengers() const {
   //every road requires one driver, hov2 requires another and hov3 requires yet another
   return 1 + (edge->forward_access() & kHOV2) + edge->HOV3_;
}
```

kevinkreiser:
@dnesbitt61 i'm not sure if you are directing your comment to me or not but if you are i'm not advocating to use two access bits in this case (though i did mention that as an option previously). [here](https://github.com/valhalla/valhalla/pull/3273#issuecomment-901537568) im advocating to rename the existing hovaccess bit mask value to `kHOV2` and to use a spare bit from directed edge for `HOV3_`

dnesbitt61:
I'm sorry - I should have looked more closely at @kevinkreiser  latest comment and code fragment! I misunderstood. I do wish we could get back the HOV access bit but that must wait for a V4 data specification.

ktatterso:
@kevinkreiser I've thought about this HOV stuff a lot (over the past few weeks).  I initially had the same idea that you just proposed:  to specify the number of passengers.  After doing a bunch of research it didn't feel like the right approach.  There are a few reasons:
- as mentioned, motorcycles frequently qualify for HOV lanes even as a single driver
- in some areas, electric vehicles qualify even with a single driver
- in some areas, two-seater cars with two occupants qualify for HOV3 lanes
- buses frequently (always?) qualify regardless of passenger count
- emergency vehicles frequently (always?) qualify

To me, it would seem like an odd interface if an EV driver would have to specify s/he has 3 passengers to qualify for HOV3 routing even though they are a single driver.

Consequently, I see HOV more like this: "as the driver of this vehicle I assert I am qualified for routing on HOV2/HOV3".

Regarding the bits, even with the renaming of the access-mask to `kHOV2` it will feel a bit wonky.  To me at least.  But I'll code it up.

kevinkreiser:
@ktatterso good point on the API spec, what you said makes sense to me.

as far as the grease is concerned i think its worth the savings. i suspose i should caution you not to look to closely at the rest of the structures because there are definitely other greasy places where we've had to be less than perfect because of mistakes (the header for example). these things happen but at the end of the day we need to optimize for too many things:

1. compatibility
2. spare for future
3. tile size
4. readibilty

inevitably i think we have to leave the last one as least important and strive to wrap the grease in well named methods so that we can hit 1-3 while doing the best we can with 4

ktatterso:
@kevinkreiser Totally fair.  It's a balancing act.

And just so we're all on the same page... the access-mask bit `kHOVAccess` is going away.  Hence, we will no longer have an HOV costing...?

Instead, we will repurpose the `kHOVAccess` access-mask bit as a flag indicating (NotHOV, HOV).  I might name it `kIsHOV`?

kevinkreiser:
@ktatterso yes, that all sounds correct with the addition of the thing i said in a previous comment:

> i suggest that we should completley deprecate the hov costing by simply setting to auto cost and adding the request option to use HOV if someone specifies the deprecated hov costing in their request

meaning you delete the hov costing model but if someone asks for it in a request we give them "auto" and set HOV2 as their costing option

does that make sense? that way externally it doesnt look like we did anything but internally we got to shed a costing model.

ktatterso:
@kevinkreiser Makes sense.  Let's see what I can whip up.

ktatterso:
@kevinkreiser Ok here's what I've found.  I have gotten this to work, but I don't see a way to make this approach to work with existing data.  Existing data will have the `kHOVAccess` access-mask bit set for many non-HOV ways.  If I take that bit to mean (NotHOV=0, HOV=1), Valhalla thinks the majority of roads are HOV.  And since my code changes avoid HOV by default... lots of no-route's.

If I build the tile data, it works correctly, since I'm able to set the bit in the access-mask to correctly indicate (NotHOV=0, HOV=1).

I think it's a good design ambition to have the software work with older data.  The trick is to steal a bit from somewhere that is already always 0 in existing data...  The other bit we might steal is the last access-bit in the access-mask... 11/12 bits are in use... can you think of anywhere else we can steal a bit?

ktatterso:
@kevinkreiser One more thought.  I wondered if we should shake the dust off `DirectedEdgeExt`... but then I see it doesn't have a ctor like `DirectedEdge` that `memset`'s everything to zero... so I don't think that would work. 

In any event, we probably should add a ctor to `DirectedEdgeExt` with a `memset`.  Thoughts?

kevinkreiser:
@ktatterso why does old data make the code think that many edges are HOV? all we are doing is just changing the name of the mask value. the only way that this would change behavior is if the old code set HOV on everything or something. i didnt review the current implementation in master but i had assumed that that bit is only set when HOV tags are present on a given way. is that not the case?

kevinkreiser:
we definitely should not use DirectedEdgeExt, that will bloat the data by huge amounts. adding a constructor and zeroing the struct if that isnt there should be done though its a mute point because we dont actually use it at the moment.

kevinkreiser:
lua parsing does seem to have some trickiness. it seems to say if its tagged hov then use whatever auto access we had, that is sane, but then if we find out that its actually only some lanes of the way we keep it tagged as HOV still: https://github.com/valhalla/valhalla/blob/480cc790ea43f1da0415d85466af7697aeaac475/lua/graph.lua#L1621-L1668

this goes back to you changing the definition of what HOV means in the data. i remember you said you only want to mark stuff that is separately digitized, my question then would be are you thinking only about a commercial dataset or are you considering OSM tagging conventions? the data we have today seems to say both autoaccess and hov access and not differentiate whether its HOV only. why cant we keep this understanding and simply add resolution meaning of how one "qualifies" to use the HOV (the 2 vs 3 part)?

if im correct in why you are saying what you said above, i think we can change the meaning of HOV less and keep compatibility and use the one bit to differentiate between the qualifications of HOV. the mask is like, do we have HOV on this edge and the new bit is like do you need to qualify with HOV2 or HOV3. cant we just use that understanding? THen old data will indeed asume anything marked HOV access is HOV2 but i think this is completely fine. You want new features you need new data to go with your new code, thats how it has worked and wil always have to work

gknisely:
@ktatterso @kevinkreiser I am starting to wonder if we really want to remove the HOV access.  I was thinking about the continuous lane guidance logic and the access would be needed to determine which lane has what access.  For example, `hov:lanes = designated|||||`  means that the left most lane is HOV only or designated.  So if we got rid of the HOV access, I am not sure how I would handle these lane based restrictions.  They can exist like this too `hov:lanes:conditional : designated @ (Mo-Fr 16:00-18:00;Mo-Fr 07:00-09:00)|||||`   So can we use some bits from elsewhere?

kevinkreiser:
@gknisely that was my point exactly. i think we should keep the HOV access. with respect to more exotic tagging i could see using restrictions to model some of those but even at present we would have enough information to say "use HOV lanes" or "use non-HOV lanes" depending on what the user requested. at the moment we cant say which lanes are which but we can at least tell when a road has both HOV and non HOV access which is important imho. there is no reason we couldnt store it elsewhere but why do that when we already store it now in the mask.

i think the scope of this PR is just about differentiating between HOV2 and HOV3 and HOV+TOLL. do you think its possible to continue with this piece and save per lane differentiation for a later discussion or do we need to get them all at once because they are all too related to break into pieces?

gknisely:
> @gknisely that was my point exactly. i think we should keep the HOV access. with respect to more exotic tagging i could see using restrictions to model some of those but even at present we would have enough information to say "use HOV lanes" or "use non-HOV lanes" depending on what the user requested. at the moment we cant say which lanes are which but we can at least tell when a road has both HOV and non HOV access which is important imho. there is no reason we couldnt store it elsewhere but why do that when we already store it now in the mask.
> 
> i think the scope of this PR is just about differentiating between HOV2 and HOV3 and HOV+TOLL. do you think its possible to continue with this piece and save per lane differentiation for a later discussion or do we need to get them all at once because they are all too related to break into pieces?

@kevinkreiser yes save it for a later discussion.  If we ever get back to continuous lane guidance, I will have to handle it.

gknisely:
@kevinkreiser I have re-read this and I just want to confirm.  You still want the following correct?  
> ... ie why not use access to mean HOV access and then use a bit to specify what type of HOV 2 or 3?

I think the discussion got a bit long and there was confusion on our side from me and @ktatterso Let us know and thanks.

kevinkreiser:
Yes I would still prefer it. I don't see why it breaks compatibility and it saves a bit too. I can't understand why we wouldn't go that route, pun intended.

kevinkreiser:
@ktatterso i see now why the old data doesnt work:

https://github.com/valhalla/valhalla/blob/480cc790ea43f1da0415d85466af7697aeaac475/lua/graph.lua#L1621-L1629

i linked it before but convinced myself to look at the lanes part, but that first part is the most important it says that, as long as it doesnt say "no" to hov then we set the hov bit to match the auto access bit which basically screws it all up :smile: 

so in short the only thing we can do is just banish that bit to deprecated and indeed use 2 full bits on the directed edge.

dnesbitt61:
yea, the access bit is meant to say "Can an HOV vehicle access this edge, or this restriction". It is not meant as this is only HOV access. I will say that there may be some restriction cases where the HOV access bit is needed for restrictions. That would be a reason to keep the access bit (apart from needing to work with legacy code/data).

src/sif/autocost.cc
```diff
@@ -295,6 +317,10 @@ class AutoCost : public DynamicCost {
   virtual bool Allowed(const baldr::DirectedEdge* edge,
                        const graph_tile_ptr& tile,
                        uint16_t disallow_mask = kDisallowNone) const override {
+
+    if (!EvaluateHOVAllowed(edge))
```
gknisely, line 321: @ktatterso shouldn't we be calling `EvaluateHOVAllowed` in `AutoCost::Allowed` and `AutoCost::AllowedReverse` versus here as the comment says, "Function to be used in location searching..."  Bidirectional_astar calls it from [here](https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L260-L261) and [here](https://github.com/valhalla/valhalla/blob/master/src/thor/bidirectional_astar.cc#L267-L268)

gknisely (changes requested):

src/sif/autocost.cc
ktatterso: @gknisely Good catch.  My testing was limited so I missed this.  I've fixed the `Allowed()` and `AllowedReverse()` methods that are called by BidirAstar to check for HOV routability.  I've also added some new tests that capture this fix.

ktatterso:
Just a quick note to say I am building planet data with "correct" HOV bits and will be doing a bunch of adhoc testing - hopefully tomorrow (Fri Sept 3).

src/sif/autocost.cc
gknisely: @ktatterso can you move !EvaluateHOVAllowed(edge) to the huge if statement.  just to be consistent

ktatterso: @gknisely I can but... I hate those big conditionals... they're hard to mentally dissect and also hard to debug.  :/

ktatterso: @gknisely ok I pushed the changes you requested.  Still waiting on the data to build...

gknisely (dismissed):
:shipit: after you RAD test

ktatterso:
> :shipit: after you RAD test

Running RAD now...

lua/graph.lua
```diff
@@ -1645,6 +1646,10 @@ function filter_tags_generic(kv)
     end
 
     if only_hov_allowed then
+      -- If we get here we know the way is a true hov-lane (not mixed).
+      -- As a result, none of the following profiles can use it.
```
kevinkreiser, line 1650: nit: `s/profiles/costings`

lua/graph.lua
```diff
@@ -1618,21 +1622,18 @@ function filter_tags_generic(kv)
     kv["seasonal"] = "true"
   end
 
+  kv["hov_tag"] = "true"
```
kevinkreiser, line 1625: can you explain what is going on here. it reads like we are saying that all ways are always tagged wrt hov (ie we dont have to infer hov-ness the tag always is there to specify it).

src/baldr/directededge.cc
```diff
@@ -203,6 +203,16 @@ void DirectedEdge::set_bridge(const bool bridge) {
   bridge_ = bridge;
 }
 
+// Sets the hov type.
+void DirectedEdge::set_hov_type(const HOVEdgeType hov_type) {
+  hov_type_ = static_cast<uint32_t>(hov_type);
+}
+
+// Is this edge strictly hov? (if so, it could be hov2 or hov3, check the hov_type_)
```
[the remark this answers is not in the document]
kevinkreiser: frankly i wonder if it would matter if we didnt just change the semantics of the bit to not include autoaccess. how many people could have possibly been actually using the HOV costing and how many of those wont update their data and code soon after we release this? famous last words i know, but this is the first time i feel pretty confident that HOV was an under-used thing

src/mjolnir/pbfgraphparser.cc
```diff
@@ -589,6 +589,19 @@ struct graph_callback : public OSMPBF::Callback {
...
+      // If this tag is set then the way is either HOV-2 or HOV-3.
+      // There are no other real-world hov levels.
+      std::string hov_type = tag_.second;
+      if (hov_type == "HOV2") {
+        way_.set_hov_type(valhalla::baldr::HOVEdgeType::kHOV2);
+      } else if (hov_type == "HOV3") {
+        way_.set_hov_type(valhalla::baldr::HOVEdgeType::kHOV3);
+      } else {
+        // assert? For now just set to the max hov level.
+        way_.set_hov_type(valhalla::baldr::HOVEdgeType::kHOV3);
```
kevinkreiser, line 602: maybe warn and ignore is better?

src/sif/autocost.cc
```diff
@@ -137,7 +137,7 @@ class AutoCost : public DynamicCost {
    * Construct auto costing. Pass in cost type and costing_options using protocol buffer(pbf).
    * @param  costing_options pbf with request costing_options.
    */
-  AutoCost(const CostingOptions& costing_options, uint32_t access_mask = kAutoAccess);
+  AutoCost(const CostingOptions& costing_options, uint32_t access_mask = (kAutoAccess | kHOVAccess));
```
kevinkreiser, line 140: dont forget to completely delete HOV costing below and add the code to worker.cc to see that costing string and make the changes to the request to turn on HOV2 or somethign for the auto costing

src/baldr/directededge.cc
```diff
@@ -586,6 +596,8 @@ json::MapPtr DirectedEdge::json() const {
       {"destination_only", static_cast<bool>(dest_only_)},
       {"tunnel", static_cast<bool>(tunnel_)},
       {"bridge", static_cast<bool>(bridge_)},
+      {"is_hov", static_cast<bool>(is_hov_only())},
+      {"hov_type", to_string(static_cast<HOVEdgeType>(hov_type_))},
```
kevinkreiser, line 600: nit: non hov edges return the string "null" which is kind of odd for json. instead of doing it inline you can put an if below that says `if(is_hov_only) map.emplace("hov_type", to_string(static_cast<HOVEdgeType>(hov_type_)));` so that it only shows up when its not "null"

kevinkreiser:
This looks ready to ship to me, all that is left to do is to deprecate the HOV costing by deleting its code, and then when someone requests it and we parse it over in worker.cc, we can change the costing to auto costing, change the costing options to auto as well and then add HOV2 to the renamed costing option. that way existing requests still work but we can have one less costing model.

kevinkreiser (changes requested):
Can we remove HOV costing and emulate it with auto costing, and can we also update the api docs?

lua/graph.lua
ktatterso: Ah yes.  I remain puzzled myself the true intention behind all the `auto_tag`, `hov_tag`, etc.  I mean, throughout the lua we check them for `nil` and we set them to `true/false`... but weirdly, the `tag_handler` logic doesn't care what you set the `hov_tag` value to, only that it is set:

```
      tag_handlers_["hov_tag"] = [this]() {
      osm_access_.set_hov_tag(true);
      has_user_tags_ = true;
    };
```
Once set, there is later logic (can't recall where) that says "hey, the access bit is set for hov, I need to query the way for the `hov_forward/hov_reverse` flags".

So I figured `hov_tag` can always be set - and what is most important is that the `hov_forward/hov_reverse` flags are correct. 

But it's quite possible I'm missing the wisdom/intention behind these `*_tag`'s.

src/mjolnir/pbfgraphparser.cc
ktatterso: Good call on `LOG_WARN`.  I still think we should set it to the max hov level... to ignore would leave it at `kHOV2`.

src/sif/autocost.cc
ktatterso: @kevinkreiser I'm fuzzy on what exactly it means to "delete the HOV costing".  Like, literally?  Or leave it as a shell class that just defaults to AutoCost with HOV2?

kevinkreiser: i meant literally :smile: 

basically 2 steps:

1. delete HOV costing from autocost.cc
2. in worker.cc if someone asks for a `costing` in the request as `hov` then we set `costing` to `auto` (in the actual request json) and we rewrite `costing_options.hov` to `costing_options.auto` (again in the actual request json). and then we add to `costing_options.auto` the `hov2` flag. then we let the request parsing go on as normal.

there are a number of other deprecated features that work this way. have a look over in worker.cc and search for "deprecate" and you'll see a couple places where this happens. not specifically for costing but for other parameters we've sunset

nilsnolde: There’s even the deprecated “auto_shorter” costing in there as well IIRC

ktatterso:
@kevinkreiser I removed the HOV costing.  Let me know if I captured the spirit of what you were thinking - in remapping requests for `hov` costing to `auto` with `use_hov2=true`.

I can update the docs after this gets released.

src/worker.cc
```diff
@@ -614,6 +614,18 @@ void from_json(rapidjson::Document& doc, Options& options) {
...
+  // use_hov2=true costing option.
+  if (costing_str == "hov") {
+    costing_str = "auto";
+    rapidjson::SetValueByPointer(doc, "/costing", "auto");
+    auto json_options = rapidjson::GetValueByPointer(doc, "/costing_options/hov");
+    if (json_options) {
+      rapidjson::SetValueByPointer(doc, "/costing_options/auto", *json_options);
+    }
+    rapidjson::SetValueByPointer(doc, "/costing_options/auto/use_hov2", true);
+  }
```
kevinkreiser, line 627: perfect!

src/sif/dynamiccost.cc
```diff
@@ -452,6 +452,11 @@ void ParseBaseCostOptions(const rapidjson::Value& value,
   // closure_factor
   pbf_costing_options->set_closure_factor(base_cfg.closure_factor_(
       rapidjson::get<float>(value, "/closure_factor", base_cfg.closure_factor_.def)));
+
+  // HOT/HOV
+  pbf_costing_options->set_use_hot(rapidjson::get<bool>(value, "/use_hot", base_cfg.use_hot_));
+  pbf_costing_options->set_use_hov2(rapidjson::get<bool>(value, "/use_hov2", base_cfg.use_hov2_));
+  pbf_costing_options->set_use_hov3(rapidjson::get<bool>(value, "/use_hov3", base_cfg.use_hov3_));
```
kevinkreiser, line 459: i think we should use different nomenclature. all of the other options which are prefixed with `use_` are a sliding scale. this is a boolean option. we should name it something else to remove that confusion or unintuitiveness. maybe `include_` or something to match the include/exclude language used in other boolean flags and attribute filters?

kevinkreiser:
@ktatterso i added the docs changes for you, they are trivial and shouldnt be left out. i do think we should change the parameter names though `use_` is always a sliding scale rather than a binary. got any other suggestions than the one i made? as soon as we fix the naming we can merge in my opinion

kevinkreiser:
@ktatterso there was a lot of stuff left hanging around, remnants of hov costing so i removed them all and i added a test case to your new suite to prove the fallback for deprecated hov costing works: https://github.com/valhalla/valhalla/pull/3273/commits/24f5ef3b7692f1511bede1a385efdac003229eea

src/sif/dynamiccost.cc
kevinkreiser: maybe we should just use `"hot": true, "hov2": true, "hov3": true` and avoid the extra verbage all together?

ktatterso:
@kevinkreiser Updated the input param names to "include" instead of "use".

src/sif/autocost.cc
```diff
@@ -45,9 +45,6 @@ constexpr float kTCUnfavorable = 2.5f;
 constexpr float kTCUnfavorableSharp = 3.5f;
 constexpr float kTCReverse = 9.5f;
 
-// How much to favor hov roads.
-constexpr float kHOVFactor = 0.85f;
```
kevinkreiser, line 49: this is what we would want to do once we add something like a `use_hov` option, ie a way of favoring or unfavoring an HOV edge.

src/sif/autocost.cc
```diff
@@ -285,6 +282,28 @@ class AutoCost : public DynamicCost {
     return static_cast<uint8_t>(type_);
   }
 
+  bool IsHOVAllowed(const baldr::DirectedEdge* edge) const {
+    // A non-hov edge means hov is allowed.
+    if (!edge->is_hov_only())
+      return true;
```
kevinkreiser, line 288: this is not inlined and its in the tight loop. i suspect that there is a performance impact from this. the good news is that this first `if` is almost always hit so we dont have to hit 3 more branches below. still it would be nice to at least test performacne difference. maybe you could do it as a follow up? like we can merge the pr but you can post back your findings regarding any changes in performance?

src/sif/autocost.cc
```diff
@@ -295,10 +314,11 @@ class AutoCost : public DynamicCost {
   virtual bool Allowed(const baldr::DirectedEdge* edge,
                        const graph_tile_ptr& tile,
                        uint16_t disallow_mask = kDisallowNone) const override {
+    bool allow_hov = IsHOVAllowed(edge);
```
kevinkreiser, line 317: why do this? let the logical  expression below shortcircuit and never even call this function if it doesnt need to, ie just put the function call at the end of the return statement

src/sif/autocost.cc
```diff
@@ -854,184 +886,6 @@ cost_ptr_t CreateBusCost(const CostingOptions& costing_options) {
   return std::make_shared<BusCost>(costing_options);
 }
 
-/**
```
kevinkreiser, line 857: this is beautiful

valhalla/sif/dynamiccost.h
```diff
@@ -888,6 +888,12 @@ class DynamicCost {
   bool penalize_uturns_;
 
   bool exclude_unpaved_{false};
+
+  // HOT/HOV flags
+  bool include_hot_lanes_{false};
+  bool include_hov2_lanes_{false};
+  bool include_hov3_lanes_{false};
```
kevinkreiser, line 895: nit: seems odd to have the `lanes_` suffix here. someone just glacing at this might be confused if there is some semantic difference between these and the base options

kevinkreiser (dismissed):
this looks good to me. there were some tiny nits that you could just patch up and maybe a follow on to just at least say if it does or doesnt change the performance of path finding. if you want to commit up the nits i can reship this if you ping me.

src/sif/autocost.cc
ktatterso: My preference is to write code that is readable and debuggable.  By breaking out the `bool allow_hov` I can set a breakpoint after it is initialized to inspect its value.  The optimizer can optimize it away in release builds.  That my reasoning.  But I can change it.

src/sif/autocost.cc
ktatterso: @kevinkreiser I ran some benchmarks this morning (`make run-benchmarks` and `valhalla_run_route`).  I did a bunch of runs, did some averaging... I don't think I'm measuring a difference.  That said, I struggle to get stable perf timings  on my system (mac laptop).

ktatterso:
@kevinkreiser nits are fixed.

kevinkreiser (approved):

PR #3274, opened 2021-08-17
Intermediate Waypoints:: Shape Association

### LINKED ISSUE #3268: Include a Shape Point for Any Intermediate Locations in the Request
@kdiluca was recently asking me how we could achieve the following:

1. add a shape point to a given route leg where ever that leg had intermedia locations, ie. locations with type `via` or `through`)
2. add an index to the output location that tells  which shape index in the leg maps to the shape point representing the location

Broadly I think we can reuse existing patterns in the code to achieve this. The quick summary is, in route_action.cc we should create `edge_trimming` objects for any intermediate location (currently we only do them for vias where the route does a uturn mid-edge). While we are doing that we should also store the paths edge index where the intermediate location occurs, on the location itself. Then over in triplegbuilder, we should iterate over the intermediate locations, while we are building the leg and when we encounter the edge index reprsenting the edge that maps to a given intermediate location, we should set its shape index to the last shape point after we've added (and trimmed) that edge.

I started a branch and started typing out details of how to accomplish this but didnt end up pushing it. I thought instead I'd make this issue. Here is the comment I started writing around here: https://github.com/valhalla/valhalla/blob/master/src/thor/route_action.cc#L616-L624

```c++
        // Below we only cut the edge shape if you were doing a turn around at a via locaiton
        // AND that via location wasnt snapped to a node in the graph, ie it must be in the middle of
        // an edge where you turn around.
        //
        // Notice that when we dont cut the shape we pop back on the current path so we dont have
        // duplicate edges through the through/via location since we were continuing on the same edge
        //
        // Instead of worrying about this special case, we can treat all via/through the same and
        // always have triplegbuilder cut the shape. So even when we do not have a turn around, and
        // the route is continuing on the same edge, we can send the same edge twice and trim the
        // first one on the end and trim the second copy on the beginning to form the full edge.
        //
        // The problem comes when you have a single edge route, and you have multiple throughs/vias
        // along it. In this case you need to trim both sides of an edge before you add it if you are
        // continuing on the same edge. This special case is easy to check for, basically if the path
        // that we are adding is 1 edge in length and the next location isnt the last location we need
        // to trim both ends

        // So the concrete proposal here is to build edge_trimming objects for every via and through
        // location regardless of what the resulting path is. So this if check below goes away and we
        // remove the check for a turn around in the via_discontinuity function. we also need to
        // handle the case with the multiple back to back throughs and vias but lets get the regular
        // thing working first and then add that complexity. Once this is done we'll have gotten the
        // first requirement finished where we have a shape point at every input location present in
        // the output routes shape

        // The other problem we need to solve is we need a shape index on the location that says where
        // the geometry of that location is in the shape. We have to set this information inside of
        // triplegbuilder. The only information we need to know to solve this is what edge in the path
        // the location in question belongs to. We have that information right here: path.size() - 1
        // then over in triplegbuilder, we simply keep track of all the intermediate locations with an
        // iterator and as we hit the edge index that matches that locations edge index we marked here
        // right after we put the shape in there, we set the shape index. you could actually set the
        // locations shape index right here in route_action, but set it to path.size() - 1, the paths
        // edge index and then overwrite it in triplegbuilder when we make the shape for this edge
```

The other places to make changes would be in triplegbuilder, first at the beginning of the loop to get an iterator into the "throughs" location vector and then after the shape trimming to check if the edge whose shape we just added was the one beloinging to the currently tracked through location iterator. if it is, then we set that locations shape index to `leg_shape.size() -1` and increment the through iterator. that work should happen around here: https://github.com/valhalla/valhalla/blob/master/src/thor/triplegbuilder.cc#L1628

kdiluca:
# Issue

The goals are to:

1.  Add a shape point to a given route leg where ever that leg had intermedia locations, ie. locations with type via or through)
2. Add an index to the output location that tells which shape index in the leg maps to the shape point representing the location

Kevin describes the work best in this issue: https://github.com/valhalla/valhalla/issues/3268
fixes #3268 

A new via_waypoints array on the leg is introduced that describes where a particular waypoint from the root-level array matches to the route.
Addition to response:
```
{
  "routes": [
    {
    “legs”: [
        {
          “via_waypoints”: [
            {    // This is data concerns the 5th waypoint in root level array
              “waypoint_index”: 4,
              “distance_from_leg_start”: 258,
              “geometry_index” : 7
            },
            {    // This is data concerns the 6th waypoint in root level array
              “waypoint_index”: 5,
              “distance_from_leg_start”: 500,
              “geometry_index” : 9
            },
    ],
        …
    },
    ...
    ],
```
## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/thor/route_action.cc
```diff
@@ -60,23 +60,58 @@ void via_discontinuity(
...
+      {path_index + (flip_index ? 1 : 0),
+       {{false, PointLL(), 0.0, path_index}, {true, snap_ll, dist_along, path_index}}});
+
+  // Insert a second discontinuity so the next (opposing) edge is trimmed at the end from
+  // 1-dist along to 1
+  intermediate_locs.insert(
+      {path_index + (flip_index ? 0 : 1),
+       {{true, snap_ll, 1.0 - dist_along, path_index}, {false, PointLL(), 1.0, path_index}}});
+  //}
+  for (auto& intermediate_loc : intermediate_locs) {
```
kevinkreiser, line 96: this will be quite repetative. why wouldnt you just print the last two you added.  you can do this by assigning the result of `insert` above to a variable and printing the stuff that comes out of the iterator

src/thor/route_action.cc
```diff
@@ -426,7 +461,8 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
   // Things we'll need
   TripRoute* route = nullptr;
   GraphId first_edge;
-  std::unordered_map<size_t, std::pair<EdgeTrimmingInfo, EdgeTrimmingInfo>> vias;
+  std::unordered_map<size_t, std::pair<EdgeTrimmingInfo, EdgeTrimmingInfo>>
+      intermediate_locs_edge_trimming;
```
kevinkreiser, line 465: why not just edge_trimming as it is in triplegbuilder?

src/thor/route_action.cc
```diff
@@ -469,16 +505,14 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
-          // Insert a route discontinuity if the paths meet at opposing edges and not
-          // at a graph node. Use path size - 1 as the index where the discontinuity lies.
-          via_discontinuity(*reader, *destination, path.back().edgeid, temp_path.front().edgeid, vias,
-                            temp_path.size(), true);
-        }
+
+        // We should build edge_trimming objects for every via and through
+        // location regardless of what the resulting path is.
+        // Insert a route discontinuity if the paths meet at opposing edges and not
+        // at a graph node. Use path size - 1 as the index where the discontinuity lies.
```
kevinkreiser, line 512: Maybe its better here to stop using the nomenclature about discontinuity because its very much untrue after this change

src/thor/triplegbuilder.cc
```diff
@@ -1330,11 +1331,11 @@ void TripLegBuilder::Build(
     const std::vector<PathInfo>::const_iterator path_end,
     valhalla::Location& origin,
     valhalla::Location& dest,
-    const std::list<valhalla::Location>& through_loc,
+    const std::list<valhalla::Location>& throughs,
     TripLeg& trip_path,
     const std::vector<std::string>& algorithms,
     const std::function<void()>* interrupt_callback,
-    std::unordered_map<size_t, std::pair<EdgeTrimmingInfo, EdgeTrimmingInfo>>* edge_trimming) {
+    std::unordered_map<size_t, std::pair<EdgeTrimmingInfo, EdgeTrimmingInfo>>* intermediate_locs) {
```
kevinkreiser, line 1338: this variable stores the information about where to trim the shape of the edges. renaming it is super confusing

src/thor/triplegbuilder.cc
```diff
@@ -1627,6 +1630,18 @@ void TripLegBuilder::Build(
     trip_edge->set_source_along_edge(trim_start_pct);
     trip_edge->set_target_along_edge(trim_end_pct);
 
+    // We are looping the intermediate locs on the edge and if the edge id matches the loc edge id,
```
kevinkreiser, line 1633: this wont work when there are loops in the route. this is why i was suggesting that you keep an iterator to the current "through" location outside of the for loop and increment it each time a cut is performed. its both faster to do it this way and its robust to routes with loops in them

src/thor/triplegbuilder.cc
```diff
@@ -1345,7 +1346,7 @@ void TripLegBuilder::Build(
 
   // Set origin, any through locations, and destination. Origin and
   // destination are assumed to be breaks.
-  CopyLocations(trip_path, origin, through_loc, dest, path_begin, path_end);
+  CopyLocations(trip_path, origin, throughs, dest, path_begin, path_end);
   auto* tp_orig = trip_path.mutable_location(0);
   auto* tp_dest = trip_path.mutable_location(trip_path.location_size() - 1);
```
kevinkreiser, line 1352: I was going to write a comment here about did you test whether we need to handle the special case of a single edge route, but t hinking it over, it seems its not possible. the change we made in route_action means that wherever there is an "intermediate" location we will have at least 2 edges. which means we can never hit the trivial 1 edge case below here.

in any case i would very highly  advise to start making gurka tests which excorcise this code. it will be much faster to validate the code that way than by  changing stuff here and trying it out on real data.

proto/tripcommon.proto
```diff
@@ -85,6 +85,8 @@ message Location {
   repeated PathEdge filtered_edges = 23;
   optional uint32 original_index = 24;
   optional LatLng projected_ll = 25;
+  optional uint32 leg_shape_index = 35;
+  optional double distance_from_origin = 36;
```
kevinkreiser, line 89: nit: technically its from origin of the leg so it might make sense to name it accordingly. `distance_within_leg` or `distance_from_leg_origin`? just to avoid confusion for people who are trying to figure out what it is without reverse engineering the rest of the code

run_route_scripts/gen_multipoint_route_requests.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
```
kevinkreiser, line 1: to me its sad to duplicate the code in the other script with some changes. i worry the more we do this, the more code we have that is unmaintained. is there a way to add this functionality to the existing script?

src/thor/route_action.cc
```diff
@@ -467,37 +487,45 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
       // Merge through legs by updating the time and splicing the lists
       if (!temp_path.empty()) {
         auto offset = path.back().elapsed_cost;
-        std::for_each(temp_path.begin(), temp_path.end(),
-                      [offset](PathInfo& i) { i.elapsed_cost += offset; });
+        auto distance_offset = path.back().path_distance;
+        std::for_each(temp_path.begin(), temp_path.end(), [offset, distance_offset](PathInfo& i) {
+          i.elapsed_cost += offset;
+          i.path_distance += distance_offset;
+        });
```
kevinkreiser, line 494: this is a good change but it should be noted that it doesnt work for all algorithms. basically `path_distance` isnt correct yet across the board so hopefully people dont rely on it downstream

src/thor/triplegbuilder.cc
```diff
@@ -415,49 +415,39 @@ void RemovePathEdges(valhalla::Location* location, const GraphId& edge_id) {
...
-    pe = found;
-    RemovePathEdges(trip_path.mutable_location(trip_path.location_size() - 1), pe->edgeid);
+  RemovePathEdges(&*trip_path.mutable_location()->rbegin(), path_begin->edgeid);
+  // intermediates
+  for (const auto& intermediate : intermediates) {
+    valhalla::Location* tp_intermediate = trip_path.add_location();
+    tp_intermediate->CopyFrom(intermediate);
+    // we can grab the right edge index in the path because we temporarily set it for trimming
+    RemovePathEdges(&*trip_path.mutable_location()->rbegin(),
+                    (path_begin + intermediate.leg_shape_index())->edgeid);
```
kevinkreiser, line 446: this does worry me a little bit. the comment says it all, we rely on the `leg_shape_index` here but we are only ever setting it in `route_action.cc`. currently map matching for example, isnt sending intermediate locations so this isnt a problem but anyone that calls `triplegbuilder::build` with intermediate locations but without the `leg_shape_index` set will cause a potential crash. i dont think this is happening today but we should verify that and then we should add:

```suggestion
    if (!intermediate.has_leg_shape_index()) {
      throw std::logic_error("leg_shape_index not set for intermediate location");
    }
    RemovePathEdges(&*trip_path.mutable_location()->rbegin(),
                    (path_begin + intermediate.leg_shape_index())->edgeid);
```

src/thor/unidirectional_astar.cc
```diff
@@ -210,8 +210,8 @@ inline bool UnidirectionalAStar<expansion_direction, FORWARD>::ExpandInner(
   // Skip shortcut edges for time dependent routes, if no access is allowed to this edge
```
kevinkreiser, line 210: all of the stuff we did in here to get `path_distance` set properly is good stuff but ends up being unused in this pr. and it should be noted that when bidirectional a* degenerates to unidirectional in practice we probably need to handle these same problems with the destination edge and path_distance.. probably.. but didnt check

test/gurka/test_match.cc
```diff
@@ -117,8 +117,7 @@ D--3--4--C--5--6--E)";
   // TODO: Remove the duplicate 6 when we fix odin to handle uturn maneuver generation with only one
   // turn around point
```
kevinkreiser, line 118: we fixed this in this pr, and it wasnt odins fault it was triplegbuilder

```suggestion
```

valhalla/midgard/util.h
```diff
@@ -420,17 +420,26 @@ template <class T> struct iterable_t {
   using iterator = T*;
```
kevinkreiser, line 420: we ended up not using this (were going to use it for holding a range (c++20) in an existing container). it doesnt hurt but we didnt use it

valhalla/thor/triplegbuilder.h
```diff
@@ -32,6 +32,7 @@ struct EdgeTrimmingInfo {
  */
 class TripLegBuilder {
 public:
+  using loc_itr = google::protobuf::RepeatedField<valhalla::Location>::iterator;
```
kevinkreiser, line 35:
```suggestion
```

run_route_scripts/gen_requests.py
```diff
@@ -22,20 +25,42 @@ def get_random_locs(yc, xc, radius, count):
...
 
 
 # do each input quadruplet
 locs = []
+intermediate_locs = []
 for arg in sys.argv[1:]:
-    y, x, radius, count = arg.split(',')
+    y, x, radius, count, intermediate = arg.split(',')
     locs.extend(get_random_locs(float(y), float(x), float(radius), int(count)))
+    intermediate_locs.extend(get_random_intermediate_locs(float(y), float(x), float(radius), int(count), str(intermediate)))
```
kevinkreiser, line 53: should there be an `if intermediate is not None:` here?

run_route_scripts/gen_multipoint_route_requests.py
kevinkreiser: looks like you did combine this into the other script now. can we delete this one?

run_route_scripts/gen_requests.py
```diff
@@ -22,20 +25,42 @@ def get_random_locs(yc, xc, radius, count):
...
 
 # make requests
 for start in locs:
-    for end in locs:
+  for intermediates in intermediate_locs:
+     for end in locs:
```
kevinkreiser, line 58: seems like this should be unnested and then intermediates may or may not produce a list of random intermediate locations

run_route_scripts/gen_requests.py
kdiluca: yes, I was still modifying...I think its good now

run_route_scripts/gen_requests.py
```diff
@@ -22,20 +25,53 @@ def get_random_locs(yc, xc, radius, count):
...
+        a = float(random.uniform(0.0, 1.0))
+        b = float(random.uniform(0.0, 1.0))
+        r = rads * math.sqrt(a)
+        c = 2 * math.pi * b
+        x = r * math.cos(c)
+        y = r * math.sin(c)
+        x += xc
+        y += yc
+        intermediate_locs.append({'lon':round(x, 6),'lat':round(y, 6),'type':intermediate});
+    return intermediate_locs
```
kevinkreiser, line 44: this function is identical to the one above, why duplicate? just pass the location type as a parameter right?

run_route_scripts/gen_requests.py
```diff
@@ -22,20 +25,53 @@ def get_random_locs(yc, xc, radius, count):
...
-        if start == end:
-            continue
-        req = {'locations':[start, end], 'costing': 'auto'}
+if (intermediate_locs):
+    for start in locs:
+        for intermediates in intermediate_locs:
+            for end in locs:
+                if start == end:
+                    continue
+            req = {'locations':[start, intermediates, end], 'costing': 'auto', 'shape_format': 'polyline6', 'format':'osrm'}
```
kevinkreiser, line 65: this indentation looks wrong, `end` is not in scope

run_route_scripts/gen_requests.py
```diff
@@ -22,20 +25,38 @@ def get_random_locs(yc, xc, radius, count):
...
+        y, x, radius, count = arg.split(',')
+        locs.extend(get_random_locs(float(y), float(x), float(radius), int(count), "break"))
 
 # make requests
-for start in locs:
-    for end in locs:
-        if start == end:
-            continue
-        req = {'locations':[start, end], 'costing': 'auto'}
+if (intermediate_locs):
```
kevinkreiser, line 44: both prints below also look like the wrong indentation

kevinkreiser (approved):

PR #3280, opened 2021-08-20
Fix trace_options matched_points edge indexes.

tatiana-yan:
# Issue

There are two problems with trace_options matched_points edge indexes:
* default (unset) value is 0 which is valid value. So in case of edge_index calculation errors unset value looks like valid value.
* valid first_match_idx/last_match_idx may be dropped which leads to unset edge_index for matched_points with type "matched".

## Tasklist

 - [x] Add tests

## Requirements / Relations


valhalla/meili/match_result.h
```diff
@@ -34,7 +37,7 @@ struct MatchResult {
   bool ends_discontinuity;
   // An index into the full list of edges in the path even across discontinuities (for
   // trace_attributes)
-  size_t edge_index;
+  size_t edge_index = kInvalidEdgeIndex;
```
kevinkreiser, line 40: nothing anywhere else in the code checks for this new value though so im a bit confused about what this is doing. i am guessing your point is that sometimes this values is garbage because its never set? if that is the case, with this change, wont this non-garbage value get treated as if it is a normal value?

kevinkreiser:
@tatiana-yan thank you so much for working on this! it will take my brain a bit of time to get back into map match mode but i will get it reviewed!

valhalla/meili/match_result.h
kevinkreiser: or do we have checks elsewhere that look for `-1` which would be equivalent? if that is the case we should change those to use this constant so its clear what the intent is, ie a sentinal value

tatiana-yan: Current situation is that there are lots of matched_points with type `matched` and edge_index `0` but `0` is wrong edge_index (the real situation is edge_index is not set due to error).
I try to fix this error here. But I think that it'll be easier to find and fix another possible errors if default(unset) edge index will have some invalid value (not `0` which is valid edge index).

Now I only try to provide output which will indicate that something is wrong. Another option is to check that we do not have `edge_index == meili::kInvalidEdgeIndex` for point with type `matched` (is `thor_worker_t::build_trace` good place for such check?). The only problem with the second approach is that matched_points edge indexes looks like not very popular trace attribute and I'm not sure is it ok to break other attributes getting in case of edge indexes errors.

kevinkreiser: do your unit tests fail if we undo this? to me if the point is "matched" then it must have a valid index. so if the unit tests show how this can happen then id be happy to debug further to see what the heck is going on

tatiana-yan: If you'll undo `src/meili/match_route.cc` changes you'll get `kInvalidEdgeIndex` + `type == "matched"` errors in mapmatch test. If you'll undo `valhalla/meili/match_result.h` changes too you'll get `type = matched` with `edge_index = 0` which hides unset edge index.

kevinkreiser: excellent that gives me a perfect test to debug. clearly a point being marked as matched but not having an edge it belongs to makes no sense so hopefully we can rectify that and not have to have an invalid value

kevinkreiser: ok i just re-read this. it seems to me that i misunderstood what you were saying. in fact your change to setting the proper index only if its valid does indeed at least fix the new test you added from having the invalid index set on the "matched" point. this makes me think that maybe all we still need to do is add an `assert(match_point.edge_index != kInvalidEdgeIndex);` if that never fires for all the map matching tests then id be relatively satisfied that we won't ever return invalid index from the serializer.

ill hunt around for the right place to put that assertion and give it a try!

kevinkreiser: ok so it seems we only set the edge_index as the very last thing before we serialize the result out, which means we only ever loop over the matched points as the very last step in serialization. so the only decent place to put an assert would be in serialization. also i noticed the check there is stupid, it checks that the edge id is valid but it should check if the points type is `kMatched` instead: https://github.com/valhalla/valhalla/blob/master/src/tyr/trace_serializer.cc#L390

the unfortunate part of this is that i forgot that we only set that edge index for trace attributes. that means that the bulk of the map matching tests wont exorcise this fix. im struggling to at least thing of a hacky way to make a bigger test of this change. instead i think i should focus probably on  checking your example failing test (before your fix) and see why exactly those indices are uninitialized. maybe its just a case i didnt remember to account for. unless since you just made the fix maybe you can explain it why we have to handle the match index not being set

tatiana-yan: Is there anything I can do to help?

kevinkreiser: yeah i guess im just missing the explanation as to why this happens and why the fix is the right one. i can get that by stepping through the code for the example that doesnt work but i just didnt get a chance to do that yet.

tatiana-yan: I just tried to fix some real-life samples with lots of incorrect `0` edge indexes and I discovered that indexes are unset. Then I found that `thor_worker_t::build_trace` is the only place we fill edge indexes and we use `first_match_idx`/`last_match_idx` for it. Then I debugged places where we set `first_match_idx`/`last_match_idx` and found that place where we overwrite values which are good for edge indexes with `-1`. I looked to another code which uses `first_match_idx`/`last_match_idx` and I see no reason to overwrite values with `-1` (maybe I missed something). So I found it from debug and maybe it's not a perfect fix.

kevinkreiser: no i think its totally valid i just want to debug it myself to see why we didnt think of that in the first place. seems to me either just have a bug and forgot to add the code you added or there is a bug further upstream that causes us to have to work around what we clearly expected not to happen. again sorry for the delay, im still trying to get time to look at this

src/meili/match_route.cc
```diff
@@ -234,7 +234,12 @@ std::vector<EdgeSegment> ConstructRoute(const MapMatcher& mapmatcher,
         // we modify the first segment of the new_segments accordingly to replace the previous
         // one in the route.
         new_segments.front().source = route.back().source;
-        new_segments.front().first_match_idx = route.back().first_match_idx;
+        // Prefer first_match_idx from previous segments but do not replace valid value with invalid.
+        if (route.back().first_match_idx != -1)
+          new_segments.front().first_match_idx = route.back().first_match_idx;
+        // Prefer last_match_idx from new segments but do not replace valid value with invalid.
+        if (new_segments.front().last_match_idx == -1)
+          new_segments.front().last_match_idx = route.back().last_match_idx;
```
kevinkreiser, line 242: i've finally had the time to actually run the code and think about what is going on here. so yeah here we are replacing the last segment in the last portion of the match that we found with the first segment of the next portion of the match we found. since they occur on the same edge we basically need to merge the two segment records. this means we want the first match index to be earliest point we would have seen in the match, ie from `route.back()` but its possible that really only one match point landed on this edge in which case `route.back()` would only have set the `last_match_idx` this is why we have to check for "unset". similarly when setting the `last_match_idx` of the merged record we know that the `new_segments` would have seen the "latest" point that landed on this edge, however if, again only one point would have landed on that edge, it would have only set `first_match_idx`. in this case we can be sure that `route.back()` will have its `last_match_idx` set.

basically when only one point falls on a given edge segment and we need to merge the segment records there, we should set both the first and last match_indices to the same index. i think the fix you have accomplishes that though i cant shake the feeling that it could be simplified.

kevinkreiser (approved):

kevinkreiser (approved):
thank you so much for doing this, and good eyes!

tatiana-yan:
thank you for review

guoxifeng1990:
Thank you @tatiana-yan first for all the correction done in this bug. 

But, there is still a small problem for interpolated points. Sometimes, interpolated points do not have edge_index like:

Capture

I had a look on your correction, but no easy to figure out where this error comes from. Maybe you have a better  insight on this?

Thanks again!

PR #3281, opened 2021-08-23
de-singleton tile extract in graphreader

### LINKED ISSUE #3117: De-Singleton tile_extract_t to Allow for Config (tileset) Reloading
At the moment graphreader uses a singleton especially for the most performant mode of operation, memory mapped tar file. This is done in large part because scanning and enumerating what is in the tar takes a decent amount of time if the file isnt already memory mapped. In this way the singleton is filling its typical role as a cache for other threads to make use of. This has a pretty steep down side though which is that once the extract is loaded its not possible to load another one. We've worked around this for unit tests by subclassing graphreader and adding methods to it to allow this: https://github.com/valhalla/valhalla/blob/916cea1454646917fd5c9fbc0c49087ae35208ce/test/test.cc#L404-L416

This solution is not scalable though in a multithreaded context so we must do something else to make this possible.

You'll notice that the actual thing we are talking about here is the `tile_extract_t` struct which houses both the extracts (graph tiles as well as speed tiles for real time speed). Both files reflect each other in terms of organization, a tar with tiles in it each of which has a file name that tells which tile it is. What costs time is running over the whole file to discover what and where tiles are located in the file. If getting this information were trivial the process could be sped up such that a singlton wouldnt be needed.

Many years ago we had such an implementation, not in valhalla but non-the-less for memory mapping a large file. in this implementation we passed an index which told where the bits were in the large file so that the program didnt need to scan it first. here we could do the same thing. pass an index with the tar that allows the `tile_extract_t` construct itself in very short time. The index would need a map, like that is found in the extract, of tile_id to byte_offset within the large tar file.

I think it would be nice not to require this index but to use it if its there. This means that configuring will take a long time in the case that it is not supplied but i dont see any other way to be lenient. the other option is to not be lenient and error when no index is provided.

The other question is how to provide the index. It would be really cool if we could make it the first file in the tar extract. this would require some trickery. i believe this could be done the following way:

1. create the tar archive as normal with all the files in it
2. enumerate the files and their byte offsets in an index file, format should be fixed sized struct {uint32_t tile_id, uint64_t byte_offset}
3. create a tar with only the index file in it
4. modify the index file by shifting all the offsets by the size of the tar with just the index file in it
5. recreate tar file with just the index in it (index is updated)
6. cat the larger tar with all the tiles in it onto the index only tar

Tars use separators between entries to mark the boundaries, using `cat` directly will make it so that the tar with just the index file will have what looks like an `end of archive` marker in it. valhalla is robust to this and will read past it to get the rest of the archive. if you want you could use `tar --concatinate` to avoid this problem but you'll also need to subtract the size of one tar block (512 bytes) from the byte offsets you put in the index

If we do this and we update tile_extract_t to read the index if its the first thing in the archive then we can assume that reloading takes only milliseconds and no longer requires a singleton to hold the memory map. I'm not sure if there is any issue with race conditions over creating the first memory map of a file but in the end threads should end up sharing the same map provided to them by the kernel. We shoudl probably look into that, https://man7.org/linux/man-pages/man2/mmap.2.html seems to mention using `MAP_FIXED_NOREPLACE` potentially but i havent read it carefully.

nilsnolde:
fixes #3117 

I started with this on the side, working on it here and there. to re-iterate: we want to be able to load other tile extracts while a consumer application is running (e.g. `valhalla_service`). in #3117 @kevinkreiser proposed to use a fixed-width index file storing the tile_id and byte offset in the final tar file which the graphreader can read first (if available) to really fast initialize the tile_extract_t (faster because it does not have to scan through the entire tar file thanks to the index file).

tasks:

- [x] set up python script
- [ ] tests python (needs some refactoring)
- [ ] adapt graphreader to read the index file first if present
- [ ] tests graphreader
- [ ] some open questions:
    - this (currently) **does not work for existing extracts**, so far it only builds an extract from `mjolnir.tile_dir`. is that fine with you? we could also add that ability and it sorta was what @kevinkreiser initially suggested
    - where to add this functionality: python bindings for sure IMO, what about `valhalla_service`?

@kevinkreiser I didn't do much yet on the c++ side, will comment a bit on the files, a few pointers would be cool:)

example commandline: `python scripts/valhalla_build_extract.py --config site/valhalla.json`


scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,155 @@
...
+import sys
+import tarfile
+from tarfile import BLOCKSIZE
+from typing import Dict
+
+description = """Builds a tar extract from the tiles in mjolnir.tile_dir to the path specified in mjolnir.tile_extract."""
+
+# "=" prefix means native byte order with standard size and no alignment:
+# https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment
+STRUCT_FORMAT = '=LQ'
```
nilsnolde, line 18: this confuses me a little, lacking the solid CS background I guess.. this is the byte layout specifier for `struct` for `uint32_t` and `uint64_t` which creates 12 byte packets for two numbers: https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment, the comment says it all. is this the way how to make sure any consumer architecture can read any other producer's architecture's `index.bin` file with this layout (e.g. I produce an extract on `x86_64` and someone with `ppc64` can read it)? or is there some more trickery involved in python or c++ to make sure endianess etc is properly taken care of?

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,155 @@
...
+    """Recursively fills the passed map with tile path & size info"""
+    for p in in_path.iterdir():
+        if p.is_dir():
+            get_tile_info(p, tiles)
+        if p.is_file():
+            if not p.suffix == '.gph':
+                LOGGER.warning(f"Not a graph tile: {p}")
+                continue
+
+            tiles[p.resolve()] = get_padded_tar_size(p.stat().st_size)
```
nilsnolde, line 50: this is the map that will cache the tile ID and offset in the tar file, e.g. `{'/home/nilsnolde/dev/cpp/valhalla/site/valhalla_tiles/2/000/762/146.gph': 6595072}`

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,155 @@
...
+    extract_fp: Path = Path(config["mjolnir"]["tile_extract"])
+    tiles_fp: Path = Path(config["mjolnir"]["tile_dir"])
+
+    # set the right logger level
+    if args.verbosity == 0:
+        LOGGER.setLevel(logging.CRITICAL)
+    elif args.verbosity == 1:
+        LOGGER.setLevel(logging.INFO)
+    elif args.verbosity >= 2:
+        LOGGER.setLevel(logging.DEBUG)
```
nilsnolde, line 67: no `-v`: no logging
`-v`: info logging
`-vv`: debug logging

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # build the tar with the index file as the first one
+    index_bin_fd = BytesIO(index_bin)
+    index_bin_fd.seek(0)
+    # write max 100 logs
+    reporting_size = int(len(tiles_stats) / 100) or 1
+    with tarfile.open(extract_fp, 'w') as tar:
+        tarinfo = tarfile.TarInfo(str(TAR_ARCNAME_DIR.joinpath(INDEX_FILE)))
+        tarinfo.size = index_size
+        tar.addfile(tarinfo, index_bin_fd)
```
nilsnolde, line 101: adds the in-memory `index_bin` file to `valhalla_tiles/index.bin` in the tar

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,155 @@
...
+
+        for count, tile_path in enumerate(tiles_stats.keys()):
```
nilsnolde, line 103: then add all the tile files

src/baldr/graphreader.cc
```diff
@@ -41,17 +41,28 @@ GraphReader::tile_extract_t::tile_extract_t(const boost::property_tree::ptree& p
...
+      // map files to graph ids: use precomputed index.bin if available, else scan the tar
+      auto first_file = archive->contents.find("index.bin");
+      if (first_file != archive->contents.end()) {
+        // TODO: read the file contents from the tar
+        // open file with ofstream, seek sizeof(tar::header_t)
+        // then loop through in 12 byte packets until archive->contents->second->second (file size) is
+        // reached; probably create a struct for tile_id / offset and fill tiles with tile_id:
+        // (file_name, file_size):
+        //
+        // https://www.cs.uaf.edu/2010/spring/cs202/lecture/02_04_binary_IO.html
```
nilsnolde, line 53: is that an ok plan? I'd like to do it myself actually, the only point is I'm not sure about the byte layout thing here, related to the python equivalent in https://github.com/valhalla/valhalla/pull/3281/files#r693936967

src/baldr/graphreader.cc
```diff
@@ -41,17 +41,27 @@ GraphReader::tile_extract_t::tile_extract_t(const boost::property_tree::ptree& p
...
+      // map files to graph ids: use precomputed index.bin if available, else scan the tar
+      auto first_file = archive->contents.find("index.bin");
+      if (first_file != archive->contents.end()) {
+        // TODO: read the file contents from the tar
+        // open file with ofstream, seek sizeof(tar::header_t)
+        // then loop through in 12 byte packets until archive->contents->second->second (file size) is
+        // reached; get the tile_id and compute the position (without header) & size from the offsets
+        //
+        // https://www.cs.uaf.edu/2010/spring/cs202/lecture/02_04_binary_IO.html
+      } else {
```
nilsnolde, line 53: is that an ok plan? I'd like to do it myself actually, the only point is I'm not sure about the byte layout thing here, related to the python equivalent in https://github.com/valhalla/valhalla/pull/3281/files#r693936967

nilsnolde: also: if no index.bin is provided we wouldn't want to let people reset the tile_extract right? how would that be achieved? currently there's a `static` instance in https://github.com/valhalla/valhalla/blob/nn-desingleton-tile-extract/src/baldr/graphreader.cc#L120

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,157 @@
...
+
+        index_size = STRUCT_SIZE * len(tiles_stats)
+        index_bin = b''
+        offset = get_padded_tar_size(index_size)  # initialize with the index file size inside the tar
+        for tile_path, tile_size in tiles_stats.items():
+            # extract level and ID portion to create a GraphId
+            # tile_path is the absolute path to the tile
+            tile_path = str(tile_path.relative_to(tiles_fp))[:-4]
+            level, idx = tile_path.split('/', 1)
+            tile_id = int(level) | (int(idx.replace('/', '')) << 3)
```
nilsnolde, line 84: currently this will throw if it's not parsing a valid tile filepath, even though we also only allow `.gph` files.. think I'll change that so also non-tiles can be tarred, the valhalla tar reader is also robust against invalid tile ids as tar names

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,158 @@
...
+            index_bin += struct.pack(STRUCT_FORMAT, tile_id, offset)
+            offset += tile_size
+    else:
+        LOGGER.critical(f"Neither 'tile_extract': {extract_fp} nor 'tile_dir': {tiles_fp} were found on the filesystem.")
+        sys.exit(1)
+
+
+
+    # build the tar with the index file as the first one
+    index_bin_fd = BytesIO(index_bin)
```
nilsnolde, line 97: this is like an in-memory open binary file descriptor

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,158 @@
...
+            tiles[p.resolve()] = get_padded_tar_size(p.stat().st_size)
+
+
+if __name__ == '__main__':
+    args = parser.parse_args()
+    with open(args.config) as f:
+        config = json.load(f)
+    index_fp: Path = args.config.parent.joinpath(INDEX_FILE)
+    extract_fp: Path = Path(config["mjolnir"]["tile_extract"])
+    tiles_fp: Path = Path(config["mjolnir"]["tile_dir"])
```
nilsnolde, line 55: I'll break most of the logic out of the `__main__` so it's easier to test it. basically only keeping the args parsing

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,158 @@
...
+import tarfile
+from tarfile import BLOCKSIZE
+from typing import Dict
+
+description = """Builds a tar extract from the tiles in mjolnir.tile_dir to the path specified in mjolnir.tile_extract."""
+
+# "=" prefix means native byte order with standard size and no alignment:
+# https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment
+STRUCT_FORMAT = '=LQ'
+STRUCT_SIZE = struct.calcsize(STRUCT_FORMAT)
```
nilsnolde, line 19: this resolves to 12 bytes with the above format

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,158 @@
...
+INDEX_FILE = "index.bin"
+TAR_ARCNAME_DIR = Path('valhalla_tiles')
+
+parser = argparse.ArgumentParser()
```
nilsnolde, line 23: forgot to add the description from above..

src/baldr/graphreader.cc
```diff
@@ -41,17 +41,27 @@ GraphReader::tile_extract_t::tile_extract_t(const boost::property_tree::ptree& p
     try {
       // load the tar
       archive.reset(new midgard::tar(pt.get<std::string>("tile_extract")));
```
nilsnolde, line 43: just realized that the block below is too late, the archive should of course not even get initialized..

scripts/valhalla_build_extract
```diff
@@ -0,0 +1,158 @@
...
+import sys
+import tarfile
+from tarfile import BLOCKSIZE
+from typing import Dict
+
+description = """Builds a tar extract from the tiles in mjolnir.tile_dir to the path specified in mjolnir.tile_extract."""
+
+# "=" prefix means native byte order with standard size and no alignment:
+# https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment
+STRUCT_FORMAT = '=LQ'
```
nilsnolde, line 18: this confuses me a little, lacking the solid CS background I guess.. this is the byte layout specifier for struct to create a uint32_t and a uint64_t which results in 12 byte packets for two numbers: https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment, the comment says it all. is this the way how to make sure any consumer architecture can read any other producer's architecture's index.bin file with this layout (e.g. I produce an extract on x86_64 and someone with ppc64 can read it)? or is there some more trickery involved in python or c++ to make sure endianess etc is properly taken care of?

scripts/valhalla_build_extract
```diff
@@ -0,0 +1,158 @@
...
+
+    # build the tar with the index file as the first one
+    index_bin_fd = BytesIO(index_bin)
+    index_bin_fd.seek(0)
+    # write max 100 logs
+    reporting_size = int(tiles_count / 100) or 1
+    with tarfile.open(extract_fp, 'w') as tar:
+        tarinfo = tarfile.TarInfo(str(TAR_ARCNAME_DIR.joinpath(INDEX_FILE)))
+        tarinfo.size = index_size
+        tar.addfile(tarinfo, index_bin_fd)
```
nilsnolde, line 85: adds the in-memory index_bin file to valhalla_tiles/index.bin in the tar

scripts/valhalla_build_extract
```diff
@@ -0,0 +1,158 @@
...
+import tarfile
+from tarfile import BLOCKSIZE
+from typing import Dict
+
+description = """Builds a tar extract from the tiles in mjolnir.tile_dir to the path specified in mjolnir.tile_extract."""
+
+# "=" prefix means native byte order with standard size and no alignment:
+# https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment
+STRUCT_FORMAT = '=LQ'
+STRUCT_SIZE = struct.calcsize(STRUCT_FORMAT)
```
nilsnolde, line 19: this resolves to 12 bytes with the above format

src/baldr/graphreader.cc
```diff
@@ -37,31 +37,55 @@ tile_gone_error_t::tile_gone_error_t(std::string prefix, baldr::GraphId edgeid)
...
         }
-      }
-      // couldn't load it
-      if (tiles.empty()) {
-        LOG_WARN("Tile extract contained no usuable tiles");
-      } // loaded ok but with possibly bad blocks
-      else {
-        LOG_INFO("Tile extract successfully loaded with tile count: " + std::to_string(tiles.size()));
-        if (archive->corrupt_blocks) {
-          LOG_WARN("Tile extract had " + std::to_string(archive->corrupt_blocks) + " corrupt blocks");
```
nilsnolde, line 64: hard to see, but really only shuffling around the conditions, nothing changed actually in this bit

scripts/valhalla_build_extract
```diff
@@ -0,0 +1,158 @@
...
+                level, idx = tile_path.split('/', 1)
+                tile_id = int(level) | (int(idx.replace('/', '')) << 3)
+            except ValueError:
+                # ignore invalid tile ids, so other files can be tarred
+                continue
+
+            LOGGER.debug(f"Tile {tile_path} (ID {tile_id}) with padded tar size: {tile_size}")
+
+            # add 4 & 8 byte ints to the index file
+            index_bin += struct.pack(STRUCT_FORMAT, tile_id, tile_size)
```
nilsnolde, line 72: instead of byte offset, we store the padded file size, which I think is easier for the graphreader logic

src/baldr/graphreader.cc
```diff
@@ -37,31 +37,60 @@ tile_gone_error_t::tile_gone_error_t(std::string prefix, baldr::GraphId edgeid)
...
+        auto remainder = index_size % block_size;
+        auto tile_pos =
+            block_size + index_size + block_size - (remainder == 0 ? block_size : remainder);
+        while (fin.tellg() != first_header.get_file_size() + block_size) {
+          uint32_t tile_id = 0;
+          uint64_t tile_padded_size = 0;
+          fin.read(reinterpret_cast<char*>(&tile_id), sizeof(tile_id));
+          fin.read(reinterpret_cast<char*>(&tile_padded_size), sizeof(tile_padded_size));
+
+          tiles[tile_id] = std::make_pair(reinterpret_cast<char*>(&tile_pos), tile_padded_size);
```
nilsnolde, line 63: tile_id & tile_size are correctly set, `tile_pos` is not. missing the right type casting..

also, same problem with endianess etc as in the python equivalent: https://github.com/valhalla/valhalla/pull/3281/files#r694160065.

any suggestions here would be welcome!

src/baldr/graphreader.cc
```diff
@@ -37,31 +37,61 @@ tile_gone_error_t::tile_gone_error_t(std::string prefix, baldr::GraphId edgeid)
...
+        auto tile_pos =
+            block_size + index_size + block_size - (remainder == 0 ? block_size : remainder);
+        while (fin.tellg() != first_header.get_file_size() + block_size) {
+          uint32_t tile_id = 0;
+          uint64_t tile_padded_size = 0;
+          fin.read(reinterpret_cast<char*>(&tile_id), sizeof(tile_id));
+          fin.read(reinterpret_cast<char*>(&tile_padded_size), sizeof(tile_padded_size));
+
+          // add header size to position, subtract header size from file size
+          tiles[tile_id] = std::make_pair(reinterpret_cast<char*>(&tile_pos + block_size),
```
nilsnolde, line 64: tile_id & tile_size are correctly set, `tile_pos` is not. missing the right type casting..

also, same problem with endianess etc as in the python equivalent: https://github.com/valhalla/valhalla/pull/3281/files#r694160065.

any suggestions here would be welcome!

scripts/valhalla_build_extract
kevinkreiser: we should be using little endian for everything. little endian won the war, any platform we target will will be LE.

there are things we need to do in c++ to be robust to running on a different architecture than where the data was created but we've sided stepped that and pretty much assume little endian everywhere.

kevinkreiser: I think it maybe it should be `<LQL`, which would have a matching c++ struct of:

```c++
struct tile_entry {
  uint32_t tile_id; //just level and tileindex hence fitting in 32bits
  uint64_t offset; //number of bytes offset from the beginning of the tar to the beginning of this tile
  uint64_t size;   //size of the tile in bytes
};
```

kevinkreiser:
@nilsnolde i didnt get to review the code in full yet but im not sure how you can do the index without the offset to the tile in the archive. i did want to make a suggestion at how to implement this though in a way that i think would be fool proof in terms of tar building. broadly i would describe in these steps:

1. count the tiles that are going to go into the tar
2. make an index file that will be big enough to fit one entry per tile, you'll need 32bits for the tile id, 64bits for the byte offset in the tar and 32 more bits for the size of the tile. the index file you make doesnt need real info in it yet just has to be the right size
3. make the tar with the blanked index file first and the rest of the tiles in sorted order (we my try different packing algorithms later)
4. as you are adding the tiles to the tar or after the fact in a loop, get the information about what tile was where in the tar and how big it was, ie store the actual info about each tile
5. open the tar file in binary write mode, seek to the spot where the index file is and over write it with this information now that we have it

This way you wont have to make sure your math is right or that you correctly computed headers and trailers and padding. You simply do a two pass algorithm where to make the sapce for the index in the first pass and then write the actual info into it in the second pass.

As far as the c++ goes, i think we shouldnt need to make too many changes over there. We can simply say, if the first thing i encounter in the tar is an index file, call another function to read that file and bail, otherwise continue reading the file as we already do today. This will mean that people can screw themselves by reloading the tar and not having the index, but we can warn about this potentially. 

let me know if any of this makes sense, id be happy to pair up with you for a bit to clear it up. this is a huge feature thanks again for all of your tireless work.

src/baldr/graphreader.cc
```diff
@@ -37,31 +37,61 @@ tile_gone_error_t::tile_gone_error_t(std::string prefix, baldr::GraphId edgeid)
 
 GraphReader::tile_extract_t::tile_extract_t(const boost::property_tree::ptree& pt) {
   // if you really meant to load it
-  if (pt.get_optional<std::string>("tile_extract")) {
+  auto extract_fp = pt.get_optional<std::string>("tile_extract");
+  if (extract_fp) {
     try {
-      // load the tar
-      archive.reset(new midgard::tar(pt.get<std::string>("tile_extract")));
```
kevinkreiser, line 43: yeah i would just keep this exactly the way it is already and instead of doing work here, i would go into midgard::tar and i would add a new feature to that interface that allows us to inject functionality for parsing the contents from an index file. let me hack up a proposal somehow...

nilsnolde:
thanks @kevinkreiser sounds reasonable enough. I'd actually like to keep most of the logic the way it is, we do have to do some basic header + padding calculation anyways to compute the right position of the data files even if we write the full dummy index and rip through the tar. but yeah, I did sacrifice a lot of readability for saving a few MB not tracking the position in the index.bin but only the padded data size. I'll get rid of that. (btw, it does work actually with only the data size tracking, you just need to have the position of the very first tile header as anchor, but it's ridiculous logic in c++..)

I'll need to take a little of a break, but be sure I'll get back to this soonish. and thanks for the draft over in #3283 !

src/baldr/graphreader.cc
nilsnolde: yeah much nicer..

nilsnolde:
oh wow, I totally didn't see the `offset_data` member! yep, then it makes more sense to me:)

kevinkreiser:
truth be told i didnt know for sure it existed before i looked but i had a strong suspicion it had to just from having implemented this stuff on the c++ side :smile:

nilsnolde:
I knew that I looked in the documentation haha if you look here and search for "offset" there's no mention of it argh.. https://docs.python.org/3/library/tarfile.html wtf python, will be my first core PR ;)

kevinkreiser:
fwiw i often find python docs leave out some details. in this case its particularly agreedious that they dont spell out the entire `member` object but in other places i'm often wondering things like what is the return value etc. i think this is just par for the course in weakly typed languages

src/baldr/graphreader.cc
```diff
@@ -36,22 +43,45 @@ tile_gone_error_t::tile_gone_error_t(std::string prefix, baldr::GraphId edgeid)
...
+  auto index_loader = [this](const std::string& filename, const char* index_begin,
+                             size_t size) -> decltype(midgard::tar::contents) {
+    // has to be our specially named index.bin file
+    if (filename != "index.bin")
+      return {};
+
+    // get the info
+    decltype(midgard::tar::contents) contents;
+    tile_index_entry t = {};
+    for (const auto& entry : midgard::iterable_t<tile_index_entry>(&t, size)) {
```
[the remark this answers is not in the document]
nilsnolde: works now @kevinkreiser , thanks!

nilsnolde:
@kevinkreiser it works on both parts now, c++ & python. can you advise what needs to be done on the graphreader's usage now that we won't maintain a `static` tar extract anymore?

I'll also write some tests for both languages once I have a better understanding what needs to be adapted.

scripts/valhalla_build_extract
```diff
@@ -0,0 +1,111 @@
...
+import tarfile
+from tarfile import BLOCKSIZE
+from time import time
+from typing import Dict, Tuple, List
+
+description = """Builds a tar extract from the tiles in mjolnir.tile_dir to the path specified in mjolnir.tile_extract."""
+
+# "<" prefix means little-endian and no alignment
+# order is important! if uint64_t is not first, c++ will use padding bytes to unpack
+STRUCT_FORMAT = '<QLL'
```
nilsnolde, line 20: `QLL` aligns better, c++ would've padded all int32 to 8 bytes otherwise

src/baldr/graphreader.cc
```diff
@@ -106,6 +140,9 @@ GraphReader::tile_extract_t::tile_extract_t(const boost::property_tree::ptree& p
   }
 }
 
+// TODO: delete this and call the extract constructor directly from within graphreader
+// before we do this we need to figure out what to do with traffic tile extract
+// they should support index reading as well for fast reload
```
nilsnolde, line 145: ah was looking for that comment!

I don't have a traffic tile extract handy, I hardly know their folder layout when they're not archived. you think we could use the current `valhalla_build_extract` script would work as it is right now to produce a new traffic tar? I guess it should, seems like graphreader expects the exact same path/folder layout as for the routing tile extract. the valhalla.json is missing a `mjolnir.traffic_dir` though, to know which directory will be archived.

scripts/valhalla_build_extract
```diff
@@ -0,0 +1,111 @@
...
+    print(f"Finished tarring {tiles_count} tiles to {extract_fp}")
+
+
+if __name__ == '__main__':
+    args = parser.parse_args()
+    with open(args.config) as f:
+        config = json.load(f)
+    extract_fp: Path = Path(config["mjolnir"]["tile_extract"])
+    tiles_fp: Path = Path(config["mjolnir"]["tile_dir"])
+
```
nilsnolde, line 102: if we want to create traffic extracts with indices as well, we'd need to add a `mjolnir.traffic_dir` to the config

nilsnolde:
I pulled the tile extract into the graphreader ctor, getting rid of the `get_extract_instance()` stuff. I registered 3 instantiations of the graphreader **per thread** while starting up `valhalla_service`:
- https://github.com/valhalla/valhalla/blob/nn-desingleton-tile-extract/src/meili/map_matcher_factory.cc#L32
- https://github.com/valhalla/valhalla/blob/nn-desingleton-tile-extract/src/baldr/connectivity_map.cc#L120
- https://github.com/valhalla/valhalla/blob/nn-desingleton-tile-extract/src/loki/worker.cc#L165

just checking in that that's expected:)

nilsnolde:
it's actually not expected, fixing it over in #3307

nilsnolde:
@kevinkreiser we talked about also creating the indexed tar file for the traffic extract. I looked at that yesterday and I'm wondering: we actually have to know the directed edge count for every traffic tile in advance right? Otherwise I don't see how we could make sure we don't create a traffic tar skeleton that's too small in size for the updater process downstream.. If so, the only viable option is to crack open the tile headers and inspect the `directededgecount_` no (without bloating most traffic tiles to an abnormous size)? Or is there some other way to get that info?

If that's the case, I'd propose to make it a separate script `valhalla_build_traffic_extract` since the logic will be very different from the routing tile extract. I can totally include that in this PR, just wanna confirm first which way's best or if I'm misunderstanding smth grossly here.

kevinkreiser:
@nilsnolde thats right we'd need to create it by looping over an existing graph tileset. this should be straightforward in python though. we just need to make a struct on the python side that looks like the graphtileheader and as you said pull out the directed edge count. to me i think it would be fine to add an option to the script to make a traffic extract rather than building a whole new script but i didnt think about it too hard so its up to you. just seems that the thing is very related and would be less confusing than having multiple scripts

>  is it ok if I include a mini tar extract
to get the full test, you should probably have cmake run the python script to generate the tar from a tileset set that we already create. we already build utrecht, so why not have cmake then tar it and then add a unit test to check the resulting tar is accessable in the same was as the other tileset? that would be my suggestion

nilsnolde:
> to me i think it would be fine to add an option to the script to make a traffic extract

yes you're right actually, has to happen after the graph tarring anyways, so why not include it there

> so why not have cmake then tar it

my worry was more that the c++ graphreader test (which I assume we want as well for the new behavior) is then also depending on the output of the python test (tar with index). I'd have imagined that to be painful enough to justify a small tar file in the repo but happy to do it in cmake if you figure that's easy enough @kevinkreiser

kevinkreiser:
regarding the test i think we can do something just like this: https://github.com/valhalla/valhalla/blob/master/test/CMakeLists.txt#L117-L142

except that our custom command will depend on the `utrecht_tiles` target and that our new unit test will depend on whatever our custom command target will be named. im no cmake wizzard but i am happy to help if youd like!

nilsnolde:
Ok cool thx, I’ll try that soon and cry got help if I get stuck😉

scripts/valhalla_build_extract
nilsnolde: Resolved, works totally different ..

nilsnolde:
Ok I'm done with the traffic extract builder on the python side in case you wanna have a look @kevinkreiser . will build it into c++ tmrw.

with regards to testing: also here I'd like to wait for #3308 to be merged so the python stuff is set up

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,179 @@
...
+class TileHeader(ctypes.Structure):
+    """
+    Resembles the uint64_t bit field bytes 40 - 48 of the
+    graphtileheader to get the directededgecount_.
+    """
+    _fields_ = [
+        ("nodecount_", ctypes.c_ulonglong, 21),
+        ("directededgecount_", ctypes.c_ulonglong, 21),
+        ("predictedspeeds_count_", ctypes.c_ulonglong, 21),
+        ("spare1_", ctypes.c_ulonglong, 1)
```
nilsnolde, line 41: this emulates a bit field, really handy..

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,179 @@
...
+            if not tile_in.name.endswith('.gph'):
+                continue
+            # jump to the data's offset and skip the uninteresting bytes
+            in_fileobj.seek(tile_in.offset_data + GRAPHTILE_SKIP_BYTES)
+
+            # read the appropriate size of bytes from the tar into the TileHeader struct
+            tile_header = TileHeader()
+            b = BytesIO(in_fileobj.read(ctypes.sizeof(TileHeader)))
+            b.readinto(tile_header)
+            b.close()
```
nilsnolde, line 147: jumping straight to the one single `uint64_t` we're interested in and reading it into the TileHeader bit fielded struct

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,179 @@
...
+            # create the traffic tile
+            size = TRAFFIC_HEADER_SIZE + TRAFFIC_SPEED_SIZE * tile_header.directededgecount_
+            tarinfo = tarfile.TarInfo(tile_in.name)
+            tarinfo.size = TRAFFIC_HEADER_SIZE + TRAFFIC_SPEED_SIZE * tile_header.directededgecount_
+            tarinfo.mtime = int(time())
+            tar_traffic.addfile(tarinfo, BytesIO(b'\0' * size))
+
+    index_fd.close()
+
+    LOGGER.info(f"Finished creating the traffic extract at {traffic_fp}")
```
nilsnolde, line 160: oops, I forgot to actually overwrite the index.bin with the right offsets..

src/baldr/graphreader.cc
```diff
@@ -36,22 +42,59 @@ tile_gone_error_t::tile_gone_error_t(std::string prefix, baldr::GraphId edgeid)
...
+          std::make_pair(std::to_string(entry.tile_id),
+                         std::make_pair(const_cast<char*>(file_begin + entry.offset), entry.size)));
+      if (tar_type == tar::TarType::kTiles) {
+        tiles.emplace(std::piecewise_construct, std::forward_as_tuple(entry.tile_id),
+                      std::forward_as_tuple(const_cast<char*>(file_begin + entry.offset),
+                                            entry.size));
+      } else if (tar_type == tar::TarType::kTraffic) {
+        traffic_tiles.emplace(std::piecewise_construct, std::forward_as_tuple(entry.tile_id),
+                              std::forward_as_tuple(const_cast<char*>(file_begin + entry.offset),
+                                                    entry.size));
```
nilsnolde, line 70: a new `TarType` enum which can be passed to the `tar` constructor and further to this lambda so we can decide which container to write to

valhalla/midgard/sequence.h
```diff
@@ -658,7 +659,21 @@ struct tar {
...
 
-  tar(const std::string& tar_file, bool regular_files_only = true)
+  std::string tar_file;
+  mem_map<char> mm;
+  using entry_name_t = std::string;
+  using entry_location_t = std::pair<const char*, size_t>;
+  std::unordered_map<entry_name_t, entry_location_t> contents;
+  size_t corrupt_blocks;
+
+  enum class TarType { kTiles = 0, kTraffic = 1 };
```
nilsnolde, line 669: maybe there's a cleaner way to do this?

nilsnolde:
alright, that's mostly it except for tests. the python script accepts a `--with-traffic` flag now, builds the empty skeleton for traffic.tar with the right binary size and the c++ `tar` constructor reads it nicely, e.g.

`./scripts/valhalla_build_extract.py --config site/valhalla.json --with-traffic -vv` (will rename again without the extension, also the cause for CI failures)

scripts/valhalla_build_extract.py
```diff
@@ -0,0 +1,183 @@
...
+    tarinfo.type = tarfile.REGTYPE
+
+    return tarinfo
+
+
+def write_index_to_tar(tar_fp_: Path):
+    """Loop through all tiles and write the correct index.bin file to the tar"""
+    # get the offset and size from the tarred tile members
+    index: List[Tuple[int, int, int]] = list()
+    with tarfile.open(tar_fp_, 'r|') as tar:
```
nilsnolde, line 84: On second look: get rid of the write further down by using tarfile‘s fileobj in this loop which is already an open file descriptor to the tar file. so we don’t need to keep the index in memory, I think we can write it dynamically while parsing the members.

nilsnolde: well that didn't pan out, guess it's getting confused that its fileobj is changing underneath while it's traversing the files inside the tar..

nilsnolde:
@kevinkreiser you were worried that the tar(s) will be mmap'ed more than once and it seems that's the case.. I printed the ptr from here: https://github.com/valhalla/valhalla/blob/aacd79a57629f732ddc09515ae3c60c2cd4e032e/valhalla/midgard/sequence.h#L189

on two threads with valhalla_service, all four times when the graph is read from loki & thor, it creates another ptr:

```
0x7ffff020b000
0x7ffff0050000
2021/09/27 14:26:43.923347 [INFO] Tile extract successfully loaded with tile count: 7
2021/09/27 14:26:43.923299 [INFO] Tile extract successfully loaded with tile count: 7
0x7fffe063e000
2021/09/27 14:26:43.923603 [INFO] Tile extract successfully loaded with tile count: 7
0x7fffe0483000
2021/09/27 14:26:43.923850 [INFO] Tile extract successfully loaded with tile count: 7
```

kevinkreiser:
@nilsnolde not really worried, we rely on this even now to be the case. i just thought it would be nice to be sure :smile:

nilsnolde:
I added a tiny c++ test. it's getting closer:)

valhalla/midgard/sequence.h
```diff
@@ -658,7 +659,20 @@ struct tar {
     }
   };
 
-  tar(const std::string& tar_file, bool regular_files_only = true)
+  std::string tar_file;
+  bool success_index = false;
```
nilsnolde, line 663: this is really only there so I can test that the index really did smth.. not sure how else to test that, open for suggestions

valhalla/baldr/graphreader.h
```diff
@@ -922,6 +922,7 @@ class GraphReader {
     std::unordered_map<uint64_t, std::pair<char*, size_t>> traffic_tiles;
     std::shared_ptr<midgard::tar> archive;
     std::shared_ptr<midgard::tar> traffic_archive;
+    bool traffic_from_index = false;
```
nilsnolde, line 925: to tell the lambda whether it should insert tile indices into `tiles ` or `traffic_tiles`

nilsnolde:
hm.. strangely the isochrone test is failing now.. 
```
[FAIL] isochrone
[==========] Running 3 tests from 1 test suite.
[----------] 3 tests from Isochrones
[ RUN      ] Isochrones.Basic
/root/project/test/isochrone.cc:37: Failure
The difference between a.GetArray()[1].GetDouble() and b.GetArray()[1].GetDouble() is 2.0999999996718088e-05, which exceeds 0.00002, where
a.GetArray()[1].GetDouble() evaluates to 52.098908999999999,
b.GetArray()[1].GetDouble() evaluates to 52.098888000000002, and
0.00002 evaluates to 2.0000000000000002e-05.
[  FAILED  ] Isochrones.Basic (1427 ms)
```

I can't see how that's caused by the tiles.. it's using the indexed tar file right now and if I remove the `mjolnir.tile_extract` before for the isochrone's test config, it passes. this suggests that smth is off in the tar file, but this tiny difference seems so strange.. 

anyone got an idea what the cause could be?

PR #3282, opened 2021-08-24
Some Suggestions for Faster Tar Loading

kevinkreiser:
Over in #3281 i wanted to make some suggestions but it was kind of difficult to do inline so i figured id give this a go


kevinkreiser:
somethign got messed up with the history, was using github web ui... anyway opend another pr

PR #3283, opened 2021-08-24
tar callback for index parsing

kevinkreiser:
some suggestions on how to parse the index on the c++ side. none of this code is tested, it probably doesnt compile and its definitely not linted :smile: but the idea i think is sound and the implementation is almost completely

### DIFF
--- a/src/baldr/graphreader.cc
+++ b/src/baldr/graphreader.cc
@@ -58,6 +58,8 @@ GraphReader::tile_extract_t::tile_extract_t(const boost::property_tree::ptree& p
       // store this info for ourselves so we dont have to parse tile names again below
       tiles[entry.tile_id] = inserted.first->second;
     }
+    // hand it back to the tar parser
+    return contents;
   };    
     
   // if you really meant to load it
--- a/src/baldr/graphreader.cc
+++ b/src/baldr/graphreader.cc
@@ -20,6 +20,12 @@ namespace {
 constexpr size_t DEFAULT_MAX_CACHE_SIZE = 1073741824; // 1 gig
 constexpr size_t AVERAGE_TILE_SIZE = 2097152;         // 2 megs
 constexpr size_t AVERAGE_MM_TILE_SIZE = 1024;         // 1k
+    
+struct tile_index_entry {
+  uint32_t tile_id; //just level and tileindex hence fitting in 32bits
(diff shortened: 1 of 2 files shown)


src/baldr/graphreader.cc
```diff
@@ -36,40 +42,32 @@ tile_gone_error_t::tile_gone_error_t(std::string prefix, baldr::GraphId edgeid)
...
+    decltype(midgard::tar::contents) contents;
+    midgard::iterable_t<tile_index_entry> index(position, size);
+    for(const auto& entry : index) {
+      // informs the tar that we parsed the index, even though we write a bogus file name
+      auto inserted = contents.insert(std::to_string(entry.tile_id),
+                        std::make_pair(const_cast<char*>(file_begin + entry.offset), entry.size));
+      // store this info for ourselves so we dont have to parse tile names again below
+      tiles[entry.tile_id] = inserted.first->second;
+    }
+  };
```
kevinkreiser, line 61: i basically reverted the changes in this file and added only this bit and the updated call to the tar constructor below

src/baldr/graphreader.cc
```diff
@@ -36,40 +42,32 @@ tile_gone_error_t::tile_gone_error_t(std::string prefix, baldr::GraphId edgeid)
...
+    
+    // get the info
+    decltype(midgard::tar::contents) contents;
+    midgard::iterable_t<tile_index_entry> index(position, size);
+    for(const auto& entry : index) {
+      // informs the tar that we parsed the index, even though we write a bogus file name
+      auto inserted = contents.insert(std::to_string(entry.tile_id),
+                        std::make_pair(const_cast<char*>(file_begin + entry.offset), entry.size));
+      // store this info for ourselves so we dont have to parse tile names again below
+      tiles[entry.tile_id] = inserted.first->second;
```
nilsnolde, line 59: sounds good @kevinkreiser thanks, much better.. `midgard::iterator_t` sounds awesome:)

I guess it only needs updating to actually return the `contents` to the tar and the right values for `tiles` (offset & size)

kevinkreiser: oh haha yeah i forgot the return haha

kevinkreiser:
@PureW it was targetted at a branch that is a work in progress, not at master 

i always make this mistake too :smile:

PR #3299, opened 2021-08-31
Fix Broken Trivial Unidirectional Routes

kevinkreiser:
While I was lending a hand over in #3274 i added a simple test that relied on the results of unidirecitonal path finding (in the forward direction). The route in question is a single edge starting at the beginning of the edge and going 2/3rds along it. The algorithm currently (for this specific ordering of the input data) returns a two edge route starting at the end of the opposing edge and then making its way to the correct edge and ending 2/3rds along. at least from my quick debug session that is what looked like happened.

i thought we removed node snap edge candidates  that were the full distance along...

### DIFF
--- a/src/thor/unidirectional_astar.cc
+++ b/src/thor/unidirectional_astar.cc
@@ -623,25 +623,47 @@ void UnidirectionalAStar<expansion_direction, FORWARD>::SetOrigin(
                   has_other_edges = has_other_edges || (FORWARD ? !e.end_node() : !e.begin_node());
                 });
 
-  // Check if the origin edge matches a destination edge at the node.
-  auto trivial_at_node = [this, &destination](const valhalla::Location::PathEdge& edge) {
+  /**
+   * Consider the route from 1 to 2 on the following graph:
+   *
+   * 1          2
+   * A----------B
+   *
+   * Here both locations get both directions of the edge AB, but it doesnt make sense to use all of
+   * these candidates in practice. The edge candidate for the origin which are at 100% along the edge
+   * amount to no distance traveled. Similarly the edge candidate for the destination at 0% along
+   * would result in an edge with no distance traveled. In other words when its a node snap, you want
+   * to use only those edge candidates that are leaving the origin and those that are arriving at the
+   * destination. In addition to the node snapping, this route is trivial because it consists of one
+   * edge.
+   *
+   * Now consider the route from 1 to 2 on the following graph:
+   *            2
+   * A----------B
+   *            1
+   * In this case the edge candidates for both locations are node snapped to B meaning both locations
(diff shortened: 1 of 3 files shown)


src/thor/unidirectional_astar.cc
```diff
@@ -623,25 +623,47 @@ void UnidirectionalAStar<expansion_direction, FORWARD>::SetOrigin(
...
     auto p = destinations_percent_along_.find(edge.graph_id());
-    if (p != destinations_percent_along_.end()) {
-      for (const auto& destination_edge : destination.path_edges()) {
-        if (destination_edge.graph_id() == edge.graph_id()) {
-          return true;
-        }
-      }
-    }
-    return false;
+    return p != destinations_percent_along_.end() && edge.percent_along() == p->second;
```
kevinkreiser, line 659: this here was the fix, the comparison of percent along. see, we need to confirm for the super trivial case that the edge candidates are indeed snapped to the exact same side (node) of the same edge. we can compare floats here because when an edge gets node snapped its percent_along is not the result of arithmetic/trig/etc its forced to either 0 or 1 depending on the direction of the edge.

kevinkreiser:
... and indeed we do remove node snap edge candidates that are superfluous unless its possible that we need them to find the node snapped destination edge candidate to complete what im now calling a "super trivial route". that is a route which is just routing from a node in the graph to that same node in the graph.

it turned out our check for this case casted too wide of a net and gave false positives, specifically, in the forward direction if you had a node snapped origin that was on the same edge (regular trivial route) of a NOT node snapped destination, it would see the edges were the same (trivial) but not that it wasnt actually both snapped to the same node (missed the super part). the same phenomenon happened in the reverse astar.

danpat:
@kevinkreiser Looks like `astar_bss.cc` also has this bit of logic - what's the backstory for that, and who's maintaining it?

src/thor/unidirectional_astar.cc
kdiluca: Awesome find & fix @kevinkreiser!

kdiluca (approved):

kevinkreiser:
@danpat i tried the same failure scenario with bss and it doesnt fail because for some reason the expansion happens in a different order, though they of course have different costings (ped vs auto). in both of them they start with the wrong edge, the one we should have rejected because it arrives at the origin with 0 length. but then after that unidirectional astar would keep expanding that path from the useless edge onto the correct one while bss actually starts a new path from the correct one. seems to me that somehow unidirectional astar sorts the uturn path lower than the correct origin edge. i'll write another issue about it.

regarding maintainance i would say that all who are willing are maintainers of the algorithm but @xlqian is the main author

PR #3300, opened 2021-09-01
bss trivial path doesnt use superfluous edge

kevinkreiser:
over in #3299 we found a bug with unidirectional/timedep astar in which it favored a 2 edge route (with the first edge being 0 length) over a single edge trivial route. this was because of a bug in the code that proactively removes noded snapped candidate edges when they are superfluous.

that same bug, exists in astar bss however for some reason that algorithm does not prefer the path with the 0 length edge and returns the path with the single edge trivial route. this pr adds a test proving that the algorithm already worked, updates the algorithm to remove the bug and should (but doesnt yet) figure out what else is wrong with either astar or the bss flavor that makes it not suffer the same consequence given the same initial conditions

### DIFF
--- a/src/thor/astar_bss.cc
+++ b/src/thor/astar_bss.cc
@@ -364,15 +364,15 @@ void AStarBSSAlgorithm::SetOrigin(GraphReader& graphreader,
                   has_other_edges = has_other_edges || !e.end_node();
                 });
 
-  // Check if the origin edge matches a destination edge at the node.
-  auto trivial_at_node = [this, &destination](const valhalla::Location::PathEdge& edge) {
-    auto p = destinations_.find(edge.graph_id());
-    if (p != destinations_.end()) {
-      for (const auto& destination_edge : destination.path_edges()) {
-        if (destination_edge.graph_id() == edge.graph_id()) {
-          return true;
-        }
-      }
+  // its super trivial if both are node snapped to the same end of the same edge
+  // note the check for node snapping is in the if below and not in this lambda
+  auto super_trivial = [this, &destination](const valhalla::Location::PathEdge& edge) {
+    if (destinations_.find(edge.graph_id()) != destinations_.end()) {
+      auto p = std::find_if(destination.path_edges().begin(), destination.path_edges().end(),
+                            [&edge](const auto& path_edge) {
+                              return edge.graph_id() == path_edge.graph_id();
+                            });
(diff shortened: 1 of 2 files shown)


test/gurka/test_route.cc
```diff
@@ -928,4 +928,8 @@ TEST(AlgorithmTestTrivial, unidirectional_regression) {
                                 {"/date_time/value", "2111-11-11T11:11"},
                             });
   gurka::assert::raw::expect_path(result, {"AB"});
+
+  // again with reverse a* search direction
+  result = gurka::do_action(valhalla::Options::route, map, {"A", "3"}, "bikeshare");
+  gurka::assert::raw::expect_path(result, {"AB"});
```
kevinkreiser, line 934: this test doesnt fail anyway even without the changes  to the node snap edge removal. we need to figure out why because it might tell us something else is wrong with unidirectional astar

src/thor/astar_bss.cc
```diff
@@ -364,15 +364,15 @@ void AStarBSSAlgorithm::SetOrigin(GraphReader& graphreader,
...
-      }
+  // its super trivial if both are node snapped to the same end of the same edge
+  // note the check for node snapping is in the if below and not in this lambda
+  auto super_trivial = [this, &destination](const valhalla::Location::PathEdge& edge) {
+    if (destinations_.find(edge.graph_id()) != destinations_.end()) {
+      auto p = std::find_if(destination.path_edges().begin(), destination.path_edges().end(),
+                            [&edge](const auto& path_edge) {
+                              return edge.graph_id() == path_edge.graph_id();
+                            });
+      return p != destination.path_edges().end() && p->percent_along() == edge.percent_along();
```
kevinkreiser, line 375: if we remove the second qualification here by doing the following:

`return p != destination.path_edges().end();` 

then the code is equivalent to before the change (ie it leaves the superfluous edge in as a route origin) but miraculously it doesnt suffer the fate of actually finding the path starting with the superfluous edge. when expansion happens it expands the correct edge as an origin rather than as the second edge in a path starting with the superfluous edge as the origin.

this makes me think there is some kind of problem in unidirectional astar that has the sorting wrong. intuitively the cost of the zero length edge plus the cost of the right edge should be at least the transition cost plus the cost of the right edge, where as the right edge as an origin should be just the cost of that right edge, so it should be less. maybe the double bucket queue has a different granularity for one algorithm vs the other?

PR #3302, opened 2021-09-02
Change sharing criterion to avoid artificial detours

genadz:
Changes:
- use continuous function to calculate sharing coefficient (instead of piecewise constant function);
- change definition of "sharing": it was modified  from "part of edges that a candidate shares" to "part of edges that already chosen path shares with alternative";
- extend alternatives search.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/alternates.cc
+++ b/src/thor/alternates.cc
@@ -180,8 +180,7 @@ bool validate_alternate_by_sharing(std::vector<std::unordered_set<GraphId>>& sha
     }
 
     // throw this alternate away if any of the chosen paths shares more than at_most_shared with it
-    assert(paths[i].back().path_distance > 0);
-    if ((shared_length / paths[i].back().path_distance) > at_most_shared) {
+    if (shared_length > at_most_shared * paths[i].back().path_distance) {
       LOG_DEBUG("Candidate alternate rejected by sharing");
       return false;
     }
--- a/src/thor/alternates.cc
+++ b/src/thor/alternates.cc
@@ -169,20 +169,19 @@ bool validate_alternate_by_sharing(std::vector<std::unordered_set<GraphId>>& sha
 
     // if an edge on the candidate_path is encountered that is also on one of the existing paths,
     // we count it as a "shared" edge
-    float shared_length = 0.f, total_length = 0.f;
+    float shared_length = 0.f;
     for (const auto& cpi : candidate_path) {
       const auto length = &cpi == &candidate_path.front()
                               ? cpi.path_distance
                               : cpi.path_distance - (&cpi - 1)->path_distance;
-      total_length += length;
       if (shared.find(cpi.edgeid) != shared.end()) {
         shared_length += length;
       }
     }
 
-    // throw this alternate away if it shares more than at_most_shared with any of the chosen paths
-    assert(total_length > 0);
-    if ((shared_length / total_length) > at_most_shared) {
+    // throw this alternate away if any of the chosen paths shares more than at_most_shared with it
+    assert(paths[i].back().path_distance > 0);
+    if ((shared_length / paths[i].back().path_distance) > at_most_shared) {
       LOG_DEBUG("Candidate alternate rejected by sharing");
       return false;
     }
--- a/src/thor/alternates.cc
+++ b/src/thor/alternates.cc
@@ -28,24 +28,18 @@ namespace thor {
 float get_max_sharing(const valhalla::Location& origin, const valhalla::Location& destination) {
   PointLL from(origin.path_edges(0).ll().lng(), origin.path_edges(0).ll().lat());
   PointLL to(destination.path_edges(0).ll().lng(), destination.path_edges(0).ll().lat());
-  auto distance = from.Distance(to);
+  float distance = from.Distance(to);
 
   // 10km
-  if (distance < 10000.) {
-    return 0.50;
-  }
-  // 20km
-  else if (distance < 20000.) {
-    return 0.60;
-  }
-  // 50km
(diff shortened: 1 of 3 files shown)


genadz:
## Quantity

### ETA < 10m

| | master | branch |
| --- | --- | --- |
| 1st alternative | 1454 | 1463 |
| 2nd alternative | 918 | 930 |

### 10m < ETA < 30m

| | master | branch |
| --- | --- | --- |
| 1st alternative | 511 | 508 |
| 2nd alternative | 356 | 352 |

### 30m < ETA < 1h

| | master | branch |
| --- | --- | --- |
| 1st alternative | 707 | 754 **(+7%)** |
| 2nd alternative | 545 | 602 **(+10%)** |

### 1h < ETA < 5h

| | master | branch |
| --- | --- | --- |
| 1st alternative | 3537 | 4045 **(+14%)** |
| 2nd alternative | 2360 | 3201 **(+36%)** |

## Quality

Let me share several examples:

### with alternatives only in master

| before | after |
| --- | --- |
| master_1 | branch_1 |
| master_2 | branch_2 |
| master_3 | branch_3 |

As expected, some alternatives that have unreasonable detours were filtered out.

### with alternatives only in branch

| before | after |
| --- | --- |
| master_4 | branch_4 |
| master_1 | branch_1 |

These changes caused by increased `kAlternativeCostExtend` parameter.

### different alternatives were proposed

| before | after |
| --- | --- |
| master_1 | branch_1 |
| master_3 | branch_3 |
| master_2 | branch_2 |

src/thor/alternates.cc
```diff
@@ -28,24 +28,18 @@ namespace thor {
...
-  // 20km
-  else if (distance < 20000.) {
-    return 0.60;
-  }
-  // 50km
-  else if (distance < 50000.) {
-    return 0.65;
-  }
+  if (distance < 10000.f)
+    return 0.6f;
```
genadz, line 35: changed min `at_most_shared` coefficient from `0.5` to `0.6`

src/thor/alternates.cc
```diff
@@ -28,24 +28,18 @@ namespace thor {
...
   // 100km
-  else if (distance < 100000.) {
-    return 0.70;
+  if (distance < 100000.f) {
+    // Uniformly increase 'at_most_shared' value from 0.6 to 0.75 for routes
+    // from 10km to 100km
+    return 0.6f + (kAtMostShared - 0.6f) * (distance - 10000.f) / (100000.f - 10000.f);
```
genadz, line 40: replaced piecewise constant function with continuous one

src/thor/alternates.cc
```diff
@@ -175,20 +169,19 @@ bool validate_alternate_by_sharing(std::vector<std::unordered_set<GraphId>>& sha
...
         shared_length += length;
       }
     }
 
-    // throw this alternate away if it shares more than at_most_shared with any of the chosen paths
-    assert(total_length > 0);
-    if ((shared_length / total_length) > at_most_shared) {
+    // throw this alternate away if any of the chosen paths shares more than at_most_shared with it
+    assert(paths[i].back().path_distance > 0);
+    if ((shared_length / paths[i].back().path_distance) > at_most_shared) {
```
genadz, line 184: this is  important change. What was before:
"sharing value is a part of shared edges of an alternative candidate with some already chosen route"
now is
"sharing value is a part of shared edges of some already chosen route with an alternative candidate".

src/thor/bidirectional_astar.cc
```diff
@@ -28,7 +28,7 @@ constexpr float kThresholdDelta = 420.0f;
 // an upper bound value cost for alternative routes we're looking for. Due to the fact that
 // we can't estimate route cost that goes through some particular edge very precisely, we
 // can find alternatives with costs greater than the threshold.
-constexpr float kAlternativeCostExtend = 1.1f;
+constexpr float kAlternativeCostExtend = 1.2f;
```
genadz, line 31: It doesn't drop performance too much, I will add numbers later

src/thor/alternates.cc
genadz: Let's consider a particular example:
```
A------------------------------------B---------C----------------------------------------D
                                     |         |
                                     |         |
                                     |         |
                                     |         |
                                     E---------F
```
here `ABCD` is an optimal route and `ABEFCD` is an alternative candidate.
Let's assume that `len(AB + CD) / len(ABCD) = 0.85` and `len(AB + CD) / len(ABEFCD) = 0.75`. Current **master** will allow `ABEFCD` to be an alternative despite the fact that `BEFC` detour is obviously unreasonable. This **branch** will reject this alternative due to the fact that `sharing = 0.85 > 0.75 = at_most_shared` value.

kevinkreiser: this was what was originally intended actually... good catch! i should also note though that in another pr we are looking into problems with teh path_distance not being set on the last edge label, which is where the pathinfo gets it. it is my h unch that this is in the case that we find a path using only the forward search tree but i havent proven it yet.

kevinkreiser (dismissed):

src/thor/alternates.cc
genadz: Hm, thanks for the comment,  I will check it

genadz: @kevinkreiser replaced division with multiplication to avoid possible problems

merkispavel:
I guess this is a sign that valhalla's Gods like your changes 
image

merkispavel (approved):
reapprove, looks good

Issue #3286, opened 2021-08-25
Crash of valhalla_build_tiles on OSM planet


mesozoic-drones:
Hello! Some time ago ```valhalla_build_tiles``` core dumped on latest OSM planet. Release tag is 3.1.2. Crash is reproduced on OSM planets approximately since august, 15. 

Callstack:
```
#0  0x00007ff383f737ff in raise () from /usr/lib64/libc.so.6
#1  0x00007ff383f5dc35 in abort () from /usr/lib64/libc.so.6
#2  0x00007ff384b4909b in __gnu_cxx::__verbose_terminate_handler() [clone .cold.1] () from /usr/lib64/libstdc++.so.6
#3  0x00007ff384b4f53c in __cxxabiv1::__terminate(void (*)()) () from /usr/lib64/libstdc++.so.6
#4  0x00007ff384b4f597 in std::terminate() () from /usr/lib64/libstdc++.so.6
#5  0x00007ff384b4f7f8 in __cxa_throw () from /usr/lib64/libstdc++.so.6
#6  0x00007ff384b4b06b in std::__throw_out_of_range_fmt(char const*, ...) [clone .cold.5] ()
   from /usr/lib64/libstdc++.so.6
#7  0x00000000004ece9f in std::vector<valhalla::mjolnir::OSMWayNode, std::allocator<valhalla::mjolnir::OSMWayNode> >::_M_range_check (__n=<optimized out>, this=0x7ffc42def870)
    at /opt/rh/gcc-toolset-9/root/usr/include/c++/9/bits/char_traits.h:365
#8  std::vector<valhalla::mjolnir::OSMWayNode, std::allocator<valhalla::mjolnir::OSMWayNode> >::at (
    __n=<optimized out>, this=0x7ffc42def870) at /opt/rh/gcc-toolset-9/root/usr/include/c++/9/bits/stl_vector.h:1091
#9  valhalla::midgard::sequence<valhalla::mjolnir::OSMWayNode>::iterator::operator valhalla::mjolnir::OSMWayNode (
    this=<optimized out>) at /usr/src/debug/valhalla/midgard/sequence.h:467
#10 valhalla::midgard::sequence<valhalla::mjolnir::OSMWayNode>::iterator::operator* (this=<optimized out>)
    at /usr/src/debug/valhalla/midgard/sequence.h:473
[...]

PR #3303, opened 2021-09-04
Overflow in sequence

molind:
Priority_queue holds indexes to the elements. And after node insert to the output sequence index is incremented and put back to the priority_queue. When we have more than INT_MAX elements it overflows and drops part of points.

Discussed in https://github.com/valhalla/valhalla/issues/3286


kevinkreiser:
the changes look great to me, but it seems one of the unit tests is throwing an exception

kevinkreiser (dismissed):

molind:
Ha. Thats strange.

kevinkreiser (approved):

PR #1805, opened 2019-05-13
move complex restrictions building to before validate.

gknisely:
# Issue

Complex restrictions must be done after adding elevation data because it does not have a true builder to serialize the complex restrictions.

## Tasklist

 - [ ] Review - you must request approval to merge any PR to master
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [x] Update the [changelog](CHANGELOG.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #3304, opened 2021-09-04
Run elevation builder after restrictions builder

nilsnolde:
Just wanted to re-run a graph build from restrictions builder which then started by building elevation. This fixes it by building elevation after restrictions.

### DIFF
--- a/src/mjolnir/util.cc
+++ b/src/mjolnir/util.cc
@@ -363,17 +363,17 @@ bool build_tile_set(const boost::property_tree::ptree& original_config,
     LOG_INFO("Skipping hierarchy builder and shortcut builder");
   }
 
-  // Add elevation to the tiles
-  if (start_stage <= BuildStage::kElevation && BuildStage::kElevation <= end_stage) {
-    ElevationBuilder::Build(config);
-  }
-
   // Build the Complex Restrictions
   // ComplexRestrictions must be done after elevation. The reason is that building
   // elevation into the tiles reads each tile and serializes the data to "builders"
   // within the tile. However, there is no serialization currently available for complex restrictions.
   if (start_stage <= BuildStage::kRestrictions && BuildStage::kRestrictions <= end_stage) {
     RestrictionBuilder::Build(config, cr_from_bin, cr_to_bin);
+  }
+
+  // Add elevation to the tiles
+  if (start_stage <= BuildStage::kElevation && BuildStage::kElevation <= end_stage) {
+    ElevationBuilder::Build(config);
   }
 
   // Validate the graph and add information that cannot be added until full graph is formed.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
    * FIXED: Fix distance value in a 0-length road [#3185](https://github.com/valhalla/valhalla/pull/3185)
    * FIXED: Trivial routes were broken when origin was node snapped and destnation was not and vice-versa for reverse astar [#3299](https://github.com/valhalla/valhalla/pull/3299)
    * FIXED: Tweaked TestAvoids map to get TestAvoidShortcutsTruck working [#3301](https://github.com/valhalla/valhalla/pull/3301)
+   * FIXED: Run elevation builder after restriction builder, as expected [#3304](https://github.com/valhalla/valhalla/pull/3304)
 
 * **Enhancement**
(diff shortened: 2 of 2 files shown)


kevinkreiser:
i feel like there are several things, add historical traffic, adding elevation, adding transit which should all be completely decoupled from the build process since they are optional. some of them already are, elevation should become one too. nothing we need to do now but it certainly would be better if we pulled all thsi optional stuff into stand alone programs  to avoid problems exactly like this one. it could also  potentially simplify the configuration by pulling stuff out and making it arguments of separate programs.. anyway just wanted to throw that idea out there so its not just in my head

src/mjolnir/util.cc
```diff
@@ -376,6 +371,11 @@ bool build_tile_set(const boost::property_tree::ptree& original_config,
     RestrictionBuilder::Build(config, cr_from_bin, cr_to_bin);
   }
 
+  // Add elevation to the tiles
+  if (start_stage <= BuildStage::kElevation && BuildStage::kElevation <= end_stage) {
+    ElevationBuilder::Build(config);
```
kevinkreiser, line 376: i havent verified this but does this use the version of tilebuilder constructor that deserializes the whole tile so that we make sure to put everything back the way it was (including the restrictions? thats the only question i have for this i just havent had time to go verify its the case

nilsnolde: yes, it does deserialize everything. it's a little confusing, the comment says:
https://github.com/valhalla/valhalla/blob/49e4a18d2a55dfc4894cbe1da966c663a303e2dd/src/mjolnir/graphtilebuilder.cc#L154-L160

but further down it looks like it does serialize complex restrictions as well:
https://github.com/valhalla/valhalla/blob/49e4a18d2a55dfc4894cbe1da966c663a303e2dd/src/mjolnir/graphtilebuilder.cc#L217-L220

stale comment? looks like it should be fine no?

kevinkreiser: yeah @nilsnolde im pretty sure it is stale because i seem to remember @gknisely changing the code to store restriction data in a sequence (ie no longer in temporary datastructures) so that we can write them in their own separate phase of the pipeline

kevinkreiser: this bit also seems to mention it though... :thinking: 
https://github.com/valhalla/valhalla/blob/49e4a18d2a55dfc4894cbe1da966c663a303e2dd/src/mjolnir/util.cc#L371-L374

kevinkreiser: this: https://github.com/valhalla/valhalla/pull/1805

kevinkreiser: but i see elevation builder does use `StoreTileData` and indeed: https://github.com/valhalla/valhalla/blob/49e4a18d2a55dfc4894cbe1da966c663a303e2dd/src/mjolnir/graphtilebuilder.cc#L312-L339

kevinkreiser: so  im kind of not following the comments... @dnesbitt61 @gknisely refresh my memory was to why complex restrictions have to be after elevation?

nilsnolde: was supposed to be a low impact fix :sweat_smile: fwiw, I tried a few turn restrictions and they still apply (not sure what exactly qualifies as complex restriction for valhalla, but the ones I tried were simple `via` node TRs). however the md5sum for a tarred extract was different from master. that might be insignificant (?!), though I'd have guessed builds are pretty deterministic (or are there datetimes in headers or smth?). anyways, just my findings while avoiding diving into the code too much;)

gknisely: @kevinkreiser @nilsnolde I think this is the reason.  "Complex restrictions must be done after adding elevation data because it does not have a true builder to serialize the complex restrictions."  which was stated 1805 as you noted above.  I think if you move this, then complex restrictions will be broken.  Correct @dnesbitt61 ?

nilsnolde: ok thanks @gknisely ! if that's so, this tiny bug (probably there forever) should probably not bother us much more.. then I'd personally rather wait until we eventually pull out all those optional extra bits as adding elevation, transit etc to their individual programs as kevin suggested further up

gknisely: @nilsnolde so do you want to close this pr and create a ticket with what @kevinkreiser noted.

nilsnolde: yeah you're right, will do

PR #3307, opened 2021-09-06
fix graphreader instantiations while setting up the workers

nilsnolde:
As discussed with @kevinkreiser, we found that the workers are setting up quite a few graphreader instances where they shouldn't have to, see https://github.com/valhalla/valhalla/pull/3281#issuecomment-913354775.

I think this PR should fix that. does it need testing?

### DIFF
--- a/src/thor/worker.cc
+++ b/src/thor/worker.cc
@@ -70,10 +70,9 @@ thor_worker_t::thor_worker_t(const boost::property_tree::ptree& config,
       bidir_astar(config.get_child("thor")), bss_astar(config.get_child("thor")),
       multi_modal_astar(config.get_child("thor")), timedep_forward(config.get_child("thor")),
       timedep_reverse(config.get_child("thor")), isochrone_gen(config.get_child("thor")),
-      matcher_factory(config, graph_reader), reader(graph_reader), controller{} {
-  // If we weren't provided with a graph reader make our own
-  if (!reader)
-    reader = matcher_factory.graphreader();
+      reader(graph_reader ? graph_reader
+                          : std::make_shared<baldr::GraphReader>(config.get_child("mjolnir"))),
+      matcher_factory(config, reader), controller{} {
 
   // Select the matrix algorithm based on the conf file (defaults to
   // select_optimal if not present)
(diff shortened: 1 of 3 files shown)


src/thor/worker.cc
```diff
@@ -70,10 +70,9 @@ thor_worker_t::thor_worker_t(const boost::property_tree::ptree& config,
       bidir_astar(config.get_child("thor")), bss_astar(config.get_child("thor")),
       multi_modal_astar(config.get_child("thor")), timedep_forward(config.get_child("thor")),
       timedep_reverse(config.get_child("thor")), isochrone_gen(config.get_child("thor")),
-      matcher_factory(config, graph_reader), reader(graph_reader), controller{} {
-  // If we weren't provided with a graph reader make our own
-  if (!reader)
-    reader = matcher_factory.graphreader();
+      reader(graph_reader ? graph_reader
+                          : std::make_shared<baldr::GraphReader>(config.get_child("mjolnir"))),
+      matcher_factory(config, reader), controller{} {
```
kevinkreiser, line 75: this code change is good but, you'll want to reorder the members in the header file so that `reader` is declared before `matcher_factor` otherwise some compilers will warn

kevinkreiser: flip these two lines: https://github.com/valhalla/valhalla/blob/master/valhalla/thor/worker.h#L130-L131 so that reader comes first

kevinkreiser: ah  you did that already hahha, nice!

kevinkreiser:
> does it need testing?

id be happy if you put a print in the graphreader constructor and then started `valhalla_service` and told me that you only saw the print out once. that'd be good enough for me!

src/thor/worker.cc
nilsnolde: yeah my clang did warn ;)

nilsnolde:
> id be happy if you put a print in the graphreader constructor and then started valhalla_service and told me that you only saw the print out once. that'd be good enough for me!

haha ok that I did of course;) since the other PR is not merged yet it also printed only once before this change, but I did check that the graphreader constructor is only called once per module per service thread. 

I'll fix the changelog after circleci is done.

kevinkreiser (approved):

kevinkreiser:
no need for changelog imho

Issue #3298, opened 2021-08-31
Elevation downloader in python?


nilsnolde:
we're just having a use case where we want to load elevation based on arbitrary polygon geometries instead of bbox, so we'd like to implement a small python script instead of the existing bash script. the geometries we want to have elevation for are pretty concave, so bboxes is a disk waste. additionally, we'd accept an [osmium-tool config file](https://docs.osmcode.org/osmium/latest/osmium-extract.html#config-file) as input to extract polygon geometries from the same input one might create OSM extracts with.

would that be interesting for upstream? realistically that'd need a spatial library though, e.g. `shapely` to ease intersection stuff. so a user would need to install that. we could automate that as well as part of the cmake config or so, but IMO it's fair to print a sane message saying `shapely` needs to be manually installed.

cc @chrstnbwnkl

nilsnolde:
just did an extract for half US + 11 canadian cities and saved more than 35% of disk space:
```
2021-09-01 14:23:35,628 DEBUG: Bbox extraction would've yielded 1103 tiles.
2021-09-01 14:23:35,628 INFO: Successfully saved 698 tiles to /custom_files/elevation_data!
```

kevinkreiser:
@nilsnolde what i have advocated for a while (but i cant figure out where) was to just download each srtm tile as the tile builder needs it (ie lazily). in this case we will only ever download exactly the tiles we need and only when we need them (though we should cache them to disk). i did a prototype of this some 3 years ago but the code is lost to time. i think in the meantime a python script is just fine. actually we could replace the existing scripts if you can add the same functionality that exists in them (ie extract them and do it all in parallel).

nilsnolde:
Sure, c++ would be deluxe. Would be nice if we can save that for another day. We’ll polish the script a little then and PR.

nilsnolde:
the only downside is that hyperthreading is not really supported in python. `multiprocessing` can only use physical cores (for me that's 4 instead of 8), and `threading` is not an option anyways with python's GIL.. hope that's not a dealbreaker..

nilsnolde:
[...]

PR #3308, opened 2021-09-07
better scripts testing

nilsnolde:
with all the recent work (and some bugs) in the vital scripts here (valhalla_build_config/_extract etc) in several PRs now I was wondering if we couldn't come up with a sane testing infrastructure to make sure we don't break these scripts along the way. 

I drafted up smth which would work IMO and be easily maintainable with just a light sprinkle of hacky/patchy :smile: . it's just a proof of concept to show how this could be done. in cmake we could simply run those tests with `python -m unittest discover -v -s test/scripts`.

When finalizing the PR I realized I'd have to change `valhalla_build_config` quite a bit as well to make it easier testable. I took the chance and cleaned that script up a bit as well.


test/scripts/helper.py
```diff
@@ -0,0 +1,19 @@
...
+    # Get the right path to the module
+    path = str(CWD.parent.parent.joinpath('scripts', module_name))
+    spec = importlib.util.spec_from_loader(
+        module_name,
+        importlib.machinery.SourceFileLoader(module_name, path)
+    )
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    sys.modules[module_name] = module
+    return module
```
nilsnolde, line 19: thanks to python's convoluted import machinery we have to go to these lengths to import modules which don't have a `.py` suffix

test/scripts/test_valhalla_build_config.py
```diff
@@ -0,0 +1,13 @@
+from unittest import TestCase
+
+try:
+    from test.scripts import helper
+except ModuleNotFoundError:
+    import helper
```
nilsnolde, line 6: again the import machinery's trickiness.. this makes sure test can be run from any project folder as working directory.

nilsnolde: although I don't _quite_ understand why I can run this from e.g. `valhalla/test` directory.. but it does work..

test/scripts/test_valhalla_build_config.py
```diff
@@ -0,0 +1,18 @@
+from unittest import TestCase
+import unittest
+
+try:
+    from test.scripts import helper
+except ModuleNotFoundError:
+    import helper
+
+valhalla_build_config = helper.import_path('valhalla_build_config')
```
nilsnolde, line 9: basically this is the same as saying `from valhalla.scripts import valhalla_build_config`

test/scripts/helper.py
nilsnolde: full disclosure: I also found that the way this can be done changed at least twice in an incompatible way since py3.0 was established.. I do hope this way is a little longer lasting..

scripts/valhalla_build_config
```diff
@@ -1,35 +1,36 @@
...
 import json
-import pprint
+from typing import List, Union
 
-class optional:
-  def __init__(self, t=None):
-    self.type = t
 
-#global default configuration
+class Optional:
```
nilsnolde, line 7: just more pythonic to have capital class names

scripts/valhalla_build_config
```diff
@@ -557,56 +558,67 @@ help_text = {
...
+def override_config(args_: dict, leaves_: list, config_: dict):
+    """override the defaults given what was passed"""
+    for leaf in leaves_:
+        keys = leaf.split('\0')
+        v = config_
+        for k in keys[:-1]:
+            v = v[k]
+        v[keys[-1]] = args_.get(leaf.replace('\0', '_'))
+        if isinstance(v[keys[-1]], type(Optional())):
+            del v[keys[-1]]
```
nilsnolde, line 607: I broke out this stuff into its own function so it's easier to test

scripts/valhalla_build_config
```diff
@@ -557,56 +558,67 @@ help_text = {
...
-def parse_bool(arg):
-  if arg.lower() in ('yes', 'true', 't', 'y', '1'):
-    return True
-  elif arg.lower() in ('no', 'false', 'f', 'n', '0'):
-    return False
-  else:
-    raise argparse.ArgumentTypeError('Boolean value expected.')
+def comma_separated_list(arg: str) -> List[str]:
+    """used to convert comma separated list of strings into a list"""
+    return arg.split(',')
```
nilsnolde, line 564: really hard to see what's going on in this diff.. 

- I just moved the comments above the functions into their docstrings
- added the type hints

scripts/valhalla_build_config
```diff
@@ -557,56 +558,67 @@ help_text = {
...
+def parse_bool(arg: str):
+    if arg.lower() in ('yes', 'true', 't', 'y', '1'):
+        return True
+    elif arg.lower() in ('no', 'false', 'f', 'n', '0'):
+        return False
+    else:
+        raise ArgumentTypeError('Boolean value expected.')
+
+
+def add_leaf_args(path: str, tree: Union[dict, bool, str, list, Optional], leaves_: List[str], parser_: ArgumentParser, help: dict):
```
nilsnolde, line 576: e.g. here I renamed `leaves` to `leaves_`, `parser` to `parser_` to not shadow those outer scope vars

test/scripts/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+find_package(Python COMPONENTS Interpreter)
+add_custom_command(OUTPUT scripts.log
+    COMMAND
+      LOCPATH=${VALHALLA_SOURCE_DIR}/locales
+      ASAN_OPTIONS=verify_asan_link_order=0:detect_leaks=0
+      /bin/bash -cx "${Python_EXECUTABLE} -m unittest discover -s ${VALHALLA_SOURCE_DIR}/test/scripts -v > ${CMAKE_BINARY_DIR}/test/scripts.log 2>&1"
```
nilsnolde, line 6: for some reason `unittest` prints to stderr

test/scripts/helper.py
```diff
@@ -0,0 +1,19 @@
...
```
nilsnolde, line 19: a helper to be able to import modules whose file suffix are not ending in `.py` like `valhalla_build_config` (with `valhalla_build_config.py` as filename we could just use `from ...scripts import valhalla_build_config`)

test/scripts/test_valhalla_build_config.py
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        leaves = list()
+        valhalla_build_config.add_leaf_args('', config, leaves, mock_parser, help_text)
+
+        # make sure the parser.add_argument was called with the right stuff while parsing the config
+        for name, default in config.items():
+            mock_parser.add_argument.assert_any_call(
+                f'--{name.replace("_", "-")}', type=value_types[name], help=help_text[name], default=default
+            )
```
nilsnolde, line 51: testing if `parser.add_argument()` gets  called with the right arguments for each of the config members

nilsnolde:
I'm finished with this for now. Looks like lots of changes, but it's mostly just a few things:
- make generally more pythonic (4 spaces "tabs", type hints for all kinds of signatures, capital class name etc)
- take out more logic from the `if __name__ == '__main__'` block and put it into functions so it's easier to test
- bit of testing infrastructure for the scripts stuff in `test/scripts` and CMakeLists, also looking towards more scripts implemented in python (e.g. #3298, #3281 )
- added the `run-scripts` target to the `check` target

test/scripts/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+find_package(Python COMPONENTS Interpreter)
```
nilsnolde, line 1: I guess we should bail if python is not found..

test/scripts/test_valhalla_build_config.py
```diff
@@ -0,0 +1,87 @@
...
+        leaves = list()
+        valhalla_build_config.add_leaf_args('', config, leaves, mock_parser, help_text)
+
+        # create the args mock
+        args = {'0_str': 'new string', "0_opt_str": valhalla_build_config.Optional(str), "0_1_int": 5}
+
+        valhalla_build_config.override_config(args, leaves, config)
+        # the Optional arg should be removed
+        self.assertTrue(len(config["0"]), 2)
+        self.assertEqual(config, {"0": {"str": "new string", "1": {"int": 5}}})
```
nilsnolde, line 83: that's the actual test, the rest is just preparing it

test/scripts/CMakeLists.txt
nilsnolde: done

scripts/valhalla_build_config
```diff
@@ -534,54 +535,65 @@ help_text = {
...
+        # lists are supported as comma separated and bools support a bunch of conventions
+        # the rest of the primatives (strings and numbers) parse automatically
+        if isinstance(tree, Optional):
+          t = tree.type
+        elif isinstance(tree, list):
+          t = lambda arg: arg.split(',')
+        elif isinstance(tree, bool):
+          t = lambda arg: bool(strtobool(arg))
+        else:
+          t = type(tree)
```
nilsnolde, line 566: I found this more idiomatic than before

scripts/valhalla_build_config
```diff
@@ -534,54 +535,65 @@ help_text = {
...
-  if isinstance(tree, dict):
-    for k in tree:
-      v = tree[k]
-      add_leaf_args('\0'.join([path, k]) if len(path) else k, v, leaves, parser)
-  #we've reached a leaf
-  else:
-    keys = path.split('\0')
-    h = help_text
-    for k in keys:
-      h = h[k]
```
nilsnolde, line 550: really hard to see what's going on in this diff.. I hardly changed the logic

- I just moved the comments above the functions into their docstrings
- added the type hints
- renamed some variables to not shadow those from the outer scope (even though it's exactly those which they're referencing, I find it less confusing)

scripts/valhalla_build_config
```diff
@@ -534,54 +535,65 @@ help_text = {
...
+parser = ArgumentParser(description='Generate valhalla configuration', formatter_class=ArgumentDefaultsHelpFormatter)
+add_leaf_args('', config, leaves, parser, help_text)
+
+# entry point to program
+if __name__ == '__main__':
+    # TODO: add argument to set base path and use in all other path based values
+    args = parser.parse_args()
+
+    # encapsulate to test easier
+    override_config(args.__dict__, leaves, config)
```
nilsnolde, line 596: `args` is a argparser.Namespace, `__dict__` contains all k, v of the arguments and nothing else. again, easier to test that function this way

nilsnolde:
@kevinkreiser `ln -s $PWD/scripts/valhalla_build_config $PWD/scripts/valhalla_build_config.py` worked :) now I could get rid of that helper script and just import normally :+1:

test/scripts/valhalla_build_config.py
```diff
@@ -0,0 +1 @@
+/home/nilsnolde/dev/cpp/valhalla/scripts/valhalla_build_config
```
nilsnolde, line 1: this is the symlink

kevinkreiser: of course this wont work on CI. i think you need to link it like this with relative paths

```bash
# clean up bad link
cd /home/nilsnolde/dev/cpp/valhalla/test/scripts
rm valhalla_build_config.py
# add new relative link
ln -s ../../scripts/valhalla_build_config valhalla_build_config.py
```

nilsnolde: 😂 didn’t even realize that..

kevinkreiser:
i think its fine skip the coverage for now. we didnt have coverage of these files before. at least now we have a test that would fail if its broken.

PR #3312, opened 2021-09-08
In graphtile.h excessive commit removed

Neyromancer:
In graphtile.h excessive commit removed

### DIFF
--- a/valhalla/baldr/graphtile.h
+++ b/valhalla/baldr/graphtile.h
@@ -523,11 +523,11 @@ public:
    *                       if nullptr is passed in flow_sources does nothing.
    * @return Returns the speed for the edge.
    */
-  inline uint32_t GetSpeed(const DirectedEdge* de,
-                           uint8_t flow_mask = kConstrainedFlowMask,
-                           uint32_t seconds = kInvalidSecondsOfWeek,
-                           bool is_truck = false,
-                           uint8_t* flow_sources = nullptr) const {
+  uint32_t GetSpeed(const DirectedEdge* de,
+                    uint8_t flow_mask = kConstrainedFlowMask,
+                    uint32_t seconds = kInvalidSecondsOfWeek,
+                    bool is_truck = false,
+                    uint8_t* flow_sources = nullptr) const {
     // if they dont want source info we bind it to a temp and no one will miss it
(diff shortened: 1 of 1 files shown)


kevinkreiser:
@Neyromancer can you maybe explain what you are trying to accomplish here? This method is very critical when it comes to performance as it's in the tight loop for the routing computation. Removing the inline will reduce performance but will probably also make the binary size a bit smaller. Any hints as to what problem you are solving here would be helpful.

Neyromancer:
sure, as the method GetSpeed is entirely defined inside the class GraphTile we do not need to use inline specifier as the method is inlined anyway. 
check here
https://en.cppreference.com/w/cpp/language/inline#:~:text=A%20function%20defined,since%20C%2B%2B20).

nilsnolde:
that's also valid für c++14 (valhalla std)? seems to mention that's only the case in c++20

maksimandrianov:
> that's also valid für c++14 (valhalla std)? seems to mention that's only the case in c++20

only this the case in c++20:
```
 if it is attached to the global module (since C++20)
```

kevinkreiser (changes requested):
yeah we are no where near using modules in c++20 so the `inline` is required. i kind of think that it gets inlined anyway regardless of version of c++.. but having the keyword doesnt hurt us. unless there is a good reason i dont think we should merge this pr

kevinkreiser:
i dont think this applies only to c++20 but also the keyword doesnt hurt anything. i dont have the time righ tnwo to prove that there is no effect to performance. if someone else does then be my guest


https://isocpp.org/wiki/faq/inline-functions#inline-member-fns-more

Neyromancer:
I guess, i could do this. @kevinkreiser, what would be the best way to do this?

kevinkreiser:
@Neyromancer i guess i wonder why is this work important to you? Do you have a compiler that warns about it? It seems to be an extreme nitpick. If you do want to test it would talk to someone like @genadz about getting a dataset with at least historical traffic in it and then also how to bench mark the code with it.

Neyromancer:
@kevinkreiser, actually it thought to close this pr but just wanted to try to test, only of pure curiosity. I will ask @genadz, thank you for reference.

PR #3314, opened 2021-09-10
nit: Uses switch statements for ensuring full coverage

purew:
Curious if there's any downside to an approach like this with a switch statement. 

Upside is that the compiler would yell at us if we forgot one of the enum values.

### DIFF
--- a/valhalla/baldr/directededge.h
+++ b/valhalla/baldr/directededge.h
@@ -713,19 +713,13 @@ public:
   bool unpaved() const {
     switch (surface()) {
       case Surface::kPavedSmooth:
-        return false;
       case Surface::kPaved:
-        return false;
       case Surface::kPavedRough:
         return false;
       case Surface::kCompacted:
-        return true;
       case Surface::kDirt:
-        return true;
       case Surface::kGravel:
-        return true;
       case Surface::kPath:
-        return true;
       case Surface::kImpassable:
         return true;
     }
--- a/valhalla/baldr/directededge.h
+++ b/valhalla/baldr/directededge.h
@@ -711,7 +711,24 @@ public:
    * Is this edge unpaved or bad surface?
    */
   bool unpaved() const {
-    return (surface() >= Surface::kCompacted);
+    switch (surface()) {
+      case Surface::kPavedSmooth:
+        return false;
+      case Surface::kPaved:
+        return false;
+      case Surface::kPavedRough:
+        return false;
+      case Surface::kCompacted:
+        return true;
+      case Surface::kDirt:
+        return true;
+      case Surface::kGravel:
+        return true;
+      case Surface::kPath:
+        return true;
+      case Surface::kImpassable:
+        return true;
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ktatterso (approved):

mandeepsandhu (commented):
+1 from me. As long as someone doesn't sneak in a `default` case, the compiler should catch unhandled enums.

dnesbitt61 (commented):
What was wrong with the prior code? Seems like a switch would be much less efficient than prior code were this to be used in costing.

PureW:
> What was wrong with the prior code? Seems like a switch would be much less efficient than prior code were this to be used in costing.

I'll run the gbench tests on the astar loop once I get this building [locally](https://github.com/valhalla/valhalla/pull/3120#discussion_r706329653).

My hope was that this wouldn't impact performance while giving more protections against someone tweaking the enum and then forgetting to update those conditionals or `std::unordered_map` lookups.

dnesbitt61:
unpaved is not used in any costing methods as far as I can tell so you likely won't see a difference. Nobody should tweak this enum since there are only 8 values allowed by the 3 bits that are used for surface type in the DirectedEdge. Someone could alter the mappings of OSM tags to these values, but we cannot add any more values to this enum without breaking the tile format.

dnesbitt61:
I take that back - it seems like it is used in costing.

kevinkreiser:
The string method doesn't matter but the branches in the unpaved method will result in some performance loss but it's likely unmeasurable when combined with the rest of the routing computation

kevinkreiser:
I would also say though that we don't miss any enum values in the costing use case so the switch statement is superfluous there imho. Surely the compiler didn't complain about that code anyway

kevinkreiser:
oh you are saying you want it to complain when we add one. well heres the thing about that, the equality has to always work, for one the thing is unsigned so we can only add increasing values. secondly we cannot reorder the values because they are built into the data and old code will expect them to be the same values. thirdly we cannot even add more here that would be considered paved because again old code expects anything "paved" to be less than a certain amount.

dnesbitt61:
I would just add an exception in set_surface where if the value exceeds the 3 bits allowed then it throws the exception.

PureW:
Gotcha, so in practice, the `Surface` enum is closed, i.e. we can't add or change it. That removes most of my worry.

What about simply keeping the change at https://github.com/valhalla/valhalla/pull/3314/files#diff-335903af8df761a6bae5f8f0aeca979d8b1980280d71e08e030ebb3e68c6cc5bR457-R473 which seems like a pure improvement and leaving the conditional as is in that case?

kevinkreiser:
> What about simply keeping

fine with me but i would warn that people can send values to that function other than those in the enum by casting, so you might want to handle the case when an unknown value appears

maksimandrianov:
If we imagine that enum can be more than 3 bits, could you give an example of when would adding a new enum value cause an error in logic in this code? 
Values of surface enum are ordered. The bigger values, the worse surfaces. This rule is used in other parts of the project, for example https://github.com/valhalla/valhalla/blob/master/src/sif/motorscootercost.cc#L379

kevinkreiser:
@maksimandrianov exactly we cannot reorder it and we cannot add values to it. As @PureW said, it's locked at this point

Issue #3316, opened 2021-09-10
Build Problems on Newer GCC: static_assert(is_destructible


kevinkreiser:
Disccussion is here: https://github.com/valhalla/valhalla/pull/3120#discussion_r706329653

kevinkreiser:
so i took a quick look at this and it seems the presence of a destructor in `tile_data` causes `is_destructable` to be `false`. this is highly unintuitive but :man_shrugging: anyway seems like we can fix this if we can figure out a way to not have a user defined constructor.. i think i have an idea for that but its kind of hacky. ill draft a pr and see what you all think

molind:
gcc with libstdc++ 11.1.0 throws this error if `tile_data` forward declared before cache_t
https://godbolt.org/z/8sYhf8Wqb
and only in 11+ versions. gcc < 11 compiles it without errors.

while clang 12.0.1 with libc++ works fine. -stdlib=libc++
https://godbolt.org/z/vPYTdb3aW

I'll send bug report to libstd++ and try to reorder declarations to fix this bug.

nilsnolde:
Fixed in #3317 :)

PR #3317, opened 2021-09-11
Reordered code to fix gcc bug

molind:
Moved tile_data declaration up, to avoid forward declaration before std::future.

Discussed in: https://github.com/valhalla/valhalla/issues/3316

### DIFF
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -276,7 +276,6 @@ struct cache_t {
   tile_data source(uint16_t index);
 };
 
-
 tile_data::tile_data(cache_t* c, uint16_t index, bool reusable, const int16_t* data)
     : c(c), data(data), index(index), reusable(reusable) {
   if (reusable)
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -161,34 +161,6 @@ public:
   }
 };
 
-class tile_data;
-
-struct cache_t {
-  // Cached tiles
-  std::vector<cache_item_t> cache;
-  // Set of reusable tile indexes
-  std::unordered_set<uint16_t> reusable;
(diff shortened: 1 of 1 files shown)


molind:
@nilsnolde @PureW Could you check https://github.com/GLMap/valhalla/commit/3fbd2578bb5cbc606a124c9bfc91fa6f1fb15fde please?
BTW If arch is one of valhalla targets may be arch-build should be added as one of tests.

nilsnolde:
Thanks a lot @molind , I really appreciate it! Wouldn’t have known where to start debugging.. Yeah it happens here and there, mostly with some dependencies. CI checks would be great, also for the project, to catch possible future errors early on. I’ll look into either arch or fedora.

I’ll check this branch once I’m home

kevinkreiser:
Interesting. I had made a small test program and I thought avoiding the forward declare was the first thing I tried. With my small test program it seemed just having a destructor caused it not to compile

molind:
@kevinkreiser that's was my first check too. https://godbolt.org is really god for that tasks:
```
#include <future>

struct test_data;
struct test_data2 {};

class test {
    // error
    std::promise<test_data> a;
    // OK
    std::promise<test_data2> b;
};

struct test_data {};

int main() {
    test a;
}
```
It seems they broke libstd++ on gcc >= 11. gcc 10.* compiles this example without an error.

kevinkreiser (approved):

nilsnolde:
little late, but yeah, that fixed it @molind , thanks a bunch!

molind:
After clarification from libstdc++ mailing list, that's was my mistake actually to try to create template objects based on forward declared structs, but "it compiles!". :) Learn something new every day.

PR #3318, opened 2021-09-13
Improved elevation tile downloading in Python

### LINKED ISSUE #3298: Elevation downloader in python?

chrstnbwnkl:
# Issue

fixes #3298 by replacing `valhalla_build_elevation` bash script with a corresponding Python script, which
- adds shapely support while still supporting the bbox method
- uses hyperthreading
- accepts a directory path as an argument that will be parsed for GeoJSON files
- handles Polygons and MultiPolygons

## Tasklist

 - [ ] Add tests

## Requirements / Relations
Waiting to merge #3308 so this can be properly tested


nilsnolde:
we already had some rounds internally reviewing this. until final review I'd ideally like to merge #3308 before. anyone chiming in is appreciated!

scripts/valhalla_build_elevation
```diff
@@ -1,81 +1,224 @@
...
+        or max_y > 90
+    ):
+        return False
+
+    return True
+
+
+def download(tile: Tile, output_dir, decompress):
+    dest_directory = Path(output_dir, tile.dir)
+    Path.mkdir(dest_directory, parents=True, exist_ok=True)
```
nilsnolde, line 165: Just realized we either need to lock the mkdir call here or create all parent directories first in the main thread @chrstnbwnkl . Preferably use a lock, somewhat nicer to also test eventually IMO

chrstnbwnkl: I just added a lock, but thinking about it further, I think we might not need it since `exist_ok=True` catches any `OSError` that occur when creating a directory that already exists (see source [here](https://github.com/python/cpython/blob/80d9ff16483b6c1898bcdcc811b5450b57a5e573/Lib/pathlib.py#L1168). If we want to further make sure this `OSError` does not occur for other reasons, we could further inspect it as described [here](https://stackoverflow.com/a/273227/10955832). What do you think?

nilsnolde: > I think we might not need it since exist_ok=True catches any OSError that occurs when creating a directory that already exists

yea you're right! then we revert that change. fwiw, the lock wouldn't have an effect here since it's created for each thread separately. it's important that the lock is shared between the threads. smth like this (answer is for processes, but might hold for threads): https://stackoverflow.com/a/25558333/2582935

nilsnolde:
@chrstnbwnkl #3308 is merged, you can look there to see how testing works for these scripts: the most important part is that you relatively symlink the script you wanna test into the folder where the test file is with a `.py ` extension, then you can import the script you wanna test easily

dnesbitt61:
I wrote that a couple years ago as a way of reducing the elevation data size by not downloading data over oceans and places where no roads exist. This reduced the size of worldwide elevation from ~1.5TB to ~500GB. It would be great to add this logic to valhalla_build_elevation as an option.

nilsnolde:
I added the option to only download tiles the graph covers to `valhalla_build_elevation` and re-arranged a little but didn't change much more @chrstnbwnkl . was in a bit of a hurry, we want to use that feature for the fossgis server, so we only download 500 gigs.

should be ready for review. CI will fail, we first need the docker image with python3-shapely installed. I'll PR real quick

scripts/valhalla_build_elevation
```diff
@@ -1,81 +1,249 @@
...
+    local_dir = graph_dir.joinpath('1')
+    tile_width = int(360 / LOCAL_SIZE)
+    for tile_fp in local_dir.rglob('*.gph'):
+        # turn the path into a tile ID
+        tile_id = int(str(tile_fp.parent.relative_to(local_dir).joinpath(tile_fp.stem)).replace(os.sep, ''))
+        s, w = (tile_id % tile_width) * LOCAL_SIZE - 180, int(tile_id / tile_width) * LOCAL_SIZE - 90
+
+        tiles.append(get_tile_info(s, w))
+
+    return tiles
```
kevinkreiser, line 185: i love that someone is finally implementing this idea however this doesnt look right to me. for this to work you need to turn the gph tile into a bounding box and see which srtm tiles it intersects.to mee it looks like you just turn the tile id into its index, thats a good start, but you need to take the tile level into account because that determins what size it is. you cannot just look at at level 1 tiles because they only hold middle class roads. you will be missing places most likely. do you want me to share a snippet?

nilsnolde: ah yeah I did mess smth up in my brain.. somehow my line of thought was: if there's a 2nd level tile, there'll definitely be the 1st level tile. but that's not correct of course (but the other way around is correct: there'll always have to be a 2nd level tile if there's features in any other level, for the bins). other than that the logic is correct, we don't need really a bbox in this case bcs 1st level graph and elevation tiles have the same dimension and conventions, so turning the tile ID into a lower-left corner is actually enough to get the correct elevation tile.

I'll turn it into 2nd level logic.

scripts/valhalla_build_elevation
```diff
@@ -1,81 +1,249 @@
...
+    tile_infos = set()
+    local_dir = graph_dir.joinpath('2')
+    tile_width = int(360 / LOCAL_SIZE)
+    for tile_fp in local_dir.rglob('*.gph'):
+        # turn the path into a tile ID
+        tile_id = int(str(tile_fp.parent.relative_to(local_dir).joinpath(tile_fp.stem)).replace(os.sep, ''))
+        s, w = floor(int(tile_id % tile_width) * LOCAL_SIZE - 180), floor(int(tile_id / tile_width) * LOCAL_SIZE - 90)
+
+        tile_infos.add(get_tile_info(s, w))
+
```
nilsnolde, line 184: I think this should be correct now @kevinkreiser : we floor the lower-left corner of local tiles giving us the integer degree which corresponds to the elevation tile notation. it's now a `set` so we don't collect dups.

test/scripts/test_valhalla_build_elevation.py
```diff
@@ -0,0 +1,77 @@
...
+                    "geometry": {
+                        "type": "Polygon",
+                        "coordinates": [[[0, 0], [0, 3], [1, 3], [1, 1], [3, 1], [3, 0], [0,0]]]
+                    }
+                 }
+            ]
+        }
+        gj_fp = TILE_DIR.joinpath('test.geojson')
+        with open(gj_fp, 'w') as f:
+            json.dump(gj, f)
```
nilsnolde, line 46: should be deleted after the test too. I'll add that before merging

kevinkreiser (approved):

PR #3324, opened 2021-09-14
Speed up nearest-neighbor searches when large search cutoff is used

danpat:
# Issue

If "someone" (could be anyone, definitely not me) happened to want to use a very large search cutoff with Loki, and placed a coordinate a long way from any actual edge, it can take Loki quite a while to snap to the nearest edge.

Doing some basic work with a profiler, it looks like there were a lot of small allocations/deallocations, and some unnecessary floating point math.

Performance improvements will be negligible for most requests.  However, the problem can be highlighted by making a request with `search_cutoff: 21000000` (approx half the planet), and using Point Nemo (-123.45,-48.89, furthest point from land) against OSM data:

## Before
```
$ time curl -d '{"locations":[{"lon":-123.45,"lat":-48.89,"radius":0,"search_cutoff":21000000},{"lon":-123.45,"lat":-48.89,"radius":0,"search_cutoff":21000000}],"costing":"auto"}' http://localhost:8002/route
real	0m4.073s
user	0m0.003s
sys	0m0.007s
```

## After
```
$ time curl -s -d '{"locations":[{"lon":-123.45,"lat":-48.89,"radius":0,"search_cutoff":21000000},{"lon":-123.45,"lat":-48.89,"radius":0,"search_cutoff":21000000}],"costing":"auto"}' http://localhost:8002/route
real	0m1.238s
user	0m0.004s
sys	0m0.006s
```

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/src/midgard/tiles.cc
+++ b/src/midgard/tiles.cc
@@ -55,7 +55,8 @@ template <class coord_t> struct closest_first_generator_t {
   robin_hood::unordered_set<int32_t> queued;
   int32_t subcols, subrows;
   using best_t = std::pair<double, int32_t>;
-  std::priority_queue<best_t, std::vector<best_t>, std::function<bool(const best_t&, const best_t&)>> queue;
+  std::priority_queue<best_t, std::vector<best_t>, std::function<bool(const best_t&, const best_t&)>>
+      queue;
 
   // re-usable, pre-allocated vector used by dist
   std::vector<coord_t> corners;
--- a/src/midgard/tiles.cc
+++ b/src/midgard/tiles.cc
@@ -11,6 +11,8 @@
 #include "midgard/tiles.h"
 #include "midgard/util.h"
 
+#include <robin_hood.h>
+
 namespace {
 
 // this is modified to include all pixels that are intersected by the floating point line
@@ -51,10 +53,13 @@ void bresenham_line(double x0,
 template <class coord_t> struct closest_first_generator_t {
   valhalla::midgard::Tiles<coord_t> tiles;
   coord_t seed;
-  std::unordered_set<int32_t> queued;
+  robin_hood::unordered_set<int32_t> queued;
   int32_t subcols, subrows;
   using best_t = std::pair<double, int32_t>;
   std::set<best_t, std::function<bool(const best_t&, const best_t&)>> queue;
+
+  // re-usable, pre-allocated vector used by dist
+  std::vector<coord_t> corners;
 
   closest_first_generator_t(const valhalla::midgard::Tiles<coord_t>& tiles, const coord_t& seed)
       : tiles(tiles), seed(seed), queued(100), queue([](const best_t& a, const best_t& b) {
@@ -70,6 +75,8 @@ template <class coord_t> struct closest_first_generator_t {
     queued.emplace(subdivision);
     queue.emplace(std::make_pair(0, subdivision));
     neighbors(subdivision);
+    constexpr int max_tested_corners = 8;
+    corners.reserve(max_tested_corners);
   }
 
   // something to measure the closest possible point of a subdivision from the given seed point
@@ -81,7 +88,11 @@ template <class coord_t> struct closest_first_generator_t {
     auto y0 = tiles.TileBounds().miny() + y * tiles.SubdivisionSize();
     auto y1 = tiles.TileBounds().miny() + (y + 1) * tiles.SubdivisionSize();
     auto distance = std::numeric_limits<double>::max();
-    std::list<coord_t> corners{{x0, y0}, {x1, y0}, {x0, y1}, {x1, y1}};
+    corners.clear();
+    corners.emplace_back(x0, y0);
+    corners.emplace_back(x1, y0);
+    corners.emplace_back(x0, y1);
+    corners.emplace_back(x1, y1);
     if (x0 < seed.first && x1 > seed.first) {
       corners.emplace_back(seed.first, y0);
       corners.emplace_back(seed.first, y1);
@@ -91,7 +102,7 @@ template <class coord_t> struct closest_first_generator_t {
       corners.emplace_back(x1, seed.second);
     }
     for (const auto& c : corners) {
-      auto d = seed.Distance(c);
+      auto d = seed.DistanceSquared(c);
       if (d < distance) {
         distance = d;
       }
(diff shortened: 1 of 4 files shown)


src/midgard/tiles.cc
```diff
@@ -70,6 +75,8 @@ template <class coord_t> struct closest_first_generator_t {
     queued.emplace(subdivision);
     queue.emplace(std::make_pair(0, subdivision));
     neighbors(subdivision);
+    constexpr int max_tested_corners = 8;
```
mandeepsandhu, line 78: Why is this 8? we're testing 4 corners, right? or maybe I'm missing something?

mandeepsandhu: never mind, I see we insert _at most_ 8.

danpat: I dunno, I just counted the max number of things that were being added in the `dist()` function, and reserved that amount of space here.

danpat: See here: https://github.com/valhalla/valhalla/pull/3324/files#diff-57cf239a14a55d61c34bbd9378f74eaebafeba721c2c466058337088192863d3L84

src/midgard/tiles.cc
```diff
@@ -81,7 +88,11 @@ template <class coord_t> struct closest_first_generator_t {
     auto y0 = tiles.TileBounds().miny() + y * tiles.SubdivisionSize();
     auto y1 = tiles.TileBounds().miny() + (y + 1) * tiles.SubdivisionSize();
     auto distance = std::numeric_limits<double>::max();
-    std::list<coord_t> corners{{x0, y0}, {x1, y0}, {x0, y1}, {x1, y1}};
+    corners.clear();
```
ktatterso, line 91: Another option would be to remove the `corners` container altogether.  Just have one `coord_t corner` and unroll the loop, checking for a smaller distance on a "as you go" basis.

ktatterso: In the spirit of "maximum performance".  :)

kevinkreiser:
yeah when i wrote this i didnt consider normal use-cases to include results from 2700km away. the changes look fine to me but the unit tests are failing with incorrect results. i dont think they should be based on a change in ordering due to different unordered_map implementation... the square is just comparative so it should be ok too though i feel like i had it as the square before and more recently changed to regular distance.. have to check git blame to be sure though

src/midgard/tiles.cc
```diff
@@ -91,7 +102,7 @@ template <class coord_t> struct closest_first_generator_t {
       corners.emplace_back(x1, seed.second);
     }
     for (const auto& c : corners) {
-      auto d = seed.Distance(c);
+      auto d = seed.DistanceSquared(c);
```
mandeepsandhu, line 105: I'm not familiar with the algorithm used here, but will it work even though we;re returning square of the distance instead of the distance? Also maybe rename the variable for clarity (if sq distance can be returned)?

danpat: Yeah - I also updated the caller of this code to compare the squared distance, so it works out.

src/midgard/tiles.cc
danpat: I thought about trying that - I'll test locally and see if it makes any further improvement.

src/midgard/tiles.cc
kevinkreiser: the algorithm is just something home grown from me. iirc it just uses the distance as a comparison, the absolute values of those dont really matter just so that order-wise you have them working out the same is all that should be needed

mandeepsandhu: Thanks for the explanation @kevinkreiser @danpat . So as long as the values returned by the `binner` is not used directly, it should be fine.

src/midgard/tiles.cc
danpat: No noticeable gain by unrolling the loop.

ktatterso: 👍

danpat:
@kevinkreiser Can you sanity check the changes I made to `dist()` - if I'm understanding the approach correctly, all it needs to do is calculate the "nearest" subdivision.  For that, a distance to the midpoint of the neighbours should order them properly, avoiding testing each corner point.

Screen Shot 2021-09-14 at 5 34 24 PM

kevinkreiser:
@danpat i think about it this way:

image

each edge of the subdivision/cell has a triangle that descibes the part of the cell closest to it. the reason i am focused on the edges of the cell is because that is the closest place a directededge could be in an adjacent bin. anyway.. if we take the center points of the adjecent cells i think you can make the same claim, that is: the middle point of the cell to the right is closer to any point in the right most triangle than any other midpoint of any other adjacent cell. in fact if you draw out all combinations i reckon it would look like a voronoi diagram. that is intersecting halfplanes between all closest cell mid points.

BUT.... its easy to think about this stuff in the plane and trick yourself that it will work on the sphere. if we put my drawing above on a sphere the edges of the triangles would be arcs and they would be shaped differently in the northern vs the southern hemisphere.

its too late for me to rationally think about whether or not i believe the following propsition: if you properly compute the midpoint (taking into acount spherical coordinates) of a given subdivision/cell then this approach will work. indeed maybe im completely wrong about all the other crap i said. ill think about it with a fresh brain tomorrow :smile:

danpat:
@kevinkreiser you're right, it doesn't hold under spherical geometry, as demonstrated by this (1/2 degree by 1/2 degree boxes, red = old method, blue = new method):

Screen Shot 2021-09-15 at 11 56 15 AM

I'll revert that change from this PR, it's likely leading to incorrect edges being snapped.

ktatterso (approved):

kevinkreiser:
as i was saying before the code looks good to me but CI says it needs linted, and iirc even before the previous change and linting the unit tests were failing. if you have a second to lint then we'll see if the tests are happy now

danpat:
The test failures are due to the switch from `.Distance()` to `.DistanceSquared()`.    Looking at the `DistanceApproximator`, it's not surprising - the test cases cover very large parts of the sphere, I bet it breaks down over huge distances.

I'll revert that part of the change.  It makes the performance improvements slightly less impressive, but we still retain the bulk of the improvements (less memory realloc).

src/midgard/tiles.cc
```diff
@@ -51,14 +52,18 @@ void bresenham_line(double x0,
...
+  std::priority_queue<best_t, std::vector<best_t>, std::function<bool(const best_t&, const best_t&)>>
+      queue;
+
+  // re-usable, pre-allocated vector used by dist
+  std::vector<coord_t> corners;
 
   closest_first_generator_t(const valhalla::midgard::Tiles<coord_t>& tiles, const coord_t& seed)
       : tiles(tiles), seed(seed), queued(100), queue([](const best_t& a, const best_t& b) {
-          return a.first == b.first ? a.second < b.second : a.first < b.first;
+          return a.first == b.first ? b.second < a.second : b.first < a.first;
```
danpat, line 66: Sort order is inverted from before - the `std::priority_queue`'s `top()` function returns the "largest item", ergo we need to sort in reverse (smallest distance has highest priority).

ktatterso (approved):

mandeepsandhu (approved):
lgtm

PR #3326, opened 2021-09-16
Making HOV transition cost small

ktatterso:
# Issue

I noticed that hov lanes are frequently avoided when they should be routed upon.  I traced this back to the `base_transition_cost` routine, which saw the hov lane transition as having a non-trivial cost.  For `hot` lanes, the transition cost was even higher since the `toll` transition cost was also added.

For transitioning from the highway onto a hot lane (tolled road), if this is an hov transition I do not apply the toll penalty.

For the maneuver penalty, I have made the hov-transition 1/10 the typical.

An existing unit test was modified to capture this fix.

## Tasklist

 - [x] Add tests

## Requirements / Relations


valhalla/sif/dynamiccost.h
```diff
@@ -1035,6 +1036,12 @@ class DynamicCost {
 
     // shortest ignores any penalties in favor of path length
     c.cost *= !shortest_;
+
+    // switching into/out-of an hov-lane is just a lane change.
+    bool hov_transition = (pred->is_hov_only() && !edge->is_hov_only()) ||
+                          (!pred->is_hov_only() && edge->is_hov_only());
+    c.cost *= !hov_transition;
```
kevinkreiser, line 1043: i worry this is too broad. it basically says if we are getting onto or leaving an HOV edge then its always free no matter what other factors are at play. imagine you cross a country border where the road becomes or stops being HOV. then the penalty for that border crossing is removed. this would not be expected.

for the example you cited, i think the right change would be to check if the costing has HOT enabled, and if so dont do the toll penalty when its a transition onto an HOV only toll edge

gknisely: @ktatterso yeah zero is a bit much.  Did you try to reduce it versus making it zero?

gknisely: @ktatterso maybe we can check for HOT and if we are transitioning onto a ramp we can reduce the cost.

valhalla/sif/dynamiccost.h
```diff
@@ -1011,7 +1015,8 @@ class DynamicCost {
          node->private_access();
     c += bike_share_cost_ * (node->type() == baldr::NodeType::kBikeShare);
     c += toll_booth_cost_ *
-         (node->type() == baldr::NodeType::kTollBooth || (edge->toll() && !pred->toll()));
+         (node->type() == baldr::NodeType::kTollBooth || (edge->toll() && !pred->toll())) *
+         !is_hov_transition;
```
kevinkreiser, line 1019: exactly what i was suggesting :pray:

valhalla/sif/dynamiccost.h
```diff
@@ -1021,7 +1026,8 @@ class DynamicCost {
     c.cost += destination_only_penalty_ * (edge->destonly() && !pred->destonly());
     c.cost +=
         alley_penalty_ * (edge->use() == baldr::Use::kAlley && pred->use() != baldr::Use::kAlley);
-    c.cost += maneuver_penalty_ * (!edge->link() && !edge->name_consistency(idx));
+    c.cost += maneuver_penalty_ * (!edge->link() && !edge->name_consistency(idx)) *
+              (is_hov_transition ? 0.1 : 1.0);
```
kevinkreiser, line 1030: its hard for me to imagine the scenario where this occurs. do you have an example?

ktatterso: @kevinkreiser An HOV lane - in nearly every case I've studied - is just a lane change (a gentle turn; no need to slow down).  However, HOV lanes are also: 1) connected by a link, and 2) are named differently.

Here's a way for the link that connects "US 36" (not hov) to "US 36 Express" (hov):  https://www.openstreetmap.org/way/632700110

If I understand what I'm seeing in the debugger, Valhalla sees two non-trivial transitions (for what is otherwise really just a lane change):
1) from the highway onto the link
2) from the link onto the hov-lane

I've pushed some code changes that represent my latest thoughts (and appear to work).  However, I'm not sure if 1) I absolutely need to steal another bit from `EdgeLabel` to solve problem #2 above, and 2) I'm really not sure whether I've implemented `AutoCost::TransitionCostReverse` correctly (still getting my mind around the reverse stuff).

So yea, like I said, to me, transitioning onto/off-of an hov lane is just a simple lane change.  I'm just trying to figure out how to code it up.

PR #3329, opened 2021-09-19
fix setting the wrong edge candidates in shape walking map matcher

kevinkreiser:
There were two bugs in the shape walking map matching algorithm both to do with setting the correct edge candidate that was used at both the origin and destination locations. This causes triplegbuilder to remove all the locations candidates which has negative side effects downstream mainly for serialization. We caught this bug as part of the work in #3274 where we strictly require that you do not send through locations with edge candidates not properly set (ie matching the routes path_infos)

### DIFF
--- a/src/thor/route_matcher.cc
+++ b/src/thor/route_matcher.cc
@@ -60,10 +60,9 @@ float length_comparison(const float length, const bool exact_match) {
 //
 // Get a map of end edges and the start node
 // of each edge. This is used to terminate the edge walking method.
-end_node_t GetEndEdges(GraphReader& reader,
-                       const google::protobuf::RepeatedPtrField<valhalla::Location>& correlated) {
+end_node_t GetEndEdges(GraphReader& reader, const valhalla::Location& destination) {
   end_node_t end_nodes;
-  for (const auto& edge : correlated.rbegin()->path_edges()) {
+  for (const auto& edge : destination.path_edges()) {
     // If destination is at a node - skip any outbound edge
     GraphId graphid(edge.graph_id());
     if (edge.begin_node() || !graphid.Is_Valid()) {
@@ -328,7 +327,8 @@ bool RouteMatcher::FormPath(const sif::mode_costing_t& mode_costing,
   // Process and validate end edges (can be more than 1). Create a map of
   // the end edges' start nodes and the edge information.
   // TODO: when we want to do more than one leg we need to create correlated path_edges on the fly
-  auto end_nodes = GetEndEdges(reader, options.locations());
+  const auto& destination = *options.locations().rbegin();
+  auto end_nodes = GetEndEdges(reader, destination);
 
   // We support either the epoch timestamp that came with the trace point or
   // a local date time which we convert to epoch by finding the first timezone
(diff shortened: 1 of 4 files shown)


src/thor/route_matcher.cc
```diff
@@ -60,10 +60,9 @@ float length_comparison(const float length, const bool exact_match) {
 //
 // Get a map of end edges and the start node
 // of each edge. This is used to terminate the edge walking method.
-end_node_t GetEndEdges(GraphReader& reader,
-                       const google::protobuf::RepeatedPtrField<valhalla::Location>& correlated) {
+end_node_t GetEndEdges(GraphReader& reader, const valhalla::Location& destination) {
   end_node_t end_nodes;
-  for (const auto& edge : correlated.rbegin()->path_edges()) {
+  for (const auto& edge : destination.path_edges()) {
```
kevinkreiser, line 65: this was just a refactory for clarity, here we are only concerned about the destination so i refactored to have the function have you send just the destination

src/thor/route_matcher.cc
```diff
@@ -427,20 +427,40 @@ bool RouteMatcher::FormPath(const sif::mode_costing_t& mode_costing,
...
             return false;
           }
 
+          // When the route ends at a node in the graph we have an ambiguous case. Multiple
+          // destination edge candidates could have ended at this node but we use an unordered_map
+          // instead of a multimap, which means when we go to insert the other candidates that end
+          // there, they dont get inserted, only the first one does. This is all we really need for
+          // finding the path but it means that once we do find the path to that node, the edge that
+          // was in the value portion of the map entry might be the wrong edge. So here we need to
+          // go find the edge candidate that was actually used in the path
```
kevinkreiser, line 443: have a read of this. this fixes using the wrong edge candiate for the destination location. basically if the destination is at a node we need to figure out which of the edges ending there got used in the path.

src/thor/route_matcher.cc
```diff
@@ -427,20 +427,40 @@ bool RouteMatcher::FormPath(const sif::mode_costing_t& mode_costing,
...
+                             });
+            if (found_edge == destination.path_edges().end()) {
+              throw std::logic_error("Could not find destination candidate in shape-walked path");
+            }
+          }
+
           // TODO: when we actually have more than one leg, we have to do when we break legs
           // Store the matching edge candidates in the shapes locations
-          const auto& end_edge = n->second.first;
-          options.mutable_shape(0)->mutable_path_edges()->Add()->CopyFrom(end_edge);
```
kevinkreiser, line 439: this was a copy-paste-fail which caused the origin edge candidate to pretty much always be wrong except for in trivial routes.

src/thor/triplegbuilder.cc
```diff
@@ -416,7 +416,7 @@ void RemovePathEdges(valhalla::Location* location, const GraphId& edge_id) {
                             return e.graph_id() == edge_id;
                           });
   if (pos == location->path_edges().end()) {
-    location->mutable_path_edges()->Clear();
+    location->mutable_path_edges()->Clear(); // should never happen
```
kevinkreiser, line 419: i kept this here though as the comment says it should never happen. over in #3274 we will change this to a `throw`

test/mapmatch.cc
```diff
@@ -218,9 +218,13 @@ TEST(Mapmatch, test_matcher) {
...
 
     try {
+      Api api;
       walked_json = actor.trace_attributes(
           R"({"date_time":{"type":1,"value":"2019-10-31T18:30"},"costing":"auto","shape_match":"edge_walk","encoded_polyline":")" +
-          json_escape(encoded_shape) + "\"}");
+              json_escape(encoded_shape) + "\"}",
+          nullptr, &api);
+      EXPECT_NE(api.trip().routes(0).legs(0).location(0).path_edges_size(), 0);
+      EXPECT_NE(api.trip().routes(0).legs(0).location().rbegin()->path_edges_size(), 0);
```
kevinkreiser, line 227: these and the 2 below were failing in the other PR #3274 without the fixes in this pr

kdiluca (approved):

PR #3330, opened 2021-09-21
More real-world hov lane change/transition test case

ktatterso:
# Issue

I was led to believe that the transition cost for getting onto/off-of an hov lane was too high.  However, it was my test case that was not "real world" enough.

I've modified my preexisting test case to make it more "real world".  This means:
- The distance between the highway lane and hov/hot lane is more realistic
- The length and angle of entry/exit onto and off-of the hov/hot lane is more realistic
- The hov lane itself would typically be much longer, but I didn't want to go too crazy
- Entering the hov lane is (as convention) now a `highway:motorway-link`
- Note that exiting the hov lane and reentering the highway is (as convention) not a `link`

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/test/gurka/test_hov.cc
+++ b/test/gurka/test_hov.cc
@@ -467,8 +467,8 @@ protected:
       D--------------E--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------F----------------G
     )";
     const gurka::ways ways =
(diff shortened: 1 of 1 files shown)


CuriOusQuOkka (approved):
LGTM

PR #3332, opened 2021-09-22
Better hashing of GraphId

oleg-liatte:
# Issue

https://github.com/martinus/robin-hood-hashing

> Depends on good Hashing. For a really bad hash the performance will not only degrade like in std::unordered_map, the map will simply fail with an std::overflow_error. In practice, when using the standard robin_hood::hash, I have never seen this happening.

`static_cast<size_t>(k.value)` is not good hashing. Especially in case of 32-bit `size_t`, when `robin_hood::unordered_map` fails pretty fast.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
    * FIXED: Make combine_route_stats.py properly quote CSV output (best practice improvement) [#3328](https://github.com/valhalla/valhalla/pull/3328)
    * FIXED: Merge edge segment records in map matching properly so that resulting edge indices in trace_attributes are valid [#3280](https://github.com/valhalla/valhalla/pull/3280)
    * FIXED: Shape walking map matcher now sets correct edge candidates used in the match for origin and destination location [#3329](https://github.com/valhalla/valhalla/pull/3329)
+   * FIXED: Better hash function of GraphId [#3332](https://github.com/valhalla/valhalla/pull/3332)
 
 * **Enhancement**
    * CHANGED: Favor turn channels more [#3222](https://github.com/valhalla/valhalla/pull/3222)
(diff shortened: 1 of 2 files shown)


kevinkreiser (approved):

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
    * FIXED: Make combine_route_stats.py properly quote CSV output (best practice improvement) [#3328](https://github.com/valhalla/valhalla/pull/3328)
    * FIXED: Merge edge segment records in map matching properly so that resulting edge indices in trace_attributes are valid [#3280](https://github.com/valhalla/valhalla/pull/3280)
    * FIXED: Shape walking map matcher now sets correct edge candidates used in the match for origin and destination location [#3329](https://github.com/valhalla/valhalla/pull/3329)
+   * FIXED: Better hash function of GraphId [#3332](https://github.com/valhalla/valhalla/pull/3332)
```
kevinkreiser, line 22: I'm not sure this should be considered a fix. Probably more of an enhancement but no need to change this change log entry

valhalla/baldr/graphid.h
```diff
@@ -224,7 +224,14 @@ struct GraphId {
 namespace std {
 template <> struct hash<valhalla::baldr::GraphId> {
   inline std::size_t operator()(const valhalla::baldr::GraphId& k) const {
-    return static_cast<size_t>(k.value);
+    // simplified version of murmur3 hash for 64 bit
+    uint64_t v = k.value;
```
SiarheiFedartsou, line 228: nit: should we extract it into separate function in order to (maybe) re-use it for other types?

SiarheiFedartsou (approved):

CHANGELOG.md
oleg-liatte: Specifically in case of `robin_hood::unordered_map` this caused a bug (insertion to unordered_map threw an exception), that is why I considered this a bugfix.

valhalla/baldr/graphid.h
oleg-liatte: There is already similar function `robin_hood::hash_int`, however it is even more simplified and requires `#include <robin_hood.h>` in this header

kevinkreiser: Yeah we don't robinhood in our public headers

CHANGELOG.md
kevinkreiser: I couldn't remember where we used it with that container. Actually in most cases I expected we would send the value directly to the container and let the container worry about hashing it. In other words the container template type would be uint64_t

mandeepsandhu:
What problem was this trying to solve? Sorry I couldn't gather much from the description (was the current scheme causing too many collisions?)

Would love to have seen some tests demonstrating the fix.

SiarheiFedartsou:
@mandeepsandhu it works perfectly on 64-bit, but not on 32-bit, because `size_t` is `uint32_t` on 32-bit platforms.

oleg-liatte:
> What problem was this trying to solve? Sorry I couldn't gather much from the description (was the current scheme causing too many collisions?)
> 
> Would love to have seen some tests demonstrating the fix.

@mandeepsandhu `GraphId::value` layout is (from lower to higher):
* 3 bits: hierarchy level
* 22 bits: tile id
* 21 bit: unique id within tile

So if we just strip it to 32 bits (the size of `size_t` on 32-bit platforms) we get 26 bits which are constant per tile and 6 bits from unique id, which is not much unique and yes, causing too many collisions.

PR #3337, opened 2021-09-25
Update all python usage to python3

mandeepsandhu:
Updates all python usage to python3. For the scripts/ dir 2to3 was used
to convert python2 code to python3.


docker/install-shared-deps.sh
```diff
@@ -54,7 +54,6 @@ env DEBIAN_FRONTEND=noninteractive apt-get install --yes --quiet \
     pkg-config \
     prime-server${primeserver_version}-bin \
     protobuf-compiler \
-    python-all-dev \
     python3-all-dev \
     python3-minimal \
```
kevinkreiser, line 58: these seem contradictory haha

scripts/create_taginfo_items.py
```diff
@@ -10,4 +10,4 @@
   for value in values:
     items.append({'key':key,'value':value,'object_types':argv[5].split(','),'description':argv[6]})
 
-print(json.dumps(items))
+print((json.dumps(items)))
```
kevinkreiser, line 13: these kinds of changes make no sense to me

kevinkreiser:
some of the stuff seems unneeded to me but ill let @nilsnolde do the actual review since he is way more current on python best practices than i am

scripts/create_taginfo_items.py
mandeepsandhu: yeah i  think this happens because `2to3` thinks `print` (without `print_function` import) is printing a tuple and encapsulates it in another set of `()`. I think using `2to3 -p` will fix the issue. Let me try that.

scripts/list_tiles.py
```diff
@@ -158,7 +158,7 @@ def check_args(argv):
 
     while (len(bounding_boxes) != 0):
       b_box = bounding_boxes.pop(0)
-      for level, tiles in tile_hierarchy.levels.items():
+      for level, tiles in list(tile_hierarchy.levels.items()):
```
nilsnolde, line 161: all of the changes inserting a `list` can be reverted IMO

nilsnolde (changes requested):
just one thing about inserting lists where it's not needed. not that performance is of essence here, I'd still revert those redundant changes.

scripts/list_tiles.py
nilsnolde: I guess it does that because in 2 it was lists returned by those functions, in 3 it's iterators/generators

mandeepsandhu: Yeah looks like `2to3` matched the types and converted it into a `list`. I'll remove the list conversions.

kevinkreiser (approved):

nilsnolde (approved):

PR #3341, opened 2021-09-28
Exclude cash-only-tolls

ktatterso:
# Issue

Support to avoid routing through/on cash-only tolls.

A new request parameter is added:  `exclude_cash_only_tolls`.

Tolls appear on nodes as `barrier=toll_booth`.

There are 60 types of `payment` keys, e.g., `payment:coins` and `payment:credit_card`.  There are only three `payment` keys that are cash:
`payment:cash`
`payment:coins`
`payment:notes`

On a given tollednode there are frequently multiple `payment` keys specified, indicating the types of payment dis/allowed.

This PR examines the types of payments on a tolled node and determines if the toll is cash-only.  This detail is saved in the tile.  At runtime we can then inspect/honor this detail in the `Allowed` methods.

## Tasklist

 - [x] Add tests

## Requirements / Relations


kevinkreiser:
can you say more about this:

> Tolls can appear on nodes and ways.

looking at what you are trying to achieve i dont understand why this isnt sequestered to just nodes in the graph. surely toll collection only ever happens at a node. if osm supports tagging this on the way (i saw a tiny minority did), then i think we should either ignore it or try to mark the node at one end of the way as the collection point. in other words i dont understand why we need this on the directededge. to me it seems we can avoid cash only toll collection by disallowing costing to pass through a node with cash only collection and completely forget about marking the edges with this info.

also i looked at your entry for taginfo and it doesnt seem to match what osm tagging exists. you mentioned it in your description and i see the parser code is correctly pulling out the payment tags but the taginfo entry is talking about our internal boolean that we make in lua not the actual tags from the osm wiki. you need to change the taginfo entry to have the tag `payment:*` and a value of `*` i guess or something more like that.

ktatterso:
@kevinkreiser I did some more digging.  There appear to be _very_ few ways tagged as `toll=yes` and requiring cash-only payment.

Just out of interest I also dug into ways tagged as `barrier=toll_booth`.  (These make up 6% of the total entities with this tag, nodes making up the majority.)  While there are 1000's of ways with this tag, at least in the cases I examined it appears that those ways _are_ frequently connected to a node tagged `barrier=toll_booth`.

Point taken, I think it is safe to restrict this to nodes.

ktatterso:
@kevinkreiser Re taginfo, I'm still learning how to update it.

taginfo.json
```diff
@@ -2788,6 +2788,14 @@
       ],
       "description": "Allows access for hov."
     },
+    {
+      "key": "payment:*",
+      "object_types": [
+        "way",
```
kevinkreiser, line 2794: we are ignoring it on the way for now right?

ktatterso: oops, right

gknisely (approved):
looks good now.

valhalla/sif/dynamiccost.h
```diff
@@ -204,7 +204,9 @@ class DynamicCost {
    * @return  Returns true if access is allowed, false if not.
    */
   inline virtual bool Allowed(const baldr::NodeInfo* node) const {
-    return (node->access() & access_mask_) || ignore_access_;
+    return ((node->access() & access_mask_) &&
+            !(exclude_cash_only_tolls_ && node->cash_only_toll())) ||
+           ignore_access_;
```
kevinkreiser, line 209: this doesnt look right, i think it should be:

```c++
return ((node->access() & access_mask_) || ignore_access_) &&
            !(exclude_cash_only_tolls_ && node->cash_only_toll());
```

basically the ignoring access has to do with the access and the cash toll is a separate concern

cc @gknisely @ktatterso

gknisely: @kevinkreiser ty

Issue #2119, opened 2019-12-17
Libprotobuf 3.11 deprecates api's we use


purew:
Deprecation warnings (some of them) from building with ` protobuf-3.11.1-2`:

```
In file included from ../src/mjolnir/valhalla_convert_transit.cc:35:
../valhalla/mjolnir/transitpbf.h: In function ‘valhalla::mjolnir::Transit valhalla::mjolnir::read_pbf(const string&)’:
../valhalla/mjolnir/transitpbf.h:102:37: warning: ‘void google::protobuf::io::CodedInputStream::SetTotalBytesLimit(int, int)’ is deprecated: Please use the single parameter version of SetTotalBytesLimit(). The second parameter is ignored. [-Wdeprecated-declarations]
  102 |   cs.SetTotalBytesLimit(limit, limit);
      |                                     ^
In file included from src/valhalla/proto/tripcommon.pb.h:23,
                 from ../valhalla/baldr/datetime.h:21,
                 from ../src/mjolnir/valhalla_convert_transit.cc:21:
/usr/include/google/protobuf/io/coded_stream.h:397:8: note: declared here
  397 |   void SetTotalBytesLimit(int total_bytes_limit, int) {
      |        ^~~~~~~~~~~~~~~~~~
In file included from ../src/mjolnir/valhalla_convert_transit.cc:35:
../valhalla/mjolnir/transitpbf.h: In function ‘void valhalla::mjolnir::write_pbf(const valhalla::mjolnir::Transit&, const boost::filesystem::path&)’:
../valhalla/mjolnir/transitpbf.h:131:29: warning: ‘int google::protobuf::MessageLite::ByteSize() const’ is deprecated: Please use ByteSizeLong() instead [-Wdeprecated-declarations]
  131 |   auto size = tile.ByteSize();
```

mloskot:
I believe this can be closed by https://github.com/valhalla/valhalla/pull/3418

nilsnolde:
I'll try to catch all of them now

PR #3346, opened 2021-10-01
remove linking to boost

### LINKED ISSUE #3220: Reduce Boost Usage to Header-Only and Vendor Whatever Version We Like
a while ago we made a big push to reduce the dependencies, particularly boost, required to build valhalla in #1465. at this moment in time we use lots of header only bits of boost but only a tiny handful (maybe only `program_options`) of boost that need linking and actually prebuild library. in order to reduce the possibility that valhalla is not buildable on a target platform we can bring our own version of boost and just the bits that we need. to do that we need to find and stomp out any boost linking that we are doing. there is one catch though, we do expose boost in our public apis which means if we vendor it we need it to not conflict with any system boost that is installed. that means it has to live under the same include path where we install our own headers. cmake will handle all of that for us if we do the work right but still its worth noting that vendoring it means even to the install destination.

@nilsnolde what do you think about this?

nilsnolde:
fixes #3220 

replaces boost.program_options with https://github.com/jarro2783/cxxopts at v2.2.1, v3.x has no stable release yet. that's the last boost we need to link, so we can vendor all the header libs we need and finally remove boost from dependencies!

so far I only migrated `valhalla_build_tiles` which has multiple positional args so I wanted to make sure I can straighten that out first with the new lib. I'll wait for initial :+1: before I migrate all other scripts. IMO readability improved, it's a little less convoluted than the boost lib. I tested all options and it works fine.

boost is still there, I just pulled in the new submodule. would someone be willing to help with the cmake work? I can take care of the other executables.

tasks:
- [x] migrate exectuables to use cxxopts and remove all references to boost::program_options in the sources (with the exception of a file which isn't even built anymore:  https://github.com/valhalla/valhalla/blob/master/src/valhalla_tiles_for_elevation.cc)
- [x] vendor header boost libs
- [x] adapt cmake
- [x] adapt CI
- [x] adapt docker
- [x] adapt documentation


nilsnolde:
btw, here's the help text, not much changed:
```
valhalla_build_tiles v3.1.3

valhalla_build_tiles is a program that creates the route graph
from one or multiple osm.pbf extract(s)

Usage:
  valhalla_build_tiles [OPTION...] OSM PBF file(s)

  -h, --help               Print this help message.
  -v, --version            Print the version of this software.
  -c, --config arg         Path to the configuration file
  -i, --inline-config arg  Inline JSON config
  -s, --start arg          Starting stage of the build pipeline (default:
                           initialize)
  -e, --end arg            End stage of the build pipeline (default: cleanup)

Build stage strings (in order)
    initialize
    parseways
    parserelations
    parsenodes
    constructedges
    build
    enhance
    filter
    transit
    bss
    hierarchy
    shortcuts
    restrictions
    elevation
    validate
    cleanup
 ```

File: src/mjolnir/valhalla_build_tiles.cc [hunk 1]
```diff
@@ -28,115 +26,113 @@ void list_stages() {
...
+                                                      "Print the version of this software.")(
+        "c,config", "Path to the configuration file",
+        cxxopts::value<std::string>())("i,inline-config", "Inline JSON config",
+                                       cxxopts::value<std::string>())(
+        "s,start", "Starting stage of the build pipeline",
+        cxxopts::value<std::string>()->default_value(
+            "initialize"))("e,end", "End stage of the build pipeline",
+                           cxxopts::value<std::string>()->default_value(
+                               "cleanup"))("f,files", "positional arguments",
+                                           cxxopts::value<std::vector<std::string>>());
```
nilsnolde, line 51: this is disgusting formatting.. the main difference here is that we don't pass vars by ref, but we retrieve an arg from the `result` obj further down

src/mjolnir/valhalla_build_tiles.cc
```diff
@@ -28,115 +26,113 @@ void list_stages() {
...
+
+    options.parse_positional({"files"});
```
nilsnolde, line 53: since `files` is a `std::vector<std::string>>` this fetches multiple positional args

src/mjolnir/valhalla_build_tiles.cc
```diff
@@ -28,115 +26,115 @@ void list_stages() {
...
+
+    // Read the config file
+    boost::property_tree::ptree pt;
+    if (result.count("inline-config")) {
+      std::stringstream ss;
+      ss << result["inline-config"].as<std::string>();
+      rapidjson::read_json(ss, pt);
+    } else if (result.count("config") &&
+               filesystem::is_regular_file(config_file_path = result["config"].as<std::string>())) {
+      std::cout << config_file_path << std::endl;
```
nilsnolde, line 113: opps, I'll remove that

kevinkreiser:
i will gladly help with cmake!

src/mjolnir/valhalla_build_tiles.cc
kevinkreiser: you can wrap sections of code with `nolint` directives to tell the linter to leave it alone. i have done this in various places when i cant stand the stupidity that it does

nilsnolde:
ok cool thanks @kevinkreiser ! I guess when CI is green that's pretty much all confirmation we need that the changes are fine and I can go ahead with the rest right? since valhalla_build_tiles is extensively used in the tests?

src/mjolnir/valhalla_build_tiles.cc
kevinkreiser: https://github.com/valhalla/valhalla/blob/deeba182f24e6a4aa4f5eb30e76d81cbb8cd80cc/src/thor/trace_route_action.cc#L97-L105

kevinkreiser:
@nilsnolde yeah we have a number of tilesets that are built on the command line in cmake. they all share the same format but its a pretty decent test imho as it even uses the inline config option

src/mjolnir/valhalla_build_tiles.cc
nilsnolde: what a relief..

src/mjolnir/valhalla_build_tiles.cc
```diff
@@ -28,115 +26,112 @@ void list_stages() {
...
+
+    // clang-format off
+    options.add_options()
+      ("h,help", "Print this help message.")
+      ("v,version","Print the version of this software.")
+      ("c,config", "Path to the configuration file", cxxopts::value<std::string>())
+      ("i,inline-config", "Inline JSON config", cxxopts::value<std::string>())
+      ("s,start", "Starting stage of the build pipeline", cxxopts::value<std::string>()->default_value("initialize"))
+      ("e,end", "End stage of the build pipeline", cxxopts::value<std::string>()->default_value("cleanup"))
+      ("f,files", "positional arguments", cxxopts::value<std::vector<std::string>>());
```
nilsnolde, line 49: this doesn't get printed in the help btw

nilsnolde:
Btw, since there’s no pressure here, I’d convert a few executables whenever I have half an idle hour over the next weeks. Just happy it’s kicked off:)

src/mjolnir/valhalla_benchmark_admins.cc
```diff
@@ -181,8 +181,7 @@ void Benchmark(const boost::property_tree::ptree& pt) {
   }
 
   // Graphreader
-  auto hierarchy_properties = pt.get_child("mjolnir");
-  GraphReader reader(hierarchy_properties);
+  GraphReader reader(pt);
```
nilsnolde, line 184: was a bug

src/mjolnir/valhalla_build_connectivity.cc
```diff
@@ -14,15 +14,13 @@ using namespace valhalla::baldr;
...
-#include <boost/program_options.hpp>
 #include <boost/property_tree/ptree.hpp>
+#include <cxxopts.hpp>
 #include <ostream>
 
-namespace bpo = boost::program_options;
 using namespace valhalla::midgard;
 
 filesystem::path config_file_path;
-std::vector<std::string> input_files;
```
nilsnolde, line 25: there's not input files here. you'll see that in a few exes, guess a copy/paste fail

src/mjolnir/valhalla_query_transit.cc
```diff
@@ -39,7 +38,7 @@ Transit read_pbf(const std::string& file_name) {
   google::protobuf::io::CodedInputStream cs(
       static_cast<google::protobuf::io::ZeroCopyInputStream*>(&as));
   auto limit = std::max(static_cast<size_t>(1), buffer.size() * 2);
-  cs.SetTotalBytesLimit(limit, limit);
+  cs.SetTotalBytesLimit(limit);
```
nilsnolde, line 41: deprecated warning from protobuf

src/mjolnir/valhalla_ways_to_edges.cc
```diff
@@ -38,55 +34,40 @@ struct EdgeAndDirection {
...
+      exit(0);
+    }
 
-  if (vm.count("version")) {
-    std::cout << "ways_to_edges " << VALHALLA_VERSION << "\n";
-    return true;
-  }
+    if (result.count("version")) {
+      std::cout << "valhalla_ways_to_edges " << VALHALLA_VERSION << "\n";
+      exit(0);
```
nilsnolde, line 59: need to exit here

nilsnolde:
somehow wanted to have it over with.. all exes are migrated now @kevinkreiser , boost::program_options is effectively not used anymore:) though there's still one file: https://github.com/valhalla/valhalla/blob/master/src/valhalla_tiles_for_elevation.cc. it's not being built anymore though, maybe we can remove it altogether? I left that one alone for now..

I tested the exes that were easy in input, but by far not all of them extensively, though I did make sure help/version works. missing the input for some and didn't have the patience to generate the input for others;) so many executables.. 

can people please check if the exes they care about are still working as expected? I'll test the very crucial ones again in the next days.. no stress here of course, would be nice to close the lid on it though:)

nilsnolde:
yeahiii, I think I got it to work with `bcp`, pretty painless approach. in case you're not familiar with it (as me until an hour ago): it's a tool which copies boost libraries with their boost dependencies in a specified directory. now that we only use header-only libs that's enough for us. but that also means they are files, not submodules, so it's static and weighs ~80 MB on top of the repo. then if we want to update the boost version, someone will have to:
- download the boost sources for that version
- build `bcp`
- use `bcp` to update the `third_party/boost` tree:

```
 bcp algorithm iterator property_tree optional format range make_shared noncopyable geometry shared_ptr archive tokenizer foreach shared_array intrusive_ptr smart_ptr variant lexical_cast heap functional  --boost=../../test/boost_1_77_0 third_party
 ```

(bcp doesn't recurse directories on its `--scan` mode and I was lazy..)

IMO that's fine as long as that procedure is documented somewhere in the readme. 

what's your take @kevinkreiser @mloskot ?

**Comment by nilsnolde** [comment]
holy shit.. hahaha I did **not** think about the consequences to contributions.. viewing the diff almost made my pc choke! jeez, no idea what to do.. :sweat_smile: that even trumps the ancient format commit from patrick argh..

honestly, git submodules are not really an option, adds gigs of git clones. I guess it doesn't matter so much who'd do it, but fwiw  I happily let someone else do that if this is the direction we'd like to go.. in that case I'd simply contribute boost in a separate PR independently so this diff is easier to review.

**Comment by kevinkreiser** [comment]
@nilsnolde i havent checked all of the files :smile: but isnt it the case that we only use a few specific things from boost: ptree, algorithm, geometry, format... is it more than that? let me check...

```
algorithm
archive
filesystem
foreach
format
functional
geometry
heap
intrusive_ptr
iterator
lexical_cast
make_shared
noncopyable
optional
program_options
property_tree
range
shared_ptr
smart_ptr
tokenizer
variant
```

a quick `grep -rIF 'include <boost' valhalla/* src/* | sed 's/.*://g' | sed -e "s@.*boost/\([a-Z0-9_]\+\).*@\1@g" | sort | uniq` shows its not too much and there are definitely a few we can just straight up get rid of but is there a way to make that tool target only the headers you want? instead of 5k+? i cant imagine we really need that many..

on another note, i definitely dont think that we should be dumping boost sources into the valhalla repository. we should write a cmake target that pulls the dependencies down at build time and caches them. we could do that with bcp and roll our own or... we could just use `mason`. it already has header-only boost at version 1.75 which is more than enough for us and easily addable to our cmake with a one liner: https://github.com/mapbox/mason#masoncmake

**Comment by nilsnolde** [comment]
Right, mason I forgot about, I’ll give that a try.

don’t forget the internal boost deps, it’s a huge spider.. then there’s around 200 headers it’ll copy just to make it platform independent, and so on. Before you know you’re at a crazy amount.. (likely some docs and shit in there too, didn’t spend much time cleaning it up)

**Comment by mloskot** [comment]
> we could just use mason

For Windows, please use vcpkg or conan :)

**Comment by kevinkreiser** [comment]
I was thinking about Conan too but I convinced myself it didn't matter in this case (meaning mason should work for all platforms) because it's header only. Is there something different about vendoring header only boost on windows vs other platforms?

**Comment by mloskot** [comment]
OK

> Is there something different about vendoring header only boost on windows vs other platforms?

I'm not aware of any differences.

**Comment by kevinkreiser** [comment]
I should say I also have no real preference in which tool we use here. If conan is similarly easy to use then i dont see why we we dont try that, here seems to be their example: https://github.com/conan-io/cmake-conan

**Comment by mloskot** [comment]
I just had impression Mason is a plain Unix (Linix+OSX) tool.

For Conan for Boost.GIL I did this trick
- https://github.com/boostorg/gil/blob/develop/conanfile.txt
- https://github.com/boostorg/gil/blob/develop/CMakeLists.txt#L158-L177

and it works quite transparently for users.

**Comment by nilsnolde** [comment]
@kevinkreiser that's it now more or less. it's using conan now, thanks @mloskot , was really helpful to see your example. 

not much to add other than yayy, finally done!

**Comment by nilsnolde** [comment]
oops the full boost third_party tree is still registered :sweat_smile:

File: .circleci/config.yml [hunk 9]
```diff
@@ -10,9 +10,12 @@ executors:
...
   install_macos_dependencies:
     steps:
-      - run: brew install protobuf cmake ccache libtool boost libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
-      - run: pip3 install requests shapely
+      - run: brew install protobuf cmake ccache libtool libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
+      - run: pip3 install requests shapely conan
       - run: git clone https://github.com/kevinkreiser/prime_server --recurse-submodules && cd prime_server && ./autogen.sh && ./configure && make -j4 && make install
+  install_conan:
+    steps:
+      - run: apt-get update -y && apt-get install -y python3-pip && pip install conan
```
nilsnolde, line 18: just temp to see if ci works, will PR to update the build image before

CMakeLists.txt
```diff
@@ -133,6 +133,44 @@ endif()
 find_package(Threads REQUIRED)
 find_package(ZLIB REQUIRED)
 
+# try to find an installed boost
+set(boost_VERSION "1.71")
+find_package(Boost ${boost_VERSION} QUIET)
+if (NOT Boost_Found)
```
nilsnolde, line 139: few questions here:

- boost version to pin? this is ubuntu 20.04 currently. I have no real feelings about what we're pinning to, can be 1.77.0 if you want, I tried, works as well
- now I'm pretty sure the logic here is not robust. in src/CMakeLists.txt it seems we have to link `boost::headers` to the module targets (I couldn't compile without) and I guess that wouldnt be a "library" if it's not a conan installation.. maybe there's another way generally? more in the other cmakelists

CMakeLists.txt
```diff
@@ -133,6 +133,44 @@ endif()
...
+  if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
+    message(STATUS "Valhalla: Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+    file(DOWNLOAD "https://raw.githubusercontent.com/conan-io/cmake-conan/0.17.0/conan.cmake"
+                  "${CMAKE_BINARY_DIR}/conan.cmake"
+                  EXPECTED_HASH SHA256=3bef79da16c2e031dc429e1dac87a08b9226418b300ce004cc125a82687baeef
+                  TLS_VERIFY ON)
+  endif()
+
+  list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR})
+  list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
```
nilsnolde, line 156: basically I took the lower example from here: https://docs.conan.io/en/latest/integrations/build_system/cmake/cmake_find_package_generator.html#in-a-conanfile-txt

CMakeLists.txt
```diff
@@ -133,6 +133,44 @@ endif()
...
+    GENERATORS cmake_find_package
+    OPTIONS boost:header_only=True)
+  conan_cmake_autodetect(settings)
+  conan_cmake_install(PATH_OR_REFERENCE .
+                      INSTALL_DIR ${VALHALLA_SOURCE_DIR}/test_conan
+                      REMOTE conancenter
+                      SETTINGS ${settings}
+                      OUTPUT_QUIET)
+
+  find_package(Boost ${boost_VERSION})
```
nilsnolde, line 170: the `boost:header_only=True` is still pulling the headers for all libraries, incl the ones which need linking. in total it's ~160 MB (down from multiple GB in git submodules).

for future ref @mloskot : this is apparently what conan [recommends](https://github.com/conan-io/cmake-conan#conan_cmake_run-high-level-wrapper) these days, also doesn't need a `conanfile.txt`, but a few more lines.

CMakeLists.txt
```diff
@@ -133,6 +133,44 @@ endif()
...
+  list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
+
+  include(${CMAKE_BINARY_DIR}/conan.cmake)
+
+  conan_cmake_configure(REQUIRES boost/${boost_VERSION}.0
+    GENERATORS cmake_find_package
+    OPTIONS boost:header_only=True)
+  conan_cmake_autodetect(settings)
+  conan_cmake_install(PATH_OR_REFERENCE .
+                      INSTALL_DIR ${VALHALLA_SOURCE_DIR}/test_conan
```
nilsnolde, line 165: that's mistake..

CMakeLists.txt
```diff
@@ -216,25 +254,20 @@ if(ENABLE_SERVICES)
...
     INTERFACE_LINK_LIBRARIES "${LUA_LIBRARIES}"
     INTERFACE_INCLUDE_DIRECTORIES "${LUA_INCLUDE_DIR}")
-## Top level executables
-elseif(ENABLE_TOOLS)
-  find_package(Boost 1.51 REQUIRED COMPONENTS program_options)
-## Header only boost for the library without mjolnir
-elseif(NOT Boost_FOUND)
-  find_package(Boost 1.51 REQUIRED)
 endif()
-add_definitions(-DBOOST_NO_CXX11_SCOPED_ENUMS)
```
nilsnolde, line 237: just moved this up

docker/Dockerfile-run
```diff
@@ -1,6 +1,6 @@
 ####################################################################
 # base image contains all the dependencies we need to build the code
-FROM valhalla/valhalla:build-3.1.0 as builder
+FROM valhalla/valhalla:build-latest as builder
```
nilsnolde, line 3: this I don't understand.. how can we have fresh valhalla:run containers if this was always fixed to build-3.1.0 before?

src/CMakeLists.txt
```diff
@@ -98,6 +98,8 @@ function(valhalla_module)
      PRIVATE
        ${VALHALLA_BINARY_DIR}
        ${VALHALLA_BINARY_DIR}/valhalla)
+  
+  target_link_libraries(${library} Boost::headers)
```
nilsnolde, line 102: it seems to also work with `boost::boost` as link. is this the right way to do it? conan cmake seems to do it like this themselves for tests: https://github.com/conan-io/conan-center-index/blob/dc40d949a49ac8e04f3a4020a0b208c1795680f5/recipes/boost/all/test_package/CMakeLists.txt#L127

valhalla/mjolnir/transitpbf.h
```diff
@@ -135,7 +135,7 @@ void write_pbf(const Transit& tile, const filesystem::path& transit_tile) {
   if (!filesystem::exists(transit_tile.parent_path())) {
     filesystem::create_directories(transit_tile.parent_path());
   }
-  auto size = tile.ByteSize();
+  auto size = tile.ByteSizeLong();
```
nilsnolde, line 138: just getting rid of a deprecated warning

nilsnolde:
if windows doesn't work right away, I'll try another day on win. it seems easier than vcpkg. not sure if all deps are available with compatible versions on windows, but maybe worth a thought to use only conan @mloskot ? I feel like vcpkg has better maintenance, but I really like that conan lets you choose versions independently from each other. but not much experience there..

CMakeLists.txt
nilsnolde: updated `boost::headers` to `boost::boost` in scr/CMakeLists.txt, seems to work as well. I'll try tmrw if it works with a system boost installation > 1.71. maybe there's a system out there which has boost >= 1.71 but not python3?  doesn't sound likely;) anyways, 160 MB of useless space saved is also an argument

CMakeLists.txt
mloskot: @nilsnolde The link points to the `conan_cmake_run() high level wrapper` section which says _"This function is **not** the recommended way of using cmake-conan any"_. Do you mean that section ?

mloskot:
@nilsnolde My not to extensive experience so far tells me to put `=` sign in between the vcpkg and conan. Both have worked for me well. 👍 to try conan and if it works, then it works.

CMakeLists.txt
nilsnolde: yeah right they recommend using the single functions like `conan_cmake_configure()` etc further up in that document. was more of a FYI, maybe avoids some confusion if your boost.gil conan is not functioning properly at some point.

mloskot: Yes, thanks for the update. I'll have to catch the GIL scripts up.

nilsnolde:
I tried reproducing the azure failure but couldn't. obviously, because no file anywhere references boost.regex.. it's not a caching issue either, the build cache never worked that I'm aware of (and not on this PR for sure) and the dependency cache was missed as it should've. don't get it.. 

@kevinkreiser can you trigger azure CI again pls? don't have access.. (I remember you inviting me azure, but that never quite worked I think)

nilsnolde:
I can also only encourage people (unfortunately there's no "all" @ on github..) to go and check the executables they care about. remember there's **hardly any test coverage** on executables. the ones that are essential I tested pretty thoroughly and so did CI. but not e.g. anything transit related or convenience/testing exes I never use. of course that can be rectified in a separate PR later on, but now you'd get me to fix it:) can't promise for after this is merged..

mloskot:
IMHO, if things work for the core team and yourself, I'd not worry. If anything braks for the rest of the world, then let's wait and hear anything from them, and receive fixes :)

I will check the stuff on my end as soon as I upgrade and will offer fixes if any necessary.

CMakeLists.txt
```diff
@@ -133,6 +133,43 @@ endif()
...
+  endif()
+
+  if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
+    message(STATUS "Valhalla: Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+    file(DOWNLOAD "https://raw.githubusercontent.com/conan-io/cmake-conan/0.17.0/conan.cmake"
+                  "${CMAKE_BINARY_DIR}/conan.cmake"
+                  EXPECTED_HASH SHA256=3bef79da16c2e031dc429e1dac87a08b9226418b300ce004cc125a82687baeef
+                  TLS_VERIFY ON)
+  endif()
+
```
kevinkreiser, line 154: can we just pull this down and keep it then you dont need to do the module path stuff again below since we do it right at the beginning

CMakeLists.txt
```diff
@@ -133,6 +133,43 @@ endif()
...
+  include(${CMAKE_BINARY_DIR}/conan.cmake)
+
+  conan_cmake_configure(REQUIRES boost/${boost_VERSION}.0
+    GENERATORS cmake_find_package
+    OPTIONS boost:header_only=True)
+  conan_cmake_autodetect(settings)
+  conan_cmake_install(PATH_OR_REFERENCE .
+                      REMOTE conancenter
+                      SETTINGS ${settings}
+                      OUTPUT_QUIET)
```
kevinkreiser, line 167: this is what populates the stuff in `.conan` we should add that as one of the places to cache in CI

kevinkreiser (dismissed):
a cuople of little nit picks, but after we have those we can merge this!

kevinkreiser:
same failure on windows trying to link boost regex :thinking:

nilsnolde:
couple of fixes after our review session yesterday:
- pull conan.cmake into this repo to avoid downloading every time
- cache conan installed libs in CI
- minor other stuff which hopefully lets azure pass

src/mjolnir/CMakeLists.txt
```diff
@@ -88,6 +88,7 @@ valhalla_module(NAME mjolnir
     valhalla::baldr
     SpatiaLite::SpatiaLite
     SQLite3::SQLite3
+    Boost::boost
```
nilsnolde, line 91: really doubt this is the missing piece to properly "link" the mjolnir exes but who knows..

CMakeLists.txt
mloskot: So, Boost via conan remains optional. Good.

mloskot (dismissed):
LGTM, thanks!

src/mjolnir/CMakeLists.txt
mloskot: I recall I had to apply similar patch in the past, but then it wasn't necessary. I could not grasp what was going on, so I did not PR it.

CMakeLists.txt
```diff
@@ -133,6 +133,31 @@ endif()
...
+  # bail if there's no conan installed
+  message(STATUS "No compatible boost version detected, using conan...")
+  find_program(conan_FOUND conan)
+  if (NOT conan_FOUND)
+    message(FATAL_ERROR "conan needs to be installed for boost, see https://docs.conan.io/en/latest/installation.html.")
+  endif()
+
+  # build dir needs to be added, that's where conan writes some info
+  list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR})
+  # list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
```
nilsnolde, line 149: I'm really not sure why this line should be needed. there's nothing being installed to the build directory that needs to be found with `find_package/program/xx` unless some magic in conan cmake is requiring it.. that's how the documentation uses conan's `cmake_find_package` generator. linux configs and builds fine without, lets see what CI says for osx & win

mloskot: I don't know either. Let's dance like the Conan folks do.

nilsnolde: yeah this line on azure cmake config might be related:
```
ERROR: Not able to automatically detect 'C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe' version
ERROR: Unable to find a working compiler
``` 
it's not actually erroring out though, it still installs fine (not sure why it'd need a compiler to install in the first place..). anyways, I'll go with the documented way and uncomment that line again

src/mjolnir/CMakeLists.txt
nilsnolde: still not working on azure.. https://dev.azure.com/valhalla1/valhalla/_build/results?buildId=10260&view=logs&j=8c5a3053-cc94-5e24-da52-c43aabf65e0e&t=b45e0669-c54f-5dea-18c2-f681995b4f24

it seems like it's only affecting `mjolnir` executables though (i.e. `DATA_TOOLS`), `valhalla_validate_transit` just being the first trying to link libs. I have **no** idea what the problem could be: it's not CI caching and the only `mjolnir` specific instructions is about including luajit.. 

some win c++ lib requiring boost::regex in that exact msvc version? it **works locally** (or did a few commits ago when azure was failing exactly the same) and the only difference to CI setup could be a msvc version. I'll switch to win again sigh..

nilsnolde: tried again locally after upgrading to latest msvc (confirmed it's the same as on azure), `valhalla_validate_transit` still builds without issues.. out of ideas now.. if you any clue off the top of your head @mloskot on what to try/log, I'm all ears:)

nilsnolde: hm, one last thing to try is excluding all boost libs we don't need in the conan recipe (unfortunately they opted for `--without-xxx` logic rather than opt-in). it's possible that it's somehow messing smth up. and that's also the ultimate difference to my local setup: I never uninstalled boost from my local vcpkg. argh, that makes sense actually.. also there's a suspicious reference in the conan boost source where I'm not entirely sure what it's doing (even if it seems fairly unrelated): https://github.com/conan-io/conan-center-index/blob/306d3ab1bbf71aea6bf0f3f032fe03a420e376a0/recipes/boost/all/conanfile.py#L1062

failing that, I'm sorta tempted to just freaking install boost.regex with conan just to have CI pass and wait for people to come and complain.. terrible band-aid without knowing what's wrong..

nilsnolde: got the bastard, will comment more where it's more visible for future reference

nilsnolde:
finally found out what the problem is with windows boost (well not really actually):

for some _terribly opaque_ reason conan wants to link to boost.regex when building on windows/visual studio. that's clearly a conan bug in the boost package configuration. I tried a few things with the existing conan configs (btw got rid of unwanted headers like program_options in the process), but nothing worked. so my fix for the time being:

install boost.regex via vcpkg so msvc can link what needs linking (effectively removing this whole PR's intent for windows), without polluting osx/linux environments. the alternative could be to have a win-specific conanfile.txt (possible via env vars). which is slightly less transparent IMO than using vcpkg.

2 last questions: 
- boost v1.71.0 is ok or should we go higher? my reasoning was it's ubuntu 20.04's current version where lots of folks maybe won't have to use the conan installation at all. 
- is the way I handle windows now ok?

nilsnolde:
and anyone got any idea how to make this easily reproducible for the conan packagers? I simply fail to understand the relation between mjolnir exes and this conan bug.. what could trigger this? that some library that mjolnir includes/links to triggers conan to try to link to boost.regex (possibly another boost include that's exclusive to mjolnir?)? why not the other executables which require boost like the `TOOLS` stuff which seems to link fine (actually didn't really check that)?

kevinkreiser:
@nilsnolde i dont understand how it would be only the mjolnir executables, because they all depend on boost in the same way other ones do, the only other possibility is what youve already suggested, that on windows one of the other targets that mjolnir exe's link have boost regex as a dependency. that seems pretty wild to me but its not impossible. my problem is im unsure how to tell. under linux we could use  `VERBOSE=1` or so to tell what the link line looked like. although i guess cmake puts a bunch of configuration artifacts, maybe we can sift through those for regex to see where its coming from, ah ha, what about this:

```
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/cmake-build-debug$ grep -rF regex * | grep boost
src/mjolnir/CMakeFiles/valhalla-mjolnir.dir/CXX.includecache:boost/algorithm/string/regex.hpp
```

What is that doing in there?

```
kkreiser@X570-I-AORUS-PRO-WIFI:~/sandbox/valhalla/cmake-build-debug$ grep -rF regex ../src/mjolnir/* | grep -F boost
../src/mjolnir/validatetransit.cc:#include <boost/algorithm/string/regex.hpp>
```

maybe that is whats doing it? we can totally get rid of that im sure.

nilsnolde:
good 🕵️ work @kevinkreiser :) this is really only anecdotal, I just don't understand:

indeed that was what triggered the (wrong) linking. it's not even used so the offending `#include` is gone. what I cannot understand: why the hell would it _not_ link to boost.regex on linux & osx? indeed the header-only boost.algorithm needs to be linked to the (compiled?) boost.regex on the `boost/algorithm/string/regex.hpp` header: https://github.com/boostorg/algorithm/blob/2a05506d52c6b6d838addd64acfb7f1d9a9fd057/include/boost/algorithm/string/regex.hpp#L15.

regex is also required in the boost.algorithm CMakeLists.txt.. is it possible that boost.algorithm doesn't actually need to link against boost.regex but only have access to some parts of the header? and the boost builder (or whatever) has a bug on windows (or conan?) and thinks it needs to link anyways? can't think of any other explanation for that platform-dependent behavior..

mloskot:
I wonder if the Boost autolinking is playing there. It can be disabled w/ preprocessor macro `BOOST_ALL_NO_LIB`. The macro should be defined when `Boost::disable_autolinking` target is added as a dependency (along with `Boost::boost`:

https://github.com/Kitware/CMake/blob/e5d9fce03feeece7f53451da9fb958db2613d31e/Modules/FindBoost.cmake#L187-L189

nilsnolde:
ahhh thanks @mloskot , now that seems to be a very plausible reason.. see here where they generate the boost conan config: https://github.com/conan-io/conan-center-index/blob/1020df5ba3f3c5ffd8ec5c24e119aa50d5211234/recipes/boost/all/conanfile.py#L1398-L1408

so if I'd put `boost:magic_autolink=False` into the conanfile.txt it might work. then clang should fail as well? I'll try that I think when I'm back on linux

nilsnolde:
so this is now ready for merging @kevinkreiser ..

CMakeLists.txt
```diff
@@ -133,6 +133,31 @@ endif()
 find_package(Threads REQUIRED)
 find_package(ZLIB REQUIRED)
 
+# try to find an installed boost or install locally with conan
+set(boost_VERSION "1.71")
+# find_package(Boost ${boost_VERSION} QUIET)
+# if (NOT Boost_FOUND)
```
kevinkreiser, line 139: this looks like you were testing and accidentally didnt put it back. ill give it a try

docker/Dockerfile-run
```diff
@@ -1,6 +1,6 @@
 ####################################################################
 # base image contains all the dependencies we need to build the code
-FROM valhalla/valhalla:build-3.1.0 as builder
+FROM valhalla/valhalla:3.1.0 as builder
```
kevinkreiser, line 3: this doesnt look like it belongs

kevinkreiser (approved):

kevinkreiser:
seems like there were two little things to clean up so i did that quick hoping that it will be all good and we can merge :crossed_fingers:

kevinkreiser:
im not 100% certain but it seems like now the build prints out everything from all the test targets that build tiles using `valhalla_build_tiles` this makes me think that somehow the inline config that is set to disable logging is not being properly handled. we'll want to look into this!

nilsnolde:
jep, seems like it. I'll have a look

valhalla/mjolnir/transitpbf.h
mloskot: Yes, we may have to bump minimum required version of Protobuf 
Related are https://github.com/valhalla/valhalla/issues/2119 and https://github.com/valhalla/valhalla/pull/3418

nilsnolde: haha thanks, was about to plea for not bumping. jep, that's what I'll do, ifdef another one..

kevinkreiser: yeah we dont have many of these, `ifdef` is a good solution

PR #3349, opened 2021-10-07
Avoid calling out "keep left/right" when passing an exit

ktatterso:
# Issue

If passing by an exit that is preceded by a "standard length" deceleration lane, avoid calling out a maneuver, e.g., "Keep left/right" to stay on the main road.

# Examples

### Before

Screen Shot 2021-11-02 at 4 20 14 PM

### After

Screen Shot 2021-11-02 at 4 51 51 PM

### Before

Screen Shot 2021-11-02 at 4 33 28 PM

### After

Screen Shot 2021-11-02 at 4 47 28 PM

# Solution

If we detect that the lane bifurcation/split is due to a deceleration lane, we avoid calling out the "keep left/right" maneuver.

The length of a deceleration lane is predictable as a function of road speed.  I used the results from [this research paper](https://www.researchgate.net/publication/282505831_Selecting_Optimal_Deceleration_Lane_Lengths_at_Freeway_Diverge_Areas_Combining_Safety_and_Operational_Effects).

The logic works backwards from the bifurcation, adding up edge lengths until 1) the extra lane goes away, or 2) we exceed the standard (+tol) length of a deceleration lane.

If we determine that the length of the additional lane fits the standard length of a deceleration lane (within tolerance), we do not call out the "keep left/right" maneuver.

## Tasklist

 - [x] Add tests


src/odin/maneuversbuilder.cc
```diff
@@ -2354,15 +2424,13 @@ bool ManeuversBuilder::IsFork(int node_index,
       return false;
     };
 
-    // The curr edge and intersecting edge must be highway or ramp but not both
-    // validate lane bifurcation
     if (prev_edge->IsHighway() &&
         ((curr_edge->IsHighway() && (xedge->use() == TripLeg_Use_kRampUse)) ||
          (xedge->IsHighway() && curr_edge->IsRampUse())) &&
+        has_lane_bifurcation(trip_path_, node_index, prev_edge, curr_edge, xedge) &&
```
ktatterso, line 2430: Checking for the lane-bifurcation now has precedence over the fork-forward angle checks.

dgearhart:
@ktatterso I am trying to understand this diff between master and branch - since I do not see a deceleration lane

## Request: 
`j '{"locations":[{"lat":36.089000,"lon":-96.032177,"street":"I 44 West"},{"lat":36.068354,"lon":-96.065311,"street":"I 44 West"}],"costing":"auto","units":"miles"}'`

## Master
image

## Branch
image

## Transition point:
image

ktatterso:
@dgearhart  Brilliant find.  There are a couple things going on.  I'm testing a fix.

ktatterso:
I've tried a bunch of things to more accurately determine the length of a deceleration lane.  In short, I've empirically determined that 1) decel lane length as a function of speed is linear, and 2) I've run a whole bunch of routes so I could apply a linear regression to determine the best real world equation for "deceleration lane length as a function of speed".

dgearhart (approved):
Excellent - test results looked good to me. Thanks for the explanation of your analysis. Thank you @ktatterso :ship:

PR #3361, opened 2021-10-18
Probable restrictions

gknisely:
# Issue

This PR consumes and utilizes probable route restrictions.  The lowest probability allowed is set in the costing.  Currently, only using probable restrictions that have a 100% probability.   

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Documentation is stored in the `docs/` folder in this GitHub repository. It can 
 
 ### Docker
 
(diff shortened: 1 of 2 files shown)


gknisely:
> Seems like `OSMRestriction` should have a ctor that `memset`'s itself to all 0's (like `OSMWay`). Otherwise the member data is unset.

@ktatterso going to have a small follow up pr for this as this has been like this for some time.  Just don't want to mess up the functionality and this shouldn't, but you never know.

PR #3380, opened 2021-10-29
overhaul python bindings

### LINKED ISSUE #3370: Overhaul python bindings
a while ago I packaged the python bindings as binary wheels in a fork and while doing so I changed quite a few things. while I'll probably continue to use the fork for packaging, I'd like to take some of that stuff upstream. a non-exhaustive list of things I'd like to improve on (I'll keep on adding here over the next days):
- make the behavior more pythonic:
  - get rid of the `Configure` step and pass the config to `Actor`, that's been grieving me since forever
  - accept and return python `dict` in the actions: might require some pybind trickery
- include & use `valhalla_build_config` in the bindings to optionally build the config from python
- when #3281 is merged, we can also allow changing tilesets in a running session (what would actually be the consequences of having multiple `Actor` with different tilesets in the same session? just more stuff mmap'd, twice the cache etc?)

I think that's actually pretty exhaustive what I'd like to do. if you don't agree with anything let me know.. I'd start soonish (1-2 weeks), most is already done anyways in that fork.

nilsnolde:
fixes #3370 

all things done from the issue. will comment more inline.

tests will fail. will migrate them to use `unittest` as well (as the script tests do) once this has some initial approval


src/bindings/python/actor.py
```diff
@@ -0,0 +1,60 @@
...
+    def wrapped(*args):
+        if isinstance(args[1], dict):
+            return json.loads(func(args[0], json.dumps(args[1])))
+        elif not isinstance(args[1], str):
+            raise ValueError("Request must be either of type str or dict")
+        return func(*args)
+    return wrapped
+
+
+class Actor(_Actor):
```
kevinkreiser, line 20: appears to be missing transit availble

src/bindings/python/actor.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def decorator(func):
+    def wrapped(*args):
+        if isinstance(args[1], dict):
+            return json.loads(func(args[0], json.dumps(args[1])))
+        elif not isinstance(args[1], str):
+            raise ValueError("Request must be either of type str or dict")
+        return func(*args)
+    return wrapped
```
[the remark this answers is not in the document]
kevinkreiser: this makes me ask the question, do we want a json request/response interface for the actor itself, or do we want those to be dicts as well? dicts feels more pythonic but parsing and serializing json is wasteful if you dont actually need it. should we offer more options to the main api functions to specify what type of interface you are using? anyway not asking you to do anything just thinking about the design

src/bindings/python/config.py
```diff
@@ -0,0 +1,39 @@
...
+
+from .valhalla_build_config import config as default_config, Optional
+
+def _sanitize_config(dict_: dict = None) -> dict:
+    """remove the "Optional" values we get back from the config generator"""
+
+    int_dict_ = dict_.copy()
+    for k, v in int_dict_.items():
+        if isinstance(v, Optional):
+            del dict_[k]
```
kevinkreiser, line 12: we should add a method to the config generation that does this like `--minimal` or something

src/bindings/python/actor.py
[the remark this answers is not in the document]
nilsnolde: it's really wasteful.. on the other hand it's what a python client would almost surely do anyways (serializing a request dict, then parsing the response string), don't you think?

having a real JSON interface would be useful here (probably also for c++ consumers?). the python json module is entirely written in python, so definitely it'd be faster. but then we'd still have to come up with a ptree/rapidjson <-> dict converter. truth be told, I didn't try taking a `py::dict` (pybind type) as arg, so not sure what it natively casts to and how well it deals with nesting.

should I just leave a TODO for now?

test/bindings/python/valhalla.json
```diff
@@ -1,14 +1,46 @@
 {
+  "mjolnir": {
+    "max_cache_size": 1000000000,
+    "id_table_size": 1300000000,
+    "use_lru_mem_cache": false,
+    "lru_mem_cache_hard_control": false,
+    "use_simple_mem_cache": false,
```
nilsnolde, line 7: oops.. seems like that's the current of `valhalla_build_config`.. or at least what the python `json` library makes of it. hope that's ok.

scripts/valhalla_build_config
```diff
@@ -588,8 +588,19 @@ def override_config(args_: dict, leaves_: list, config_: dict):
         for k in keys[:-1]:
             v = v[k]
         v[keys[-1]] = args_.get(leaf.replace('\0', '_'))
-        if isinstance(v[keys[-1]], type(Optional())):
-            del v[keys[-1]]
+        sanitize_config(v)
```
kevinkreiser, line 591: this doesnt seem quite right, we replaced a single `if` with a recursive function. i think perhaps my earlier suggestion didnt make any sense! i had forgotten that of course we already do remove the optionals since they arent actually holding usable values. now i guess im wondering, can we just call `override_config` as is from the new improved bindings? sorry im a bit dense at the moment!

test/CMakeLists.txt
```diff
@@ -327,10 +327,10 @@ if(ENABLE_PYTHON_BINDINGS AND ENABLE_DATA_TOOLS)
       LOCPATH=${VALHALLA_SOURCE_DIR}/locales
       PYTHONPATH=src/bindings/python
       ASAN_OPTIONS=verify_asan_link_order=0:detect_leaks=0
-      /bin/bash -cx "${Python_EXECUTABLE} ${VALHALLA_SOURCE_DIR}/test/bindings/python/utrecht-test.py > test/python_valhalla.log"
+      /bin/bash -cx "${Python_EXECUTABLE} -m unittest discover -s ${VALHALLA_SOURCE_DIR}/test/bindings/python -v > test/python_valhalla.log 2>&1"
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
     DEPENDS
-      ${VALHALLA_SOURCE_DIR}/test/bindings/python/utrecht-test.py
+      ${VALHALLA_SOURCE_DIR}/test/bindings/python/test_utrecht.py
```
kevinkreiser, line 333: it looks like you didnt rename the file but rather deleted it only. did you maybe forget to add it back and push it? or am i just blind?

scripts/valhalla_build_config
nilsnolde: hehe, yeah the suggestion was borderline, I was about to say smth, then I thought like "well 2 lines saved, just do it". the code is "correct" in that it does the right thing, but I agree, it doesn't "seem" right:)

we can't just call `override_config` as-is, it's expecting lots of args like the config string with `\0` termination and so on. I'll pull that function again to the other file as it was before, don't really feel like touching again the slightly brittle logic in valhalla_build_config just to save a 3-4 lines of code. ok for you?

src/bindings/python/actor.py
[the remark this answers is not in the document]
nilsnolde: left a todo

test/CMakeLists.txt
nilsnolde: my bad..

nilsnolde:
from my side this is done @kevinkreiser . no hurry here though, if we get it merged within the next 1-2 weeks that'd be cool.

kevinkreiser (approved):

PR #3384, opened 2021-11-02
Update build instructions with new dependency

mandeepsandhu:


mandeepsandhu:
https://github.com/valhalla/valhalla/pull/3318 added a new dependency - `shapely` python module. While this commit [added the new dependency to macOS CI build](https://github.com/valhalla/valhalla/pull/3318/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R14), its not there for other linux platforms. TBH I don't know why linux CI builds work, since my local `make check` is broken without explicitly installing the python module.

Also, I'm recommending installing it via `apt` instead of `pip` since we don't have instructions/advocated for virtualenv's for our python dependencies.

kevinkreiser:
i also prefer this to pip. we should probably also do it that way in the docker deps script

kevinkreiser (approved):

nilsnolde:
ah damn, forgot the docs.. classic..

@mandeepsandhu that commit added the mac ci dep, this commit added the dep to docker before: https://github.com/valhalla/valhalla/pull/3383. with the linux ci lifecycle here we need to pre-merge new deps in the containers, since that's the env in ci.

I usually don't install these kinda modules via pip either but in mac ci there was already `requests` installed pip and didn't feel like rounds of ​ci building with brew/python/system stuff ;)

mandeepsandhu:
> @mandeepsandhu that commit added the mac ci dep, this commit added the dep to docker before: #3383. with the linux ci lifecycle here we need to pre-merge new deps in the containers, since that's the env in ci.

Ah ok, that explains it. Thanks!

PR #3391, opened 2021-11-09
Elevation tile on demand fetching

Neyromancer:
standalone on-demand elevation tile loader which allows to download elevation tiles from remote storage.

- added service add_elevation_on_demand - allows to fetch elevation from remote server based on provided tile and config
- added tests to test add_elevation_on_demand functionality


kevinkreiser:
Excellent I've been wanting to do this forever! I'll take a look in the next few days

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
    * FIXED: Some refactoring that was proposed for the PR 3379 [3381](https://github.com/valhalla/valhalla/pull/3381)
 
 * **Enhancement**
+   * ADDED: Ondemand elevation tile fetching [#3391(https://github.com/valhalla/valhalla/pull/3391)
```
kevinkreiser, line 9: Should be at the bottom of the list

scripts/valhalla_build_config
```diff
@@ -61,7 +61,10 @@ config = {
     }
   },
   'additional_data': {
-    'elevation': '/data/valhalla/elevation/'
+    'elevation': '/data/valhalla/elevation/',
+    'elevation_url': Optional(str),
+    'elevation_url_gz': Optional(bool),
+    'elevation_dir': Optional(bool)
```
kevinkreiser, line 67: isnt this redundant with the existing directory above? i agree the naming is unfortunate and it would be nice to be the same like `tile_dir` is but still we already have this so sadly i think we have to stick with just `elevation` rather than duplicate the idea

scripts/valhalla_build_config
```diff
@@ -329,7 +332,10 @@ help_text = {
     }
   },
   'additional_data': {
-    'elevation': 'Location of srtmgl1 elevation tiles for using in valhalla_build_tiles'
+    'elevation': 'Location of elevation tiles',
+    'elevation_url': 'Http location to read elevations from. this address is used if elevation tiles were not found in the elevation_dir. Ex.: http://your_valhalla_tile_server_host:8000/some/Optional/path/{DataPath}?some=Optional&query=params. Valhalla will look for the {DataPath} portion of the url and fill this out with a given elevation path when it makes a request for tha particular elevation',
+    'elevation_url_gz': 'Whether or not to request for compressed tiles true/false',
+    'elevation_dir': 'Location on remote server to read elevations from'
```
kevinkreiser, line 338: i feel like "remote" cant be the right word here, but as i said above i think this dir is redundant with elevation

src/mjolnir/elevationbuilder.cc
```diff
@@ -34,6 +34,148 @@ constexpr double POSTING_INTERVAL = 60;
 // Do not compute grade for intervals less than 10 meters.
 constexpr double kMinimumInterval = 10.0f;
 
+typedef std::unordered_map<uint32_t, std::tuple<uint32_t, uint32_t, float, float, float, float>>
+    cache_t;
+
+std::unordered_set<PointLL> get_coord(const std::string& tile_dir, const std::string& tile) {
```
kevinkreiser, line 40: this method is used only for testing? could it be moved to the test itself?

src/mjolnir/elevationbuilder.cc
```diff
@@ -34,6 +34,148 @@ constexpr double POSTING_INTERVAL = 60;
 // Do not compute grade for intervals less than 10 meters.
 constexpr double kMinimumInterval = 10.0f;
 
+typedef std::unordered_map<uint32_t, std::tuple<uint32_t, uint32_t, float, float, float, float>>
```
kevinkreiser, line 37: we use the newer  style for typedef `using cache_t = ...`

src/mjolnir/elevationbuilder.cc
```diff
@@ -63,102 +204,7 @@ void add_elevation(const boost::property_tree::ptree& pt,
     tilequeue.pop_front();
     lock.unlock();
 
-    // Get the tile. Serialize the entire tile?
-    GraphTileBuilder tilebuilder(graphreader.tile_dir(), tile_id, true);
```
kevinkreiser, line 67: this chunk was just moved out into the method above right? were there any actual changes to the code block?

src/mjolnir/elevationbuilder.cc
```diff
@@ -34,6 +34,148 @@ constexpr double POSTING_INTERVAL = 60;
...
+        resampled = valhalla::midgard::resample_spherical_polyline(shape, POSTING_INTERVAL);
+      }
+      for (auto&& el : resampled)
+        result.insert(std::move(el));
+    }
+  }
+  return result;
+}
+
+void add_elevations_to(GraphReader& graphreader,
```
kevinkreiser, line 76: the naming here is a bit odd, maybe `add_elevation_to_tile`?

src/mjolnir/elevationbuilder.cc
```diff
@@ -167,6 +213,35 @@ void add_elevation(const boost::property_tree::ptree& pt,
 namespace valhalla {
 namespace mjolnir {
 
+bool ElevationBuilder::load_tile_elevations(const std::string& tile,
```
kevinkreiser, line 216: the naming of this method is also odd, maybe `add_elevation`

test/sample.cc
```diff
@@ -186,6 +191,38 @@ TEST(Sample, hgt_file_name) {
...
+                               {{"http://s3.amazonaws.com/mapzen.valhalla{DataPath}"},
+                                {152.967888, -27.533658},
+                                {"http://s3.amazonaws.com/mapzen.valhalla/S28/S28E152.hgt"}},
+                               {{"http://s3.amazonaws.com/mapzen.valhalla{DataPath}"},
+                                {139.737345, 35.628096},
+                                {"http://s3.amazonaws.com/mapzen.valhalla/N35/N35E139.hgt"}},
+                               {{"http://s3.amazonaws.com/mapzen.valhalla{DataPath}"},
+                                {4.898484, 52.380697},
+                                {"http://s3.amazonaws.com/mapzen.valhalla/N52/N52E004.hgt"}}};
+
```
kevinkreiser, line 216: i mean this is great for test coverage but... its pretty expensive for a unit test right? off the top of my head i think each of these is like 25 megabytes in uncompressed form.

nilsnolde:
can you elaborate a little on what "on demand" means? didn't look too much into the changes, but it seems like it's mostly a new executable, from what I can see duplicating what we just added to `valhalla_build_elevation` python script (#3318 ), i.e. fetching elevation only for graph tile coverage? can you please say a few words how `valhalla_add_elevation` is adding functionality compared with `valhalla_build_elevation` @Neyromancer ? for now it's pretty confusing to me, but I might miss smth..

scripts/valhalla_build_config
Neyromancer: yeah, i decided to bring this as a proposal. i will remove "elevation_dir" and leave only "elevation" then.

src/mjolnir/elevationbuilder.cc
Neyromancer: no changes at this particular part, i just needed to reuse this logic in another method ["load_tile_elevations"](https://github.com/valhalla/valhalla/blob/4a991391bfc532afeecbb5e84e802aca975beb0a/valhalla/mjolnir/elevationbuilder.h#L19) and so decided to move it to a separate method so it will help me to avoid code repetition by reusing available one.

src/mjolnir/elevationbuilder.cc
Neyromancer: yeah, "add_elevation_to_tile" sounds much better

nilsnolde:
ah think I got it.. was just adjusting a planet build to only cover the graph with elevation and in the current situation I have to do the build in stages, stopping after the graph initially built, then run the py script to download the elevation, then start again at enhance. with this PR I guess you don't have to do multi-stage builds right? argh, then I wish I would've waited 2 more weeks before implementing exactly that in the py script, would've saved me a few hours :sweat_smile: much simpler now, thanks for that!

test/sample.cc
Neyromancer: well... i just consider more tests better than more space. space is cheap.

merkispavel:
> ah think I got it.. was just adjusting a planet build to only cover the graph with elevation and in the current situation I have to do the build in stages, stopping after the graph initially built, then run the py script to download the elevation, then start again at enhance. with this PR I guess you don't have to do multi-stage builds right? argh, then I wish I would've waited 2 more weeks before implementing exactly that in the py script, would've saved me a few hours 😅 much simpler now, thanks for that!

The general idea is not to fetch unnecessary elevation tiles(e.g around the oceans) and this is what both `valhalla_build_elevation.py` script and this PR changes are good at. 
1) For small input pbfs c++ on demand fetching might be a better choice(as you noticed)
2) Though I'm not sure about the whole-planet-build. This is where python script can win in the sense of performance compared to lazy on demand fetching: the script fetches all at once(rather than many individual fetches) so I guess it might be "cheaper"

Speaking of new `valhalla_add_elevation` binary. We deal with spot instances. So I'm thinking about caching the progress of the elevation stage not to lose progress when an instance dies. This is where I would need `valhalla_add_elevation` binary
```python
for tile in valhalla_tiles:
   if tile not processed:
    ./valhalla_add_elevation --graph-tile tile
    cache tile
```

We can probably move `valhalla_add_elevation` to the private fork not to burden public contributors to support it as it's usage is pretty specific. As you wish guys

nilsnolde:
right, thanks for the explanation @merkispavel ! maybe for planet py might be slightly cheaper. eventually I hope I don't have to download the full planet for a while though;)

IMHO `valhalla_add_elevation` is mildly specific but not a lot of code, could be useful for others and burden is minimal, so from me :+1:

test/sample.cc
Neyromancer: @kevinkreiser, these are not requests to a remote storage. this test is purely for checking if the function which substitutes "{DataPath}" with elevation tiles path works correctly.

kevinkreiser:
> then I wish I would've waited 2 more weeks before implementing exactly that in the py script, would've saved me a few hours ￼ much simpler now, thanks for that!

@nilsnolde yeah this is what i was trying to suggest in that recent pr from @molind , its great to see someone just knock this out

test/sample.cc
kevinkreiser: ah ok sorry, i was trying to do a quick pass on the review!

kevinkreiser:
> Though I'm not sure about the whole-planet-build. This is where python script can win in the sense of performance compared to lazy on demand fetching: the script fetches all at once(rather than many individual fetches) so I guess it might be "cheaper"

@merkispavel im not sure about that, i still think this implementation can win because we coalesce requests for the same tile to the same `future` and we write those tiles to the elevation cache dir so that subsequent requests will h it the directory first. the only way we could be even better is if we spawn a background thread that pre-warms the elevation directory cache while other parts of the tile building process are running

src/skadi/sample.cc
```diff
@@ -35,10 +34,43 @@ constexpr int16_t NO_DATA_HIGH = 16384;
...
 constexpr size_t TILE_COUNT = 180 * 360;
 constexpr int8_t UNPACKED_TILES_COUNT = 50;
+const constexpr char* kDataPathPattern = "{DataPath}";
 
-// macro is faster than inline function for this..
+// macro is faster than inline function for this...
 #define out_of_range(v) v > NO_DATA_HIGH || v < NO_DATA_LOW
 
+template <typename T>
+void parallel_call(const std::function<double(const T&)>& func,
```
kevinkreiser, line 43: nit: the name is too generic

src/skadi/sample.cc
```diff
@@ -35,10 +34,43 @@ constexpr int16_t NO_DATA_HIGH = 16384;
...
+                   std::vector<T> st,
+                   std::vector<double>& result) {
+  if (!func || st.empty())
+    return;
+
+  std::uint32_t size = std::max(1U, std::thread::hardware_concurrency());
```
kevinkreiser, line 49: its fine to use max parallelism as a default but this needs to be from config otherwise.

src/skadi/sample.cc
Neyromancer: well... it was done intentionally. As this function is just a utility function i tried to keep it as general as possible. i thought it might be good idea to move it to a separate unit and so to make it available to other library modules and reused if needed (probably not with this pr). what do you think?

Neyromancer: i can rename of course if you think it is better

src/skadi/sample.cc
```diff
@@ -35,10 +34,44 @@ constexpr int16_t NO_DATA_HIGH = 16384;
...
 
+template <typename T>
+void parallel_call(const std::function<double(const T&)>& func,
+                   std::vector<T> st,
+                   std::vector<double>& result,
+                   std::uint32_t num_threads = 1) {
+  if (!func || st.empty())
+    return;
+
+  std::uint32_t size = std::max(1U, std::min(std::thread::hardware_concurrency(), num_threads));
```
PureW, line 50: Why limit it to `hardware_concurrency`? If the work being split is IO bound like network requests, you can definitely run a lot more threads than number of cores.

src/skadi/sample.cc
```diff
@@ -449,11 +553,79 @@ template <class coords_t> std::vector<double> sample::get_all(const coords_t& co
...
+      return get_from_cache(coord);
+
+    st_.insert(uri);
+  }
+
+  LOG_INFO("Start loading data from remote server address: " + uri);
+  int repeat{3};
+  auto result = remote_loader_->get(uri);
+  while (repeat-- > 0 && result.status_ != baldr::tile_getter_t::status_code_t::SUCCESS) {
+    result = remote_loader_->get(uri);
```
PureW, line 606: Perhaps do some sleep here with exponential backoff?

valhalla/tile_server.h
```diff
@@ -8,9 +8,30 @@ struct context_t;
...
+    m_proxy_endpoint = end_point;
+  }
+
+  void set_request_interrupt(const std::string& interrupt_address) {
+    m_request_interrupt = interrupt_address;
+  }
+
+  void set_result_endpoint(const std::string& result_endpoint) {
+    m_result_endpoint = result_endpoint;
+  }
```
merkispavel, line 34: this may be redundant. What we need is to assign unique names for all these endpoints. I would rather append `m_url` orport to the existing names instead of setting endpoint of them manually

merkispavel (changes requested):
1. I'm not sure we need concurrent fetching on the `sample::get_all` level
2. I would think about merging `sample::get_from_remote` method with `cache_t` class so that we could just call `cache->tile(index)` and all fetching/saving/caching things are hidden in the cache class.
Maybe we can wrap existing `cache_t` class and enhance it with fetching logic

kevinkreiser:
In my opinion this is so incredibly similar to the graph tile fetching we should move that implementation into the curler header and just use it in both places (graph tile and sample). Otherwise we are duplicating the logic but with slight differences. The only real difference they need is to fetch different urls

src/skadi/sample.cc
```diff
@@ -449,11 +553,79 @@ template <class coords_t> std::vector<double> sample::get_all(const coords_t& co
...
+  auto tile = cache_->source(get_tile_index(coord));
+  if (!tile)
+    return get_no_data_value();
+
+  double u = (coord.first - std::floor(coord.first)) * (HGT_DIM - 1);
+  double v = (1.0 - (coord.second - std::floor(coord.second))) * (HGT_DIM - 1);
+  return tile.get(u, v);
+}
+
+template <class coord_t> double sample::get_from_remote(const coord_t& coord) {
```
maksimandrianov, line 586: clang-format?

Neyromancer:
thank you all for the detailed reviews. the code now is under refactoring. i will write a bit later when finish making the changes.

valhalla/tile_server.h
Neyromancer: just wanted to have general solution

src/skadi/sample.cc
Neyromancer: at the end decided to remove retry from here, it seems that its place in cur.h at the end. but left this improvement for later

src/skadi/sample.cc
Neyromancer: removed parallel_call

src/skadi/sample.cc
Neyromancer: original idea was to avoid the overhead of too many threads as this could potentially lead to performance decrease due to oversubscription. but i understood that if to play with the number of threads we could have more threads and achieve better performance.
anyway this function "parallel_call" was removed.

src/mjolnir/elevationbuilder.cc
```diff
@@ -34,22 +32,127 @@ constexpr double POSTING_INTERVAL = 60;
...
+    float max_down_slope = forward ? std::get<3>(found->second) : std::get<5>(found->second);
+    directededge.set_max_up_slope(max_up_slope);
+    directededge.set_max_down_slope(max_down_slope);
+  }
+
+  // Update the tile
+  tilebuilder.StoreTileData();
+
+  // Check if we need to clear the tile cache
+  if (graphreader.OverCommitted()) {
```
Neyromancer, line 134: test coverage not required here as the whole method emerged from previously existing method, it was done just for reuse

src/mjolnir/elevationbuilder.cc
```diff
@@ -34,22 +32,127 @@ constexpr double POSTING_INTERVAL = 60;
...
+
 /**
  * Adds elevation to a set of tiles. Each thread pulls a tile of the queue
  */
-void add_elevation(const boost::property_tree::ptree& pt,
-                   std::deque<GraphId>& tilequeue,
-                   std::mutex& lock,
-                   const std::unique_ptr<valhalla::skadi::sample>& sample,
-                   std::promise<uint32_t>& /*result*/) {
+void add_elevations_to_multiple_tiles(const boost::property_tree::ptree& pt,
```
Neyromancer, line 144: test coverage not required here as the whole method emerged from previously existing method, it was done just for reuse

src/mjolnir/elevationbuilder.cc
```diff
@@ -34,22 +32,127 @@ constexpr double POSTING_INTERVAL = 60;
...
+                                      std::promise<uint32_t>& /*result*/) {
   // Local Graphreader
   GraphReader graphreader(pt.get_child("mjolnir"));
 
   // We usually end up accessing the same shape twice (once for each direction along an edge).
   // Use a cache to record elevation attributes based on the EdgeInfo offset. This includes
   // weighted grade (forward and reverse) as well as max slopes (up/down for forward and reverse).
-  std::unordered_map<uint32_t, std::tuple<uint32_t, uint32_t, float, float, float, float>>
-      geo_attribute_cache;
+  cache_t geo_attribute_cache;
```
Neyromancer, line 155: only name of the type was changed not the type itself

src/mjolnir/elevationbuilder.cc
```diff
@@ -63,102 +166,7 @@ void add_elevation(const boost::property_tree::ptree& pt,
...
-    // Update the tile
-    tilebuilder.StoreTileData();
-
-    // Check if we need to clear the tile cache
-    if (graphreader.OverCommitted()) {
-      lock.lock();
-      graphreader.Trim();
-      lock.unlock();
-    }
+    add_elevations_to_single_tile(graphreader, lock, geo_attribute_cache, sample, tile_id);
```
Neyromancer, line 169: test coverage not required here as the whole method emerged from previously existing method, it was done just for reuse

src/mjolnir/elevationbuilder.cc
```diff
@@ -210,8 +247,8 @@ void ElevationBuilder::Build(const boost::property_tree::ptree& pt) {
   // Spawn the threads
   for (auto& thread : threads) {
     results.emplace_back();
-    thread.reset(new std::thread(add_elevation, std::cref(pt), std::ref(tilequeue), std::ref(lock),
-                                 std::ref(sample), std::ref(results.back())));
+    thread.reset(new std::thread(add_elevations_to_multiple_tiles, std::cref(pt), std::ref(tilequeue),
```
Neyromancer, line 250: test coverage not required here as the whole method emerged from previously existing method, it was done just for reuse

src/mjolnir/elevationbuilder.cc
```diff
@@ -210,8 +247,8 @@ void ElevationBuilder::Build(const boost::property_tree::ptree& pt) {
...
+                                 std::ref(lock), std::ref(sample), std::ref(results.back())));
```
Neyromancer, line 251: test coverage not required here as only parameter's names changed.

src/baldr/graphtile.cc
```diff
@@ -214,19 +221,15 @@ graph_tile_ptr GraphTile::CacheTileURL(const std::string& tile_url,
     return nullptr;
   }
 
-  auto uri = MakeSingleTileUrl(tile_url, graphid);
-  auto result = tile_getter->get(uri);
+  auto fname = valhalla::baldr::GraphTile::FileSuffix(graphid.Tile_Base(),
+                                                      valhalla::baldr::SUFFIX_NON_COMPRESSED, false);
+  tile_getter->set_path_pattern(valhalla::baldr::GraphTile::kTilePathPattern);
```
kevinkreiser, line 226: do we need to do this every time? why cant we do this once in the constructor?

valhalla/filesystem.h
```diff
@@ -491,4 +494,100 @@ inline std::chrono::time_point<std::chrono::system_clock> last_write_time(const
   return std::chrono::system_clock::from_time_t(FS_MTIME(s));
 }
 
+inline std::string generate_tmp_suffix() {
```
kevinkreiser, line 497: all of the valhalla specific stuff in here should be removed and put into a different header. this header is supposed to be a drop in replacement for the functionality we would get from `std::filesystem` if we were on c++17, we shouldnt put valhalla specific logic here. sorry for the confusion

valhalla/baldr/curl_tilegetter.h
```diff
@@ -1,11 +1,12 @@
 #pragma once
```
kevinkreiser, line 1: what if we added an optional string parameter on the end of the get function that, if non empty, will be the place to store the result to. that way we can:

1. avoid having to write the code both in sample and in graphtile
2. we can move the specific filesystem stuff here that we need

valhalla/mjolnir/elevationbuilder.h
```diff
@@ -16,6 +15,7 @@ class ElevationBuilder {
    * Add elevation information to the graph tiles.
    */
   static void Build(const boost::property_tree::ptree& pt);
+  static bool add_elevations(const std::string& tile, const boost::property_tree::ptree& pt);
```
kevinkreiser, line 18: we should keep this to one method `Build` so that it matches with the rest of the builders. What logic do we have in `valhalla_add_elevation` that is missing from the existing `Build` function? whether we called it from a standalone run of `valhalla_add_elevation` or from within `valhalla_build_tiles` it should do the same thing right? if the problem is about unit testing, i would suggest a private header which we can include for unit tests. let me know if this makes sense

Neyromancer: well... they are a bit different as Build tries to load elevation tiles for all valhalla tiles from the directory which is specified in the config file but valhalla_add_elevation should load elevation tiles only for a single valhalla tile passed as a parameter (although we still need config)

src/baldr/graphtile.cc
Neyromancer: right, my bad

src/mjolnir/valhalla_add_elevation.cc
```diff
@@ -0,0 +1,121 @@
...
+namespace bpo = boost::program_options;
+
+using namespace valhalla::baldr;
+using namespace valhalla::midgard;
+using namespace valhalla::mjolnir;
+
+enum Input { CONFIG, TILES };
+
+/*
+ * This service downloads elevations from remote storage for each provided tile.
```
kevinkreiser, line 22: ok this isnt a service though right, its a program that runs once and terminates. its not a daemon.

valhalla/mjolnir/elevationbuilder.h
kevinkreiser: i mean they are structured slightly differently but they do the same thing. the only real difference is once is inside of valhalla_build_tiles as a step in many steps, and the other is inside of valhalla_add_elevation as a single step.

both of them accomplish the same goal, pretty much the exact same way. they loop over all the tiles and add elevation to them. i definitely understand why we want to run build elevation into the tiles separate from the tile building process, but i dont understand why its useful to specifiy the list tiles separately. why isnt specifying the tileset (via config) enough?

its hard to understand why you would want to only add elevation to certain tiles in a tile set and not others or why you'd want to have to write code that asks whether or not the tile is in the tileset before fetching the elevation data. maybe you explain this use-case somewhere and i missed it?

merkispavel: @kevinkreiser the last part of the https://github.com/valhalla/valhalla/pull/3391#issuecomment-965057330 might answer your question. My intention is to split valhalla tiles into batches and cache the batch result when it's ready
However(!), it was a visible issue when we fectched useless over-the-ocean elevation tiles(1.5TB in total). As I noticed in the comment we might don't need it when we stop fetching those elevation tiles as the stage duration is linear to  the elevation tileset size

And as I noted in that comment - there's no need to push this binary to the public repo. We can probably move it to the private fork only as it is kind of beta(I haven't seen it in work yet)

kevinkreiser: sorry, maybe im not describing this very well. i think having a mode where we can fetch the data on demand and only pull the tiles that intersect the graphtiles is what ive always been advocating for. no problem there. all im saying is that it would be cool if the binary `valhalla_add_elevation` would be very thin wrapper around `ElevationBuilder::Build` basically take all the logic out of `valhalla_add_elevation` and put it in the `Build` method so that regardless of when we add the elevation, whether its in `valhalla_build_tiles` or in a separate call via `valhalla_add_elevation` its using the same code path. and the way that it works would be determined completely by config. so if someone didnt configure a tile url for elevation then it would only add elevation data that it found already int he directory, but if they did configure a tile directory and a url then it would fetch that data and store it in the tile directory while it was adding elevation. i'm just trying to keep things simple and not add so many ways to do the same thing.

i hope this makes sense just let me know if not, and we can always just have a quick video chat. its always possible that im misunderstanding why we cant do it like this :smile:

dnesbitt61: Sorry I haven't been following that closely, but will this on-demand fetching method work on systems with less storage? For example if I only have 200GB available for elevation and I want to process the whole planet? Does it require disk space for the entire set of elevation tiles accessed or will at cache them and remove as needed so new ones can be retrieved?

Neyromancer: for now automatic elevation disposal is not implemented but as for me this idea sounds really cool. maybe we could have optional parameter in our config for this

kevinkreiser: thats how the tile caching works for graph tiles as well (LRU). yet another opportunity to share code

src/mjolnir/elevationbuilder.cc
```diff
@@ -167,69 +206,46 @@ void add_elevation(const boost::property_tree::ptree& pt,
 namespace valhalla {
 namespace mjolnir {
 
-void ElevationBuilder::Build(const boost::property_tree::ptree& pt) {
-
-  auto hierarchy_properties = pt.get_child("mjolnir");
-  std::string tile_dir = hierarchy_properties.get<std::string>("tile_dir");
-
-  // Crack open some elevation data if its there. Return if it is not.
+void ElevationBuilder::Build(const boost::property_tree::ptree& pt, const std::string& tile) {
```
merkispavel, line 209: In my view it's the binary's responsibility to parse tileId string to tileId GraphId. I think it's better to have collection of graphIds for example
```suggestion
void ElevationBuilder::Build(const boost::property_tree::ptree& pt, std::deque<GraphId> tiles) {
```

Neyromancer: for now ElevationBuilder has everything which is needed for this job more over it does this already

cons of this proposal:
1. unnecessary complexity in stand alone binary (additional work)
2. if the logic of worknig with tilesets would change we would need to change in at least 2 places, in binary and in the current Build implementation
3. additional public method(not so negative effect but still additional code)

pros
1. can pass all tiles in one call(instead of calling Build in a loop)

kevinkreiser: i was thinking it would be all config driven. if nothing is provided in the config it uses the `tile_dir` to find out what graph tiles are in the tileset and gets elevation data for them. otherwise we can add a key to the `additional_data` part of the config and call it `elevation_shape`. its value can be the path to a file on the filesystem, it could point to a geojson file, it could point to a file with a list of tiles in it, it could point to whatever, but then its all configuration based and again then the `valhalla_add_elevation` can be made very stupid and all the logic can live inside of `Build`. thats just my opinion though.

kevinkreiser: it seems to me you guys want to have to tell the Build which tiles to get elevation for rather than build figure that out on its own. why is that? i cant understand the use case. the only thing that that allows you to add is the ability to only add elevation to a subset of tiles in your tileset. i suppose this could be useful if you had more than one elevation data source OR! if you were trying to distribute the processing so that you could do parts of the tileset in parallel on multiple machines.... i guess its the latter one then :smile:  ok cool, you guys should have just mentioned that up front, it makes it much easier to understand the motivations haha apologies if i missed that earlier. so yeah it makes sense here. i think passing a file with the list of tiles in it too the valhalla_add_elevation binary and passing a `deque` to Build makes sense. we just need to update the other calls to Build to compute the tileset first before calling, would it be ok to change it to an `unordered_set` since graphreader.gettileset already returns that type?

src/baldr/graphtile.cc
Neyromancer: at the end, i think i will disagree with this as CacheTileURL method (static method, not sure it is good to change this for now) receives tile_getter as a parameter and only before trying to load tiles from remote storage inside CacheTileURL we know that we have to look for kTilePathPattern. I mean GraphTile constructor knows nothing about tile_getter, GraphReader knows nothing about GraphTile implementation and tile_getter_t/curl_tile_getter_t can be used with different use cases where we don't need to search for kTilePathPattern.

kevinkreiser: I mean it should go right here:
https://github.com/valhalla/valhalla/blob/master/src/baldr/graphreader.cc#L485
In the tile getter constructor

kevinkreiser: Probably the real question I should have asked is why put the url pattern into tile getter at all?

Neyromancer: > I mean it should go right here: https://github.com/valhalla/valhalla/blob/master/src/baldr/graphreader.cc#L485 In the tile getter constructor

yeah, i understood, that is why i wrote that  "GraphReader knows nothing about GraphTile implementation". it just seems for me that this is work for the current GraphTile implementation.

valhalla/baldr/curl_tilegetter.h
Neyromancer: 2. filesystem.h cleaned, all valhalla specific functions removed and only general filesystem functions left.

Neyromancer: Concerning 1, as I understood you meant to hide the logic of storing data to filesystem inside the get function (from tile_getter_t). At first the get function will retrieve data from remote storage and then inside the same get function ( in case we pass additional parameter (path where to store data locally after retrieving them from remote storage) to this method) we will store the retrieved data.
As for me this sounds like not so good idea 
1.  difficult to guess from the function name that this method not only retrieves data but also stores them.
2. Do we always need/can to store data when we pass this additional parameter or not? 
3. Logic of storing data and logic of retrieving data from remote storage looks completely separate/unrelated, as for me. It might be better just to reuse (any of the already implemented ) cache from (e.g.) GraphReader, combine it with the logic from tile_getter_t and create something like StorageManager. StorageManager will check the cache and use the get function if nothing was found in the cache. This will help to keep code more structured, reusable and understandable.

valhalla/skadi/sample.h
```diff
@@ -31,6 +38,12 @@ class sample {
    */
   template <class coord_t> double get(const coord_t& coord);
 
+  /**
+   * Get a single sample from a remote source
+   * @param coord the single posting at which to sample the datasource
+   */
+  template <class coord_t> double get_from_remote(const coord_t& coord);
```
merkispavel, line 45: could you move it to the private/protected section? Looks like there's no need to reveal this method

valhalla/skadi/sample.h
```diff
@@ -42,6 +55,13 @@ class sample {
    */
   static double get_no_data_value();
 
+  /**
+   * @brief Add tile to cache and store it in local filesystem.
+   * @param[in] path path to the tile
+   * @param[in] raw_data Data to store.
+   */
+  bool store(const std::string& path, const std::vector<char>& raw_data);
```
merkispavel, line 63: could you move it to a private/protected section?

valhalla/skadi/sample.h
```diff
@@ -61,6 +81,21 @@ class sample {
...
+   * @param coord the single posting at which to sample the datasource
+   */
+  template <class coord_t> double get_from_cache(const coord_t& coord);
+
+private:
+  std::string url_;
+  std::unique_ptr<baldr::tile_getter_t> remote_loader_;
+  std::unordered_set<std::string> st_;
+  std::mutex st_lck_;
+  std::uint32_t num_threads_{1};
```
merkispavel, line 96: I guess this is an abandoned line the previous commits. Could drop it?

src/skadi/sample.cc
```diff
@@ -442,18 +439,89 @@ template <class coords_t> std::vector<double> sample::get_all(const coords_t& co
...
+  auto elev = get_hgt_file_name(index);
+  auto uri = baldr::make_single_point_url(url_, elev, remote_path_);
+
+  {
+    std::lock_guard<std::mutex> _(st_lck_);
+    if (st_.count(uri))
+      return get_from_cache(coord);
+
+    st_.insert(uri);
+  }
```
merkispavel, line 500: It is not clear how all these `get_from_remote` and `get_from_cache` works together
I'll come back to this thread soon, I need to think about it...

merkispavel (changes requested):
One major concern is how sample constructors look like. I guess there should be only one of the that accepts config

src/mjolnir/elevationbuilder.cc
merkispavel: you're kinda right @kevinkreiser . My idea was to use it in the distributed or/and "cache-able" pipeline when you split planet tileset into the batches(e.g 1x1 degree squres, or 2x2 degree, or just single tile in a batch) (see pseudo code in the [comment](https://github.com/valhalla/valhalla/pull/3391#issuecomment-965057330))
- process batches sequentially and cache result of the batch not to lose progress
- process batches in parallel so caching might not be even needed
Yeah, this is not a secret at all :) we should've mentioned it somewhere in the PR description

src/baldr/graphreader.cc
```diff
@@ -549,8 +549,6 @@ class TarballGraphMemory final : public GraphMemory {
 // Get a pointer to a graph tile object given a GraphId. Return nullptr
 // if the tile is not found/empty
 graph_tile_ptr GraphReader::GetGraphTile(const GraphId& graphid) {
-  // TODO: clear the cache automatically once we become overcommitted by a certain amount
-
```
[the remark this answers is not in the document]
Neyromancer: cache is checked below

src/baldr/graphtile.cc
```diff
@@ -205,28 +197,38 @@ void GraphTile::SaveTileToFile(const std::vector<char>& tile_data, const std::st
...
+  }
+}
+
 graph_tile_ptr GraphTile::CacheTileURL(const std::string& tile_url,
                                        const GraphId& graphid,
                                        tile_getter_t* tile_getter,
                                        const std::string& cache_location) {
   // Don't bother with invalid ids
-  if (!graphid.Is_Valid() || graphid.level() > TileHierarchy::get_max_level()) {
+  if (!graphid.Is_Valid() || graphid.level() > TileHierarchy::get_max_level() || !tile_getter) {
```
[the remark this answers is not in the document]
Neyromancer: i could add of course but this comment looks redundant as the idea is clear from the code. @merkispavel, what do you think?

src/tile_server.cc
```diff
@@ -100,19 +100,15 @@ worker_t::result_t disk_work(const std::list<zmq::message_t>& job,
...
-  std::string request_interrupt = "ipc:///tmp/http_test_request_interrupt";
-  std::string proxy_endpoint = "ipc:///tmp/http_test_proxy_endpoint";
-
+  std::string result_endpoint{"ipc:///tmp/http_test_result_endpoint" + m_url};
+  std::string request_interrupt{"ipc:///tmp/http_test_request_interrupt" + m_url};
+  std::string proxy_endpoint{"ipc:///tmp/http_test_proxy_endpoint" + m_url};
   // server
   std::thread server(std::bind(&http_server_t::serve,
-                               http_server_t(context, "tcp://*:8004", proxy_endpoint + "_upstream",
+                               http_server_t(context, "tcp://" + m_url, proxy_endpoint + "_upstream",
```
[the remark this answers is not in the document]
Neyromancer: yeah, indeed. it might be better to changed to the original implementation then

valhalla/baldr/tilegetter.h
```diff
@@ -57,5 +57,21 @@ class tile_getter_t {
...
 };
 
+/**
+ * @brief Build uri address to make remote call
+ * @name[in] tile_url Base url address
+ * @name[in] fname File to call for
+ * @name[in] remote_path This parameter is used only in testing
+ * @return full uri address
+ */
+inline std::string make_single_point_url(const std::string& tile_url,
```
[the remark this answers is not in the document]
Neyromancer: right, my thoughts were that this function is very general and could belong to the class higher in the hierarchical order but i agree and move it to curl_tile_getter.

valhalla/skadi/sample.h
Neyromancer: it is still used in [sample](https://github.com/valhalla/valhalla/blob/5eb2fdd718103452f516b729f0d913be50050d12/src/skadi/sample.cc#L369) construction

merkispavel: I mean `num_threads_` is not used anywhere beside `constructor`

src/skadi/sample.cc
merkispavel: I think the current implementation doesn't handle a case correctly when two threads ask for the same tile
For example
- 1 thread called `get_from_remote`. There's no such tile on the filesystem so it started fetching it.
- 2 thread called `get_from_remote`. The check above(`st.count(uri)`) is passed as the first thread already added such key. However, the tile has been fetched completely yet

So we need some kind of blocking for the same tile. Maybe something similar to what we have at the moment(I mean `pending_tiles` https://github.com/valhalla/valhalla/blob/dd8699dc8436ec9da69c5dc6925f51b65471febb/src/skadi/sample.cc#L258)

Neyromancer: i think i found different solution at the end but let's discuss this one either.

merkispavel: I think there are a few main points I'm trying to say
- If multiple threads want to fetch the same tile - only one thread initiates fetching and other threads should wait(block) for the result
- If I'm not mistaken the current (master) implementation doesn't assume  that tiles suddenly land on the filesystem. So we need to adopt current implementation for on-demand use case and make it thread safe
- Make sure everything else is thread safe and wrapped with synchronization(mutex-es) if needed

src/skadi/sample.cc
```diff
@@ -431,74 +419,135 @@ template <class coords_t> std::vector<double> sample::get_all(const coords_t& co
...
   }
-  name.append(std::to_string(y));
-  name.append(name);
 
-  name.append(x < 0 ? "W" : "E");
-  x = std::abs(x);
-  if (x < 100) {
-    name.push_back('0');
+  LOG_INFO("Start loading data from remote server address: " + uri);
+  auto result = remote_loader_->get(uri);
```
merkispavel, line 497: I believe there's still possible to fetch same tile from multiple threads because you split mutex-ed logic into 2 blocks
```
1) check with lock
2) remote_loader->fetch
3) mark tile as fetch with lock
```

It is possible that multiple threads pass 1) and start fetching tile in 2) at the same time. However, it's better to block all other threads if one of them has already started fetching the tile(or had intention and grabbed the lock)

test/CMakeLists.txt
```diff
@@ -73,6 +73,15 @@ if (UNIX AND ENABLE_SINGLE_FILES_WERROR)
     PROPERTIES COMPILE_FLAGS "-Wall -Werror")
 endif()
 
+if (APPLE)
+  find_package(OpenSSL REQUIRED)
+#  target_link_libraries(${test}  OpenSSL::SSL)
+#  target_link_libraries(${test} OpenSSL::Crypto)
+  target_link_libraries(valhalla_test OpenSSL::SSL)
+  target_link_libraries(valhalla_test OpenSSL::Crypto)
+endif()
```
kevinkreiser, line 82: can you explain why we need to link this?

Neyromancer: already removed, i was thinking to use openssl to calculate file's hashsum to check if valhalla tiles are the same after applying elevation tiles (elevation tiles loaded from remote storage vs local storage). But this solution was changed to just checking files byte by byte.

valhalla/filesystem.h
```diff
@@ -491,4 +494,65 @@ inline std::chrono::time_point<std::chrono::system_clock> last_write_time(const
   return std::chrono::system_clock::from_time_t(FS_MTIME(s));
 }
 
+/**
+ * @brief Saves data to the path.
+ * @attention Will replace the contents in case if fpath already exists. Will create
+ * new directory if directory did not exist before hand.
+ * */
+template <typename T> inline bool save(const std::string& fpath, const T& data = {}) {
```
merkispavel, line 502: You're assuming that data is a container with sequential memory layout(e.g, vector, array). Does it make sense to specify it explicitly?
```suggestion
template <typename T> inline bool save(const std::string& fpath, const vector<T>& data = {}) {
```

src/skadi/sample.cc
```diff
@@ -456,9 +443,8 @@ template <class coord_t> double sample::get(const coord_t& coord) {
 
   // get the proper source of the data
   const auto& tile = cache_->source(index);
-  if (!tile) {
-    return NO_DATA_VALUE;
-  }
+  if (!tile)
+    return get_from_remote(coord);
```
merkispavel, line 447: I believe it's better to implement it like this
```cpp
const auto& tile = cache_->source(index);
if (!tile) {
  // fetch tile
  fetch_from_remote(tile);
  // try to load it from the filesystem one more time
  tile = cache_->source(index);
}
```
In this case you don't need to have any coordinate-related logic in the fetch_from_remote/get_from_remote: it will do pure fetching

src/skadi/sample.cc
```diff
@@ -474,74 +460,136 @@ template <class coords_t> std::vector<double> sample::get_all(const coords_t& co
...
-  y = std::abs(y);
-  if (y < 10) {
-    name.push_back('0');
+  auto index = get_tile_index(coord);
+  auto elev = get_hgt_file_name(index);
+  auto uri = baldr::make_single_point_url(url_, elev, remote_path_);
+
+  {
+    std::shared_lock<std::shared_timed_mutex> _(st_lck);
+    if (st_.count(uri))
```
merkispavel, line 532: It's probably redundant. `If tile exists on filesystem`-check is enough. I mean if we invoke `get_from_remote` function we're sure that we need to fetch this tile

src/skadi/sample.cc
```diff
@@ -474,74 +460,136 @@ template <class coords_t> std::vector<double> sample::get_all(const coords_t& co
...
 
-  return name;
+  auto res = get_from_cache(coord);
+  if (res == get_no_data_value()) {
+    LOG_WARN("Fail to load data from remote server address: " + uri);
+    return get_no_data_value();
+  }
+
+  LOG_INFO("Data loaded from remote server address: " + uri);
+  return res;
```
merkispavel, line 562: I would drop this piece of code to separate fetching logic from the get-elevation-for-coordinate logic.

src/skadi/sample.cc
```diff
@@ -456,9 +443,8 @@ template <class coord_t> double sample::get(const coord_t& coord) {
 
   // get the proper source of the data
   const auto& tile = cache_->source(index);
```
merkispavel, line 445: When we fetch tile - we do cache insert with locked mutex. Need to double check if we should lock here as well...

merkispavel (changes requested):

src/skadi/sample.cc
```diff
@@ -474,74 +461,146 @@ template <class coords_t> std::vector<double> sample::get_all(const coords_t& co
...
+    LOG_WARN("Fail to load data from remote server address: " + uri);
+    return get_no_data_value();
   }
-  if (x < 10) {
-    name.push_back('0');
+
+  {
+    std::lock_guard<std::shared_timed_mutex> _(st_lck);
+    if (!st_.count(uri)) {
+      // TODO(neyromancer): переиспользовать функционал кэша, можно избавиться от st_.
```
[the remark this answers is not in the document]
Neyromancer: this comment was for me and the fact that you saw it during your final review means that the file was outdated. last my commit was 9 days ago and files had not changed since then. i message you on slack after i updated all the files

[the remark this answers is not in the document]
Neyromancer: this comment was for me. it just happened that you used outdated code. messaged you on skack about update.

src/skadi/sample.cc
```diff
@@ -474,74 +460,136 @@ template <class coords_t> std::vector<double> sample::get_all(const coords_t& co
...
     values.emplace_back(value);
   }
+
   return values;
 }
 
-double sample::get_no_data_value() {
-  return NO_DATA_VALUE;
-}
+template <class coord_t> double sample::get_from_cache(const coord_t& coord) {
```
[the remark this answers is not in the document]
Neyromancer: get call from remote either, this function get data only form cache

src/baldr/graphtile.cc
```diff
@@ -104,9 +97,19 @@ graph_tile_ptr GraphTile::DecompressTile(const GraphId& graphid,
 graph_tile_ptr GraphTile::Create(const std::string& tile_dir,
                                  const GraphId& graphid,
                                  std::unique_ptr<const GraphMemory>&& traffic_memory) {
-
   // Don't bother with invalid ids
   if (!graphid.Is_Valid() || graphid.level() > TileHierarchy::get_max_level() || tile_dir.empty()) {
+    std::string error_msg;
+    if (!graphid.Is_Valid())
+      error_msg += "GraphId is invalid\n\n";
```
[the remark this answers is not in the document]
Neyromancer: at the end i decided to remove logging here as on building it creates a lot of unnecessary noise

src/baldr/graphtile.cc
```diff
@@ -104,9 +97,19 @@ graph_tile_ptr GraphTile::DecompressTile(const GraphId& graphid,
...
+
+    if (graphid.level() > TileHierarchy::get_max_level())
+      error_msg += "GraphId level exceeds tile hierarch max level\n\n";
```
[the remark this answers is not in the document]
Neyromancer: the same as above

src/skadi/sample.cc
Neyromancer: the idea was to reduce the number of times we are trying to access filesystem but as discussed yesterday i will remove this as the whole method is going to change.

valhalla/filesystem.h
Neyromancer: needed it here https://github.com/valhalla/valhalla/blob/3146fc64fbb992a1754516f9bf8a7de4a018df0c/test/elevation_builder.cc#L407

src/skadi/sample.cc
Neyromancer: i don't see why a lock is needed here. it seems that the state of the cache is guarded inside of the source method

src/skadi/sample.cc
Neyromancer: yeah, it seems it will look better

src/skadi/sample.cc
```diff
@@ -460,11 +447,13 @@ template <class coord_t> double sample::get(const coord_t& coord, tile_data& til
 
   // the caller can pass a cached tile, so we only fetch one if its not the one they already have
   if (index != tile.get_index()) {
-    // get the proper source of the data
-    tile = cache_->source(index);
-  }
-  if (!tile) {
-    return NO_DATA_VALUE;
+    if (!(tile = cache_->source(index))) {
```
merkispavel, line 450: I think there's still a race between the following calls
1) `cache_source(index) ->   auto& item = cache[index];`
2)  `fetch(index) -> store -> cache_->insert -> cache[pos].init(..)`

2nd call is safe(as you lock `cache_lck` on `cache_->insert`), however 1st call is not under mutex

```suggestion
    {
      std::lock_guard<std::mutex> guard(cache_lck);
      tile = cache_->source(index);
    }
    if (!tile) {
```

merkispavel (changes requested):
Thanks for the fixes 👍 . I think there's the last data race that need to be fixed(see comment in 
 `sample::get`)
Plus I'm going to test it locally. And I think we are good  after that

src/skadi/sample.cc
Neyromancer: reason i didn't place lock there:
threads

As i understood you are concerned of the case when we have **thread 4** and **thread 2**.
In my view we don't have race condition here because there is a lock  before insert method ( inside store method ) which limits access to this method to a single thread. This means that even if **thread 4** would get nullptr at cache[index] ( inside source method ) at the time when **thread 2** get value from the method get ( inside fetch method ) we would still successfully store the result with **thread 2** and initialise the cache with respective value. When the **thread 4** arrive we would have 2 possibilities: the store method is still locked or already unlocked but in both cases the **thread 4** will just contain the same data as the **thread 2** a moment ago and so we would not spoil the data.

merkispavel: The concern is about how `cache_item_t` is accessed. 
From the writer side(`fetch` method):
1) lock the mutex
2) call `cache_item_t::init method`

From the reader side(`get` method):
1) get cache_item_t
2) if it is unitialized(nullptr) - call `item.init`([link](https://github.com/valhalla/valhalla/blob/4d02424fa5c60e1c65fd692b51bfcab80e9037a2/src/skadi/sample.cc#L334))

So it's possible that `cache_item_t::init` is called from the 2 threads at the same time. This is one of the possible races beside others(if compiler reorders instructions)

In theory, you could omit locking the mutex on reader side if `cache_item_t` was completely thread safe. But it is not

merkispavel (changes requested):

src/mjolnir/elevationbuilder.cc
```diff
@@ -63,174 +164,62 @@ void add_elevation(const boost::property_tree::ptree& pt,
...
   std::vector<std::shared_ptr<std::thread>> threads(nthreads);
-
-  // Setup promises. Hold the results for the threads
   std::vector<std::promise<uint32_t>> results(nthreads);
 
-  LOG_INFO("Adding elevation to " + std::to_string(tilequeue.size()) + " tiles with " +
-           std::to_string(nthreads) + " threads...");
-
-  // Spawn the threads
+  LOG_INFO("Loading elevations to tiles");
```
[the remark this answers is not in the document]
Neyromancer: I just think that information which have implementation details should be at debug level, not info

merkispavel (approved):
thank you!

merkispavel:
I'm sorry that it took so long. And thank you for the work @Neyromancer. I tested it locally: works as expected

@kevinkreiser Let us know if you'd like to take a look or it's ok to merge it

valhalla/baldr/curl_tilegetter.h
```diff
@@ -63,7 +63,7 @@ class curl_tile_getter_t : public tile_getter_t {
  * @return full uri address
  */
 inline std::string make_single_point_url(const std::string& tile_url,
-                                         const std::string& fname,
```
merkispavel: const reference handles temporary objects. There is no need to change it

merkispavel (approved):
re-approve

PR #3393, opened 2021-11-10
Fix out-of-range error in GeoPoint::HeadingAtEndOfPolyline

oleg-liatte:
# Issue

Fix iterator decrement beyond begin()

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/midgard/pointll.cc
```diff
@@ -290,6 +290,10 @@ PrecisionT GeoPoint<PrecisionT>::HeadingAtEndOfPolyline(const std::vector<GeoPoi
                     pt1->lat() + ((pt0->lat() - pt1->lat()) * pct));
         return ll.Heading(pts[idx1]);
       } else {
+        if (pt0 == pts.begin()) {
```
kevinkreiser, line 293: ok so this happens when we never have enough total length to reach the distance threshold. can you make one small clean up here and remove the else case completely? it just adds more nesting where its not needed

oleg-liatte: Indeed. Done.

kevinkreiser (approved):

PR #3394, opened 2021-11-10
Do not create statsd client in workers if it is not configured

mskurydin:
# Issue

Right now the statsd client is created all the time.
If there is no configuration, the defaults are used.
However, under some conditions we noticed unwanted side effects.

For example, the use of `getaddrinfo` during initialization may cause DNS requests under the hood.
In the case of the bogus network at the start, it may cause uncontrollable hangs.

The idea here is not to pay for what is not used.

## Tasklist

 - [ ] Add tests

### DIFF
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -1143,50 +1143,46 @@ void service_worker_t::set_interrupt(const std::function<void()>* interrupt_func
   interrupt = interrupt_function;
 }
 void service_worker_t::cleanup() {
-  // sends metrics to statsd server over udp
   if (statsd_client) {
+    // sends metrics to statsd server over udp
     statsd_client->flush();
   }
 }
 void service_worker_t::enqueue_statistics(Api& api) const {
   // nothing to do without stats
-  if (!api.has_info() || api.info().statistics().empty())
+  if (!statsd_client || !api.has_info() || api.info().statistics().empty())
     return;
 
   // these have been filled out as the request progressed through the system
   for (const auto& stat : api.info().statistics()) {
     float frequency = stat.has_frequency() ? stat.frequency() : 1.f;
-
-    if (statsd_client) {
-      switch (stat.type()) {
-        case count:
-          statsd_client->count(stat.key(), static_cast<int>(stat.value() + 0.5), frequency,
-                               statsd_client->tags);
-          break;
-        case gauge:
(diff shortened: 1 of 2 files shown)


PureW (dismissed):

src/worker.cc
```diff
@@ -1142,7 +1144,9 @@ void service_worker_t::set_interrupt(const std::function<void()>* interrupt_func
 }
 void service_worker_t::cleanup() {
   // sends metrics to statsd server over udp
-  statsd_client->flush();
+  if (statsd_client) {
```
kevinkreiser, line 1147: this is unfortunate. we should have just checked if no host was set here: https://github.com/vthiery/cpp-statsd-client/blob/master/include/cpp-statsd-client/UDPSender.hpp#L184 and returned immedately. then all of this would be unneeded

src/worker.cc
```diff
@@ -1153,31 +1157,34 @@ void service_worker_t::enqueue_statistics(Api& api) const {
...
     float frequency = stat.has_frequency() ? stat.frequency() : 1.f;
 
-    switch (stat.type()) {
-      case count:
-        statsd_client->count(stat.key(), static_cast<int>(stat.value() + 0.5), frequency,
-                             statsd_client->tags);
-        break;
-      case gauge:
-        statsd_client->gauge(stat.key(), static_cast<unsigned int>(stat.value() + 0.5), frequency,
+    if (statsd_client) {
```
kevinkreiser, line 1160: to reduce the diff can you change this one and the others to be more like

```c++
if (!statsd_client)
  return;
```

src/worker.cc
kevinkreiser: https://github.com/vthiery/cpp-statsd-client/pull/40

kevinkreiser (dismissed):

src/worker.cc
```diff
@@ -1132,27 +1132,30 @@ struct statsd_client_t : public Statsd::StatsdClient {
...
-  statsd_client->flush();
+  if (statsd_client) {
+    // sends metrics to statsd server over udp
+    statsd_client->flush();
+  }
 }
 void service_worker_t::enqueue_statistics(Api& api) const {
   // nothing to do without stats
-  if (!api.has_info() || api.info().statistics().empty())
+  if (!statsd_client || !api.has_info() || api.info().statistics().empty())
```
kevinkreiser, line 1153: :+1:

kevinkreiser (approved):

PR #3397, opened 2021-11-11
WIP:  add docker - build check

ImreSamu:
# Issue

Testing - in Container "test"

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/docker/Dockerfile-run
+++ b/docker/Dockerfile-run
@@ -14,6 +14,9 @@ RUN mkdir build
 WORKDIR /usr/local/src/valhalla/build
 RUN cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo
 RUN make all -j$(nproc)
+RUN make utrecht_tiles
+RUN make -j$(nproc) tests
+RUN make -j$(nproc) check
 RUN make install
 
 # we wont leave the source around but we'll drop the commit hash we'll also keep the locales
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,7 @@ jobs:
           name: Build Valhalla Docker Images
           command: |
             docker build --pull -f ./docker/Dockerfile-build --tag valhalla/valhalla:build-latest ./docker
+            docker build -f ./docker/Dockerfile-run --tag valhalla/valhalla:run-latest .
             if [[ "${CIRCLE_BRANCH}" == "master" ]] || [[ ! -z "${CIRCLE_TAG}" ]]; then
               docker build -f ./docker/Dockerfile-run --tag valhalla/valhalla:run-latest .
             fi


nilsnolde:
hm.. did you have build/run-time problems with the run image @ImreSamu ? 

IMHO this does not add considerably to a more releases, mostly it slows down the `-run` image build considerably. while it's totally possible that a `-run` build fails bcs of outdated dependencies from the `-build` base image, we'd as well catch that during circleci, which uses the `-build` image as environment. the `-run` image doesn't really do much itself before we build valhalla, it's just these lines:
```
ADD . /usr/local/src/valhalla
RUN ls
RUN git submodule sync && git submodule update --init --recursive
RUN mkdir build

# configure the build with symbols turned on so that crashes can be triaged
WORKDIR /usr/local/src/valhalla/build
```

so basically adding this pretty expensive test (15 mins or so) in the docker image will only really test those fairly generic lines. what do you think?

docker/Dockerfile-run
```diff
@@ -14,6 +14,9 @@ RUN mkdir build
 WORKDIR /usr/local/src/valhalla/build
 RUN cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo
 RUN make all -j$(nproc)
+RUN make utrecht_tiles
+RUN make -j$(nproc) tests
+RUN make -j$(nproc) check
```
kevinkreiser, line 19: i mean i know we can do this, but i wonder what is the point. the run docker build only happens when we merge to `master` and we only ever do that after all the checks have passed. this just feels like something that will take extra time in CI but not benefit us.

ImreSamu:
Sorry for the noise;  I will close this PR.

Context /  My "root" problem: 
* my local docker build is failing ( with some special tunings );
* I have re-checked the current master:  I have added 3 extra `make` (( utrecht_tiles, check, test ))  to the `Dockerfile-run` - and the check failing again :cry:  

in the https://github.com/valhalla/valhalla/blob/master/.circleci/config.yml  - the `make check`  is running in the `build-release` -part;
but the [`cmake` parameters are different](https://github.com/valhalla/valhalla/blob/master/.circleci/config.yml#L111-L113)
```cmake
          cd build && cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=On -DENABLE_PYTHON_BINDINGS=On \
                -DLOGGING_LEVEL=TRACE \
                -DCPACK_GENERATOR=DEB -DCPACK_PACKAGE_VERSION_SUFFIX="-0ubuntu1-$(lsb_release -sc)" -DENABLE_SANITIZERS=ON
```

vs.  in the Dockerfile-run very simple `cmake` is calling : https://github.com/valhalla/valhalla/blob/master/docker/Dockerfile-run#L15         
```
RUN cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo
```

IMHO:  in extreme cases, different cmake can also cause problems. ( with extremely low probability )
Summary:  I will be happier if the 2 `cmake` will be the same.  ( the `build-release`  version and the `docker-run` version )

ps:
The build error was probably caused by some local configuration error; ( still debugging; re-installing; etc .. )

kevinkreiser:
@ImreSamu its a valid point you make we dont do exactly the same invokation for the build we could probably make them mostly the same, python bindings are on by default so that is superfluous, shared trace logging has caused failures in the past when it wasnt part of ci but now that it is that shouldnt be your problem because by default our log level is info (ie in the docker build). that leaves 3 things. the first two are about testing code distribution (shared libs and debian packaging), imho those are a very unimportant difference and wouldnt cause you problems (unless you rely on those specifically). that leaves the sanitizers which indeed can catch things that wouldnt be caught in the simple docker build but again they are in the CI build so it would be unlikely that CI passes and then the docker runner build fails since its less strict.

in any case if you care to share what your "special tunings" are we might be able to give you some info about where to look. just let us know!

PR #3398, opened 2021-11-11
Correct ETA calculation

CuriOusQuOkka:
# Issue

- [x] Smoothing live traffic so that it won't be used after a long time. Use a certain percentage of live speed only in the next hour relative to the current time.
- [x] support `bidirectional_a*` algo for `depart_at` query. It uses recosting based on the passing time (opposite to invariant behaviour)
- [x] Update timedependent_forward_a* to build routes depending on the correct speeds. 

## Tasklist

 - [x] Add tests

## Requirements / Relations


kevinkreiser:
fading out realtime speeds and fading in historical or default speeds is what we always wanted to do, thanks for making that happen! but i think there is a major problem with the current way this is implemented. the API supports the date time type `invariant`, this was already shipped in a previous version. we cant just change the way it works without allow users to opt into that new behavior. basically when someone requests it they should expect that the time of day doesn't vary along the route. i know its annoying but if we want to allow a mode that is essentially "run bidirectional astar with a single time of day but when we go to do the final costing actually treat it like a time dependent route in terms of the evolution of time along the path" we need a way for the user to choose that new behavior. otherwise we have broken the API. you need another time type or a way to tell a the API you want bidirectional astar with a time that varies. perhaps thats what we should do, allow the user to say i want to bidirectional astar even though im using depart_at, arrive_by, or current time with the caveat that today the calculated route is using invarient time but recosts at the end to vary time over the route. basically leave invarient intact in the recosting etc but add a new option that allows using bidirectional astar with time varied routes.

there is also a smaller nitpick in the relatioinship between `DynamicCost::EdgeCost` and `GraphTile::GetSpeed`. these could use a giant disclaimer regarding the new parameter `travel_time_seconds`. i saw you put "since the beginning" but its really important that people dont call EdgeCost in the reverse direction or they'll have the fading happening from the end of the route as well. its too bad there isnt a good way to protect this from happening. its kind of like we want only recosting to be allowed to call it with `travel_time_seconds` because then we can be sure its only in the forward direction

src/sif/recost.cc
```diff
@@ -26,7 +25,6 @@ void recost_forward(baldr::GraphReader& reader,
                     float source_pct,
                     float target_pct,
                     const baldr::TimeInfo& time_info,
-                    const bool invariant,
```
kevinkreiser, line 29: removing this is a breaking API change. but we can su pport what you want to do in another way. see my larger pr comment for details or let me know if you want to meet to talk about it in more detail.

kevinkreiser (changes requested):

src/sif/autocost.cc
```diff
@@ -232,7 +232,8 @@ class AutoCost : public DynamicCost {
   virtual Cost EdgeCost(const baldr::DirectedEdge* edge,
                         const graph_tile_ptr& tile,
                         const uint32_t seconds,
-                        uint8_t& flow_sources) const override;
+                        uint8_t& flow_sources,
+                        const uint32_t travel_time_seconds) const override;
```
maksimandrianov, line 236: Usually immutable parameters go first, and mutable ones go last. Could you change the order please? and below

CuriOusQuOkka:
@kevinkreiser
I was checking how the invariant option appeared in the code. As I understand from this PR https://github.com/valhalla/valhalla/pull/2660, originally it was included to have more consistent version of bidirectional A*. Because when we go in a reverse direction, we can not predict the timestamp at which the edge will be passed. To avoid incorrect behaviour connected to this uncertainty the invariant parameter was used.

After that in PR https://github.com/valhalla/valhalla/pull/2697, invariant was added to **recosting** and **TripLegBuilder** functionality. It was not the main fix of the PR, but looks like it was written to finish the full pipeline of the parameter. Let's see how it works.

image

I’ve compared ETA on a route from Portland to Seattle (time on the plot is in UTC). Route takes around 3 hours. The routing algorithm (bidirectional a*) returns the same path at each time-point (as it was not fixed). The only difference is the recosting function (as in #3398 PR). 
* If you look at the base chart you see that the ETA rises significantly close to the rush hours, at 8am Portland time or 4pm Portland time, it happens daily.
* Fixed version takes into account the arrival point as well. ETA starts rising a few hours earlier, around 6am, so it includes rush hours in Seattle too and makes ETA calculations much better. 


In the [documentation](https://github.com/valhalla/valhalla/blame/master/docs/api/turn-by-turn/api-reference.md#L309) I see  

> 3 - Invariant specified time. Time does not vary over the course of the path.

Nothing is said about the estimated time of arrival, the original problem was connected to the path building. For me current behaviour looks like a bug: when someone uses a routing algorithm, they want to see a **correct** ETA on a **given** route. Currently we use a bad model for ETA calculation and there is no need to use it when a better solution is proposed.

I believe it would make much more sense if we fix the description of the parameter than to create another one.

kevinkreiser:
@CuriOusQuOkka 

> To avoid incorrect behaviour connected to this uncertainty the invariant parameter was used ... let's see how it works

the point was to be able to use the more performant bidirectional a* algorithm with traffic data. traffic data requires a time component to the algorithm so you could look up the speed at that time. invariant (meaning not varying) time allows us to add this to bidira* because we don't need to guess the time at the end or beginning of the route, we have just one time. one time that is used for the computation of the route, costing, elasped time, etc. it is therefore not a bug but intended behavior. the fact that i didnt implement it all in a single pr but we got it cleaned up in a few subsequent prs is irrelevant. the important part is we didnt do a release until it was complete. sometimes its better to break things up into small pieces rather than rewrite everything in one go you know?

> and makes ETA calculations much better

i am definitely not arguing that picking a constant time will do better than varying the time over the route when it comes to ETA accuracy. that is not my argument and that was never the point. the change in this PR is neither a "fix" nor a novel idea. in fact we intended to do exactly this blending of live and historical conditions all along. its why I added the time tracking object in the first place, so we could easily send that info to GetSpeed to compute the coefficients of the linear combination. actually what i wanted to do was not just recost and vary the time along the recosting but i also wanted to maybe try machine learning a way to pick a close estimate of what the ending time of a route would be so that we could even try to properly support timed restrictions etc. basically have all the functionality in bidirectional a* and completely remove the forward and reverse a* variants that do time dependence. in any case, i would not and am not arguing against varying the time to get better ETA results. i am arguing that you are breaking the API by hijacking the meaning of "invariant"

> Nothing is said about the estimated time of arrival, the original problem was connected to the path building

no the original problem was connected to using traffic conditions in the bidirectional a* algorithm. regarding documentation, yes, you're right the documentation doesn't spell out every single computation that is done when the user specifies invariant time. it doesnt do that for any of the other time options either. it is assumed that you understand that the whole of the computation uses that single time that you specified. i dont see how you could make any other assumption based on the documentation. do you actually think a user should assume that if they selected: `type: invariant` and `value: 2016-03-29T08:00` that after the route has been computed we then recost it using the current traffic conditions? that is complete nonsense and i guess it brings me to a bug in this pr.

i noticed you arent using the proper travel time when calling EdgeCost. you are supposed to be using `TimeInfo::seconds_from_now`. What it looks like you have done is just assumed that every route is a route that is starting at the current time because you seem to just be passing the elapsed seconds from the start of the route. the problem is the route doesnt have to start at "now" it can start tomorrow or last week or whenever.

anyway... my whole point isnt that i want you to rewrite your PR. 90% of it is good. all i am asking you to do leave invariant intact, the way it is, and add a new parameter that lets you call bidirectional a* with varying time. something like `"force_bidirectional": true`. then you can do exactly what you wanted. call the api with `type: current` and have recosting at the end vary the time starting at `now`. of course the time wont vary during the path computation and you can note that in the documentation if you wish and say that in a future version when we figure out how to properly do time dependent bidirectional a* that it will vary and become "more correct". let me know if that makes sense or not.

CuriOusQuOkka:
> i noticed you arent using the proper travel time when calling EdgeCost. you are supposed to be using TimeInfo::seconds_from_now. What it looks like you have done is just assumed that every route is a route that is starting at the current time because you seem to just be passing the elapsed seconds from the start of the route. the problem is the route doesnt have to start at "now" it can start tomorrow or last week or whenever.

This should not be the problem because in case you pass not current time live traffic is ignored on flow_mask evaluation stage.

kevinkreiser:
> This should not be the problem because in case you pass not current time live traffic is ignored on flow_mask evaluation stage.

true, but it shouldnt be. if i pass a time of 5 minutes ago and eventually the time reaches current time then it would be nice to use the current conditions. that was the whole point of the timeinfo seconds_from_now, which i still suggest you should use

src/sif/recost.cc
CuriOusQuOkka: fixed it, added an option for using invariant parameter on path postprocessing.

valhalla/baldr/graphtile.h
```diff
@@ -535,51 +535,62 @@ class GraphTile {
...
-    //               the request is not for "now", or we're some X % along the route
+    //               the request is not for "now"
     // TODO(danpat): for short-ish durations along the route, we should fade live
     //               speeds into any historic/predictive/average value we'd normally use
+
+    // This parameter describes the weight of live-traffic on a specific edge. In the beginning of the
+    // route live-traffic gives more information about current congestion situation. But the further
+    // we go the less consistent this traffic is. We prioritize predicted traffic in this case.
+    // Want to have a smooth decrease function.
+    float live_traffic_multiplier = 1. - std::min(travel_time_seconds / 3600., 1.);
```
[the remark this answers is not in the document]
CuriOusQuOkka: thanks

CuriOusQuOkka:
@kevinkreiser updated the PR using seconds from now variant. It sounds useful as flow_mask can be set in a more flexible way

proto/options.proto
```diff
@@ -227,4 +227,5 @@ message Options {
   optional bool linear_references = 45;                                   // Include linear references for graph edges returned in certain responses.
   repeated CostingOptions recostings = 46;                                // Costing options to use to recost a path after it has been found
   repeated Ring exclude_polygons = 47;                                    // Rings/polygons to exclude entire areas during path finding
+  optional bool invariant_postprocess = 48 [default = true];              // Use invariant time after a route is built, on recosting or leg building stages.
```
kevinkreiser, line 230: this is so strange to me. is there something you dont like about my suggestion to avoid misusing the word `invariant`? from this i understand you want to keep using the `invarient` date time type but then you want recosting to ignore it and use the `depart_at` type. i think it would be way less confusing for the api to just allow a `date_time.type` of  `depart_at/current` (or even `arrive_by` once we have `recost_reverse`) to go to the bidirectional algorithm, where it can just not vary the time (for now) while finding the path and then use either the `depart_at` time or `current` time in the recosting part as specified in the request.

let me know if you have some time to talk this over. i think hijacking `invariant` to do what you want is poor API design. It would be easier if you let the requester say they want to use date_time in bidirectional astar. there is another option that i dont like which would be to add a second date_time parameter to be used in recosting only. but its a bad option as the recosting that already exists is complicated enough.

maybe im missing something and it has to be the way you've laid out, but i cant see the reason why. do let me know when you have the chance.

kevinkreiser (dismissed):

src/thor/bidirectional_astar.cc
```diff
@@ -1270,7 +1270,7 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
     // recost edges in final path; ignore access restrictions
     try {
       sif::recost_forward(graphreader, *costing_, edge_cb, label_cb, source_pct, target_pct,
-                          time_info, invariant, true);
+                          time_info, invariant && options.invariant_postprocess(), true);
```
kevinkreiser, line 1273: so based on my previous comments, here for example we would change it to something like this:

```suggestion
                          time_info, options.has_date_time_type() ? options.date_time_type() : boost::none, true);
```

valhalla/baldr/graphtile.h
```diff
@@ -535,51 +535,71 @@ class GraphTile {
...
-    //               the request is not for "now", or we're some X % along the route
     // TODO(danpat): for short-ish durations along the route, we should fade live
     //               speeds into any historic/predictive/average value we'd normally use
+
+    constexpr double LIVE_SPEED_FADE = 1. / 3600.;
+    // This parameter describes the weight of live-traffic on a specific edge. In the beginning of the
+    // route live-traffic gives more information about current congestion situation. But the further
+    // we go the less consistent this traffic is. We prioritize predicted traffic in this case.
+    // Want to have a smooth decrease function.
+    float live_traffic_multiplier = 1. - std::min(seconds_from_now * LIVE_SPEED_FADE, 1.);
```
kevinkreiser, line 565: :+1:

proto/options.proto
CuriOusQuOkka: I got what you mean by `force_bidirectional` sounds like a good idea. The second with additional parameter will make the code complicated and messy, agree with you. I'll implement the solution with `force_bidirectional` on Wednesday, OOO now.

CuriOusQuOkka: I've implemented a draft version without tests for now. Bidirectional algo will use invariant time for all date_time_types untill we find a good implementation that works with time updates. @kevinkreiser What do you think about it now? Recosting uses invariant only when it was specified in date_time_type.

CuriOusQuOkka: I use `prioritize_bidirectional`, because bidirectional a* has problems on short routes, when origin and destination edges are connected. Don't want to call it in such cases

proto/options.proto
```diff
@@ -227,4 +227,5 @@ message Options {
   optional bool linear_references = 45;                                   // Include linear references for graph edges returned in certain responses.
   repeated CostingOptions recostings = 46;                                // Costing options to use to recost a path after it has been found
   repeated Ring exclude_polygons = 47;                                    // Rings/polygons to exclude entire areas during path finding
+  optional bool prioritize_bidirectional = 48 [default = false];          // Prioritize bidirectional a* over timedependent_forward a*.
```
kevinkreiser, line 230: yep, but only when a time dependent (depart_at arrive_by or current time) route is requested!

src/sif/recost.cc
```diff
@@ -121,8 +121,11 @@ void recost_forward(baldr::GraphReader& reader,
     Cost transition_cost = node ? costing.TransitionCost(edge, node, label) : Cost{};
     // update the cost to the end of this edge
     uint8_t flow_sources;
+    const int64_t seconds_from_now =
+        int64_t(offset_time.seconds_from_now) * (offset_time.negative_seconds_from_now ? -1 : 1);
```
kevinkreiser, line 125: yep i wondered what you would want to do here. i suppose one could do fading on either side of "now" and just ignore the sign, but i see you took the other approach. i think it would be interesting to actually allow it though. i dont have a strong opinion on it so what you did is fine with me.

valhalla/baldr/graphtile.h
```diff
@@ -535,51 +535,71 @@ class GraphTile {
...
+    if (seconds_from_now < 0) {
+      live_traffic_multiplier = 0;
+    }
```
kevinkreiser, line 568: this is what i was just talking about, you could actually totally ignore the sign and let the fading happen on either side of seconds_from_now == 0.

src/thor/bidirectional_astar.cc
```diff
@@ -507,8 +507,9 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
   PointLL destination_new(destination.path_edges(0).ll().lng(), destination.path_edges(0).ll().lat());
   Init(origin_new, destination_new);
 
+  // Set invariant true for bidirectional a*, change it when the algo allows varying time.
+  bool invariant = true;
```
kevinkreiser, line 511: this doesnt seem right, because you can still make a route that doesnt ask for timedependence. i didnt think about it all the way but i was kind of expecting this to be the change:

    // we use a non varying time for all time dependent routes until we can figure out how to vary the time during
    // the path computation in the bidirectional algorithm
    bool invariant = options.has_date_time_type();

src/thor/route_action.cc
```diff
@@ -351,7 +351,8 @@ thor::PathAlgorithm* thor_worker_t::get_path_algorithm(const std::string& routet
 
   // If the origin has date_time set use timedep_forward method if the distance
   // between location is below some maximum distance (TBD).
-  if (origin.has_date_time() && options.date_time_type() != Options::invariant) {
+  if (origin.has_date_time() && options.date_time_type() != Options::invariant &&
+      !options.prioritize_bidirectional()) {
```
kevinkreiser, line 355: :+1:

src/thor/bidirectional_astar.cc
kevinkreiser: so actually i guess we can completely delete this boolean and modify `Expand` directly... let me make another comment elsewhere in the diff

kevinkreiser: maybe the variable would be better labeled as `time_dependent` and then over in `Expect` we could add the comment that currently bidira* cant vary the time for time dependent routes so we use `0` seconds for offset? any thoughts to make this more clear ))

kevinkreiser:
this looks great to me! i only saw two issues, the one with the boolean in bidira* which technically doesnt matter because they do the same thing but i think its unclear the way it is now and the other one where i was thinking you can fade current conditions on either side of "now" ie even if the seconds_from_now is negative. let me know what you think about those two comments. otherwise this looks done to me! thanks for dealing with my annoying reviews )

src/sif/recost.cc
CuriOusQuOkka: your idea sounds useful! Also it requires less operations so I modified this logic.

CuriOusQuOkka:
I've made time performance check for query with specific `date_time` value. Here are quantiles for `response_time`
query | 5% | 50 % | 95 % | 99%
-- | -- | -- | -- | --
master + invariant time | 0.0292 | 0.0628 | 0.1179 | 0.1403
branch + invariant time | 0.0291 | 0.0632 | 0.1193 | 0.1422
branch + depart_at | 0.0291 | 0.0622 | 0.1180 | 0.1410

src/thor/bidirectional_astar.cc
CuriOusQuOkka: I would leave it invariant, it is more clear what it is created for in this case. Once we update the algo and time updates will be possible, we can separate behaviour. Now the behaviour is more visible.

@kevinkreiser What did you mean by `Expect` comment? I've updated the documentations to emphasize this behaviour.

CuriOusQuOkka:
@kevinkreiser thanks for your help! Your proposals significantly improved the logic here, it looks much better now 👍

src/thor/bidirectional_astar.cc
kevinkreiser: Sorry it was a very stupid typo, kids were probably yelling in the background at the time... I meant the `Expand` method haha sorry for that

kevinkreiser: And yeah I'm fine leaving it I just thought some way to call attention to that spot would be good so future us will understand/remember it

src/sif/autocost.cc
```diff
@@ -232,7 +232,8 @@ class AutoCost : public DynamicCost {
   virtual Cost EdgeCost(const baldr::DirectedEdge* edge,
                         const graph_tile_ptr& tile,
                         const uint32_t seconds,
-                        uint8_t& flow_sources) const override;
+                        uint8_t& flow_sources,
+                        const uint64_t seconds_from_now) const override;
```
genadz, line 236: Can we pass one parameter `time_info` instead of `seconds` and `seconds_from_now` ?

genadz (changes requested):

src/sif/autocost.cc
```diff
@@ -491,9 +492,10 @@ bool AutoCost::ModeSpecificAllowed(const baldr::AccessRestriction& restriction)
 Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
                         const graph_tile_ptr& tile,
                         const uint32_t seconds,
-                        uint8_t& flow_sources) const {
+                        uint8_t& flow_sources,
+                        const uint64_t seconds_from_now) const {
   // either the computed edge speed or optional top_speed
-  auto edge_speed = tile->GetSpeed(edge, flow_mask_, seconds, false, &flow_sources);
+  auto edge_speed = tile->GetSpeed(edge, flow_mask_, seconds, false, &flow_sources, seconds_from_now);
```
genadz, line 498: and here the same. looks like we can replace both `seconds` and `seconds_from_now` by `time_info`

src/thor/bidirectional_astar.cc
```diff
@@ -507,8 +507,10 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
   PointLL destination_new(destination.path_edges(0).ll().lng(), destination.path_edges(0).ll().lat());
   Init(origin_new, destination_new);
 
+  // we use a non varying time for all time dependent routes until we can figure out how to vary the
+  // time during the path computation in the bidirectional algorithm
+  bool invariant = options.has_date_time_type();
```
genadz, line 512: reasonable

test/gurka/test_speed_smoothing.cc
```diff
@@ -0,0 +1,211 @@
...
+        {"AB", {{"highway", "primary"}}},
+    };
+    const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10000);
+    map = gurka::buildtiles(layout, ways, {}, {}, "test/data/small_route_speeds",
+                            {
+                                {"mjolnir.shortcuts", "false"},
+                                {"mjolnir.timezone", VALHALLA_BUILD_DIR "test/data/tz.sqlite"},
+                            });
+    map.config.put("mjolnir.traffic_extract", "test/data/small_route_speeds/traffic.tar");
+    map.config.put("thor.invariant_recosting", false);
```
genadz, line 23: have we introduced a new option in the config ?

test/gurka/test_speed_smoothing.cc
```diff
@@ -0,0 +1,211 @@
...
+TEST_F(SmallRouteSpeedTest, LiveTrafficSpeed) {
+  auto reader = test::make_clean_graphreader(map.config.get_child("mjolnir"));
+  for (auto tile_id : reader->GetTileSet()) {
+    auto tile = reader->GetGraphTile(tile_id);
+    for (const auto& e : tile->GetDirectedEdges()) {
+      auto speed = tile->GetSpeed(&e, baldr::kCurrentFlowMask, 0);
+      EXPECT_EQ(speed, current_);
+
+      speed = tile->GetSpeed(&e, baldr::kCurrentFlowMask, 10000);
+      EXPECT_EQ(speed, current_);
```
genadz, line 75: could you add more comments what you check ? (I see comments only for some `EXPECT_EQ` checks)

test/gurka/test_speed_smoothing.cc
```diff
@@ -0,0 +1,211 @@
...
+
+      uint8_t* flow_sources = nullptr;
+      speed = tile->GetSpeed(&e, baldr::kCurrentFlowMask, 0, false, flow_sources, 0);
+      EXPECT_EQ(speed, current_);
+
+      int highway_speed = 75;
```
genadz, line 81: how did you get this values ?

test/gurka/test_speed_smoothing.cc
```diff
@@ -0,0 +1,211 @@
...
+  static gurka::map map;
+  static uint32_t current_, historical_max_, historical_min_, constrained_, freeflow_;
+};
+
+gurka::map SmallRouteSpeedTest::map = {};
+uint32_t SmallRouteSpeedTest::current_ = 0, SmallRouteSpeedTest::historical_max_ = 0,
+         SmallRouteSpeedTest::historical_min_ = 0, SmallRouteSpeedTest::constrained_ = 0,
+         SmallRouteSpeedTest::freeflow_ = 0;
+
+TEST_F(SmallRouteSpeedTest, LiveTrafficSpeed) {
```
genadz, line 66: add a description for this test

test/gurka/test_traffic_routes.cc
```diff
@@ -0,0 +1,342 @@
...
+                       make_request("A", "B", speeds, false, 0, "2021-11-08T00:00")));
+  // depart_at = current_date_time
+  auto date_time = date_time_in_N_minutes(0);
+  results.emplace_back(gurka::do_action(valhalla::Options::route, map,
+                                        make_request("A", "B", speeds, false, 1, date_time)));
+  results.emplace_back(gurka::do_action(valhalla::Options::route, map,
+                                        make_request("A", "B", speeds, true, 1, date_time)));
+  for (const auto& result : results) {
+    // time_dependent_forward_a* is used for one-edge routes.
+    EXPECT_EQ(result.trip().routes(0).legs(0).algorithms(0), "time_dependent_forward_a*");
```
genadz, line 127: why do we need this check ?

test/gurka/test_traffic_routes.cc
```diff
@@ -0,0 +1,342 @@
...
+    EXPECT_EQ(result.trip().routes(0).legs(0).algorithms(0), "time_dependent_reverse_a*");
+  }
+}
+
+TEST_F(TrafficSmoothing, AlgoPrioritizationWithDepartAt) {
+  std::string speeds = "\"predicted\",\"current\"";
+
+  auto result = gurka::do_action(valhalla::Options::route, map,
+                                 make_request("A", "D", speeds, false, 1, "2021-11-08T00:00"));
+  EXPECT_EQ(result.trip().routes(0).legs(0).algorithms(0), "time_dependent_forward_a*");
```
genadz, line 150:
```suggestion
  ASSERT_EQ(result.trip().routes(0).legs(0).algorithms(0), "time_dependent_forward_a*");
```

test/gurka/test_traffic_routes.cc
```diff
@@ -0,0 +1,342 @@
...
+  auto time_dependent_eta = calculate_eta(result);
+
+  result = gurka::do_action(valhalla::Options::route, map,
+                            make_request("A", "D", speeds, true, 1, "2021-11-08T00:00"));
+  EXPECT_EQ(result.trip().routes(0).legs(0).algorithms(0), "bidirectional_a*");
```
genadz, line 155:
```suggestion
  ASSERT_EQ(result.trip().routes(0).legs(0).algorithms(0), "bidirectional_a*");
```

test/gurka/test_traffic_routes.cc
```diff
@@ -0,0 +1,342 @@
...
+  auto bidirectional_eta = calculate_eta(result);
+
+  // ETA calculated by two algos should be equal
+  EXPECT_EQ(time_dependent_eta, bidirectional_eta);
```
genadz, line 159: I would compare floating values with `EXPECT_NEAR`

valhalla/baldr/graphtile.h
```diff
@@ -535,51 +535,67 @@ class GraphTile {
    *                       week so we modulus the time to day based seconds
    * @param  flow_sources  Which speed sources were used in this speed calculation. Optional pointer,
    *                       if nullptr is passed in flow_sources does nothing.
+   * @param  seconds_from_now  Absolute number of seconds from now till the moment the edge is passed.
+   * Be careful when setting the value in reverse direction algorithms to use proper value. It affects
+   * the percentage of live-traffic usage on the edge. The bigger seconds_from_now is set the less
+   * percentage is taken. Currently this parameter is set to 0 when building a route with reverse and
+   * bidirectional a*.
```
genadz, line 542:
```suggestion
   * @param  seconds_from_now  Absolute number of seconds from now till the moment the edge is passed.
   *                           Be careful when setting the value in reverse direction algorithms to use proper value. It affects
   *                           the percentage of live-traffic usage on the edge. The bigger seconds_from_now is set the less
   *                           percentage is taken. Currently this parameter is set to 0 when building a route with reverse and
   *                           bidirectional a*.
```

src/sif/autocost.cc
kevinkreiser: for sure, that was what i was originally planning to do, encapsulate all the timing information into one thing that we coudl pass around. its great to finally do this last part

valhalla/baldr/graphtile.h
CuriOusQuOkka: It is done automatically by `scripts/format.sh`. I can not set it as you show whatever I do :(

src/sif/autocost.cc
CuriOusQuOkka: GetSpeed can not be changed in this PR, as timeinfo object refers it through graphreader, I'll try to fix this relation in another PR. (It's strange that TimeInfo class has dependance on graph

src/sif/autocost.cc
CuriOusQuOkka: Moved to time_info for EdgeCost function. @genadz, @kevinkreiser could you please check the latest commit? Not 100% sure if setting invariant time info is a good option everywhere.

src/thor/unidirectional_astar.cc
```diff
@@ -236,9 +236,8 @@ inline bool UnidirectionalAStar<expansion_direction, FORWARD>::ExpandInner(
 
   // Compute the cost to the end of this edge
   uint8_t flow_sources;
-  auto edge_cost = FORWARD ? costing_->EdgeCost(meta.edge, tile, time_info.second_of_week,
-                                                flow_sources, time_info.seconds_from_now)
-                           : costing_->EdgeCost(opp_edge, t2, time_info.second_of_week, flow_sources);
+  auto edge_cost = FORWARD ? costing_->EdgeCost(meta.edge, tile, time_info, flow_sources)
+                           : costing_->EdgeCost(opp_edge, t2, time_info, flow_sources);
```
genadz, line 240: before we pass `seconds_from_now=0` for reverse search, but now we pass the whole `time_info`. what is the value of `time_info.seconds_from_now`

CuriOusQuOkka: I validated, that time is calculated correctly starting from the endpoint, and back to the origin. So it is a good option to use our upgraded traffic model. In case `arrive_by` is in 10 hours, we will not use live traffic. But when it is in 30 minutes speed will be interpolated both from `predicted` and `live`. ETA will be more correct in this case.

test/gurka/test_speed_smoothing.cc
CuriOusQuOkka: I moved the tests to test_traffic_routes file. Added comments everywhere

test/gurka/test_speed_smoothing.cc
CuriOusQuOkka: It was a default value on the edge, hardcoded for highways in valhalla. I've set another one in the test, not to have this magic

test/gurka/test_traffic_routes.cc
CuriOusQuOkka: sounds reasonable, no matter what the algo is, live traffic should be used

test/gurka/test_traffic_routes.cc
CuriOusQuOkka: fixed

test/gurka/test_traffic_routes.cc
CuriOusQuOkka: agree

src/sif/dynamiccost.cc
```diff
@@ -176,7 +176,8 @@ bool DynamicCost::AllowMultiPass() const {
 // default parameters and inheritance (which are a bad mix)
 Cost DynamicCost::EdgeCost(const baldr::DirectedEdge* edge, const graph_tile_ptr& tile) const {
   uint8_t flow_sources;
-  return EdgeCost(edge, tile, kConstrainedFlowSecondOfDay, flow_sources);
+  baldr::TimeInfo time_info = TimeInfo::invalid();
```
genadz, line 179: why do we need local variable here ?

genadz (approved):

src/sif/dynamiccost.cc
CuriOusQuOkka: agree, put it as a parameter

CuriOusQuOkka:
@genadz I added mapmatch tests, could you look into the latest commits?

genadz (approved):

PR #3399, opened 2021-11-12
add heading to PathLocation::PathEdge for /locate 

nilsnolde:
we'd like to add the correlated point's `heading` to the `/locate` response.

as it's not important info for any other than loki::search I guess it just wasn't worth keeping track of the heading. we added it to PathLocation::PathEdge struct, but don't carry it over to PBF.

this is WIP, first I'd like to hear your thoughts about it.

### DIFF
--- a/src/meili/candidate_search.cc
+++ b/src/meili/candidate_search.cc
@@ -87,7 +87,7 @@ CandidateCollector::WithinSquaredDistance(const midgard::PointLL& location,
         } else if (dist == 0.0) {
           snapped_node = opp_edge->endnode();
         }
-        correlated.edges.emplace_back(edgeid, dist, point, sq_distance, 0.f);
+        correlated.edges.emplace_back(edgeid, dist, point, sq_distance);
       }
     }
 
@@ -106,7 +106,7 @@ CandidateCollector::WithinSquaredDistance(const midgard::PointLL& location,
         } else if (dist == 0.0) {
           snapped_node = edge->endnode();
         }
-        correlated.edges.emplace_back(opp_edgeid, dist, point, sq_distance, 0.f);
+        correlated.edges.emplace_back(opp_edgeid, dist, point, sq_distance);
       }
     }
 
--- a/src/meili/candidate_search.cc
+++ b/src/meili/candidate_search.cc
@@ -87,7 +87,7 @@ CandidateCollector::WithinSquaredDistance(const midgard::PointLL& location,
         } else if (dist == 0.0) {
           snapped_node = opp_edge->endnode();
         }
-        correlated.edges.emplace_back(edgeid, dist, point, sq_distance);
+        correlated.edges.emplace_back(edgeid, dist, point, sq_distance, 0.f);
       }
     }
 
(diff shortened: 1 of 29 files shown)


kevinkreiser:
i feel like we may as well add it to the pbf. i had thought that we didnt even compute the angle if the heading wasnt asked for but it looks like we changed that since the last time my memory took a snapshot of the code. other than that, this already looks good to me

valhalla/baldr/pathlocation.h
```diff
@@ -35,7 +35,8 @@ struct PathLocation : public Location {
              const double score,
              const SideOfStreet sos = NONE,
              const unsigned int outbound_reach = 0,
-             const unsigned int inbound_reach = 0);
+             const unsigned int inbound_reach = 0,
+             const float projected_heading = 0.f);
```
kevinkreiser, line 39: this should probably be some sentinal value that indicates not valid. `-1.f` for example.

kevinkreiser: the other option is to change the order of the parameters and make it a manditory parameter. although that is more code change its more of the kind of change we are making here. basically every path edge must have a heading. another problem with that one though is that there may be several callers outside of loki::search that dont know or care about the heading.

kevinkreiser:
looks good to me. we should just tie up that loose end about differentiating between not set and actually a value of 0

valhalla/baldr/pathlocation.h
nilsnolde: I made it mandatory, there were hardly any other callers other than in tests

src/meili/candidate_search.cc
```diff
@@ -87,7 +87,7 @@ CandidateCollector::WithinSquaredDistance(const midgard::PointLL& location,
         } else if (dist == 0.0) {
           snapped_node = opp_edge->endnode();
         }
-        correlated.edges.emplace_back(edgeid, dist, point, sq_distance);
+        correlated.edges.emplace_back(edgeid, dist, point, sq_distance, 0.f);
```
nilsnolde, line 90: this method doesn't seem to be used.. should I do it properly anyways? would require some changes to the helper method meili::helpers::Project to take a shape instead of a lazy_shape

nilsnolde: since it's not used I don't know if heading is even important to whoever is calling this method.. doesn't sound like it judging by the name `WithinSquareDistance`

kevinkreiser: Heading is not used for mapmatching. But 0 is certainly not correct. And this method is used but it's templates so it's hard for IDEs to tell. I feel like given this we probably should have just made it optional and defaulted to -1.sorry I should have done more looking at the calls before adding that as a potential sokution

nilsnolde: found that in the veeery end.. very annoying..

nilsnolde: jup done

nilsnolde:
not sure if it came out, this is done I think @kevinkreiser . no rush though :)

kevinkreiser (approved):

PR #3401, opened 2021-11-13
Add support for LZ4 compressed elevation tiles.

lseelenbinder:
~This is not fully baked, but since I'm by no means a CMake or C++ expert, I wanted to ensure this was a reasonable start.~

~I'll still need to add tests and some additional documentation.~

According to the tests, it does what it says on the cover. :smile: 

# Issue

Adds drop-in support for lz4 elevation tiles. Briefly discussed here: https://github.com/valhalla/valhalla/issues/3071#issuecomment-967591165

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -14,8 +14,8 @@
 #include <unordered_set>
 
 #include <boost/optional.hpp>
-#include <sys/stat.h>
 #include <lz4frame.h>
+#include <sys/stat.h>
 
 #include "baldr/compression_utils.h"
 #include "filesystem.h"
@@ -62,7 +62,7 @@ int16_t flip(int16_t value) {
 
 uint64_t file_size(const std::string& file_name) {
   // TODO: detect gzip and actually validate the uncompressed size?
-  struct stat s{};
+  struct stat s {};
   int rc = stat(file_name.c_str(), &s);
   return rc == 0 ? s.st_size : -1;
 }
@@ -160,7 +160,7 @@ public:
           format = format_t::UNKNOWN;
           return false;
         }
-      } while (result != 0 );
+      } while (result != 0);
     } else {
       LOG_WARN("Corrupt elevation data of unknown type");
       format = format_t::UNKNOWN;
@@ -358,7 +358,7 @@ tile_data cache_t::source(uint16_t index) {
 
   // we have it raw or we don't
   if (item.get_format() == format_t::RAW) {
-      return {this, index, false, (const int16_t*)item.get_data()};
+    return {this, index, false, (const int16_t*)item.get_data()};
   }
 
   // we were able to load it but the format wasn't RAW, which only leaves compressed formats
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -376,7 +376,7 @@ tile_data cache_t::source(uint16_t index) {
   }
 
   // item in cache is already unpacked
-  auto unpacked = item.get_unpacked();
+  const char* unpacked = item.get_unpacked();
   if (unpacked) {
     auto rv = tile_data(this, index, true, (const int16_t*)unpacked);
     mutex.unlock();
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -153,15 +153,19 @@ public:
       size_t src_size = data.size();
       size_t dest_size = HGT_BYTES;
       size_t result;
+
       do {
         result = LZ4F_decompress(decode, const_cast<char*>(this->unpacked), &dest_size, data.get(),
                                  &src_size, &options);
         if (LZ4F_isError(result)) {
+          LZ4F_freeDecompressionContext(decode);
           LOG_WARN("Corrupt lz4 elevation data");
           format = format_t::UNKNOWN;
           return false;
         }
       } while (result != 0);
+
+      LZ4F_freeDecompressionContext(decode);
     } else {
       LOG_WARN("Corrupt elevation data of unknown type");
       format = format_t::UNKNOWN;
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -15,6 +15,7 @@
 
 #include <boost/optional.hpp>
 #include <sys/stat.h>
+#include <lz4frame.h>
 
 #include "baldr/compression_utils.h"
 #include "filesystem.h"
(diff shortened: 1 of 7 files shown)


kevinkreiser:
We'll want to add some unit tests just to exorcise the code paths, you can find some here for gzip, a similar approach can be used to check that lz4 is working properly: https://github.com/valhalla/valhalla/blob/master/test/sample.cc

src/skadi/sample.cc
```diff
@@ -71,14 +72,14 @@ uint64_t file_size(const std::string& file_name) {
...
 
-enum class format_t { UNKNOWN = 0, RAW = 1, GZIP = 2 };
+enum class format_t { UNKNOWN = 0, RAW = 1, GZIP = 2, LZ4 = 3 };
 
 class cache_item_t {
 private:
   format_t format;
   valhalla::midgard::mem_map<char> data;
   int usages;
-  const char* unpacked;
```
kevinkreiser, line 81: this is unfortunate but i assume its an artifact of `LZ4F_decompress` taking a non const parameter

kevinkreiser: the good news is you dont need all of these removals of `const` because when we create the bytes where we unpack the data, we create it as non-const. so we can const cast it back before you make that call. ill make a suggestion below and then we can remove a lot of the changes here

src/skadi/sample.cc
```diff
@@ -110,47 +111,83 @@ class cache_item_t {
...
+      LZ4F_decompressionContext_t decode;
+      LZ4F_decompressOptions_t options;
+      LZ4F_createDecompressionContext(&decode, LZ4F_VERSION);
+
+      // Take these two values locally, since LZ4F_decompress expects pointers...
+      size_t src_size = data.size();
+      size_t dest_size = HGT_BYTES;
+      size_t result;
+      do {
+        result = LZ4F_decompress(decode, this->unpacked, &dest_size, data.get(), &src_size, &options);
```
kevinkreiser, line 157: we can revert all the changes you made to remove `const`ness from this and just cast the `const` away at the time when we need to do that. this is ok because when we create `unpacked` we do so without `const`.

```suggestion
        result = LZ4F_decompress(decode, const_cast<char *>(this->unpacked), &dest_size, data.get(), &src_size, &options);
```

src/skadi/sample.cc
```diff
@@ -353,6 +390,7 @@ tile_data cache_t::source(uint16_t index) {
       }
     }
   }
+
   if (!unpacked) {
     unpacked = (char*)malloc(HGT_BYTES);
```
kevinkreiser, line 395: ^^ this is why we can use `const_cast`, because here `unpacked` is not actually `const` so casting it away will not be undefined behavior

kevinkreiser:
@lseelenbinder this looks generally good, i think you can revert the changes to constness and add a unit test and we can get it reviewed. thanks!

src/skadi/sample.cc
lseelenbinder: Ah, great! I cleaned it all up as suggested.

lseelenbinder:
I'm having difficulty tracking down where the test data files are for elevation. I found the test, but `test/data/(sample|samplegz)` don't exist on my branch so the current tests fail, as well? Is there some automated process for generating them, or should I download these from somewhere?

kevinkreiser:
@lseelenbinder its kind of tricky but we create this header file called `test/pixels.h` which has the binary data of a real tile (though not all of it) as an array of pixel values. we include this header with the vector and then in the test we turn it into elevation tiles: https://github.com/valhalla/valhalla/blob/master/test/sample.cc#L30-L70 then the actual tests check those.

so you'll want to lz4 compress that data in that same function and write it to another location. then you can add a unit test that checks that it loads ok like the other tests do: https://github.com/valhalla/valhalla/blob/master/test/sample.cc#L102-L104

lseelenbinder:
Hey @kevinkreiser, I think this is ready for a final review now. Thanks for your help finishing it up!

src/skadi/sample.cc
```diff
@@ -122,35 +123,76 @@ class cache_item_t {
...
+      // we have to unzip it
+      if (!baldr::inflate(src_func, dst_func)) {
+        LOG_WARN("Corrupt gzip elevation data");
+        format = format_t::UNKNOWN;
+        return false;
+      }
+    } else if (format == format_t::LZ4) {
+      LZ4F_decompressionContext_t decode;
+      LZ4F_decompressOptions_t options;
+      LZ4F_createDecompressionContext(&decode, LZ4F_VERSION);
```
kevinkreiser, line 150: can you refresh my memory, does this also support LZ4_HC or other variants? i vaguely remember there were 3 formats, one was roughly normal but not fixed endian, another was fixed endian (probably network order) and one was "high compression". does this support all of them automatically or just one of them, and if the latter, which one? it was 3 years ago when i implemented it and i think i ended up going with the first one becuase its what the command line tool spits out maybe the others were containerless and more for streaming situations? sorry i just cant remember

lseelenbinder: No worries, @kevinkreiser.

I just went with whatever the command line tool spit out. There's other formats that seem to be more for streaming, but it seemed the most logical target was the command line results. I made a few mistakes on the decompression side using `lz4.h` functions instead of `lz4_frame.h` functions, so it appears at least the `lz4` command line outputs a framed format?

Should we just state we support files compressed with `lz4` command line in the release notes? The command line doesn't seem to offer any options for this (https://linuxcommandlibrary.com/man/lz4).

kevinkreiser: nope thats all good. i just wanted to confirm my recollection. ill review this tonight hopefully when i have a free moment. thanks for the efforts!

kevinkreiser:
this looks good to me but you need to resolve the conflicts

kevinkreiser (approved):

src/skadi/sample.cc
```diff
@@ -334,7 +376,7 @@ tile_data cache_t::source(uint16_t index) {
   }
 
   // item in cache is already unpacked
-  const char* unpacked = item.get_unpacked();
```
kevinkreiser, line 337: please revert this change, or use `const auto *`. auto is great and all but when we use it to hide qualifiers it makes everyones lives harder

lseelenbinder: Whoops, that was an oversight. Good catch!

lseelenbinder:
Thanks again @kevinkreiser. Assuming tests pass, this should be good to go!

kevinkreiser (dismissed):

kevinkreiser (approved):

Issue #3387, opened 2021-11-04
Import bike share station information


xlqian:
Hi, 

Following our contribution to BSS, we'd like to enrich this feature a little bit more. 
In the current version, a bike share station node is imported with minimum information of it, that is to say, only its coordinates are imported. 
In this example, the bss node contains actually more information:
https://www.openstreetmap.org/node/1248493028#map=19/48.11798/-1.68641

Our goal is to display that information in the final API but we have no idea how to import that information and display it.
  
We need your enlightening. :)

thank you 

team navitia

kevinkreiser:
no problem. yes at the moment we dont have an easy way to tie unstructured information to nodes in the graph however we do have a way of storing unstructured information to edges in the graph, and the good news is that the BSS connection edges would be a good candidate there. I would suggest that we do what we do for things like: pronounciation, bridge names and tunnel names, that is we use the "tag value" support in `edgeinfo`. with the place for final storage out of the way the next thing we need to focus on is how to get that information to the point in the code where we create the connection edges. see all the rest of the code assumes the edge exists up front when we make the edgeinfos but in this case we only create the connection edges later. we'd need to add a bit of code to act as if these edges exist before we do the connecting. then the connecting will be a matter of finding the prexisting edgeinfo record and using that when creating the connection directededge. let me know if any of this is making sense :smile:

PR #3411, opened 2021-11-16
Import bike share station information

AzHicham:
Hello :) 
Following [this issue](https://github.com/valhalla/valhalla/issues/3387) opened by my colleague @xlqian, here is the PR that allow retrieving bike share station information in valhalla from an osm.
This is still a work in progress, but reviews are very welcomed. 

## Tasklist

 - [x] Add tests

### DIFF
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -120,7 +120,6 @@ message BikeShareStationInfo {
   optional string operator = 5;
   optional float rent_cost = 6;
   optional float return_cost = 7;
-  optional string osm_node_id = 8;
 }
 
 message TransitPlatformInfo {
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -115,7 +115,7 @@ message TransitStationInfo {
 message BikeShareStationInfo {
   optional string name = 1;
   optional string ref = 2;
-  optional string capacity = 3;
+  optional uint32 capacity = 3;
   optional string network = 4;
   optional string operator = 5;
   optional float rent_cost = 6;
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -112,6 +112,16 @@ message TransitStationInfo {
   optional LatLng ll = 3;                // Latitude, longitude of the station
 }
 
+message BikeShareStationInfo {
+  optional string name = 1;
+  optional string ref = 2;
+  optional string capacity = 3;
+  optional string network = 4;
+  optional string operator = 5;
+  optional float rent_cost = 6;
+  optional float return_cost = 7;
+}
+
 message TransitPlatformInfo {
   enum Type {
     kStop = 0;
@@ -175,6 +185,12 @@ message TaggedValue {
     // see https://github.com/valhalla/valhalla/issues/3262
     kTunnel = 49; // static_cast<uint8_t>('1')
     kBridge = 50; // static_cast<uint8_t>('2')
+    kBssName = 51;
+    kBssRef = 52;
+    kBssNetwork = 53;
+    kBssCapacity = 54;
+    kBssSource = 55;
+    kBssOperator = 56;
   }
   optional string value = 1;   // The actual tagged name value, examples: Ted Williams Tunnel
   optional Type type = 2;      // The type of tagged name (tunnel or bridge)
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -120,6 +120,7 @@ message BikeShareStationInfo {
   optional string operator = 5;
   optional float rent_cost = 6;
   optional float return_cost = 7;
+  optional string osm_node_id = 8;
 }
 
 message TransitPlatformInfo {
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -181,16 +181,11 @@ enum RoadClass {
 message TaggedValue {
   enum Type {
     kLayer = 1;
+    kBssInfo = 3;
     // we used to have bug when we encoded 1 and 2 as their ASCII codes, but not actual 1 and 2 values
     // see https://github.com/valhalla/valhalla/issues/3262
     kTunnel = 49; // static_cast<uint8_t>('1')
     kBridge = 50; // static_cast<uint8_t>('2')
-    kBssName = 51;
-    kBssRef = 52;
-    kBssNetwork = 53;
-    kBssCapacity = 54;
-    kBssSource = 55;
-    kBssOperator = 56;
   }
   optional string value = 1;   // The actual tagged name value, examples: Ted Williams Tunnel
   optional Type type = 2;      // The type of tagged name (tunnel or bridge)
--- a/src/mjolnir/pbfgraphparser.cc
+++ b/src/mjolnir/pbfgraphparser.cc
@@ -1444,6 +1444,8 @@ public:
           } else if (key_value.first == "operator") {
             bss_info.set_operator_(key_value.second);
           }
+
+          bss_info.set_osm_node_id(std::to_string(osmid));
         }
 
         std::string buffer;
--- a/src/mjolnir/pbfgraphparser.cc
+++ b/src/mjolnir/pbfgraphparser.cc
@@ -26,6 +26,7 @@
 #include "midgard/sequence.h"
 #include "midgard/tiles.h"
 #include "mjolnir/timeparsing.h"
+#include <valhalla/proto/directions.pb.h>
 
 using namespace valhalla::midgard;
 using namespace valhalla::baldr;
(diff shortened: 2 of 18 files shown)


proto/tripcommon.proto
```diff
@@ -175,6 +185,12 @@ message TaggedValue {
     // see https://github.com/valhalla/valhalla/issues/3262
     kTunnel = 49; // static_cast<uint8_t>('1')
     kBridge = 50; // static_cast<uint8_t>('2')
+    kBssName = 51;
+    kBssRef = 52;
+    kBssNetwork = 53;
+    kBssCapacity = 54;
+    kBssSource = 55;
+    kBssOperator = 56;
```
kevinkreiser, line 193: what if we combined these into a single record? one interesting idea would be to construct the protobuf object above `BikeShareStationInfo` serialize it to bytes and then b64 encode the bytes and store that as the value. this would make our lives really easy when getting the info back out of the tiles.

another question i have is why dont these fields match the protobuf above?

xlqian:
Hi, many thanks to @AzHicham for this great job. We managed to import some additional bike share station information(which meet our demands so far) and somehow integrated this information into the manoeuvres. 

Here are some more questions regarding which we'd love to hear from you:

1. we imported, as yet, just a part of bike share station information, there are still a bunch of information not imported, do you prefer to let it be so, or an exhaustive importing of all tagged values?
    
2. What do you think about displaying that bss information in the API? If yes, in which section do you want to display BSS info? and can you tell us which part of the codes we can start with?

Thank you

valhalla/baldr/graphconstants.h
```diff
@@ -364,6 +364,12 @@ enum class TaggedValue : uint8_t { // must start at 1 due to nulls
   // see https://github.com/valhalla/valhalla/issues/3262
   kTunnel = static_cast<uint8_t>('1'),
   kBridge = static_cast<uint8_t>('2'),
```
kevinkreiser, line 366: these were a bug and thus casted from char to int. you should start counting from int=2

proto/tripcommon.proto
kevinkreiser: also this s hould use a value of `2`. tunnel and bridge were mistakenly using chars when first implemented

kevinkreiser:
so i only had one major suggestion here about how to store the data, @AzHicham and @xlqian what do you guys think? maybe its a crazy idea, but it might be a good pattern to store more complex things in the future as well. no worries if it seems like overkill.

maybe @gknisely or @dnesbitt61 have opinions as well?

kevinkreiser:
> do you prefer to let it be so, or an exhaustive importing of all tagged values

i think one of the cool things about the suggestion to use pbf to encapsulate the value of the tagged value is that we can support optional pieces being there and we can add new fields to it later without breaking compatibility. so if we go with the pbf approach i suggested, then we dont need to be ehaustive until later when we need more

> What do you think about displaying that bss information in the API

i think it makes sense to display it in the API yes. i havent thought about how to do that yet. actually i think id prefer to hear @dgearhart 's opinion on that if he has a minute

valhalla/baldr/graphconstants.h
AzHicham: thank you @kevinkreiser

AzHicham:
> > do you prefer to let it be so, or an exhaustive importing of all tagged values
> 
> i think one of the cool things about the suggestion to use pbf to encapsulate the value of the tagged value is that we can support optional pieces being there and we can add new fields to it later without breaking compatibility. so if we go with the pbf approach i suggested, then we dont need to be ehaustive until later when we need more
> 
> > What do you think about displaying that bss information in the API
> 
> i think it makes sense to display it in the API yes. i havent thought about how to do that yet. actually i think id prefer to hear @dgearhart 's opinion on that if he has a minute

Hello @kevinkreiser I followed your advice i my latest commit, so now we store Bss informations as a serialized protobuf message

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1415,17 +1416,44 @@ struct graph_callback : public OSMPBF::Callback {
...
+          } else if (key_value.first == "capacity") {
+            bss_info.set_capacity(key_value.second);
+          } else if (key_value.first == "operator") {
+            bss_info.set_operator_(key_value.second);
+          }
+        }
+
+        std::string buffer;
+        bss_info.SerializeToString(&buffer);
+        n.set_bss_info_index(osmdata_.node_names.index(buffer));
```
xlqian, line 1451: :+1:

kevinkreiser:
@AzHicham @xlqian im glad you liked my suggestion, i think its awesome that the implementation is so clean! @gknisely @dgearhart we should have done this for the phoneme/language stuff. though it might have used more space the implementation is so clean!

kevinkreiser:
@AzHicham @xlqian do you guys think we should get this reviewed and merged and then make the API response changes in a separate PR after we discuss where that would go? Its nice to have such a complete little chunk of work here that is ready to go.

xlqian:
> @AzHicham @xlqian do you guys think we should get this reviewed and merged and then make the API response changes in a separate PR after we discuss where that would go? Its nice to have such a complete little chunk of work here that is ready to go.

That's a very good idea for us since we have got everything that we need in this PR.  We can contribute later on to make more changes in API response.

proto/tripcommon.proto
```diff
@@ -112,6 +112,17 @@ message TransitStationInfo {
...
 
+message BikeShareStationInfo {
+  optional string name = 1;
+  optional string ref = 2;
+  optional string capacity = 3;
+  optional string network = 4;
+  optional string operator = 5;
+  optional float rent_cost = 6;
+  optional float return_cost = 7;
+  optional string osm_node_id = 8;
```
kevinkreiser, line 123: A couple things about this. This should be and unsigned 64bit value, definitely not a string but also we have another in progress pr that adds support for this which is already quite close to done. Would you be OK to remove this with the knowledge that we will soon have support for osm node ids from this PR #3245 by @sleepingwell?

AzHicham: Perfect I'll remove it :)

proto/tripcommon.proto
```diff
@@ -112,6 +112,16 @@ message TransitStationInfo {
   optional LatLng ll = 3;                // Latitude, longitude of the station
 }
 
+message BikeShareStationInfo {
+  optional string name = 1;
+  optional string ref = 2;
+  optional string capacity = 3;
```
kevinkreiser, line 118: i believe this should also be a number. 

```suggestion
  optional uint32 capacity = 3;
```

then below you'll want to use `strtoul` and if it cant parse the string then dont set the capacity

AzHicham: Fixed ! it's better this way

src/mjolnir/pbfgraphparser.cc
```diff
@@ -26,6 +26,7 @@
 #include "midgard/sequence.h"
 #include "midgard/tiles.h"
 #include "mjolnir/timeparsing.h"
+#include <valhalla/proto/directions.pb.h>
```
kevinkreiser, line 29:
```suggestion
#include"proto/directions.pb.h"
```

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1415,17 +1416,47 @@ struct graph_callback : public OSMPBF::Callback {
       } else {
         results = empty_node_results_;
       }
+      bool is_bss = false;
```
kevinkreiser, line 1419: i think we can simplify this whole thing by just bailing out early if we dont have the bike rental value:

```c++
    // Handle bike share stations separately
    boost::optional<Tags> results = boost::none;
    if (bss_nodes_) {
      // Get tags - do't bother with Lua callout if the taglist is empty
      if (tags.size() > 0) {
        results = lua_.Transform(OSMType::kNode, osmid, tags);
      } else {
        results = empty_node_results_;
      }
      
      // bail if there is nothing bike related
      Tags::const_iterator found;
      if (!results.has_value() || (found = results->find("amenity")) == results->end() || found->second != "bicycle_rental") {
        return;
      }

      // Create a new node and set its attributes
      OSMNode n{osmid};
      n.set_latlng(lng, lat);
      n.set_type(NodeType::kBikeShare);
      valhalla::BikeShareStationInfo bss_info;

      for (auto& key_value : *results) {
        if (key_value.first == "name") {
          bss_info.set_name(key_value.second);
        } else if (key_value.first == "network") {
          bss_info.set_network(key_value.second);
        } else if (key_value.first == "ref") {
          bss_info.set_ref(key_value.second);
        } else if (key_value.first == "capacity") {
          auto capacity = std::strtoul(key_value.second.c_str(), nullptr, 10);
          if (capacity > 0) {
            bss_info.set_capacity(capacity);
          }
        } else if (key_value.first == "operator") {
          bss_info.set_operator_(key_value.second);
        }
      }

      std::string buffer;
      bss_info.SerializeToString(&buffer);
      n.set_bss_info_index(osmdata_.node_names.index(buffer));
      ++osmdata_.node_name_count;

      bss_nodes_->push_back(n);
      return; // we are done because we are just handling bss this parsing pass
    }
```

AzHicham: Done ! :)

kevinkreiser (approved):
this looks good to me, @gknisely @dgearhart do you have any opinions here?

src/odin/maneuver.cc
```diff
@@ -1119,6 +1119,14 @@ void Maneuver::set_long_street_name(bool has_long_street_name) {
   has_long_street_name_ = has_long_street_name;
 }
 
+BikeShareStationInfo Maneuver::bss_info() const {
```
dgearhart, line 1122: Did you want to return by value or const ref?

dgearhart (approved):
Just a comment about return type - otherwise looks good to me
@gknisely what say you?

dgearhart:
> i think it makes sense to display it in the API yes. i havent thought about how to do that yet. actually i think id prefer to hear @dgearhart 's opinion on that if he has a minute

@xlqian can you give more details - typical use case for user?

src/odin/maneuver.cc
AzHicham: My bad I’ll correct it !

xlqian:
> > i think it makes sense to display it in the API yes. i havent thought about how to do that yet. actually i think id prefer to hear @dgearhart 's opinion on that if he has a minute
> 
> @xlqian can you give more details - typical use case for user?

In our software, we use some attributes from bike-share stations(e.g. : "ref", "network", "operator"  etc) to request some tiers services in order to get "real-time" information of BSS (e.g. number of available rental bikes, number of available slots, if the BSS is closed? etc...). In the future, we'd love to integrate all this real-time information into the path finding algorithm(a real-time BSS a*).  

So far, we have got everything we need to carry on our project, since we don't use the API directly(our software is compiled with Valhalla's libraries directly and we just use the protobuf to obtain what we want). To be honest, API is not out major concern but we just think that it would be a pity if we don't display that information while it is existant.

dgearhart (approved):
Thanks for the update

Issue #2574, opened 2020-09-02
Bus costing does not route using new 14th St NY busways and Market St SF busways


disambiguator:
When routing on Valhalla using the `costing: bus` option, we do not route on the new major busways in SF and NY.

When investigating on OSM, I found that the 14th St ways in NY have `motor_vehicle: no` tags without `bus: yes`. (For example: https://www.openstreetmap.org/way/196117073#map=17/40.73627/-73.99596)

On Market St, I found `bus: designated` tags https://www.openstreetmap.org/way/225847526

Example repro case: Routing from 1356 Market St to 691 Market St takes you along Mission St instead of the Market St busway 
<img width="856" alt="image" src="https://user-images.githubusercontent.com/2306794/92024570-32c2a500-ed13-11ea-81eb-18f720f38d66.png">

Routing from 527 6th Ave, NY to 206 E 14th St, NY avoids 14th St as well
<img width="625" alt="image" src="https://user-images.githubusercontent.com/2306794/92025255-1b37ec00-ed14-11ea-8100-7e2b167ec575.png">

anthonyNkuinjeu:
Hey. Just to say that I encounter the same problem on another road for another country (France).

**This is the road I want to take. It is exclusive to buses and OSM does have the tag "bus" set to "yes"**

![road_i_want](https://user-images.githubusercontent.com/681293/95090455-992d4f80-0725-11eb-98e1-50f118448853.png)

**This is the response from Valhalla. Making a big turn to get out of the bus zone. I have no idea why...**

![road_i_get_valhalla](https://user-images.githubusercontent.com/681293/95090487-a1858a80-0725-11eb-96e1-8c8ea8d37817.png)

PR #3413, opened 2021-11-17
Add graph support for highway=busway

landonreed:
# Issue

This PR targets #2574. While it may not handle all of the cases described in that issue, it aims to expand the routable bus-only ways for bus costing to include the OSM tag `highway=busway`, which according to [OSM docs](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbusway) should imply `bus=designated`. This is specifically designed to support [some ways in the Brisbane area](https://overpass-turbo.eu/s/1cqh) that are correctly identified as `highway=busway`.

I'm not sure if this warrants a test, but am open to adding one if the maintainers think that would be helpful!

## Tasklist

 - N/A Add tests
 - N/A Update the docs with any new request parameters or changes to behavior described

## Requirements / Relations

 N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
    * FIXED: Some refactoring that was proposed for the PR 3379 [3381](https://github.com/valhalla/valhalla/pull/3381)
    * FIXED: Avoid calling out "keep left/right" when passing an exit [3349](https://github.com/valhalla/valhalla/pull/3349)
    * FIXED: Fix iterator decrement beyond begin() in GeoPoint::HeadingAtEndOfPolyline() method [#3393](https://github.com/valhalla/valhalla/pull/3393)
+   * FIXED: Parse "highway=busway" OSM tag: https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbusway [#3413](https://github.com/valhalla/valhalla/pull/3413)
 
 * **Enhancement**
    * CHANGED: Pronunciation for names and destinations [#3132](https://github.com/valhalla/valhalla/pull/3132)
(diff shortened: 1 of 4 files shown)


kevinkreiser:
@gknisely maybe this is a good opportunity to throw in a quick gurka test for `bus`?

gknisely:
> @gknisely maybe this is a good opportunity to throw in a quick gurka test for `bus`?

Sure as soon as I get a minute.  Unless @landonreed wants to give it a try

landonreed:
Hey all, I don't have a ton of experience in the valhalla code base, but would be happy to give it a look if that would be helpful!

kevinkreiser:
We dont have any bus specific tests yet but a good place to add some might be in here: https://github.com/valhalla/valhalla/blob/master/test/gurka/test_access_psv.cc where we could confirm `bus` can but `auto` cannot take these ways. Or over in the general routing test: https://github.com/valhalla/valhalla/blob/master/test/gurka/test_route.cc

@landonreed take a look at the general lay out of the gurka tests and give this a read: https://github.com/valhalla/valhalla/tree/master/test/gurka#valhalla-gurka-tests to understand the general philosophy.

landonreed:
@kevinkreiser, thanks for the pointers! Does https://github.com/valhalla/valhalla/pull/3413/commits/a54ac387be25aeda467e0cb2584d5029225b71d5 look to be on the right track? Happy to separate that out into its own `TEST` if that would be better.

kevinkreiser:
@landonreed that test change looks great to me. Now we just have to fix the rest of the build and resolve merge conflicts

kevinkreiser:
@landonreed i modified your testing to make the test cases a bit stronger and also fix the cases you were aiming at. we'll see what the build says but locally everything worked for me after these changes :crossed_fingers:

kevinkreiser (approved):

landonreed:
That looks great! Thanks for all of the help moving this along, @kevinkreiser!

landonreed:
Hi @kevinkreiser, thanks once again for helping shepherding this along!

I wanted to inquire as to when you expect to cut the next release of valhalla? We have some customers that are interested in some of these changes, but I think we're blocked until the new version with this PR ships.

We're happy to help with anything that might grease the wheels!

kevinkreiser:
@landonreed im not exactly sure what your deployment process is but the only thing our release process does:

1. tagging the repo
2. pushing a tagged docker build to docker hub

You are free to pull and retag our `run-latest` docker image anytime, which has the tip of master in it, if you want to use this for your deployment. In other words nothing about our release cadence should hold you back from deploying this code. Does that make sense?

landonreed:
@kevinkreiser, thanks for the info! That's very helpful. I'll check on how our deployment is handled. We use a third-party hosted service, so there's a little bit of grey area in how they handle releasing new code. An official tagged release might be helpful, but if that's become less common of a practice for the project (or has overhead) maybe we can recommend that they use `run-latest`. 

I'll follow up here if a tagged release seems like our only path forward.

kevinkreiser:
It's just not always a perfect time for a tagged release to be honest. At the moment we have partial support for indoor routing such that you can rout on them but the routes you get don't have maneuvers fleshed out yet. That pr is in progress and once we merge it we can probably tag

PR #3416, opened 2021-11-18
Add string for kPedestrianCrossing.

### LINKED ISSUE #3415: Use to_string doesn't include kPedestrianCrossing
When looking at the Use of edges along a route I noted that some are printed as null when using to_string(Use u). I tracked this down to a new Use - kPedestrianCrossing that was added to the enum but not added to the to_string method.

dnesbitt61:
Prior to this was causing a null output in to_string(Use).

# Issue

fixes #3415 

## Tasklist

 - [ ] Add tests


dgearhart (approved):
:shipit:

PR #3417, opened 2021-11-19
better recosting

CuriOusQuOkka:
# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


scripts/valhalla_build_config
```diff
@@ -98,7 +98,8 @@ config = {
     },
     'max_reserved_labels_count': 1000000,
     'clear_reserved_memory': False,
-    'extended_search': False
+    'extended_search': False,
+    'invariant_recosting': True
```
kevinkreiser, line 102: i think this should be decided at runtime, not via config. simply add a new field to the request structure that tells valhalla to use bidirectional a* even if the date_time type isnt invariant and  you're basically done (other than the actual work of doing the linear combination in GetSpeed and passing the seconds_from_now in the costing function)

src/sif/recost.cc
```diff
@@ -122,7 +122,8 @@ void recost_forward(baldr::GraphReader& reader,
     // update the cost to the end of this edge
     uint8_t flow_sources;
     cost += transition_cost +
-            costing.EdgeCost(edge, tile, offset_time.second_of_week, flow_sources) * edge_pct;
+            costing.EdgeCost(edge, tile, offset_time.second_of_week, flow_sources, seconds_offset) *
```
kevinkreiser, line 125:
```suggestion
            costing.EdgeCost(edge, tile, offset_time.second_of_week, flow_sources, offset_time.seconds_from_now) *
```

scripts/valhalla_build_config
kevinkreiser: sorry maybe a little more info, you'd need to account for this in thor/route_action where we select the algorithm to use.

CuriOusQuOkka:
it was a draft branch, see original one here https://github.com/valhalla/valhalla/pull/3398

Issue #3421, opened 2021-11-22
Python bindings build error when Valhalla is used as a Git submodule


szymone:
Build fails when using Valhalla as a Git submodule and adding Valhalla with CMake add_subdirectory(...). This is caused by the following line:

https://github.com/valhalla/valhalla/blob/61c748d49d47845cd57dbbb8595ab4e1d422f05d/src/bindings/python/CMakeLists.txt#L1

szymone:
Here is PR for this issue #3422

PR #3422, opened 2021-11-22
Build fix

szymone:
# Issue

This PR targests #3421 issue.


nilsnolde (approved):
thx! imo no need for changelog.

if you have a cmake recipe for integrating & building valhalla as git submodule, we'd appreciate a share:)

kevinkreiser (approved):

kevinkreiser:
As far as the recipe goes it should be mostly just add_subdirectory

szymone:
@nilsnolde In the simplest way, you can just add `add_subdirectory(${PROJECT_SOURCE_DIR}/extern/valhalla ${CMAKE_BINARY_DIR}/extern/valhalla)` in your main CMake file.

PR #3425, opened 2021-11-23
some cleanup for old gcc & boost

nilsnolde:
some more cleanup. took me much too long to figure out a missing `include` was the problem;)

on a side note, the function https://github.com/valhalla/valhalla/blob/8739361bc463277de489395b84d9410e73f75bce/src/odin/markup_formatter.cc#L21 is again defined exactly the same here:
https://github.com/valhalla/valhalla/blob/8739361bc463277de489395b84d9410e73f75bce/src/odin/enhancedtrippath.cc#L30

couldn't that go into proto_conversions.cc @gknisely ? or maybe you avoided that for a particular reason?


src/mjolnir/pbfgraphparser.cc
```diff
@@ -1419,7 +1419,7 @@ struct graph_callback : public OSMPBF::Callback {
 
       // bail if there is nothing bike related
       Tags::const_iterator found;
-      if (!results.has_value() || (found = results->find("amenity")) == results->end() ||
+      if (!results.is_initialized() || (found = results->find("amenity")) == results->end() ||
```
nilsnolde, line 1422: has_value is >= 1.68

nilsnolde:
just had a look again: the `Pronunciation_Alphabet_Name` function wasn't used in `enhancedtrippath.cc` so I just removed it and left the rest as it was. does that look good to you @gknisely @kevinkreiser ?

nilsnolde:
oops, turns out that function was used in a trace logging protected block. so I reverted to back to the original idea and put the pbf lookup function to `proto_conversions.h`, should be fine now

src/mjolnir/pbfgraphparser.cc
kevinkreiser: i think you can just use the boolean cast operator

nilsnolde:
I think that's it now @kevinkreiser :)

src/odin/enhancedtrippath.cc
```diff
@@ -26,18 +29,7 @@ constexpr int kIsStraightestBuffer = 10;                   // Buffer between str
...
-      values{{valhalla::Pronunciation_Alphabet::Pronunciation_Alphabet_kIpa, "kIpa"},
-             {valhalla::Pronunciation_Alphabet::Pronunciation_Alphabet_kXKatakana, "kXKatakana"},
-             {valhalla::Pronunciation_Alphabet::Pronunciation_Alphabet_kXJeita, "kXJeita"},
-             {valhalla::Pronunciation_Alphabet::Pronunciation_Alphabet_kNtSampa, "kNtSampa"}};
-  auto f = values.find(alphabet);
-  if (f == values.cend())
-    throw std::runtime_error("Missing value in protobuf Pronunciation_Alphabet enum to string");
-  return f->second;
-}
+} // namespace
```
kevinkreiser, line 32: this looks like a mistake, you took some things out of the anonymous namespace

src/proto_conversions.cc
```diff
@@ -305,6 +305,18 @@ const std::string& FilterAction_Enum_Name(const FilterAction action) {
   return i == actions.cend() ? empty : i->second;
 }
 
+const std::string& PronunciationAlphabetToString(valhalla::Pronunciation_Alphabet alphabet) {
```
kevinkreiser, line 308: we shuld follow the same naming conventions as the others with _Name and _Parse

kevinkreiser:
looks good to me other than tiny nitpicks!

src/odin/enhancedtrippath.cc
nilsnolde: it's actually on purpose, it's moved to proto_conversions.h as that exact function was created twice in anonymous namespaces (see markup_formatter.cc below)

kevinkreiser: by moving this up to here to pull `RoadClass_Name` `TripLeg_Traversability_Name` `TripLeg_Use_Name` `TripLeg_TravelMode_Name`  `TripLeg_VehicleType_Name` and many more out of the anonymous namespace. totally cool to move that function over to there but the other functions below here should stay in the anonymous namespace

kevinkreiser (approved):

dgearhart:
> couldn't that go into proto_conversions.cc @gknisely ? or maybe you avoided that for a particular reason?

@nilsnolde they were not returning the same values - one was for trace debugging and the other was the returned value for the interface. For now it is ok to return the same - just so it is the interface one as you selected

src/odin/enhancedtrippath.cc
nilsnolde: ah yeah, I was blind, thanks for fixing @kevinkreiser !

nilsnolde:
apologies @gknisely should've really seen that it's different return values..

nilsnolde:
@dgearhart jeez, I'm really confused.. won't happen again

dgearhart (changes requested):
@nilsnolde let's have the two methods return the values they did originally to avoid any confusion

nilsnolde:
jep sure @dgearhart . sure I just revert that part of the changes entirely?

dgearhart:
> jep sure @dgearhart . should I just revert that part of the changes entirely (i.e. not move the function to proto_conversions.h)?

@nilsnolde sounds good - thanks

kevinkreiser (approved):

PR #3427, opened 2021-11-24
Use std::multimap for admin and timezone polygon results

dnesbitt61:
Admin polygon lookup previously was using std::unordered_multimap for the polygon set returned when querying the admin database. This results in a polygon set that is unordered. It was found that often the country polygon was tested before state polygons in GetMultiPolyId. Since we always want to find a state polygon first (if one exists) this unordered list of results is sub-optimal.

this PR changes to using std::multimap for the resulting polygon list. This can now be checked in the proper order and if a state polygon is found as the first case where the lat,lng is within the admin polygon we can avoid testing the (usually) larger country polygon.

I tested this on North America OSM data. The build stage was done with 8 threads (set concurrency=8 in the config) with the following performance:

master: 2:57:55
branch: 1:32:36

The total size of tiles matches between master and the branch (which I take to mean it is likely the admin results are consistent).

# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/mjolnir/admin.cc
+++ b/src/mjolnir/admin.cc
@@ -47,6 +47,9 @@ uint32_t GetMultiPolyId(const std::multimap<uint32_t, multi_polygon_type>& polys
   for (const auto& poly : polys) {
     if (boost::geometry::covered_by(p, poly.second)) {
       const auto& admin = graphtile.admins_builder(poly.first);
+      if (!admin.state_offset())
+        index = poly.first;
+      else
         return poly.first;
     }
   }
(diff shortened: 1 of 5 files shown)


gknisely (approved):

src/mjolnir/admin.cc
```diff
@@ -10,6 +10,8 @@
 namespace valhalla {
 namespace mjolnir {
 
+using polygon_type = boost::geometry::model::polygon<point_type>;
```
kevinkreiser, line 13: this looks unused

kevinkreiser (approved):

src/mjolnir/admin.cc
dnesbitt61: Thanks. I added that for an early change where I tested if the tile's bounding box was totally inside the 1 state query resulting multi-polygon. If so I was going to just return that one admin and then set tile_withing_one_admin within graphbuilder.cc. This would have removed the need to do point in poly for all other nodes within the tile. Unfortunately this led to a mismatch in tile sizes - so some admin information must have changed. I will leave this for a later (possible) update.

PR #3428, opened 2021-11-24
Chinesepostman

ismailsunni:
# Issue

This Pull Request will add a Chinese Postman Problem solver support for Valhalla. It uses the algorithm explained [here](https://algorithms.discrete.ma.tum.de/graph-algorithms/directed-chinese-postman/index_en.html) to solve the problem. This Chinese Postman service handles a not strongly connected graph or unbalanced graph that is required by the algorithm. It will try to find a route outside the area of interest to make the graph representation balanced and strongly connected.

### Input

The input is very similar to the normal route algorithm. The only difference is the introducing `chinese_postman_polygon` to indicate which area that its route must be visited and the number of `locations` must be 2 (origin and destination only).

### Output

The output is in the same format as in the routing including the maneuver.  Below is an example of the output:
Selection_352

### Known Issues

1. For some cases, it failed to find the shortest route between two locations (to make the graph balanced or strongly connected). Thus, it will connect those two locations with a straight line. Based on our testing, it happens in ~10% of the random test cases. When we compute the route using Valhalla directly, it can find the route. An example of this issue is below:
   Selection_349
2. Unvisited road inside the Chinese polygon
   In some cases, it failed to cover all the roads inside the Chinese polygon. Until the current investigation, it failed to build a balanced graph. See the image below for example:
   image

Both issues were found when we tested on real data. We haven't been able to reproduce the bug with a unit test to fix it.

Suggestion and review are very welcomed :)

## Tasklist

 - [x] Add tests

## Requirements / Relations


docs/api/chinese-postman/api-reference.md
```diff
@@ -0,0 +1,46 @@
+# Chinese Postman service API reference
```
kevinkreiser, line 1: we should slap a big fat (BETA) on this top level lable and then put a note below that says something like: While this API may be fully functional we are currently treating it as beta and may still make breaking changes to the API in the future.

kevinkreiser: would be also good to add the information about the two failure cases that you added to the PR discription into the text here. i think we are totally fine to iterate on this rather than get it perfect on the first go.

src/thor/Hungarian.cc
```diff
@@ -0,0 +1,423 @@
...
+//
+// This is a C++ wrapper with slight modification of a hungarian algorithm implementation by Markus
+// Buehren. The original implementation is a few mex-functions for use in MATLAB, found here:
+// http://www.mathworks.com/matlabcentral/fileexchange/6543-functions-for-the-rectangular-assignment-problem
+//
+// Both this code and the orignal code are published under the BSD license.
+// by Cong Ma, 2016
+//
+
+#include "thor/Hungarian.h"
```
kevinkreiser, line 12: can you rename this to lower case in both the header and the source files. our file naming convention is always lower case and used to be no spaces but is more recently underscores for spaces

src/loki/polygon_search.cc
```diff
@@ -186,8 +201,89 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Options_Ring>&
     LOG_TRACE("Avoided edges GeoJSON: \n" + to_geojson(avoid_edge_ids, reader));
   }
 #endif
-
   return avoid_edge_ids;
 }
+
+// TODO: Probable merged the logic with edges_in_rings above (if possible)
```
kevinkreiser, line 207: i agree i think with a little work we can merge the two into one function with different options depending on who is calling it

src/thor/Hungarian.cc
kevinkreiser: actually no, scratch that, we should be submoduling this into the thirdparty directory and using it from there directly not copy pasting the code into our repo. you can achieve this by doing something like:

    git submodule add https://github.com/mcximing/hungarian-algorithm-cpp third_party/hungarian-algorithm-cpp

you'll need to change CMakeLists.txt to properly reference them as well. I would even suggest that since we only call them from some other source files in thor that we make them private. let me know and i can help you to modify cmake to make this change, but basically in src/thor/CMakeLists.txt you set that thirdparty dir we submoduled as a private include, and then you just add the header and source file to the list of files to compile.

ismailsunni:
Thanks @kevinkreiser 
Yeah, we tried to make it as perfect as possible on the first go but it seems it's better to make a PR since it doesn't break other parts of the library.
I will update the PR addressing the review!

Thanks again :)

src/loki/polygon_search.cc
```diff
@@ -94,12 +100,21 @@ std::string to_geojson(const std::unordered_set<vb::GraphId>& edge_ids, vb::Grap
...
 namespace loki {
 
+// This custon within is created because using the boost::within for edge_info
+// shape will give different result depending on the order of the point.
+bool IsWithin(vb::EdgeInfo edge_info, polygon_type polygon) {
+  // create temporary line from edge_info
+  point_type p1(edge_info.shape()[0].first, edge_info.shape()[0].second);
+  point_type p2(edge_info.shape()[1].first, edge_info.shape()[1].second);
+  linestring_t edge_line{p1, p2};
+  return bg::within(edge_line, polygon);
```
nilsnolde, line 110: this doesn't looks right.. you're basically only testing the very first segment of the whole edge (or the last if it's reverse). I don't think your problems would become any less when you use the already existing (geographic) geometry types, but who knows, maybe that resolves the problem you describe in the comment. 

`shape will give different result depending on the order of the point.`: can you elaborate what "order of the point" means?

nilsnolde: in any case I don't think you should only test the very first segment here which might be the first 50 cm of a whole 2 km edge..

src/loki/polygon_search.cc
nilsnolde: yeah, as far as I can see the only differences are:
- you call `within` instead of `intersects`
- you treat opp edges differently

both make sense, just mold them into arguments to the function somehow

src/loki/worker.cc
```diff
@@ -142,6 +142,19 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
     }
   }
 
+  if (options.has_chinese_polygon()) {
+    const auto chinese_edges =
+        edges_in_ring(options.chinese_polygon(), *reader, costing, max_exclude_polygons_length);
```
nilsnolde, line 147: I'm not sure if piggy backing on `max_exclude_polygons_length` is good.. maybe it is, instead of introducing yet another, almost identical config option? both are very expensive, though chinese postman much more so..

src/thor/chinese_postman_action.cc
```diff
@@ -0,0 +1,484 @@
...
+
+    loc.mutable_ll()->set_lng(l.first);
+    loc.mutable_ll()->set_lat(l.second);
+    source_location_list.Add()->CopyFrom(loc);
+    target_location_list.Add()->CopyFrom(loc);
+  }
+
+  try {
+    auto locations = PathLocation::fromPBF(source_location_list, true);
+    const auto projections = loki::Search(locations, *reader, costing);
```
nilsnolde, line 130: hm not sure it's good to call loki from thor..

nilsnolde (commented):
Thanks for submitting the PR @ismailsunni ! 

I only spent a little while going over the stuff I'm most familiar with.. for the rest it'd help if you could comment more extensively, either in the PR or in-code comments. especially the chinese postman graph and thor's chinese_postman action. those are pretty big and would really benefit from more comments to understand things better

src/loki/polygon_search.cc
```diff
@@ -24,6 +26,10 @@ using line_bg_t = bg::model::linestring<vm::PointLL>;
 using ring_bg_t = std::vector<vm::PointLL>;
 using namespace vb::json;
 
+typedef bg::model::d2::point_xy<double> point_type;
+typedef bg::model::polygon<point_type> polygon_type;
+typedef bg::model::linestring<point_type> linestring_t;
```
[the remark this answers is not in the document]
nilsnolde: just some explanation on the existing boost geom lines:
`BOOST_GEOMETRY_REGISTER_POINT_2D(vm::PointLL, double, bg::cs::geographic<bg::degree>, first, second)`: registers valhalla's PointLL as geographic point type, so now we can do all bg spatial operations on native types instead of creating bg model points.
`BOOST_GEOMETRY_REGISTER_RING(std::vector<vm::PointLL>)` does the same for a ring (mind windedness!). also means we can't use `std::vector<PointLL>` for linestring model, that's why we had to create a bg linestring model a few lines down.

docs/api/chinese-postman/api-reference.md
ismailsunni: Added.

src/loki/worker.cc
ismailsunni: Ah yes, you are right. I forgot to add a new one. Added now as max_chinese_polygon_length

src/loki/polygon_search.cc
ismailsunni: I managed to merge the function, and it seems I also do not need to use a custom `within` function.

Besides those two differences, I also throw a different exception. Since 167 is specific to avoid polygons.

src/loki/polygon_search.cc
ismailsunni: Yep, you are correct. It was my bad unit test (most of the edge only has one segment). Anyway, this part is no longer exists after I merge to edges_in_rings.

src/loki/polygon_search.cc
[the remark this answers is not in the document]
ismailsunni: Code is removed now.

src/loki/chinese_postman_action.cc
```diff
@@ -0,0 +1,87 @@
...
+
+  // Taken from loki route_action
+  // correlate the various locations to the underlying graph
+  std::unordered_map<size_t, size_t> color_counts;
+  try {
+    auto locations = PathLocation::fromPBF(options.locations(), true);
+    const auto projections = loki::Search(locations, *reader, costing);
+    for (size_t i = 0; i < locations.size(); ++i) {
+      const auto& correlated = projections.at(locations[i]);
+      PathLocation::toPBF(correlated, options.mutable_locations(i), *reader);
```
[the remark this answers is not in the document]
ismailsunni: Haha, thanks @nilsnolde, this one slipped. The first part is indeed not needed. It seems it was commented out then un-commented back, and still works (since the 2nd part is undone what the 1st part did). I will remove it.

src/thor/chinese_postman_action.cc
ismailsunni: Any suggestion for this one @nilsnolde @kevinkreiser ? I am out of idea for this one

ismailsunni:
Thanks, @kevinkreiser, and @nilsnolde for the review.

I have addressed all the reviews and added more comments for the Chinese postman, except the one above (calling Loki:Search from Thor). I have also fixed the failed unit test (hopefully my last commit doesn't break anything).

Let me know if there is something that I can improve before merging or some suggestion about the Loki:Search.

kevinkreiser:
@ismailsunni thank you so much, we will take another review and try to help explain on any remaining issues. i see osx is still failing so we'll take a look at that as well!

ismailsunni:
Thanks, @kevinkreiser , I have fixed the osx one. It returns a different route, but a valid one.

src/thor/chinese_postman_action.cc
kevinkreiser: yeah this will be straight forward. ill stub in a method for you later tonight, perhaps i can even make it so that loki::Search could reuse the method that i stub it when it detects a "node-snapped" projection, then we can get a little bit of code re-use out of it too!

kevinkreiser:
actually before i remove the `loki::search` call, let me first fix the conflicts i caused you :smile:

kevinkreiser:
ok i got rid of the conflicts and updated it to work with all the other changes ive been makign lately, whew! i'll come back tomorrow and do the node snapped loki search stuff, too tired to continue at present :smile:

kevinkreiser:
i couldnt get to this last night but ill give it a try again tonight!

purew:
I wonder why github shows the protobuf3 migration as part of the diff for this PR?

kevinkreiser:
@purew good question... i wonder if that was my fault somehow

kevinkreiser:
this seems to be the culpret: https://github.com/valhalla/valhalla/pull/3428/commits/b7caf3416a00fdb097692bde4b873d200e16b805

kevinkreiser:
i'll see if i can clean up that commit as well to remove the extra diffs too before we move on

kevinkreiser:
wow that was stupid, this fixed it: https://stackoverflow.com/a/46782679/5251867

kevinkreiser:
sorry i havent forgotten about this. i will get to it this week. apologies for the delay

kevinkreiser:
still havent forgotten about this, its on my list for this week

kevinkreiser:
i'm scheduled to take a stab at implementing the bypassing of calling loki from thor tomorrow night. thanks for the patience here @ismailsunni

kevinkreiser:
@ismailsunni i pushed up the change to directly construct the edge candidates rather than calling loki from thor. this should both be faster and more correct. all of the unit tests passed locally but i did not check the visual output. would you mind building it yourself and just verifying that everything looks correct on your end?

thanks again for all of this work and for sticking with it for so long. let me know if you see anything wrong with my changes here and if not ill give it a final review and we can get it merged.

ismailsunni:
@kevinkreiser thanks a lot, Kevin. I will build and verify it this week.

kevinkreiser:
@ismailsunni there is one particular thing i was thinking about this which was the node to node routing part and how that works with the heirarchy. mainly im just writing this comment as a note to myself to review where the node ids come from. specifically if we are using nodes from all levels of the heirarchy we'll want to collapse those to just one of the nodes but still get all the edges from al the levels. and if its just one of the nodes already then we'll still need to write the little bit of logic to look across levels for edges at the node. sorry this didnt occure to me until this moment or i'd perhpas forgotten about it until now

ismailsunni:
Hi @kevinkreiser I checked with some of my examples that were previously tested. Some work fine, some don't. Most of the failed requests give this error message:
```
{"error_code":349155920,"error":":Could not find matching edge candidate","status_code":0,"status":""}
```
I checked it happens when this line is being called:
```
  TripLegBuilder::Build(options, controller, *reader, mode_costing, path.begin(), path.end(),
                        originLocation, destinationLocation, leg, algorithms, interrupt);
```
Any idea or hint why this might happen?

For the one that works, it looks good.

One other issue is our Windows Valhalla installer build is broken, but probably it's caused by a different issue, and not related to this PR.

kevinkreiser:
@ismailsunni yeah i know where that is, at the beginning of building the builder figures out w hich edge candidate on the orign and destination locations path edges matches those in the actual route. should be straight forward. can you check one failing request into the unit test so i can debug it directly?

ismailsunni:
Hi @kevinkreiser, I have added a unit test that throws the error. See: https://github.com/valhalla/valhalla/pull/3428/commits/9857c5063fb10619d6724c1e21a242cd1c5404f5#diff-074eb29c2de5047ffca9405f19edf11c4db700ea3b59d7059e4150abe4bc67f8R537

kevinkreiser:
this is still on my list i will get to fixing that test!

kevinkreiser:
Ok this is takign too long for me to solve the issues with it. I think it would be best if I clean up where my testing left off and merge what we have. Then in some subsequent PRs we can fix the remaining issues. I'll try to do that this coming week

kevinkreiser:
ok, ive been slowly working on this pr for a long time. many apologies @ismailsunni . at any rate let me give a summary of the progress and how i propose to move forward

1. after bringing the branch up to date with master after many big refactorings the tests were still passing
2. we noticed that from within thor we were calling loki to make locations for running path/matrix computations
3. since the locations are just graph nodes already, no search actually needed to be performed
4. i rewrote that bit of the code, all the tests still passed
5. i asked @ismailsunni to check with any of his local test cases to see how the code was still working and he noticed some problems
6. he was then able to write some unit tests that indeed expose these problems
7. naturally i assumed it was something about my code changes that caused this so here's what i did:
8. i reverted the code back to https://github.com/valhalla/valhalla/pull/3428/commits/4aea4ac755f0fe6abf5c71de8a02f14f7e36afce which was before i had touched it in any meaningful way (didnt even have the merge conflict fixes from master)
9. i checked out the latest version of the tests here: https://github.com/valhalla/valhalla/pull/3428/commits/9857c5063fb10619d6724c1e21a242cd1c5404f5 with the new test that should have the regression from my work
10. i back ported a couple lines of the test since they were written after all the master refactoring
11. i ran the tests and still the newest test case that shows a failure fails with the same issue:

```
[          ] Valhalla request is: {"locations":[{"lon":0.0018864552939636647,"lat":-0.00008983120447680498},{"lon":0.0022008645096242755,"lat":-0.00008983120447680498}],"costing":"auto","chinese_postman_polygon":[[0.001616961680540284,0.0],[0.0027847673387082674,0.0],[0.0027847673387082674,-0.0006288184313121905],[0.001616961680540284,-0.0006288184313121905]]}
unknown file: Failure
C++ exception with description "Could not find matching edge candidate" thrown in the test body.
[  FAILED  ] ChinesePostmanProfilesTest/ChinesePostmanTest.TestChinesePostmanCouldNotFindingMatchingEdge/0, where GetParam() = "auto" (253 ms)
```

this tells me that it's not actually my changes to filling out Locations without using loki::search but rather somethign thats already there in the implementation.

so what should we do to move forward? my proposal is the following. i'll finish up cleaning up the pr removing loki::search and other small nits i find. ill disable the failing tests. and since its all marked as beta we will understand that there is a bug out there to find and fix in a separate pr. i think this is too good of work to let sit out on a branch any longer even if it doesnt work perfectly. what does everyone else think?

kevinkreiser:
ok assuming all of this stuff builds, all that is left to do is clean up all the nitpicks and we can get it reviewed and merged. sorry again for the horrendous wait.

kevinkreiser:
@ismailsunni i am certain we can find and fix the problem. its essentially that we have a path with certain edge ids, but when we go to make the leg of the route, the edge ids that the location object has dont match the path we give it. since the path is happening only at nodes i suspect we are switching between and incoming edge and and outgoing one and so we have a mis match of edge ids. i havent been able to fix it previously but i was less focused on that than on the removing loki so... at any rate we will get it fixed!

ismailsunni:
Yes, agree,, sooner or later, the bug will be fixed!

src/thor/chinese_postman_action.cc
```diff
@@ -0,0 +1,469 @@
...
+// ordered by the best.
+int get_node_candidate_index(const valhalla::Location& location,
+                             const GraphId& edge_id,
+                             double* percent_along) {
+  // see if this edge is an edge candidate for this location
+  for (int i = 0; i < location.correlation().edges_size(); ++i) {
+    const auto& e = location.correlation().edges(i);
+    if (e.graph_id() == edge_id) {
+      // TODO: this will always be 0, this is definitely a bug
+      *percent_along = int(e.percent_along());
```
kevinkreiser, line 40: @ismailsunni i started cleaning up this source file a bit, simplifying things, refactoring a bit, adding some protections, and updating some comments but one thing i foudn that looks like a bug to me was this. previously you had passed `percent_along` as `int*` which forced the floating point number (always between 0-1) to be truncated, most likely to 0. if i let this thing stay as a floating point number lots of tests fail. is this really the behavior we wanted here? i saw down below where you check it to see whether its closer to 0 or closer to 1, whcih made me think this is not the behavior we really wanted. let me know what you think!

ismailsunni: Hi @kevinkreiser you are correct, it's not supposed to be an `int*`. I have tried your latest commit and removed the int conversion. It indeed failed a lot of tests, but actually only 2 (`TestChinesePostmanCouldNotFindingMatchingEdge` and `TestChinesePostmanMiddleEdge`). Both are throwing the same error: `Could not find matching edge candidate`. I have time today to debug it, hopefully, I got something.

kevinkreiser: awesome! let me know what you find. i'll keep working on cleaning up the code on my end too! there are a lot of ID to string conversions and ordered maps that I want to switch to ID hash maps. i'll get to that probably this weekend. sorry again this is taking so long but i finally feel like i have just enough time to make progress on it!

ismailsunni: hi @kevinkreiser 
It seems the issue is related to how the starting and end points handle the percent_along. I have checked that if I use double for the percent along, all tests are ok except when the one that the starting or endpoints are not in the node.
In this case, the algorithm seems ok, but how I handle the percent along is not ok if the percent_along is not 0 or 1.

The previous success test was a lucky case.

I have disabled and commented out on the failed test. I hope I can do more debugging next week.

kevinkreiser: sounds great! i might be able to take a look as well, ill let you know. i have some more changes coming too to speed up the cp graph a bit. its great to make progress on getting this ready for merge. thanks again for all the work!

nilsnolde:
I wanted to take a peek here, never really looked at the code too much. Still lots of cleanup to do IMO.. I followed basically Valhalla's processing pipeline when looking at the code, I got until the start of the heavy lifting in `thor_worker_t::chinese_postman` where I left a comment, will continue there at a later time. I'd like to understand exactly what it's doing, maybe the bug will appear at some point.

CMakeLists.txt
```diff
@@ -135,7 +135,7 @@ find_package(ZLIB REQUIRED)
 
 # try to find an installed boost or install locally with conan
 set(boost_VERSION "1.71")
-find_package(Boost ${boost_VERSION} QUIET)
+find_package(Boost ${boost_VERSION} EXACT QUIET)
```
nilsnolde, line 138: Oh yeah, this is messed up.. Basically I did a bit of OSRM work lately, which also switched to conan but uses boost v1.79.0. When I configured this fork, CMake picked up that version and after I did some cleanup in loki's `polygon_search` some tests failed and I couldn't see my mistake. Turns out boost::geometry must have a regression where the `within(line, poly)` also catches Linestrings which only intersect.. I'll comment in the test below and hope I can find the time to make sure it's really a regression and file a bug over at bg.

src/loki/polygon_search.cc
```diff
@@ -152,42 +169,55 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Ring>& rings_p
...
         }
       }
     }
   }
 
 // log the GeoJSON of avoided edges
 #ifdef LOGGING_LEVEL_TRACE
-  if (!avoid_edge_ids.empty()) {
-    LOG_TRACE("Avoided edges GeoJSON: \n" + to_geojson(avoid_edge_ids, reader));
+  if (mode == "avoid_polygons") {
```
nilsnolde, line 213: oops, must amend, this is stale

src/loki/worker.cc
```diff
@@ -142,6 +147,21 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
...
   }
 
+  // Save the edges for chinese postman graph
+  if (options.has_chinese_polygon()) {
+    const auto chinese_edges = edges_in_rings(options.chinese_polygon(), *reader, costing,
+                                              max_chinese_polygon_length, Purpose::CHINESE);
+    for (const auto& edge_id : chinese_edges) {
+      if (avoid_edges.find(edge_id) != avoid_edges.end()) {
+        continue;
+      }
```
nilsnolde, line 157: before this was done in thor, which seems pointless to carry useless chinese edges over in PBF

src/thor/chinese_postman_action.cc
```diff
@@ -0,0 +1,455 @@
...
+  int candidateDestinationNodeIndex;
+  double destinationPercentAlong;
+
+  // Add edges to the CP Graph, which were found from the spatial bins in loki
+  // avoided_edges
+  ChinesePostmanGraph G;
+  graph_tile_ptr tile;
+
+  // TODO(nils): originVertex and destinationVertex is literally the only things
+  // this loop looks up. that seems super wasteful, iterating over all correlated edges of both
```
nilsnolde, line 281: I got until here

test/gurka/test_chinese_postman.cc
```diff
@@ -0,0 +1,574 @@
...
+    const auto layout = gurka::detail::map_to_coordinates(ascii_map, 10);
+    // Add low length limit for excluded_polygons so it throws an error
+    chinese_postman_map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_chinese_postman",
+                                            {{"service_limits.max_exclude_polygons_length", "1000"},
+                                             {"service_limits.max_chinese_polygon_length", "1000"}});
+
+    // TODO: this is kinda of a bad graph for testing as there's more than one solution for e.g. F ->
+    // E, when covering the whole graph, it could be either one of these two and both are exactly
+    // equal in cost: { "FD", "DE", "EA", "AB", "BC", "CD", "DE", "EA", "AF", "FE", "EA", "AC", "CD",
+    // "DE" } { "FD", "DE", "EA", "AB", "BC", "CD", "DE", "EA", "AC", "CD", "DE", "EA", "AF", "FE" }
```
nilsnolde, line 369: this also took a while to understand. when I refactored a bit, some tests were failing using this map. when I went through edge-by-edge I realized they're the exact same.. I'm sure it's deterministic, but it definitely depends somehow on the order of edge insertion into the chinese graph or so

test/gurka/test_chinese_postman.cc
```diff
@@ -0,0 +1,574 @@
...
+    // Setup chinese_postman_map
+    const std::string ascii_map = R"(
+    p---------q-------r------s-----t
+    |    A----|--B----|--C---|--G  |
+    | i--|--j-|--|--k |  |   |  |  |
+    | |  |  | |  |  | |  |   |  |  |
+    | l--|--m-|--|--n |  |   |  |  |
+    |    |    |  |    |  |   |  |  |
+    |    D----|--E----|--F---|--H  |
+    u---------v-------w------x-----y
```
nilsnolde, line 344: Edge EF was found by boost.geometry 1.79.0 to be **within** the polygon `prwu` which it obviously is not.. 1.71.0 doesn't have that bug. I should make a small test case out of this soon and see if I can reproduce it stand-alone to file a bug.

test/gurka/test_chinese_postman.cc
nilsnolde: Uff, those maps are _very_ flaky.. Even in the test code it's expected to have different results for OSX and Linux. Now OSX is failing with what _seems_ like legitimate problems on first glance, but Linux is fine. If possible it'd be very good to have _only_ **one** possible solution, though I realize that might not be too easy and maybe even impossible, there often are multiple equivalent solutions/paths, probably all the more when the map gets more complex. I think there should be more lower level testing of functions rather than so many end-to-end tests which have the aim to test code of lower-level functions.

nilsnolde:
@ismailsunni a few questions I’m not totally clear on:

1. Reading some literature, I’m still not positive what the Eulerian path really represents in a _directed_ graph: does it visit each _directed_ edge once or each pair of (edge, opp_edge) once? The unit tests suggest the former. In that case, do you think it‘d be easily be possible with the current implementation to also support the second option? A typical use case of garbage collection usually wouldn’t want to visit each side of a minor road but only visit the road once and take care of both sides at the same time (at least in Berlin). I’d imagine the same is true for the military patrolling use case you implemented this for, no?

2. Are you sure your second issue in the PR description really is one? 

> Unvisited road inside the Chinese polygon
In some cases, it failed to cover all the roads inside the Chinese polygon. Until the current investigation, it failed to build a balanced graph. See the image below for example:

The edge that’s not covered doesn’t look like it’s entirely inside the polygon, but the Chinese edges are currently filtered with a „within“. Maybe it’d make sense to let the user control that in the future if it should be „within“ or „intersects“.

kevinkreiser:
I was working on this branch as well. I believe there is a logic error regarding the edges that intersect the polygon. I commented about it earlier. After I thought about it more it seemed to me that we should always include the full edge and not do anything partial. I suppose the one arunent against it would be to allow uturns at the intersection point to keep you inside the polygon but this seems somewhat unusable.

@nilsnolde the branch was building on the last commit I pushed. did changes in master break it?

nilsnolde:
> I was working on this branch as well.

Yeah I know, wanted to help a bit to make it mergeable:)

> I suppose the one arunent against it would be to allow uturns at the intersection point to keep you inside the polygon but this seems somewhat unusable.

Wouldn't be a fan of that either. I also think it's best to simply include the full edge. Which edges to include though is a matter of taste (either `within` or `intersects`), so IMO that could be a request parameter at some point.

> did changes in master break it?

Nope, the Linux build failed bcs I messed up some trace log block (which my IDE doesn't catch since I'm not configuring with trace logging), just fixed it. The tests are flaky though, that's why OSX is failing, see my comment here: https://github.com/valhalla/valhalla/pull/3428#discussion_r987623402.

ismailsunni:
Hi, @nilsnolde and @kevinkreiser thanks so much for the review. I will try to find some time to make it works again. It's been busy lately for me since I am moving back to Indonesia

ruz76:
Hi  @nilsnolde, @kevinkreiser and @ismailsunni. Is there any chance that thiss pull request will be accepted? This functionality is what I am seeking for. I would like to use it in a SAR (Search and Rescue) plugin for QGIS that is used by Police of the Czech Republic and Mountain Rescue Service.

kevinkreiser:
@ruz76 yes it will be merged. there were a number of improvements and issues we've been making for it that kept it from being merged up to this point but we definitely have it on the short list to be merged. we just need some time to make sure its doing what we expect. sorry fo rthe wait

nilsnolde:
If there's people who're interested in funding this PR, please reach out to me (public Github email), we'd be very interested. Otherwise, from my side at least, there's no plan to work on this anytime soon, even though I'd love to have this merged. And I very much doubt that the company who's behind this PR is re-activating it just for showing commitment to open source projects (with zero offense to the respective developers, not their call).

ruz76:
Thank you. I will ask the manager of the SAR project for possible funding. If not possible I can offer my time to help you with fixing code before the merge.

ismailsunni:
Hello, sorry for the late reply.
Actually, last week I planned to check again on this PR (I know it's been a super long time) since I got a small R&D time to continue this PR. Unfortunately, it was my luck this time that prevent me because I was sick for almost the whole week.
I will see what I can do this week.

ismailsunni:
Hi @nilsnolde 

> @ismailsunni a few questions I’m not totally clear on:
> 
> 1. Reading some literature, I’m still not positive what the Eulerian path really represents in a _directed_ graph: does it visit each _directed_ edge once or each pair of (edge, opp_edge) once? The unit tests suggest the former. In that case, do you think it‘d be easily be possible with the current implementation to also support the second option? A typical use case of garbage collection usually wouldn’t want to visit each side of a minor road but only visit the road once and take care of both sides at the same time (at least in Berlin). I’d imagine the same is true for the military patrolling use case you implemented this for, no?

Yes, it visits each pair of (edge, opp_edge) once. We want to start with something simpler to implement. Based on my reading, if we are allowed to visit each _directed_ edge once, the problem become Mix Chinese Postman Problem, and it's much more complex/difficult (NP-Complete) see: https://en.wikipedia.org/wiki/Mixed_Chinese_postman_problem.


> 2. Are you sure your second issue in the PR description really is one?
> 
> > Unvisited road inside the Chinese polygon
> > In some cases, it failed to cover all the roads inside the Chinese polygon. Until the current investigation, it failed to build a balanced graph. See the image below for example:
> 
> The edge that’s not covered doesn’t look like it’s entirely inside the polygon, but the Chinese edges are currently filtered with a „within“. Maybe it’d make sense to let the user control that in the future if it should be „within“ or „intersects“.

If I chooses different starting and end point, with the same polygon, it cover the uncovered road. From this, I think it's not because the filter. But it's a good idea to let the use control the filter, either within or intersect.

CMakeLists.txt
ismailsunni: @nilsnolde I also found this bug, and can be replicated here: https://gist.github.com/ismailsunni/f1b1640e1e93d5e52384c287778f70a0 

image

Line 3 should be within the polyon, but boost says no. Although boost says that the line is intersect. The result should be similar with line 1, 2, and 6.

nilsnolde: Ah awesome, thanks so much for putting together a test case @ismailsunni ! Look very legit to me when reading over (i.e. no floating point issues (you're using double), it's not trying to be spherical/geoidal etc). So, you also tried with boost v1.79 or similar?

Damn, I was sorta hoping it's us doing weird things with `within` of bg line instead of polygon/ring objects. But your test case seems to prove there's legitimate problems with boost's `within`. I didn't check your test thoroughly: are we sure it's the right winding? I keep forgetting what winding an external ring is supposed to be with bg, but I think counter-clockwise?! It might be worth it to throw a `bg::correct(poly)` in there just to make sure that's not the case, that'll correct the winding and closedness of rings.

If all that checks out, it'd be best to file a bug upstream and see what they say. It's only a minor problem for us really, we can enforce using a possibly bug-free boost 1.71 via conan and CMake. But it's prohibiting us from ever updating that version to past the version that introduced the potential bug.

nilsnolde: Also wondering if we shouldn't update this whole thing to use `geos` instead. We won't get rid of it for the time being, but @kevinkreiser recently did a major refactor of the whole admin stuff (and in fact he has an open PR to remove bg from that part entirely), and while using the C lib is somewhat ugly/hard, it might be time well spent.

I'm not saying we want to push that on you, I think for the sake of this PR it's fine for now to live with the possible bug in bg and require v1.71..

kevinkreiser: long story short yes, we should drop boost for geos. boost is not ready for spherical coordinate systems at this time

ismailsunni:
hi @kevinkreiser @nilsnolde 
To make my effort more focused, on this PR to be merged, what are the missing items that I need to take care of?

- Fix the case where the starting or endpoint is not in the node (somewhere on the edge). E.g. Fix unit test: `TestChinesePostmanMiddleEdge`
- Make sure the CIs pass
- 
- Should I fix those two cases?
- Any other?

Btw, I have a question regarding the pre-commit. I follow the instruction [here](https://github.com/valhalla/valhalla/blob/master/CONTRIBUTING.md#code-contributions). But, whenever I try to commit something, it gets reverted. And when I ran `pre-commit run --all-files `, there were a bunch of files that are modified (even if I am on the master branch). Do you know something that might be that problem?

```
isunni@wrk136:~/dev/cpp/valhalla ((HEAD detached at 28889b6b1))$ git status 
HEAD detached at 28889b6b1
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   src/baldr/curler.cc
        modified:   src/bindings/python/python.cc
        modified:   src/midgard/point2.cc
        modified:   src/mjolnir/graphbuilder.cc
        modified:   src/mjolnir/pbfgraphparser.cc
        modified:   src/odin/narrativebuilder.cc
        modified:   src/sif/dynamiccost.cc
        modified:   src/thor/triplegbuilder.cc
        modified:   src/tyr/route_serializer_osrm.cc
        modified:   test/edgecollapser.cc
        modified:   test/gurka/gurka.h
        modified:   test/gurka/test_recost.cc
        modified:   test/loki_service.cc
        modified:   test/skadi_service.cc
        modified:   test/test.cc
        modified:   test/timeparsing.cc
        modified:   valhalla/baldr/double_bucket_queue.h
        modified:   valhalla/midgard/tiles.h
```

Thanks! :)

nilsnolde:
Thanks for renewed interest @ismailsunni ! 

From my side, I think it'd be great to solve before a merge:
1. your 1. issue in the OP, i.e. the one with straight lines between locations (not sure if that coincides with `Fix the case where the starting or endpoint is not in the node`?)
2. try updating the tests so the result of the chinese postman is not architecture dependent, e.g. https://github.com/valhalla/valhalla/pull/3428#discussion_r987623402

IMO then it'd be in shape for a beta version. You're mentioning in your 1. problem that it's happening to around 10% of the requests. I'm torn a bit if I'd be comfortable labeling it even beta if that's still the case, but others might ok with it?

nilsnolde:
Oh and with pre-commit: we changed the setup recently. Maybe you best remove the `rm -f .git/hooks/pre-comit` and `pre-commit install` again. Let us know if that doesn't work, but also what commands you did and what the actual output is.

ismailsunni:
Thanks @nilsnolde I will just manually run the format script. Removing and re-installing the pre-commit back doesn't resolve the issue in my case.

Ok, I will try my best for those two items you mentioned.

ruz76:
Thank you @ismailsunni for you work. I am now in a situation to select some solution in a couple of weeks. For now I am testing also https://github.com/rkistner/chinese-postman solution. Generally speaking I checked 6-7 solutions and only your and solution from Kistner seems to be working. Since my solution for Seach and Rescue system is based on docker I do not see a problem to use Valhalla as backend for routing solutions. I have asked manager of our project for possible financial support. They did not say no or yes, so I think that it is still open. We have a meeting next Tuesday, so I will ask again. The project has limited budget, but I think that something like 500 USD may not be a problem to spent on a support.

nilsnolde:
Thanks for the notice @ruz76 and nice of you to offer financial support. However, we prefer to keep those things off Github, it's about pure technical discussions here (no worries, I did the same mistake before too..). Both @ismailsunni and me have public email addresses, better use those. Though not sure about Ismail's current work situation. 

I can say from my side I won't have time to work on this anytime soon, i.e. before June/July.

PR #3431, opened 2021-11-29
add directions_type from directions_options

liyinxiao:
# Issue

Fix bug - there's a mismatch between [doc](https://valhalla.readthedocs.io/en/latest/api/turn-by-turn/api-reference/) and code. In the [doc](https://valhalla.readthedocs.io/en/latest/api/turn-by-turn/api-reference/), `directions_type` is listed as an option inside `directions_options`, and I'm adding it back.  

Options | Description
-- | --
directions_type | An enum with 3 values.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
    * FIXED: Add string for Use:kPedestrianCrossing to fix null output in to_string(Use). [#3416](https://github.com/valhalla/valhalla/pull/3416)
    * FIXED: Remove simple restrictions check for pedestrian cost calculation. [#3423](https://github.com/valhalla/valhalla/pull/3423)
    * FIXED: Parse "highway=busway" OSM tag: https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbusway [#3413](https://github.com/valhalla/valhalla/pull/3413)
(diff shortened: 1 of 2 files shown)


kevinkreiser:
[the docs clearly state:](https://github.com/valhalla/valhalla/blob/master/docs/api/turn-by-turn/api-reference.md#directions-options)

> Directions options should be specified at the top level of the JSON, and usage of the `directions_options` nested structure is deprecated.

I think maybe you misunderstand what this means. `units`, `language`, `directions_type` and `narrative` should be top level keys in the request, nesting them within a `directions_options` object is deprecated. If you think we could word this better to remove confusion please feel free to submit another pr with a documentation change! Sorry for the confusion!

liyinxiao:
Thanks for the clarification on how this originated. The most confusing part in the doc is while you have three options listed under `directions_options`, which is deprecated, I would expect all three options to work, or all three options to not work. Now, two options work while one option doesn't..

Options | Description
-- | --
units | Distance units for output. 
language | The language of the narration instructions based on the IETF BCP 47 language tag string.
directions_type | An enum with 3 values.

I would propose to remove `directions_type` from the table in the doc, so no one would even attempt to nest `directions_type` in a `directions_options` object (which is not only deprecated but also will not work). What do you think?

PR #3436, opened 2021-12-03
clean up to fix library error

xlqian:
# Issue

We checked out the latest version (master) to compile inside of our project. Some errors occurred when compiling due to the "bad" include in some headers. Here is the fix 

## Tasklist

### DIFF
--- a/valhalla/mjolnir/osmway.h
+++ b/valhalla/mjolnir/osmway.h
@@ -7,13 +7,11 @@
 #include <vector>
 
 #include <valhalla/baldr/graphconstants.h>
+#include <valhalla/mjolnir/osmpronunciation.h>
 #include <valhalla/mjolnir/uniquenames.h>
 
 namespace valhalla {
 namespace mjolnir {
-
-// Forward declaration
-struct OSMPronunciation;
 
 // OSM way
 struct OSMWay {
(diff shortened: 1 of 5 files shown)


valhalla/mjolnir/osmway.h
```diff
@@ -6,13 +6,15 @@
 #include <string>
 #include <vector>
 
-#include "mjolnir/osmpronunciation.h"
```
kevinkreiser, line 9: couldnt the fix have been as simple as <valhalla/mjolnir/osmpronunciation.h> here, and then don't do any of the other changes?

xlqian: unluckily No, that was the first thing I did LOL

xlqian: oops, that idea seems should work.. I have done something wrong certainly...

kevinkreiser:
for some odd reason the build is breaking recently on the edge filter tests in release mode. we might want to look into that. for now ive just rerun the build but its happened a few times now so there is definitely something up. not related to this change, this looks good to me

xlqian:
I was just trying to find out what has broken the CI in my PR :smiling_face_with_tear:

kevinkreiser (approved):

PR #3444, opened 2021-12-06
add section about the new demo server

nilsnolde:
we now finalized the fossgis demo server which hosts an app now: https://valhalla.openstreetmap.de. the repo is available here: https://github.com/gis-ops/valhalla-app. 

I guess it has the best reach if we include it fairly top-level in the root README, so I put it right under the LICENSE. I also re-arranged a little to make the build parts more coherent. I don't have strong feelings about the wording or whatever, I'd just like to include that section where it's seen immediately ~~without scrolling~~. ok for everyone?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,6 @@ Valhalla is an open source routing engine and accompanying libraries for use wit
 
 Valhalla, and all of the projects under the Valhalla organization, use the [MIT License](COPYING).  Avatar/logo by [Jordan](https://www.instagram.com/jaykaydraws/)
 
-## Demo Server
-
-[FOSSGIS e.V.](https://fossgis.de) hosts a demo server which is open to the public and includes a full planet graph on https://valhalla.openstreetmap.de wih an [open-source web app](https://github.com/gis-ops/valhalla-app). The HTTP API is accessible on the same domain, simply append the appropriate endpoint path, e.g. https://valhalla.openstreetmap.de/isochrone. Usage of the demo server follows the usual fair-usage policy as OSRM & Nominatim demo servers (somewhat enforced by [rate limits](https://github.com/valhalla/valhalla/discussions/3373#discussioncomment-1644713)).
-
 ## Overview
 
 There are several key features that we hope can differentiate the Valhalla project from other routing and network analysis engines. They are:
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -27,6 +27,10 @@ Valhalla is an open source routing engine and accompanying libraries for use wit
 
 Valhalla, and all of the projects under the Valhalla organization, use the [MIT License](COPYING).  Avatar/logo by [Jordan](https://www.instagram.com/jaykaydraws/)
 
+## Demo Server
```
kevinkreiser, line 30: maybe this should go at least below the overview after we've described a little bit more of what valhalla is.

nilsnolde: agreed, I moved it below the overview

TimMcCauley:
@nilsnolde  @kevinkreiser there is a part at https://github.com/valhalla/valhalla/blob/master/docs/api/turn-by-turn/api-reference.md that has a link to the old demo server ("View an interactive demo"), maybe this could also be replaced?

nilsnolde:
yep, good idea @TimMcCauley , I also added it to the isochrone api-reference, in case that's fine for the others

docs/api/turn-by-turn/api-reference.md
```diff
@@ -2,7 +2,7 @@
 
 Valhalla's routing service (a.k.a. turn-by-turn), is an open-source routing service that lets you integrate routing and navigation into a web or mobile application.
 
-[View an interactive demo](http://valhalla.github.io/demos/routing)
+[View an interactive demo](https://valhalla.openstreetmap.de)
```
kevinkreiser, line 5: i dont know im iffy on these types of changes. i think there is room for both types of demos. the programs we have have in the demos repo are for developers who are running their own valhalla and want to play with it. the demo at valhalla.openstreetmap.de is more for high level people who want to just see vaguely how it works. i'm happy to have both linked but i dont want to lose visibility of the tools we have in the demos repo. i personally think linking to the new interactive demos that are hosted makes the most sense on the main README because that is where people who are only taking a shallower read and who would appreciate a front end would tend to go.

basically i dont want to lose the links to the existing demos and tools even if they require you to be running your own service and are terrible as UI goes hahahha, not saying we cant put in the link to the new stuff, just dont want to lose the old stuff as i think it still serves a purpose (debugging for developers)

how can we serve both use-cases?

nilsnolde: no worries, I think you're right. this is more the area where a dev is roaming around, makes sense to link more of a dev tool. I removed those changes. that reminds me though that the demos could be updated to hit the open server rather than the key-protected current server to work out-of-the-box, right?

kevinkreiser (approved):

PR #3450, opened 2021-12-08
Fix ubsan error due to possible invalid shift exponent

mandeepsandhu:
ubsan complained about the shift exponent in NodeInfo::heading() being
possibly too large under some conditions.

Fixed by checking for valid localidx. If invalid index is passed in, we
log a warning and return a heading of 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
    * FIXED: Parse "highway=busway" OSM tag: https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbusway [#3413](https://github.com/valhalla/valhalla/pull/3413)
    * FIXED: workaround python's ArgumentParser bug to not accept negative numbers as arguments [#3443](https://github.com/valhalla/valhalla/pull/3443)
    * FIXED: Undefined behaviour on some platforms due to unaligned reads [#3447](https://github.com/valhalla/valhalla/pull/3447)
+   * FIXED: Fixed undefined behavior due to invalid shift exponent when getting edge's heading [#3450](https://github.com/valhalla/valhalla/pull/3450)
 
 * **Enhancement**
    * CHANGED: Pronunciation for names and destinations [#3132](https://github.com/valhalla/valhalla/pull/3132)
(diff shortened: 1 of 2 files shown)


kevinkreiser:
i saw this too but what i wonder is why is the local index out of range in the first place?

mandeepsandhu:
> what changed that exposed this particular situation?

I haven't run bisect, so can't say which change introduced this bug. Was ubsan always on? or was it switched on recently?

mandeepsandhu:
> i saw this too but what i wonder is why is the local index out of range in the first place?

I see some places where we're looping the edge count and passing that as index into `NodeInfo::heading` (without capping it at 7)), so seems quite likely that this index goes out of valid range. eg: [`IsInternalIntersection`](https://github.com/valhalla/valhalla/blob/master/src/mjolnir/graphenhancer.cc#L640) passes the `idx` arg as-is into `NodeInfo::heading` (`idx` if fetch from a [loop](https://github.com/valhalla/valhalla/blob/master/src/mjolnir/graphenhancer.cc#L1489) through the node's edges). There might be more instances of such usage.

dnesbitt61:
This might only be used in map-matching. I am thinking we should use turn types there and then we wouldn't even need this

kevinkreiser:
yep @mandeepsandhu i totally get why it happens i just dont get why it started happening in the `check` target semi-recently. something changing elsewhere must have caused it and thats what im interested in.

and yeah @dnesbitt61 has made this point before i think, perhaps we can just jetison it completely?

dgearhart:
> This might only be used in map-matching. I am thinking we should use turn types there and then we wouldn't even need this

@dnesbitt61 @kevinkreiser we use the intersecting edge begin heading - can not remove it

dgearhart (approved):
:ship: we will keep our eyes out for reason of recent failure

mandeepsandhu:
> yep @mandeepsandhu i totally get why it happens i just dont get why it started happening in the `check` target semi-recently. something changing elsewhere must have caused it and thats what im interested in.

As I said, a time-consuming git-bisect can telll us exactly when it started occurring. Initially I though usan must've been enabled recently, but I think its been on for some time now (unless some ubsan configuration changed recently).
> 
> and yeah @dnesbitt61 has made this point before i think, perhaps we can just jetison it completely?

dnesbitt61:
Unfortunately we need it (even if we change map matching) to get the heading of intersecting edges along the path. If we remove it we would have to access other tiles (level 2 in particular) along the route which would slow things down considerably.

PR #3451, opened 2021-12-08
Fix crash in DoubleBucketQueue

andreasjunghans:
When decreasing label costs, std::remove already removes the label from prevbucket. And since std::remove returns a past-the-end iterator, feeding it to erase() is illegal and caused crashes for us.

### DIFF
--- a/valhalla/baldr/double_bucket_queue.h
+++ b/valhalla/baldr/double_bucket_queue.h
@@ -140,7 +140,7 @@ public:
     if (prevbucket != newbucket) {
       // Add label to newbucket and remove from previous bucket
       newbucket.push_back(label);
-      prevbucket.erase(std::remove(prevbucket.begin(), prevbucket.end(), label));
+      std::remove(prevbucket.begin(), prevbucket.end(), label);
     }
   }
 


dnesbitt61:
I do not think this is right. Looks like tests fail. The code (prior to your change) uses the [erase-remove idiom ](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom) which I think is the correct way to efficiently remove the element from the vector. std::remove does not actually erase the element - it logically rearranges the vector without physically removing the element.

andreasjunghans:
Oops, you're right. I will have to investigate some more. We did get crashes on this line for long routes (segmentation fault in __memmove_avx_unaligned_erms), and removing the erase() call "fixed" it. But as you pointed out, that's obviously wrong - sorry for jumping to conclusions and not RTFM properly.

andreasjunghans:
Figured it out. We made same changes that caused decrease() to be called for a label that was already popped. That in turn made the std::remove() call return prevbucket.end(), which made prevbucket.erase() crash. Sorry for the noise!

Issue #3099, opened 2021-05-25
Warnings section in response


nilsnolde:
Would be nice to tell a user when certain things happen while processing, e.g. clamping values to their allowed maximum, enabling `destination_only` roads in 2. pass, deprecating request parameters etc.

IMO it should be a top-level `warnings` array of strings which is empty by default and all modules can add their share during processing: probably implement the most obvious ones first and add others wherever it feels appropriate in the future.

Suggestion:
1.  Include the `warning` field in `proto/api.proto` as `repeated string`. Or do we need extra info to justify a `Warning` message?
2.  Add warnings where sensible in the code and serialize for every action in `tyr/serializers.cc`

Opinions?

Arkadipta-Chatterjee:
Can i work on this?

kevinkreiser:
I think we should take @nilsnolde's suggestions above about which ones to implement, and actually flesh out the details a bit. I think for someone who doesnt know the code it will be hard otherwise to contribute.

Regrading where to keep this information, this was one of the ideas behind the info.proto message. I think we can add repeated strings to the `Info` message type to track warnings that should be user facing, i could also be confivinced that this structure isnt user facing and that we sh ould add it somewhere else to a part of the structure we already serialize out like `Options`

nilsnolde:
that's kind of you to offer @Arkadipta-Chatterjee .. on second thought this might actually not be a `good first issue` for the reasons @kevinkreiser mentioned: it does need a fairly in-depth knowledge of the code/processing to know when to add a "warning" along the processing. the scaffolding around serialization, setting up pbf etc does not need that expertise, but that's also very quickly done.

if you want to contribute, may I suggest the following ones:
https://github.com/valhalla/valhalla/issues/3212 (making isochrone resolution more dynamic so we can get rid of the most glaring artifacts for smaller isochrones)
https://github.com/valhalla/valhalla/issues/3010 (if you have access to a Windows machine, this would be awesome to finally solve, should be fairly mechanical around spatialite integration in the code)

nilsnolde:
but this issue would be nice for our use case as well, so after all those other todo's I'd like to close this as well.

> i could also be confivinced that this structure isnt user facing and that we sh ould add it somewhere else to a part of the structure we already serialize out like Options

"user-facing" you mean "returned in the response" @kevinkreiser ? if so, how would these warnings be beneficial if they're not returned to a user?

kevinkreiser:
[...]

PR #3454, opened 2021-12-11
Remove All PBF Defaults from Interface

kevinkreiser:
This is another step closer to proto3 for our protos. The main change here is to remove the use of `default` which is not supported in proto3. There were two ways to do this.

The first way is to simply remove the default value from the definition. This works if the default is what proto2/3 already default a scalar value to, from the manual:

> If the default value is not specified for an optional element, a type-specific default value is used instead: for strings, the default value is the empty string. For bytes, the default value is the empty byte string. For bools, the default value is false. For numeric types, the default value is zero. For enums, the default value is the first value listed in the enum's type definition.

For some of the values they defaulted to a certain enum value which was not the first enum value so i re-ordered those enums so that the default would be the first one. For some of the values that didnt default to `false` i simply change the name of the field to have the semantically opposite meaning so that a default of `false` would make sense.

The second way is to move the defaulting inside of the code. This was required for those defaults which were not trivial like those above.

In addition to these things I did a couple more clean ups. The first one was I got rid of the `do_no_track` option as no one ever used it and we didnt really adhere to it. The second is that i deprecated the `best_paths` option in favor of the more descriptive and already existing `alternates` option. this parameter is still supported of course but it is translated into `alternates` at request time. I also made sure that all enums started with the first value at `0` which is required by proto3.

NOTE: since deprecating `best_paths` i've also updated the config elements that configure the maximum values for that parameter. if you use an existing config the service will fail to start. the good news is you can just rename the config option (and subtract 1) and move on with your life. if you generate your configs before running the service you wont notice an issue.


test/loki_service.cc
```diff
@@ -90,14 +90,6 @@ const std::vector<http_request_t> valhalla_requests{
...
         GET,
         R"(/route?json={"locations":[{"lon":0,"lat":0},{"lon":0,"lat":0}],"costing":"pedestrian","exclude_locations":[{"lon":0,"lat":0}]})"),
-    http_request_t(
-        POST,
-        "/trace_attributes",
-        R"({"shape":[{"lat":37.8077440,"lon":-122.4197010},{"lat":37.8077440,"lon":-122.4197560},{"lat":37.8077450,"lon":-122.4198180}],"shape_match":"map_snap","best_paths":0,"costing":"pedestrian","directions_options":{"units":"miles"}})"),
-    http_request_t(
-        POST,
-        "/trace_attributes",
-        R"({"shape":[{"lat":37.8077440,"lon":-122.4197010},{"lat":37.8077440,"lon":-122.4197560},{"lat":37.8077450,"lon":-122.4198180}],"shape_match":"map_snap","best_paths":5,"costing":"pedestrian","directions_options":{"units":"miles"}})"),
```
kevinkreiser, line 100: instead of sending an error about asking for too many alternates in map matching i changed it to do what we do for regular routing which is to cap it. in the end this isnt a breaking change as the user wasnt getting back a usable response in the first place when asking for too few or too many routes.

nilsnolde: good as it's more inline with the general philosophy. I still think these implicit request parameter changes should be communicated in the response, e.g. #3099 . some day.. :)

nilsnolde (approved):
:+1:

PR #3455, opened 2021-12-13
Include roads under construction into the graph

genadz:
# Issue

Add a config option to include/exclude roads under construction from the graph. Such roads are still completely excluded from routing, but they might be useful in matching algorithm.

Implementation details:
- add a new config option `mjolnir.include_constructions`. In case it's `true` -> we include roads with tags combination `highway=construction+construction=<road_class>` into the graph. **By default** this option is **turned off**.
- add a special 'use'-type  `Use::kConstruction` for such roads. This is **the only difference** between ordinary edges and constructions in valhalla graph. It means that constructions, for example, may have `access` tag that allows `auto`,`bicycle`,`pedestrian` and etc.;
- check 'use'-type in costing to completely **exclude** roads under construction **from routing for all modes**;
- check 'use'-type in shortcutsbuilding procedure to **prevent constructions being a part of any shortcut**.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


lua/graph.lua
```diff
@@ -26,7 +26,6 @@ highway = {
 ["pedestrian"] =        {["auto_forward"] = "false", ["truck_forward"] = "false", ["bus_forward"] = "false", ["taxi_forward"] = "false", ["moped_forward"] = "false", ["motorcycle_forward"] = "false", ["pedestrian_forward"] = "true",  ["bike_forward"] = "false"},
 ["steps"] =             {["auto_forward"] = "false", ["truck_forward"] = "false", ["bus_forward"] = "false", ["taxi_forward"] = "false", ["moped_forward"] = "false", ["motorcycle_forward"] = "false", ["pedestrian_forward"] = "true",  ["bike_forward"] = "true"},
 ["bridleway"] =         {["auto_forward"] = "false", ["truck_forward"] = "false", ["bus_forward"] = "false", ["taxi_forward"] = "false", ["moped_forward"] = "false", ["motorcycle_forward"] = "false", ["pedestrian_forward"] = "false", ["bike_forward"] = "false"},
-["construction"] =      {["auto_forward"] = "false", ["truck_forward"] = "false", ["bus_forward"] = "false", ["taxi_forward"] = "false", ["moped_forward"] = "false", ["motorcycle_forward"] = "false", ["pedestrian_forward"] = "false", ["bike_forward"] = "false"},
```
genadz, line 29: `construction` tag shouldn't influence on `access`

lua/graph.lua
```diff
@@ -851,12 +850,15 @@ end
 --returns 1 if you should filter this way 0 otherwise
 function filter_tags_generic(kv)
 
-  if (kv["highway"] == "construction" or kv["highway"] == "proposed") then
+  if (kv["highway"] == "construction" and kv["construction"] == nil) or kv["highway"] == "proposed" then
```
genadz, line 853: filter out constructions with incomplete tag info

lua/graph.lua
```diff
@@ -851,12 +850,15 @@ end
...
     return 1
   end
 
   --figure out what basic type of road it is
   local forward = highway[kv["highway"]]
+  if kv["highway"] == "construction" then
+    forward = highway[kv["construction"]]
```
genadz, line 860: road class should be defined in `construction` tag https://wiki.openstreetmap.org/wiki/Tag:highway%3Dconstruction

lua/graph.lua
```diff
@@ -1426,7 +1433,9 @@ function filter_tags_generic(kv)
   local use = use[kv["service"]]
 
   if kv["highway"] then
-     if kv["highway"] == "track" then
+     if kv["highway"] == "construction" then
+        use = 55
```
genadz, line 1437: assign a corresponding `use`-type

lua/graph.lua
```diff
@@ -1467,7 +1476,8 @@ function filter_tags_generic(kv)
     use = 0 --general road, no special use
   end
 
-  if (kv["access"] == "emergency" or kv["emergency"] == "yes") and
+  -- do not change 'construction' use
+  if use ~= 55 and (kv["access"] == "emergency" or kv["emergency"] == "yes") and
```
genadz, line 1480: `construction` 'use' shouldn't be overrided

scripts/valhalla_build_config
```diff
@@ -35,6 +35,7 @@ config = {
     'hierarchy': True,
     'shortcuts': True,
     'include_driveways': True,
+    'include_constructions': False,
```
genadz, line 38: configure include or exclude roads under construction

src/mjolnir/graphenhancer.cc
```diff
@@ -862,7 +862,8 @@ uint32_t GetDensity(GraphReader& reader,
           // Exclude non-roads (parking, walkways, ferries, etc.)
           if (directededge->is_road() || directededge->use() == Use::kRamp ||
               directededge->use() == Use::kTurnChannel || directededge->use() == Use::kAlley ||
-              directededge->use() == Use::kEmergencyAccess) {
+              directededge->use() == Use::kEmergencyAccess ||
+              directededge->use() == Use::kConstruction) {
```
genadz, line 866: exclude roads under construction from density calculation

src/mjolnir/shortcutbuilder.cc
```diff
@@ -109,7 +109,8 @@ GraphId GetOpposingEdge(const GraphId& node,
   for (uint32_t i = 0, n = nodeinfo->edge_count(); i < n; i++, directededge++, ++edgeid) {
     if (directededge->use() == Use::kTransitConnection ||
         directededge->use() == Use::kEgressConnection ||
-        directededge->use() == Use::kPlatformConnection) {
+        directededge->use() == Use::kPlatformConnection ||
+        directededge->use() == Use::kConstruction) {
```
genadz, line 113: exclude constructions from shortcuts building

src/mjolnir/shortcutbuilder.cc
```diff
@@ -199,7 +200,7 @@ bool CanContract(GraphReader& reader,
     if (!directededge->is_shortcut() && directededge->use() != Use::kTransitConnection &&
         directededge->use() != Use::kEgressConnection &&
         directededge->use() != Use::kPlatformConnection && !directededge->start_restriction() &&
-        !directededge->end_restriction()) {
+        !directededge->end_restriction() && directededge->use() != Use::kConstruction) {
```
genadz, line 203: exclude constructions from shortcuts building

src/mjolnir/shortcutbuilder.cc
```diff
@@ -386,7 +387,8 @@ uint32_t AddShortcutEdges(GraphReader& reader,
     const DirectedEdge* directededge = tile->directededge(edge_id);
     if (directededge->use() == Use::kTransitConnection ||
         directededge->use() == Use::kEgressConnection ||
-        directededge->use() == Use::kPlatformConnection || directededge->bss_connection()) {
+        directededge->use() == Use::kPlatformConnection || directededge->bss_connection() ||
+        directededge->use() == Use::kConstruction) {
```
genadz, line 391: exclude constructions from shortcuts building

test/CMakeLists.txt
```diff
@@ -124,19 +124,19 @@ set_target_properties(build_timezones PROPERTIES FOLDER "Tests")
 add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/test/data/utrecht_tiles/traffic.tar
   COMMAND ${CMAKE_COMMAND} -E make_directory test/data/utrecht_tiles/
   COMMAND ${CMAKE_BINARY_DIR}/valhalla_build_tiles
-      --inline-config '{"mjolnir":{"id_table_size":1000,"tile_dir":"test/data/utrecht_tiles","timezone":"test/data/tz.sqlite","admin":"${VALHALLA_SOURCE_DIR}/test/data/netherlands_admin.sqlite","hierarchy":true,"shortcuts":true,"concurrency":1,"logging":{"type":""}}}'
+      --inline-config '{"mjolnir":{"id_table_size":1000,"tile_dir":"test/data/utrecht_tiles","timezone":"test/data/tz.sqlite","admin":"${VALHALLA_SOURCE_DIR}/test/data/netherlands_admin.sqlite","include_constructions":true,"hierarchy":true,"shortcuts":true,"concurrency":1,"logging":{"type":""}}}'
```
genadz, line 127: include constructions into `utrecth` test map

test/gurka/test_construction.cc
```diff
@@ -0,0 +1,159 @@
...
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/construction",
+                               {{"mjolnir.include_constructions", "false"}});
+
+  baldr::GraphReader reader(map.config.get_child("mjolnir"));
+
+  auto AB = gurka::findEdge(reader, layout, "AB", "B");
+  ASSERT_NE(std::get<0>(AB), baldr::GraphId{}) << "missing valid road AB";
+
+  auto BC = gurka::findEdge(reader, layout, "BC", "C");
+  EXPECT_EQ(std::get<0>(BC), baldr::GraphId{}) << "constructions should be excluded";
```
genadz, line 43: `include_constructions=false` -> it's expected that constructions will be excluded from the graph

test/gurka/test_construction.cc
```diff
@@ -0,0 +1,159 @@
...
+
+  baldr::GraphReader reader(map.config.get_child("mjolnir"));
+
+  auto AB = gurka::findEdge(reader, layout, "AB", "B");
+  ASSERT_NE(std::get<0>(AB), baldr::GraphId{}) << "missing valid road AB";
+
+  auto BC = gurka::findEdge(reader, layout, "BC", "C");
+  ASSERT_NE(std::get<0>(BC), baldr::GraphId{}) << "constructions should be included";
+  EXPECT_EQ(std::get<1>(BC)->use(), baldr::Use::kConstruction)
+      << "invalid 'use'-type for construction edge";
```
genadz, line 62: check that the road under construction is included and has a proper `use`

test/gurka/test_construction.cc
```diff
@@ -0,0 +1,159 @@
...
+
+  auto CD = gurka::findEdge(reader, layout, "CD", "D");
+  EXPECT_EQ(std::get<0>(CD), baldr::GraphId{})
+      << "constructions with incomplete tags should be excluded";
```
genadz, line 66: even if `include_constructions=true` roads with missing `construction` tag should be filtered out

test/scripts/test_valhalla_build_extract.py
```diff
@@ -22,10 +22,10 @@ def test_create_extracts(self):
         tile_count = valhalla_build_extract.get_tile_count(TILE_PATH)
 
         # test that the index has the right offsets/sizes
-        exp_tuples = ((2560, 25568, 290368), (294912, 410441, 658856), (955392, 6549282, 6051320))
+        exp_tuples = ((2560, 25568, 291912), (296448, 410441, 662496), (960512, 6549282, 6059608))
```
genadz, line 25: result of including constructions into the test `utrecht` map

valhalla/baldr/graphconstants.h
```diff
@@ -314,7 +314,7 @@ enum class Use : uint8_t {
   kEgressConnection = 52,   // Connection to a egress node
   kPlatformConnection = 53, // Connection to a platform node
   kTransitConnection = 54,  // Connection to multi-use transit stop
-
+  kConstruction = 55,       // Road under construction
```
genadz, line 317: add new `use`

valhalla/sif/dynamiccost.h
```diff
@@ -233,7 +233,7 @@ class DynamicCost {
         ((disallow_mask & kDisallowEndRestriction) && edge->end_restriction()) ||
         ((disallow_mask & kDisallowSimpleRestriction) && edge->restrictions()) ||
         ((disallow_mask & kDisallowShortcut) && edge->is_shortcut());
-    return accessible && !assumed_restricted;
+    return accessible && !assumed_restricted && (edge->use() != baldr::Use::kConstruction);
```
genadz, line 236: return `false` in `Allowed` method for roads under construction

valhalla/sif/dynamiccost.h
```diff
@@ -247,9 +247,10 @@ class DynamicCost {
     // you have forward access for the mode you care about
     // you dont care about what mode has access so long as its forward
     // you dont care about the direction the mode has access to
-    return (edge->forwardaccess() & access_mask_) ||
-           (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
-           (ignore_oneways_ && (edge->reverseaccess() & access_mask_));
+    return ((edge->forwardaccess() & access_mask_) ||
+            (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
+            (ignore_oneways_ && (edge->reverseaccess() & access_mask_))) &&
+           (edge->use() != baldr::Use::kConstruction);
```
genadz, line 253: return `false` in `IsAccessible` method for roads under construction.

Probably this is not the best place for it. The only reason I added this check here - is to avoid copying. This method is used in every costing model to check if an edge is routable or not.

genadz:
I will add performance results a little bit later.

genadz:
> I will add performance results a little bit later.

No changes

taginfo.json
```diff
@@ -2767,6 +2767,206 @@
...
       "description": "Allows access for all modes unless the mode override is set except for bicycles."
     },
+    {
+      "key": "construction",
+      "value": "motorway",
+      "object_types": [
+        "way"
+      ],
+      "description": "Motorway road under construction."
+    },
```
genadz, line 2777: added values based on https://wiki.openstreetmap.org/wiki/Tag:highway%3Dconstruction and values in `highway` tag from this file

valhalla/sif/dynamiccost.h
dnesbitt61: this seems fine to me.

dnesbitt61 (commented):
Can you provide reasons why you would allow roads under construction in a route?

genadz:
> Can you provide reasons why you would allow roads under construction in a route?

no. that's why I explicitly check if `edge->use() == Use::kConstruction` in `Allowed` methods. The reason why somebody would like to see roads under in the graph it's not a routing case. Constructions may be useful in cases when you want to associate some information with valhalla graph. And, if this information (for example, OpenLR record that refers to some road) for some reason relates to the road under construction, it will be mismatched.

kevinkreiser:
this looks great to me ill give it a review later today!

proto/trip.proto
```diff
@@ -60,6 +60,7 @@ message TripLeg {
     kEgressConnectionUse = 52;   // Connection between transit station and transit egress
     kPlatformConnectionUse = 53; // Connection between transit station and transit platform
     kTransitConnectionUse = 54;  // Connection between road network and transit egress
+    kConstructionUse = 55;          // Road under construction
```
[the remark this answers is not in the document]
genadz: good catch, thank you! will update

valhalla/sif/dynamiccost.h
genadz: hmm. looks like this code may break backward compatibility of valhalla tiles and router. I mean, if you have old version of valhalla router but trying to use new tiles, you might get a route through constructions, because constractions have non-zero access

genadz: so, probably we should turn off access for constructions

genadz: updated

scripts/valhalla_build_config
kevinkreiser: nit: the plural `constructions` sounds odd. probably just the singular `construction` is better.

src/mjolnir/directededgebuilder.cc
```diff
@@ -163,6 +163,16 @@ DirectedEdgeBuilder::DirectedEdgeBuilder(const OSMWay& way,
       }
     }
   }
+  // Explicitely turn off access for constructions. It's done for backward compatibility
+  // of valhalla tiles and valhalla routing. In case we allow non-zero access then older
+  // versions of router will work with new tiles incorrectly. They would start to route
+  // on roads under construction because they're not awary about new 'Use::kConstruction'
```
kevinkreiser, line 169:
```suggestion
  // on roads under construction because they're not aware about new 'Use::kConstruction'
```

kevinkreiser (approved):
other than tiny nitpicks i didnt see anything left to do here. i saw in shortcut building you excluded them, do you need to do the same in shortcut recovery to make sure the recovery rate stays high?

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
    * CHANGED: Refactored GraphBuilder::CreateSignInfoList [#3438](https://github.com/valhalla/valhalla/pull/3438)
    * ADDED: Add support for LZ4 compressed elevation tiles [#3401](https://github.com/valhalla/valhalla/pull/3401)
    * CHANGED: Rearranged some of the protobufs to remove redundancy [#3452](https://github.com/valhalla/valhalla/pull/3452)
+   * ADDED: Add `include_constructions` option into the config to include/exclude roads under construction from the graph [#3455](https://github.com/valhalla/valhalla/pull/3455)
```
kevinkreiser, line 37: nit: should get moved down to the bottom before merge

src/mjolnir/directededgebuilder.cc
```diff
@@ -163,6 +163,16 @@ DirectedEdgeBuilder::DirectedEdgeBuilder(const OSMWay& way,
...
+  // and use only access mode to check if an edge is routable or not.
+  if (way.use() == Use::kConstruction) {
+    forward_access = 0;
+    reverse_access = 0;
+    LOG_DEBUG("Turn off access mode for road under construction.");
```
kevinkreiser, line 174: if we left this lua in:
https://github.com/valhalla/valhalla/pull/3455/files#r767592222

then we wouldnt need to do this here right? or does something further down in the updated lua turn access back on?

genadz: yes, there is a bunch of logic in the lua function (`function filter_tags_generic(kv)`) that may override access tags. Moreover, some access tags may be changed in c++ code that goes after `parse_ways` stage.

CHANGELOG.md
genadz: ok

genadz:
> other than tiny nitpicks i didnt see anything left to do here. i saw in shortcut building you excluded them, do you need to do the same in shortcut recovery to make sure the recovery rate stays high?

yeah, make sense.

genadz:
@kevinkreiser I fixed your comments but still digging into RAD diff

kevinkreiser:
@genadz any luck on the RAD diffs?

genadz:
> @genadz any luck on the RAD diffs?

Partially. I fixed some cases but still have small diff. Continue debugging

genadz:
@kevinkreiser I've finished with RAD tests. Out of ~4000 test routes there are 5 unique diffs (1 in routing and 4 in instructions). I've reviewed them manually, 4 of them look OK, 1 is incorrect but it's also incorrect on master because of bad data. On correct data instructions also became correct (for master and for the branch).

lua/graph.lua
```diff
@@ -1404,17 +1406,20 @@ function filter_tags_generic(kv)
    end
 
   --set a few flags
-  local road_class = road_class[kv["highway"]]
+  local rc = road_class[kv["highway"]]
+  if kv["highway"] == "construction" then
+    rc = road_class[kv["construction"]]
```
kevinkreiser, line 1411: maybe i didnt see this but how do we get away with not checking that `rc` isn't `nil`. it seems to me someone could but a crazy value in there that isnt in our map. maybe we do this on the c++ side somehow or i propbably just missed it

src/mjolnir/graphenhancer.cc
```diff
@@ -1219,8 +1219,8 @@ void ProcessEdgeTransitions(const uint32_t idx,
     uint32_t left_count = 0;
     if (ntrans > 2) {
       for (uint32_t j = 0; j < ntrans; ++j) {
-        // Skip the from and to edges
-        if (j == i || j == idx) {
+        // Skip the from and to edges; also skip roads under construction
+        if (j == i || j == idx || edges[j].use() == Use::kConstruction) {
```
kevinkreiser, line 1223: @genadz can you just explain this a little bit? when we look up stop impact later on, dont we do that by local index? if we skip edges under construction, do we get an off by one index problem when we look up the stop impact later?

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1859,7 +1868,7 @@ struct graph_callback : public OSMPBF::Callback {
 
     // We need to set a data processing flag so we need to
     // process in pbfgraphparser instead of lua because of config option use_rest_area
-    if (use_rest_area_ && service_ == "rest_area") {
+    if (use_rest_area_ && service_ == "rest_area" && way_.use() != Use::kConstruction) {
```
kevinkreiser, line 1871: this is kind of unfortunate, there are several places where we need to detect if its under construction and then if it is we avoid modifying the use or the access. it doesnt seem like there is any way around having to remember to check this for every one. i was going to suggest doing it inside of the setters in the way class so its less spread out over the code but its stil the same amound of checking unfortunately..

valhalla/sif/dynamiccost.h
kevinkreiser: should we really block roads under construction when the person says "ignore_access=true"?

valhalla/sif/dynamiccost.h
kevinkreiser: in my opinion, if the person wants to ignore access they probably want to allow roads under construction

kevinkreiser:
i had a couple of questions, but really the code is pretty much done here and ready to merge. if you have a change to answer those questions i think we can get it merged pretty quickly

lua/graph.lua
genadz: if `rc == nil` here, we will set `rc = 7` in the `if-else` blocks later (line 1418-1419)

src/mjolnir/pbfgraphparser.cc
genadz: agree with you, ideally all logic about modifying `use` or `access` should be concentrated in one place. Right now it's not true. So, let me check if it can be moved into "way_setters"

valhalla/sif/dynamiccost.h
genadz: I can imagine usecases in real life when roads with forbidden access were actually passed, it's physically may happen. In case of roads under construction it can't be true. Another point is that these changes save current behavior for routing.

src/mjolnir/graphenhancer.cc
genadz: this impacts only `left_count` and `right_count` variables that are used to set `edge_to_left` and `edge_to_right` flags. So, this change says even if there is a road under construction between two real edges, it will not be considered to calculate `edge_to_left` and `edge_to_right` flags

valhalla/sif/dynamiccost.h
kevinkreiser: it totally can be true though, especially when OSM hasnt been yet updated to reflect reality. anyway i was only advocating for when `ignore_access == true` then `custruction` roads could be used, which shouldnt break current behavior. for now though i think its fine the way you have it, just to me its something that is not intuitive when you specify `ignore_access` in your request

src/mjolnir/pbfgraphparser.cc
kevinkreiser: yep i came to the same conclusion, i was really just kind of sad that we had to put this all over the place and then remember to put it the next time we start parsing something new that has to do with access/use.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
    * ADDED: isochrone action for /expansion endpoint to track dijkstra expansion [#3215](https://github.com/valhalla/valhalla/pull/3215)
    * CHANGED: remove boost from dependencies and add conan as prep for #3346 [#3459](https://github.com/valhalla/valhalla/pull/3459)
    * CHANGED: Remove boost.program_options in favor of cxxopts header-only lib and use conan to install header-only boost. [#3346](https://github.com/valhalla/valhalla/pull/3346)
+   * ADDED: Add `include_construction` option into the config to include/exclude roads under construction from the graph [#3455](https://github.com/valhalla/valhalla/pull/3455)
```
kevinkreiser, line 50: can you put this at the bottom of the list?

kevinkreiser (approved):
changelog nitpick otherwise good

CHANGELOG.md
genadz: yea, sorry

kevinkreiser (approved):

genadz:
@kevinkreiser could you reapprove ?

genadz:
thanks!

genadz:
thanks!

PR #3460, opened 2021-12-15
top_speed option: ignore live speed for speed based penalties

merkispavel:
# Description
`top_speed` option does 2 things:
- limits edge speed with the given value
- penalizes edges with higher speed

This PR touches penalty logic of the top_speed. In the context of speed based penalties, these changes forces edge speed to be taken from "static" speed layers rather than dynamic ones(current, predictive) 

### Why
Live traffic speed is dynamic and can change quickly. For example, if you only want to use roads below 40kph, a high speed motorway may have big congestion with speeds at 20kph but it is possible that by the time the vehicle reaches the motorway, the traffic clears up and live speed returns back to normal 100kph
In my view, the same logic is applied to predicted speeds as well

### Questions to reviewers
- Does this change make sense to you?
- Are changes consistent with original design of `top_speed` option(if there was any idea behind it)?

## Next steps
If everybody is ok with this changes - I'll apply these changes to other places(other costings) and probably add some tests


nilsnolde:
hm, maybe I'm thinking right about this, also not very familiar with time-dependent astar, but anyways: the changes make total sense to me in case of live traffic. we simply don't know what the speed will be by the time the car actually gets there. but for predicted speed we do know the right speed at the right time during traversal no? to me it seems that's cool to use for speed-based penalization, happy to hear your thoughts:)

src/sif/autocost.cc
```diff
@@ -517,7 +517,15 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
...
-  float speed_penalty = (edge_speed > top_speed_) ? (edge_speed - top_speed_) * 0.05f : 0.0f;
+  float average_historic_speed = edge_speed;
+  // disable dynamic(current/predicted) speed layers and use static
+  // ones(constrained_flow/free_flow/base)
+  if (top_speed_ != kMaxAssumedSpeed && (flow_sources & kDynamicFlowMask)) {
+    // do not pass seconds_of_week not to make speed penalties independent of datetime
+    average_historic_speed = tile->GetSpeed(edge, flow_mask_ & kStaticFlowMask);
+  }
+  float speed_penalty =
+      (average_historic_speed > top_speed_) ? (edge_speed - top_speed_) * 0.05f : 0.0f;
```
genadz, line 528:
```suggestion
      (average_historic_speed > top_speed_) ? (average_historic_speed - top_speed_) * 0.05f : 0.0f;
```
?
otherwise, if `edge_speed < top_speed_` we can get negative speed penalty

genadz:
> hm, maybe I'm not thinking right about this, also not very familiar with time-dependent astar, but anyways: the changes make total sense to me in case of live traffic. we simply don't know what the speed will be by the time the car actually gets there. but for predicted speed we do know the right speed at the right time during traversal no? to me it seems that's cool to use for speed-based penalization, happy to hear your thoughts:)

valid note. it depends what is the goal of `top_speed` parameter. I see 2 situations:
1) we want to penalize roads where actual speed will be higher than `top_speed` - in this case predicted speeds should be taken because we approximately know what speed will be on the edge;
2) we want to penalize roads where actual speed **may be** higher than `top_speed` - here we should take some static data.

nilsnolde:
thanks. so basically option 1 is saying "we trust the speed data completely so please give me the (however accurate) speed at that time and edge".. for option 2 we'd assume (rightfully I guess) that the speeds are inaccurate bcs time will be off due to unexpected live events along the expansion etc and rather take the "safe" approach of the static average speed, constrained or free.

ha, that's more of a philosophical question than I expected :sweat_smile: since in the case where this applies we're not using the original `edge_speed` but the `top_speed_` we really use it to penalize, not for the base cost. still, I think I'd prefer actually option 1, because in cases where the `edge_speed` is lower than `top_speed_` we also pretend to trust the dynamic speed values enough to take it as base cost. so to me that'd be more consistent no? but then raises actually the same conclusion for realtime speeds, ie the whole PR. anyways, just my thoughts, hoping for people with more experience to chime in. it's also not such a huge decision, so sorry for maybe overthinking this..

src/sif/autocost.cc
merkispavel: ah, right. Thanks @genadz

kevinkreiser:
the way i see it was more that it is a matter of the legal speed not the measured speed. if my top speed is 65mph and the speed limit is 55mph, but crazy people are driving 85mph I dont think i should be penalized and suggested to stay off that road. On the other hand if I'm conservative maybe I do want to avoid the crazy people so they don't run me over (an argument I often hear from my pappy).

nilsnolde:
but doesn't the same thought apply to using "measured" (live/historical) speed in general? using it as base cost has the same implications no?

kevinkreiser:
i guess my point is that setting top speed should cap the travel speed to that, so measured speed will be capped when used as a base cost. my understanding was that when a user says my top speed is X, they are telling us I cannot drive faster than X, even if the measure speed is X + 20, i will be driving X.

nilsnolde:
aaah now I got it, yes.. makes sense, sorry for the noise!

merkispavel:
> the way i see it was more that it is a matter of the legal speed not the measured speed
agreed, this is how I think about it as well

> if my top speed is 65mph and the speed limit is 55mph, but crazy people are driving 85mph I dont think i should be penalized and suggested to stay off that road.

makes total sense. Any historic speed source(predicted even predicted one with N mins buckets) should smooth such extreme values(like 85mph on a road with 55mph speed limit)

merkispavel:
Looking at your thoughts, I would suggest to move one speed layer at a time. Let's disable live speeds(i.e current) source for `top_speed` feature and see how it goes
I'll update the PR soon

merkispavel:
I update the PR, folks. It's ready for review

src/sif/autocost.cc
```diff
@@ -517,7 +517,15 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
   }
 
   // TODO: speed_penality hasn't been extensively tested, might alter this in future
-  float speed_penalty = (edge_speed > top_speed_) ? (edge_speed - top_speed_) * 0.05f : 0.0f;
+  float average_edge_speed = edge_speed;
+  // dont use current speed layer for penalties as live speeds might be too low/too high
+  // better to use layers with smoothed/constant speeds
+  if (top_speed_ != kMaxAssumedSpeed && (flow_sources & kCurrentFlowMask)) {
+    average_edge_speed =
+        tile->GetSpeed(edge, flow_mask_ & kNotCurrentFlowMask, time_info.second_of_week);
```
kevinkreiser, line 525: i think this should be:

```suggestion
        tile->GetSpeed(edge, flow_mask_ & (~kCurrentFlowMask), time_info.second_of_week);
```

Otherwise you may change what sources are used rather than those that were specified in the request.

valhalla/baldr/graphconstants.h
```diff
@@ -629,6 +629,7 @@ constexpr uint8_t kFreeFlowMask = 1;
 constexpr uint8_t kConstrainedFlowMask = 2;
 constexpr uint8_t kPredictedFlowMask = 4;
 constexpr uint8_t kCurrentFlowMask = 8;
+constexpr uint8_t kNotCurrentFlowMask = kFreeFlowMask | kConstrainedFlowMask | kPredictedFlowMask;
```
kevinkreiser, line 632: i dont th ink we need this we can just use bitwise operators directly right?

merkispavel: I don't mind. I didn't measure but I hope it won't impact performance

kevinkreiser (dismissed):
this looks fine to me, i suggested a small change to use negation to flip the current flow bit rather than having another constant.

src/sif/autocost.cc
merkispavel: Sure, I'll drop this alias

valhalla/baldr/graphconstants.h
kevinkreiser: bitwise negation of a constant? if the compiler doesnt optimize that then ill be a monkeys uncle!

merkispavel: Oh, damn, indeed.. I forgot that they are constants(constexpr-s). The complier definitely handles such things 👍  Thanks for the catch, @kevinkreiser

kevinkreiser:
Shouldn't we do other motorized costings like truck etc?

merkispavel:
> Shouldn't we do other motorized costings like truck etc?

My bad. Originally I grepped only `autocost.cc` file for `top_speed_` usages.
Now all costings(truck, motor*, etc) have these change. Nice catch @kevinkreiser  🙏

mandeepsandhu:
Some of the closure penalty tests are failing. I remember we penalized closed roads by costing it as a very low-speed road. Looks like that penalty is no longer applied?

test/gurka/test_closure_penalty.cc
```diff
@@ -172,9 +172,6 @@ TEST_P(ClosurePenalty, AvoidClosure) {
             .str();
 
     std::vector<std::string> expected_path = {"QT", "TU", "RU", "OR", "LO", "IL", "FI", "CF"};
-    if (costing == "motor_scooter") {
```
merkispavel, line 175: @mandeepsandhu I've seen a comment about this change in your PR https://github.com/valhalla/valhalla/pull/2964#issuecomment-806217728
I guess this is insignificant(as you noted)

merkispavel:
Tests are finally fixed. All tests pass and the PR ready for the final review.
FYI I added on more commit to extract speed penalty calculation and avoid its duplication across costings. Let me know if this is redundant: it's a standalone commit so it's easy to revert

mandeepsandhu (approved):
lgtm

merkispavel:
Going to merge it, folks. Feel free to leave comments/review if there's something that needs to be fixed/reverted/etc

nilsnolde:
no worries, was waiting for tests passing before approving. good with me, thanks!

kevinkreiser:
yep you covered my comments so i was good thanks

PR #3461, opened 2021-12-15
fix logger issue introduced when migrating to cxxopts

nilsnolde:
in #3459 we migrated all exes to cxxopts lib and in the process it seems I messed up the logger configuration. in some exes that was done too late, after LOG_x was called which basically configures the static logger as `std_out` and ignores whatever was set in the config.

most notably this happened in `valhalla_build_tiles`: https://github.com/valhalla/valhalla/pull/3346#issuecomment-994053989. I checked all other exes and amended the ones which had similar issues.

### DIFF
--- a/src/mjolnir/valhalla_add_predicted_traffic.cc
+++ b/src/mjolnir/valhalla_add_predicted_traffic.cc
@@ -309,6 +309,16 @@ int main(int argc, char** argv) {
     return EXIT_FAILURE;
   }
 
+  // configure logging
+  boost::optional<boost::property_tree::ptree&> logging_subtree =
+      config.get_child_optional("mjolnir.logging");
+  if (logging_subtree) {
+    auto logging_config =
+        valhalla::midgard::ToMap<const boost::property_tree::ptree&,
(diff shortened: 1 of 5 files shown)


src/mjolnir/valhalla_add_predicted_traffic.cc
```diff
@@ -329,16 +339,6 @@ int main(int argc, char** argv) {
...
-  // configure logging
-  boost::optional<boost::property_tree::ptree&> logging_subtree =
-      config.get_child_optional("mjolnir.logging");
-  if (logging_subtree) {
-    auto logging_config =
-        valhalla::midgard::ToMap<const boost::property_tree::ptree&,
-                                 std::unordered_map<std::string, std::string>>(logging_subtree.get());
-    valhalla::midgard::logging::Configure(logging_config);
-  }
-
```
nilsnolde, line 341: really just a bunch of blocks moving around..

kevinkreiser (approved):

kevinkreiser (approved):

PR #3462, opened 2021-12-15
github actions workflow

missinglink:
this `DRAFT` PR is a work-in-progress to *attempt* to move from Circle to GH Actions for CI tasks.
feel free to comment below, the git history will inevitably become a mess which will need to be cleaned up before merging.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,51 @@
+name: Continuous Integration
+on: push
+defaults:
+  run:
+    shell: bash
+jobs:
+  build:
+    runs-on: '${{ matrix.os }}'
+    strategy:
+      matrix:
+        os:
+          - ubuntu-20.04
+    steps:
+      - name: 'checkout'
+        uses: actions/checkout@v2
+      - name: 'lint source'
+        run: |
+          set -euxo pipefail
+          ./scripts/format.sh
+          ./scripts/error_on_dirty.sh
+      - name: 'git submodules'
+        run: |
+          set -euxo pipefail
+          git submodule sync
+          git submodule update --init
+      - name: 'install shared dependencies'
+        run: |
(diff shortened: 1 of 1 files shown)


kevinkreiser:
@missinglink we basically need to supplant all of the required circle checks: `build-osx build-release lint-build-debug build-docker-images`. if you have a look in https://github.com/valhalla/valhalla/blob/master/.circleci/config.yml each of those should be spelled out in there. there are a couple convoluted parts like we need to be using ccache and restoring that cache on subsequent runs (or else they take ages) and probably some other important things im forgetting :smile:

kevinkreiser:
i dont want to be a downer but have you seen this? https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

the best we can get is 2 cores... that will take ages to build anything. i know they have that self hosted thing but i havent looked into it. i havent been able to figure out if you can pay github to have larger machines either.

kevinkreiser:
that self-hosted running thing seems really cool but out of reach at the moment (i dont have a dedicated machine i could donate to the cause at the moment but this is something i always wished for, thank you github/msft). some good reading here: https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners and here: https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow

missinglink:
yeah, I also noticed that it's not easy to do multiarch builds like you would on circle/Travis, again that's available with self hosted runners.

so no ARM macs, no 32 bit etc etc.
this two things combined make it less desirable.

I think I'll tap out now rather than lose time on something which may never eventuate 🙇

kevinkreiser:
@missinglink no worries, thanks for the attempt and great to hear from you!

PR #3463, opened 2021-12-16
Fix: missing heading for intersections

Neyromancer:
Originally only 2 edges were added along the route after the fix any possible number of intersecting edges are added to the final route.

### DIFF
--- a/src/tyr/route_serializer_osrm.cc
+++ b/src/tyr/route_serializer_osrm.cc
@@ -439,13 +439,11 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
   auto intersections = json::array({});
   uint32_t n = arrive_maneuver ? maneuver.end_path_index() + 1 : maneuver.end_path_index();
   EnhancedTripLeg_Node* prev_node = nullptr;
-
   for (uint32_t i = maneuver.begin_path_index(); i < n; i++) {
     auto intersection = json::map({});
 
     // Get the node and current edge from the enhanced trip path
     // NOTE: curr_edge does not exist for the arrive maneuver
-
     auto node = etp->GetEnhancedNode(i);
     auto curr_edge = etp->GetCurrEdge(i);
     auto prev_edge = etp->GetPrevEdge(i);
@@ -545,7 +543,6 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
     // Add the incoming edge except for the first depart intersection.
     // Set routeable to false except for arrive.
     // TODO - what if a true U-turn - need to set it to routeable.
-
     if (i > 0) {
       bool entry = (arrive_maneuver) ? true : false;
       uint32_t prior_heading = prev_edge->end_heading();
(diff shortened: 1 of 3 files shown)


Neyromancer:
Tets to be added

Neyromancer:
Tests added

src/tyr/route_serializer_osrm.cc
```diff
@@ -549,6 +549,14 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
       edges.emplace_back(((prior_heading + 180) % 360), entry, true, false);
     }
 
+    if (i > 0 && !arrive_maneuver) {
+      bool entry = (arrive_maneuver) ? true : false;
```
merkispavel, line 553: 1. `arrive_maneuver` is always equal to `false` in this branch so this variable/ternary operator is redundant
2. The entry is set to false if the intersecting edge is not routable in that direction. I'm not sure that is should be always false..

src/tyr/route_serializer_osrm.cc
```diff
@@ -549,6 +549,14 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
...
+      for (uint32_t m = 0; m < node->intersecting_edge_size(); m++) {
+        auto intersecting_edge = node->GetIntersectingEdge(m);
+        edges.emplace_back(intersecting_edge->begin_heading(), entry, true, false);
```
merkispavel, line 556: as these are outgoing edges - boolean flags should be set vice versa, nope?
```suggestion
        edges.emplace_back(intersecting_edge->begin_heading(), entry, false, true);
```

src/tyr/route_serializer_osrm.cc
```diff
@@ -549,6 +549,14 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
       edges.emplace_back(((prior_heading + 180) % 360), entry, true, false);
     }
 
+    if (i > 0 && !arrive_maneuver) {
```
merkispavel, line 552: I would at it to the code where "main" outgoing edge is added
```cpp
// Add the edge departing the node
    if (!arrive_maneuver) {
      edges.emplace_back(curr_edge->begin_heading(), true, false, true);
      // add other outgoing edges
      ...
    }
```

merkispavel (changes requested):

src/tyr/route_serializer_osrm.cc
Neyromancer: the logic here is different, we are adding intersecting edges that are not starting or not ending, only in the middle of the road.

src/tyr/route_serializer_osrm.cc
```diff
@@ -532,16 +532,23 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
...
+    // the arrival step.
     std::vector<IntersectionEdges> edges;
 
     // Add the edge departing the node
     if (!arrive_maneuver) {
       edges.emplace_back(curr_edge->begin_heading(), true, false, true);
+      if (i > 0) {
+        for (uint32_t m = 0; m < node->intersecting_edge_size(); m++) {
+          auto intersecting_edge = node->GetIntersectingEdge(m);
+          bool routable = intersecting_edge->IsTraversableOutbound(curr_edge->travel_mode());
```
merkispavel, line 544: wow, what a nice method. Exactly what we need!

merkispavel (approved):
Nice, looks good

merkispavel (approved):
Re approve after merging master with conflicts

Issue #3466, opened 2021-12-19
proto3-related bug for HTTP services


nilsnolde:
first reported in #3465 , opening new here for clarity.

#3457 seems to have introduce a kinda subtle bug: when running valhalla_service in HTTP mode there's an exception triggered for _any_ `/route` and `/optimized_route` request: `[libprotobuf ERROR google/protobuf/wire_format_lite.cc:577] String field 'valhalla.TaggedValue.value' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes`

however, one-shot mode works fine, so the problem might be related with serializing the pbf for prime_server somehow? in valhalla a HTTP request fails here: https://github.com/valhalla/valhalla/blob/ce94c7b2266d804024ce423ff2738a2233b1d2f6/src/odin/worker.cc#L77-L81

so I guess either loki or thor fails to fill the pbf properly? I couldn't find the problem so far, not easy to debug with prime_server in the middle..

nilsnolde:
fixed in #3468

PR #3468, opened 2021-12-21
fixes #3465 #3466

### LINKED ISSUE #3465: String field 'valhalla.TaggedValue.value' contains invalid UTF-8 data when parsing a protocol buffer.
I'm trying to use the optimized_route with the use of the docker image.

However, I get an error when trying to use it:
```String field 'valhalla.TaggedValue.value' contains invalid UTF-8 data when parsing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes. ```

The post returns with the following message:
```{"error_code":299,"error":"Unknown:Failed to parse intermediate request format:Failed parsing pbf in Odin::Worker","status_code":400,"status":"Bad Request"}```

I'm posting to http://localhost:8002/optimized_route with:
```{"locations":[{"lat":52.1,"lon":5.1,"name":"Carlos"},{"lat":52.0,"lon":5.0,"name":"Cato"}],"costing":"auto","directions_options":{"units":"kilometers"}}```

I'm using this docker-compose:
```
version: '3.0'
services:
  valhalla:
    image: gisops/valhalla:latest
    container_name: valhalla_latest
    ports:
      - 8002:8002
    #build:
    #  context: .
    #  args:
    #    - VALHALLA_UID=1000
    #    - VALHALLA_GID=1000
    volumes:
      - ./custom_files/:/custom_files
    environment:
      # Auto-download PBFs from Geofabrik
      - tile_urls=https://download.geofabrik.de/europe/netherlands-latest.osm.pbf
      - server_threads=2  # determines how many threads will be used to run the valhalla server
      - use_tiles_ignore_pbf=False  # load existing valhalla_tiles.tar directly
      - build_elevation=False  # build elevation with "True" or "Force": will download only the elevation for areas covered by the graph tiles
      - build_admins=False  # build admins db with "True" or "Force"
      - build_time_zones=False  # build timezone db with "True" or "Force"
      - build_tar=True  # build an indexed tar file from the tile_dir for faster graph loading times
      - force_rebuild=False  # forces a rebuild of the routing tiles with "True"
      # - path_extension=graphs  # this path will be internally appended to /custom_files; no leading or trailing path separator!
    # adapt to the expected build time
    # healthcheck: 
    #   test: curl --fail -s http://localhost:8002/status || exit 1
    #   interval: 1s
    #   retries: 10
    #   start_period: 2s
    #   timeout: 1s
```
This is the log of my output:
https://gist.github.com/CvRXX/76c6f7e13c28fbbd09f539315f655117

The current files in costume tiles are:
```
drwxrwxr-x 1 root   root    206 Dec 18 19:01 .
drwxr-xr-x 1 carlos carlos  128 Dec 18 18:45 ..
-rw-rw-r-- 1 root   root    272 Dec 18 18:53 duplicateways.txt
-rw-rw-r-- 1 root   root     65 Dec 18 19:01 file_hashes.txt
-rw-rw-r-- 1 root   root   1.1G Dec 18 18:46 netherlands-latest.osm.pbf
-rw-rw-r-- 1 root   root   6.8K Dec 18 18:46 valhalla.json
drwxrwxr-x 1 root   root      6 Dec 18 19:01 valhalla_tiles
-rw-rw-r-- 1 root   root   627M Dec 18 19:01 valhalla_tiles.tar
```

kevinkreiser:
# what was going wrong?

so when going from proto2 to proto3 we get a new scalar type called `bytes`. this is useful in languages like java and python which want to make assertions about the type of data that is in a string vs a byte array. in c++ we roll our own so the apis are exactly the same regardless.

we have this concept of tagged values which are arbitrary data which we can attach to any edge via its edge info. the tag usually refers to the tag in osm and the value is any payload we want to attach. the first types of these were strings and were additional names for things like tunnels and bridges. but then we added other stuff like `layer` which is an integer and `pronunciation` which is a custom c++ struct and finally also bike share stations which are actually serialized protobufs.

because proto3 doesnt want c++ to send non utf8 binary stuff as a pythong `str` or java `String` and have the person on the other end in those languages get screwed over, they have implemented checks in serialization/deserialization to check that even if we dont differentiate in c++.

# how to fix

there were two ways to fix this. the first way would be to not put any non textual things in the tagged value. indeed we are currently only using tunnels for the osrm serializer. so in `triplegbuilder` we could have filtered the binary ones out. the thing is though that then if you called trace attributes and you wanted all that data you would only get the filtered list. so instead of that i've just changed the type to bytes so we can keep putting all the values in there and the consumer on the other end can know how and whether or not they want to use the data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
    * FIXED: Use midgard::unaligned_read in GraphTileBuilder::AddSigns [#3456](https://github.com/valhalla/valhalla/pull/3456)
    * FIXED: Relax test margin for time dependent traffic test [#3467](https://github.com/valhalla/valhalla/pull/3467)
    * FIXED: Fixed missed intersection heading [#3463](https://github.com/valhalla/valhalla/pull/3463)
+   * FIXED: Stopped putting binary bytes into a string field of the protobuf TaggedValue since proto3 protects against that for cross language support [#3468](https://github.com/valhalla/valhalla/pull/3468)
 
 * **Enhancement**
    * CHANGED: Pronunciation for names and destinations [#3132](https://github.com/valhalla/valhalla/pull/3132)
--- a/proto/tripcommon.proto
+++ b/proto/tripcommon.proto
@@ -368,17 +368,17 @@ enum RoadClass {
 }
 
 message TaggedValue {
+  // dont renumber these they match the c++ definitions
   enum Type {
     kNone = 0;
     kLayer = 1;
+    kPronunciation = 2;
     kBssInfo = 3;
-    // we used to have bug when we encoded 1 and 2 as their ASCII codes, but not actual 1 and 2 values
-    // see https://github.com/valhalla/valhalla/issues/3262
-    kTunnel = 49; // static_cast<uint8_t>('1')
-    kBridge = 50; // static_cast<uint8_t>('2')
+    kTunnel = 49;
+    kBridge = 50;    
   }
   oneof has_value {
-    string value = 1;   // The actual tagged name value, examples: Ted Williams Tunnel
+    bytes value = 1;    // The actual tagged name value, examples: Ted Williams Tunnel
   }
   oneof has_type {
     Type type = 2;      // The type of tagged name (tunnel or bridge)


nilsnolde (approved):
ay, what an easy fix.. thanks!

brad-4k:
When might we see this in the valhalla:latest docker image?  Thanks!

kevinkreiser:
@brad-4k its merged to master so its already there

brad-4k:
Is there a way to get a version from it to verify?  I pulled yesterday and still getting the error.  Trying to update (I think) right now:
`$ docker pull docker.pkg.github.com/gis-ops/docker-valhalla/valhalla:latest
latest: Pulling from gis-ops/docker-valhalla/valhalla
Digest: sha256:17c101c48ae1b5fbfc4d1633e6b34ff99bf95e88ad2b24207830b974c526b671
Status: Image is up to date for docker.pkg.github.com/gis-ops/docker-valhalla/valhalla:latest`

I still get:

`...
end raw bytes. 
[libprotobuf ERROR google/protobuf/wire_format_lite.cc:626] String field 'valhalla.TaggedValue.value' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes. 
2021/12/23 22:13:04.209036 [INFO] Got Odin Request 0
[libprotobuf ERROR google/protobuf/wire_format_lite.cc:626] String field 'valhalla.TaggedValue.value' contains invalid UTF-8 data when parsing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes. 
2021/12/23 22:13:04.209405 [ERROR] Failed parsing pbf in Odin::Worker
0 2021/12/23 22:13:04`

kevinkreiser:
sorry this repository is not connected to the gis-ops one. that is a downstream fork that @nilsnolde manages. im not 100% sure of the additions there but i would assume they are the ones that setup service configuration and starting of the service. if you have questions about that please lets keep them over in that repository: https://github.com/gis-ops/docker-valhalla

brad-4k:
Oh, right!  Sorry!  In the famous words of Homer Simpson, DOH!

kevinkreiser:
No worries, an honest mistake!

nilsnolde:
we rebuild and push a new image every saturday morning (it's a cron) as `latest`. that's the periodic reminder to stay away from `latest` images or `master/main` code if you depend on it working :)

PR #3471, opened 2021-12-24
Reuse sample::get implementation

merkispavel:
# Issue

Previously we had duplicated logic for `sample::get` and `sample::get_all`. The PR
- removes the duplication by adding helper method `get(coord, tile&)`(like we have in [GraphReader](https://github.com/valhalla/valhalla/blob/ba50b993cc02be4e610f0231a71a40b3153b936c/valhalla/baldr/graphreader.h#L482))
- minor: sets default `tile_data` index to invalid value(i.e `TILE_COUNT` instead of 0)

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -458,6 +458,7 @@ template <class coord_t> double sample::get(const coord_t& coord, tile_data& til
   auto lat = std::floor(coord.second);
   auto index = static_cast<uint16_t>(lat + 90) * 360 + static_cast<uint16_t>(lon + 180);
 
+  // the caller can pass a cached tile, so we only fetch one if its not the one they already have
   if (index != tile.get_index()) {
     // get the proper source of the data
     tile = cache_->source(index);
--- a/src/skadi/sample.cc
+++ b/src/skadi/sample.cc
@@ -212,7 +212,7 @@ private:
   bool reusable;
 
 public:
-  tile_data() : c(nullptr), index(0), reusable(false), data(nullptr) {
+  tile_data() : c(nullptr), index(TILE_COUNT), reusable(false), data(nullptr) {
   }
 
   tile_data(const tile_data& other) : c(nullptr) {
@@ -234,6 +234,10 @@ public:
 
   inline explicit operator bool() const {
     return data != nullptr;
+  }
+
+  uint16_t get_index() const {
+    return index;
   }
 
   double get(double u, double v) const {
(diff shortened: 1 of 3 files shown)


src/skadi/sample.cc
```diff
@@ -212,7 +212,7 @@ class tile_data {
   bool reusable;
 
 public:
-  tile_data() : c(nullptr), index(0), reusable(false), data(nullptr) {
+  tile_data() : c(nullptr), index(TILE_COUNT), reusable(false), data(nullptr) {
```
merkispavel, line 215: If I'm not mistaken - index=0 is actually a valid index. So it's better to fill it with invalid index value

src/skadi/sample.cc
```diff
@@ -470,29 +476,18 @@ template <class coord_t> double sample::get(const coord_t& coord) {
...
 
+template <class coord_t> double sample::get(const coord_t& coord) {
+  tile_data tile;
+  return get(coord, tile);
+}
+
 template <class coords_t> std::vector<double> sample::get_all(const coords_t& coords) {
   std::vector<double> values;
   values.reserve(coords.size());
-  uint16_t curIndex = TILE_COUNT;
```
merkispavel, line 476: `tile_data` already stores index so I just added `tile_data::get_index` getter to obtain its index

src/skadi/sample.cc
```diff
@@ -448,14 +452,16 @@ sample::~sample() {
...
-template <class coord_t> double sample::get(const coord_t& coord) {
+template <class coord_t> double sample::get(const coord_t& coord, tile_data& tile) {
   // check the cache and load
   auto lon = std::floor(coord.first);
   auto lat = std::floor(coord.second);
   auto index = static_cast<uint16_t>(lat + 90) * 360 + static_cast<uint16_t>(lon + 180);
 
-  // get the proper source of the data
-  const auto& tile = cache_->source(index);
+  if (index != tile.get_index()) {
```
kevinkreiser, line 461: this is a bit tricky. you send an unitialized `tile_data` object to the function and it detects that its not initialized by looking at the index and seeing its wrong? shouldnt we check something else that is more obvious? maybe we could add a method to `bool tile_data::initilized() const { return index != TILE_COUNT; }` or what would be more obvious to me would be: `bool tile_data::initilized() const { return data != nullptr; }`

i just find this `if` with no comment very tricky because you have to figure out why the heck its checking the index

kevinkreiser: ok so now i see why you are doing the trick with the index after i looked at the `get_all` change. caching. then i think the best thing to do here would be to add a comment above the `if` that says something like:

```suggestion
  // if the caller passed in a tile they were already working with we only fetch it if this coordinate needs a different tile
  if (index != tile.get_index()) {
```

kevinkreiser (dismissed):
this looks good to me, do you mind adding that extra comment for clarity?

merkispavel:
> this looks good to me, do you mind adding that extra comment for clarity?

Sure thing! done

kevinkreiser (approved):

PR #3481, opened 2021-12-29
switch to loki logging config for valhalla_service

### LINKED ISSUE #3479: Changing log configuration requires full tile rebuild
Currently it is not possible to simply switch on/off color encoding of log-level without thrashing the tiles. Ideally the log color attribute is read from the JSON-config file every time valhalla starts up.

nilsnolde:
fixes #3479 

valhalla_service was using `tyr.logging` to configure logging, which is not being exported anymore by the config generator. switched to loki.logging instead.


kevinkreiser (approved):
oh man good catch, i totally missed that when i refactored tyr away

PR #3486, opened 2022-01-03
potential tilebuild speedup

nilsnolde:
this is just a temp PR for easier reviewing. not at all complete!


src/mjolnir/graphbuilder.cc
```diff
@@ -452,10 +458,19 @@ void BuildTileSet(const std::string& ways_file,
...
   ////////////////////////////////////////////////////////////////////////////
   // Iterate over tiles
-  for (; tile_start != tile_end; ++tile_start) {
+  TileDef current_tile;
+  while (true) {
     try {
+      if (tilequeue.empty()) {
+        break;
+      }
+      lock.lock();
```
nilsnolde, line 467: move above the if

PR #3490, opened 2022-01-04
Trying newer xcode

kevinkreiser:
Circle is deprecating xcode 11 so we need to upgrade

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   macos:
     macos:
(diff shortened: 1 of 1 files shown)


merkispavel:
@kevinkreiser I tried various xcode versions and the newest ones(with macos 11) doesn't work out of the box(check my PR description https://github.com/valhalla/valhalla/pull/3492/)
So I would merge `xcode 12.4.0` update not to be worry about xcode deprecation in the CircleCI https://github.com/valhalla/valhalla/pull/3492

PR #3502, opened 2022-01-07
valhalla.h and config.h don't need cmake configuration (anymore)

nilsnolde:
must've been a relic from older configs that those two headers were passed through cmake config. at least now it's not needed anymore and those headers can stay in the source dir, no need to copy them over to build dir.


valhalla/config.h
```diff
@@ -6,7 +6,9 @@
 #define VALHALLA_STRINGIZE(A) VALHALLA_STRINGIZE_NX(A)
 
 /* Version number of package */
-#define VALHALLA_VERSION VALHALLA_STRINGIZE(VALHALLA_VERSION_MAJOR) "." VALHALLA_STRINGIZE(VALHALLA_VERSION_MINOR) "." VALHALLA_STRINGIZE(VALHALLA_VERSION_PATCH)
+#define VALHALLA_VERSION                                                                             \
+  VALHALLA_STRINGIZE(VALHALLA_VERSION_MAJOR)                                                         \
+  "." VALHALLA_STRINGIZE(VALHALLA_VERSION_MINOR) "." VALHALLA_STRINGIZE(VALHALLA_VERSION_PATCH)
```
nilsnolde, line 11: urgh, didn't realize that.. I'll turn off clang format for that one

valhalla/valhalla.h
```diff
@@ -3,4 +3,3 @@
 #define VALHALLA_VERSION_MAJOR 3
```
kevinkreiser, line 3: once upon a time i had a bunch of scripting to do autotools style debian packaging for valhalla. it was great, we had debian packages on ppa. then we migrated to cmake and all that stuff didnt work but we kept the template headers

at this point you are right we dont need them anymore. actually i think we should drop `config.h` completely lets just move the packaging stuff into `valhalla.h` as well and simplify all the way. what do you think @nilsnolde ?

kevinkreiser: ah wait i guess we cant move PACKAGE and PACKAGE_VERSION they need to stay in config.h as they would collide in other projects. we should also check that cmake is getting these variables properly for its packaging targets

nilsnolde: the version variables make it into the main cmakelists.txt

kevinkreiser (approved):

PR #3506, opened 2022-01-10
Refactor Costing and PathEdge Protos

kevinkreiser:
in #3464 we added support for interacting with the service via protobuf. at the time we noted several items for future work. this pr closes out the "refactor" item.

> **refactor the options message (separate PR):** the request options message was incepted before we intended to use it as a request format. Because of that it's become pretty crufty and we should take a hard look at it before taking this feature out of beta.

the specific refactoring that was done here is all contained to the "request" (ie options object). inside of the options object there are several output items, which is a bit strange as this is the request object we use to start the service interaction going. there are mainly two places where we actually store output which we would not expect the user to provide. those two places are the ones listed in the title, the costing object and the path edges that loki finds. let me summarize the changes.

**PathEdge**
Loki takes your input locations and finds candidate edges in the graph that we can use in the algorithms. We currently store this information in the `Location` object which is also where the requester specifies the lat lng etc. In addition to the path edges there are a few more bits of output info about where the location correlated to the graph. Instead of having these all directly in the location object I nested them into a new object called correlation and marked it so that its obvious that its output. I could have done something more radical and moved it completely out of the options object but this didnt feel necessary.

**Costing**
Currently we have a `Costing` enum which must be specified as well as `CostingOptions` which are optional. At the moment costing options has both inputs and outputs. The refactor puts all of the proper costing options into their own object and nests that inside of a costing object. In this way we can have the higher level non optional bits of a costing object as well as the output bits separate from the actual user supplied options. ive also changed the naming a bit so that the enum is called `Costing::Type` and costing options are now `Costing::Options`.

Thats it. There is no functional change to the code in anyway its all moves and renames to hopefully make the api a bit easier to work with in pbf mode.


nilsnolde (approved):
just looked at the structural changes of protobuf and looks reasonable to me!

dgearhart (approved):
LGTM :ship:

Issue #3469, opened 2021-12-22
Indoor routing support proposal


liyinxiao:
Pedestrians should be able to navigate through indoor elements such as corridors, elevators, stairs, escalators, and/or building entrances. OSM already supports indoor mapping, although adoption has not been very high. I am writing this proposal to see if there's any interest in supporting pedestrian indoor routing in valhalla. Briefly, here are a few potential use cases: 
- Corridors
    - Pedestrians can walk through corridors
- Elevator/stairs/escalators
    - Pedestrians can take elevator/stairs/escalators to a different level
- Building Entrance
    - Pedestrian can enter/exit buildings through building entrances

![image](https://user-images.githubusercontent.com/37345111/147153056-dab79fa5-ebef-455a-9774-29e4f3564116.png)
*An example of indoor routing: this route goes through a mall (Prudential center, Boston), a footbridge, and a second mall (Copley Place, Boston).*

### Proposed data model of indoor elements
OSM has a few tagging schemas for indoor mapping, including [Simple Indoor Tagging (SIT)](https://wiki.openstreetmap.org/wiki/Simple_Indoor_Tagging) as well as [indoor tagging schemes](https://wiki.openstreetmap.org/wiki/Indoor_Mapping#Tagging_proposals). To build routing graphs, SIT polygons are probably not the best fit, and I am proposing this list of indoor elements as a starting point for valhalla routing service to support:

- [Corridors](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dcorridor)
    - Way: tag:highway=corridor, level=* (or level:ref=*)
    - Node: level=* (or level:ref=*)
- [Elevator](https://wiki.openstreetmap.org/wiki/Tag:highway%3Delevator)
    - Node: tag:highway=elevator, level=* (or level:ref=*)
- [Stairs](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dsteps)
    - Way:  tag:highway=steps, level=* (or level:ref=*), indoor=yes
    - Node: level=* (or level:ref=*), indoor=yes
- [Escalators](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dsteps)
    - Way:  tag:highway=steps, conveying=* , level=* (or level:ref=*), indoor=yes
    - Node: level=* (or level:ref=*), indoor=yes
- [Building Entrance](https://wiki.openstreetmap.org/wiki/Key:entrance) 
[...]

PR #3509, opened 2022-01-13
indoor routing - data model, data processing

liyinxiao:
# Issue

https://github.com/valhalla/valhalla/issues/3469

In this PR, we make the changes to data processing and the data model, basically to get data into the tiles. The indoor elements are routable but all of the maneuvers will be of type "DirectionsLeg_Maneuver_Type_kContinue". Maneuver and narration support (with tests) can be added in another PR. 

A few key ideas:
- [Elevator](https://wiki.openstreetmap.org/wiki/Tag:highway%3Delevator) can be modeled as both a node (more common) and a way according to OSM Wiki, and we aim to support both. Alternatively, we can process nodes to create additional/fake ways during parsing stage and only support ways (@kevinkreiser). Use an elevator_penalty to account for the time to wait. 
- [Level](https://wiki.openstreetmap.org/wiki/Key:level) is stored as a string, to account for edges crossing multiple levels, for future use cases (such as level_filter in loki).
  - There are 1388 distinct values of levels for nodes, and 1380 distinct values for ways on OSM today, including fractional values such as 3.6 and invalid values such as GF. Levels don't always increment by 0.5.
- Prefer level:ref over level for narration. For example, prefer "take the elevator to Ground Level" to "take the elevator to Level 1", if [level:ref](https://wiki.openstreetmap.org/wiki/Key:level:ref) exists. 
- Add indoor_ field to check EnterBuilding vs. ExitBuilding. 

Test plan
```
cd build
mkdir -p valhalla_tiles
./valhalla_build_config --mjolnir-tile-dir ${PWD}/valhalla_tiles --mjolnir-tile-extract ${PWD}/valhalla_tiles.tar --mjolnir-timezone ${PWD}/valhalla_tiles/timezones.sqlite --mjolnir-admin ${PWD}/valhalla_tiles/admins.sqlite > valhalla.json
./valhalla_build_tiles -c valhalla.json ~/Downloads/massachusetts-latest.osm.pbf
find valhalla_tiles | sort -n | tar cf valhalla_tiles.tar --no-recursion -T -
./valhalla_service valhalla.json 1
curl http://localhost:8002/route --data '{"locations":[{"lat":42.361145,"lon":-71.0589,"type":"break","city":"Boston"},{"lat":42.358853,"lon":-71.057298,"type":"break","city":"Boston"}],"costing":"pedestrian","directions_options":{"units":"miles"}}' | jq '.'
```

## Tasklist

 - [x] Add tests

## Requirements / Relations


dnesbitt61 (dismissed):

kevinkreiser (dismissed):
It would be great to add some gurka unit tests which prove the edges and nodes are getting parsed properly with respect to their tags and values. it doesnt have to be in depth or anything just build the data in gurka and then assert that the properties on the nodes, edges and edgeinfos are set properly. looks great so far!

liyinxiao:
Thanks @dnesbitt61, @kevinkreiser for the fast feedback. I will add a few gurka tests, to make sure each indoor element gets parsed properly.

liyinxiao:
Adding gurka tests: 
```
make run-gurka_building_entrance
make run-gurka_elevator
make run-gurka_steps
make run-gurka_escalator
make run-gurka_levels
```

kevinkreiser:
@liyinxiao would you mind putting all of the tests into the same test file. i think it would be slightly more tractable if they were all together in an indoor routing gurka test suite rather than in separate files. you could even build one map with all the features you want to test rather than a different map for each one. since we already have over 100 gurka tests its starting to become difficult to find a particular test :smile:

liyinxiao:
Good idea! Re-running gruka tests:
```
make run-gurka_indoor
[100%] Built target gurka_indoor
[100%] Generating gurka_indoor.log
[PASS] gurka_indoor
[100%] Built target run-gurka_indoor
```

kevinkreiser:
@liyinxiao looks like you need to run `scripts/format.sh` to lint the code as well

liyinxiao:
CI tests failed. I looked into it, and when I run `VERBOSE=1 make -C build check` or `make run-scripts` on my macbook it gives me this error:
```
yinxiaoli@yinxiaoli-mbp build % VERBOSE=1 make -C build check   
[ 23%] Generating scripts.log
LOCPATH=/Users/yinxiaoli/Desktop/tmp/valhalla/locales ASAN_OPTIONS=verify_asan_link_order=0:detect_leaks=0 /bin/bash -cx "/opt/homebrew/Frameworks/Python.framework/Versions/3.8/bin/python3.8 -m unittest discover -s /Users/yinxiaoli/Desktop/tmp/valhalla/test/scripts -v > /Users/yinxiaoli/Desktop/tmp/valhalla/build/test/scripts.log 2>&1"
+ /opt/homebrew/Frameworks/Python.framework/Versions/3.8/bin/python3.8 -m unittest discover -s /Users/yinxiaoli/Desktop/tmp/valhalla/test/scripts -v
make[3]: *** [test/scripts/scripts.log] Error 1
make[2]: *** [test/scripts/CMakeFiles/run-scripts.dir/all] Error 2
make[1]: *** [test/CMakeFiles/check.dir/rule] Error 2
make: *** [check] Error 2
```
This error occurs with or without my commits. Any suggestions around this error will be appreciated.

kevinkreiser:
Ah yes, we semi-recently changed the python unit tests and unfortunately it seems we forgot to prints out the test logs when the test fails so you can tell what the heck is going. in this case just manually running `VERBOSE=1` tells me:

    LOCPATH=/home/kkreiser/sandbox/valhalla/locales python3 -m unittest discover -s /home/kkreiser/sandbox/valhalla/test/scripts -v
    
shows:

```
test_add_leaf_types (test_valhalla_build_config.TestBuildConfig) ... ok
test_nested_config (test_valhalla_build_config.TestBuildConfig)
tests if we get the values of nested dicts ... ok
test_override_config (test_valhalla_build_config.TestBuildConfig)
tests end to end if the arg parsing is working ... ok
test_get_tiles_with_bbox (test_valhalla_build_elevation.TestBuildElevation) ... ok
test_get_tiles_with_geojson (test_valhalla_build_elevation.TestBuildElevation) ... ok
test_get_tiles_with_graph (test_valhalla_build_elevation.TestBuildElevation) ... ok
test_create_extracts (test_valhalla_build_extract.TestBuildExtract) ... FAIL

======================================================================
FAIL: test_create_extracts (test_valhalla_build_extract.TestBuildExtract)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kkreiser/sandbox/valhalla/test/scripts/test_valhalla_build_extract.py", line 26, in test_create_extracts
    self.check_tar(EXTRACT_PATH, exp_tuples, tile_count * INDEX_BIN_SIZE)
  File "/home/kkreiser/sandbox/valhalla/test/scripts/test_valhalla_build_extract.py", line 36, in check_tar
    self.assertIn(t, exp_tuples)
AssertionError: (960512, 6549282, 6059792) not found in ((2560, 25568, 291912), (296448, 410441, 662496), (960512, 6549282, 6059608))

----------------------------------------------------------------------
Ran 7 tests in 0.071s

FAILED (failures=1)
```

this is basically saying the tile sizes have changed, which is completely expected in this case. the cool thing is that it means our canned Utrecht osm pbf data must have some ways tagged in such a way that you are parsing them out. unfortunately this makes for a bit of a brittle test. any time we change the data we'll get failures here. i think its ok though we simply update the numbers when we expect them to change.

in this case that first test was `(960512, 6549282, 6059608)` but should now be `(960512, 6549282, 6059792)` so it looks like the tile is bigger by ~200 bytes. so yeah id just upate the values in ` test/scripts/test_valhalla_build_extract.py`

liyinxiao:
Thanks, it worked! Yeah, your utrecht_netherlands.osm.pbf contains 128 OSM ways with level tag, which are going into the tiles.

test/gurka/gurka.cc
```diff
@@ -571,6 +571,41 @@ findEdgeByNodes(valhalla::baldr::GraphReader& reader,
...
+      }
+      // Check begin node
+      auto edge_id = tile_id;
+      edge_id.set_id(&e - tile->directededge(0));
+      auto begin_node_id = reader.edge_startnode(edge_id);
+      ll = tile->get_node_ll(begin_node_id);
+      if (ll.ApproximatelyEqual(nodes.at(node_name))) {
+        return begin_node_id;
+      }
+    }
```
kevinkreiser, line 604: you can just check the nodes directly:

```suggestion
    auto node_id = tile_id;
    for (auto node_id = tile_id; node_id.id() < tile->header()->nodecount(); ++node_id) {
      auto ll = tile->get_node_ll(node_id);
      if (ll.ApproximatelyEqual(nodes.at(node_name))) {
        return node_id;
      }
    }
```

src/sif/pedestriancost.cc
```diff
@@ -42,6 +42,9 @@ constexpr float kDefaultSpeedWheelchair = 4.0f; // 2.5  MPH  TODO
 constexpr float kDefaultStepPenaltyFoot = 30.0f;        // 30 seconds
 constexpr float kDefaultStepPenaltyWheelchair = 600.0f; // 10 minutes
 
+// Penalty to take elevator
+constexpr float kDefaultElevatorPenalty = 5.0f; // 5 seconds
```
kevinkreiser, line 46: this seems quite cheap!

on a serious note though, if we want to keep the changes to costing in this pr we should have a test that exorcises them in your test suite. do you mind adding a test? maybe you want to set a super high penalty in the request so that a person takes the stairs instead of the elevator and your test can just show  you took the stairs when you set the penalty but you take the elevator otherwise? or did you want to save all that for another pr?

proto/tripcommon.proto
```diff
@@ -376,6 +376,8 @@ message TaggedValue {
     kLayer = 1;
     kPronunciation = 2;
     kBssInfo = 3;
+    kLevel = 4;
```
kevinkreiser, line 379: this file got renamed to `common.proto` in a recent pr i merged. do you mind merging master into this branch or do you want me to do it?

src/sif/pedestriancost.cc
liyinxiao: Are you talking about adding to my gurka test or a different test? It will be great if you can help with the tests. If you feel 5s is too cheap, I'm fine with whatever number you feel is more reasonable. Another option (as you said) is to save them for my next PR for maneuvers - which I can happily take. Let me know what you would prefer!

proto/tripcommon.proto
liyinxiao: merged

src/sif/pedestriancost.cc
kevinkreiser: oh yeah i would just add it to the existing tests, that way we can keep all the indoor routing tests right there, data, and algorithm righ tnext to each other and easy to find.

regarding the number im fine with 5 seconds for now, it just seemed a little small but i was really just kidding around. and yeah i can happily make a simple test in there!

liyinxiao: Hi @kevinkreiser, I propose to move all the changes (and tests) in pedestriancost.cc to a second PR, so that we can merge this PR first, if this test is the only blocker at this point. In this way, you can take your time with the tests while I am working on another PR for indoor routing maneuvers/narrations. What do you think?

kevinkreiser: haha im literally writing the test right now! i'll put another hour into it but if it doesnt finish after that then i agree. yes that is the only blocker at hte moment. i just need to give it a final review after adding the test. sorry for the delay hard to find the time these past couple of nights

kevinkreiser: ah good thing i wrote these tests, i caught a small typo that messed up the access for the edges and nodes! ill push it up shortly!

kevinkreiser: ah it seems you have your fork configured such that i cannot push to it. would you mind changing that at least temporarily?

kevinkreiser: oh it seem the github gui will let do it that way. since its only 2 files ill just do that. one sec!

kevinkreiser: ok just pushed!

lua/graph.lua
```diff
@@ -30,8 +30,8 @@ highway = {
...
-["corridor"] =          {["auto_forward"] = "false", ["truck_forward"] = "false", ["bus_forward"] = "false", ["taxi_forward"] = "false", ["moped_forward"] = "false", ["motorcycle_forward"] = "false", ["pedestrian"] = "true",  ["bike_forward"] = "false"},
-["elevator"] =          {["auto_forward"] = "false", ["truck_forward"] = "false", ["bus_forward"] = "false", ["taxi_forward"] = "false", ["moped_forward"] = "false", ["motorcycle_forward"] = "false", ["pedestrian"] = "true",  ["bike_forward"] = "false"}
+["corridor"] =          {["auto_forward"] = "false", ["truck_forward"] = "false", ["bus_forward"] = "false", ["taxi_forward"] = "false", ["moped_forward"] = "false", ["motorcycle_forward"] = "false", ["pedestrian_forward"] = "true",  ["bike_forward"] = "false"},
+["elevator"] =          {["auto_forward"] = "false", ["truck_forward"] = "false", ["bus_forward"] = "false", ["taxi_forward"] = "false", ["moped_forward"] = "false", ["motorcycle_forward"] = "false", ["pedestrian_forward"] = "true",  ["bike_forward"] = "false"}
```
kevinkreiser, line 34: there was just an unfortunate typo here, we needed to set `pedestrian_forward` rather than just `pedestrian` then later on we derive `pedestrian_backward` from forward unless the way is a one way only footway (like an escalator)

test/gurka/test_indoor.cc
```diff
@@ -51,6 +51,10 @@ class Indoor : public ::testing::Test {
         {"GH", {{"highway", "elevator"}, {"indoor", "yes"}, {"level", "1;2"}}},
         {"HI", {{"highway", "corridor"}, {"indoor", "yes"}, {"level", "2"}}},
         {"IJ", {{"highway", "corridor"}, {"indoor", "yes"}, {"level", "3"}}},
+
+        {"Cx", {{"highway", "steps"}, {"indoor", "yes"}, {"level", "1;2"}}},
+        {"xy", {{"highway", "steps"}, {"indoor", "yes"}, {"level", "2;3"}}},
+        {"yJ", {{"highway", "steps"}, {"indoor", "yes"}, {"level", "3"}}},
```
kevinkreiser, line 57: i added an alternate path here, you can take the steps rather than use the elevator to get to the 3rd floor

test/gurka/test_indoor.cc
```diff
@@ -66,37 +70,49 @@ gurka::map Indoor::map = {};
...
 
   directededge = std::get<1>(gurka::findEdgeByNodes(graphreader, layout, "D", "E"));
   EXPECT_EQ(directededge->indoor(), false);
+  EXPECT_TRUE(directededge->forwardaccess() & baldr::kPedestrianAccess);
+  EXPECT_TRUE(directededge->reverseaccess() & baldr::kPedestrianAccess);
+
   directededge = std::get<1>(gurka::findEdgeByNodes(graphreader, layout, "E", "F"));
   EXPECT_EQ(directededge->indoor(), true);
+  EXPECT_TRUE(directededge->forwardaccess() & baldr::kPedestrianAccess);
+  EXPECT_TRUE(directededge->reverseaccess() & baldr::kPedestrianAccess);
```
kevinkreiser, line 115: i added some hardening for all of these. what i did first was add teh routing test below but it failed immediately to even route at all. this made me immediately think that the access was messed up, so to check that i hardened all of these tests to see if pedestrians were even allowed on these edges/nodes. it turned out that no they weren't in some cases which was what the lua change fixed

test/gurka/test_indoor.cc
```diff
@@ -117,4 +133,15 @@ TEST_F(Indoor, EdgeInfo) {
...
+
+TEST_F(Indoor, ElevatorPenalty) {
+  // first route should take the elevator node
+  auto result = gurka::do_action(valhalla::Options::route, map, {"E", "J"}, "pedestrian");
+  gurka::assert::raw::expect_path(result, {"EF", "FG", "GH", "HI", "IJ"});
+
+  // second route should take the stairs because we gave the elevator a huge penalty
+  result = gurka::do_action(valhalla::Options::route, map, {"E", "J"}, "pedestrian",
+                            {{"/costing_options/pedestrian/elevator_penalty", "3600"}});
+  gurka::assert::raw::expect_path(result, {"EF", "CF", "Cx", "xy", "yJ"});
```
kevinkreiser, line 146: this is pretty straightforward. first we take a route that only has a 5 second penalty to take the elevator, which is obviously cheaper than turning and then takign the steps with their penalty. then we jack up the elevator penalty so that the steps become more appealing

taginfo.json
```diff
@@ -4503,7 +4555,7 @@
       ],
       "description": "This tag contains a phonetic guide to pronouncing the official name for a country used in that country if it differs from what is in the name tag"
     },
-   {
+    {
```
kevinkreiser, line 4558: i appreciate this clean up, thank you!

kevinkreiser (approved):
@liyinxiao thanks so much for getting the ball rolling on this awesome feature, i did a couple of documentation clean up things but this looks great and it will be great to get to the next step!

lua/graph.lua
liyinxiao: Thanks for catching this! don't know how I missed this..

PR #3515, opened 2022-01-17
Corrected expansion_props to expansion_properties in api docs

chrstnbwnkl:
# Issue
The `expansion_properties` parameter for the expansion service was listed as `expansion_props` in the API docs, just corrected it.


nilsnolde (approved):
perfect thanks!

PR #3516, opened 2022-01-18
rename tripcommon proto to common

kevinkreiser:
this is just a rename of a single proto file and renames for all `import/include`s

exactly what the title says. basically the trip common proto has stuff that is common to more than just the trip and directions protos so instead of calling it trip common we'll just call it common.


nilsnolde (approved):
sounds good thanks

PR #3519, opened 2022-01-22
indoor routing maneuvers

liyinxiao:
# Issue

[#3469](https://github.com/valhalla/valhalla/issues/3469)

In this PR, we make the changes to maneuvers/narratives to have proper indoor support. Briefly, we would like to have the following narration support:
- Take the elevator to Level 2.
- Take the escalator to Level 2.
- Take the stairs to Parking Level.
- Enter/Exist the building. 

The new narration templates are added in en-US.json and synced to each new narrative language file, following [this](https://github.com/valhalla/valhalla/tree/master/locales).
```
cd locales 
./merge-en.sh
```

## Tasklist

 - [x] Add tests

## Test plan
format the files
```
./scripts/format.sh
```

gurka tests
```
make run-gurka_indoor 
make run-gurka
```
build tiles and test
```
cd build
mkdir -p valhalla_tiles
./valhalla_build_config --mjolnir-tile-dir ${PWD}/valhalla_tiles --mjolnir-tile-extract ${PWD}/valhalla_tiles.tar --mjolnir-timezone ${PWD}/valhalla_tiles/timezones.sqlite --mjolnir-admin ${PWD}/valhalla_tiles/admins.sqlite > valhalla.json
./valhalla_build_tiles -c valhalla.json ~/Downloads/massachusetts-latest.osm.pbf
find valhalla_tiles | sort -n | tar cf valhalla_tiles.tar --no-recursion -T -
./valhalla_service valhalla.json 1

curl http://localhost:8002/route --data '{"locations":[{"lat":42.361145,"lon":-71.0589,"type":"break","city":"Boston"},{"lat":42.358853,"lon":-71.057298,"type":"break","city":"Boston"}],"costing":"pedestrian","directions_options":{"units":"miles"}}' | jq '.'
```

## Requirements / Relations

Based on this PR: https://github.com/valhalla/valhalla/pull/3509


src/odin/maneuversbuilder.cc
```diff
@@ -1648,6 +1741,31 @@ void ManeuversBuilder::SetManeuverType(Maneuver& maneuver, bool none_type_allowe
...
+  // Process elevator
+  else if (maneuver.elevator()) {
+    maneuver.set_type(DirectionsLeg_Maneuver_Type_kElevatorEnter);
+    LOG_TRACE("ManeuverType=ELEVATOR");
+  }
+  // Process steps
+  else if (maneuver.steps()) {
+    maneuver.set_type(DirectionsLeg_Maneuver_Type_kStepsEnter);
+    LOG_TRACE("ManeuverType=STEPS");
+  }
```
kevinkreiser, line 1753: i feel like we maybe dont want to call out using the steps in all situations. im a novice compared to @dgearhart but my gut intuition is that there are lots of cases, especially outside on like hiking paths where we dont need to say take the stairs/steps as there is nothign else to do. indoors steps are a disambiguating feature so i could see always calling them out in that case but then my next thought is what about outdoors but like in a dense city around buildings and stuff. how can we tell when it makes sense to call them out and when not? @dgearhart opinions?

dgearhart: @kevinkreiser I will have to take a look

liyinxiao: This makes sense to me. An easy alternative is to call out indoor stairs only, while leaving outdoor steps ([example](https://wiki.openstreetmap.org/wiki/File:Stairs_with_ramp.jpg)) as it is today. It would look like this:
```
  // Steps
  if (prev_edge->IsStepsUse() && prev_edge->indoor()) {
    maneuver.set_steps(true);
  }
```
Still, I am curious on what @dgearhart thinks.

kevinkreiser: yep that is what my gut was saying for a simple approach. maybe we could start with that and evolve to be smarter in a subsequent pr, just so we arent blocked on design there, im not sure on @dgearhart 's availalbility at the moment to spend some brain cells on it.

kevinkreiser:
@liyinxiao ill review this tonight, thanks again for such quick work!

liyinxiao:
@kevinkreiser Thanks and no worries! There's no rush (didn't realize it sent out a notification)..

proto/directions.proto
```diff
@@ -96,6 +96,11 @@ message DirectionsLeg {
       kPostTransitConnectionDestination = 36;
       kMergeRight = 37;
       kMergeLeft = 38;
+      kElevatorEnter = 39;
+      kStepsEnter = 40;
+      kEscalatorEnter = 41;
+      kBuildingEnter = 42;
+      kBuildingExit = 43;
```
kevinkreiser, line 103: perfect combination, exactly what i would have expected

kevinkreiser (approved):
looks good to me, but i'll give @dgearhart just a little bit to see if he has anyy feedback

src/odin/narrativebuilder.cc
```diff
@@ -464,6 +464,31 @@ void NarrativeBuilder::Build(std::list<Maneuver>& maneuvers) {
         maneuver.set_verbal_arrive_instruction(FormVerbalArriveInstruction(maneuver));
         break;
       }
+      case DirectionsLeg_Maneuver_Type_kElevatorEnter: {
```
dgearhart, line 467: What about the verbal instructions?

liyinxiao: Thanks for reviewing this PR @dgearhart ! Yeah, I wasn't sure what's the best verbal instructions for these indoor maneuvers, and I left it blank with the intention that this could be iterated later on. 
But I could take a stab at it too. I propose for `verbal_pre_transition_instruction`, we can stick to the same instructions as `instruction` while for `verbal_post_transition_instruction`, we can use `FormVerbalPostTransitionInstruction()`.  What do you think? Let me know what your thoughts!

dgearhart (approved):
Thank you!

kevinkreiser (approved):

PR #3527, opened 2022-02-02
Reducing the Use of oneof in Protos

kevinkreiser:
In a previous PR #3457, we migrated our protobuf definitions to proto3. When we did that we needed to make use of a trick that uses the `oneof` keyword to emulate the ability to do `optional` that was removed from the language in proto3. We took the conservative approach by simply adding `oneof` to any previously `optional` parameter. This is pretty ugly and quite verbose but it allowed us to make the migration without having to think about each individual field.

This PR significantly reduces the use of `oneof`. I accomplished this in the following ways:

1. in proto3 default values for all primative types are 0 (strings are empty string), this means that if the field has no semantic use for the 0 value (or empty string) we can simply allow these values to communicate them not being set. many many cases in the code worked that way so i was able to remove a great many `oneof`s this way
2. enums which make use of the default value 0 but do need differentiate being unset get a new default value denoting its unset-ness
3. deleteing unused options such as all the address stuff in location

Cases where we cannot remove `oneof` are those in which we need to differentiate between being set and not set (optional) but also the default value is not 0 in the case its not set. For example our node snap tolerance defaults to 5 meters when you dont set it, but if we removed the `oneof` this would end up making 0 the new default which is a breaking change. In those cases we need to keep using oneof. I s hould note that this only matters for pbf input, json input allows us to check whether something was or wasnt set.


src/baldr/attributes_controller.cc
```diff
@@ -1,9 +1,10 @@
-#include <thor/attributes_controller.h>
+#include "baldr/attributes_controller.h"
+#include "midgard/logging.h"
 
 #include <string>
 
 namespace valhalla {
-namespace thor {
+namespace baldr {
```
mandeepsandhu, line 7: Whats the reason for moving this to `baldr`?

mandeepsandhu:
For string types, does this PR switch to using `!string.empty()` where previously we'd do `has_string()` ?

mandeepsandhu (dismissed):
Thanks for the cleanup! Too many diffs, I'll base my review on all tests passing! :slightly_smiling_face:

proto/common.proto
```diff
@@ -24,70 +24,40 @@ message SearchFilter {
...
-    uint32 original_index = 52;
-  }
-  LatLng projected_ll = 53;
-  oneof has_leg_shape_index {
-    uint32 leg_shape_index = 54;
-  }
-  oneof has_distance_from_leg_origin {
-    double distance_from_leg_origin = 55;
-  }
+  repeated PathEdge edges = 1;
```
mandeepsandhu, line 53: any  concerns renumbering the fields here and few other places? I remember we store the trip leg pbf for some of the pinpoint tests which would fail to read back the correct data if the pbf fields were renumbered.

I guess since the tests are passing the renumbering here does not affect the few places we've stored these protobufs on disk?

kevinkreiser: oh it broke the tests, but i fixed them by leaving the old defs there and copying them to the new defs and overwiiting the files. ive been doing this alot for all the protobuf changes, its a huge pain in the butt but its worth it to get the code cleaner. you see how it did it over here (takes a long time to load because of diff size): https://github.com/valhalla/valhalla/pull/3527/files#diff-836ee7bf7651f8100e6b6072ab85387a7fee2ea9178762aabf200726bd31a0d9R32

kevinkreiser: the other stuff i renumbered were things that were parsed from the request, so i could just call `ParseApi` and then again overwrite the proto after i've parsed the request. Also a pain in the butt but something ive been doing a lot with all these changes and organizational stuff

mandeepsandhu: ah ok I missed some of those diffs at first glance (as you said, because of the size of the change github had issues expanding all the diffs).

thanks for updating the tests! :+1:

src/loki/worker.cc
```diff
@@ -44,7 +44,7 @@ void loki_worker_t::parse_locations(google::protobuf::RepeatedPtrField<valhalla:
       else if (location.radius() > max_radius)
         location.set_radius(max_radius);
 
-      if (!location.has_heading_tolerance_case())
+      if (location.has_heading_tolerance_case())
```
mandeepsandhu, line 47: Not sure I follow this. We're now overriding `heading_tolerance_case` even if it explicitly set?

src/thor/triplegbuilder.cc
```diff
@@ -445,9 +443,9 @@ void CopyLocations(TripLeg& trip_path,
     valhalla::Location* tp_intermediate = trip_path.add_location();
     tp_intermediate->CopyFrom(intermediate);
     // we can grab the right edge index in the path because we temporarily set it for trimming
-    if (!intermediate.correlation().has_leg_shape_index_case()) {
+    /*if (!intermediate.correlation().has_leg_shape_index_case()) {
```
mandeepsandhu, line 446: Do we leave this commented?

mandeepsandhu (dismissed):
had a few clarifying question. rest lgtm. Thx for heavy lifting! :slightly_smiling_face:

src/loki/worker.cc
kevinkreiser: i do not know how that got in here, i didnt even change heading tolarance in the proto. ill add a unit test because obviously this isnt protected by existing tests

kevinkreiser: done

src/thor/triplegbuilder.cc
kevinkreiser: done

mandeepsandhu (approved):
Thx for adding new tests. lgtm

src/baldr/attributes_controller.cc
kevinkreiser: this was so that we could re-use this class in odin when we serialize the route out. im not sure its the best place, we could hvae argued to put it in tyr also i think. if you like tyr better we could totally move that in the next pr. also sorry for not answering this question earlier somehow i didnt see it!

PR #3532, opened 2022-02-07
Trim dijsktra expansion

### LINKED ISSUE #3517: trim expansion response for intervals less than 10 mins/kms
now we finally wanted to make use of the dijkstra `/expansion` tracking and I realized it's really weird that we return huge geometries even though the request just asks for 1 minute isochrone. reason is that we add some buffer to user-specified max_mins/kms of 10 mins/kms, so dijkstra finds edges with higher cost.

just had a quick look but didn't see an elegant obvious way.. I'll leave it here for now, if anyone has an idea off the to of their head, I'd appreciate it

nilsnolde:
fixes #3517 

the issue explains the intention of this change: I just moved the `expansion_callback` into the `Isochrone::ShouldExpand` function so I have access to the `max_seconds_` and `max_meters_` members which are derived from the user input.


.gitignore
```diff
@@ -1,6 +1,6 @@
 # editor temporary / backup files
 *~
-
+.venv
```
nilsnolde, line 3: oops. but in fact could we keep it? (I can move to a more sensible location in .gitignore)

nilsnolde, line 3: oops. but in fact could we keep it? (I can move to a more sensible location in .gitignore)

nilsnolde, line 3: oops. but in fact could we keep it? (I can move to a more sensible location in .gitignore)

nilsnolde: github being buggy again.. :roll_eyes:

src/thor/isochrone.cc
```diff
@@ -12,6 +12,8 @@ using namespace valhalla::sif;
 
 namespace {
 
+constexpr float METRIC_PADDING = 10.f;
```
kevinkreiser, line 15: :+1:

src/thor/isochrone.cc
```diff
@@ -329,6 +333,16 @@ ExpansionRecommendation Isochrone::ShouldExpand(baldr::GraphReader& /*graphreade
   // prune the edge if its start is above max contour
   if (time > max_seconds_ && distance > max_meters_)
     return ExpansionRecommendation::prune_expansion;
+
+  // log it back to expansion_action
+  if (expansion_callback_) {
+    if (time <= (max_seconds_ - METRIC_PADDING * kSecondsPerMinute) ||
+        distance <= (max_meters_ - METRIC_PADDING * kMetersPerKm)) {
+      expansion_callback_(graphreader, pred.edgeid(), "dijkstras", "s", pred.cost().secs,
```
kevinkreiser, line 341: i'm trying to think of a cleaner way to do this that doesnt involve moving this out of the base class. it would be awesome if here we could simply set a member variable boolean, to false or something that would cause it not to do anything when we detect we dont want to actually call it back.

actually i guess you could keep an empty callback as a memeber variable and swap it in and out based on whether or not we are in the "extra" part of the expansion or not.

do you have any thoughts on that?

nilsnolde: sure no problem, I'll see what makes sense in order to not remove it in the base dijkstra class

nilsnolde: just wanted to get to this real quick, only to (re-)discover the initial problem: `max_meters_` and `max_seconds_` are member vars of `Isochrone` not `Dijkstras` and they're set during constructing the `isotile_`. that's why I moved the expansion_callback call in the first place.

in the light of that, I'm not sure I understand how your previous suggestions would be implemented.. we'd somehow have to let dijkstras know about the metrics if the call should stay where it was, and the metrics might differ for other dijkstra derived algos. also the expansion_callback is universal for all algos, so we can't just plug in algo-specific args into the lambda easily. 

right now I have no idea how to keep the function call in the base class.. do you?

kevinkreiser (approved):

nilsnolde:
merging master dismissed your review @kevinkreiser could you approve again? thx!

nilsnolde:
:facepalm: I merged another PR after you approved again, again conflicting changelog.. I just aborted all CI where I could (not azure), can you just merge it when you find the time @kevinkreiser ?

kevinkreiser (approved):

nilsnolde:
Let’s get this shipped finally, I’ll promise to not merge anymore before😅 needs yet another approval, sorry for the hassle..

kevinkreiser (approved):

Issue #3542, opened 2022-02-19
valhalla_export_edges : missing null columns separator


ImreSamu:
Based on the `valhalla_export_edges -h`  - The default separator is : '00'
* `-c [ --column ] arg   What separator to use between columns [default=\0].`

Based on my check: 
-  3.1.4:   :heavy_check_mark:    is OK   
   -   `commit ac824e5ca839d36a489e7123c5ed2920ecc3d674 (grafted, HEAD, tag: 3.1.4)`
-  master: :bug: :lady_beetle:  missing null separator in the output
   -    `commit b9744b98abddad321ed2bfa11eb679a70b0c443b (grafted, HEAD -> master, origin/master, origin/HEAD)`

Minimal test scripts:
```bash
mkdir build 
cd build 
cmake ..
make -j$(nproc)
mkdir test_export_edges
cd test_export_edges
../../scripts/valhalla_build_config > ./config.json
cp ../../test/data/utrecht_netherlands.osm.pbf .
sed -i "s|/data/valhalla|.|g"  config.json
../valhalla_build_tiles  --config ./config.json ./utrecht_netherlands.osm.pbf
../valhalla_export_edges --config ./config.json > ./edges.0csv
# test  https://unix.stackexchange.com/a/527524 : count the number of bytes in a file, grouping the same bytes
od -t x1 -w1 -v -An ./edges.0csv | sort | uniq -c | head
```

The output:

#### 3.1.4 - "count the number of bytes in a file, grouping the same bytes" 

`00`  is exist;  looks OK. :heavy_check_mark: 

```
user@tx7:~/tmp/valhalla_314/build/test_export_edges$ od -t x1 -w1 -v -An ./edges.0csv | sort | uniq -c | head
   6057  00
   6009  0a
   2317  20
     26  27
     77  2d
    193  2e
     31  30
     66  31
    130  32
     39  33
```

#### master -  "count the number of bytes in a file, grouping the same bytes" 

`00`  is not exist in the output;  and I am expecting  .. 
:bug: :lady_beetle: 

```
user@tx7:~/tmp/valhalla_master/build/test_export_edges$ od -t x1 -w1 -v -An ./edges.0csv | sort | uniq -c | head
   6043  0a
   2383  20
     26  27
     79  2d
    193  2e
     57  30
    101  31
    170  32
    100  33
     16  34
```

ImreSamu:
Docker based test with the latest `valhalla/valhalla` docker image

```
[...]

PR #3543, opened 2022-02-19
fix valhalla_export_edges null separator ( #3542 )

ImreSamu:
# Issue

#3542 - "valhalla_export_edges : missing null columns separator"

## Tasklist

 - [ ] Add tests

## Requirements / Relations

- none


kevinkreiser (approved):
Yeah it was an std::string before. Using it as a cstring would not work. Thanks for the catch!

PR #3545, opened 2022-02-21
exit and exit_verbal should be similar.

molind:
Few more fixes in ru-RU locale.

It should use similar construction in exit and exit_verbal.


molind:
Current version of `exit` phrases means "Turn right from exit" or "Turn left from exit". Like you're already on exit from highway and now you should turn right or left. Corrected version says direction to the exit. I'm still unsure about correctness of all phrases with <NAME_SIGN>. May be it will require more fixes.

kevinkreiser (approved):
LGTM can you merge master one more time and we'll get it in?

kevinkreiser (approved):

PR #3555, opened 2022-02-28
Note for arm64 dependencies in README

nuttert:
During the first build under arm64 with Rosetta, I encountered an error linking libs that had already been previously installed in a regular terminal, their removal helped, as well as installing dependencies from the sources.

If you find this a useful comment, we can add this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,9 @@ echo "alias mbrew='arch -arm64e /opt/homebrew/bin/brew'" >> ~/.zshrc
 
 You will use them to specify the platform when installing a library. Note: use `ibrew` in `Rosetta Terminal` to install all dependencies for `valhalla` and `prime_server` projects.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -166,6 +166,9 @@ echo "alias mbrew='arch -arm64e /opt/homebrew/bin/brew'" >> ~/.zshrc
 
 You will use them to specify the platform when installing a library. Note: use `ibrew` in `Rosetta Terminal` to install all dependencies for `valhalla` and `prime_server` projects.
 
+> **_NOTE:_** If when installing packages below you get message `attempting to link with file built for macOS-arm64`, you can remove already installed packages for arm64 i.e. `mbrew uninstall ...`. Also, if there are problems with individual packages, you can install them from sources e.g. [geos](https://github.com/libgeos/geos) or [sqlite](https://www.sqlite.org/download.html).
```
kevinkreiser, line 169: i don think this should be a quote, but otherwise looks good

```suggestion
**_NOTE:_** If when installing packages below you get message `attempting to link with file built for macOS-arm64`, you can remove already installed packages for arm64 i.e. `mbrew uninstall ...`. Also, if there are problems with individual packages, you can install them from sources e.g. [geos](https://github.com/libgeos/geos) or [sqlite](https://www.sqlite.org/download.html).
```

PR #3560, opened 2022-03-07
Add matrix objects to thor worker

### LINKED ISSUE #3556: Potential memory leak in Matrix API
Hello,

When building bigger matrices >250+ I am having a memory leaks all over the place and it doesn't seem like it's caching issue since I am replaying the same matrices and memory usage increases linearly. Also, when doing these bigger matrices there is a huge memory usage, for example 572x572 is using 12+GB of RAM (and not releasing it once done) - is this normal?
If any data/log is needed I can provide!

Kind regards!

dnesbitt61:
Add matrix objects to thor worker. This allows the clear method to be called after serialization occurs - as other actions do. For CostMatrix, the clear call can take a long time (100s of ms), so this change can improve response time / latency.

Also added a few "cleanup" items in this PR: 

- changed name of Clear methods to clear
- resize and shrink to fit the CostMatrix vectors (since these can allocate a lot of memory)
- removed a couple clear calls in the main matrix method since cleanup is done in the worker::cleanup
- remove use of shared_ptr for DoubleBucketQueue in CostMatrix adjacency lists

# Issue

fixes #3556 - Ran this with valgrind and found no memory leak. Ran a large (520x520 location) request multiple times in a loop - memory grew to 11.1 GB after several requests but never grew higher.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


dnesbitt61:
With master, I ran a large 512x512 matrix (listed in issue) in a loop (using curl) 20x:
With top - memory use (RES) topped at 9.9g
With some timing, the requests took an average of 89.8 secs each.

With this branch the same test shows:
With top, memory use goes to 11.1g (after about 6 requests) - never increases afterwards.
Timing: requests took an average of 85.9 secs (cleanup occurs after serializing and averages 740ms).

valhalla/thor/timedistancematrix.h
```diff
@@ -109,7 +109,7 @@ class TimeDistanceMatrix {
    * Clear the temporary information generated during time+distance
    * matrix construction.
    */
-  void Clear();
+  void clear();
```
nilsnolde, line 112: ;)

nilsnolde (dismissed):
thanks @dnesbitt61 , makes sense to me! others are better equipped to review code changes though..

any idea where the memory increase is coming from? can't really see it based on the code changes..

dnesbitt61:
I do not know why top shows more memory use. I added code to clear targets_ - that wouldn't account for the increase but figure it should be cleared more thoroughly. Adding the matrix objects to worker I don't think would do it either - these classes do not seem to start with any large memory reservations (and I tried removing timedistancebssmatrix from worker class without any change in behavior).

src/thor/costmatrix.cc
```diff
@@ -65,24 +65,44 @@ float CostMatrix::GetCostThreshold(const float max_matrix_distance) {
 
 // Clear the temporary information generated during time + distance matrix
 // construction.
-void CostMatrix::Clear() {
+void CostMatrix::clear() {
```
kevinkreiser, line 68: this tries to reclaim all allocations. that might not be great for a long running server though. some of these containers could stand to keep some heap space so that they dont have to immediately reallocate on the next request. should we make at least some default values other than `0` to shrink to so that we dont have to always pay the reallocation penalty? other algorithms have configuration (this could too) which controls what these maximum-allocation-after-request numbers, that way mobile applications can set it to `0` to get all their ram back and servers can use some reasonable numbers to avoid the realloc penalty. @dnesbitt61 what do you think?

dnesbitt61: The problem with CostMatrix is that for an m sources by n targets request there are m+n EdgeLabel vectors, m+n EdgeStatus objects, and m+n adjacency lists. I clear, resize, and shrink these vectors so the internal container memory is deallocated. Which allocated memory would you retain? 
Contrast that to TimeDistanceMatrix which has one of each of these objects, so retaining some allocation (for EdgeLabels for example) makes sense.

kevinkreiser: yep good point about the nesting. i guess maybe only that of the top level containers could be regulated via config so we at least avoid some level of reallocation? honestly i havent measured the throughput difference so maybe for now we can do something like this, give it a test, and make it configureable in a future pr?

kevinkreiser:
@dnesbitt61 do you mind putting a changelog entry for this? it seems like a significant enough change to me, what do you think?

dnesbitt61:
update Changelog. This seems like a good change to me as it follows the pattern of cleanup after serializing the result.

kevinkreiser (approved):

Issue #3258, opened 2021-08-10
mean_elevation for EdgeInfo is initialized with 1234 m


nilsnolde:
in https://github.com/valhalla/valhalla/blob/480cc790ea43f1da0415d85466af7697aeaac475/src/mjolnir/graphbuilder.cc#L718

the `mean_elevation` is initialized with `1234`. if no elevation is provided, it'll stay that way and return it on a `/locate` or `/trace_attributes` call (or in my current `/roads_stats` implementation ;)), which should be confusing enough but couldn't find an issue about it.

I tried real quick passing a `0` instead with old tiles, new tiles, with and without elevation and was all good (also didn't see anything in the code that would actually rely on `1234`. Though 0 is also a dumb value, maybe smth like `std::numeric_limits<int32_t>::max()` as `kNoElevation`? 

Ideally I think we'd return a `null` JSON value if there's no elevation provided for an edge. Got a quick code suggestion how to make that happen in a `json::MapPtr`? e.g. 
```c++
  json::MapPtr edge_info = json::map({
      {"mean_elevation", mean_elevation() != std::numeric_limits<int32_t>::max() ? static_cast<uint64_t>(mean_elevation()) : nullptr}
  });
```

obviously doesn't work..

kevinkreiser:
cc @dnesbitt61 

https://github.com/valhalla/valhalla/pull/1648/files#diff-8eb193c3251becf80bdd36d40abdfe7681301bbfbac59302efcbb70f3a2e1fa9R655

that does seem like an accident. i would have expected a constant as you suggest.

i will note though that we store this as an unsigned value and then we offset it at runtime to get the actual value. because of this i think `0` is actually probably a pretty good value to signify no data. that would work out to -500m below sea level which is lower than the lowest spot currently (not counting subteranian i guess). we can then update the serializer to  just say if the elevation bin is 0 output null as you said otherwise use the value. this means that all old data has been broken since v3 was created but oh well, you win some you lose some. also it means that we lose the ability to represent -500m but i think that is inconsequential.

nilsnolde:
> i think 0 is actually probably a pretty good value to signify no data

I'll look through it in detail eventually in PR mode, but couldn't we store a signed int32? that doesn't break the tile format does it? don't worry, no need to scour the code base to prove your point, we can do that in the coming PR setting. it'd be nice to support below 0 (and actual 0 elevation, not signifying nodata). with love for my dutch folks :smile:

[...]

PR #3571, opened 2022-03-11
fix graphbuilder's elevation initialization

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/3258

finally getting some time to fix some of my old issues.. 

mostly fixes returning `"mean_elevation": null` when there is none, instead of `1234`. also some others:

`EdgeInfo::json` was casting elevation to an unsigned integer, leading to bogus values for the (admittedly few) roads which are sub sea level, e.g. next to dead sea:
```
curl --request POST \
  --url https://valhalla1.openstreetmap.de/locate \
  --data '{
	"costing": "auto",
	"verbose": true,
	"id": 1,
	"locations": [
		{
			"lon": 35.408936,
			"lat": 31.591404
		}
	]
}'
```
leads to `"mean_elevation": 18446744073709551254`. with this PR it's `"mean_elevation": -362`.

### DIFF
--- a/src/baldr/edgeinfo.cc
+++ b/src/baldr/edgeinfo.cc
@@ -313,11 +313,17 @@ std::string EdgeInfo::level_ref() const {
 json::MapPtr EdgeInfo::json() const {
   json::MapPtr edge_info = json::map({
       {"way_id", static_cast<uint64_t>(wayid())},
-      {"mean_elevation", static_cast<uint64_t>(mean_elevation())},
       {"bike_network", bike_network_json(bike_network())},
       {"names", names_json(GetNames())},
       {"shape", midgard::encode(shape())},
   });
+  // add the mean_elevation depending on its validity
+  const auto elev = mean_elevation();
+  if (elev == kNoElevationData) {
+    edge_info->emplace("mean_elevation", nullptr);
+  } else {
+    edge_info->emplace("mean_elevation", static_cast<int64_t>(elev));
+  }
 
   if (speed_limit() == kUnlimitedSpeedLimit) {
     edge_info->emplace("speed_limit", std::string("unlimited"));
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -732,10 +732,9 @@ void BuildTileSet(const std::string& ways_file,
               bike_network = w.bike_network();
             }
 
-            // Add edge info. Mean elevation is set to 1234 as a placeholder, set later if we have it.
             edge_info_offset =
                 graphtile.AddEdgeInfo(edge_pair.second, (*nodes[source]).graph_id,
-                                      (*nodes[target]).graph_id, w.way_id(), 1234, bike_network,
+                                      (*nodes[target]).graph_id, w.way_id(), 0, bike_network,
                                       speed_limit, shape, names, tagged_values, pronunciations, types,
                                       added, dual_refs);
             if (added) {
--- a/src/tyr/trace_serializer.cc
+++ b/src/tyr/trace_serializer.cc
@@ -111,10 +111,18 @@ json::ArrayPtr serialize_edges(const AttributesController& controller,
         edge_map->emplace("lane_connectivity", lane_connectivity);
       }
       if (controller(kEdgeMaxDownwardGrade)) {
-        edge_map->emplace("max_downward_grade", static_cast<int64_t>(edge.max_downward_grade()));
+        if (edge.max_downward_grade() == kNoElevationData) {
+          edge_map->emplace("max_downward_grade", static_cast<int64_t>(edge.max_downward_grade()));
+        } else {
+          edge_map->emplace("max_downward_grade", nullptr);
+        }
       }
       if (controller(kEdgeMaxUpwardGrade)) {
-        edge_map->emplace("max_upward_grade", static_cast<int64_t>(edge.max_upward_grade()));
+        if (edge.max_downward_grade() == kNoElevationData) {
+          edge_map->emplace("max_upward_grade", static_cast<int64_t>(edge.max_upward_grade()));
+        } else {
+          edge_map->emplace("max_upward_grade", nullptr);
+        }
       }
       if (controller(kEdgeWeightedGrade)) {
         edge_map->emplace("weighted_grade", json::fixed_t{edge.weighted_grade(), 3});
@@ -122,10 +130,13 @@ json::ArrayPtr serialize_edges(const AttributesController& controller,
       if (controller(kEdgeMeanElevation)) {
         // Convert to feet if a valid elevation and units are miles
         float mean = edge.mean_elevation();
-        if (mean != kNoElevationData && options.units() == Options::miles) {
-          mean *= kFeetPerMeter;
+        if (mean == kNoElevationData) {
+          edge_map->emplace("mean_elevation", nullptr);
+        } else {
+          edge_map->emplace("mean_elevation", static_cast<int64_t>(options.units() == Options::miles
+                                                                       ? mean
+                                                                       : mean * kFeetPerMeter));
         }
-        edge_map->emplace("mean_elevation", static_cast<int64_t>(mean));
       }
       if (controller(kEdgeWayId)) {
         edge_map->emplace("way_id", static_cast<uint64_t>(edge.way_id()));
--- a/valhalla/baldr/graphconstants.h
+++ b/valhalla/baldr/graphconstants.h
@@ -198,7 +198,8 @@ constexpr uint32_t kMaxCurvatureFactor = 15;
 constexpr uint32_t kMaxAddedTime = 255;
 
 // Elevation constants
-constexpr float kNoElevationData = 32768.0f;
+// this is the minimum we support, i.e. -500 m would result in "no elevation"
+constexpr float kNoElevationData = -500;
 
 // Node types.
 enum class NodeType : uint8_t {
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # editor temporary / backup files
 *~
+.venv
 
 # built objects
 valhalla_run_map_match
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
    * FIXED: Wrong predecessor opposing edge in dijkstra's expansion [#3528](https://github.com/valhalla/valhalla/pull/3528)
    * FIXED: exit and exit_verbal in Russian locale should be same [#3545](https://github.com/valhalla/valhalla/pull/3545)
    * FIXED: Skip transit tiles in hierarchy builder [#3559](https://github.com/valhalla/valhalla/pull/3559)
+   * FIXED: Fix "no elevation" values and /locate elevation response [#3571](https://github.com/valhalla/valhalla/pull/3571)
    
 * **Enhancement**
    * CHANGED: Pronunciation for names and destinations [#3132](https://github.com/valhalla/valhalla/pull/3132)


src/tyr/trace_serializer.cc
```diff
@@ -111,21 +111,32 @@ json::ArrayPtr serialize_edges(const AttributesController& controller,
         edge_map->emplace("lane_connectivity", lane_connectivity);
       }
       if (controller(kEdgeMaxDownwardGrade)) {
-        edge_map->emplace("max_downward_grade", static_cast<int64_t>(edge.max_downward_grade()));
+        if (edge.max_downward_grade() == kNoElevationData) {
+          edge_map->emplace("max_downward_grade", static_cast<int64_t>(edge.max_downward_grade()));
+        } else {
+          edge_map->emplace("max_downward_grade", nullptr);
+        }
```
nilsnolde, line 118: the grades also get a `kNoElevationData` in the `TripLegBuilder`

valhalla/baldr/graphconstants.h
```diff
@@ -198,7 +198,8 @@ constexpr uint32_t kMaxCurvatureFactor = 15;
 constexpr uint32_t kMaxAddedTime = 255;
 
 // Elevation constants
-constexpr float kNoElevationData = 32768.0f;
+// this is the minimum we support, i.e. -500 m would result in "no elevation"
+constexpr float kNoElevationData = -500;
```
nilsnolde, line 202: this constant was only used for /trace*, in the TripLegBuilder to set `kNoElevationData` values and in the serializer. now we also use it to serialize `EdgeInfo` (s. above)

src/baldr/edgeinfo.cc
```diff
@@ -313,11 +313,17 @@ std::string EdgeInfo::level_ref() const {
...
       {"bike_network", bike_network_json(bike_network())},
       {"names", names_json(GetNames())},
       {"shape", midgard::encode(shape())},
   });
+  // add the mean_elevation depending on its validity
+  const auto elev = mean_elevation();
+  if (elev == kNoElevationData) {
+    edge_info->emplace("mean_elevation", nullptr);
+  } else {
+    edge_info->emplace("mean_elevation", static_cast<int64_t>(elev));
```
nilsnolde, line 325: hm.. this is still missing a possible conversion to miles I just realized

nilsnolde: that's actually a rhetorical question.. a request with or without `"directions_options":{"units":"miles"}` has no impact. I'll PR that separately

kevinkreiser: Yes, locate was designed to just regurgitate exactly what is in the tile data back to the requester. It's useful for programmatic things not meant to be used directly by a human. For this reason I don't think we should at all do units conversions. The api docs should just say everything is metric. I could be convinced it makes sense to care about units though. Seems you feel like it should. Could you maybe make a case for it just so I could understand the use case you have in mind?

src/mjolnir/graphbuilder.cc
```diff
@@ -732,10 +732,9 @@ void BuildTileSet(const std::string& ways_file,
               bike_network = w.bike_network();
             }
 
-            // Add edge info. Mean elevation is set to 1234 as a placeholder, set later if we have it.
             edge_info_offset =
                 graphtile.AddEdgeInfo(edge_pair.second, (*nodes[source]).graph_id,
-                                      (*nodes[target]).graph_id, w.way_id(), 1234, bike_network,
+                                      (*nodes[target]).graph_id, w.way_id(), 0, bike_network,
```
kevinkreiser, line 737: I've not looked at this bug in a while but why not set to the kNoElevation value so that if it never gets set the bits of code that use it will know? Also we should consider what uses this value incase it would cause data compatibility issues

valhalla/baldr/graphconstants.h
kevinkreiser: The previous value is the no data value in srtm, which is the data set we use to add eleveation. I'm certain we also reference this value in skadi just not this constant. What this means is that it's possible skadi returns the previous value when it has no data. I'll have to go check on that when I get a chance or maybe you already looked?

src/baldr/edgeinfo.cc
nilsnolde: I wasn't sure myself if that's supposed to be supported by the locate api, then I saw the docs are explicitly using `miles`  in the example so I figured it should convert. very happy not to go through the process tbh, won't make an argument here;)

valhalla/baldr/graphconstants.h
nilsnolde: I _think_ actually the tiles' nodata value is the negative equivalent, which seems to be acknowledged in skadi as well: 
https://github.com/valhalla/valhalla/blob/master/src/skadi/sample.cc#L33

that constant was just used for communication between triplegbuilder and the serializer. there's no other use of `32768` in the code base, so I assume this is fine to go with. I'll re-think (and test) a little more though, understand the relationship of all these "no data" values. would be nice to consolidate a little.

src/mjolnir/graphbuilder.cc
nilsnolde: yeah I think you're right. was a little quick last night. in the issue everyone said it should be 0, didn't challenge that much it seems. don't think there can be compatibility issues, I'll have a look again though and describe the flow a bit better

src/mjolnir/graphbuilder.cc
```diff
@@ -732,12 +732,11 @@ void BuildTileSet(const std::string& ways_file,
               bike_network = w.bike_network();
             }
 
-            // Add edge info. Mean elevation is set to 1234 as a placeholder, set later if we have it.
             edge_info_offset =
                 graphtile.AddEdgeInfo(edge_pair.second, (*nodes[source]).graph_id,
-                                      (*nodes[target]).graph_id, w.way_id(), 1234, bike_network,
-                                      speed_limit, shape, names, tagged_values, pronunciations, types,
-                                      added, dual_refs);
+                                      (*nodes[target]).graph_id, w.way_id(), kNoElevationData,
```
nilsnolde, line 737: so I changed it from 0 to the no data value, which will return `null` in the geojson response. I traced all usages of `mean_elevation` & co and can't see how this change would affect anything negatively

src/thor/triplegbuilder.cc
```diff
@@ -1104,13 +1104,9 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
...
 
-  // Set mean elevation if requested (set to kNoElevationData if unavailable)
+  // Set mean elevation if requested (will be kNoElevationData if unavailable)
   if (controller(kEdgeMeanElevation)) {
-    if (graphtile->header()->has_elevation()) {
-      trip_edge->set_mean_elevation(edgeinfo.mean_elevation());
-    } else {
-      trip_edge->set_mean_elevation(kNoElevationData);
-    }
+    trip_edge->set_mean_elevation(edgeinfo.mean_elevation());
```
nilsnolde, line 1109: this was redundant for elevation. the `set_mean_elevation` will handle the case when there's no elevation set

valhalla/baldr/graphconstants.h
nilsnolde: I looked into the details and all things accessing elevation. that previous value was sorta arbitrary it seems. in skadi it's -32768.0f.

kevinkreiser: yeah the value in skadi is the NO_DATA value in SRTM which is what the raw data will return if you ask it for a location where it doesnt have data. when we store it in the tiles we have to discretize the data further to pack it into a smaller number of bits. because of this we set a min of -500 and a max of however many descrete values we can have based on how many bits wide the field is. so that makes sense to me. default value should be no data which in this case would be -500, the constant in edgeinfo that you moved over to the graph constants. so that looks all good. the only question i have after this is when we are building tiles do we check for the skadi no data value and then convert to our -500 no data value when we store it in the tiles?

nilsnolde: ah got ya.. probably not! will take a look

nilsnolde: nope, it doesn't..  it looks to me like it never did actually. when building elevation for the tiles, it gets all height samples including skadi's `NO_DATA_VALUE`, then when computing the `weighted_grade` it just skips any `NO_DATA_VALUE` which would result in 0 elevation if there's no elevation data in the tiles, which then gets set on the tiles. so in the elevation builder we have to check for that and set the -500. I'll commit quickly

src/skadi/util.cc
```diff
@@ -71,7 +75,8 @@ weighted_grade(const std::vector<double>& heights,
   // Get the average weighted grade by homogenizing total weight. Return
   // max grades (up and down) and mean elevation.
   return std::make_tuple(total_grade * (1.0 / total_weight), max_up_grade, max_down_grade,
-                         (total_elev / n));
+                         (has_only_no_data ? static_cast<double>(baldr::kNoElevationData)
+                                           : total_elev / n));
```
nilsnolde, line 79: in this function we can check if the heights were actually all no_data. if so, we return our own no data value.

valhalla/baldr/graphconstants.h
nilsnolde: jep got it @kevinkreiser . I changed the logic slightly in `skadi/util.cc:weighted_grade`

test/util_skadi.cc
```diff
@@ -24,6 +24,14 @@ TEST(UtilSkadi, Grade) {
   answer = (down_grade * down_weight * 3 + up_grade * up_weight) / (down_weight * 3 + up_weight);
   grade = std::get<0>(skadi::weighted_grade({0, 1, 0, -1, -2}, 1.0));
   EXPECT_NEAR(grade, answer, .00001) << "Weighted grade was not right";
+
+  // check mean elevation
+  auto elev = std::get<3>(skadi::weighted_grade({0.0f, 15.0f, -32768.0f}, 1.0));
+  EXPECT_NEAR(elev, 7.5f, 0.00001);
+
+  // check mean elevation
```
kevinkreiser, line 32: weighted grade is not mean elevation though its like the nominal slope of the road

kevinkreiser: meh nevermind, just the naming is unfortunate, its grade and mean elevation

src/skadi/util.cc
kevinkreiser: i actually think we shouldnt change this, i think we should detect it on the other side of this call and convert it there. would you be cool with that? the reason being is that skadi is the interface to the raw data, baldr::kNoElevationData is a convention we use in our tile data. so the caller, somewhere in mjolnir should check if it got back a no data value and and convert it there imho

src/mjolnir/edgeinfobuilder.cc
```diff
@@ -24,10 +24,10 @@ void EdgeInfoBuilder::set_wayid(const uint64_t wayid) {
 
 // Set the mean elevation.
 void EdgeInfoBuilder::set_mean_elevation(const float mean_elev) {
-  if (mean_elev < kMinElevation) {
+  if (mean_elev <= kNoElevationData) {
```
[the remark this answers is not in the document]
kevinkreiser: isnt it the case that we need to differentiate between no data and minimum value though? like when its no data then we serialize null and when its minimum value we serialize the minimum value, right?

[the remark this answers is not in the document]
nilsnolde: Minimum value (-500) will not be serialized, same as no data. Think that’s fine, there’s really no place on terrestrial earth at -500. Dead Sea depression is at -450 or so, roads near that have the lowest values in the world.

src/skadi/util.cc
nilsnolde: Totally fine with me, it’s cleaner architecturally. Then I’ll return the raw no data value and let mjolnir detect and handle it

test/util_skadi.cc
nilsnolde: It’s only getting the elevation value from the tuple though;)

src/skadi/util.cc
```diff
@@ -71,7 +76,8 @@ weighted_grade(const std::vector<double>& heights,
   // Get the average weighted grade by homogenizing total weight. Return
   // max grades (up and down) and mean elevation.
   return std::make_tuple(total_grade * (1.0 / total_weight), max_up_grade, max_down_grade,
-                         (total_elev / n));
+                         (has_only_no_data ? static_cast<double>(baldr::kNoElevationDataRaw)
+                                           : total_elev / n));
```
nilsnolde, line 80: we can't move all of the logic since the caller has no chance of knowing whether the heights were all on sea level or no_data values. actually the latter would've caused a "division by zero" error so far. guess that never happened until now

src/tyr/trace_serializer.cc
kevinkreiser: The grades aren't an elevation though, they are a slope. Are you saying we store that in the tiles now, I thought they had a different number of buts anyway

src/skadi/util.cc
```diff
@@ -31,29 +32,33 @@ weighted_grade(const std::vector<double>& heights,
   double max_up_grade = 0.0;
   double max_down_grade = 0.0;
 
+  bool has_only_no_data = true;
+
   // Accumulate elevation - to compute mean_elevation
   uint32_t n = 0;
   double total_elev = 0.0;
-  if (heights.front() != NO_DATA_VALUE) {
+  if (heights.front() != baldr::kNoElevationDataRaw) {
```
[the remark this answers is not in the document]
nilsnolde: jep, sorry, you're right, reverted those changes and calling skadi::get_no_data_value() everywhere

src/tyr/trace_serializer.cc
nilsnolde: looks like those are stored in the tiles but not as `kNoElevationData`, that value is only reserved to be stored for elevation: https://github.com/valhalla/valhalla/blob/b889f94452ad9de4661a902cca2ef5f35b2b9229/src/mjolnir/elevationbuilder.cc#L120-L125

the triplegbuilder however already assigned `kNoElevationData` to the slopes when there's no elevation available in the graph tile: https://github.com/valhalla/valhalla/blob/master/src/thor/triplegbuilder.cc#L1092-L1105

happy to change to smth more sensible, for now I just made sure it's also properly used in the serializer and not returned as 32768 (or -500 now)

src/skadi/util.cc
```diff
@@ -71,7 +76,8 @@ weighted_grade(const std::vector<double>& heights,
   // Get the average weighted grade by homogenizing total weight. Return
   // max grades (up and down) and mean elevation.
   return std::make_tuple(total_grade * (1.0 / total_weight), max_up_grade, max_down_grade,
-                         (total_elev / n));
+                         (has_only_no_data ? static_cast<double>(get_no_data_value())
+                                           : total_elev / n));
```
kevinkreiser, line 80: its already a double

```suggestion
                         has_only_no_data ? get_no_data_value() : total_elev / n));
```

src/mjolnir/edgeinfobuilder.cc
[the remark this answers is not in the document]
kevinkreiser: yeah you are saying in practice -500 wont happen so no_data can be the same as min_value but like you mentioned at the beginning it does seem a bit semantically strange in the computation later where you do an offset by the no data value. i'd be happy just duplicating the constant even if they were the same value so that when you read the code you offset by the min-value rather than the no-data-value

src/thor/triplegbuilder.cc
kevinkreiser: its redundant now that you fixed it! it wasnt redundant in the past, previously this protected us from getting `1234` in the response

src/skadi/util.cc
nilsnolde: :facepalm: of course we can.. changed it.

src/tyr/trace_serializer.cc
kevinkreiser: oh i see, thanks for pointing that out! for whatever reason i had thought that we were talking about directed edges here but actually youre talking about the serialization downstream from here.

src/mjolnir/edgeinfobuilder.cc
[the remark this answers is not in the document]
nilsnolde: yeah true, wasn't very happy with that either but with the comment above the constant in graphconstants I found it ok-ish.. then let me duplicate that constant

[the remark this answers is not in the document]
nilsnolde: reverted the change to edgeinfo.h where kMinElevation was set before, which reverted all changes to edgeinfo.h and edgeinfobuilder.h. kMinElevation is now used where appropriate

is the internet broken today?! github is being ridiculously slow or doesn't respond at all after pushing, geofabrik same, curl actually times out downloading some pbf.. maybe it's telling me to go home instead..

nilsnolde:
fixed up the last minor things @kevinkreiser . thanks for the review

kevinkreiser (approved):

src/tyr/trace_serializer.cc
```diff
@@ -111,21 +111,32 @@ json::ArrayPtr serialize_edges(const AttributesController& controller,
...
         // Convert to feet if a valid elevation and units are miles
         float mean = edge.mean_elevation();
-        if (mean != kNoElevationData && options.units() == Options::miles) {
-          mean *= kFeetPerMeter;
+        if (mean == kNoElevationData) {
+          edge_map->emplace("mean_elevation", nullptr);
+        } else {
+          edge_map->emplace("mean_elevation", static_cast<int64_t>(options.units() == Options::miles
+                                                                       ? mean
+                                                                       : mean * kFeetPerMeter));
```
christian-nils, line 138: It seems that the inline if is not correct here.

christian-nils:
```suggestion
          edge_map->emplace("mean_elevation", static_cast<int64_t>(options.units() == Options::miles
                                                                       ? mean * kFeetPerMeter
                                                                       : mean ));
```

kevinkreiser: wow oops, i missed that. thanks for the catch and detta blir jag göra!

PR #3575, opened 2022-03-15
Const graph tile memory

molind:
# Issue

Tile data was loaded into memory without const modifiers and it was hard to trace if there was no erroneous modification of this memory somewhere. I set const for all objects that are loaded from this memory.


molind:
Ok. It's used in GraphBuilder and this change requires bigger refactor. I've checked only route computation part.

PR #3576, opened 2022-03-16
Additional Fixed Speed Costing Option

### LINKED ISSUE #3504: Isochrone calculations: use fixed vehicle speed?
Is there an easy way via costing options or other request options or some underlying service configuration to use the pedestrian/bicycle/vehicle's current known speed as the speed used in calculations rather than whatever gets derived from each OSM road segment?  I'm basically looking for a simple/more naive approach that assumes the vehicle is traveling at whatever speed it's currently traveling at independent of speed limits/assumptions about the road segments in question. The goal is to provide a near real-time measure of where a vehicle might be in the next few minutes if it continues to travel at its current rate.

jbartrug:
# Issue

#fixes #3504. Adds in the option to use a fixed speed parameter to override the calculated speed for truck, auto, and motorcycle. Went with the option to add a custom costing option. Unit test written and also tested with the command line. 

## Tasklist

 - [x] Add tests

## Requirements / Relations

No requirements or relations.


src/sif/autocost.cc
```diff
@@ -80,6 +80,8 @@ constexpr ranged_default_t<float> kUseDistanceRange{0, kDefaultUseDistance, 1.0f
 constexpr ranged_default_t<float> kAutoHeightRange{0, kDefaultAutoHeight, 10.0f};
 constexpr ranged_default_t<float> kAutoWidthRange{0, kDefaultAutoWidth, 10.0f};
 constexpr ranged_default_t<uint32_t> kProbabilityRange{0, kDefaultRestrictionProbability, 100};
+constexpr ranged_default_t<uint32_t> kFixedSpeedRange{0, 0, baldr::kMaxSpeedKph};
```
nilsnolde, line 83: overall I think you should move the whole logic into the parent class in `sif/dynamiccost`, similar to how it's done for `top_speed`. that'll remove a lot of redundant code. e.g. move this line here to `dynamiccost.cc:108`. I'll give some tips where to move the other parts as well.

src/sif/autocost.cc
```diff
@@ -340,6 +342,7 @@ class AutoCost : public DynamicCost {
   // Vehicle attributes (used for special restrictions and costing)
   float height_; // Vehicle height in meters
   float width_;  // Vehicle width in meters
+  float fixed_speed_;
```
nilsnolde, line 345: this goes to dynamiccost.h:961 or so

src/sif/autocost.cc
```diff
@@ -409,6 +412,7 @@ AutoCost::AutoCost(const Costing& costing, uint32_t access_mask)
   // Get the vehicle attributes
   height_ = costing_options.height();
   width_ = costing_options.width();
+  fixed_speed_ = costing_options.fixed_speed();
```
nilsnolde, line 415: this to dynamiccost.h.get_base_costs():1060

src/sif/autocost.cc
```diff
@@ -708,6 +717,8 @@ void ParseAutoCostOptions(const rapidjson::Document& doc,
   JSON_PBF_DEFAULT(co, false, json, "/include_hot", include_hot);
   JSON_PBF_DEFAULT(co, false, json, "/include_hov2", include_hov2);
   JSON_PBF_DEFAULT(co, false, json, "/include_hov3", include_hov3);
+  JSON_PBF_RANGED_DEFAULT(co, kFixedSpeedRange, json, "/fixed_speed", fixed_speed);
```
nilsnolde, line 720: redundant as you already set that in the parent class. actually you can remove the 3 lines above as well, they're redundant too

src/sif/dynamiccost.cc
```diff
@@ -105,6 +105,7 @@ constexpr ranged_default_t<float> kClosureFactorRange{1.0f, kDefaultClosureFacto
 
 constexpr ranged_default_t<uint32_t> kVehicleSpeedRange{10, baldr::kMaxAssumedSpeed,
                                                         baldr::kMaxSpeedKph};
+constexpr ranged_default_t<uint32_t> kFixedSpeedRange{0, 0, baldr::kMaxSpeedKph};
```
nilsnolde, line 108: ah you already put it here.. 

hm, not so sure if 0 is a good default, I'll let others comment on that

src/sif/motorcyclecost.cc
```diff
@@ -54,6 +54,7 @@ constexpr float kLeftSideTurnCosts[] = {kTCStraight,         kTCSlight,  kTCUnfa
 constexpr ranged_default_t<float> kUseHighwaysRange{0, kDefaultUseHighways, 1.0f};
 constexpr ranged_default_t<float> kUseTollsRange{0, kDefaultUseTolls, 1.0f};
 constexpr ranged_default_t<float> kUseTrailsRange{0, kDefaultUseTrails, 1.0f};
+constexpr ranged_default_t<uint32_t> kFixedSpeedRange{0, 0, baldr::kMaxSpeedKph};
```
nilsnolde, line 57: redundant

src/sif/motorcyclecost.cc
```diff
@@ -272,6 +273,7 @@ class MotorcycleCost : public DynamicCost {
   float toll_factor_;        // Factor applied when road has a toll
   float surface_factor_;     // How much the surface factors are applied when using trails
   float highway_factor_;     // Factor applied when road is a motorway or trunk
+  float fixed_speed_;
```
nilsnolde, line 276: redundant

src/sif/motorcyclecost.cc
```diff
@@ -291,6 +293,8 @@ MotorcycleCost::MotorcycleCost(const Costing& costing)
   // Get the base costs
   get_base_costs(costing);
 
+  fixed_speed_ = costing_options.fixed_speed();
```
nilsnolde, line 296: redundant

src/sif/truckcost.cc
```diff
@@ -72,6 +72,7 @@ constexpr ranged_default_t<float> kTruckHeightRange{0, kDefaultTruckHeight, 10.0
 constexpr ranged_default_t<float> kTruckWidthRange{0, kDefaultTruckWidth, 10.0f};
 constexpr ranged_default_t<float> kTruckLengthRange{0, kDefaultTruckLength, 50.0f};
 constexpr ranged_default_t<float> kUseTollsRange{0, kDefaultUseTolls, 1.0f};
+constexpr ranged_default_t<uint32_t> kFixedSpeedRange{0, 0, baldr::kMaxSpeedKph};
```
nilsnolde, line 75: redundant

src/sif/truckcost.cc
```diff
@@ -276,6 +277,7 @@ class TruckCost : public DynamicCost {
   float height_;    // Vehicle height in meters
   float width_;     // Vehicle width in meters
   float length_;    // Vehicle length in meters
+  float fixed_speed_; // Vehicle speed in kmph
```
nilsnolde, line 280: redundant

src/sif/truckcost.cc
```diff
@@ -302,6 +304,7 @@ TruckCost::TruckCost(const Costing& costing)
   height_ = costing_options.height();
   width_ = costing_options.width();
   length_ = costing_options.length();
+  fixed_speed_ = costing_options.fixed_speed();
```
nilsnolde, line 307: redundant

src/sif/truckcost.cc
```diff
@@ -634,6 +646,8 @@ void ParseTruckCostOptions(const rapidjson::Document& doc,
   JSON_PBF_RANGED_DEFAULT(co, kTruckWidthRange, json, "/width", width);
   JSON_PBF_RANGED_DEFAULT(co, kTruckLengthRange, json, "/length", length);
   JSON_PBF_RANGED_DEFAULT(co, kUseTollsRange, json, "/use_tolls", use_tolls);
+  JSON_PBF_RANGED_DEFAULT(co, kFixedSpeedRange, json, "/fixed_speed", fixed_speed);
```
nilsnolde, line 649: redundant

test/gurka/test_fixed_speed.cc
```diff
@@ -0,0 +1,78 @@
...
+TEST_F(FixedSpeedTest, ClampMaxSpeed) {
+  Options options;
+  rapidjson::Document dom;
+  rapidjson::SetValueByPointer(dom, "/fixed_speed", 500);
+
+  options.set_costing_type(Costing::auto_);
+  auto& co = (*options.mutable_costings())[Costing::auto_];
+  sif::ParseBaseCostOptions(*rapidjson::GetValueByPointer(dom, ""), &co, {});
+
+  ASSERT_EQ(co.options().top_speed(), baldr::kMaxAssumedSpeed);
```
nilsnolde, line 77: copy/paste fail?

src/sif/truckcost.cc
```diff
@@ -607,6 +616,9 @@ Cost TruckCost::TransitionCostReverse(const uint32_t idx,
 // assume the maximum speed is used to the destination such that the time
 // estimate is less than the least possible time along roads.
 float TruckCost::AStarCostFactor() const {
+  if(fixed_speed_ != 0){
+    return speedfactor_[fixed_speed_];
+  }
```
nilsnolde, line 621: makes sense to me. shouldn't we do that to the other costing models too?

src/sif/motorcyclecost.cc
```diff
@@ -579,6 +588,8 @@ void ParseMotorcycleCostOptions(const rapidjson::Document& doc,
   JSON_PBF_RANGED_DEFAULT(co, kUseHighwaysRange, json, "/use_highways", use_highways);
   JSON_PBF_RANGED_DEFAULT(co, kUseTollsRange, json, "/use_tolls", use_tolls);
   JSON_PBF_RANGED_DEFAULT(co, kUseTrailsRange, json, "/use_trails", use_trails);
+  JSON_PBF_RANGED_DEFAULT(co, kFixedSpeedRange, json, "/fixed_speed", fixed_speed);
```
nilsnolde, line 591: redundant

nilsnolde (changes requested):
looks good generally, thanks! you put in a few too many things, sorry for all the "redundant" comments, also check them again for consistency, might be I put one too many or missed one. basically everything can happen in the dynamiccost parent class and you just access `fixed_speed_` in the derived costing classes to set astar heuristic and edge cost.

I'd also wait for @kevinkreiser 's review (or someone else's). `quite sure this is the way to go, but maybe not. decreases the chance that you'd have to refactor again;)

nilsnolde:
what you can already do: run `./scripts/format.sh` locally and commit the changes it does. I can recommend adding it as a git commit hook:
```
cat ./scripts/format.sh > .git/hooks/pre-commit && tail -n +7 scripts/error_on_dirty.sh >> .git/hooks/pre-commit
chmod +x .git/hooks/pre-commit
```

src/sif/autocost.cc
nilsnolde: think you already did that after reviewing everything. so you mostly need to remove some stuff

kevinkreiser:
i agree with @nilsnolde we should do this in the base class and access it from all the other classes. this does bring up the question about bikes and pedestrians though and how they already have speeds that you can specify. we can rectify that later though

src/sif/dynamiccost.cc
jbartrug: Thanks for all the feedback! I really appreciate it. Just wanted to explain the reasoning for 0 as a default. My thinking was I needed a value that was outside the range of viable numbers, so that there wouldn't be any limitations on what fixed speed you could set.  Zero made sense because if the fixed speed  is zero then you don't need any fixed speed routing you're at a standstill, you should use the estimated speed instead. Totally open to other options that was just my base reasoning.

proto/options.proto
```diff
@@ -307,6 +307,9 @@ message Costing {
     oneof has_elevator_penalty {
       float elevator_penalty = 79;
     }
+    oneof has_fixed_speed {
+      float fixed_speed = 80;
+    }
```
kevinkreiser, line 312: since you use a default of `0` you can simply make this:

```suggestion
    float fixed_speed = 80;
```

src/sif/autocost.cc
```diff
@@ -275,6 +275,9 @@ class AutoCost : public DynamicCost {
    * estimate is less than the least possible time along roads.
    */
   virtual float AStarCostFactor() const override {
+    if (fixed_speed_ != 0) {
+      return speedfactor_[fixed_speed_];
+    }
```
kevinkreiser, line 280: rather than adding this complexity perhaps we should solve the problem of when people are  setting both of these and choose a winner. and whichever the winner is we should assign to `topspeed_` and let this code unchanged.

intuitively, when someone specifies `fixed_speed` then `top_speed` should get assigned the same value. does that make sense?

src/sif/autocost.cc
```diff
@@ -498,6 +501,11 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
   auto edge_speed = tile->GetSpeed(edge, flow_mask_, time_info.second_of_week, false, &flow_sources,
                                    time_info.seconds_from_now);
   auto final_speed = std::min(edge_speed, top_speed_);
+
+  if (fixed_speed_ != 0) {
+    final_speed = fixed_speed_;
+  }
+
```
kevinkreiser, line 508: in the case someone is using fixed speed its highly inefficient to still called `tile->GetSpeed()` just to throw the value away. maybe this would be better:

```suggestion
  auto edge_speed = fixed_speed_ != 0 ? tile->GetSpeed(edge, flow_mask_, time_info.second_of_week, false, 
                                   &flow_sources, time_info.seconds_from_now) : fixed_speed_;
  auto final_speed = std::min(edge_speed, top_speed_);
```

src/sif/dynamiccost.cc
kevinkreiser: i think this will work fine however i would suggest making `0` a named constant so the check in the actual costing will be more readible

kevinkreiser: one thing that is a bit strange from the API perspective is that you can actually set it to 0. i suppose the docs sh ould say valid values are from 1 to 160kph or whatever so people dont have to worry about what meaning 0 would have

src/sif/autocost.cc
kevinkreiser: I disagree. Fixed speed means only one speed. We want that disabled unless someone specified it.

nilsnolde: not sure what you disagree with haha I was just saying I agree with you, albeit not in a straight forward way

kevinkreiser: sorry i disagree that we should assign top_speed to fixed_speed. if we do that then when someone says top_speed = 45, then we'll have fixed_speed = 45 and then every edge will be 45. maybe i am misunderstanding what you are saying? my assumption was that if fixed_speed is set, ie not 0, then we will use fixed_speed for every edge. so setting it to top_speed when top_speed is passed in will essentially break the whole top_speed feature. what am i missing here? :smile:

kevinkreiser: ah maybe this is a language thing, let me write it in code, when i read what you said:

> assign top_speed to fixed_speed

my brain saw this code:

```c++
fixed_speed = top_speed;
```

nilsnolde: ah well yeah, I meant it the other way around :smile: `top_speed = fixed_speed` when both are passed

src/sif/dynamiccost.cc
jbartrug: Sounds good made it a named constant called kDefaultFixedSpeed

src/sif/autocost.cc
jbartrug: Agreed much more efficient. Changed to save that calculation.

proto/options.proto
jbartrug: When I changed to that format I got a compile error saying the member didn't exist in costing options so I left it as it was.

src/sif/autocost.cc
nilsnolde: @jbartrug you can revert this change and rather do `top_speed = fixed_speed` if `fixed_speed != kDefaultFixedSpeed` in `dynamiccost.h::get_base_costs()`

src/sif/autocost.cc
```diff
@@ -495,9 +498,13 @@ Cost AutoCost::EdgeCost(const baldr::DirectedEdge* edge,
                         const baldr::TimeInfo& time_info,
                         uint8_t& flow_sources) const {
   // either the computed edge speed or optional top_speed
-  auto edge_speed = tile->GetSpeed(edge, flow_mask_, time_info.second_of_week, false, &flow_sources,
-                                   time_info.seconds_from_now);
-  auto final_speed = std::min(edge_speed, top_speed_);
+  auto edge_speed = fixed_speed_ == 0
```
nilsnolde, line 501: should we use `kDefaultFixedSpeed` here as well?

nilsnolde: and all the other places I guess

test/gurka/test_fixed_speed.cc
```diff
@@ -0,0 +1,78 @@
...
+  doTests("auto", {{"/costing_options/auto/fixed_speed", "50"}});
+}
+
+TEST_F(FixedSpeedTest, TruckFixedSpeed) {
+  doTests("truck", {{"/costing_options/truck/fixed_speed", "50"}});
+}
+
+TEST_F(FixedSpeedTest, MotorcycleFixedSpeed) {
+  doTests("motorcycle", {{"/costing_options/motorcycle/fixed_speed", "50"}});
+}
```
nilsnolde, line 66: would you mind adding a test covering the scenario where both `top_speed` and `fixed_speed` are specified?

valhalla/sif/dynamiccost.h
```diff
@@ -1059,6 +1060,8 @@ class DynamicCost {
     flow_mask_ = costing_options.flow_mask();
     // Set the top speed a vehicle wants to go
     top_speed_ = costing_options.top_speed();
+    // Set the fixed speed a vehicle can go
+    fixed_speed_ = costing_options.fixed_speed();
```
nilsnolde, line 1064:
```suggestion
    // Set the fixed speed a vehicle can go
    fixed_speed_ = costing_options.fixed_speed();
    top_speed_ = fixed_speed_ == kDefaultFixedSpeed ? costing_options.top_speed() : fixed_speed_;
```

src/sif/autocost.cc
```diff
@@ -926,9 +933,11 @@ class TaxiCost : public AutoCost {
...
                         const baldr::TimeInfo& time_info,
                         uint8_t& flow_sources) const override {
-    auto edge_speed = tile->GetSpeed(edge, flow_mask_, time_info.second_of_week, false, &flow_sources,
-                                     time_info.seconds_from_now);
-    auto final_speed = std::min(edge_speed, top_speed_);
+    auto edge_speed = fixed_speed_ == 0
+                          ? tile->GetSpeed(edge, flow_mask_, time_info.second_of_week, false,
+                                           &flow_sources, time_info.seconds_from_now)
+                          : fixed_speed_;
+    auto final_speed = fixed_speed_ != 0 ? fixed_speed_ : std::min(edge_speed, top_speed_);
```
nilsnolde, line 940: edge_speed is already fixed_speed from above
```suggestion
    auto final_speed = std::min(edge_speed, top_speed_);
```

nilsnolde: same is in the other costing models

nilsnolde (changes requested):
looks good so far @jbartrug thanks! just a couple of minor things, I think then it's good to :ship:

src/sif/autocost.cc
nilsnolde: ah never mind this comment.. of course you want to fix the final speed.

test/gurka/test_fixed_speed.cc
jbartrug: Sure thing, just added that test in the latest commit along with your other suggestions thanks.

nilsnolde (approved):
lgtm. let's wait for someone else to have to a look before merging.

jbartrug:
@kevinkreiser Just wanted to check on the status of the review and merge of this MR. On my end it is ready to merge in but if there is anything else let me know what I can do.

valhalla/baldr/graphconstants.h
```diff
@@ -103,6 +103,11 @@ constexpr uint32_t kMaxSpeedKph = std::max(kMaxTrafficSpeed, kMaxAssumedSpeed);
 // stretch, its unlikely to be good signal below this value
 constexpr uint32_t kMinSpeedKph = 5; // ~3 MPH
 
+// Default Fixed Speed. This is the default fixed speed that is assumed.
+// Unless otherwised specified no fixed speed will be assumed and speed will be
+// calculated from costing algorithm.
+constexpr uint32_t kDefaultFixedSpeed = 0; // ~0 MPH
```
kevinkreiser, line 109: we should name this something more insightful. like `kDisableFixedSpeed`

docs/api/turn-by-turn/api-reference.md
```diff
@@ -113,6 +113,7 @@ These options are available for `auto`, `bus`, and `truck` costing methods.
 | `country_crossing_penalty` | A penalty applied for a country crossing. This penalty can be used to create paths that avoid spanning country boundaries. The default penalty is 0. |
 | `shortest` | Changes the metric to quasi-shortest, i.e. purely distance-based costing. Note, this will disable all other costings & penalties. Also note, `shortest` will not disable hierarchy pruning, leading to potentially sub-optimal routes for some costing models. The default is `false`. |
 | `top_speed` | Top speed the vehicle can go. Also used to avoid roads with higher speeds than this value. `top_speed` must be between 10 and 252 KPH. The default value is 140 KPH. |
+| `fixed_speed` | Fixed speed the vehicle can go. Used to override the calculated speed. Can be useful if speed of vehicle is known. `fixed_speed` must be between 0 and 252 KPH. The default value is 0 KPH. If not set it will default to using the calculated speed based on the roadways. |
```
kevinkreiser, line 116:
```suggestion
| `fixed_speed` | Fixed speed the vehicle can go. Used to override the calculated speed. Can be useful if speed of vehicle is known. `fixed_speed` must be between 1 and 252 KPH. The default value is 0 KPH which disables fixed speed and falls back to the standard calculated speed based on the road attribution. |
```

proto/options.proto
kevinkreiser: im sure we can fix that, we want to have as few `oneof`'s as possible. can you post back the compiler error?

kevinkreiser:
@jbartrug i think you forgot motorscooter cost. i suspect we want this for all motorized costings

nilsnolde (dismissed):
ready to merge I'd say:) sorry for the long wait @jbartrug

jbartrug:
Thanks @nilsnolde would love to merge whenever. Although it looks like it has a linting stage error. It looks like the same error is present on master as well. I manually ran the linting locally and it didn't find anything.

kevinkreiser:
Yeah it's not a linting error it's actually #3740

nilsnolde (dismissed):

nilsnolde:
hm, weird, it's failing the `fixed_speed` tests now.. will have a look later today, let's put it back in the queue for now

nilsnolde (approved):
this is done as well now

src/sif/dynamiccost.cc
```diff
@@ -461,6 +463,9 @@ void ParseBaseCostOptions(const rapidjson::Value& json,
   JSON_PBF_DEFAULT(co, cfg.include_hot_, json, "/include_hot", include_hot);
   JSON_PBF_DEFAULT(co, cfg.include_hov2_, json, "/include_hov2", include_hov2);
   JSON_PBF_DEFAULT(co, cfg.include_hov3_, json, "/include_hov3", include_hov3);
+
+  co->set_fixed_speed(
+      kFixedSpeedRange(rapidjson::get<uint32_t>(json, "/fixed_speed", co->fixed_speed())));
```
nilsnolde, line 468: old style parsing, because we opted not to make a `oneof` out of fixed_speed in the options.proto, since it defaults to 0 anyways.

kevinkreiser (approved):

PR #3578, opened 2022-03-17
Updates to country specific access overrides

### LINKED ISSUE #3577: Bicycle Access not allowed on highway=trunk in Romania
Wiki says bicycle access should be allowed on trunk roads in Romania: https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/.
Need to update adminconstants.h. To fix this will require building a new admin.sqlite DB and rebuilding data.

dnesbitt61:
In particular, add kBicycle access to trunk roads in Romania. Add a couple countries that did not previously have any
override logic. Also, not some that are in the OSM Wiki but do not have an entry here since there are no actual overrides at this time.

These changes impact building the admin.sqlite DB and they impact the country specific access logic that occurs in the graphenhancer stage.

# Issue

fixes #3577 

## Tasklist

 - [ ] Add tests

## Requirements / Relations


dnesbitt61:
As I read it, for Spain and Netherlands which have bicycles (and mopeds and pedestrians) allowed on tunk and trunk_link by default unless the motorroad tag is set. In countryaccess it first checks motorroad tag and access which by default is false for bicycle, moped, and pedestrian. Let me know if this is incorrect!

gknisely (approved):

dnesbitt61:
Tested this with Spain. On this branch I created a new admin.sqlite DB. I then built tiles for Spain. I was able to run bicycle routes on highway=trunk - except for cases where motorroad=yes is present. (Using overpass-turbo I found a good case on LE-20 east of Leon in northern Spain). Previously I was able to generate a bicycle route on LE-20 where motorroad was present. On this branch, bicycle routing is not allowed on this stretch of motorroad.

PR #3579, opened 2022-03-18
Improve build errors reporting

oleg-liatte:
# Issue

Improve build errors reporting

## Tasklist

 - [ ] Add tests

### DIFF
--- a/scripts/valhalla_build_timezones
+++ b/scripts/valhalla_build_timezones
@@ -1,18 +1,16 @@
 #!/bin/sh
 
 error_exit() {
-  echo "$1" 1>&2
-  #exit 1
+  echo "error: $1" 1>&2
+  exit 1
 }
 
 if  ! which spatialite >/dev/null; then
-    echo "spatialite not found which is required.  Please install via:  sudo apt-get install spatialite-bin"
-    exit 1
+    error_exit "spatialite not found which is required.  Please install via:  sudo apt-get install spatialite-bin"
 fi
 
 if  ! which unzip >/dev/null; then
-    echo "unzip not found which is required.  Please install via:  sudo apt-get install unzip"
-    exit 1
+    error_exit "unzip not found which is required.  Please install via:  sudo apt-get install unzip"
 fi
 
 rm -rf dist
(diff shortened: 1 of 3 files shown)


scripts/valhalla_build_timezones
```diff
@@ -1,18 +1,16 @@
 #!/bin/sh
 
 error_exit() {
-  echo "$1" 1>&2
-  #exit 1
+  echo "error: $1" 1>&2
+  exit 1
```
oleg-liatte, line 5: This comment seems to be committed accidentally.

scripts/valhalla_build_timezones
```diff
@@ -1,18 +1,16 @@
 #!/bin/sh
 
 error_exit() {
-  echo "$1" 1>&2
-  #exit 1
+  echo "error: $1" 1>&2
```
oleg-liatte, line 4: Add "error" to emphasize that this is a critical error (and to highlight it where applicable)

test/CMakeLists.txt
```diff
@@ -114,7 +114,8 @@ set_target_properties(logging PROPERTIES COMPILE_DEFINITIONS LOGGING_LEVEL_ALL)
 
 add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/test/data/tz.sqlite
   COMMAND ${CMAKE_COMMAND} -E make_directory test/data/
-  COMMAND ${VALHALLA_SOURCE_DIR}/scripts/valhalla_build_timezones 2>/dev/null > ${CMAKE_BINARY_DIR}/test/data/tz.sqlite
+  COMMAND ${VALHALLA_SOURCE_DIR}/scripts/valhalla_build_timezones > ${CMAKE_BINARY_DIR}/test/data/tz.sqlite.tmp
```
oleg-liatte, line 117: Don't hide stderr

test/CMakeLists.txt
```diff
@@ -114,7 +114,8 @@ set_target_properties(logging PROPERTIES COMPILE_DEFINITIONS LOGGING_LEVEL_ALL)
...
+  COMMAND ${CMAKE_COMMAND} -E rename ${CMAKE_BINARY_DIR}/test/data/tz.sqlite.tmp ${CMAKE_BINARY_DIR}/test/data/tz.sqlite
```
oleg-liatte, line 118: Create output file only in case of success. Otherwise subsequent builds will consider `tz.sqlite` fresh and do not attempt to recreate it

nilsnolde (approved):
makes sense to me, thanks:)

scripts/valhalla_build_timezones
kevinkreiser: The reason this is commented is because of recent versions of libgeos double freeing some memory. In others words, on some platforms this breaks the build. The double free happens at the end of the operation so it's harmless but erroring caused everything else to fail

scripts/valhalla_build_timezones
kevinkreiser: Sadly it was on purpose.

kevinkreiser: maybe we can detect the geos version and only `exit 1` if it was a newer one?

for example `pkg-config geos --modversion` returns `3.9.1` on my system and it fails because of double free but the output sqlite is perfectly fine

kevinkreiser: interestingly on github 3.9 doesnt even exist as a tag... so perhaps thats a sign ;)

kevinkreiser: in a month we'll have a new ubuntu version so maybe this will be rectified then..

kevinkreiser, line 5: this will allow users with geos 3.9 to not error out from double free

```suggestion
  exit $(pkg-config geos --modversion | grep -cvF "3.9")
```

oleg-liatte: Didn't get the problem. How geos affects exiting with failure if some step fails?

kevinkreiser (dismissed):

scripts/valhalla_build_timezones
kevinkreiser: yeah spatialite which is what we use here, links libgeos, certain versions 3.9.* seem to have a bug where they double free. this causes spatialite to crash which causes this script to fail. i commented out the exit because of this.

kevinkreiser (approved):

Issue #3010, opened 2021-04-16
Fix spatialite support for Windows


nilsnolde:
Currently it's not possible to build tiles with admin & timezone support because spatialite is not correctly set up for some reason. So tiles built on Windows can't 
- use time-dependent routing
- driving side will be off
- no border-crossing penalties

(maybe not an exhaustive list of disadvantages..)

Arkadipta-Chatterjee:
Can i work on this?

nilsnolde:
oh yes please, that'd be great! unfortunately we can't give you a lot of guidance here, the vast majority of people here never touches windows. but I can guide you at least to where the relevant code bits are:

1. valhalla has the ability to build admin (containing mostly country polygons and some related metadata like right/left handed traffic etc) & timezone sqlite db's, where only the former is built using c++, timezones are built using a bash script. it's not strictly necessary to be able to build admins from windows, but it'd be nice to also solve that:
this is the executable: https://github.com/gis-ops/valhalla/blob/master/src/mjolnir/valhalla_build_admins.cc, which you'd run with `valhalla_build_admins --config <your_valhalla.json> <your_osm_pbf>`, with likely Windows/MSVC not being able to set up this part: https://github.com/gis-ops/valhalla/blob/master/src/mjolnir/adminbuilder.cc#L261
2. once the admin db is built, valhalla will use while building the routing graph to enhance the node information and here it's essential to make it work on windows. this is the approximate flow:
    - `valhalla_build_tiles` is the executable which builds the graph
    - it calls a function which will execute several build stages consecutively: https://github.com/gis-ops/valhalla/blob/master/src/mjolnir/util.cc#L167
    - that function has 2 stages where the sqlite dbs will be opened:
    https://github.com/gis-ops/valhalla/blob/master/src/mjolnir/graphbuilder.cc#L400 (first usage building tiles)
    https://github.com/gis-ops/valhalla/blob/master/src/mjolnir/graphenhancer.cc#L1245 (second usage enhancing already built tiles)

so in the end it's mostly necessary to figure out how to open the spatialite dbs in this function: https://github.com/gis-ops/valhalla/blob/master/src/mjolnir/admin.cc#L14

how to set up your windows environment is in the [readme](https://github.com/valhalla/valhalla#building-from-source---windows) and/or in [CI](https://github.com/valhalla/valhalla/blob/master/.azure-pipelines.yml).

let us know if smth is unclear!

PR #3580, opened 2022-03-18
fix spatialite integration for windows

nilsnolde:
fiiinalllyyy fixes https://github.com/valhalla/valhalla/issues/3010

been bugging me for ages..

I went through a few demos and chose a more "modern" approach to load spatialite. only tested it down to spatialite 4.3.0, which should be enough, even ubuntu 16.04 has it. it's also still a little rough around the edges maybe, let's see what you think.

### DIFF
--- a/src/mjolnir/admin.cc
+++ b/src/mjolnir/admin.cc
@@ -16,18 +16,10 @@ sqlite3* GetDBHandle(const std::string& database) {
   // Initialize the admin DB (if it exists)
   sqlite3* db_handle = nullptr;
   if (!database.empty() && filesystem::exists(database)) {
-    spatialite_init(0);
-    std::string sql;
     uint32_t ret = sqlite3_open_v2(database.c_str(), &db_handle,
                                    SQLITE_OPEN_READONLY | SQLITE_OPEN_NOMUTEX, nullptr);
     if (ret != SQLITE_OK) {
       LOG_ERROR("cannot open " + database);
-      sqlite3_close(db_handle);
-      return nullptr;
-    }
-
-    // loading SpatiaLite as an extension
-    if (!load_spatialite(db_handle)) {
       sqlite3_close(db_handle);
       return nullptr;
     }
--- a/src/mjolnir/adminbuilder.cc
+++ b/src/mjolnir/adminbuilder.cc
@@ -10,8 +10,8 @@
 #include "mjolnir/pbfadminparser.h"
 #include "mjolnir/util.h"
 
-// sqlite is included in util.h and must be before spatialite
 #include <spatialite.h>
+#include <sqlite3.h>
 
 #include "config.h"
 
@@ -256,8 +256,6 @@ void BuildAdminFromPBF(const boost::property_tree::ptree& pt,
     filesystem::remove(*database);
   }
 
-  spatialite_init(0);
-
   sqlite3* db_handle;
   sqlite3_stmt* stmt;
   uint32_t ret;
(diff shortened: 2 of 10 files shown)


src/mjolnir/admin.cc
```diff
@@ -3,10 +3,11 @@
 #include "filesystem.h"
 #include "midgard/logging.h"
 #include "mjolnir/util.h"
-#include <spatialite.h>
 #include <sqlite3.h>
 #include <unordered_map>
 
+#include <spatialite.h>
```
nilsnolde, line 9: need to put those at the end as sqlite3.h needs to be included before spatialite.h. otherwise format.sh will simply switch the includes

nilsnolde: also I remove all references to sqlite/spatialite from mjolnir/util.h so we now get a lot more includes in cc files

src/mjolnir/adminbuilder.cc
```diff
@@ -273,10 +271,8 @@ void BuildAdminFromPBF(const boost::property_tree::ptree& pt,
   }
 
   // loading SpatiaLite as an extension
-  if (!valhalla::mjolnir::load_spatialite(db_handle)) {
-    sqlite3_close(db_handle);
-    return;
-  }
+  auto* db_cache = spatialite_alloc_connection();
+  spatialite_init_ex(db_handle, db_cache, 0);
```
nilsnolde, line 275: this is basically the new syntax. `spatialite_init(0)` is deprecated. we need to hold a pointer to the connection's address

src/mjolnir/adminbuilder.cc
```diff
@@ -639,6 +635,8 @@ void BuildAdminFromPBF(const boost::property_tree::ptree& pt,
   }
 
   sqlite3_close(db_handle);
+  spatialite_cleanup_ex(db_cache);
+  spatialite_shutdown();
```
nilsnolde, line 639: cleanup. you'll see this pattern thoughout all files using spatialite.

nilsnolde: spatialite_shutdown I just saw in the demos, can't hurt I guess..

test/win/valhalla.json
```diff
@@ -99,7 +99,7 @@
     "verbose": false
   },
   "mjolnir": {
-    "admin": "",
+    "admin": "test\\data\\netherlands_admin.sqlite",
```
nilsnolde, line 102: trying it out on azure

nilsnolde:
Huh strange.. of all things, the Python bindings build triggers a leak in build_tile_set()..

nilsnolde:
I took a stab at implementing your suggestion, thanks for that! much cleaner.

I thought I'd include it all in a private header in ./src/mjolnir/spatialite_conn.h, but the linker complained that there are multiple definitions in graphbuilder.cc and graphenhancer.cc, though I didn't see what might be causing that.

kevinkreiser:
@nilsnolde you need to make the functions inline or make a cc file and put the guts of the implementation in there with just the definitions in the header

nilsnolde:
ah inlining would've helped there? good to know.. but yeah, I went for your second suggestion, .h/.cc file.

CHANGELOG.md
```diff
@@ -75,6 +76,7 @@
    * ADDED: Expose reverse isochrone parameter for reverse expansion [#3528](https://github.com/valhalla/valhalla/pull/3528)
    * CHANGED: Add matrix classes to thor worker so they persist between requests. [#3560](https://github.com/valhalla/valhalla/pull/3560)
    * CHANGED: Remove `max_matrix_locations` and introduce `max_matrix_location_pairs` to configure the allowed number of total routes for the matrix action for more flexible asymmetric matrices [#3569](https://github.com/valhalla/valhalla/pull/3569)
+   * CHANGED: modernized spatialite syntax [#3580](https://github.com/valhalla/valhalla/pull/3580)
```
nilsnolde, line 79: I somehow felt like 2 entries, let me know if I should remove one

kevinkreiser (approved):

Issue #3582, opened 2022-03-23
Unidirectional A* variants have issues with multiple destination edges


derolf:
In SetOrigin and ExpandInternal, the Uni A* checks if a new edge E is a destination edge.

Let’s consider the case where E is a destination edge.

Then the cost of E is adjusted (minus percentage, plus lateral distance) and inserted into the queue. If the lateral distance is high, the adjusted cost of E might be much bigger than the vanilla cost of E.

However, it could be that the actual shortest path travels along E to another destination edge F with a much smaller lateral distance and hence much smaller total cost.

Because of the adjusted cost of E, it is impossible for any such shortest path to travel along E towards F (costs are distorted). So any such path will not be found, resulting in strange detours around the destination: They will avoid E and try to reach F via a detour.

NOTE: We have a workaround in our codebase that fixes the issue: We always inserts a vanilla edge label into the queue first. If the edge is also a destination edge, we compute the adjusted cost and add another edge label to the queue and mark it as a destination edge label. When pulling from the queue, we check if it’s a destination edge label and terminate the search.

derolf:
Pretty sure that this is the root cause for https://github.com/valhalla/valhalla/issues/1952

kevinkreiser:
@derolf excellent suggestion! this issue has been plaguing us for years, how to penalize candidate start/end edges appropriately. the suggestion sounds straight forward and requires only one bit on the edge label. we should write a unit test and give this a try!

cc @dnesbitt61

derolf:
Thanks @kevinkreiser !

Right now our “flag” is to ride on the fact that destination edge labels have dist = 0.0. However that feels a bit like a hack and I only did it to keep the changes local since edge label is used at many places. What do you think?

kevinkreiser:
@derolf i dont have strong feelings that that would be "hacky" having path distance to destination being 0 on the destination edge is kind of intuitive, but also we have a lot of spare bits in edgelabel at the moment and we could easily just taking one of those and make an explicitly named flag with a nice comment explaining what it means.

i'd say if you are already working on this then throw up a pr and we can have a few folks make comments on it and get it merged, thanks again for bringing this up!

derolf:
Guten Abend @kevinkreiser !

So, I added some tests in: https://github.com/derolf/valhalla/commit/0429fabdb08c2e995f08eb3e179da63c1cef8cc6

[...]

Issue #1952, opened 2019-09-25
large search radius causes strange behavior... 


xlqian:
Hello Valhalla Team, 
it's Navitia again... I just found a strange behavior when working on BSS (still WIP, but I have a first version that returns correct journeys, i'll make a Draft/PR very shortly)

Context: 
In Navitia, we have to use a relatively large search radius(ex. 30m) to project the railway stations onto the map, because some of our stations' GPS coordinate may be found among railways, which may be very far from roads. 
![image](https://user-images.githubusercontent.com/6093486/65611358-fa6b0080-dfb2-11e9-94ee-c49dccf321f6.png)

With small search radius, we won't be able to find any solutions to get to those stations by car/bicycle.

The bug:

With that said, here is a simple route with 30m as search radius when projecting:
![image](https://user-images.githubusercontent.com/6093486/65612509-af51ed00-dfb4-11e9-803a-b728064456c0.png)
Everything goes well in the above route.

Now we move the destination slightly to the left and we get this:
![image](https://user-images.githubusercontent.com/6093486/65612602-d7415080-dfb4-11e9-958d-0c80b9c8cc6e.png)

We actually find the reason why the algo behaved like this. 

First, because of the large search radius, besides those "correct" projections, the destination is also projected to the endnode of the cross of "rue des francs bourgeois" and "rue payenne", kind of bad projection and bad projections are very heavily penalized by this line:
https://github.com/valhalla/valhalla/blob/master/src/thor/worker.cc#L250

Then, this penalization is added to the Edgelabel at this line:
https://github.com/valhalla/valhalla/blob/master/src/thor/astar.cc#L177

The penalization is actually added to the edgelabel which is best solution, then this edge will be ignored in the next search because of the heavy penalization which makes it be very far from the top of the priority queue.

We actually have a solution for this but the dev will be complicated and we want hear from you if you have any workaround for this issue.

 
thank you

kevinkreiser:
@xlqian we have struggled with this issue for years. how to properly penalize different edge candidates so that route will choose the one that makes the most sense. sometimes we find that the penalization should be different based on the applications assumptions. in fact we added even a request feature to disable this penalization per location (`rank_candidates`).

in our first implementation we penalized all of the edges by the square of the distance of the snap location. this worked pretty well except that the high penalty made the algorithm do much more expansion than it would have to do normally.

[...]

PR #3585, opened 2022-03-29
Multiple origin and destination candidates results in strange detours in unidirectional A*

derolf:
- Added tests for multiple origin and destination candidates to reproduce the issue -> tests fail
- Fixed unidirectional A* to properly handle these cases -> tests pass

# Issue

https://github.com/valhalla/valhalla/issues/3582
https://github.com/valhalla/valhalla/issues/1952

## Tasklist

 - [x] Add tests


derolf:
Dear reviewers, 

the basic idea is to wrap label creation into a lambda (`add_label`) in both `SetOrigin` and `ExpandInner`. `add_label` is always called to at least add the "normal" edge. If the edge is also a destination, `add_label` is called for each correlation on that destination, adjusting costs.

Cheers

Daniel

derolf:
The test `Standalone, NotThruPruning` fails. I looked through my changes and can't figure out what's wrong there. Any idea?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Release Date: 2021-??-?? Valhalla 3.1.5
 * **Removed**
 * **Bug Fix**
+   * FIXED: Strange detoures with multi-origin/destination unidirectional A* [#3585](https://github.com/valhalla/valhalla/pull/3585)
```
kevinkreiser, line 4: can you put this at the bottom of the list, we have them in descending chronological order

kevinkreiser:
@dgearhart it looks like the pencilpoint tests are failing on this one with finding completely different routes? i just looked at ci and saw the logs and thats what it looked like. do you have a second to take a look? otherwise i can look maybe in a couple of days.

@derolf thanks very much for the massive refactor of these algorithms. i'm hoping that @dnesbitt61 and i can take a look in the next few days. anyone else who has time (maybe  @purew ? ) would also be much appreciated!

valhalla/sif/edgelabel.h
```diff
@@ -270,6 +271,21 @@ class EdgeLabel {
...
+  /**
+   * Is this edge a destination edge?
+   * @return  Returns true if this edge is a destination edge.
+   */
+  bool destination() const {
+    return destination_;
+  }
+
+  /**
+   * Sets this edge as an origin.
```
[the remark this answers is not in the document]
derolf: An edge can be both origin and destination at the same time.

dgearhart:
> @dgearhart it looks like the pencilpoint tests are failing on this one with finding completely different routes? i just looked at ci and saw the logs and thats what it looked like. do you have a second to take a look? otherwise i can look maybe in a couple of days.

@kevinkreiser will not have any time today

valhalla/sif/edgelabel.h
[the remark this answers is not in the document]
kevinkreiser: sorry im just pointing out that the comment says "set this edge as an origin" but the code sets the destination bit. that seems contradictory though yes i understand that its possible an edge is both a destination and an origin for certain routes. really i just wonder why the comment isnt "sets this edge as a destination"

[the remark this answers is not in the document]
derolf: Ah, my bad!!!! Sorry

kevinkreiser:
@derolf apologies its taken so long to get to this, but i just wanted to say its still on my radar. just have to appease the tests at this point.. bear with us

derolf:
Hey @kevinkreiser,

Any idea how to fix the tests?

Assuming I would fix them in the next days, would it be ready to get merged?

cheers

Daniel

kevinkreiser:
@derolf we were just starting to look at them last week to see what is going on. As soon as I can qualify the exact cause I'll post back to this issue. All I see at the moment is it takes a different path. It might be that the test itself should change

derolf:
@kevinkreiser Both the old algorithm and the test are buggy:

- Pencil point turns are turns from an edge that is traversed shorter than 8m (kShortInternalLength)
- In the test, the C-D edge in the test is over 40m long. So that's actually NOT a pencil point turn.
 
- The old algorithm by chance (bug) didn't take the left turn, so the test passed
- My bug fix made the algorithm take the turn, so the test failed

- I fixed the test by making the C-D edge shorter than 8m, now the test passes

nilsnolde:
yayy, thanks @derolf ! we'll look at it latest next week. in the meantime we installed `pre-commit` to run the format.sh script automatically. if you run `./scripts/format.sh` you'll get pre-commit installed automatically and don't have to worry about it anymore.

derolf:
Looks like a mapmatching case fails, on it...

derolf:
@kevinkreiser  there are some slight deviations in recosting:

```
/home/ubuntu/mac/dev/valhalla/test/gurka/test_recost.cc:199: Failure
  n.cost().elapsed_cost().cost()
    Which is: 31.8683
  n.recosts(0).elapsed_cost().cost()
    Which is: 31.8682
/home/ubuntu/mac/dev/valhalla/test/gurka/test_recost.cc:199: Failure
  n.cost().elapsed_cost().cost()
    Which is: 39.1633
  n.recosts(0).elapsed_cost().cost()
    Which is: 39.1632
fJqhHgJncE?t`@?fJ?dJ?fJ?xO?zO
/home/ubuntu/mac/dev/valhalla/test/gurka/test_recost.cc:283: Failure
  n.cost().elapsed_cost().cost()
    Which is: 168.806
  n.recosts(0).elapsed_cost().cost()
    Which is: 168.806
/home/ubuntu/mac/dev/valhalla/test/gurka/test_recost.cc:283: Failure
  n.cost().elapsed_cost().cost()
    Which is: 192.166
  n.recosts(0).elapsed_cost().cost()
    Which is: 192.166
[  FAILED  ] recosting.same_historical (2400 ms)
```

I will just used EXPECT_NEAR instead of EXPECT_EQ!?

derolf:
@nilsnolde @kevinkreiser Coverage went down a tiny bit. Any idea for an easy gain here?

nilsnolde:
Ah that's fine @derolf. Can happen with bigger refactors where lots of (apparently very well tested) code is removed. As for me, the more important thing is the patch diff and that looks as good as it can be. 

Though I have a hard time wrapping my head around the overall decrease if the patch diff is so high. I could understand if the patch diff was like 84.9 or so, then I could reason that code was removed that was covered at a much higher degree. But with 96% coverage.. Anyways, maybe I understand those metrics wrong. Again, it's fine with me at least.

derolf:
@nilsnolde Ship it!

nilsnolde:
jep, we'll review it again in a bit. I'm pretty sure it's good though! was just saying it's fine to decrease coverage. thanks again and sorry for the long wait.

src/thor/unidirectional_astar.cc
```diff
@@ -194,150 +194,169 @@ inline bool UnidirectionalAStar<expansion_direction, FORWARD>::ExpandInner(
...
-                                  nodeinfo->timezone(), restriction_idx) ||
-        costing_->Restricted(meta.edge, pred, edgelabels_, tile, meta.edge_id, false, &edgestatus_,
-                             time_info.local_time, nodeinfo->timezone())) {
-      return false;
-    }
+  if (!FORWARD) {
+    opp_edge_id = endtile->GetOpposingEdgeId(meta.edge);
+    opp_edge = endtile->directededge(opp_edge_id);
+    assert(opp_edge_id);
+    assert(opp_edge);
```
nilsnolde, line 207: if we have the tile, it's not possible that we don't have the edge

nilsnolde (approved):
Finally had some time to look into this. Looks good, thanks @derolf . I think the `assert` are redundant, other than that all good. Will go over it with @kevinkreiser tmrw, then we can merge.

src/thor/unidirectional_astar.cc
```diff
@@ -649,50 +652,59 @@ void UnidirectionalAStar<expansion_direction, FORWARD>::SetOrigin(
...
-    if (FORWARD) {
-      if (costing_->AvoidAsOriginEdge(edgeid, edge.percent_along())) {
-        continue;
-      }
+    // Disallow any user avoided edges if the avoid location is behind the destination along the
+    // edge (check has to be done BEFORE we invert the edge if !FORWARD)
+    if (FORWARD ? costing_->AvoidAsOriginEdge(edgeid, percent_along)
+                : costing_->AvoidAsDestinationEdge(edgeid, percent_along)) {
+      continue;
     }
```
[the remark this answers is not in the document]
nilsnolde: Hm, not so convinced anymore after all.. For **reverse** this might not do what we'd expect I think. Basically it seems this relies on the edge as it's being traversed by the algorithm, not its opposing. That exact edge might not even be in the `avoid_edges` set if the current mode isn't allowed on it. However, its opposing (the one we actually care about) might've been. That is to say, it's better than before, where we didn't look at avoided edges at all for the reverse tree, but might still need some more fiddling to get it really right, so we should leave a TODO.

nilsnolde (approved):

derolf:
Thanks for approving the PR. What would you like me to do before you merge the PR?

kevinkreiser:
I'd still like to test the pr. We need to run a suite of routes through it and look at the differences vs master. I'll do that and report back. I'd also like to reason about the changes myself i just haven't had the time. I started looking weeks ago but other things came up.

nilsnolde:
Maybe we can do that tmrw together if there’s enough time.

nilsnolde:
Maybe we can do that tmrw together if there’s enough time.

valhalla/thor/unidirectional_astar.h
```diff
@@ -185,8 +185,9 @@ class UnidirectionalAStar : public PathAlgorithm {
   // Edge status. Mark edges that are in adjacency list or settled.
   EdgeStatus edgestatus_;
 
-  // Destinations, id and percent used along the edge
-  std::unordered_map<uint64_t, float> destinations_percent_along_;
+  // Mark if edge is a destination
+  std::unordered_multimap<valhalla::baldr::GraphId, std::reference_wrapper<const valhalla::PathEdge>>
```
kevinkreiser, line 189: im not too sur that we need a multimap here. the correlated edges can only have one copy of each unique edge id so you wont ever push in the same edge twice as a destination. unless im missing something here.

derolf: That's not correct. The map-matcher could match the same edge multiple times. Just think about a hard loop shaped like a U with the car between.

kevinkreiser: while it is true that map matching needs to handle this case, *this* algorithm has nothing to do with our map matching implementation. all our map matching code is in the `meili` namespace (and a little bit of it in thor/map_matcher thor/trace_*). this is because we do a one-to-many routing between map matching candidates and this algorithm only does a one-to-one routing.

valhalla/sif/edgelabel.h
```diff
@@ -441,6 +457,7 @@ class EdgeLabel {
   uint64_t shortcut_ : 1;
   uint64_t dest_only_ : 1;
   uint64_t origin_ : 1;
+  uint64_t destination_ : 1;
```
kevinkreiser, line 460: this is convenient good call

valhalla/thor/unidirectional_astar.h
derolf: I mean if the incoming destination set has the same edge multiple times in the list, then we would need to add complex logic for tie breaking. Using a multimap usually nicely handles that case without any other checks required.

kevinkreiser: yep sorry what im saying is that that cannot happen. the list is populated by this line: https://github.com/valhalla/valhalla/blob/4532f7be1910349bb30ccdad36fb2e9c1d930881/src/thor/unidirectional_astar.cc#L822

that list in turn is populated by loki search and im just telling you that loki search will only have unique edge ids in the list. this is because the whole job of loki search is to find the closest point on a given edge and return that as a candidate. there can be only 1 closest point on a given edge per location. so im just saying that multimap is an unneeded complication. i can see you are using it to nicely handle the event that there would be the same id more than once but what im saying is that there will never be.

derolf: One should not couple implementation details of two algorithms, esp. across public APIs.

So relying on the fact that LOKI behaves that way is not good because the A* could also be used in a different way.

kevinkreiser:
so we ran master and this branch on a 14k route file and the diff results were good for depart at. only 351 diffs total. we will run arrive by and also look at differences in performance. i'll post some of the screen shots of the differences so we can see what kinds of diffs we have. the routes i ran were with a 10m radius so that the logic to select different destination candidates would be triggered. looking good so far! thanks again for all the work @derolf

kevinkreiser:
ok i've checked performance and as expected there is no impact (we are doing the same amount of unordered::finds as we were before basically and the lambda is basically free).

@nilsnolde and i looked at a bunch of the diffs and just about all of them make sense and are better results. there are a few where i'd  say the results are worse and not intuitive but on the whole the ones we checked (maybe 20 manually reviewed) id say only a couple were noticeably worse.

algorithmically the change makes sense to me so i think we are good to merge.

valhalla/thor/unidirectional_astar.h
kevinkreiser: one should probably also not constantly change the interface to those public APIs but i guess it depends on the guarantees one wants to provide with a given API.

since its inception the project has not really concerned itself with maintaining compatibility (commit to commit) between the APIs exposed in the headers. we pretty much only gaurantee the json API. in fact it is a long term goal to vastly reduce the scope of methods exposed in the library headers (a battle for another day.. or maybe slowly plugging away at it rather).

furthermore, i can tell you for a fact that other parts of thor assume that there are no duplicates in the list of a locations edge candidates (eg: https://github.com/valhalla/valhalla/blob/master/src/thor/triplegbuilder.cc#L411 if there were duplicates here it would randomly pick one candidate, not necessarily the right one)

anyway the whole reason i was telling you this was because i saw the `multimap` and i saw the call to `equal_range`  and i was wondering: why is he protecting against this case that doesn't occur? which lead to me really just trying to do a polite: FYI you dont need to worry about this, it cant happen. but i think from your round about answers afterward i could understand that you may be calling this algorithm with your own set of edge candidates which may indeed have duplicates?

my point is i understand how one *could* put more than one correlation point for the same edge in the list, it just doesnt happen in any of the code here on github. if you have other designs elsewhere or future plans, best to just say that and shortcut the discussion :smile:

kevinkreiser (approved):

valhalla/thor/unidirectional_astar.h
derolf: Nay, no plans right now, but could happen.

I did that change for the simple reason of my edge-cases-tuned brain :-)

PR #3586, opened 2022-03-29
fix trace mean elevation serialization

kevinkreiser:
as mentioned here https://github.com/valhalla/valhalla/pull/3571#discussion_r837481014 by @christian-nils


nilsnolde:
argh :sweat_smile: tusen takk!!

PR #3588, opened 2022-04-05
Add serialized warnings array to JSON output

### LINKED ISSUE #3099: Warnings section in response

MuluhGodson:
# Issue
fixes #3099 

## Testing 
```sh
cmake -B build -DENABLE_TESTS=ON
cd build
make run-gurka_warnings
```

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/tyr/serializers.cc
```diff
@@ -157,6 +158,18 @@ void openlr(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper
   writer.end_array();
 }
 
+void serializeWarnings(const valhalla::Api& api, rapidjson::writer_wrapper_t& writer){
```
nilsnolde, line 161: so, `warnings` could (and should eventually) be added for all actions/endpoints. and the problem there is that some (or even all) other actions like isochrones/locate etc use another syntax/model for JSON. the one you see above is based on rapidjson. there's another one in baldr/json.h, which is a homegrown JSON builder.

so to support all endpoints, we'll probably have to overload this function. at which point I'm not sure it has such merit to even have it centrally in `serializers.cc`, but let's go with that for now.

src/tyr/serializers.cc
```diff
@@ -157,6 +158,18 @@ void openlr(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper
...
+    valhalla::Info warning;
```
nilsnolde, line 162: does this work for you?.. don't you have to use the `api` above to see the warning which were added during processing?

src/tyr/serializers.cc
```diff
@@ -157,6 +158,18 @@ void openlr(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper
...
+    writer.start_array("warnings");
+    writer.start_object();
+    for(int i = 0; i < warning.warnings_size(); i++){
+       writer(std::to_string(i), warning.warnings(i).description());
+    }
+    writer("warning_count", std::to_string(warning.warnings_size()));
```
nilsnolde, line 168: the whole format can be like 
```
"warnings": [
  "warning 1",
  "warning 2"
]
```
you can ignore the codes for now.

src/worker.cc
```diff
@@ -640,6 +651,11 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
   // auto_data_fix is deprecated and will be turned into
   // ignore all the things costing option. maybe remove in v4?
   if (costing_str == "auto_data_fix") {
+
+    // warning for auto data fix
+    valhalla::Info info;
+    info.add_warnings()->set_description("auto_data_fix is deprecated and will be turned to ignore all the things costing option");
```
nilsnolde, line 657: I'd try somewhere at the top: `auto* warnings = api->mutable_info()->mutable_warnings()` and then when you want to set a warning
```
auto* warn = warnings->Add()->set_description
```

nilsnolde: untested

src/tyr/serializers.cc
MuluhGodson: I've refactored the function to use the `api` directly.

src/worker.cc
MuluhGodson: I've refactored this to include a function. Then I just call the function whenever I need to add a warning.

src/tyr/serializers.cc
MuluhGodson: Refactoring the output of the warning as a bare array, the parser still assigns keys to each warning.
image

src/worker.cc
```diff
@@ -153,6 +153,13 @@ rapidjson::Document from_string(const std::string& json, const valhalla_exceptio
   return d;
 }
 
+// fuction to add warnings to api object
+void add_warning(valhalla::Api& api, std::string warning_text){
+  
+  auto* warning = api.mutable_info()->mutable_warnings()->Add();
+  warning->set_description(warning_text);
+}
```
nilsnolde, line 161: good impuls! as long as we're just using it in this one function, we can also make it a lambda inside the function and you can capture the `api` pointer from the outer scope.

src/tyr/serializers.cc
nilsnolde: well, I actually really meant a flat array of strings. no object. as long as we're not actually using codes for warnings there's no point numbering them as an array is inherently numbered and any json parser will know the length.

nilsnolde: ah just kinda reading the first example. the array is right, but of course without the numbers in there;)

src/tyr/route_serializer_valhalla.cc
```diff
@@ -4,6 +4,7 @@
 #include "midgard/aabb2.h"
 #include "midgard/logging.h"
 #include "odin/util.h"
+#include "proto/info.pb.h"
```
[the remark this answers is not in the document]
MuluhGodson: Ah yes. 
I'll do a quick clean up and remove them.

src/worker.cc
```diff
@@ -517,6 +517,12 @@ void parse_contours(const rapidjson::Document& doc,
  */
 void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
   // if its a pbf request we want to keep the options and clear the rest
+
+  // lambda fuction to add warnings to api object
+  auto add_warning = [](Api& api, const std::string& warning_text) {
+    api.mutable_info()->mutable_warnings()->Add()->set_description(warning_text);
+  };
```
kevinkreiser, line 524: seems like we can use this anywhere we want to add  a warning potentially? maybe we should just add this as a function in worker.h so anyone can call it from anywhere

valhalla/tyr/serializers.h
```diff
@@ -16,6 +16,7 @@
 #include <valhalla/meili/match_result.h>
 #include <valhalla/midgard/gridded_data.h>
 #include <valhalla/proto/api.pb.h>
+#include <valhalla/proto/info.pb.h>
```
kevinkreiser, line 19: again superfluous

src/tyr/serializers.cc
```diff
@@ -157,6 +157,26 @@ void openlr(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper
   writer.end_array();
 }
 
+void serializeWarnings(const valhalla::Api& api, rapidjson::writer_wrapper_t& writer) {
+  writer.start_array("warnings");
+  for (int i = 0; i < api.info().warnings_size(); i++) {
+    writer(std::to_string(i), api.info().warnings(i).description());
+  }
+  writer.end_array();
```
kevinkreiser, line 165: if we just do an array of strings and we want to use codes with the strings in the future then this will be a breaking change. so we have to make the decision now whether or not we want coded warnings.

i see why for deprecation we wouldnt need a code. its a warning that is fairly unimportant and at some unspecified date in the future may disappear. however there are other more important and likely permanent warnings about limiting input parameters (max radius for example) or when we need to do a two pass run of a route because we needed to relax hierarchy constraints. i think having codes for messages makes them obviously way easier to detect on the client side and also doesnt hurt us for the less important messages. thoughts?

MuluhGodson: I agree. For simple deprecation notices a plain description will suffice. 

However, for the more permanent warnings, what format will these codes take, so that they aren't confused with error codes? 

We have error codes already from 100 up to 503. (Not all values in-between are used).

src/worker.cc
MuluhGodson: Okay. I've pushed this update in the latest commits.

src/tyr/route_serializer_valhalla.cc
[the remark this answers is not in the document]
MuluhGodson: done.

valhalla/tyr/serializers.h
MuluhGodson: This has been fixed in the latest commit.

src/tyr/serializers.cc
MuluhGodson: :smile: i just pushed the latest commit with the numbers removed.

src/tyr/serializers.cc
MuluhGodson: done..

valhalla/worker.h
```diff
@@ -74,6 +74,12 @@ void ParseApi(const prime_server::http_request_t& http_request, Api& api);
 #endif
 
 std::string serialize_error(const valhalla_exception_t& exception, Api& options);
+
+// lambda function to add warnings to proto info object
+const auto add_warning = [](Api& api, const std::string& warning_text) {
```
kevinkreiser, line 79: we should convert this to a proper function rather than a lambda

src/tyr/serializers.cc
kevinkreiser: right but what about the larger point that for things that are important to know but dont stop the request from being processed, i think its important to use a code to distinguish it aside from the text so that its easier to maintain on the client side. with the current implementation in this pr its just an array of strings. can we make it an array of objects, and inside each object we have a code and a string. then `add_warning` can take the code as a parameter and look up the warning from the full list, similar to how the `valhalla_exception_t` stuff works

MuluhGodson: Yes this is a great reason. 
My question is what format should the codes take so as not to confuse them with error codes or HTTP codes on the client side or is it inconsequential?

valhalla/worker.h
MuluhGodson: Okay. I'll update that in the next commit.

src/tyr/serializers.cc
nilsnolde: there's no way of confusion, all those things are different entities. you can just add a `unordered_map` with `{int code, string msg}`, put that in src/worker.cc (or .h, I personally don't really mind), then adapt your `add_warning` code

MuluhGodson: > there's no way of confusion, all those things are different entities. you can just add a `unordered_map` with `{int code, string msg}`, put that in src/worker.cc (or .h, I personally don't really mind), then adapt your `add_warning` code

Yes this is what I am currently implementing.

docs/api/turn-by-turn/overview.md
```diff
@@ -6,7 +6,7 @@ The Valhalla route service (a.k.a. turn-by-turn) is an open-source routing servi
...
 
-The service [route results](api-reference.md#outputs-of-a-route) provide details about the trip, including locations, a summary with basic information about the entire trip, and a list of legs. Each leg has its own summary, a shape, which is an encoded polyline of the route path, and a list of maneuvers. These maneuvers provide written narrative instructions, plus verbal alerts that can be used as audio guidance in navigation apps.
+The service [route results](api-reference.md#outputs-of-a-route) provide details about the trip, including locations, a summary with basic information about the entire trip, a list of legs and a warnings array if any deprecated parameters were included in the request. Each leg has its own summary, a shape, which is an encoded polyline of the route path, and a list of maneuvers. These maneuvers provide written narrative instructions, plus verbal alerts that can be used as audio guidance in navigation apps.
```
nilsnolde, line 9: it's fine with me to put it there but it's not only about deprecated parameters, we'll also include other important warnings about clamped values and such

docs/route_overview.md
```diff
@@ -47,6 +47,6 @@ The route narrative/guidance generating code is located in the *odin* directory.
       - toJSON - Returns a trip object formated to JSON
     - **jsonToProtoRoute**(json_route, proto_route) → used by navigator
       - Transfers the JSON route information returned from a route request into the Route pbf object passed in by reference.
-  - To summarize, Valhalla builds a JSON “trip” object which contains the locations, route summary with basic info about the trip, a list of legs of the trip, status, units and language.
+  - To summarize, Valhalla builds a JSON “trip” object which contains the locations, route summary with basic info about the trip, a list of legs of the trip, status, units, language and a warnings array if any deprecated parameters are used in the request.
```
nilsnolde, line 50: same here

docs/api/turn-by-turn/overview.md
MuluhGodson: Okay I'll update the Readme documents.

test/gurka/test_warnings.cc
```diff
@@ -0,0 +1,239 @@
...
+// test case for routes endpoint
+TEST(warnings, routes_endpoint) {
+  const std::string ascii_map = R"(
+          A----------B-------------C-----P
+          |          |                   |
+          |          |                   |
+          |          3--------U     G----S
+          |          |        |     |    |
+          D----------E--------4-----F----L
+     )";
```
nilsnolde, line 17: I think we can reduce the complexity drastically here ;) it's good that you're testing all the things you added. this is what I'd change:
- have a single map for everything and an easy one. we're only testing deprecated params (so far) so the map doesn't really matter (yet). a `A--B--C` is totally fine. otherwise people (me) will be stunned in the future and frantically look for the reason for such a detailed map ;)
- look into other gurka files, you'll see how you can use gtest's `SetUpTestSuite` function that only runs once and all test fixtures can use the same map
- (bonus point, but not really necessary): parameterized test suite with just one function calling all permutations of endpoints and deprecated parameters. wasn't sure if it's possible with gtest, but seems so: https://stackoverflow.com/a/59377529

valhalla/worker.h
```diff
@@ -74,6 +75,26 @@ void ParseApi(const prime_server::http_request_t& http_request, Api& api);
...
+
+// function to add warnings to proto info object
+inline void add_warning(valhalla::Api& api, unsigned code) {
+  auto message = warnings_object.find(code);
+  if (message != warnings_object.end()) {
+    auto* warning = api.mutable_info()->mutable_warnings()->Add();
+    warning->set_description(message->second);
+    warning->set_code(message->first);
+  }
+}
```
nilsnolde, line 96: I think this should all go to the .cc file. well, except the `add_warning` definition, but the function body should move as well. the warning codes shouldn't be part of the public api, they can stay "private", just like the error codes

nilsnolde (changes requested):
a few minor things, otherwise it looks pretty good to me!

valhalla/worker.h
nilsnolde: yeah, the signature of the function should be in the header file. but the implementation (IMHO) should be in the .cc file. the reason is that with the above the warning codes are in `valhalla` namespace and "pollute" the public api. if another project uses valhalla as a library they have no use for those codes. in order for them not to be seen we put them in an anonymous namespace (and they're just dead weight in a header in my experience). IMHO it's also nicer style to keep the header files as slim/clean as reasonable to immediately see what it's containing.

it's not a big deal, bit nit picky, but there's at least actual reasons;)

kevinkreiser: yep the `declaration` should go in the header and the `definition` or `implementation` should go in the source file

test/gurka/test_warnings.cc
MuluhGodson: :smile: thank you for this. having one map for all the test cases really simplified it. The file even looks way cleaner.

test/gurka/test_warnings.cc
```diff
@@ -0,0 +1,77 @@
...
+ * Endpoints defined in proto/options.proto
+ * route = 1
+ * locate = 2
+ * trace_route = 6
+ * trace_attributes = 7
+ * transit_available = 9
+ * height = 8
+ */
+
+INSTANTIATE_TEST_CASE_P(WarningsTests, Warning, ::testing::Values(1, 2, 6, 7, 8, 9));
```
nilsnolde, line 77: sorry, maybe I was being overzealous (and didn't think enough about the consequences..). actually I meant to parameterize a map of endpoint: msg or so, but I think that's hard to realize and not worth the effort. you mind reverting back to what you had before?

test/gurka/test_warnings.cc
```diff
@@ -0,0 +1,77 @@
...
+TEST_P(Warning, matrix) {
+  Api request;
+  for (auto& costing : deprecated_costing_methods) {
+    std::string request_str = R"({"sources":[{"lat":)" + std::to_string(map.nodes["A"].lat()) +
+                              R"(,"lon":)" + std::to_string(map.nodes["A"].lng()) +
+                              R"(}],"targets":[{"lat":)" + std::to_string(map.nodes["C"].lat()) +
+                              R"(,"lon":)" + std::to_string(map.nodes["C"].lng()) +
+                              R"(}],"costing":")" + costing + R"("})";
+    ParseApi(request_str, Options::sources_to_targets, request);
+    ASSERT_FALSE(request.info().warnings_size() == 0);
```
nilsnolde, line 62: this is superfluous. if the size is 1 it's obviously not 0 ;)

src/tyr/transit_available_serializer.cc
```diff
@@ -15,20 +15,27 @@ json::MapPtr serialize(const PathLocation& location, bool istransit) {
...
                                       const std::vector<baldr::Location>& locations,
                                       const std::unordered_set<baldr::Location>& found) {
   auto json = json::array({});
   for (const auto& location : locations) {
     json->emplace_back(serialize(location, found.find(location) != found.cend()));
   }
+
+  // add warnings to json response
+  if (request.info().warnings_size() >= 1) {
+    json->emplace_back(json::map({{"warnings", valhalla::tyr::serializeWarnings(request)}}));
```
nilsnolde, line 36: we should remove `warnings` from this endpoint. you can see it returns an array of defined objects, we can't just throw a `warnings` object in the mix, that'd break the output format.

nilsnolde (changes requested):
sorry for the wait. a few more minor things

test/gurka/test_warnings.cc
MuluhGodson: Okay, I've reverted the test to the version without the parameterization.

test/gurka/test_warnings.cc
MuluhGodson: :smile: Oh I've taken out all the superfluous assertions.

src/worker.cc
```diff
@@ -138,6 +138,25 @@ const std::unordered_map<unsigned, valhalla::valhalla_exception_t> error_codes{
     {503, {503, "Leg count mismatch", 400, HTTP_400, OSRM_INVALID_URL, "wrong_number_of_legs"}},
 };
 
+// unordered map for warning pairs
+const std::unordered_map<int, std::string> warnings_object = {
+    {100, "auto_shorter is deprecated and will be turned into the shotest costing option"},
```
SiarheiFedartsou, line 143:
```suggestion
    {100, "auto_shorter is deprecated and will be turned into the shortest costing option"},
```

src/tyr/height_serializer.cc
```diff
@@ -94,6 +94,11 @@ std::string serializeHeight(const Api& request,
     json->emplace("id", request.options().id());
   }
 
+  // add warnings to json response
+  if (request.info().warnings_size() >= 1) {
+    json->emplace("warnings", serializeWarnings(request));
+  }
```
nilsnolde, line 100: `"warnings"` is only returned if there was at least one warning, so it's conditionally present in JSON output. I'd personally prefer to output an empty array, I think it's nicer for end users in most scenarios. Other opinions?

test/gurka/gurka.cc
```diff
@@ -64,9 +64,9 @@ std::vector<midgard::PointLL> to_lls(const nodelayout& nodes,
  */
 std::string build_valhalla_request(const std::string& location_type,
                                    const std::vector<midgard::PointLL>& waypoints,
-                                   const std::string& costing = "auto",
-                                   const std::unordered_map<std::string, std::string>& options = {},
-                                   const std::string& stop_type = "break") {
+                                   const std::string& costing,
+                                   const std::unordered_map<std::string, std::string>& options,
+                                   const std::string& stop_type) {
```
nilsnolde, line 69: I made this function public so it can be used in tests. Hope that's fine with you..

src/tyr/height_serializer.cc
MuluhGodson: Yeah thanks for bringing this up. I've done some more work with APIs and this is definitely a good way to go as some users may write their logic keeping in mind there's a warmings output and if not properly configures for null safety could be problematic. 
I'll update the PR. I certainly have a lot to catch up on.  😁

nilsnolde:
From my side this is pretty much ready for merging. Just cleaned up a bit, @MuluhGodson did the vast majority of it, thanks for that!

Windows is failing.. I don't understand the error. It seems like it's trying to install `curl` from the system's included `C:\vcpkg`, but that shouldn't even be visible as we use our custom built `vcpkg` for all commands in the `.azure-pipelines.yml`. Annoying (and seemingly pointless to have `vcpkg` pre-installed, at what commit it pointing? current master?).. and not sure how to fix that.. Do you have any idea @mloskot ?
``` 
Error: There are invalid fields in the control or manifest file of C:\vcpkg\ports\curl\vcpkg.json
The following fields were not expected:
    In a dependency: host, host
    In manifest: version
``` 
https://dev.azure.com/valhalla1/valhalla/_build/results?buildId=10781&view=logs&j=8c5a3053-cc94-5e24-da52-c43aabf65e0e&t=2eecd64f-9377-5e39-fbc9-8cd385513ff3

src/tyr/height_serializer.cc
nilsnolde: no worries, I just carried it over the finish line, wasn't much to left to do:)

MuluhGodson: Great, thanks.

mloskot:
@nilsnolde I'm not entirely clear, but long story short, versioning specification changed and curl manifest uses later one

From https://vcpkg.io/en/docs/examples/versioning.getting-started.html

> Previously, you could only declare versions for your projects using the `version-string` property. 

Valhalla seems to be using the `vcpkg.exe` old enough to expect this `version-string` only, but `ports/curl/vcpkg.json` uses newer scheme for version specification. It also looks like curl has become an indirect dependency (something started requiring curl? mystery...) but old `vcpkg.exe` can not parse its (new format) manifest.

Looks like vcpkg update may be required here.

nilsnolde:
Urgh, thanks @mloskot ! I'd just like it to use what I specify: our vcpkg commit/version and the manifests it uses. Not the manifests of whatever is installed system-wide already. Are you aware of any flag to `vcpkg` that makes ignore other installations? Is it maybe setting a `VCPKG_ROOT` env var? 

I wouldn't like to go through the ordeal of setting up my win machine right now :sweat_smile: I'll try a more recent version of `vcpkg`, if that doesn't work out-of-the-box, it'll have to wait for a bit. Damn, all 3 platforms have failing jobs right now..

nilsnolde:
>  It also looks like curl has become an indirect dependency (something started requiring curl?

@mloskot I think that's because `DATA_TOOLS` is turned on since a while on CI and that requires curl:
https://github.com/valhalla/valhalla/blob/a6d46d93043bc94a0460a026404909bb76772578/CMakeLists.txt#L163

mloskot:
@nilsnolde  Yes, the data tools need curl. I'm not clear what causes the vcpkg confusion (does image on Azure Pipelines already come with preinstalled vcpkg, etc.), but I'm a bit swamped. However, if you experience problems w/ upgrading vcpkg, I will do my best to chime in, so feel free to ping me please.

nilsnolde:
> (does image on Azure Pipelines already come with preinstalled vcpkg, etc.)

That's what it looks like to me. But we're using our version of vcpkg explicitly, also in `CMAKE_TOOLCHAIN_FILE`. Anyways, I opened #3754 with an attempt to naively upgrade, it failed and I'll try to replicate locally soon-ish. And thanks, I'll scream loud enough for sure when my low MS frustration bar is exceeded;)

nilsnolde (dismissed):

src/tyr/serializers.cc
```diff
@@ -157,6 +157,26 @@ void openlr(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper
   writer.end_array();
 }
 
+void serializeWarnings(const valhalla::Api& api, rapidjson::writer_wrapper_t& writer) {
+  writer.start_array("warnings");
+  for (int i = 0; i < api.info().warnings_size(); i++) {
+    writer.start_object();
+    writer("code", api.info().warnings(i).code());
+    writer("text", api.info().warnings(i).description());
```
kevinkreiser, line 165:
```suggestion
  for (const auto& warning : api.info().warnings()) {
    writer.start_object();
    writer("code", warning.code());
    writer("text", warning.description());
```

src/tyr/serializers.cc
```diff
@@ -157,6 +157,26 @@ void openlr(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper
...
+  }
+  writer.end_array();
+}
+
+json::ArrayPtr serializeWarnings(const valhalla::Api& api) {
+  auto warnings = json::array({});
+  for (int i = 0; i < api.info().warnings_size(); i++) {
+    warnings->emplace_back(json::map(
+        {{"code", api.info().warnings(i).code()}, {"text", api.info().warnings(i).description()}}));
+  }
```
kevinkreiser, line 176:
```suggestion
  for (const auto& warning : api.info().warnings()) {
    warnings->emplace_back(json::map(
        {{"code", warning.code()}, {"text", warning.description()}}));
  }
```

kevinkreiser (dismissed):

nilsnolde (dismissed):
only missing changelog conflict update after the other PRs are merged into master

nilsnolde (approved):
Resolved conflicts with master, should be good to merge now

kevinkreiser (dismissed):

nilsnolde:
Think I’ll try to revert the changes to gurka here before merging

nilsnolde (dismissed):
Jeez, last time.. reverted the changes to gurka.cc and did a literal instead in the request.

nilsnolde (approved):

PR #3590, opened 2022-04-07
Added highway and surface factor in Truck Cost

### LINKED ISSUE #3325: truck cost missing some factors in edgecost
just realized that truck routing doesn't care about the `use_highways` option. I didn't do an extensive review what else might be missing, on a quick scan it's at least:
- highway factor
- surface factor

labeling as bug since it's not expected behavior reading the documentation

mr-loop-1:
Added mssing cost calculations for Highway and Surface factors in **truckcost.cc** (`/src/sif/truckcost.cc`) which should have been present according to the [docs](https://valhalla.readthedocs.io/en/latest/api/turn-by-turn/api-reference/#automobile-and-bus-costing-options)

# Issue

Fixes #3325 

## Testing

Added a `test_truck_highway.cc` file in `/test/gurka`. The tests pass after proposed fixes.


src/sif/truckcost.cc
```diff
@@ -64,6 +65,28 @@ constexpr float kLeftSideTurnCosts[] = {kTCStraight,         kTCSlight,  kTCUnfa
 // How much to favor truck routes.
 constexpr float kTruckRouteFactor = 0.85f;
 
+constexpr float kMaxHighwayBiasFactor = 8.0f;
```
nilsnolde, line 68: while pulling the entire logic (more or less) into the base class (which I'd prefer) is quite a bit bigger refactor, we can at least declare `kMaxHighwayBiasFactor` in `dynamiccost.cc` so we don't define it 3 times over. actually the arrays are all the same as well, but we can tackle that at a later stage

src/sif/truckcost.cc
```diff
@@ -310,6 +336,23 @@ TruckCost::TruckCost(const Costing& costing)
     speedfactor_[s] = (kSecPerHour * 0.001f) / static_cast<float>(s);
   }
 
+  // Sets the surface factor for truck
+  surface_factor_ = 0.5f;
```
nilsnolde, line 340: this doesn't make much sense here. also shouldn't truck be a little more sensitive to surfaces? IMO this can easily be `1.0f` in which case you can omit the whole `surface_factor_` and only use the array for costing further down

test/gurka/test_truck_highway.cc
```diff
@@ -0,0 +1,73 @@
...
+gurka::map TruckHighwayTest::map = {};
+
+TEST_F(TruckHighwayTest, Truckcost) {
+  doTests("truck", {{"/costing_options/truck/use_highways", "0"}});
+}
+
+
+// TEST_F(TruckHighwayTest, MotorcycleCost) {
+//   doTests("motorcycle", {{"/costing_options/motorcycle/use_highways", "0"}});
+// }
```
nilsnolde, line 72: I'd rename the file to `test_use_highways.cc` and also include motorcycle and car here.

nilsnolde:
thanks @mr-loop-1 ! good start! few minor comments.

also, you need to run format.sh so CI passes. I prefer it as pre-commit hook but that's up to you:
```
cat ./scripts/format.sh > .git/hooks/pre-commit && tail -n +5 scripts/error_on_dirty.sh >> .git/hooks/pre-commit
chmod +x .git/hooks/pre-commit
```

src/sif/truckcost.cc
mr-loop-1: I did that because the surface factor was `1` only for Motorcycle in `Autocost.cc`.

https://github.com/valhalla/valhalla/blob/6bdc1e64eaa2196697b5d07b7735cc0dc465adaa/src/sif/autocost.cc#L355-L371

mr-loop-1: I did that because the surface factor was `1` only for Motorcycle in `Autocost.cc`.

https://github.com/valhalla/valhalla/blob/6bdc1e64eaa2196697b5d07b7735cc0dc465adaa/src/sif/autocost.cc#L355-L371

nilsnolde: you just uncovered another undocumented parameter:) `type` for car, where `motorcycle` seems odd and probably from days before there was a proper costing model for it. 

anyways, I'm also not sure if 0.5 or 1.0 is better here, though leaning towards 1.0. my point is we don't need that variable because a truck is a truck (for us so far at least), so you can just take that factor directly into the `EdgeCost` and remove the definition of `surface_factor_`

src/sif/truckcost.cc
mr-loop-1: Maybe there would have been possibility for array values becoming different for modes, that they provided them separately in each file. As for Bias Factor I moved it to the header.

nilsnolde (approved):
great thanks @mr-loop-1, for me this looks pretty much done. let's see what others say before merging

nilsnolde:
we could merge this as well actually. wdyt @kevinkreiser ? just came across this again when one of our apps wouldn't avoid highways on the truck costing. maybe needs a bit revised factors for surface penalties, IIRC it just copied it from the car model.

kevinkreiser (approved):

nilsnolde (approved):

PR #3593, opened 2022-04-13
Add defaults for a few parameters

nilsnolde:
some more explicit defaults for a few parameters.


kevinkreiser (approved):
very much appreciated

Issue #3482, opened 2021-12-29
shortcuts should be capped in length


nilsnolde:
recently ran a US tomtom dataset where we were handling ferries pretty naively and valhalla attempted to build a 24 000 km shortcut :joy: 

```
2021/12/20 21:33:30.842885 [INFO] Done HierarchyBuilder
2021/12/20 21:33:30.846544 [INFO] Creating shortcuts on level 1
2021/12/20 21:35:21.394332 [ERROR] Exceeding max. edge length: 24087463
terminate called after throwing an instance of 'std::runtime_error'
  what():  DirectedEdgeBuilder: exceeded maximum edge length
```

@kevinkreiser noted it'd be best to cap the maximum length of shortcuts. I'll take care of it soon. guess we just cap it at the max allowed value right? 
https://github.com/valhalla/valhalla/blob/ba50b993cc02be4e610f0231a71a40b3153b936c/valhalla/baldr/graphconstants.h#L175

nilsnolde:
urgh, I just realized this is now also failing OSM.. somewhere some edge/shortcut, which was added in the last 2 weeks, is throwing the same error. interestingly remarkably close edge length with 26888 km. now this becomes more of a high priority issue @chris-jpark :)

PR #3597, opened 2022-04-19
fixed issue #3482 capped shortcut max length

chris-jpark:
# Issue
## Issue #3482  

Patched an issue where shortcuts were being created with edges greater than the max length. 

Valhalla will now throw an error when an edge greater than kMaxEdgeLength.

File in `/src/mjolnir/shortcutbuilder.cc/` was edited for implementation. 

Tested with local gurka environment, ensures that the error is thrown

File in `/test/gurka/test_shortcut.cc/` was edited for testing.


nilsnolde:
I think you committed the full build tree?! Can’t see much but there’s the CMakeCache.txt in the root of the repo. It’s best to build into a „build“ folder, that’s in the .gitignore and can’t be accidentally committed. Easiest if you now just nuke the whole project and start from scratch.

nilsnolde:
Also you need to run scripts/format.sh

dividividib:
Thank you for the reply, I'll just grab the parts that I edited and make a separate clone.

CMakeLists.txt
```diff
@@ -47,6 +47,8 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 
+set(CMAKE_CXX_FLAGS "-DGEOS_INLINE")
```
nilsnolde, line 50: you can pass that via command line to cmake

format.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
nilsnolde, line 1: this file is wrong here, just delete it

src/mjolnir/Makefile
```diff
@@ -0,0 +1,1117 @@
+# CMAKE generated file: DO NOT EDIT!
```
nilsnolde, line 1: no cmake generated makefiles, guess a relict from your previous in-source build?

src/mjolnir/directededgebuilder.cc
```diff
@@ -36,6 +36,9 @@ DirectedEdgeBuilder::DirectedEdgeBuilder(const OSMWay& way,
   // Protect against 0 length edges
   set_length(std::max(length, kMinimumEdgeLength), true);
 
+  // Protect against values greater than maximum edge length;
+  // set_length(std::min(length, kMaxEdgeLength), true);
```
nilsnolde, line 40: left from debugging?

src/mjolnir/shortcutbuilder.cc
```diff
@@ -406,6 +406,12 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       DirectedEdge newedge = *directededge;
       uint32_t length = newedge.length();
 
+      //    if shortcut edge length is greater than kMaxEdgeLength, return 0
+      if (length > kMaxEdgeLength) {
+        LOG_ERROR("length too big");
+        throw std::runtime_error("edge length too big");
+      }
```
nilsnolde, line 413: hm, I had smth else in mind tbh. this just raises the same error as before, just with a slightly different error message, so to me it doesn't really add value yet. I think there's a fairly straight-forward way to handle this gracefully and still finish a valid graph build. what I thought we should do is detect when the max edge length is about to be exceeded, and if it would, we finish the current shortcut with a length less than the max and then create the next shortcut starting with the next edge. does that make sense?

test/gurka/test_shortcut.cc
```diff
@@ -74,6 +74,38 @@ TEST(Shortcuts, CreateInvalid) {
...
+        {"maxspeed:backward", "80"}}},
+  };
+
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize);
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_openlrjoiner_shortcut_speed");
+
+  baldr::GraphReader graph_reader(map.config.get_child("mjolnir"));
+
+  // check that there are no shortcut edges
+  EXPECT_ANY_THROW(gurka::findEdgeByNodes(graph_reader, layout, "A", "C"));
```
nilsnolde, line 105: we'll also have to update the test after changing the behavior, basically we don't want it to fail.

src/mjolnir/graphvalidator.cc
```diff
@@ -134,6 +134,10 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode,
         continue;
       }
 
+      if (edge.length() > kMaxEdgeLength) {
+        throw std::runtime_error("Edge Length too Long");
+      }
```
nilsnolde, line 139: this is probably already taken of: https://github.com/gis-ops/valhalla/blob/6dac5913835663881178537d53e3385cfdf6d695/src/baldr/directededge.cc#L117-L123

nilsnolde (changes requested):

src/mjolnir/shortcutbuilder.cc
dividividib: For a given shortcut, I understand that we want the total length of the shortcut to be shorter than the kMaxEdgeLength. When a shortcut is created, does the entire shortcut become one edge, or does it become a structure of connected edges?

kevinkreiser: the entire shortcut will be a single edge. what we want to do, as @nilsnolde mentioned was to modify the `CanContract` method to return false if the contraction would make the final edge length longer than `kMaxEdgeLength`

CMakeLists.txt
dividividib: Using the cmd line was giving me errors in my build - probably because of apple silicon's bugs, I will edit it out in the push build.

src/mjolnir/shortcutbuilder.cc
dividividib: I made sure the CaContract returns false if the node's length is greater than the maxLength, but does the scope of the CanContract function allow us to keep track of the total shortcut length? It looks like it's being called before the shortcuts are built and does not know the length of the shortcut. I thought the length of the shortcuts could only be determined as FormShortcuts is called.

src/mjolnir/Makefile
dividividib: this shouldn't have happened, as I deleted all files inside the previous build and re-cloned.

src/mjolnir/shortcutbuilder.cc
kevinkreiser: I was speaking more to the semantics rather than how the code is today. Surely the call could have the total length passed in so far and bail if exceeded so that all the decision code is in one function. Or up one level we could track this decision separately. Either way will work I wasn't being specific

dividividib: Alright! Thank you for the clarification, that makes sense. I was wondering to what extent I could alter the code - looks like I have quite some freedom to work with!

dividividib: I've reflected on your advice, and added a condition to check for length in the AddShortcutEdges function. This seems to work, as tested on the test_shortcut file. I tested so that a shortcut is created between paths less than the max, but is not for edges greater than the cap.

src/mjolnir/shortcutbuilder.cc
```diff
@@ -218,6 +218,12 @@ bool CanContract(GraphReader& reader,
   if (!EdgesMatch(tile, edge1, edge2))
     return false;
 
+  //if edges are greater than the maximum legnth constnat, do not contract them
+  uint32_t totalLength = edge1->length() + edge2->length(); 
+  if(totalLength > kMaxEdgeLength) {
+    return false; 
+  }
+
```
nilsnolde, line 226: is this one necessary? is there a case in the main function where it might become a problem if we don't add this check? otherwise I'd just throw it

src/mjolnir/shortcutbuilder.cc
```diff
@@ -472,6 +483,12 @@ uint32_t AddShortcutEdges(GraphReader& reader,
           break;
         }
 
+        //terminate shortcut if edge length has exceeded
+        shortcutLength +=tile->directededge(next_edge_id)->length(); 
+        if(shortcutLength > kMaxEdgeLength){
+          break; 
+        }
```
nilsnolde, line 490: you can remove the return from ConnectEdges() and do this instead:

```suggestion
        length += tile->directededge(next_edge_id)->length(); 
        if(length > kMaxEdgeLength){
          break; 
        }
        
        // further down
        ConnectEdges(..)
```

test/gurka/test_shortcut.cc
```diff
@@ -74,6 +74,42 @@ TEST(Shortcuts, CreateInvalid) {
...
+        {"maxspeed:backward", "80"}}},
+      {"BC",
+       {{"highway", "primary"},
+        {"name", "Independence Avenue"},
+        {"maxspeed:forward", "80"},
+        {"maxspeed:backward", "80"}}},
+  };
+
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize);
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_openlrjoiner_shortcut_speed");
```
nilsnolde, line 100: `gurka_openlrjoiner_shortcut_speed`?

test/gurka/test_shortcut.cc
```diff
@@ -74,6 +74,42 @@ TEST(Shortcuts, CreateInvalid) {
...
+
+  baldr::GraphReader graph_reader(map.config.get_child("mjolnir"));
+
+  // check that there are no shortcut edges
+  auto shortcut_edge = std::get<1>(gurka::findEdgeByNodes(graph_reader, layout, "A", "C"));
```
nilsnolde, line 105: does that mean you tuned the test map in a way that you'll get a shortcut from A to C right? that'd be wonderful. but shouldn't there then also be a shortcut from C -> E? if not, is it possible to tune the map even further, so that the whole thing is too long, but we do get 2 shortcuts out of it in total?

ah wait.. you didn't specify the other ways in the map.. you only did `AB` and `BC`, the others don't even exist as ways as far as the map is concerned.

nilsnolde (changes requested):
thanks for the next round. still some things to get cleaned up. in the end we only want a change in shortcutbuilder.cc and test_shortcut.cc without any pure whitespace changes.

src/mjolnir/shortcutbuilder.cc
```diff
@@ -406,6 +405,7 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       DirectedEdge newedge = *directededge;
       uint32_t length = newedge.length();
 
+
```
nilsnolde, line 408: irrelevant white space change

nilsnolde, line 408: irrelevant white space change

src/mjolnir/shortcutbuilder.cc
```diff
@@ -449,6 +449,10 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       // For turn duration calculation during contraction
       uint32_t opp_local_idx = directededge->opp_local_idx();
       GraphId next_edge_id = edge_id;
+
+      //keep track of shortcut length so it does not exceed max length
+      uint32_t shortcutLength = 0; 
+
```
nilsnolde, line 455: redundant

test/gurka/test_shortcut.cc
```diff
@@ -74,6 +74,52 @@ TEST(Shortcuts, CreateInvalid) {
...
+       {{"highway", "primary"},
+        {"name", "Independence Avenue"},
+        {"maxspeed:forward", "60"},
+        {"maxspeed:backward", "60"}}},
+      {"DE",
+       {{"highway", "primary"},
+        {"name", "Independence Avenue"},
+        {"maxspeed:forward", "60"},
+        {"maxspeed:backward", "60"}}},
+  };
```
nilsnolde, line 107: still, I don't think this test is proving that the changes work now. you change properties going from `BC` to `CD`, so no matter the edge length it's not possible to have a shortcut from A -> D/E. to prove a point, all edges must have the same properties (you can only keep name & highway tag). 

and _ideally_ the map produces two shortcuts from that stretch (might be you have to extend it a little). the first one aborts due to length issues and starts a second one.

nilsnolde (commented):
I think the test is not quite right yet. and a few redundant changes.

src/mjolnir/shortcutbuilder.cc
```diff
@@ -622,7 +628,7 @@ uint32_t FormShortcuts(GraphReader& reader, const TileLevel& level) {
       const auto& admin = tile->admininfo(nodeinfo.admin_index());
       nodeinfo.set_edge_index(tilebuilder.directededges().size());
       nodeinfo.set_admin_index(tilebuilder.AddAdmin(admin.country_text(), admin.state_text(),
-                                                    admin.country_iso(), admin.state_iso()));
+                                                    admin.country_iso(), admin.state_iso()));
```
kevinkreiser, line 631: nit: whitespace

chris-jpark:
Found a bigger issue than expected - basically, even if I cut the shortcut short when the max length is reached, the code doesn't know to continue building a second shortcut. 

It creates the shortcuts by looping through the nodes, and it determines whether to create a shortcut based on `CanContract`, so we would like that to return false if a 'maxLength' condition is met. 

I think we could create a hashmap in `FormShortcuts` and feed it into the argument of `CanContract` and `AddShortcutEdges` so `AddShortcutEdges` can append to it when we find a shortcut too long, and `CanContract` can read that and return false. The hashmap should contain the edges that raised the condition, since that is what `AddShortcutEdges` loops through, and `CanContract` already reads the edges of each node anyways.

nilsnolde:
I think some state to remember which ones failed is probably the best way. you probably want to put nodes to that map though since that's what `CanContract` checks (is it contractable) and return `false` if that node is in the new map, even though it would be contractable but we have to pretend it's not. 

one problem this approach has it that at the point we're aborting a shortcut we're actually looping through one specific tile's nodes while the node where the abortion happens is 99.99% in another tile, which might come after or before the current tile in the loop. in other words we actually would have to process all tiles first, see where we have too long shortcuts, then go back to pick up those nodes and create new shortcuts starting at those. doable but quite a bit more work to refactor. considering that, I'm wondering if we shouldn't just skip the further shortcut creation and keep just the plain edges after we hit the maximum. the issue this pr results from had problems with ferry edges. those are a) pretty rare (and not present in OSM) and b) unlikely to have a measurable impact on performance. it's not 100% correct but it's also not really wrong, it's just practical.

what do you think @kevinkreiser ?

kevinkreiser:
Yeah i think this was my intuition at the beginning when i was suggesting to control the behavior via `CanContract`. Give bit a bit of time to refresh my knowledge of the code and I'll see what I can suggest, sorry for the delay!

chris-jpark:
Ok, I'll give that a shot and update on how it goes. : )

chris-jpark:
When I added the hashmap, the test case throws an error for accessing a directed edge that is 'out of bounds'. I suspect that this is because it expects a shortcut from A-F based on the 'endnode' we create initially. I'll look and see to fix that over the next couple days.

nilsnolde:
yeah I think we'd need to track those nodes where we aborted a shortcut (as you already do) and have another pass over only those nodes and create those additional shortcuts. it's a little more of a refactor, since you'll want to re-use all the logic in `FormShortcuts`, but it's mostly breaking the node logic out into a separate function it seems. this is just glancing at the code, but I think that could work?

chris-jpark:
Although refactoring code further would improve pathfinding efficiency, the problem of shortcuts being extremely long shouldn't occur often. Sending a pull request where code terminates shortcuts if maxLength is reached. If the tiles that include the nodes that follow are processed afterwards, the code will create additional shortcuts that continue the original shortcut path.

src/mjolnir/shortcutbuilder.cc
```diff
@@ -366,13 +365,18 @@ uint32_t AddShortcutEdges(GraphReader& reader,
...
+  // this node can be contracted and is not a terminus of a pervious shortcut.
   EdgePairs edgepairs;
-  if (CanContract(reader, tile, start_node, edgepairs)) {
+  std::unordered_set<GraphId>::const_iterator prev_shortcut = last_nodes.find(start_node);
+
+  if (CanContract(reader, tile, start_node, edgepairs) && prev_shortcut == last_nodes.end()) {
     return 0;
   }
+  // Variable to check if this is a sliced shortcut w shortcut was too long
+  bool sliceShortcut = false;
```
nilsnolde, line 379: this can go, it's unused

src/mjolnir/shortcutbuilder.cc
```diff
@@ -366,13 +365,18 @@ uint32_t AddShortcutEdges(GraphReader& reader,
                           const GraphId& start_node,
                           const uint32_t edge_index,
                           const uint32_t edge_count,
-                          std::unordered_map<uint32_t, uint32_t>& shortcuts) {
+                          std::unordered_map<uint32_t, uint32_t>& shortcuts,
+                          std::unordered_set<GraphId>& last_nodes) {
   // Shortcut edges have to start at a node that is not contracted - return if
-  // this node can be contracted.
+  // this node can be contracted and is not a terminus of a pervious shortcut.
   EdgePairs edgepairs;
```
nilsnolde, line 372: I'd leave a TODO here to tell ourselves that we _could_ improve on this by looping over all `last_nodes` after processing all tiles, else it's random whether a broken shortcut will ever be picked up in another tile

test/gurka/test_shortcut.cc
```diff
@@ -74,6 +74,57 @@ TEST(Shortcuts, CreateInvalid) {
   EXPECT_ANY_THROW(gurka::findEdgeByNodes(graph_reader, layout, "C", "A"));
 }
 
+// Here no shortcuts are created.
+// The grid size should be greater than the max length that allows shortcuts to be created
+TEST(Shortcuts, CreateTooLong) {
+  constexpr double gridsize = 1500000;
+  // Maximum edge length is    16777215
```
nilsnolde, line 81: can you indent this number better to align with the number over it? so it's easier to see that it's a full magnitude difference

nilsnolde (commented):
few nits, other than that we're done here I think:) thanks @chris-jpark . now on to gtfs :D

src/mjolnir/shortcutbuilder.cc
```diff
@@ -366,11 +365,18 @@ uint32_t AddShortcutEdges(GraphReader& reader,
...
+  // this node can be contracted and is not a terminus of a pervious shortcut.
+
+  // TODO: Could loop through last_nodes after proessing all tiles
+  // to continue the shortcut path of the ones terminated because they were
+  // too long. Currently, a shortcut path is continued by another tile at random.
   EdgePairs edgepairs;
-  if (CanContract(reader, tile, start_node, edgepairs)) {
+  std::unordered_set<GraphId>::const_iterator prev_shortcut = last_nodes.find(start_node);
+
+  if (CanContract(reader, tile, start_node, edgepairs) && prev_shortcut == last_nodes.end()) {
```
kevinkreiser, line 379: is the idea with this change that we can try to continue making shortcuts for nodes that we aborted previous shortcuts at? basically we only bail on a node if its contractable but we havent ended/aborted a too-long shortcut at it? thus allowing the creation of another shortcut from that abort point? i just worry about the rest of the logic making sure not to duplicate the previously aborted shortcut and to continue in the other direction. are we sure this wouldnt cause any trouble?

dnesbitt61: also, when creating a shortcut in one direction the code needs to make sure the "reverse" shortcut is created in the opposing direction. I didn't look at this closely, but making sure that rule is obeyed is essential.

kevinkreiser: yeah exactly. basically changing anything about this is potentially brittle if we dont have a good understanding of the assumptions downstream. i have many worries like this in the back of my h ead as well. its not that i dont think we should try to tackle this but we need to mitigate any regressions this would cause. daves example is a good one. how do we make sure the opposing shortcut of the aborted one lines up where it stopped?

nilsnolde:
I think it’s safe to say that we went down a rabbit hole here. Sorry for the time waste @chris-jpark ! And good work on the actual project:)

PR #3607, opened 2022-04-30
Enhance valhalla_build_elevation with LZ4 recompression support

### LINKED ISSUE #3606: Add support for LZ4 recompression in valhalla_build_elevation
Building on the work of #3401, the elevation download script should be enhanced to support optional re-compression of elevation tiles from gzip into lz4.

ianthetechie:
# Issue

Closes #3606.

## Tasklist

 - [ ] Add tests - **I don't see applicable tests for similar download portions of the script; let me know if you want a test**
 - [ ] If you made changes to the lua files, update the [taginfo](taginfo.json) too. - **N/A**

## Requirements / Relations

Complementary to #3401.

# Commentary

The main goal of this PR is to add support for re-compressing downloaded elevation tiles as LZ4. LZ4 comes at a slight space cost (roughly 12% larger than gzip for the planet) and of course the upfront CPU cost of recompressing. In real-world request loads, we have observed significant reductions in p95 and higher response times after switching from gzip to LZ4 elevation tiles.

Secondarily, this PR makes a subtle type change from a boolean decompress flag to explicitly supporting multiple compression types in case something better comes along later. This also ensures the added cases are cleanly handled. It also removes the requirement for a JSON config file. While a JSON config file is probably the more typical usage pattern, it's also useful to be able to specify a bbox and output directory on the command line directly when downloading the whole world to a local file server or similar applications (and it technically doesn't even have to be Valhalla-specific).

### DIFF
--- a/scripts/valhalla_build_elevation
+++ b/scripts/valhalla_build_elevation
@@ -71,7 +71,7 @@ method.add_argument(
 method.add_argument(
     "-t",
     "--from-tiles",
-    help="Only download tiles covered by the Valhalla graph.",
+    help="Only download tiles covered by the Valhalla graph (requires a Valhalla config JSON file).",
     action="store_true"
 )
 parser.add_argument(
@@ -307,7 +307,11 @@ if __name__ == "__main__":
     elif args.from_bbox:
         tiles = get_tiles_with_bbox(args.from_bbox)
     elif args.from_tiles:
-        tiles = get_tiles_with_graph(Path(config["mjolnir"]["tile_dir"]))
+        if config is not None:
+            tiles = get_tiles_with_graph(Path(config["mjolnir"]["tile_dir"]))
+        else:
+            LOGGER.critical("--from-tiles requires a config to be specified.")
+            sys.exit(1)
     else:
         LOGGER.critical("No download method specified.")
         sys.exit(1)
--- a/scripts/valhalla_build_elevation
+++ b/scripts/valhalla_build_elevation
@@ -307,11 +307,10 @@ if __name__ == "__main__":
     elif args.from_bbox:
         tiles = get_tiles_with_bbox(args.from_bbox)
     elif args.from_tiles:
-        if config is not None:
-            tiles = get_tiles_with_graph(Path(config["mjolnir"]["tile_dir"]))
-        else:
+        if config is None:
             LOGGER.critical("--from-tiles requires a config to be specified.")
             sys.exit(1)
+        tiles = get_tiles_with_graph(Path(config["mjolnir"]["tile_dir"]))
     else:
         LOGGER.critical("No download method specified.")
         sys.exit(1)
--- a/scripts/valhalla_build_elevation
+++ b/scripts/valhalla_build_elevation
@@ -2,27 +2,29 @@
 
 import argparse
 import os
-from collections import namedtuple
-from functools import partial
 import gzip
 import json
 import logging
+import sys
+
+from collections import namedtuple
+from enum import Enum, unique
+from functools import partial
 from math import ceil, floor
 from multiprocessing import cpu_count
 from multiprocessing.dummy import Pool as ThreadPool
 from pathlib import Path
-import sys
 from typing import List, Iterable, Set
 from urllib import request
 from urllib.error import URLError
-
 
 # hack so ArgumentParser can accept negative numbers
 # see https://github.com/valhalla/valhalla/issues/3426
 for i, arg in enumerate(sys.argv):
     if not len(arg) > 1:
         continue
-    if (arg[0] == '-') and arg[1].isdigit(): sys.argv[i] = ' ' + arg
+    if (arg[0] == '-') and arg[1].isdigit():
+        sys.argv[i] = ' ' + arg
 
 description = """Downloads Tilezen's elevation tiles that either intersect with features in all .geojson files in
               --input-geojson-dir or that intersect with --bbox. NOTE: geojson method requires shapely.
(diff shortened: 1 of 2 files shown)


scripts/valhalla_build_elevation
```diff
@@ -2,27 +2,29 @@
...
-from collections import namedtuple
-from functools import partial
 import gzip
 import json
 import logging
+import sys
+
+from collections import namedtuple
+from enum import Enum, unique
+from functools import partial
```
nilsnolde, line 12: messing with our carefully ordered imports huh;) (though just seeing that `os` escaped) can we revert back to what it was before and just add enum alphabetically?

scripts/valhalla_build_elevation
```diff
@@ -34,13 +36,31 @@ handler = logging.StreamHandler()
...
+    LZ4 = 3
+
+    @property
+    def extension(self):
+        if self is TileCompression.UNCOMPRESSED:
+            return ""
+        elif self is TileCompression.GZIP:
+            return ".gz"
+        elif self is TileCompression.LZ4:
+            return ".lz4"
```
nilsnolde, line 54: this is really nice!

scripts/valhalla_build_elevation
```diff
@@ -208,24 +246,45 @@ def download(tile: Tile, output_dir, decompress):
...
+                        f.write(uncompressed)
+                    elif compression is TileCompression.LZ4:
+                        # Local imports are yuck, even though it only takes effect once. the alternative is
+                        # requiring everyone to have the LZ4 package available, regardless of whether the user needs it.
+                        # Compression level 6 was chosen after some benchmarking as having a reasonable balance between
+                        # compression time and space savings. The end result will be larger than the maximally gzipped
+                        # tiles from AWS, but only by around 12%. To narrow the gap to ~10% requires compression level
+                        # 9, and more than double the compression time, so this is the approximate efficient frontier.
+                        # for our purposes. Note that in general LZ4 decompression cost does not vary much with
+                        # compression level, unless using extremely high compression levels.
```
nilsnolde, line 264: local import is perfect here IMO, we do the same already with shapely in case of geojson downloads. I'd cut the comment a little maybe :sweat_smile:

scripts/valhalla_build_elevation
```diff
@@ -235,6 +294,19 @@ if __name__ == "__main__":
...
 
+    config = None
+    if args.config:
+        with open(args.config) as f:
+            config = json.load(f)
+
+    if args.outdir:
+        elevation_fp = args.outdir
+    elif config is not None:
+        elevation_fp = Path(config["additional_data"]["elevation"] or "elevation")
```
nilsnolde, line 305: this is not quite enough to make `config` optional as it's also used further down when the user wants to get tiles covering the graph (often the most sensible option IMO in case elevation is only used to enrich the graph, not the `/height` service). admittedly there was no exception handling in place before either, `config` wasn't even `required`. 

guess it's enough to protect line 315 from missing config path.

nilsnolde (changes requested):
thanks, makes sense! just a few minor things.

scripts/valhalla_build_elevation
ianthetechie: haha ;) On my own projects I usually just run `black` so I don't have to think about it and usually from imports are in different blocks, but I'll just revert this and throw in enum alphabetically since this isn't my codebase :P

nilsnolde: yeah, would love to do that here too but installation blabla.. really, smth like `pre-commit` which runs `format.sh` would make SO much sense for valhalla as a whole (but also to run black/flake8 on python code), thousands of pointless CI hours of formatting commits saved. think I'll tinker with a shell script to do that soonish!

> usually from imports are in different blocks

interesting, never noticed! thought it's just std, installed and local modules being separated

scripts/valhalla_build_elevation
ianthetechie: I guess I should point out the motivation for touching this as well. While the normal case would have a Valhalla JSON config, I think it's generally useful to have an option to run this script without that context, such as to download the whole planet elevation data to an internal file server. This is probably a minority use case, but I think it's useful to be able to operate as the previous shell script without a JSON config.

That said, good call. I neglected the possible code path on line 315. Any objection to checking that config is not none, exiting if it's not, and updating docs? This might be possible in argparse, but I don't know a great way off the top of my head.

scripts/valhalla_build_elevation
ianthetechie: Yeah I didn't actually run it here, but in previous pipelines I'd set up with Python code I just run black in CI and fail if it detects any changes. Might be something to reduce cognitive overhead in the future.

scripts/valhalla_build_elevation
nilsnolde: yeah I figured so, should've commented that I thought it's a valid use case, now the script can be used independently of valhalla. just get it from github's raw endpoint and execute it to get some bbox's tilezen tiles, that's great. would you mind adding some notes about that to the PR description for future reference?

> Any objection to checking that config is not none, exiting if it's not, and updating docs

that's what I had in mind too:)

scripts/valhalla_build_elevation
```diff
@@ -235,22 +289,45 @@ if __name__ == "__main__":
...
         tiles = get_tiles_with_geojson(args.from_geojson)
     elif args.from_bbox:
         tiles = get_tiles_with_bbox(args.from_bbox)
     elif args.from_tiles:
-        tiles = get_tiles_with_graph(Path(config["mjolnir"]["tile_dir"]))
+        if config is not None:
+            tiles = get_tiles_with_graph(Path(config["mjolnir"]["tile_dir"]))
+        else:
+            LOGGER.critical("--from-tiles requires a config to be specified.")
+            sys.exit(1)
```
nilsnolde, line 314: really nit:
```suggestion
        if config is None:
            LOGGER.critical("--from-tiles requires a config to be specified.")
            sys.exit(1)
        tiles = get_tiles_with_graph(Path(config["mjolnir"]["tile_dir"]))
```

mostly to save one line I guess :sweat_smile: feel free to disregard

nilsnolde (dismissed):

scripts/valhalla_build_elevation
ianthetechie: Sure thing; I'll update the PR description.

nilsnolde (approved):
thanks for the time and effort @ianthetechie !

PR #3616, opened 2022-05-04
remove geos in favor of boost geometry

### LINKED ISSUE #3388: Can't build Valhalla with geos-3.10.x
Hello,
I tried to build Valhalla on macOS Monterey.  It seems that the new geos version has changed something in it's API. The build produces the following error:
`Undefined symbols for architecture x84_64: geos::geom::Coordinate::Coordinate(), ....`
I could fix this on my system by building geos-3.9.3 from source and install it.

Homebrew by default installs geos-3.10.1 at the moment and it is not possible to install older versions of geos.

### LINKED ISSUE #3396: Replace geos with boost.geometry?
was just looking at admin building and realized that geos seems to only be used in https://github.com/valhalla/valhalla/blob/master/src/mjolnir/adminbuilder.cc. 

quickly skimming the code I don't think I saw stuff that's not possible with boost.geometry.. it's mostly polygon-in-polygon checks (I think) and while bg does not yet support all kinds of geom types for `within` checks, (multi)polygons should work well: https://www.boost.org/doc/libs/1_77_0/libs/geometry/doc/html/geometry/reference/algorithms/within/within_2.html.

in case we eventually get to choose our own boost version (hopefully close to latest) after finishing #3346 , it might be little effort to ditch geos in favor of bg? one more linked lib gone would be awesome (esp such a troublesome one for valhalla, using the c++ interface)

nilsnolde:
fixes #3396, fixes #3388

I didn't build more than andorra yet with this, but the tests pass at least.. I'll build the world tmrw and compare to the most recent fossgis admin db. md5's _should_ be the same, the inner rings are inserted the same and windedness rules are same in both libs.

realized too late that I based the branch off the spatialite fix PR branch.. it's just the commits though, the rest was merged already, and we squash anyways..

and I wanna add how much I like this elegant algorithm, kudos @ktatterso :)


src/loki/polygon_search.cc
```diff
@@ -33,26 +33,12 @@ static const auto Haversine = [] {
...
-  if (!ring.empty() && is_open) {
-    ring.push_back(ring[0]);
-  }
-
-  // reverse ring if counter-clockwise
-  if (vm::polygon_area(ring) > 0) {
-    std::reverse(ring.begin(), ring.end());
-  }
-}
-
```
nilsnolde, line 49: we needed that in pre-header-boost times, not anymore

src/mjolnir/adminbuilder.cc
```diff
@@ -12,53 +18,21 @@
...
 
 namespace {
 
 struct polygondata {
-  Polygon* polygon;
-  LinearRing* ring;
+  polygon_t polygon;
   double area;
-  int iscontained;
   unsigned containedbyid;
```
nilsnolde, line 36: these members weren't needed anymore.. `ring` is unnecessary with boost geometry and `isocontained` is redundant since it's only 1 if `containedbyid != 0` so all checks can be made with the latter instead

src/mjolnir/adminbuilder.cc
```diff
@@ -75,83 +49,47 @@ int polygondata_comparearea(const void* vp1, const void* vp2) {
...
-#endif
-
-  LineMerger merger;
-  merger.add(mline.get());
-#if GEOS_VERSION_MAJOR == 3 && GEOS_VERSION_MINOR >= 9
-  std::vector<std::unique_ptr<LineString>> merged(merger.getMergedLineStrings());
-#else
-  std::unique_ptr<std::vector<LineString*>> merged(merger.getMergedLineStrings());
-#endif
-  WKTWriter writer;
```
nilsnolde, line 94: I didn't fully understand why there was so much geometry massaging going on when the end geometry was actually pretty simple. probably some geos niceties.. that could now be simplified a lot. basically we already give this function polygons, do the processing based on them and then later build up the final `MultiPolygon` which is returned as WKT. no intermediary geometry massaging

src/mjolnir/adminbuilder.cc
```diff
@@ -75,83 +49,47 @@ int polygondata_comparearea(const void* vp1, const void* vp2) {
...
-#else
-      polys[totalpolys].polygon = gf->createPolygon(gf->createLinearRing(pline->getCoordinates()), 0);
-      polys[totalpolys].ring = gf->createLinearRing(pline->getCoordinates());
-#endif
-      polys[totalpolys].area = polys[totalpolys].polygon->getArea();
-      polys[totalpolys].iscontained = 0;
+  for (const auto& polygon : polygons) {
+    if (polygon.outer().size() > 3) {
+      polys[totalpolys].polygon = polygon;
+      polys[totalpolys].area = bg::area(polygon);
```
nilsnolde, line 62: bg makes it so much more legible:)

src/mjolnir/adminbuilder.cc
```diff
@@ -75,83 +49,47 @@ int polygondata_comparearea(const void* vp1, const void* vp2) {
...
-  WKTWriter writer;
-
-  // Procces ways into lines or simple polygon list
-#if GEOS_VERSION_MAJOR == 3 && GEOS_VERSION_MINOR >= 9
-  polygondata* polys = new polygondata[merged.size()];
-#else
-  polygondata* polys = new polygondata[merged->size()];
-#endif
+std::string GetWkt(std::vector<polygon_t>& polygons) {
+  std::ostringstream wkt;
```
nilsnolde, line 53: it's easy to miss and not so important: but this function only returns a single WKT string now. before that was a vector of WKT strings. however, the `wkts` vector never had more than one element, so no need for a vector at all

src/mjolnir/adminbuilder.cc
```diff
@@ -398,82 +305,63 @@ void BuildAdminFromPBF(const boost::property_tree::ptree& pt,
...
       if (has_data) {
-
-        std::unique_ptr<Geometry> mline(gf->createMultiLineString(lines.release()));
-        std::vector<std::string> wkts = GetWkts(mline);
+        std::string wkt = GetWkt(polygons);
         std::string name;
         std::string name_en;
         std::string iso;
 
-        for (const auto& wkt : wkts) {
```
nilsnolde, line 433: again, this `wkts` vector only ever had one element

src/mjolnir/adminbuilder.cc
```diff
@@ -398,82 +305,63 @@ void BuildAdminFromPBF(const boost::property_tree::ptree& pt,
...
-            sqlite3_bind_text(stmt, 2, iso.c_str(), iso.length(), SQLITE_STATIC);
-          } else {
-            sqlite3_bind_null(stmt, 2);
-          }
+        if (admin.iso_code_index()) {
+          iso = osm_admin_data.name_offset_map.name(admin.iso_code_index());
+          sqlite3_bind_text(stmt, 2, iso.c_str(), iso.length(), SQLITE_STATIC);
+        } else {
+          sqlite3_bind_null(stmt, 2);
+        }
```
nilsnolde, line 335: nothing happened here.. just git weirdness

nilsnolde: ah bullshit, it's the removed loop..

src/mjolnir/adminbuilder.cc
```diff
@@ -75,83 +49,47 @@ int polygondata_comparearea(const void* vp1, const void* vp2) {
...
-#if GEOS_VERSION_MAJOR == 3 && GEOS_VERSION_MINOR >= 9
-  polygondata* polys = new polygondata[merged.size()];
-#else
-  polygondata* polys = new polygondata[merged->size()];
-#endif
+std::string GetWkt(std::vector<polygon_t>& polygons) {
+  std::ostringstream wkt;
+
+  // something to hold the polygons for processing
+  polygondata* polys = new polygondata[polygons.size()];
```
nilsnolde, line 56: should change that to `polygondata polys[polygons.size()]`. already deleted the `delete` further down or this `new`

gknisely:
Hi @nilsnolde.  What testing has been done with this update?  I would probably run a conversion on a full planet with master code and then again on the branch for the building of the admin db.  Also, we should figure out how to diff the two admins.  Maybe just a dump of the two tables from each db and compare?  We should do this as we only have minimal testing for the admin db since gurka testing was introduced much later on.

nilsnolde:
I mentioned it already I think: I didn't change any of the actual logic, it's only simplifying and refactoring the existing logic to boost

nilsnolde:
oh jeez, I'm an idiot.. total misconception.. will take a bit longer

nilsnolde:
I'll open another one, basically all comments are worthless :sweat_smile:

PR #3619, opened 2022-05-09
Few admin fixes & suggestions

nilsnolde:
one french territory changed its name. also one suggestion, but I'm not so sure about it myself.


lua/admin.lua
```diff
@@ -141,7 +141,7 @@ function rels_proc (kv, nokeys)
      end
 
      if kv["admin_level"] == "2" then 
-        if kv["name"] ==  "France" then
+        if kv["name"] ==  "France" or kv["name"] == "United Kingdom" then
```
nilsnolde, line 144: I think I just found it strange to think of england/scotland/wales as a state. the change suggestion is very similar to how we handle france IMO. we promote french territories to countries and throw away the french "kingdom"/empire. england etc don't have a concept of states, but they do some autonomy. plus scotland has its own national soccer team ;)

I also wanted to do the same with netherlands, but that's a mess in osm, wouldn't really work..

nilsnolde:
in case anyone thinks this broke master: I just accidentally canceled the build while cancelling the docs PR

PR #3621, opened 2022-05-10
support nulls in config-based speed assignment

kevinkreiser:
A while back we added support for configuring the default speeds of edges in the graph via data provided over in https://github.com/OpenStreetMapSpeeds/schema/blob/master/default_speeds.json

Semi-recently we were able to do a full planet build of that data but where there are holes in it we put nulls in the json. What we forgot to do was come back to the code here and make it robust certain entries being `null`. This PR does that.

I've added a unit test that shows a config with mixed `null` and non-`null` values properly do a partial speed assignment, setting those configured values to a given edge but not setting them when `null` is present.

### DIFF
--- a/src/mjolnir/speed_assigner.h
+++ b/src/mjolnir/speed_assigner.h
@@ -65,6 +65,7 @@ The json basically looks like this:
 
 constexpr uint32_t kMinSuburbanDensity = 5;
 constexpr uint32_t kMaxSuburbanDensity = 10;
+constexpr uint32_t kUnconfiguredSpeed = 0;
 
 /**
  * This class has two methods of deciding the default speed of an edge. The default method is to use
@@ -93,7 +94,7 @@ protected:
         throw std::runtime_error(name + " must have " + std::to_string(entries.size()) + " speeds");
       size_t i = 0;
       for (const auto& speed : arr) {
-        entries[i++] = speed.GetUint();
+        entries[i++] = speed.IsNull() ? kUnconfiguredSpeed : speed.GetUint();
       }
     }
 
@@ -102,10 +103,12 @@ protected:
       parse(obj, "link_exiting", link_exiting);
       parse(obj, "link_turning", link_turning);
       parse(obj, "roundabout", roundabout);
-      service[0] = obj["driveway"].GetUint();
-      service[1] = obj["alley"].GetUint();
-      service[2] = obj["parking_aisle"].GetUint();
-      service[3] = obj["drive-through"].GetUint();
+      service[0] = obj["driveway"].IsNull() ? kUnconfiguredSpeed : obj["driveway"].GetUint();
+      service[1] = obj["alley"].IsNull() ? kUnconfiguredSpeed : obj["alley"].GetUint();
+      service[2] =
+          obj["parking_aisle"].IsNull() ? kUnconfiguredSpeed : obj["parking_aisle"].GetUint();
+      service[3] =
+          obj["drive-through"].IsNull() ? kUnconfiguredSpeed : obj["drive-through"].GetUint();
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/mjolnir/speed_assigner.h
```diff
@@ -65,6 +65,7 @@ The json basically looks like this:
 
 constexpr uint32_t kMinSuburbanDensity = 5;
 constexpr uint32_t kMaxSuburbanDensity = 10;
+constexpr uint32_t kUnconfiguredSpeed = 0;
```
kevinkreiser, line 68: sentinal value that indicates no configured speed exists for a particular directed edge

src/mjolnir/speed_assigner.h
```diff
@@ -93,7 +94,7 @@ class SpeedAssigner {
         throw std::runtime_error(name + " must have " + std::to_string(entries.size()) + " speeds");
       size_t i = 0;
       for (const auto& speed : arr) {
-        entries[i++] = speed.GetUint();
+        entries[i++] = speed.IsNull() ? kUnconfiguredSpeed : speed.GetUint();
```
kevinkreiser, line 97: this and the other several below are used to fill the speed tables with an invalid sentinal value so that when the caller asks the assigner for a speed that would have been a null it will get back that invalid value. previously calling `GetUnit` on a `null` value threw and exception and disabled the config based speed assignment entirely :frowning_face:

src/mjolnir/speed_assigner.h
```diff
@@ -115,23 +118,28 @@ class SpeedAssigner {
...
-                  const std::string& country,
-                  const std::string& state) const {
+  uint32_t FromConfig(DirectedEdge& directededge,
+                      const uint32_t density,
+                      const std::string& country,
+                      const std::string& state) const {
+    // nothing is configured
+    if (tables.empty()) {
+      return kUnconfiguredSpeed;
+    }
```
kevinkreiser, line 137: this was down below in the calling function, i encapsulated it here just for organizational purposes

src/mjolnir/speed_assigner.h
```diff
@@ -253,8 +254,10 @@ class SpeedAssigner {
                    const std::string& country_code,
                    const std::string& state_code) const {
 
-    // If we have config loaded we'll use that
-    if (!tables.empty() && FromConfig(directededge, density, country_code, state_code)) {
+    // See if we can get a valid speed loaded from configuration
+    auto configured_speed = FromConfig(directededge, density, country_code, state_code);
+    if (configured_speed != kUnconfiguredSpeed) {
+      directededge.set_speed(configured_speed);
```
kevinkreiser, line 260: the main change here, other than storing an invalid value in the speed tables when `null` is encountered, is that we changed the signature of the `FromConfig` method. instead of setting the value on the edge and returning a boolean saying whether it did the assignment or not, it simply returns the speed it finds and the caller (this function) does with it what it wants. in this case it says, hey, was that a valid speed, if so i'll take it and move on with my life otherwise ill use the heuristic to figure out a speed to use for this edge

test/gurka/test_config_speed.cc
```diff
@@ -623,3 +624,32 @@ TEST(Standalone, Malformed) {
...
+        "roundabout": [11,11,11,11,11,11,11,11], "driveway": 11, "alley": 11, "parking_aisle": 11, "drive-through": 11
+      }
+    }]
+  )";
+  speed_config.close();
+  testable_assigner assigner("test/data/speed_config.json");
+  ASSERT_FALSE(assigner.tables.empty());
+  ASSERT_FALSE(assigner.UpdateSpeed(edge, 6, true, "foo", "bar"));
+  ASSERT_TRUE(assigner.UpdateSpeed(edge, 1, true, "foo", "bar"));
+  ASSERT_EQ(edge.speed(), 11);
```
kevinkreiser, line 654: here we mix nulls and valid speeds and we make sure that:

1. the speed tables load
2. asking for a nulled out value doesnt assign a speed from the config
3. asking for a speed that does exist sets it properly

nilsnolde (approved):
looks good, thanks!

PR #3622, opened 2022-05-10
build: add `pre-commit` script

### LINKED ISSUE #3613: proper pre-commit script
before I go ahead with it: what you think about adding a proper pre-commit script to valhalla? mostly to try and prevent all these format.sh commits.. it can also be used in CI of course.

I was thinking a bash script which installs python's `pre-commit` library. `pre-commit` needs a small simple config file in this repo but is battle-tested and very flexible. with that we could add the existing scripts as target, but also do some python linting/formatting with appropriate tools (I feel we have enough python in this repo to be a bit more mindful about it;)). in total it'd be 2 new files (one script, one config).

so a dev would only run `install_pre-commit.sh` or so and that installs the `pre-commit` library to their python env which will then run all targets on each commit. we could also add the tidy script on pre-push events (though that's much less likely to cause failures and often takes a long time to run).

thoughts?

vinayakkulkarni:
# Issue

- [x] Fixes #3613

## Requirements / Relations

- [x] Enables `pre-commit` & `commit-msg` git hook

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 99
+
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 99
--- a/.github/workflows/lint-pr.yml
+++ b/.github/workflows/lint-pr.yml
@@ -1,0 +1,17 @@
+name: 'Lint PR'
+
+on:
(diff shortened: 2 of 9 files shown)


nilsnolde:
thanks @vinayakkulkarni ! honestly IMHO we shouldn't include js frameworks at this point if we don't have to. that requires everyone to have at least `npm`. and it's quite a lot of files touched. when I opened the issue I was more thinking about [`pre-commit`](https://pre-commit.com), which only needs one small config file, runs in python (default available on linux/mac), easily runs our `format.sh` script and can also lint/format our python code of which we have quite a bit.

kevinkreiser:
i agree with @nilsnolde a single pre-commit config yaml is all we need and python is already a requirement

PR #3635, opened 2022-05-17
Fix how max_matrix_distance is found

### LINKED ISSUE #3634: valhalla_run_matrix fails
Fails to find the proper max_matrix_distance and exits.

dnesbitt61:
# Issue

#fixes #3634

## Tasklist

 - [ ] Add tests

## Requirements / Relations


kevinkreiser:
we really have to refactor these `run` binaries to use `actor` or `worker`. every time we refactor or add stuff we always break these things...

kevinkreiser (dismissed):

dnesbitt61:
I thought about that for a minute or so, but don't have much time today. one thing this does though is compare CostMatrix to TimeDistanceMatrix which can be useful.

kevinkreiser:
yep, to do it right we need to build into the workers/actor a way to control a bit what it does underneath. totally understand not having enough time for it!

PR #3640, opened 2022-05-24
chore(tooling): add support for different linters in `pre-commit`

vinayakkulkarni:
# Issue

Adds support for linting `Dockerfile`, `.yml`, `.json` & `.md`

## Requirements / Relations

https://github.com/valhalla/valhalla/pull/3637#discussion_r876815876


nilsnolde (changes requested):
personally I wouldn't have thought (or maybe even cared :sweat_smile:) of all the aux files (*.json/.yaml/.md etc) but it's not hurting that much either I guess (except for the first time someone runs it and has to pull the hadolint image). btw, in this PR we should include CI to run `pre-commit` before merging or there'll be a mismatch between people installing this and people who don't and CI still passes.

some weird things in the debug build, but unrelated to your changes.

nilsnolde:
I'll close this one, IMO too many new dependencies for little gain. The important parts for us are c++ and python files IMO and those are covered now. Sorry for the wasted work, best to open an issue next time and discuss specifics before PRing.

PR #3648, opened 2022-06-06
adding max axles count

mattbenson247:
Adds `axle_count` to the truck costing options.


src/mjolnir/pbfgraphparser.cc
```diff
@@ -748,6 +748,14 @@ struct graph_callback : public OSMPBF::Callback {
       restriction.set_modes(kTruckAccess);
       osmdata_.access_restrictions.insert(
           AccessRestrictionsMultiMap::value_type(osmid_, restriction));
+    };
+     tag_handlers_["maxaxles"] = [this]() {
+      OSMAccessRestriction restriction;
+      restriction.set_type(AccessType::kMaxAxles);
+      restriction.set_value(std::stof(tag_.second) * 100);
```
kevinkreiser, line 755: whats the 100 here for? i couldnt seem to find an entry in the wiki for this tag and taginfo shows its very sparingly populated BUT of the 79 ways that do have it, the values are 2 and 3.

kevinkreiser: oh, also axles should be positive whole numbers i guess :D maybe `std::stoul` makes more sense

mattbenson247: Damn you are quick. Thanks for that! I was just creating the shell of what we need to gather and will submit it later, likely with a few questions attached. The calc there is not obviously unnecessary. 

Additionally, I agree that the taginfo:maxaxle(s) is sparingly used. However, I work for a trucking technology company, and we have seen issues in the past with a few places [(the Holland Tunnel specifically that we provided routing for)](https://valhalla.github.io/demos/routing/#loc=15,40.734104,-74.025385&locations=[{%22lat%22:40.7416970923627,%22lon%22:-74.04574133455752},{%22lat%22:40.73642941794871,%22lon%22:-74.00102354586123}]&costing=%22auto%22). 
I recently added the Holland Tunnel maxheight to OSM for security there. However, since there is no weight restriction for the HT, no axleLoad restriction, and only an axle count restriction, the routing is incorrectly allowing a >3 axle truck to drive through, when the Port Authority of NY/NJ doesn't allow this.  [HT OSM way](https://www.openstreetmap.org/edit#map=17/40.72823/-74.02449). [PANYNJ Restriciton PDF](https://www.panynj.gov/content/dam/bridges-tunnels/pdfs/truckers-guide.pdf)

I understand this is a one off circumstance, however, a simple fix to add the axle count to the restrictions should a decent safeguard for the trucking community. That being said, we will likely also be updating OSM data when we receive updates/comments from our truck driver community.

kevinkreiser: oh yeah i think its great to add the axle restruction regardless of how well populated it is today, no worries about that! the extra `100` factor was why i was saying i couldnt find any information about it. i was imaging that maybe maxaxles were in the units `centiaxles` or something  ;) hence the scaling it back to integer but it seems to be the more obvious unit of axles itself haha

src/mjolnir/pbfgraphparser.cc
```diff
@@ -748,6 +748,14 @@ struct graph_callback : public OSMPBF::Callback {
       restriction.set_modes(kTruckAccess);
       osmdata_.access_restrictions.insert(
           AccessRestrictionsMultiMap::value_type(osmid_, restriction));
+    };
+     tag_handlers_["maxaxles"] = [this]() {
+      OSMAccessRestriction restriction;
+      restriction.set_type(AccessType::kMaxAxles);
+      restriction.set_value(std::stoul(tag_.second));
```
mattbenson247, line 755: @kevinkreiser Thanks. I am not a C dev, so any help would be much appreciated here. I believe your were correct with this recommendation. Additionally, I changed the graph.lua to parse the kv to a number, and the truckcost.cc to a `cast int`.  Let me know what else I might be missing here. Or if there is anything larger I really need to get this merged. Thanks again for the help man. Much appreciated.

test/parse_request.cc
```diff
@@ -138,6 +138,7 @@ constexpr float kDefaultTruck_CountryCrossingPenalty = 0.0f;   // Seconds
 constexpr float kDefaultTruck_LowClassPenalty = 30.0f;         // Seconds
 constexpr float kDefaultTruck_TruckWeight = 21.77f;            // Metric Tons (48,000 lbs)
 constexpr float kDefaultTruck_TruckAxleLoad = 9.07f;           // Metric Tons (20,000 lbs)
+constexpr float kDefaultTruck_TruckAxles = 3.f;                // Count
```
kevinkreiser, line 141: the rest of the defaults are for semi trucks which have 5 axles right? 3 for the tractor and 2 more for the trailer

mattbenson247: There are a bunch of classes/configurations and trailer types for hauling. Here is a decent overview: https://www.fhwa.dot.gov/policyinformation/tmguide/tmg_2013/vehicle-types.cfm

As for the 2, 3, 5 I added... those were just random.

kevinkreiser: oh jeez im sorry i thought i was looking at the truck costing defaults. my eyes saw the names of the constants and i didnt even look at the file name. i thought i was looking here: https://github.com/valhalla/valhalla/blob/e8220b7e82411b310eb275b01f4774f5bfdb2409/src/sif/truckcost.cc#L50-L54

CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 ## Release Date: 2021-??-?? Valhalla 3.1.5
+
+## Release Date: 2016-01-04
+* **Removed**
+* **Bug Fix**
+* **Enhancement** 
+   * ADDED: MaxAxle count to restrictions [#3648](https://github.com/valhalla/valhalla/pull/3648)
+
```
nilsnolde, line 8: put this item under the existing `Enhancement` section further down at the end of the list and remove the rest

nilsnolde (changes requested):
IMO this looks good. just some changelog issues and you'd need to run `./scripts/format.sh` locally to satisfy CI (since a few weeks this'll set up `pre-commit` so you won't have to worry about it anymore in the future:)).

nilsnolde:
@mattbenson247 if you don't want/have time to finish this off, I'll clone your fork in the next days

nilsnolde:
The whole request parsing stuff was missing, I added it quickly.

nilsnolde (dismissed):

proto/options.proto
```diff
@@ -307,6 +307,9 @@ message Costing {
     oneof has_elevator_penalty {
       float elevator_penalty = 79;
     }
+    oneof has_axle_count {
+      int32 axle_count = 80;
+    }
```
kevinkreiser, line 312: we can use defaults here to avoid `oneof` since 0 is meaningless

src/sif/truckcost.cc
```diff
@@ -268,12 +270,13 @@ class TruckCost : public DynamicCost {
...
-  float height_;    // Vehicle height in meters
-  float width_;     // Vehicle width in meters
-  float length_;    // Vehicle length in meters
+  bool hazmat_;         // Carrying hazardous materials
+  float weight_;        // Vehicle weight in metric tons
+  float axle_load_;     // Axle load weight in metric tons
+  float height_;        // Vehicle height in meters
+  float width_;         // Vehicle width in meters
+  float length_;        // Vehicle length in meters
+  uint32_t axle_count_; // Vehicle axle count
```
kevinkreiser, line 279: can use `uint8_t` here and for the range above

proto/options.proto
kevinkreiser: we alrady use 80 for a recent pr we just merged.. have to update this

src/sif/truckcost.cc
nilsnolde: changed it to uint8_t

nilsnolde:
I fixed up the remaining thing from the review and resolved the master conflicts. I think I'll throw in a quick actual test, so far it's only testing the json parsing. Then we can merge on Wednesday.

test/gurka/test_truck_restrictions.cc
```diff
@@ -0,0 +1,66 @@
...
+        // all restrictions should be higher than our defaults, so we can actually see the impact of
+        // any single one
+        {"BC",
+         {{"highway", "residential"},
+          {"maxheight", "5"},
+          {"maxlength", "25"},
+          {"maxwidth", "3"},
+          {"hazmat", "no"},
+          {"maxaxles", "8"},
+          {"maxaxleload", "10"}}},
```
nilsnolde, line 29: I think truck restrictions weren't tested before, so I added a test for all of them. The map allows for Valhalla's truck defaults (4.11 m height etc) so we can verify that it's actually fine if no `costing_options` are set.

docs/api/turn-by-turn/api-reference.md
```diff
@@ -121,8 +121,8 @@ The following options are available for `auto`, `bus`, `taxi`, and `truck` costi
 
 | Vehicle Options | Description |
 | :-------------------------- | :----------- |
-| `height` | The height of the vehicle (in meters). Default 1.9 for car, bus, taxi and 2.6 for truck. |
-| `width` | The width of the vehicle (in meters). Default 1.6 for car, bus, taxi and 4.11 for truck. |
+| `height` | The height of the vehicle (in meters). Default 1.9 for car, bus, taxi and 4.11 for truck. |
+| `width` | The width of the vehicle (in meters). Default 1.6 for car, bus, taxi and 2.6 for truck. |
```
nilsnolde, line 125: I mixed up these two when I added them..

kevinkreiser (approved):

PR #3652, opened 2022-06-12
Forgot to adapt driving side and country access rules for UK

nilsnolde:
we got rid of UK as a country in [#3619](https://github.com/valhalla/valhalla/pull/3619), but I totally forgot to search everything for UK :face_exhaling: I replaced all UK logic with the new countries.


nilsnolde:
this is urgent I'd say and really sorry that slipped the cracks everyone. hope not too many were routing in london on the right side..

kevinkreiser (approved):
Good news is it only affected people who built new admins since it was merged. I suspect that number isn't too high. Clearly we need better testing of this stuff but it's generally so stable it's hard to justify the time spent

PR #3654, opened 2022-06-15
Fix AABB2::Intersect() - y_intersect()

DmitryAk:
# Issue

#3656 
Found during static code analyser running
The condition:
https://github.com/valhalla/valhalla/blob/de5e4a72332a0331099d26f8becfc95f9329d96b/src/midgard/util.cc#L555
is always `false`. `NAN` was returned mistakenly.

What is the best way to cover this code with tests?

## Tasklist

 - [ ] Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## Release Date: 2021-??-?? Valhalla 3.1.5
 * **Removed**
 * **Bug Fix**
-   * FIXED: Fix AABB2::Intersect(...) - horizontal line intersections y_intersect(...) returning NAN [#3654](https://github.com/valhalla/valhalla/pull/3654)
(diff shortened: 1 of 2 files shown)


kevinkreiser:
yeah i checked and deleting this code and the tests is totally doable. fwiw i prefer to remove this code and further minimize the codebase.

kevinkreiser:
if you agree, i can push up what i have locally and we can merge that instead of this. let me know your opinion

DmitryAk:
> ng this code and the tests is totally doable. fwiw i prefer to remove this code and further minimize the codebase.


> if you agree, i can push up what i have locally and we can merge that instead of this. let me know your opinion

I agree, if this code can be deleted - this looks good to me.
Closing this,
Thanks!

PR #3655, opened 2022-06-15
Fix is_conditional_active(...)

DmitryAk:
# Issue

#3656 
Found during static code analyzer running
The same condition is checked twice: https://github.com/valhalla/valhalla/blob/de5e4a72332a0331099d26f8becfc95f9329d96b/src/baldr/datetime.cc#L368
It seems like in the second case `!e_week` should be checked.

What is the best way to cover this with test?

## Tasklist

 - [ ] Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Release Date: 2021-??-?? Valhalla 3.1.5
 * **Removed**
 * **Bug Fix**
+   * FIXED: DateTime::is_conditional_active(...) incorrect end week handling [#3655](https://github.com/valhalla/valhalla/pull/3655)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Release Date: 2021-??-?? Valhalla 3.1.5
 * **Removed**
 * **Bug Fix**
+   * FIXED: DateTime::is_conditional_active(...) incorrect end week handling [#3655](https://github.com/valhalla/valhalla/pull/3655)
```
kevinkreiser, line 4: can you please move this to the bottom of the list. for whatever reason we do newest at the bottom

kevinkreiser (approved):

kevinkreiser (approved):

PR #3657, opened 2022-06-15
remove unused methods from midgard

kevinkreiser:
in #3654 it was noticed that a pretty serious typo was breaking the intercept function. upon further inspection though, it turns out these methods arent even used anywhere (must have been at one time or came over with the initial code dump from @dnesbitt61) at any rate, im of the opinion that we should trim out as much unused code as possible since this reduces maintainance and increases readability (findability really i guess)

### DIFF
--- a/scripts/valhalla_build_timezones
+++ b/scripts/valhalla_build_timezones
@@ -2,7 +2,7 @@
 
 error_exit() {
   echo "error: $1" 1>&2
-  #exit $(pkg-config geos --modversion | grep -cvF "3.9")
+  exit $(pkg-config geos --modversion | grep -cvF "3.9")
 }
 
 if  ! which spatialite >/dev/null; then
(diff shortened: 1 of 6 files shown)


scripts/valhalla_build_timezones
```diff
@@ -2,7 +2,7 @@
 
 error_exit() {
   echo "error: $1" 1>&2
-  exit $(pkg-config geos --modversion | grep -cvF "3.9")
+  #exit $(pkg-config geos --modversion | grep -cvF "3.9")
```
kevinkreiser, line 5: this is unrelated to this pr and i can remove it but a previous, semi-recent pr, broke this for geos 3.9 users. i'll have a fix for it in the chinese postman pr

dnesbitt61 (commented):
Wouldn't this have use to determine tiles/cells intersected by a route shape? I guess a more efficient method uses the Bresenham code?

kevinkreiser:
the code is completley unused, other than the unit tests. yes for the cell intersection stuff i rolled my own in bresenhams

PR #3659, opened 2022-06-16
Fix tiny bug in timedistance bss matrix

xlqian:
# Issue

The PR targeting a tiny bug in timedistance bss matrix

move the flag into the loop

## Tasklist

### DIFF
--- a/src/thor/timedistancebssmatrix.cc
+++ b/src/thor/timedistancebssmatrix.cc
@@ -590,12 +590,12 @@ void TimeDistanceBSSMatrix::SetDestinationsOneToMany(
   // For each destination
   uint32_t idx = 0;
 
-  // Only skip outbound edges if we have other options
-  bool has_other_edges = false;
-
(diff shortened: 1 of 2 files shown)


src/thor/timedistancebssmatrix.cc
```diff
@@ -590,13 +590,13 @@ void TimeDistanceBSSMatrix::SetDestinationsOneToMany(
...
 
-  // Only skip outbound edges if we have other options
-  bool has_other_edges = false;
-
   for (const auto& loc : locations) {
     // Set up the destination - consider each possible location edge.
     bool added = false;
 
+    // Only skip outbound edges if we have other options
+    bool has_other_edges = false;
```
kevinkreiser, line 598: I see, it's per individual location not for all locations. Great catch!

xlqian: copy-paste is evil :D

Issue #3628, opened 2022-05-16
GTFS Support: Data Access and Parsing


kevinkreiser:
The first step in adding GTFS support to the library will be to make sure we can read GTFS and that we have tests proving that that is the case. To my mind this would be best accomplished by:

1. adding a submodule to the thirdparty directory this branch of th is repo: https://github.com/mapsme/just_gtfs/tree/cpp14 (also its behind master might want to check to see what is missing there or ask the authors if they are thinking they will update it)
2. make a new gurka test, the test wont need a map just yet so dont worry about that instead add a test that should use the just_gtfs library to try to load some gtfs data
3. in the test, before loading the data you should create some gtfs data directly (by having the test write some gtfs files to disk). no need to make it  realistic data, in fact use strings like foo and bar or whatever, be creative. focus on adding all of the different features we can take advantage of. its important to add all of the things that our downstream processing will be able to make use of so that we'll have full test coverage of the code. use specification section here: https://github.com/valhalla/valhalla/issues/3539#issue-1140844006 to see all of the differen files/fields the code can support. all we want is a minimal example so you only need to get one of each field in the mocked gfts data. of course one trip could use most of the fields in just 2 stops so yeah feel free to keep it very minimal. start small and build up. we can spread this out over multiple prs if that is helpful.
4. once you have some amount of the data loading from your faked up gtfs data now you need to throw some gtest expect/asserts into the tests to make sure that the parser is getting all of the stuff you put in the files into its api. this will both familiarize you with the gtfs spec and the libraries apis for accessing it.

This task can be considered done when we have:

1. just_gtfs's c++14 branch as a thirdparty submodule (and the required changes to CMakeLists.txt to get access to it, can follow other header only examples)
2. a mocked up gtfs feed or two that expose all the required/optional fields
3. a suite of gurka tests (no gurka map yet) which use the parser to load the gtfs data and assert that the data we expected is accessible from the library

kevinkreiser:
we should also use a gtfs validator to check our mocks before moving forward

kevinkreiser:
this was fixed in #3665

PR #3660, opened 2022-06-19
GTFS Support: Data Access and Parsing

chris-jpark:
# Issue

GTFS Support: Data Access and Parsing #3628 

Tests the ability to read and write GTFS files using the [just_gtfs](https://github.com/mapsme/just_gtfs/tree/cpp14) submodule. Imports the submodule into valhalla and creates a gurka test writing a GTFS folder that includes all necessary attributes used by valhalla. Tests that we can read and write GTFS files properly in valhalla.


nilsnolde (commented):
thanks for the start @chris-jpark ! got a few comments:
- we need to take out all the stuff from the shortcuts PR (easiest is to revert the gtfs commits and change to a new branch based on master, the commit in that branch and open a new PR)
- I think we're ok without the `test/data/gtfs_sample` data which seems to be an altered toronto dataset, since we're able to roll our own gtfs data

 a bit more inline.

what do you think @kevinkreiser ?

test/gurka/test_gtfs.cc
```diff
@@ -0,0 +1,137 @@
...
+  // write frequencies.txt
+  struct Frequency freqOne {
+    .trip_id = "bar", .start_time = Time(0, 0, 0), .end_time = Time(2, 0, 0), .headway_secs = 1800,
+    .exact_times = gtfs::FrequencyTripService::FrequencyBased,
+  };
+  feed.add_frequency(freqOne);
+  feed.write_frequencies("../test/data/gtfs_test/");
+
+  // make sure files are actually written
+  const auto& trips = feed.get_trips();
```
nilsnolde, line 125: you'll need to read the feed off disk, i.e. create a new feed from the directory path

test/gurka/test_gtfs.cc
```diff
@@ -0,0 +1,137 @@
...
+// test to write gtfs files
+TEST(GtfsExample, WriteGtfs) {
+
+  Feed feed;
+
+  // write agency.txt
+  struct Agency ttc {
+    .agency_id = "TTC", .agency_name = "Toronto Commission", .agency_url = "http://www.ttc.ca",
+    .agency_timezone = "America/Toronto"
+  };
```
nilsnolde, line 28: for now this test is fine I think. eventually we might want to encapsulate the gtfs writer in a function which we can call from single test suites where we have a bit control on e.g. coordinates for certain test maps and could actually integrate them with standard gurka maps even. anyways, future stuff. so far we're only testing `just_gtfs` actually, no valhalla api yet. but also there isn't any new valhalla api yet.

ideally we'd also make sure we produce valid gtfs data. you could run this tool on the `gtfs_test` dataset: https://github.com/MobilityData/gtfs-validator. seems to be well maintained. how we'll integrate that tool in the future to make sure all our (synthetic) test gtfs data is eventually valid, not sure yet. if we take this current dataset for the time being, let's validate it manually first before moving on.

kevinkreiser:
@chris-jpark lets close this in favor of the other one you opened up

Issue #3629, opened 2022-05-16
GTFS Support: Transcoding to GraphTiles


kevinkreiser:
After we have a mock gtfs feed and our tests prove we can load it with the thirdparty library its time to start moving the gtfs data into the tiled format that Valhalla's data processing expects. When we build the routing data we may optionally supply a root directory within which transit data can be found. If that data is present, the routing tile building process will enumerate it and attach it to the regular street graph.

The data format that this process expects to find our own binary GraphTile format but this format is generated from a protobuf encoded one via the `valhalla_convert_transit` executable. If you are not familiar with protobuf, please familiarize your self with it: https://en.wikipedia.org/wiki/Protocol_Buffers

In brief, we have an existing (though currently semi broken) process for getting transit data into the graph:

1. fetch aggregated transit data from the transitland services json api using the transit_fetcher: https://github.com/valhalla/valhalla/blob/master/src/mjolnir/valhalla_fetch_transit.cc
2. the fetcher tiles this transit data (very similar to gtfs) into the protobuf format defined here: https://github.com/valhalla/valhalla/blob/master/proto/transit.proto (actually transit_fetch.proto but the former is a superset so we can use that)
3. these tiled pbfs are converted into GraphTiles via `valhalla_convert_transit`
4. the tile build process then attaches those tiles to the regular graph if it finds them in the provided directory

As you can see at the end of the day we need GraphTiles of gtfs data. There are two approaches one could take here:

1. rewrite `valhalla_transit_fetcher` to, instead of calling transitland, translate gtfs (via just_gtfs) into protobuf transit tiles and then use the rest of the existing code to load the data
2. rewrite `valhalla_convert_transit` to, instead of converting protobuf transit tiles, create GraphTiles directly from gtfs via just_gtfs

Before starting this task we should look at both avenues and determine the best coarse of action. Once we have decided we can get into the details of how to move the data around.

This task will be complete once:

1. we have a test (we can add it to the previous suite) which, using our faked up gtfs, generates GraphTiles with transit information
2. we have gtest expect/asserts that verify the information is found in the GraphTiles

pranavpandey1998official:
I am in favour of 2.
1. This would give us more flexibility and room to integrate GTFS-RT and GTFS flex in the future
2. with option 1 we will lose some of the GTFS data (I am not sure about the specifics)

kevinkreiser:
[...]

PR #3669, opened 2022-06-29
GTFS: Transcoding to GraphTiles

chris-jpark:
# Issue

Targetting #3629 where we want to format the GTFS data into valhalla's tiles system and create protobuf files to connect them to the graph. Instead of fetching data from a third-party API, the goal is to load them from local GTFS files. 

The first task is to convert the `which_tiles` function that sorts the data spatially by unique tiles.


chris-jpark:
The high level plan for the file is as so: 

The first step is to refactor the priority queue of tile data into a hashmap mapped by stop_ids. 
This allows us to fill out the `stop_pairs` class of the protobuf as the `tileTransitInfo` struct contains most of the information needed for the class. We'll loop through the trips-> stop_times to do that. Will use the function Kevin sent to calculate `dist_traveled` for origin and destination (the attributes from `shapes.txt`), which should be the only information we need that `tileTransitInfo` can't directly fetch. 

I don't think we need to use `tileTransitInfo` to fill the Nodes (stops) and Routes classes as it looks like the same complexity as just accessing the info directly through just_gtfs.

chris-jpark:
I've completed serializing all the classes of the protobuf other than the shape-related objects. 
I've been looking at [the method](https://github.com/valhalla/valhalla/blob/master/src/loki/search.cc#L555-L575) to line up the shape points and their matching stops, but I'm not sure how to convert to PointLL types from Coordinates. 

Is there a way to do this such that I can build an almost replica function, or should I just use those lines of code as reference and build a similar function to do the job with coordinates.

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,777 @@
...
+    } while (filesystem::exists(file_name));
+  }
+}
+} // namespace
+
+namespace valhalla {
+namespace mjolnir {
+
+// Read from GTFS feed, sort data into the tiles they belong to
+std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_tree::ptree& pt) {
```
chris-jpark, line 528: move this to the anonymous namespace

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,777 @@
...
+    write_shapes(tile, current);
+
+    // list of dangling tiles
+    if (dangles) {
+      dangling.emplace_back(current.graphId);
+    }
+
+    if (tile.stop_pairs_size()) {
+      //      auto curr_file_name = GraphTile::FileSuffix(current.graphId, SUFFIX_NON_COMPRESSED);
+      //      filesystem::path curr_transit_tile = pt.get<std::string>("mjolnir.transit_dir") +
```
chris-jpark, line 392: remove this comment

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,777 @@
...
+    node->set_onestop_id(tile_stop.stop_id);
+    stop_graphIds[tile_stopId] = node_id;
+  }
+  return stop_graphIds;
+}
+
+// read feed data per stop, given shape
+float add_stop_pair_shapes(const gtfs::Stop& stop_connect, const gtfs::Shape& trip_shape) {
+  // check which segment would belong to which tile
+  float dist_traveled = 0;
```
chris-jpark, line 145: if dist_traveled already exists in gtfs data  just return it directly

valhalla/midgard/sequence.h
```diff
@@ -156,7 +156,7 @@ template <class T> class mem_map {
   void create(const std::string& new_file_name, size_t new_count, int advice = POSIX_MADV_NORMAL) {
     auto target_size = new_count * sizeof(T);
     struct stat s;
-    if (stat(new_file_name.c_str(), &s) || s.st_size != target_size) {
+    if (stat(new_file_name.c_str(), &s) || (unsigned long)s.st_size != target_size) {
```
chris-jpark, line 159: upcast target size (decltype) @kevinkreiser

test/gurka/test_gtfs.cc
```diff
@@ -146,4 +155,60 @@ TEST(GtfsExample, WriteGtfs) {
   const auto& calendarGTFS = feed_reader.get_calendar();
   EXPECT_EQ(calendarGTFS.size(), 1);
   EXPECT_EQ(calendarGTFS[0].service_id, serviceOneID);
```
chris-jpark, line 157: write some tests about frequency / schedule information

chris-jpark: test for missing days / calendar exceptions

test/gurka/test_gtfs.cc
```diff
@@ -146,4 +155,60 @@ TEST(GtfsExample, WriteGtfs) {
...
+  // it's gonna write protobufs
+  filesystem::recursive_directory_iterator transit_file_itr("test/data/transit_tiles");
+  filesystem::recursive_directory_iterator end_file_itr;
+
+  // for each pbf.
+  for (; transit_file_itr != end_file_itr; ++transit_file_itr) {
+    if (filesystem::is_regular_file(transit_file_itr->path())) {
+      std::string fname = transit_file_itr->path().string();
+      std::mutex lock;
+      mjolnir::Transit transit = mjolnir::read_pbf(fname, lock);
```
chris-jpark, line 191: call the function without mutex

kevinkreiser:
@chris-jpark looks like only very minor things to go on the lint debug build, just a couple clang tidy things const reference basically.

src/CMakeLists.txt
```diff
@@ -223,6 +223,7 @@ target_include_directories(valhalla
     ${VALHALLA_SOURCE_DIR}/third_party/rapidjson/include
     ${VALHALLA_SOURCE_DIR}/third_party/date/include
     ${VALHALLA_SOURCE_DIR}/third_party/cpp-statsd-client/include
+    ${VALHALLA_SOURCE_DIR}/third_party/just_gtfs/include
```
kevinkreiser, line 226: this should be in the private section. i hope we arent actually including this from any of our headers

kevinkreiser: or wait, its already in mjolnir below.. why did we need it here again? i know we needed previously in the tests by why at the library level?

src/mjolnir/valhalla_fetch_transit.cc
```diff
@@ -28,8 +28,8 @@
 
 #include "filesystem.h"
```
kevinkreiser, line 29: it seems like this file should be unchanged right? can you revert it and leave it untouched. once we have raw gtfs working properly we can completely delete it but it should probably live for a while yet, untouched

valhalla/midgard/sequence.h
```diff
@@ -154,8 +154,9 @@ template <class T> class mem_map {
 
   // create a new file to map with a given size
   void create(const std::string& new_file_name, size_t new_count, int advice = POSIX_MADV_NORMAL) {
-    auto target_size = new_count * sizeof(T);
+    decltype(stat::st_size) target_size = new_count * sizeof(T);
     struct stat s;
+
```
kevinkreiser, line 159: please try to avoid whitespace only changes

valhalla/mjolnir/convert_transit.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <boost/property_tree/ptree.hpp>
+#include <unordered_set>
+
+#include <valhalla/baldr/graphid.h>
+
+namespace valhalla {
+namespace mjolnir {
+
```
kevinkreiser, line 10: javadoc please

valhalla/mjolnir/ingest_transit.h
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * @brief Processes the data from 'select_transit_tiles' into GraphTiles, adding the data into the
+ * valhalla graph.
+ *
+ * @param pt Property tree containing the hierarchy configuration
+ *             and other configuration needed to build transit.
+ * @param tiles queue of all tiles that have transit data (tiles that need processing)
+ * @param thread_count
+ * @return std::list<GraphId>
+ */
```
kevinkreiser, line 29: javadoc is wrong

valhalla/mjolnir/ingest_transit.h
```diff
@@ -0,0 +1,99 @@
...
+std::list<baldr::GraphId> ingest_transit(const boost::property_tree::ptree& pt);
```
kevinkreiser, line 30: why is this a list? if it can be a  vector or an unordered_set that is better

valhalla/mjolnir/ingest_transit.h
```diff
@@ -0,0 +1,99 @@
...
+  std::vector<baldr::GraphId> intrastation; // List of intra-station connections
+  std::vector<TransitLine> lines;           // Set of unique route/stop pairs
+};
+
+Transit read_pbf(const std::string& file_name, std::mutex& lock);
+Transit read_pbf(const std::string& file_name);
+
+// Get PBF transit data given a GraphId / tile
+Transit read_pbf(const baldr::GraphId& id, const std::string& transit_dir, std::string& file_name);
+void write_pbf(const Transit& tile, const filesystem::path& transit_tile);
```
kevinkreiser, line 91: we should add javadoc on these

valhalla/mjolnir/ingest_transit.h
```diff
@@ -0,0 +1,99 @@
...
+  uint32_t shapeid;
+  float orig_dist_traveled;
+  float dest_dist_traveled;
+};
+
+struct StopEdges {
+  baldr::GraphId origin_pbf_graphid;        // GraphId (from pbf) of the origin stop
+  std::vector<baldr::GraphId> intrastation; // List of intra-station connections
+  std::vector<TransitLine> lines;           // Set of unique route/stop pairs
+};
```
kevinkreiser, line 84: since we dont pass these around, cant we just move these to the cc file?

src/mjolnir/valhalla_fetch_transit.cc
chris-jpark: I will have to change removing `transitpbf.h` and including `ingest_transit.h` to not get build errors because the first file was removed and the functions were moved to the second file, but I will revert back other changes.

valhalla/mjolnir/ingest_transit.h
chris-jpark: It's because in line 625 -> the list is spliced and copied into the promises list.

valhalla/mjolnir/ingest_transit.h
chris-jpark: moved to convert_transit!

chris-jpark:
## Fixes issue #3629
Removes Transitland dependency when creating Transit Tiles. 
Introduces the ability to read GTFS data directly.
Protobufs are created from the GTFS data which then is converted to Transit Tiles. 
Each process is tested in `test_gtfs.cc`. 
Implementation is done in `mjolnir/ingest_transit.cc`.

src/CMakeLists.txt
chris-jpark: Removed from this CMakeList - must have been a redundant include 😅

valhalla/mjolnir/convert_transit.h
chris-jpark: documentation added!

valhalla/mjolnir/ingest_transit.h
chris-jpark: completed!

src/mjolnir/valhalla_fetch_transit.cc
kevinkreiser: Ah OK yeah changing includes is fine just revert the other diffs

chris-jpark: Yup, reverted all other edits

.circleci/config.yml
```diff
@@ -67,7 +67,7 @@ jobs:
             && cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=On -DCPACK_GENERATOR=DEB \
                 -DENABLE_COMPILER_WARNINGS=On -DENABLE_WERROR=Off -DCMAKE_EXPORT_COMPILE_COMMANDS=On \
                 -DCMAKE_CXX_FLAGS="-fuse-ld=lld" \
-                -DLOGGING_LEVEL=NONE \
+                -DLOGGING_LEVEL=INFO \
```
kevinkreiser, line 70: so a while ago we disabled logging in the debug build because by default it logs way too much. then when we moved the guts from valhalla_convert_transit into the library, the file we added it to was part of -werror and so variables that are only used for logs become errors (since the logs are ifdef'd out from the compilers perspective). we could actually avoid making this change and use the preprocessor to conditionally make those variables over in the source file. @nilsnolde what do you think about that? should we try to keep logging at NONE? i kind of like logging at INFO in case its useful in the build

test/gurka/test_gtfs.cc
```diff
@@ -146,4 +160,94 @@ TEST(GtfsExample, WriteGtfs) {
...
+          transit.nodes(0).onestop_id() == stopOneID || transit.nodes(0).onestop_id() == stopTwoID;
+      EXPECT_EQ(stops_written, true);
+      // routes info
+      EXPECT_EQ(transit.routes_size(), 1);
+      EXPECT_EQ(transit.routes(0).onestop_id(), "2");
+
+      // stop_pair info
+      EXPECT_EQ(transit.stop_pairs_size(), 2);
+      bool stop_pairs_written = transit.stop_pairs(0).origin_onestop_id() == stopOneID ||
+                                transit.stop_pairs(0).origin_onestop_id() == stopTwoID;
```
kevinkreiser, line 233: does the code coverage warning about not hitting line 233 mean that the id is always stopOneID and never stopTwoID?

chris-jpark: Yup, this is expected - stopTwoID should be stored in `transit.stop_paris(1)`. 
However, `transit.stop_pairs(0) = stopTwoID` on my device, which made me realize that depending on threading, it might store the data in a different order. 
But on the same device, the order should be the same.

valhalla/mjolnir/ingest_transit.h
kevinkreiser: yeah i think we dont technically have to splice them together we could just move ranges out of vectors into another vector but honestly it doesnt matter. just ignore this comment :smile:

kevinkreiser (dismissed):
i didnt put a lot of comments on the important parts of the implementation in ingest_transit.cc but also we paired a lot on that so i think its ok. im sure there will be some bugs in there that we uncover as we get closer to full integration. @nilsnolde did you have a minute to look over the changes for anything?

test/gurka/test_gtfs.cc
kevinkreiser: oh jeez for some reason i didnt see the `(0)` i thought that the loop just checked both but i see you check 0 and 1 separately in the loop, sorry about that!

kevinkreiser: actually it should be reproducable, if we set concurrency to 1 in the config (which i think it is..) only one thread will ever be there to pick of stuff to do in order. we dont randomize anything else on purpose do we?

chris-jpark: No, that is all for checking for randomness

nilsnolde:
Sorry for being a bit passive lately @kevinkreiser and @chris-jpark. I’ll have a more thorough check tmrw morning. And thanks for all the good work so far!

.circleci/config.yml
nilsnolde: > for some reason unused variables are treated as errors even if its not part of sources_with_warnings

was it using the ccache files then for years I guess?

> avoid making this change and use the preprocessor

yep, if the impact of INFO logging is so huge on debug builds then let's do that. hope it's not too many other long untouched files needing that, but we'll see:)

can you add that to the PR @chris-jpark ?

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,778 @@
...
+  auto node_id = tile_info.graphId;
+  feedCache stopFeeds;
+
+  std::unordered_map<gtfs::Id, GraphId> stop_graphIds;
+  // loop through all stops inside the tile
+  for (const feedObject& feed_stop : tile_stopIds) {
+    const auto& tile_stopId = feed_stop.id;
+    const auto& feed = stopFeeds(feed_stop);
+    auto* node = tile.add_nodes();
+    gtfs::Stop tile_stop = *(feed.get_stop(tile_stopId));
```
nilsnolde, line 185: there's no way this doesn't return a valid stop? think I'd still protect for it, not too much experience with GTFS real data but I'm sure they can be very messed up.

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,778 @@
...
+    node->set_lon(tile_stop.stop_lon);
+    node->set_lat(tile_stop.stop_lat);
+    node->set_type((int)tile_stop.location_type);
```
nilsnolde, line 188: nit: usually this project uses c++ casts, e.g. `std::static_cast<int>()`

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,778 @@
...
+  }
+  return dist_traveled;
+}
+
+struct unique_transit_t {
+  std::mutex lock;
+  std::unordered_map<std::string, size_t> trips;
+  std::unordered_map<std::string, size_t> block_ids;
+  std::unordered_map<std::string, size_t> lines;
+};
```
nilsnolde, line 238: nit: can we move this to the top of the file? I do like it better when I know structs are always declared at the top of the file

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,778 @@
...
+  const auto& tile_tripIds = tile_info.tile_trips;
+  feedCache tripFeeds;
+  bool dangles = false;
+
+  for (const feedObject& feed_trip : tile_tripIds) {
+    const auto& tile_tripId = feed_trip.id;
+    const std::string currFeedPath = feed_trip.feed;
+    const auto& feed = tripFeeds(feed_trip);
+
+    gtfs::Trip currTrip = *(feed.get_trip(tile_tripId));
```
nilsnolde, line 255: same here: can't be an invalid id?

nilsnolde: I won't leave this comment with all occurrences but there's a lot more. in case you thought about it while developing and decided it's fine, never mind:)

src/mjolnir/valhalla_ingest_transit.cc
```diff
@@ -0,0 +1,22 @@
+#include "baldr/rapidjson_utils.h"
+#include "mjolnir/ingest_transit.h"
+
+int main(int argc, char** argv) {
+  if (argc < 2) {
+    std::cerr << "Usage: " << std::string(argv[0]) << " Error" << std::endl;
+    return 1;
+  }
```
nilsnolde, line 8: right now I'd have no idea how this exe works before I look into the source, on error we don't print any usage. can do it in a follow-up PR as well, but we really shouldn't leave it like this.. have a look here for copy/paste (almost):
https://github.com/valhalla/valhalla/blob/gtfs-fetch-transit/src/valhalla_export_edges.cc/#L143-L185

src/mjolnir/ingest_transit.cc
chris-jpark: In `select_transit_tiles` I loop through the stop data from the feed, so all the stops should exist, but just in case the feed doesn't include stop_id, I just won't process the stop if it doesn't have an id.

src/mjolnir/ingest_transit.cc
chris-jpark: 👍

src/mjolnir/ingest_transit.cc
chris-jpark: I'll be doing the same as above -> if there exists no id for the object (whether it be route, trip, etc...), I will filter it out so all occurrences of search by id have a valid result.

src/mjolnir/valhalla_ingest_transit.cc
kevinkreiser: yeah we just copied this from fetch transit  which doesnt use nice args its standard old c style stuff

.circleci/config.yml
kevinkreiser: i think the impact of INFO isnt huge. and i just looked and release is using TRACE which is crazy (but makes sure to test all code paths). unless you are totally opposed i would keep INFO

src/mjolnir/valhalla_convert_transit.cc
```diff
@@ -1,1281 +1,7 @@
...
-             " midnight departures were added: " + std::to_string(percent) + "% increase.");
-  }
-
-  auto t2 = std::chrono::high_resolution_clock::now();
-  uint32_t secs = std::chrono::duration_cast<std::chrono::seconds>(t2 - t1).count();
-  LOG_INFO("Finished building transit network - took " + std::to_string(secs) + " secs");
-}
-
 int main(int argc, char** argv) {
   if (argc < 2) {
```
kevinkreiser, line 6: if we did the other executable with cxxopts we should probably do this one too. thankfully there are only what 2 arguments to parse, the config and a test file. should be quick!

src/mjolnir/convert_transit.cc
```diff
@@ -0,0 +1,1345 @@
+#include <cmath>
```
kevinkreiser, line 1: nothing is changed here really this is just moved from the executable and the include paths updated. we might have changed the signature slightly to make it use the config only to do its work. the cool thing here is that now that its in the library we will be testing it and test coverage will inevitably go up. what i can reason about though is why it goes down in this pr. the executable was untested before so i assume it was also a coverage hit before..

src/mjolnir/ingest_transit.cc
kevinkreiser: yeah i do think its impossible for this to get screwed up. the feedObject is populated by looping over all the stops in the feed as chris is saying. this is just going back on a second pass and pu lling out the rest of the info for the stop (once we know all the stops that will be in a tile)

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,790 @@
...
+    const auto& tile_stopId = feed_stop.id;
+    const auto& feed = stopFeeds(feed_stop);
+    auto* node = tile.add_nodes();
+    gtfs::Stop tile_stop = *(feed.get_stop(tile_stopId));
+    node->set_lon(tile_stop.stop_lon);
+    node->set_lat(tile_stop.stop_lat);
+    node->set_type(static_cast<int>(tile_stop.location_type));
+    node->set_graphid(node_id);
+    node_id++;
+    // node->set_prev_graphid(); what is prev?
```
kevinkreiser, line 224: used in convert_transit i believe

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,790 @@
...
+    node->set_name(tile_stop.stop_name);
+    // node->set_osm_way_id(); where is this?
```
kevinkreiser, line 226: this will be our next problem. how to connect the stop to the regular graph :smile:  

the transitland integration works by  having them call valhalla/locate to get near by edges in the regular graph and then storing the most interesting one's way id on the transitland side. then when valhalla ingests transitlands data it can say oh yeah you think it should go to this way id, let me find edges with that way id and i'll connect it to that.

instead we will have to do what @xlqian did with bike share station integration and connect it ourselves. i think his work over there was execllent so i'm expecting better results and possibly the ability to reuse that code

src/mjolnir/ingest_transit.cc
```diff
@@ -0,0 +1,790 @@
...
+    node->set_timezone(tile_stop.stop_timezone);
+    // 0 is No Info ; 1 is True ; 2 is False
+    bool wheelchair_accessible = (tile_stop.wheelchair_boarding == "1");
+    node->set_wheelchair_boarding(wheelchair_accessible);
+    // node->set_generated(feed_stop.generated); ?
```
kevinkreiser, line 231: i this is a value add from transitland: https://www.transit.land/documentation/datastore/stops#generated-platforms-and-egresses

basically if there is a complex station you want to know the entry and exit points of it to connect it to the road network. some feeds dont have this information and indeed some stations/stops are trivial (bus stop along the road for example). in those cases one can generate an in/egress (if the downstream consumer, valhalla) expects there to always be in/egress information. what we can do is look in convert transit to see how we actually use it as i dont remember off the top of my head. not important for now tbh

src/mjolnir/ingest_transit.cc
kevinkreiser: i forget if this one is referenced from another item  (like route for example) or if we got it by looping over trips. if we got it by looping over the thing directly we dont need to check if its not there now but if its derived from another thing (ie joined from another table) then yeah we s hould check it.  any place that you skip something you should definitely put a LOG_WARN about what feed and what thing and what id was getting skipped

src/mjolnir/ingest_transit.cc
kevinkreiser: also can you mark anything similar to this where we arent sure yet or havent implemented it with a `// TODO:` so its easier to search for in the code

chris-jpark:
Pushed a commit testing for stitching as well, had to change a couple of lines on tile level for it to work properly.
Also added cxxopts for valhalla_convert_transit.
A little unsure about positional arguments inside cxxopts, but we'll see how it works out next PR as most of the changes for convert_transit will be happening on the next PR.
the commit also includes TODO comments wherever necessary as you suggested.

src/mjolnir/valhalla_ingest_transit.cc
```diff
@@ -0,0 +1,69 @@
...
+    cxxopts::Options options(
+      "valhalla_ingest_transit",
+      "valhalla_ingest_transit " VALHALLA_VERSION "\n\n"
+      "valhalla_ingest_transit is a program that reads GTFS data. It converts a directory of transit feeds into protobuf tiles."
+      "\n\n");
+
+    options.add_options()
+      ("h,help", "Print this help message.")
+      ("v,version", "Print the version of this software.")
+      ("c,config", "Path to the json configuration file.", cxxopts::value<std::string>());
```
nilsnolde, line 21: it's good to have it not positional, more in line with most of valhalla's executables

chris-jpark: Is it possible to implement non-positional without flags, as the command line for `valhalla_convert_transit` is `valhalla_config [target_directory] [test_file]`. `valhalla_ingest_transit` looks already non positional?

nilsnolde: we can leave `valhalla_convert_transit` for the next PR IMO.

we can then take the opportunity to change that a bit and make it more consistent with the other exes (`config` as optional, even though required, and `test_file` as positional). to clarify, every argument that's not named is called "positional", so the previous behavior was 2 positional arguments. 

`implement non-positional without flags`, not sure what that means :sweat_smile:

src/mjolnir/valhalla_convert_transit.cc
```diff
@@ -1,35 +1,89 @@
...
+      "valhalla_convert_transit " VALHALLA_VERSION "\n\n"
+      "valhalla_convert_transit is a program that reads protobuf files and creates Level 3 Transit Tiles."
+      "\n\n");
+
+    options.add_options()
+      ("h,help", "Print this help message.")
+      ("v,version", "Print the version of this software.")
+      ("c,config", "Path to the json configuration file.", cxxopts::value<std::string>())
+      ("target_directory", "Path to write transit tiles", cxxopts::value<std::string>())
+      ("test_file", "file where tests are written", cxxopts::value<std::string>());
```
kevinkreiser, line 27: in the next part we can change to get target from config and the test file can be a flag rather than positional

kevinkreiser (approved):

kevinkreiser:
fixes https://github.com/valhalla/valhalla/issues/3629

PR #3670, opened 2022-06-30
mark edge segments with interpolated match points

kevinkreiser:
this is an alternative approach to @nilsnolde 's work on #3646 

in that pr we essentially called cut segments for every match point, which, while that probably isnt too bad, is technically a worse run time because of the nested looping and allocations as nils pointed out.

in this pr i took a slightly different approach. i refactored cut segments in the following way. previously it would simply split any edge segments where a breakpoint landed as well as the last segment between a pair of states in the hmm. now, while it does that it also marks those segments with the match point index that lies on the given segment even if we arent going to split that segment at the match point (ie even if its not a break). the runtime is unchanged because we still loop over all the segments and all the match points in tandem. the only thing that will cost run time are the two `if`s inside of the check where we avoid splitting the segment.

another option would have been to move this logic up further in the call stack to be inside of the `MergeRoute` function. while this seems like a better place for it (because we would create the segments properly in the first place) this required a lot more surgery because that function doesnt currently loop over the match points. ultimately though this is definitely where the code should be, ill make a comment in the code to that effect

### DIFF
--- a/scripts/valhalla_build_timezones
+++ b/scripts/valhalla_build_timezones
@@ -2,7 +2,7 @@
 
 error_exit() {
   echo "error: $1" 1>&2
-  exit $(pkg-config geos --modversion | grep -cvF "3.9")
+  #exit $(pkg-config geos --modversion | grep -cvF "3.9")
 }
 
 if  ! which spatialite >/dev/null; then
--- a/src/meili/match_route.cc
+++ b/src/meili/match_route.cc
@@ -122,25 +122,21 @@ bool MergeRoute(const State& source,
  * @param match_results  The matched points representing the matched input trace points
  * @param first_idx      The index of the first match in the range
  * @param last_idx       The index of the second match in the range
- * @param segments       The segments over the range
+ * @param segments_begin The current first segment for this range
+ * @param segments_end   The end of segments iterator
  * @param new_segments   The new vector of segments over the range (should be the same or more
  * segments). This is a reference parameter to avoid constant allocation
  */
 void cut_segments(const std::vector<MatchResult>& match_results,
                   int first_idx,
                   int last_idx,
-                  const std::vector<EdgeSegment>& segments,
+                  std::vector<EdgeSegment>::iterator& first_segment,
+                  const std::vector<EdgeSegment>::iterator& segments_end,
                   std::vector<EdgeSegment>& new_segments) {
-  auto first_segment = segments.begin();
   int prev_idx = first_idx;
   for (int curr_idx = first_idx + 1; curr_idx <= last_idx; ++curr_idx) {
     const MatchResult& curr_match = match_results[curr_idx];
     const MatchResult& prev_match = match_results[prev_idx];
-
-    // skip if we do not need to cut
-    if (!curr_match.is_break_point && curr_idx != last_idx) {
-      continue;
-    }
 
     // Allow for some fp-fuzz in this gt comparison
     bool prev_gt_curr = prev_match.distance_along > curr_match.distance_along + 1e-3;
@@ -149,12 +145,12 @@ void cut_segments(const std::vector<MatchResult>& match_results,
     bool loop = (prev_match.edgeid == curr_match.edgeid) && prev_gt_curr;
 
     // if it is a loop, we start the search after the first edge
-    auto last_segment = std::find_if(first_segment + static_cast<size_t>(loop), segments.end(),
+    auto last_segment = std::find_if(first_segment + static_cast<size_t>(loop), segments_end,
                                      [&curr_match](const EdgeSegment& segment) {
                                        return (segment.edgeid == curr_match.edgeid);
                                      });
 
-    if (last_segment == segments.cend()) {
+    if (last_segment == segments_end) {
       throw std::logic_error("In meili::cutsegments(), unexpectedly unable to locate target edge.");
     }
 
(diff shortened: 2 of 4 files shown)


src/meili/match_route.cc
```diff
@@ -129,38 +134,47 @@ bool MergeRoute(const State& source,
...
+        break;
+    }
 
-    if (last_segment == segments.cend()) {
+    if (search_segment == segments.cend()) {
       throw std::logic_error("In meili::cutsegments(), unexpectedly unable to locate target edge.");
     }
 
+    // if its not a break point nor the last segment, then we dont need to split it into two
+    if (!curr_match.is_break_point && curr_idx != last_idx) {
```
kevinkreiser, line 164: so the main trick is we moved this continue down past where we did the loop to find the segment upon which the current match point lies. this way we have access to it, even if we arent going to cut it below. because i keep the `search_segment` around for each iteration of the main loop we are gauranteed that the inner loop does no more iterations than it otherwise would have and so this change shouldnt really cost us anything

src/meili/match_route.cc
```diff
@@ -129,38 +134,47 @@ bool MergeRoute(const State& source,
...
+      // but if it isnt a breakpoint and its not the last stateful match point then its interpolated
+      // and we still need to mark the segments in such a way that we which match points lie on them
+      if (search_segment->first_match_idx < 0)
+        search_segment->first_match_idx = curr_idx;
+      if (search_segment->last_match_idx < 0)
+        search_segment->last_match_idx = curr_idx;
```
kevinkreiser, line 170: i'll admit that i didnt think too hard about how correct this is but rather did the conservative thing, that is only set the index if unset. its possible that i should just always set both of them because they will be corrected below and wont have been set up to this point (ie we can probably remove the `if`s) but i didnt want to have to prove that to myself at 10pm :smile:

scripts/valhalla_build_timezones
```diff
@@ -2,7 +2,7 @@
 
 error_exit() {
   echo "error: $1" 1>&2
-  exit $(pkg-config geos --modversion | grep -cvF "3.9")
+  #exit $(pkg-config geos --modversion | grep -cvF "3.9")
```
kevinkreiser, line 5: yeah so we have a bug on geos 3.9*. and in a previous pr we patched it without really checking it. the best thing that we can do here is to not exit if its 3.9. ill change this script to do that. @nilsnolde are you cool if this comes along for the ride? im kind of sick of having to deal with it and have been too lazy to pr it seprately

nilsnolde:
that's great @kevinkreiser , thanks! much easier to understand than my PR. hope we won't have to touch for a while now :smile:

scripts/valhalla_build_timezones
nilsnolde: sure!

nilsnolde (dismissed):

nilsnolde (dismissed):

nilsnolde:
huh.. can't set you as a reviewer anymore @kevinkreiser, github weirdo again. anyways, you can let me know if the tests are ok for you like this.

kevinkreiser:
Haha yeah i opened the pr so you have to be reviewer

nilsnolde:
makes sense :D I gratefully approve :)

kevinkreiser:
just had to update the changelog and will merge when the checks are green

nilsnolde (approved):

ThomasVall:
Hi,

We experienced another edge_index problem even with your fix, we have some values like : "18446744073709551615" with version 3.3.0 of Valhalla.
`{
      "lon": 1.140924,
      "lat": 49.41262,
      "type": "interpolated",
      "edge_index": 7,
      "distance_along_edge": 0.959356,
      "distance_from_trace_point": 3.243988
    },
    {
      "lon": 1.14093,
      "lat": 49.412622,
      "type": "matched",
      "edge_index": 7,
      "distance_along_edge": 0.963251,
      "distance_from_trace_point": 3.356317
    },
    {
      "lon": 1.140932,
      "lat": 49.412623,
      "type": "interpolated",
      "edge_index": 7,
      "distance_along_edge": 0.964807,
      "distance_from_trace_point": 3.225371
    },
    {
      "lon": 1.140932,
      "lat": 49.412623,
      "type": "interpolated",
      "edge_index": 18446744073709551615,
      "distance_along_edge": 0.964807,
      "distance_from_trace_point": 3.005524
    },
    {
      "lon": 1.140932,
      "lat": 49.412623,
      "type": "interpolated",
      "edge_index": 18446744073709551615,
      "distance_along_edge": 0.964807,
      "distance_from_trace_point": 3.005524
    },
    {
      "lon": 1.140932,
      "lat": 49.412623,
      "type": "interpolated",
      "edge_index": 18446744073709551615,
      "distance_along_edge": 0.964807,
      "distance_from_trace_point": 3.005524
    },
    {
      "lon": 1.140945,
      "lat": 49.412629,
      "type": "interpolated",
      "edge_index": 18446744073709551615,
      "distance_along_edge": 0.973363,
      "distance_from_trace_point": 1.62578
    },
    {
      "lon": 1.141003,
      "lat": 49.412627,
      "type": "interpolated",
      "edge_index": 8,
      "distance_along_edge": 0.06591,
      "distance_from_trace_point": 2.697996
    },
    {
      "lon": 1.141034,
      "lat": 49.412595,
      "type": "matched",
      "edge_index": 8,
      "distance_along_edge": 0.184857,
      "distance_from_trace_point": 2.532955
    },
    {
      "lon": 1.141064,
      "lat": 49.412565,
      "type": "interpolated",
      "edge_index": 8,
      "distance_along_edge": 0.29661,
      "distance_from_trace_point": 2.773829
    },
    {
      "lon": 1.141101,
      "lat": 49.412526,
      "type": "interpolated",
      "edge_index": 8,
      "distance_along_edge": 0.436593,
      "distance_from_trace_point": 2.798364
    },`

Did we miss something or is it another bug ?

Regards,
Thomas

nilsnolde:
Yes: https://github.com/valhalla/valhalla/issues/3699

It's smth to do with duplicate waypoints (as in your example, better to sanitize it), but maybe not only. It's part of the bug scrub project, but has very little priority atm. Which can change of course once someone decides to fund a fix:)

ThomasVall:
Thanks @nilsnolde for you quick reply, I edited my example with no duplicate but I still get the same edge_index bug. Is there a workaround for now, or must I still use an older version of Valhalla ?

kevinkreiser:
@ThomasVall regardless of whether this looks like a new but similar bug to some of the others or if its a duplicate. it would be good for you to post this info in the issue so that it doesnt get lost. a reproducible case is always the first step in fixing these things. as @nilsnolde said its hard to find time to work on this bug scrub at the moment but when we do having the test case right there in the issue is very helpful.

thanks for reporting!

PR #3672, opened 2022-07-04
FIX(android):Fix for build android.

SongChiYoung:
When build android, `unused parameter` error.
So, Fix it with temp value

# Issue
when build with android NDK

`valhalla/src/midgard/logging.cc:137:67: error: unused parameter 'custom_directive' [-Werror,-Wunused-parameter]
  virtual void Log(const std::string& message, const std::string& custom_directive = " [TRACE] ") {
                                                                  ^
valhalla/src/midgard/logging.cc:166:67: error: unused parameter 'custom_directive' [-Werror,-Wunused-parameter]
  virtual void Log(const std::string& message, const std::string& custom_directive = " [TRACE] ") {
                                                                  ^`

It's from that code 

```
virtual void Log(const std::string& message, const std::string& custom_directive = " [TRACE] ") {
 #ifdef __ANDROID__
     __android_log_print(ANDROID_LOG_INFO, "valhalla", "%s", message.c_str());
 #else
```

cause, when android, does not use `custom_directive` so add to temp using hack. 

like it 

`
std::string tmp = custom_directive; //for prevent unused error
`

now we can build it.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/midgard/logging.cc
+++ b/src/midgard/logging.cc
@@ -136,6 +136,7 @@ public:
   }
   virtual void Log(const std::string& message, const std::string& custom_directive = " [TRACE] ") {
 #ifdef __ANDROID__
+    std::string tmp = custom_directive; // to prevent -Wunused-parameter
     __android_log_print(ANDROID_LOG_INFO, "valhalla", "%s", message.c_str());
 #else
     std::string output;
@@ -165,6 +166,7 @@ class StdErrLogger : public StdOutLogger {
   using StdOutLogger::StdOutLogger;
   virtual void Log(const std::string& message, const std::string& custom_directive = " [TRACE] ") {
 #ifdef __ANDROID__
+    std::string tmp = custom_directive; // to prevent -Wunused-parameter
     __android_log_print(ANDROID_LOG_ERROR, "valhalla", "%s", message.c_str());
 #else
     std::string output;


src/midgard/logging.cc
```diff
@@ -165,6 +166,7 @@ class StdErrLogger : public StdOutLogger {
   using StdOutLogger::StdOutLogger;
   virtual void Log(const std::string& message, const std::string& custom_directive = " [TRACE] ") {
 #ifdef __ANDROID__
+    std::string tmp = custom_directive; //for prevent unused error
```
nilsnolde, line 169:
```suggestion
    std::string tmp = custom_directive;  // to prevent -Wunused-parameter
```

nilsnolde (changes requested):
IMO fair enough, just some grammar corrections. also, you'd need to run `./scripts/format.sh` pls.

fwiw, probably you'd also get away with the option `-DENABLE_SINGLE_FILES_WERROR=OFF`

SongChiYoung:
@nilsnolde 
I fixed format correctly.
And I think, without error option is bad choice of solving when we want build the nice project.

Thx to review it.

src/midgard/logging.cc
SongChiYoung: I was fix it and squash my commits.

nilsnolde (approved):
thanks! no changelog needed IMHO

src/midgard/logging.cc
```diff
@@ -136,6 +136,7 @@ class StdOutLogger : public Logger {
   }
   virtual void Log(const std::string& message, const std::string& custom_directive = " [TRACE] ") {
 #ifdef __ANDROID__
+    std::string tmp = custom_directive; // to prevent -Wunused-parameter
     __android_log_print(ANDROID_LOG_INFO, "valhalla", "%s", message.c_str());
```
kevinkreiser, line 140: i was going to suggest maybe actually using the parameter, seems kind of lame to add a no-op to the code just to shut up -Werror. we could add the custom directive to the error message in the android version:

```suggestion
    __android_log_print(ANDROID_LOG_INFO, "valhalla", "%s %s", custome_directive.c_str(), message.c_str());
```

SongChiYoung: Actually, I wanted to do that too.
But in my opinion, it's not a familiar log format for Android (because there is a TAG method).

therefore,
I wanted to respect the intent of the original author of the Android log output part.

Even since I am new to the valhalla community, I wanted to respect the intentions of others.

If a lot of people agree, I agree with the amendment.

kevinkreiser: I dont feel strongly about it. We can merge what you have

SongChiYoung: I just confirmed that you are the original author, and I'm confused. haha.

I'm glad that someone else doesn't have the same troubles as me on Android builds.

Issue #3685, opened 2022-07-28
Conan error when building Docker image.


wieringen:
This morning, I tried to build the Valhalla docker image, as I have been doing for the last couple of days. However, this time it failed to build. I tried again with a freshly cloned repo, but the issue remained:

```
#14 1.005 -- No compatible boost version detected, using conan...
#14 1.038 -- Conan: checking conan executable
#14 1.039 -- Conan: Found program /usr/local/bin/conan
#14 1.387 -- Conan: Version found Conan version 1.43.0
#14 1.387 -- Conan executing: /usr/local/bin/conan install /usr/local/src/valhalla/conanfile.txt --remote conancenter --settings build_type=RelWithDebInfo --settings compiler=gcc --settings compiler.version=9 --settings compiler.libcxx=libstdc++11 --settings compiler.cppstd=14
#14 2.116 WARN: Remotes registry file missing, creating default one in /root/.conan/remotes.json
#14 2.116 ERROR: boost/1.71.0: Cannot load recipe.
#14 2.116 Error loading conanfile at '/root/.conan/data/boost/1.71.0/_/_/export/conanfile.py': Unable to load conanfile in /root/.conan/data/boost/1.71.0/_/_/export/conanfile.py
#14 2.116   File "/usr/lib/python3.8/imp.py", line 171, in load_source
#14 2.116     module = _load(spec)
#14 2.116   File "<frozen importlib._bootstrap>", line 702, in _load
#14 2.116   File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
#14 2.116   File "<frozen importlib._bootstrap_external>", line 848, in exec_module
#14 2.116   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
#14 2.116   File "/root/.conan/data/boost/1.71.0/_/_/export/conanfile.py", line 1, in <module>
#14 2.116     from conan.tools.build import cross_building
#14 2.116 ImportError: cannot import name 'cross_building' from 'conan.tools.build' (/usr/local/lib/python3.8/dist-packages/conan/tools/build/__init__.py)
#14 2.116
#14 2.154 CMake Error at cmake/conan.cmake:638 (message):
#14 2.154   Conan install failed='1'
#14 2.154 Call Stack (most recent call first):
#14 2.154   CMakeLists.txt:153 (conan_cmake_install)
#14 2.154
#14 2.154
#14 2.154 -- Configuring incomplete, errors occurred!
[...]

PR #3690, opened 2022-07-30
remove dockerfile-build and install deps directly 

nilsnolde:
ref https://github.com/valhalla/valhalla/issues/3685#issuecomment-1199818890

simplifies to only one Dockerfile and removes the `docker` directory. Dockerfile-build was merged with Dockerfile-run and circleci jobs install their own deps now.


.circleci/config.yml
```diff
@@ -13,42 +13,33 @@ commands:
...
+  install_linux_dependencies:
+    steps:
+      - run: bash ./scripts/install-linux-deps.sh
 
 jobs:
   build-docker-images:
-    docker:
-      - image: docker:18.06.3-ce-git
+    machine:
+      image: ubuntu-2004:2022.07.1
```
nilsnolde, line 23: we can use the circleci provided image so it doesn't download: circleci.com/developer/machine/image/ubuntu-2204#how-this-image-works

.circleci/config.yml
```diff
@@ -13,42 +13,33 @@ commands:
...
     steps:
-      - setup_remote_docker
       - checkout
       - run:
-          name: Build Valhalla Docker Images
+          name: Build and push Valhalla Docker image
```
nilsnolde, line 27: do it all in one go

.circleci/config.yml
```diff
@@ -13,43 +13,35 @@ commands:
...
-    docker:
-      - image: valhalla/valhalla:build-latest
+    machine:
+      image: ubuntu-2004:2022.07.1
     resource_class: xlarge
     steps:
       - checkout
+      - install_linux_dependencies
       - run: |
+            python3 -m pip install requests
```
nilsnolde, line 44: confusingly, circleci doesn't use the same python version as installed by our script.. makes no sense and is very annoying

scripts/needs_ci_run
```diff
@@ -128,6 +128,9 @@ for changed_file in changed_files:
         if fnmatch.fnmatch(changed_file, ignore_rule):
             ignored_files.append(changed_file)
 
+print(f"changed files: {changed_files}")
+print(f"ignored files: {ignored_files}")
```
nilsnolde, line 132: I actually didn't know this was still active.. I haven't ever seen it skip anything, I wanna see what it thinks are changed & ignored files

nilsnolde:
stupid circleci has serious python3 issues, I'll look at it later

scripts/install-linux-deps.sh
```diff
@@ -56,4 +56,4 @@ env DEBIAN_FRONTEND=noninteractive apt-get install --yes --quiet \
   && rm -rf /var/lib/apt/lists/*
 
 # for boost
-pip install conan
+python3 -m pip install conan
```
ImreSamu, line 59: imho: more secure with `--upgrade` 
* `python3 -m pip install --upgrade conan`

so we can remove the same conan upgrade lines from the new "Dockerfile"  ( added by me )

Dockerfile
```diff
@@ -1,11 +1,15 @@
...
 
-# get the code into the right place and prepare to build it
+# set paths
+ENV PATH /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
+ENV LD_LIBRARY_PATH /usr/local/lib:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib32:/usr/lib32
+
+# get the code into the right place, install deps and prepare to build it
 WORKDIR /usr/local/src/valhalla
 ADD . /usr/local/src/valhalla
+RUN bash ./scripts/install-linux-deps.sh
```
ImreSamu, line 12: imho: for the optimal docker layer caching 
the order of these two lines should be reversed.

proposal:
```
RUN bash ./scripts/install-linux-deps.sh
ADD . /usr/local/src/valhalla
```
and please add a [`.dockerignore` file](https://docs.docker.com/engine/reference/builder/#dockerignore-file)   :smile: 
   * ignoring a `./build` directory - makes the local docker image testing and building faster.

nilsnolde: > the order of these two lines should be reversed.

jep that's true (just gotta add the script before using it).

I'll add a .dockerignore as well. I never build this image myself, but it hurts if you do

ellenhp:
I haven't thoroughly reviewed but I like this a lot in principle. I'm doing some work on the docker image as well to get a native arm64 image and ended up doing many of these same things independently. It'll be a pain to rebase my changes but I think it'll be worth it :)

.circleci/config.yml
```diff
@@ -12,44 +12,35 @@ commands:
...
-            fi
-      - run:
-          name: Push Valhalla Docker Images
-          command: |
-            if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
-              docker push valhalla/valhalla:build-latest
-              docker push valhalla/valhalla:run-latest
-            elif [[ ! -z "${CIRCLE_TAG}" ]]; then
+              BUILD_TAG="${CIRCLE_TAG:-latest}"
```
nilsnolde, line 31: now we'd only build docker images on master and releases, but IMO that's fine, the other CI builds run on very similar systems

.dockerignore
```diff
@@ -0,0 +1,69 @@
+# same as .gitignore
```
nilsnolde, line 1: I had this on my laptop since a while apparently, it's just a copy of .gitignore. could be overhauled I guess, but meeh

Dockerfile
```diff
@@ -1,19 +1,22 @@
 ####################################################################
-# base image contains all the dependencies we need to build the code
-FROM valhalla/valhalla:build-3.1.5 as builder
+FROM ubuntu:20.04 as builder 
 MAINTAINER Kevin Kreiser <kevinkreiser@gmail.com>
 
-# get the code into the right place and prepare to build it
+# set paths
+ENV PATH /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
+ENV LD_LIBRARY_PATH /usr/local/lib:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib32:/usr/lib32
```
nilsnolde, line 7: never tried without those extra paths. they don't seem needed anymore on 20.04?

Dockerfile
```diff
@@ -1,19 +1,22 @@
...
+
+# install deps
 WORKDIR /usr/local/src/valhalla
+COPY ./scripts/install-linux-deps.sh /usr/local/src/valhalla/scripts
+RUN bash ./scripts/install-linux-deps.sh
```
nilsnolde, line 12: first copy and execute the deps script for better layer caching as @ImreSamu pointed out. this is effectively doing a similar job as having a separate build docker image, esp if (hopefully soon) remote layer caching is standard by default

scripts/needs_ci_run
```diff
@@ -1,12 +1,11 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3
```
nilsnolde, line 1: hmpf, took me longer than it should've to understand what was happening.. in circleci we have to use `sudo` to install the dependencies (also python3-requests). it uses `/usr/bin/python3` for that. this script is being run without `sudo` and references `/opt/circleci/.pyenv/versions/3.10.5/bin/python3` inside circleci's virtual env and apparently they didn't bother to include `/usr/lib/python3.10/site-packages` in the PYTHONPATH, so importing `requests` fails.. 

since it only runs in ci I think it's fair to use this line instead

.circleci/config.yml
```diff
@@ -12,44 +12,36 @@ commands:
   install_macos_dependencies:
     steps:
       - run: brew install protobuf cmake ccache libtool libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
-      - run: pip3 install requests shapely conan
+      - run: /usr/bin/python3 -m pip install --user requests
+      - run: pip3 install shapely conan
```
nilsnolde, line 16: `requests` is used by the `needs_ci_run` script which runs with `/usr/bin/python3`.

the other two libraries are used by cmake & valhalla and are run with `python3` which is apparently different from `/usr/bin/python3` so we need to install with that one as before

nilsnolde:
what a pain python can be.. this should be ready now.. to summarize:
- this removes the docker `build-*` images and instead we use a normal ubuntu ci image and install the dependencies on each job, which is only takes 40 secs per job
- we don't build any docker image anymore on PRs, only on `master` pushes and releases. as mentioned above, it's fine IMO bcs the circleci and docker base image are ubuntu 20.04. building the build image took around 2 minutes, so it's a net zero on ci time pretty much
- plain `Dockerfile`, removed `docker` dir
- moved `install_linux_deps.sh` script to `scripts`

.circleci/config.yml
```diff
@@ -12,44 +12,36 @@ commands:
...
+  install_linux_dependencies:
+    steps:
+      - run: sudo bash ./scripts/install-linux-deps.sh
 
 jobs:
   build-docker-images:
-    docker:
-      - image: docker:18.06.3-ce-git
+    machine:
+      image: ubuntu-2004:current
```
nilsnolde, line 25: this should have a current docker version installed: https://circleci.com/developer/machine/image/ubuntu-2004#how-this-image-works. also `build-essentials` and others, so installing deps will take less time

Dockerfile
```diff
@@ -1,19 +1,22 @@
...
+# get the code into the right place and prepare to build it
 ADD . /usr/local/src/valhalla
 RUN ls
 RUN git submodule sync && git submodule update --init --recursive
 RUN rm -rf build && mkdir build
 
-# upgrade Conan:
-# https://github.com/valhalla/valhalla/issues/3685#issuecomment-1198604174
-RUN pip install --upgrade conan
-
```
nilsnolde, line 16: on the other hand @ImreSamu : if we remove this line, we wouldn't actually prevent issue #3685 for docker builds. as long as the installation script doesn't change, locally cached layers will be used. I think it's best to keep it

ImreSamu: > locally cached layers will be used.

the ["old" documentation](https://github.com/valhalla/valhalla/blob/3ecf1c7d87854e9317113c1f67504f95a7b4ddc8/docker/README.md#to-buildpublish-images-manually
)  contains the `--no-cache` suggestion, probably we should also add the `--pull`
*  `docker build --pull --no-cache ...`
   *  `--no-cache`  "Do not use cache when building the image"
   *  `--pull` : "Always attempt to pull a newer version of the image"

probably the circle-CI `build-docker-images` job should contain these 2 extra safety parameters.
* now: `docker build --tag valhalla/valhalla:run-${BUILD_TAG} .`
* proposed: `docker build  --pull --no-cache  --tag valhalla/valhalla:run-${BUILD_TAG} .`

nilsnolde: I'm not so concerned about circleci builds actually, more about user builds bcs we don't cache docker in ci, but other machines do. though we could pull the newest base image, true, I'll add that.

.circleci/config.yml
nilsnolde: switch to `current` ubuntu tag

kevinkreiser (approved):
too bad ci can't conditionally run the docker stuff, that it has to still run and pretend like it did something

PR #3694, opened 2022-08-04
Blind user mode.

### LINKED ISSUE #2989: Blind users costing model
Many people with blindness are using navigational mobile apps specially designed for them, providing area or route informations as text read by TTS, or screen-reading software installed in mobile device, instead of map rendered visually on the screen.
Would it be possible to introduce "blind users" costing model in Valhalla, which would make it even more perfect as a routing engine for this group of end-users?
"Blind users" costing model could be basically the pedestrian model, with some extensions:
1. Announce in routing instructions, streets crossed by the user with the hint to cross it and go forward.
2. Generate instruction, when user should change the street side.
3. If possible, announce the type of pedestrian crossing: if has  traffic lights, even better if has tag traffic_signals:sound=yes, or traffic_signals:vibration=yes.
4. Some further improvements and optimizations would be possible in this model, e.g. costing_options to avoid huge streets with no signals if crossing needed, and so on, but not all can be possible to implement without changing the data structures, and even in the basic form, with points 1 and 2, it would be great help.

Possibly related to [#2942: Pedestrian crossing should be a separate TripLeg_Use ](https://github.com/valhalla/valhalla/issues/2942)

grzezlo:
Special optional mode implemented in Valhalla turn-by-turn, for use in Seeing Assistant applications from @transition-technologies:
- Announcing crossed streets, the stairs, bridges, tunnels, gates and bollards, which need to be passed on route.
- Information about traffic signals on crosswalks.
- Route numbers not announced for named routes.
- Added blind_user_mode switch in query parameters.

No idea how to write the tests, also where to announce this new feature  in `CHANGELOG.md` (maybe under `ADDED: Indoor routing` near line 84?) 

Added processing of crossing=traffic_signals in `mjolnir/pbfgraphparser.c`, so updated `taginfo.json' too.

# Issue
fixes #2989

## Tasklist

 - [x] Add tests

## Requirements / Relations
No special requirements.


kevinkreiser:
This is awesome I'll take a look at it more closely tonight most likely

kevinkreiser:
@grzezlo would there be any chance you could update your branch to get the latest master merged in? i know its been a long time but i'm hoping you'd still be interested in merging this

grzezlo:
No problem, @kevinkreiser, merged.

nilsnolde:
I'm just looking at this again after a long time. Sorry for the wait @grzezlo I could fully understand if you moved on by now. I'd still be super interested to merge this and I know @kevinkreiser as well. So I'd help out on this if you'd prefer that.

One thing I'm wondering: how about including the OSM tag `tactile_paving`. It's used quite well it seems, [> 3 Mio ways](https://taginfo.openstreetmap.org/keys/tactile_paving#overview) and could be used to prefer edges when `blind_user_mode = true`. Could also be done in a separate PR. Maybe that'd even be better. I'll spend some more time in the next weeks reviewing this work. Thanks again!

proto/options.proto
```diff
@@ -482,4 +482,5 @@ message Options {
     uint32 matrix_locations = 54;                                  // a one to many or many to one time distance matrix. Does not affect
   }                                                                // sources_to_targets when either sources or targets has more than 1 location
                                                                    // or when CostMatrix is the selected matrix mode.
+  bool blind_user_mode = 81;                           // route: additional maneuvers and informations for blind users.
```
nilsnolde, line 485: I was just looking at "vehicle types" and realized we already have smth you want in our request parameters, however, not documented. In PBF we have `transport_type` which is being set by either `bicycle_type` request param or `type` for all other modes, also pedestrian (not documented). Maybe we should use that instead of this boolean? Then we can make it a `TravelType` which will make more sense once we get to implement `tactile_paving` in costing. What do you think @grzezlo ?

grzezlo:
Thanks @nilsnolde for your comments and first of all for interest in the subject. Sorry for my late answer. In fact I moved to other projects, but Valhalla is still interesting for me and used as a routing engine in our app.
Regarding your idea: my implementation of the blindUserMode focused primarily on the narrative directions extension, as a next stage I thought about extending it to suggest some route planning optimisations, but I didn't manage it in given time.
But your idea opens exciting possibilities.
Preferring tactile pawing is one of them, other would be to prefer intersections with "traffic_signals:sound" or "traffic_signals:vibration" tags.
In some cases Valhalla plans route straight via the intersection, not giving precisely the crossings, although they are drawn on map: some workaround was to make our own customizations of "walkway_factor" and "sidewalk_factor" in query, but it affected some other cases which became non-optimal.
I thought about forcing the engine in blind user mode to always route via crossings on intersections if they're drawn, but didn't have enough understanding of the internal route planning mechanism to achieve it.

nilsnolde:
Thanks for your good thoughts here @grzezlo ! It's again a while gone that I said I'll look into it more. It's far down the PR list by now, I guess it's not as much hitting my attention, I'll try to change that, I really want this functionality.

Is it ok for you if I commit to your branch instead of reviewing? I mean both in the end of course, but maybe that'd be more efficient?

kevinkreiser:
i felt the same way about this but again didnt find anyy time. unless @grzezlo has time to also work on it, i would suggest you make a new branch from master and then merge @grzezlo's branch into that. that way you can let his branch unchanged and have all your work in a place you can more easily commit to

grzezlo:
@nilsnolde  modyfying this branch is ok for me, in such scenario all work on it will be in one place.

chrstnbwnkl:
I have picked this up again, and apart from resolving merge conflicts, I did two things:

  - move the request option from `/blind_user_mode` to `/costing_options/type`, making use of an existing parameter
  - add some gurka tests 

Having it as a costing option while not actually impacting the costing at all is a bit misleading maybe, but @nilsnolde [pointed out](https://github.com/valhalla/valhalla/pull/3694#issuecomment-1837269975) that also modifying the costing could be a good addition to this in the future.

grzezlo:
Thanks @chrstnbwnkl for your contributions, and especially for the tests: being myself blind user of screen reading software I have some troubles to effectively work with ASCII-art things, and the test maps looks like such.
I thout it looks I have no other choice than to return to Valhalla and try to do something more about Blind user mode, but I recall I failed earlier to implement some non-trivial algorithm decision changes I wrote about in my comment on on Dec 21, 2023.
Thanks again for making this huge step toward including blindUserMode in main code of Valhalla.

chrstnbwnkl:
@grzezlo I understand how these ASCII maps are hard to comprehend only using screen readers, glad I could help here. It's impressive how you navigate a code base as large as Valhalla's with a screen reader anyways!

Regarding your earlier comment, I think we can move it one step at a time and focus on the blind specific costing in a future PR! I am definitely interested to help with that.

chrstnbwnkl:
@grzezlo There are bi-weekly meetings Mondays and Wednesdays where we go over PRs and give and get feedback on them. If you are interested, you can join one of these and we can go over this one! They usually happen around 14:30 Berlin time.

nilsnolde:
That’d be great:) Would be able to join next Monday at 2:30 pm CET @grzezlo? I think this Wednesday we already have a date with @ianthetechie for the golf cart PR

grzezlo:
It would be great to take part in such meeting.
Next monday is perfect for me, tomorrow i wouldn't be able to join, so nice, that other PR will be discussed.

kevinkreiser:
@grzezlo ill send you an invite

src/odin/maneuversbuilder.cc
```diff
@@ -86,7 +86,10 @@ namespace valhalla {
 namespace odin {
 
 ManeuversBuilder::ManeuversBuilder(const Options& options, EnhancedTripLeg* etp)
-    : options_(options), trip_path_(etp) {
+    : options_(options), trip_path_(etp),
+      blind_mode_(options.costing_type() == Costing_Type_pedestrian &&
+                  options.costings().find(Costing::pedestrian)->second.options().transport_type() ==
+                      "blind") {
```
kevinkreiser, line 92: ok so to look a bit more like the other types we should:

over in pedestriancost.cc we parse out kFook and kWheelchair, we should add kBlind
then in triplegbuilder we call on a trip edge set_pedestrian_type which stores this type on the edges of the path so we'll need to update the proto with that blind value as well (trip edge is a proto struct)

src/mjolnir/pbfgraphparser.cc
```diff
@@ -1927,8 +1927,10 @@ struct graph_callback : public OSMPBF::Callback {
         // node
         n.set_state_iso_index(osmdata_.node_names.index(tag.second));
         ++osmdata_.node_name_count;
+      } else if (tag.first == "crossing" && tag.second == "traffic_signals") {
+        n.set_traffic_signal(true);
       } else if (tag.first == "highway") {
-        n.set_traffic_signal(tag.second == "traffic_signals");
+        n.set_traffic_signal(n.traffic_signal() || tag.second == "traffic_signals");
```
kevinkreiser, line 1933: this has the potential to make the alogrithm think that we have to stop multiple times in a row in quick succession when a car traffic light and a cross walk light are next to each other. we might want to move pedestrian crossing to be a new node type or use the spare bit on node

valhalla/odin/narrative_dictionary.h
```diff
@@ -134,6 +144,7 @@ constexpr auto kCrossStreetNamesTag = "<CROSS_STREET_NAMES>";
 constexpr auto kRoundaboutExitStreetNamesTag = "<ROUNDABOUT_EXIT_STREET_NAMES>";
 constexpr auto kRoundaboutExitBeginStreetNamesTag = "<ROUNDABOUT_EXIT_BEGIN_STREET_NAMES>";
 constexpr auto kRampExitNumbersVisualTag = "<EXIT_NUMBERS>";
+constexpr auto kObjectLabelTag = "<object_label>";
```
chrstnbwnkl, line 147: upper case

chrstnbwnkl:
After some review rounds now, the only big question remaining is how to enable announcements for passing pedestrian traffic signals without messing with `stop_impact` on `DirectedEdge` and without compromising backwards compatibility (as @gknisely pointed out, adding a new `NodeType` on `TripLeg` would lead to a runtime error).

kevinkreiser:
is there anything left to do here? from my point of view we covered most of it and since we reverted the graphparser change we can merge now and come back to the whole crosswalk (traffic_signals) thing as a separate issue

chrstnbwnkl:
From my side this is all set. In the last CI run, the tests I added failed only in the build release (and I wasn't able to reproduce it locally). Just merged master, so let's see if this was a hick up of some sort

chrstnbwnkl:
Okay, seems like I have to have another look at those tests

kevinkreiser (approved):
thank you so much @grzezlo for startign this and @chrstnbwnkl for bringing it over the line!

mkandulavm:
Hi

I am getting this exception when making a route call
Caught an exception: No such node (instructions.pass)

Any possibility above merged code requires changes for graph tile generation ?
I ask as the "instructions.pass" line in narrative_dictionary.h was recently commited.

nilsnolde:
from what I can see that key is in all 29 published locales, so your problem seems to be that you're again not really on master. updating locales for odin is done during building the library, which will write them from locales/*.json to a public header in `${CMAKE_BINARY_DIR}/odin/locales.h`. 

and as a general rule, unless we bump the major version, we **never require** new tile builds for any new commits..

nilsnolde:
are you sure you're really _re-compiling_ valhalla when you pull (vs just installing new headers..)?

mkandulavm:
@nilsnolde thank you for the comment. I have a custom locale which is missing the "pass". I am not aware of this to be added. It would be easier if the exception pointed out such a update is required for the locales. 
Either ways, thanks again ! I will confirm if this got fixed.

nilsnolde:
Well we don’t expect people to bring their own locales tbh. I’d recommend to consider adding it to transifex. If it’s > 95% translated we’ll add it to the repo and this sort of thing can’t happen.

PR #3700, opened 2022-08-08
GTFS Support: Convert PBFs to Transit Tiles

### LINKED ISSUE #3630: GTFS Support: Connecting to the Road Network
Now that we have GraphTiles of transit data all we need to do is to connect those to the rest of the graph. There are a couple of things to know though...

The current code relies on the transit stop to have a "connecting way id" attached to it. The code then looks for the closest edge in the graph with that way id before attaching it. You can find this implementation here: https://github.com/valhalla/valhalla/blob/master/src/mjolnir/transitbuilder.cc#L449

The problem is that, at present, the way id comes from transitland who in turn gets the id by calling `locate` (a valhalla service). So we have a bit of a cyclical problem here... You might think, great we can cut out the middle man but the problem there is that the middle man in question calls locate on a valhalla that has a spatial index loaded into it so that it can find the actual closest edge. At the point in which we add transit to the graph though, ie where we need to put this, that index is not yet created.. on top of that we also use reachability to throw out results that dont connect to much of the rest of the graph and finally we use it to make sure to only look at the lower class roads.

So how then do we find the appropriate edge to connect to? How do we find the closest edge in the graph if we cant call locate (loki::search)? We have some options:

1. do a best attempt looking in the tiles we have and giving up after some threshold. we will hit edges cases where we pick a bad candidate
2. write a quick spatial index with off the shelf libraries, boost in memory or spatialite. this could be slow or not fit in ram or both and it still doesnt allow us to check the reachability of the edge
3. move the code around so that connecting the transit data to the graph happens after the spatial index already exists (since the transit data is not in the spatial index anyway)

I think number 3 is the most compelling here as it gives us the best results. It also gives us the chance to take transit out of the main pipeline of tile creation and add it as a post processing step which, architecturally, is always prefered. we s hould do some investigation to see which of the 3 is best though or perhaps another option

This task is done when:

1. the tile building process can attach level 3 transit GraphTiles to the level 2 graph and those tiles are accessible via GraphReader

chris-jpark:
the basic gist of this pr:

- is a ton of clean up from the first couple prs
  - there were tons of spots where the id's were being converted from strings to integers but this is not really how the data is supposed to look
  - convert_transit had lots of places where it was incorrectly modifying the graphid to be level 4 (thought it was doing 2 + 1 but 
  - we already had 3 for the level so it was getting 4, which messed everything up)
  - also fixed the date time ingestion for the schedule information
  - lots of little changes to things like names and helper functions through out the library. maybe some are overkill please leave a comment if anyone feels strongly and we can remove
- this also fixes #3630 in that we've implemented a way to find the connections to the graph via a naive loki-like search algorithm (doesnt look in adjacent files and isnt that fast but who cares for now)
- it also allows you to put a lat lon or a way id to influence the place where a station will connect to the road network, at present there is no way to get this from gtfs but we could side load the data (simple csv file?) in the conversion process of gtfs to pbf tiles

In the end the graph is connected between level 3 and level 2 but there are some things that this PR does not do. in fact transit routing still doesnt work but its getting close. here are our next steps in semi-priority order (which i posted down below but feel would be better off in the PR description):

1. fix stitching in ingest_transit. for whatever reason the stitcher cant find the missing destination/origin for 1 of the stop pairs in the test. the result is we get a couple of logs that say we are bailing on one of the stop pairs. this is because the pair connects across the tile boundary (which is a great thing to test)
2. make sure convert transit can actually add the edges between stations (actually between stations platforms) so that you can actually route on a kRail or kBus edge
3. the routing algorithm doesnt work, says it cant find a transit station near by (shouldnt be the case we know the connection does work currently so perhaps). im hoping fixing the first 2 will make this one go away but i expect there to be more routing issues
4. fix the routing algorithms likely bugs due to the code being unexorcised for years
5. _______IF WE MADE IT HERE WE COULD SAY WE ARE IN BETA WITH CAVEATS_______
6. hierarchy builder is not fully robust to the presence of transit data, so currently we cant build hierarchies if we want to have transit data as well. this in itself wouldnt stop us from supporting transit but its definitely required to be fixed before transit can be considered fully functional again.
7. unit test multi feed imports and fix if broken
8. _______IF WE MADE IT HERE WE HAVE SOMETHING WE COULD SAY IS OUT OF BETA_______
9. transit connect edges are not always formed on both sides of the street network edge, this happens because we  currently dont look in adjacent tiles to find  the end nodes. ive put a bunch of hints in the code about what to do to fix this (and also get better matches for edges to connect stations to)
10. clean up the code in convert transit where we hook up the 3 transit node types and connect them to each other. i wrote a big todo in there
11. another group of  tasks to tackle adding transit to the graph after the graph is fully built as mentioned over in: https://github.com/valhalla/valhalla/issues/3806#issuecomment-1339231858

I can turn the above into issues that we can burn down 1 by 1.


chris-jpark:
This pull request deals wth the parsing of protobuf files to build the Valhalla transit tiles, and a clean up of building protobufs.

While building protobufs, I made changes to ensure that the test cases simulate stitching (i.e. One transit trip spans multiple tiles), and that in that case, a pair of stops can acknowledge that one of the stops can be outside the current tile. 
In this case, we process the intra-stop pairs first and then proceed to 'stitch' together the missing links inter-tiles. 

To meet the requirements to create transit tiles pre-set by Valhalla, there needed to be additional data added on top of GTFS's data. 
This includes: 
- finding the nearest OSM way_id from a given transit object, such that the transit data can be connected to the road data. This is done by projecting the transit map onto the road map, and iterating through the way_ids in the given tile, and checking the perpendicular distance of each way_id. 
- generating virtual egresses and parent stations when GTFS does not provide them. 

Most of the debugging has been completed, but there still persist bugs when simulating transit routing. Currently, it thinks that the transit station is 'too far' and refuses to create the route.

src/mjolnir/convert_transit.cc
```diff
@@ -99,6 +99,73 @@ struct StopEdges {
...
+  // TODO: if tile is null return empty vector;
+  if (transit_tile.nodes_size() == 0) {
+    return std::vector<uint64_t>{};
+  }
+  auto t1 = std::chrono::high_resolution_clock::now();
+  auto scoped_finally = make_finally([&t1, size = transit_tile.nodes_size()]() {
+    auto t2 = std::chrono::high_resolution_clock::now();
+    uint32_t secs = std::chrono::duration_cast<std::chrono::seconds>(t2 - t1).count();
+    LOG_INFO("Projection Finished - Projection of " + std::to_string(size) + " bike station  took " +
+             std::to_string(secs) + " secs");
```
kevinkreiser, line 119: we need to fix stuff like this

src/mjolnir/convert_transit.cc
```diff
@@ -99,6 +99,73 @@ struct StopEdges {
...
+
+  // auto local_level = TileHierarchy::levels().back().level;
+  std::vector<uint64_t> best_ways;
+  std::map<GraphId, size_t> edge_count;
+
+  for (const auto& egress : transit_tile.nodes()) {
+    uint64_t best_wayid = std::numeric_limits<uint64_t>::max();
+    // In this loop, we try to find the way on which to project the bss node by iterating all nodes in
+    // its corresponding tile... Not a good idea in term of performance... any better idea???
+    auto bss_ll = PointLL{egress.lon(), egress.lat()};
```
kevinkreiser, line 130: and this

src/mjolnir/convert_transit.cc
```diff
@@ -535,12 +602,24 @@ void AddToGraph(GraphTileBuilder& tilebuilder_transit,
...
+                uint32_t& no_dir_edge_count,
+                GraphReader& reader) {
   auto t1 = std::chrono::high_resolution_clock::now();
 
   // Get Transit PBF data for this tile
   Transit transit = read_pbf(tile, lock);
 
+  GraphId local_tile_id(tileid.tileid(), TileHierarchy::levels().back().level, 0);
+  graph_tile_ptr local_tile = reader.GetGraphTile(local_tile_id);
+  //  GraphTileBuilder tilebuilder_local(reader.tile_dir(), local_tile_id, true);
```
kevinkreiser, line 614: and this

src/mjolnir/convert_transit.cc
```diff
@@ -676,8 +760,12 @@ void AddToGraph(GraphTileBuilder& tilebuilder_transit,
         egress_node.set_stop_index(index);
         egress_node.set_timezone(timezone);
         egress_node.set_edge_index(tilebuilder_transit.directededges().size());
-        egress_node.set_connecting_wayid(egress.osm_way_id());
 
+        if (local_tile != nullptr && best_wayids[index] != std::numeric_limits<uint64_t>::max()) {
```
kevinkreiser, line 764: if we dont do this the station is then not connected to the graph. we should at least log a warning in that case because it means people will not be able to get in or out at that station

src/mjolnir/convert_transit.cc
```diff
@@ -1055,7 +1144,7 @@ void build_tiles(const boost::property_tree::ptree& pt,
     // a writeable instance (deserialize it so we can add to it)
     lock.lock();
 
-    GraphId transit_tile_id = GraphId(tile_id.tileid(), tile_id.level() + 1, tile_id.id());
+    GraphId transit_tile_id = GraphId(tile_id.tileid(), tile_id.level(), tile_id.id());
```
kevinkreiser, line 1147: ok so outside of this function we changed it to loop over transit tile ids now (ie level 3) so we no longer need this

src/mjolnir/convert_transit.cc
```diff
@@ -1134,6 +1224,8 @@ void build_tiles(const boost::property_tree::ptree& pt,
       }
 
       GraphId platform_pbf_graphid = GraphId(platform.graphid());
+      LOG_INFO(std::to_string(platform_pbf_graphid) + " " +
+               std::to_string(departures.begin()->second.orig_pbf_graphid));
```
kevinkreiser, line 1228: is this debug?

src/mjolnir/convert_transit.cc
```diff
@@ -1236,7 +1328,8 @@ std::unordered_set<GraphId> convert_transit(const ptree& pt) {
   // figure out which transit tiles even exist
   filesystem::recursive_directory_iterator transit_file_itr(
       pt.get<std::string>("mjolnir.transit_dir") + filesystem::path::preferred_separator +
-      std::to_string(TileHierarchy::levels().back().level));
+      std::to_string(TileHierarchy::GetTransitLevel().level));
+  // it looks like Tile Hierarchy has level 2 as a maximum
```
kevinkreiser, line 1332: nit:
```suggestion
```

CHANGELOG.md
```diff
@@ -105,6 +105,7 @@
    * CHANGED: Precise definition of types of edges on which BSS could be projected [#3658](https://github.com/valhalla/valhalla/pull/3663)
    * CHANGED: Remove duplicate implementation of `adjust_scores` [#3673](https://github.com/valhalla/valhalla/pull/3673)
    * ADDED: convert GTFS data into protobuf tiles [#3629](https://github.com/valhalla/valhalla/issues/3629)
+   * ADDED: convert GTFS protobufs into Transit Tiles [#3700](https://github.com/valhalla/valhalla/pull/3700)
```
kevinkreiser, line 108: move this down to the bottom

src/mjolnir/ingest_transit.cc
```diff
@@ -280,101 +369,127 @@ bool write_stop_pairs(Transit& tile,
     const std::string currFeedPath = feed_trip.feed;
     const auto& feed = tripFeeds(feed_trip);
 
-    gtfs::Trip currTrip = *(feed.get_trip(tile_tripId));
+    auto currTrip = feed.get_trip(tile_tripId);
 
     // already sorted by stop_sequence
-    auto tile_stopTimes = feed.get_stop_times_for_trip(tile_tripId);
+    const auto& tile_stopTimes = feed.get_stop_times_for_trip(tile_tripId);
```
kevinkreiser, line 375: we need to check things like this. the api actually returns a copy here, not a reference. i think the compiler smooths this over for us but the reader of this code would make assumptions that wouldnt be true in this case.

src/mjolnir/ingest_transit.cc
```diff
@@ -280,101 +369,127 @@ bool write_stop_pairs(Transit& tile,
...
-        // TODO: Determine what to do if there are multiple frequencies.
-        // currently, choose fist one
-        if (currFrequencies.size() > 0) {
-          stop_pair->set_frequency_end_time(DateTime::seconds_from_midnight(freq_end_time));
-          stop_pair->set_frequency_headway_seconds(currFrequencies[0].headway_secs);
-        }
+        if (!feed.get_frequencies(currTrip->trip_id).empty()) {
+          const auto& currFrequencies = feed.get_frequencies(currTrip->trip_id);
+          auto freq_start_time = (currFrequencies[0].start_time.get_raw_time());
+          auto freq_end_time = (currFrequencies[0].start_time.get_raw_time());
```
kevinkreiser, line 476: i know this didnt really change but this looks wrong. its using the same value as the start time

kevinkreiser:
ok i made a bunch of comments. most of them are small but at least a couple will require a little bit more work. i really want to merge this but id prefer not to merge it until its a little cleaner. basically im happy to merge even if its still not working but not until the review comments are addressed

kevinkreiser:
ok the reason this is currently failing is because the transit graph is not connected to the regular graph. the reason for that is because the regular graph has already gone through the creation of heirarchies but the old code expected this to not have been the case. remember the old code expected transitland to just provide the right way id from the beginning but now we have to do a pass ourselves to get it. we have two options to fix this:

1. keep using the naive algorithm which finds the closest edge by simply looking at all edges in the tile but update it to also check other levels of the heirarchy
2. use loki to find the best candidates which requires some cmake code motion to get the dependencies right. this is the best code re-use solution but makes things complicated in that mjolnir library code would depend on loki library code. an obvious proper organization of the code has not come to me yet..

nilsnolde:
Right, that's a bit painful.. Duplicating much of loki's search would be annoying, but so is depending on loki in mjolnir a bit. Would be much worse though if it'd be the other way around. We don't (yet) explicitly provide options to only build mjolnir or exclude loki in a library build, right? Are there other reasons beside removing some potential modularization and some CMake complication?

kevinkreiser:
yeah so instead of tackling using loki from inside mjolnir, i'm just keeping the naive logic that does a simple closest edge in same tile. i updated it as i mentioned above to look at other levels of the heirarchy and that seems to work though we have problems in transit builder now.

1. if you find an edge that isnt in level 2 to try to make the connection to, it complains about making a really long transit connect edge (which makes me think it wont use it). for the time being i made the graph all residential (and put a TODO) so it gets past this bit
2. when it tries to add stuff to the graph it breaks the tiles in such a way that we throw a nodeinfo index out of bounds error during heirarchy builder.

so yeah a few more things to iron out yet :smile:

kevinkreiser:
Ok so this PR is much closer than it was several months ago however its still not working and there are at least 3 problems. i believe they should be fixed in the following priority order:

1. fix stitching in ingest_transit. for whatever reason the stitcher cant find the missing destination/origin for 1 of the stop pairs in the test. the result is we get a couple of logs that say we are bailing on one of the stop pairs. this is because the pair connects across the tile boundary (which is a great thing to test)
2. make sure convert transit can actually add the edges between stations (actually between stations platforms) so that you can actually route on a kRail or kBus edge
3. the routing algorithm doesnt work, says it cant find a transit station near by (shouldnt be the case we know the connection does work currently so perhaps). im hoping fixing the first 2 will make this one go away.
4. hierarchy builder is not fully robust to the presence of transit data, so currently we cant build hierarchies if we want to have transit data as well. this in itself wouldnt stop us from supporting transit but its definitely required to be fixed before transit can be considered fully functional again.
5. transit connect edges are not always formed on both sides of the street network edge, this happens because we  currently dont look in adjacent tiles to find  the end nodes. ive put a bunch of hints in the code about what to do to fix this (and also get better matches for edges to connect stations to)
6. clean up the code in convert transit where we hook up the 3 transit node types and connect them to each other. i wrote a big todo in there
7. unit test multi feed imports
8. another group of  tasks to tackle adding transit to the graph after the graph is fully built as mentioned over in: https://github.com/valhalla/valhalla/issues/3806#issuecomment-1339231858

I can turn the above into issues that we can burn down 1 by 1. I would recommend that we are ready to merge this as a work in progress that is not yet fully supported

kevinkreiser:
@nilsnolde @dnesbitt61 @gknisely im happy to review this with you before merging, i know its large so i'd be happy to walk you through it

nilsnolde (approved):
Alright, let's (keep) do(ing) this:)

nilsnolde:
@kevinkreiser I think CI failing is just a tiny typo when you were hacking in some proto stuff.

kevinkreiser:
@nilsnolde yeah, as soon as we get end to end working, we can go back and delete one of the protos and the fetch transit binary. it currently doesnt really work anyway so there is no point in maintaining it after it stops being a useful "illustration" of how it used to work

nilsnolde (approved):

PR #3702, opened 2022-08-13
Use `starts_with()` instead of `substr(0, N)` getting and comparing to prefix

DmitryAk:
# Issue

There were several places in code such as:
https://github.com/valhalla/valhalla/blob/e598c09d7aa737f060ccffb4ac18351201567200/src/mjolnir/pbfgraphparser.cc#L1815
It seems not very reliable and not very performant:
1. It is required to specify the exact prefix size in `substr(0, N)`. If there is a mistake - the result of comparing the substr with prefix will always be `false`. It looks not very convenient and not very reliable.
2. It always creates an auxiliary substring, and this substring is compared to the prefix (so even in best case this requires O(N), where N is prefix size). At the same time it is possible that `starts_with` can just compare the first chars - and reject further comparing (so in best case this requires O(1)). Also `starts_with` doesn't require unnecessary auxiliary substring (so no unnecessary allocation is guaranteed). To check this, a benchmark test was introduced in this PR, results (Release configuration, x86_64, clang-1316.0.21.2.5):

```
With substr() it took 8224 ms
With starts_with() it took 1963 ms
```

## Tasklist

 - [x] Add tests


kevinkreiser:
Looks good to me just waiting for the build to pass

kevinkreiser (approved):

PR #3710, opened 2022-08-16
add ferry support to truck model

### LINKED ISSUE #3709: ferry for hgv not respected
we found that ferries are (basically) disabled for hgv, `use_ferry` is not parsed and hence the transition cost is 6 hours by default. however, there are lots of [ferries with hgv access](https://overpass-turbo.eu/s/1l3K) and I wonder if there was a specific reason for the current behavior or just an unintended oversight? it seems it's very intended, it's the only use of `BaseCostingOptionsConfig.disable_ferry_`.

it should be a quick addition if wanted.

nilsnolde:
fixes #3709 

still owing it a test. will do tmrw.

### DIFF
--- a/test/gurka/test_ferry_connections.cc
+++ b/test/gurka/test_ferry_connections.cc
@@ -92,6 +92,42 @@ TEST(Standalone, ShortFerry) {
   EXPECT_EQ(DE_edge->classification(), baldr::RoadClass::kPrimary);
 }
 
+TEST(Standalone, TruckFerryDuration) {
+  // corresponds to 33.3 m/sec (120 km/h) in the below map
+  uint32_t ferry_secs = 9;
+
+  const std::string ascii_map = R"(
+          A--B--C--D
+          |        |
+          E--------F
+    )";
+
+  const gurka::ways ways = {{"AB", {{"highway", "secondary"}}},
+                            {"BC",
+                             {{"motor_vehicle", "yes"},
+                              {"motorcar", "yes"},
+                              {"bicycle", "yes"},
+                              {"hgv", "yes"},
+                              {"duration", "00:00:0" + std::to_string(ferry_secs)},
+                              {"route", "ferry"},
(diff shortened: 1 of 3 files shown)


kevinkreiser:
@gknisely any reason why we didnt allow ferries before? is it because the tags dont tell us enough info about whether or not a specific config of HGV would work? eg maybe the ferry will allow a truck with certain length, height, width, weight but not others but we dont know because its not tagged reliable and so to avoid routing on a ferry a truck cant actually take?

nilsnolde:
The few ferries in German waters that carry hgv are simply tagged with hgv=yes. I guess the normal restrictions (maxweight etc) would also work on ferry edges right? But not sure if that’s the OSM way of tagging ferries..

test/gurka/test_ferry_connections.cc
```diff
@@ -92,6 +92,42 @@ TEST(Standalone, ShortFerry) {
...
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_ferry_duration");
+
+  valhalla::Api fastest = gurka::do_action(valhalla::Options::route, map, {"A", "D"}, "truck",
+                                           {{"/costing_options/truck/use_ferry", "1"},
+                                            {"/costing_options/truck/ferry_cost", "0"}});
+
+  // verify we took the ferry edge and the duration tag was respected
+  auto ferry_edge = fastest.trip().routes(0).legs(0).node(1).edge();
+  ASSERT_EQ(ferry_edge.use(), valhalla::TripLeg_Use::TripLeg_Use_kFerryUse);
+  ASSERT_NEAR(ferry_edge.speed(), ferry_edge.length_km() / (ferry_secs * kHourPerSec), 0.1);
```
nilsnolde, line 128: was a good opportunity to add a test for a ferry's `duration` tag

kevinkreiser (approved):
perfect, nothing to add here

PR #3713, opened 2022-08-17
format & lint python code

### LINKED ISSUE #3613: proper pre-commit script

nilsnolde:
following up on #3637 to close #3613

only python files in the `./scripts` and `src/bindings` directories will be linted/formatted, other files are not too relevant IMHO. also adding a check to circleci.


.pre-commit-config.yaml
```diff
@@ -1,8 +1,33 @@
 repos:
-  # Autoformat: Shell files
+  # C++
   - repo: local
     hooks:
     - id: format
       name: format
       entry: bash ./scripts/format.sh
       language: system
+      pass_filenames: false
```
nilsnolde, line 9: There were some strange in the pre-commit CI run and I realized by default, it runs each hook on every file it catches in the diff, so format.sh would run x times (also locally when committing btw). `pass_filename: false` fixes that so pre-commit only runs format.sh once.

.circleci/config.yml
```diff
@@ -102,7 +102,6 @@ jobs:
                 echo "Changes in last commit do not need CI. Skipping step"
                 circleci-agent step halt
             fi
-      - run: ./scripts/format.sh && ./scripts/error_on_dirty.sh
```
kevinkreiser, line 105: by not doing this we waste CI hours. i had this here so that builds would fail fast rather than having you find out they all passed except the lint one. i suppose we could make a separate ci job to do linting first that the other ones wait on to avoid this? but i thought it was cheap enough we could just have them all do it

scripts/args.py
```diff
@@ -1,14 +1,25 @@
 #!/usr/bin/env python3
```
kevinkreiser, line 1: i feel like this is an opportunity to delete some scripts that we never use :smile:

.pre-commit-config.yaml
```diff
@@ -1,8 +1,33 @@
...
+      - id: flake8
+        # E501: ignore line length (black auto-formats)
+        # E731: ignore replace-lambda-with-def warnings
+        # E203: ignore "whitespace before ':'"
+        args: [
+          '--max-line-length=105',
+          '--extend-ignore=E501,E731,E203',
+          '--extend-exclude=src/bindings/python/__init__.py'
+        ]
+        files: ^(src/bindings/python/|scripts/)
```
kevinkreiser, line 33: is there any chance we can allow the use of single quotes and not force double quotes. for whatever reason i am partial to them, unless you have a strong opinion

.circleci/config.yml
nilsnolde: ah right, makes sense. it is cheap enough, I'll revert

.pre-commit-config.yaml
nilsnolde: na, I don't and also don't understand why that's an issue for a formatter. I included it but will wait for the master hot fix at #3714 before pushing

.pre-commit-config.yaml
```diff
@@ -1,8 +1,33 @@
...
       language: system
+      pass_filenames: false
+
+  # Python formatter
+  - repo: https://github.com/psf/black
+    rev: 22.6.0
+    hooks:
+      - id: black
+        language_version: python3
+        args: ['--line-length=105', '--skip-string-normalization']
```
nilsnolde, line 17: `--skip-string-normalization` prevents single to double quote conversion

nilsnolde:
had to force push after getting rid of the single-quote rule.

also included the docker lines of #3715 already, guess they'll be merged either way. changelog is still slightly messed up, but I wanna wait until circleci built successfully and see if `needs_ci_run` still works (I have a feeling it doesn't).

.circleci/config.yml
```diff
@@ -20,6 +20,13 @@ commands:
       - run: sudo bash ./scripts/install-linux-deps.sh
 
 jobs:
+  lint:
+    machine:
+      image: ubuntu-2004:current
+    steps:
+      - checkout
+      - run: sudo python3 -m pip install pre-commit && pre-commit run --all-files && ./scripts/error_on_dirty.sh
```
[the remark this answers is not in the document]
nilsnolde: this new job messes with the `needs_ci_run` check. I really don't feel like debugging on CI now.. I'll do it the other way as before..

kevinkreiser (approved):
now with some ci cache i hope the release build for master goes fast again! looked like the docker image build is now down to 35 min which is half the time!

PR #3714, opened 2022-08-17
fix docker setup

nilsnolde:
docker CI workflwo is failing on master after merging #3690 .

seems it's only missing having a script made executable. actually, let's wait, I'll build myself locally first (I now think I didn't do that before..)


Dockerfile
```diff
@@ -2,14 +2,16 @@
 FROM ubuntu:20.04 as builder 
 MAINTAINER Kevin Kreiser <kevinkreiser@gmail.com>
 
+ARG CONCURRENCY=$(nproc)
```
nilsnolde, line 5: the docker build killed my computer a few times, so I added this:

`docker build --build-arg CONCURRENCY=20 -t valhalla/valhalla:run-latest .`

nilsnolde:
that should do it @kevinkreiser

Dockerfile
```diff
@@ -2,14 +2,16 @@
...
 # set paths
 ENV PATH /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
 ENV LD_LIBRARY_PATH /usr/local/lib:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib32:/usr/lib32
 
 # install deps
 WORKDIR /usr/local/src/valhalla
-COPY ./scripts/install-linux-deps.sh /usr/local/src/valhalla/scripts
-RUN bash ./scripts/install-linux-deps.sh
+COPY ./scripts/install-linux-deps.sh /usr/local/src/valhalla/scripts/install-linux-deps.sh
+RUN bash /usr/local/src/valhalla/scripts/install-linux-deps.sh
```
kevinkreiser, line 14: so just manually use bash, that should work if its about not being executable

kevinkreiser (approved):

Dockerfile
nilsnolde: it was actually that docker copied the script and names it literally `/usr/local/src/valhalla/scripts`.. always getting confused with these docker copy files shite.. anyways, I could've it built it myself in the first PR :sweat_smile:

nilsnolde:
damn it.. still failing.. seems docker doesn't like defaulting to an expression for a build arg.. 

give me a minute, I'll quickly finish the work on the other branch and come up with a diff, so maybe you can commit directly to master.

PR #3715, opened 2022-08-17
no expressions to default build args

nilsnolde:
really sorry.. I forgot to _not_ set the build arg for a docker build locally before merging.. now it'll work, ffs!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM ubuntu:20.04 as builder 
 MAINTAINER Kevin Kreiser <kevinkreiser@gmail.com>
 
-ARG CONCURRENCY=$(nproc)
+ARG CONCURRENCY
 
 # set paths
 ENV PATH /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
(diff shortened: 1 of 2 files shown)


nilsnolde:
made it a pr but canceled all ci jobs manually

nilsnolde:
so a fresh build (without cache) takes 19 mins on the `release` job and 5 mins on the `debug` job. any idea why? there's not much difference between the 2 builds and I'd have expected a release build to be quicker. it can't be only the `LOGGING_LEVEL=TRACE` option right?

kevinkreiser:
the release build has: `-DENABLE_SANITIZERS=ON` which will be the biggest factor in my opinion

Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN pip install --upgrade conan
 # configure the build with symbols turned on so that crashes can be triaged
 WORKDIR /usr/local/src/valhalla/build
 RUN cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_COMPILER=gcc
-RUN make all -j${CONCURRENCY}
+RUN make all -j${CONCURRENCY:-$(nproc)}
```
kevinkreiser, line 29: i was wondering why you didnt do it like this in the previous one but couldnt be bothered to check docker to see what would happen

kevinkreiser (approved):

nilsnolde:
🤞

PR #3720, opened 2022-08-21
rename turkey admin to türkiye

### LINKED ISSUE #3718: Turkey -> Türkiye
Apparently the name change is now reflected on OSM as well, so we need to update our admin entries. If no one else jumps ahead, I’ll do it later when I’m back at the computer.

nilsnolde:
fixes #3718


kevinkreiser (approved):
too bad for all the text to speech engines which will choke on this but i guess i should expect political machinations generally never consider things such as practicality haha

PR #3735, opened 2022-09-07
Fix elevation serializing

DmitryAk:
# Issue

Fixes some possible mistakes introduced in https://github.com/valhalla/valhalla/pull/3571

@nilsnolde @kevinkreiser please review

## Tasklist

### DIFF
--- a/src/tyr/trace_serializer.cc
+++ b/src/tyr/trace_serializer.cc
@@ -112,16 +112,16 @@ json::ArrayPtr serialize_edges(const AttributesController& controller,
       }
       if (controller(kEdgeMaxDownwardGrade)) {
         if (edge.max_downward_grade() == kNoElevationData) {
-          edge_map->emplace("max_downward_grade", static_cast<int64_t>(edge.max_downward_grade()));
-        } else {
           edge_map->emplace("max_downward_grade", nullptr);
+        } else {
+          edge_map->emplace("max_downward_grade", static_cast<int64_t>(edge.max_downward_grade()));
         }
       }
       if (controller(kEdgeMaxUpwardGrade)) {
-        if (edge.max_downward_grade() == kNoElevationData) {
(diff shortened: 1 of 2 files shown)


src/tyr/trace_serializer.cc
```diff
@@ -112,16 +112,16 @@ json::ArrayPtr serialize_edges(const AttributesController& controller,
       }
       if (controller(kEdgeMaxDownwardGrade)) {
         if (edge.max_downward_grade() == kNoElevationData) {
-          edge_map->emplace("max_downward_grade", static_cast<int64_t>(edge.max_downward_grade()));
-        } else {
           edge_map->emplace("max_downward_grade", nullptr);
+        } else {
+          edge_map->emplace("max_downward_grade", static_cast<int64_t>(edge.max_downward_grade()));
```
DmitryAk, line 117: It looks like branches were misplaced here

src/tyr/trace_serializer.cc
```diff
@@ -112,16 +112,16 @@ json::ArrayPtr serialize_edges(const AttributesController& controller,
...
         }
       }
       if (controller(kEdgeMaxUpwardGrade)) {
-        if (edge.max_downward_grade() == kNoElevationData) {
-          edge_map->emplace("max_upward_grade", static_cast<int64_t>(edge.max_upward_grade()));
-        } else {
+        if (edge.max_upward_grade() == kNoElevationData) {
```
DmitryAk, line 121: It looks like 2 mistakes were here:
1. `max_downward_grade` was used instead of `max_upward_grade`
2. branches were misplaced

dnesbitt61 (approved):
Good catch. I will say though that the max grades are perhaps not that useful without some filtering/smoothing. I tried to use these fields for wheelchair routing a long time ago and abandoned that effort as the grades were a bit inaccurate and noisy - especially on short edges.

kevinkreiser:
build failure was the ephemeral one we get with the elevation builder stuff. there is a separate issue about this

PR #3738, opened 2022-09-10
Fix TimeDistanceMatrix results sequence

DmitryAk:
# Issue

TimeDistanceMatrix results sequence in vector was broken for the case when `sources count` > `targets count`. E.g., if sources == {A,B,C} and targets == {D,E}, it is expected to return a vector
```
{A->D A->E
 B->D B->E
 C->D C->E}
```
but actually before this fix the result was:
```
{A->D B->D
 C->D A->E
 B->E C->E}
```

This PR adds a unit test reproducing the issue + fixes the issue.

## Tasklist

 - [x] Add tests

### DIFF
--- a/test/gurka/test_time_tracking.cc
+++ b/test/gurka/test_time_tracking.cc
@@ -189,7 +189,8 @@ TEST(TimeTracking, routes) {
   };
   const auto layout = gurka::detail::map_to_coordinates(ascii_map, 100);
   auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_time_tracking_make",
-                               {{"mjlonir.timezone", "/path/to/timezone.sqlite"}});
+                               {{"mjlonir.timezone", "/path/to/timezone.sqlite"},
+                                {"thor.source_to_target_algorithm", "timedistancematrix"}});
 
   // pick out a start and end ll by finding the appropriate edges in the graph
   baldr::GraphReader reader(map.config.get_child("mjolnir"));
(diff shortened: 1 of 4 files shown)


test/gurka/test_time_tracking.cc
```diff
@@ -254,6 +255,27 @@ TEST(TimeTracking, routes) {
...
+  d.Parse(matrix.c_str());
+  rapidjson::Value a = d["sources_to_targets"].GetArray()[0].GetArray();
+
+  // check the timings
+  times.clear();
+  for (rapidjson::SizeType i = 0; i < a.Size(); ++i) {
+    times.push_back(a[i]["time"].GetDouble());
+  }
+
+  const std::vector<double> expectedTimes = {51, 193}; // {"A"->"D", "A"->"E"}
```
DmitryAk, line 277: before the fix: actual values were `{51, 34}`

nilsnolde:
I’ll take a look today at this one, been around the matrix code recently anyways.

test/gurka/test_time_tracking.cc
```diff
@@ -254,6 +255,27 @@ TEST(TimeTracking, routes) {
...
+      boost::format(
+          R"({"costing":"auto","sources":[{"lon":%1%,"lat":%2%},{"lon":%3%,"lat":%4%},{"lon":%5%,"lat":%6%}],"targets":[{"lon":%7%,"lat":%8%},{"lon":%9%,"lat":%10%}]})") %
+      map.nodes["A"].first % map.nodes["A"].second % map.nodes["B"].first % map.nodes["B"].second %
+      map.nodes["C"].first % map.nodes["C"].second % map.nodes["D"].first % map.nodes["D"].second %
+      map.nodes["E"].first % map.nodes["E"].second;
+  std::string matrix = actor.matrix(req.str(), nullptr, &api);
+  rapidjson::Document d;
+  d.Parse(matrix.c_str());
+  rapidjson::Value a = d["sources_to_targets"].GetArray()[0].GetArray();
+
```
nilsnolde, line 270: This is a bit of weird place for testing matrix stuff. I'd suggest to glue it to the end of `test/matrix.cc` where the others live too. It's not gurka map enabled but there's other requests already.

nilsnolde (commented):
Thanks a lot @DmitryAk , in fact that bug must've been there forever! More sources than targets is a more unusual case, yet a fully valid one. I'd just think about putting the test to the other matrix tests. Other than that it's good for me.

test/gurka/test_time_tracking.cc
DmitryAk: I'll try to move the test into `test/matrix.cc` today, thank you.

DmitryAk: Done, I've just added a new test into `test/matrix.cc`, and removed the changes from `gurka/test_time_tracking.cc`.

kevinkreiser (approved):

Issue #3135, opened 2021-06-08
Bridleways with foot=true tag results in edges without bicycle access 


davidjcohen:
When an OSM way (in the UK) has the following tags:

highway=bridleway
foot=true

e.g. https://www.openstreetmap.org/way/33114857

we lose bicycle access

see http://localhost:8002/locate?json={"verbose":true,"locations":[{"lat": 53.5992164, "lon": -1.5830084}],"costing":"pedestrian","directions_options":{"units":"miles"}}

Is this the intended behaviour?

It looks like lua/graph.lua sets the "use" to pedestrian in this case, and we don't get the defaults from the admin database.

Thanks in advance

dnesbitt61:
interesting - especially since this is part of a cycling route (relation)

kungfu303:
Having a similar issue with a pedestrian bike/walk bridge in Providence, RI. [Here](https://www.openstreetmap.org/way/609950799) it is marked highway=cycleway, but it has foot=designated. This seems to be removing it completely from bicycle access. What's a resolution (maybe an override as mentioned [in this issue ](https://github.com/valhalla/valhalla/issues/3705)?

kungfu303:
Update: I'll try modifying graph.lua with some extra conditionals ([similar to this](https://github.com/valhalla/valhalla/blob/master/lua/graph.lua#L1813)) at the end so that it checks for bicycle=designated and foot=designated, and ensures that it does have bike_forward access. If there's a better solution, feel free to share

kevinkreiser:
@gknisely would be better to advise here but what you're saying is where I would look first

kungfu303:
Thanks! Also quick question on recompiling after modifying the graph.lua. It says in one of the READMEs that if you modify the lua file, you'd be able to build valhalla without needing to remake the tiles. I tried running from the `valhalla_build_extract -c valhalla.json -v` command, but it seems to not have taken the new graph.lua into account.

gknisely:
This should be working correctly as the user added the bicycle access tags.  This should override any country overrides in the code.  It does look like this is happening for pedestrian access, but it is removing bicycle access for some reason.  I will have to dig more.   @kungfu303 you need to make sure you have a admin db in use for country overrides.  Do you have it created?  Also, in order to build tiles, you use `./valhalla_build_tiles -c ../../conf/valhalla.conf ../Path_to_your_pbf/your.pbf`

kungfu303:
[...]

PR #3746, opened 2022-09-12
Wales and Scotland name change

gknisely:
# Issue

Wales and Scotland had a name change.  We need to come up with a better design and handle name changes.  Not sure what to do yet.  Anyhow #fixes https://github.com/valhalla/valhalla/issues/3135  Added a small test showing how just one user inputted access can shut off country access overrides.

## Tasklist

- [x] Add tests

## Requirements / Relations


lua/admin.lua
```diff
@@ -64,7 +64,7 @@ drive_on_right = {
 ["South Africa"] = "false",
 ["Sri Lanka"] = "false",
 ["Suriname"] = "false",
-["Scotland"] = "false",
+["Alba / Scotland"] = "false",
```
kevinkreiser, line 67: doesnt this mean that someone converting an older planet will also get screwed? i think the compatible fix would be to include both variants right? just support both for a time. its a bit of baggage to carry but its better than getting bug reports :D

lua/admin.lua
```diff
@@ -191,7 +191,7 @@ function rels_proc (kv, nokeys)
            end
          end
        end
-       if kv["name"] == "England" or kv["name"] == "Scotland" or kv["name"] == "Wales" or kv["name"] == "Northern Ireland" then
+       if kv["name"] == "England" or kv["name"] == "Alba / Scotland" or kv["name"] == "Cymru / Wales" or kv["name"] == "Northern Ireland" then
```
gknisely, line 194: > doesnt this mean that someone converting an older planet will also get screwed? i think the compatible fix would be to include both variants right? just support both for a time. its a bit of baggage to carry but its better than getting bug reports :D

@kevinkreiser Yeah I had the following, but was concerned folks would complain.  Thoughts?

```
if kv["name"] == "Alba / Scotland" then
   kv["name"] = "Scotland"
end

if kv["name"] == "Cymru / Wales" then
   kv["name"] = "Wales"
end
```

nilsnolde: IMO fine fix:)

nilsnolde (dismissed):

nilsnolde:
urgh.. windows fails with some vcpkg idiocy.. can someone pls re-trigger that workflow, hoping it's just a hiccup? lint build is the usual coverage fail

lua/admin.lua
kevinkreiser: I had actually meant just leave both checks in and use whichever we find. Which means doing it at the other places where you have changes as well. Basically support both names all the way through the code

kevinkreiser: Yep I know that's what I'm suggesting

kevinkreiser: Do we know when those relations changed? If it was a long while ago then just ignore my comment. If it was recent then it's a pretty huge gotcha to have to know to use the bleeding edge changeset

gknisely: @kevinkreiser Scotland was 8 months ago and Wales I am not sure as it keeps timing out on me.

kevinkreiser: ok cool then we can just ship it as is. we'll assume wales was also a while ago. sorry for the noise i just didnt want to knowingly put a gotcha in there

kevinkreiser (dismissed):

lua/admin.lua
gknisely: @kevinkreiser np.  thanks for the feedback

kevinkreiser (approved):

Issue #3705, opened 2022-08-15
Footway for bikes in Belarus


molind:
I checked state of the [wiki](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions) and [valhalla code](https://github.com/valhalla/valhalla/blob/master/valhalla/mjolnir/adminconstants.h#L43-L46) related to footway permissions in Belarus, it should allow to use footways for bikes instead of the driveways. But it doesn't. In Minsk users explicitly added bicycle=yes on multiple footways (e.g. https://www.openstreetmap.org/way/110906819). But it should be redundant as I understand.

Example: https://valhalla.openstreetmap.de/directions?profile=bicycle&wps=30.3162905%2C53.8753305%2C30.3331368%2C53.8853113

dnesbitt61:
I believe the country specific overrides for Belarus footways are correct. See valhalla/mjolnir/adminconstants.h.
I am able to generate bicycle routes on footways in the area you show - but these have bicycle=yes. Do you have an example where the OSM way is highway=footway without a bicycle=yes tag?

molind:
@dnesbitt61 Yes I've seen and mentioned [valhalla code](https://github.com/valhalla/valhalla/blob/master/valhalla/mjolnir/adminconstants.h#L43-L46) related to footway permissions in Belarus. And example was already in my initial message.

> Example: https://valhalla.openstreetmap.de/directions?profile=bicycle&wps=30.3162905%2C53.8753305%2C30.3331368%2C53.8853113

@kevinkreiser It's more like an issue, not just a question. Because it should allow to drive on footways, but it doesn't allow to do so in Mogilev (which is in Belarus).

kevinkreiser:
@molind i think im not understanding what you are asking for. i had understood you are saying that bikes are NOT using highway=footway when they SHOULD use it. is that what you were saying? when i check the route that you linked i see it taking this way: https://www.openstreetmap.org/way/427635271#map=18/53.87579/30.31477 which is marked as highway=footway. sorry if im misunderstanding.

molind:
@kevinkreiser Sorry if I wasn't clear enough. You got me right. I draw an expected result in blue. Footway should be preferred over highway for bikes in Belarus.
<img width="1192" alt="Screenshot 2022-08-15 at 21 34 55" src="https://user-images.githubusercontent.com/102496/184705326-6689b0a1-210b-4e0b-b489-25ac95da77c7.png">

[...]

Issue #3740, opened 2022-09-11
Elevation Builder Timesout on the Coverage Build


kevinkreiser:
The elevation builder test is easily the longest test of all the tests. On CI it takes over 10 minutes during the coverage build and since it doesnt output anything during that time CI bails on it. A quick fix would be to have cmake somehow output a character every so often to work around it but long term it doesnt make sense to test with such a large extract (building all of utrecht again but with elevation this time..). What we should do is:

1. Convert this test to a gurka test as soon as possible
2. Fix the bug in `get_coord` which it assumes it will return an empty set when a tile doesnt exist but actually since it uses the tilebuilder interface (which will not alert you when you try to open an existing tile that doesnt actually exist) it segfaults trying to access graph nodes which arent there (not sure how CI isnt barfing on this already)

Neyromancer:
i would dive in

nilsnolde:
@Neyromancer you still plan to work on this? Else I'll fix it soonish.

Neyromancer:
@nilsnolde, sure, sorry for delay

nilsnolde:
No worries, just checking:)

PR #3751, opened 2022-09-16
Pedestrian crossings are allowed for bikes.

molind:
# Issue

Few more fixes for #3705 .

I've checked it on Belarus dump and now it works correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
    * FIXED: Fix elevation serializing [#3735](https://github.com/valhalla/valhalla/pull/3735)
    * FIXED: Fix returning a potentially uninitialized value in PointXY::ClosestPoint [#3737](https://github.com/valhalla/valhalla/pull/3737)
    * FIXED: Wales and Scotland name change. [#3746](https://github.com/valhalla/valhalla/pull/3746)
+   * FIXED: Pedestrian crossings are allowed for bikes [#3751](https://github.com/valhalla/valhalla/pull/3751)
 
 * **Enhancement**
    * CHANGED: Pronunciation for names and destinations [#3132](https://github.com/valhalla/valhalla/pull/3132)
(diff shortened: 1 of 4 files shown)


gknisely:
@molind Need to add to the test too.  Did you test locally too?

molind:
Yes, only on local machine. Imported dump of belarus and tried to compute route from the issue.

gknisely:
@molind I can't modify your branch, but I updated the test.  Can you add this to your branch under `valhalla/test/gurka` and delete the old test named `test_admin_sidewalk_override.cc`  Of course please push up the changes.  Thank you.

[test_admin_sidewalk_crossing_override.cc.gz](https://github.com/valhalla/valhalla/files/9616936/test_admin_sidewalk_crossing_override.cc.gz)

gknisely:
@molind I forgot to mention that you need to update the changelog

molind:
More like it's hangup somewhere in `make -C build -j8 coverage`

see bottom task in https://app.circleci.com/pipelines/github/valhalla/valhalla/9482/workflows/4fb8bf4b-7314-4303-b345-c78f7c336608/jobs/54247
`Too long with no output (exceeded 10m0s): context deadline exceeded`

kevinkreiser:
@molind the lint-build-debug one is indeed failing because of a known issue which we can safely ignore: #3740 

whatever the osx build is though we cannot ignore

nilsnolde:
Just realized that the OSX build error was already in `master` and came from the most recent PR at https://github.com/valhalla/valhalla/pull/3746. I guess that went unnoticed with lint/debug failing as well. Could you amend that when you have the time @gknisely ?

kevinkreiser:
That's no good. While we are waiting for the elevation builder test to be fixed I think we should disable some of it so that we don't get in the habit of ignoring build statuses

gknisely:
> Just realized that the OSX build error was already in `master` and came from the most recent PR at #3746. I guess that went unnoticed with lint/debug failing as well. Could you amend that when you have the time @gknisely ?
> 
> EDIT: Win is failing for the same reason as over here: [#3588 (comment)](https://github.com/valhalla/valhalla/pull/3588#issuecomment-1254111359)

@molind you will have to update your test with this small change.  https://github.com/valhalla/valhalla/pull/3757

molind:
@gknisely done.

gknisely (approved):

PR #3752, opened 2022-09-18
Show distance limit in error message

reillywynn:
# Issue

Show distance limit in error message when distance is too high, for example distance limit for bicycle is 500000 meters which is 500 km. It will now display that the max limit is 500 km.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -85,7 +85,7 @@ const std::unordered_map<unsigned, valhalla::valhalla_exception_t> error_codes{
     {151, {151, "Exceeded max time", 400, HTTP_400, OSRM_INVALID_VALUE, "too_large_time"}},
     {152, {152, "Exceeded max contours", 400, HTTP_400, OSRM_INVALID_VALUE, "too_many_contours"}},
     {153, {153, "Too many shape points", 400, HTTP_400, OSRM_INVALID_VALUE, "too_large_shape"}},
-    {154, {154, "Path distance exceeds the max distance limit", 400, HTTP_400, OSRM_DISTANCE_EXCEEDED, "too_large_distance"}},
+    {154, {154, "Path distance exceeds the max distance limit, max distance is " + to_string(max_distance) + " Kilometers.", 400, HTTP_400, OSRM_DISTANCE_EXCEEDED, "too_large_distance"}},
     {155, {155, "Outside the valid walking distance at the beginning or end of a multimodal route", 400, HTTP_400, OSRM_INVALID_URL, "too_large_first_last_walking_distance"}},
     {156, {156, "Outside the valid walking distance between stops of a multimodal route", 400, HTTP_400, OSRM_INVALID_URL, "too_large_in_between_walking_distance"}},
     {157, {157, "Exceeded max avoid locations", 400, HTTP_400, OSRM_INVALID_VALUE, "too_many_avoids"}},


src/worker.cc
```diff
@@ -85,7 +85,7 @@ const std::unordered_map<unsigned, valhalla::valhalla_exception_t> error_codes{
     {151, {151, "Exceeded max time", 400, HTTP_400, OSRM_INVALID_VALUE, "too_large_time"}},
     {152, {152, "Exceeded max contours", 400, HTTP_400, OSRM_INVALID_VALUE, "too_many_contours"}},
     {153, {153, "Too many shape points", 400, HTTP_400, OSRM_INVALID_VALUE, "too_large_shape"}},
-    {154, {154, "Path distance exceeds the max distance limit", 400, HTTP_400, OSRM_DISTANCE_EXCEEDED, "too_large_distance"}},
+    {154, {154, "Path distance exceeds the max distance limit, max distance is " + to_string(max_distance) + " Kilometers.", 400, HTTP_400, OSRM_DISTANCE_EXCEEDED, "too_large_distance"}},
```
nilsnolde, line 88: That can't work, `max_distance` is not defined anywhere for this block of code. `valhalla_exception_t` accepts an `extra` as a string, so every time `154` is raised, we'd need to pass the current `max_distance` within that extra string to the exception.

nilsnolde (changes requested):
Thanks for trying to tackle this.

Best to build the branch with the fix successfully before declaring it fixed ;)

reillywynn:
Sorry about that, I didn't know that I declared it fixed, I'm brand new at open source so I'm not completely sure how to do things. To build it on my end after making a change, I just follow the "Building from Source - Windows" heading and this will let me know if the change I made actually works?

src/worker.cc
reillywynn: valhalla_exception_t(unsigned code, const std::string& extra = "");
So by passing in the max_distance and converting to a string anywhere 154 is raised, this should pass the current max_distance as the variable "extra" so that it can be used in the error message?

kevinkreiser:
yes you have to pass it as the extra parameter, and you'll have to change the tests to match it (loki_service)

kevinkreiser:
honestly im not a huge fan of us changing only a single error message to expose the limit while the bulk of them do not do this. the right thing to do would be either to have them all omit the limit or all show it. would you be game for the latter @reillywynn ? the change is pretty mechanical. find all the places where these types of errors are thrown and `std::to_string` the integer. we might want to reword some of the messages so that they read better. a quick grep shows the following sources have error codes in the 100s, not all of them will have errors that need this treatment though:

```
src/loki/isochrone_action.cc
src/loki/locate_action.cc
src/loki/matrix_action.cc
src/loki/polygon_search.cc
src/loki/route_action.cc
src/loki/status_action.cc
src/loki/trace_route_action.cc
src/loki/transit_available_action.cc
src/loki/worker.cc
src/sif/dynamiccost.cc
src/tyr/actor.cc
src/worker.cc
```

reillywynn:
@kevinkreiser yes, I can work on that. Just to clarify, rather than just changing 154 to show the distance limit, anything that has a max limit of something should be changed. (and reworded if need be)?

kevinkreiser:
@reillywynn yep that's right! Tha KS you!

reillywynn:
@kevinkreiser  I'm having trouble locating where exactly the last portion of the error message is printed when it says "for bicycles" "OSM: Path distance exceeds the max distance limit for bicycle.". The solution I'm working on just adds reads as follows "Path distance exceeds the max distance limit of [max_distance] kilometers". will the "for bikes" populate in there automatically? and also, I think the max distance is in meters correct? If so, would it be appropriate to convert to kilometers before passing the value into valhalla_exception_t or keep it as meters?

kevinkreiser:
regarding the units, everything is in meters in the configs so the messages should match in my opinion.

regarding this blurb:

> "OSM: Path distance exceeds the max distance limit for bicycle."

i dont think our software creates this. that "OSM:" prefix is a clue in my opinion. it looks to me that you are getting this from somewhere else rather than directly from valhalla. its totally possible that the thing you are getting it from wraps the valhalla message in "OSM: " and " for bicycle"

reillywynn:
okay I will keep it in meters then. That makes sense with the error message because I couldn't find it anywhere. It seems like it pulls the error message and then wraps it with that with openStreetMap.

nilsnolde:
> it looks to me that you are getting this from somewhere else rather than directly from valhalla

Jep, our Valhalla web app running on osm.de adds those bits:) (we also do TomTom & HERE, that’s why we specify “OSM”)

nilsnolde:
superseded by #3779

PR #3754, opened 2022-09-21
Bump vcpkg

nilsnolde:
Trying to bump vcpkg version to the latest ["release"](https://github.com/microsoft/vcpkg/releases/tag/2022.08.15) (whatever that means with vpkg..), see discussion at https://github.com/valhalla/valhalla/pull/3588#issuecomment-1254111359.

We could also switch to VS 2022 while we're at it. No idea if it's best practice with VS or we'd have to run both 2019 & 2022 for a while. Back when we switched from 2017 to 2019 we did a hard switch.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -37,7 +37,7 @@ jobs:
       BUILD_CFG: 'Release'
       BUILD_DIR: '$(Agent.WorkFolder)\build'
       VCPKG_DIR: '$(Build.SourcesDirectory)\vcpkg'
-      VCPKG_REF: 'f4bd6423'
+      VCPKG_REF: '2022.08.15'
       TRIPLET: 'x64'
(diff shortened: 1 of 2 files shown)


nilsnolde:
ah crap.. based off of wrong branch.. if win goes through, I'll rebase and force push

nilsnolde:
Damn, CMake still doesn't configure right on Win, can't find `zlib`, same as before the vcpkg version bump. No idea what's the problem.. Nothing changed, it's the same CI system as always, same VS, same commands.. I'll try this myself one of these days exactly as CI tries it. Gotta re-activate Win sooner or later anyways (I moved computers lately).

nilsnolde:
Ah cool, seems like the env vars have done the trick!

kevinkreiser (approved):
win build is passing and you havent changed anything else so this is good to go

nilsnolde:
Can you merge @kevinkreiser , I'm missing the privileges

Issue #3768, opened 2022-09-29
SetDestinations only once for all origins in TimeDistance matrix


nilsnolde:
I didn't really look too closely, it definitely involves some refactor, but nothing too bad.

I think we can squeeze out a bit more performance in TimeDistance matrix if we only do a `SetDestinations` once for a whole request. Currently it's done for every single origin:
https://github.com/valhalla/valhalla/blob/3febdd297c71f8ca7a74577c2b0af4e9a0e90573/src/thor/timedistancematrix.cc#L173-L174

Which entails getting to the tile, the edge, edge cost etc for all edges every destination is close to, which all in all seems a little wasteful. Right now, we remove objects from the destinations in `UpdateDestinations`, which I guess is the reason why we call `SetDestinations` for every origin. My gut feeling is there's other ways to do that which lets us preserve all the expensive info about destinations across all origins.

Didn't profile (but will before tackling it), but I'd think for lots of destinations not too far apart from the origin(s), this should have quite an impact (i.e. when Dijkstra doesn't have too much to do).

Issue #3765, opened 2022-09-28
Add time info to TimeDistance matrix


nilsnolde:
I already mentioned it to @kevinkreiser and want to get broader feedback on it:

I'd like to add time info to the timedistance matrix algo. From a more detailed look into the code, it seems fairly straight-forward, so I don't think we'd need to really discuss a design. And in case of live traffic, performance shouldn't get hurt too much compared to no time info. Of course people will have to put timedistance algo as their matrix algo in the config and that'll make many:many car computations quite a bit slower, but since it's opt-in it's fine IMO.

One design question does come up though: if we include time info for matrix computations, it'd make sense to allow passing one time info object per location. A rare use case I'd imagine, to have individual times for each source, but a valid one I think. Generally not that useful for `/route`, but maybe it would even make sense for `break` waypoints, e.g. when one knows they'll spent a few hours or so at a certain waypoint, they can pass a time object for that location? 

When we parse the locations in `worker.cc`, we already push the date-time info down to the PBF locations so it seems natural to allow this to happen in the request parameters as well:
https://github.com/valhalla/valhalla/blob/57f9657d0c385384aa3a1a79fbdfc1efe73f3ebc/src/worker.cc#L453

The existing date-time parameter can stay top-level (and it makes sense for routing) and we'd need to decide how to deal with both being present, a global date-time object and location-individual ones. That's a bit of an open UX/design question, but I'm happy to outline a few options once you guys are on-board with this proposal in general terms.

PR #3773, opened 2022-10-02
Templatize TimeDistanceMatrix

### LINKED ISSUE #3764: Templatize forward/reverse Dijkstra in TimeDistance matrix
While looking a bit more into the matrix code, I wondered if we couldn't get rid of the whole reverse traversal code in timedistance matrix. I think it's a bit superfluous and would reduce the code to more or less half of what it is now. 

If we have more targets than sources, we can switch them and do forward expansion instead of doing a reverse traversal from the targets (and switch them back before returning them). (or the more elaborate way of templating the traversal functions similar to what was done semi-recently on the uni-directional astar)

Cc @dnesbitt61

nilsnolde:
fixes #3764 

Since I'll be working more with the matrix code soon, I thought this refactoring would be a good primer PR. Since @DmitryAk added a test for the reverse tree in #3738 , I'm fairly confident it's fine.

This is the first of a series of PRs to close a few more issues: #3765 and #3768. I'm not sure there shouldn't be another optimization/issue: base the TimeDistance algo on `Dijkstras`? (didn't really look into it, but should be possible with bit of a rewrite I guess)


src/thor/matrix_action.cc
```diff
@@ -42,10 +43,17 @@ std::string thor_worker_t::matrix(Api& request) {
...
+        forward_search
+            ? time_dist_matrix_forward_.SourceToTarget(options.sources(), options.targets(), *reader,
+                                                       mode_costing, mode,
+                                                       max_matrix_distance.find(costing)->second,
+                                                       options.matrix_locations())
+            : time_dist_matrix_reverse_.SourceToTarget(options.sources(), options.targets(), *reader,
+                                                       mode_costing, mode,
+                                                       max_matrix_distance.find(costing)->second,
+                                                       options.matrix_locations());
+    return time_distances;
```
nilsnolde, line 56: either forward or reverse, depending if there's more sources or targets. I'd like to have said `td_algo = forward_search ? time_dist_matrix_forward_ : time_dist_matrix_reverse_`, but it gave an error about `deleted function`. if anyone has the c++ sugar handy to make that happen (if possible), I'd appreciate it

src/thor/timedistancematrix.cc
```diff
@@ -141,416 +203,217 @@ void TimeDistanceMatrix::ExpandForward(GraphReader& graphreader,
...
+    const sif::travel_mode_t mode,
+    const float max_matrix_distance,
+    const uint32_t matrix_locations) {
+  // Run a series of one to many calls and concatenate the results.
+  const auto& origins = FORWARD ? source_location_list : target_location_list;
+  const auto& destinations = FORWARD ? target_location_list : source_location_list;
+
+  std::vector<TimeDistance> many_to_many(origins.size() * destinations.size());
+  for (size_t origin_index = 0; origin_index < origins.size(); ++origin_index) {
+    const auto& origin = origins.Get(origin_index);
```
nilsnolde, line 226: I removed the `OneToMany` function, there's no use for it anymore

src/thor/timedistancematrix.cc
```diff
@@ -141,416 +203,217 @@ void TimeDistanceMatrix::ExpandForward(GraphReader& graphreader,
...
-    GraphReader& graphreader,
+template std::vector<TimeDistance> TimeDistanceMatrix<ExpansionType::forward>::SourceToTarget(
+    const google::protobuf::RepeatedPtrField<valhalla::Location>& source_location_list,
+    const google::protobuf::RepeatedPtrField<valhalla::Location>& target_location_list,
+    baldr::GraphReader& graphreader,
     const sif::mode_costing_t& mode_costing,
     const sif::travel_mode_t mode,
-    const float max_matrix_distance) {
-  return SourceToTarget(locations, locations, graphreader, mode_costing, mode, max_matrix_distance);
-}
```
nilsnolde, line 400: Also removed the `ManyToMany`, it wasn't ever used and seemed pretty pointless (though it was public API.. as I understand we don't necessarily care about the C++ API while versioning?)

src/thor/timedistancematrix.cc
```diff
@@ -141,416 +203,217 @@ void TimeDistanceMatrix::ExpandForward(GraphReader& graphreader,
...
-                              const uint32_t matrix_locations) {
-  // Set the mode and costing
-  mode_ = mode;
-  costing_ = mode_costing[static_cast<uint32_t>(mode_)];
-  current_cost_threshold_ = GetCostThreshold(max_matrix_distance);
-
-  // Construct adjacency list, edge status, and done set. Set bucket size and
-  // cost range based on DynamicCost. Initialize A* heuristic with 0 cost
-  // factor (needed for setting the origin).
-  astarheuristic_.Init({origin.ll().lng(), origin.ll().lat()}, 0.0f);
```
nilsnolde, line 167: I removed all references to AStarHeuristic, this is a Dijkstra, no need for that

valhalla/thor/timedistancematrix.h
```diff
@@ -113,7 +51,7 @@ class TimeDistanceMatrix {
                  const google::protobuf::RepeatedPtrField<valhalla::Location>& target_location_list,
                  baldr::GraphReader& graphreader,
                  const sif::mode_costing_t& mode_costing,
-                 const sif::TravelMode mode,
+                 const sif::travel_mode_t mode,
```
nilsnolde, line 54: just consistency

src/thor/timedistancematrix.cc
```diff
@@ -84,33 +97,70 @@ void TimeDistanceMatrix::ExpandForward(GraphReader& graphreader,
...
+    if (FORWARD) {
+      if (!costing_->Allowed(directededge, is_dest, pred, tile, edgeid, 0, 0, restriction_idx) ||
+          costing_->Restricted(directededge, pred, edgelabels_, tile, edgeid, true)) {
+        continue;
+      }
+    } else {
+      // TODO: is the Restricted() function correct here passing directed_edge & pred & tile?
+      if (!costing_->AllowedReverse(directededge, pred, opp_edge, t2, opp_edge_id, 0, 0,
+                                    restriction_idx) ||
+          (costing_->Restricted(directededge, pred, edgelabels_, tile, edgeid, FORWARD))) {
```
nilsnolde, line 148: I copied most things from the `ExpandReverse` of course, but I do wonder if this is really correct.. Shouldn't we pass the `opp_edge` and `opp_pred_edge` and `t2` here?

nilsnolde: never mind.. all other reverse trees also work with the forward edge, so I guess this is correct..

src/thor/matrix_action.cc
```diff
@@ -42,10 +43,17 @@ std::string thor_worker_t::matrix(Api& request) {
                                       mode, max_matrix_distance.find(costing)->second);
   };
   auto timedistancematrix = [&]() {
-    return time_distance_matrix_.SourceToTarget(options.sources(), options.targets(), *reader,
-                                                mode_costing, mode,
-                                                max_matrix_distance.find(costing)->second,
-                                                options.matrix_locations());
+    time_distances =
+        forward_search
+            ? time_dist_matrix_forward_.SourceToTarget(options.sources(), options.targets(), *reader,
```
nilsnolde, line 48: hmm.. just realized that this sucks a bit that the caller has to know which version to request.. I'll re-write a bit

src/valhalla_run_matrix.cc
```diff
@@ -244,12 +244,16 @@ int main(int argc, char* argv[]) {
...
-    res = tdm.SourceToTarget(options.sources(), options.targets(), reader, mode_costing, mode,
-                             max_distance);
-    tdm.clear();
+    if (forward_search) {
+      res = TimeDistMatrixForward().SourceToTarget(options.sources(), options.targets(), reader,
+                                                   mode_costing, mode, max_distance);
+    } else {
+      res = TimeDistMatrixReverse().SourceToTarget(options.sources(), options.targets(), reader,
+                                                   mode_costing, mode, max_distance);
+    }
```
nilsnolde, line 256: This is what I mean.. The matrix algo should know which one to apply. I'm not sure how to handle this. I'm not sure how to deal with this, and would appreciate some feedback. AFAICT there's 2 options:
1. keep the templated function and introduce another public function (outside of the TimeDistanceMatrix class) which takes the `options` as argument and returns the right algo based on `sources` & `targets`
2. make the class non-templated and only template the functions which need to know about the direction. Also keep `SourceToTarget` non-templated so it can decide what to do based on the arguments (possibly introduce a protected `ComputeMatrix` or so to do what `SourceToTarget` is doing right now)

Opinions?

src/thor/matrix_action.cc
nilsnolde: never mind, this is a stupid pattern, see https://github.com/valhalla/valhalla/pull/3773#discussion_r985280041

src/valhalla_run_matrix.cc
nilsnolde: I went for 2., cleaner IMO.

valhalla/thor/timedistancematrix.h
```diff
@@ -150,30 +98,21 @@ class TimeDistanceMatrix {
...
+  template <const ExpansionType expansion_direction,
+            const bool FORWARD = expansion_direction == ExpansionType::forward>
+  std::vector<TimeDistance>
+  ComputeMatrix(const google::protobuf::RepeatedPtrField<valhalla::Location>& source_location_list,
+                const google::protobuf::RepeatedPtrField<valhalla::Location>& target_location_list,
+                baldr::GraphReader& graphreader,
+                const sif::mode_costing_t& mode_costing,
+                const sif::travel_mode_t mode,
+                const float max_matrix_distance,
+                const uint32_t matrix_locations = kAllLocations);
```
nilsnolde, line 112: this function's body is now what `SourceToTarget` was before, while `SourceToTarget` now only decides which direction to pick for the template. That way there's absolutely no change to calling functions.

nilsnolde:
This is ready now for review.

valhalla/sif/dynamiccost.h
```diff
@@ -408,7 +408,7 @@ class DynamicCost {
       return next_pred;
     };
     auto reset_edge_status =
-        [&edgestatus, &forward](const std::vector<baldr::GraphId>& edge_ids_in_complex_restriction) {
```
nilsnolde, line 411: I got carried away randomly by a tidy warning, I can revert

nilsnolde:
Ah, and of course the same could be done to `TimeDistanceBSSMatrix`.

nilsnolde:
@kevinkreiser I pushed the bit of less code from last review.

@dnesbitt61 if you wanna have a look too, appreciated:)

nilsnolde:
Frankly, I have no idea why Github thinks rapidjson would be updated. rapidjson is already at `[73063f5](https://github.com/miloyip/rapidjson/compare/b16cec1a1aff158f5002250139528cf8a4b413d8...73063f5002612c6bf64fe24f851cd5cc0d83eef9)`: https://github.com/valhalla/valhalla/tree/master/third_party. Also the two commits it's referencing seems to be pretty much the same.. Weird..

nilsnolde:
WTF Github.. Now there's all those changes of the axle PR in this changeset.. which is already merged and there can't be any changes.. Do you see that as well? :sweat_smile:

I did have an unfortunate force push, but how can it get this messed up?

kevinkreiser (approved):

Issue #3776, opened 2022-10-03
Add matrix action to gurka


nilsnolde:
Also quite a bit of refactor required, but it'd be nice to not be locked in with utrecht_tiles for matrix tests.

PR #3778, opened 2022-10-03
Templatize TimeBSSDistanceMatrix

nilsnolde:
Follow up on #3773 to apply the same treatment to BSS matrix.

Of course we forgot to fix this algo when merging https://github.com/valhalla/valhalla/pull/3738, so the `ManyToOne` response sequence was wrong, yet the tests pass. That's because there was no many:many test for the reverse tree, only a single many:one, which didn't trigger that bug. So, I quickly added one, but even now it's failing. It seems the reverse Matrix Dijkstra is off more compared to the forward BSS AStar than the forward Matrix Dijkstra. Shouldn't they all lead to the exact same results unless there's a bug either in the BSS AStar or the Matrix code? (no time info, no shortcuts.. just seeing that the AStar is using shortcuts.. or better, trying to use shortcuts.. all costing models involved will skip shortcuts inherently in the Allowed() method.. it's just a few lines of code though and can't lead to any bug I guess)

Would you be able to help @xlqian ? I really like that you're comparing the routing results with the matrix results. I'd just like to understand why the matrix trees are "so different" compared to the BSS AStar.


nilsnolde:
It follows the exact same logic as #3738 , so all comments apply here too. I guess once that's merged, this here should be a no-brainer.

nilsnolde:
To be able to reason better about the differences of the matrix Dijkstra vs the undirectional AStar, I think it’d be best if we do #3776 so we can run matrices on controlled gurka maps.

nilsnolde:
I disabled the failing test for the reverse tree for now. It's not specific to this PR, whatever is causing the test to fail was there before too, I'll create a new issue.

kevinkreiser (approved):

PR #3779, opened 2022-10-03
add distance limit in error message for 154 error code

### LINKED ISSUE #3722: Show distance limit in error message
When I attempt to plan a longer route at https://valhalla.openstreetmap.de/ I get the error message:

> OSM: Path distance exceeds the max distance limit for bicycle.

It would be nice if the error message said what the distance limit is.

nilsnolde:
fixes #3722 

@jackrea07 & @reillywynn , in case you want to have a look. Since it's really annoying for the web app to not display this info, I did it quickly myself.

@kevinkreiser, regarding your comment here: https://github.com/valhalla/valhalla/pull/3752#issuecomment-1250877235

In fact, most other errors are already returning the server-side limit. Maybe it'd make sense to return both, the actual limit and the passed value, since the passed value is often not readily comprehensible (e.g. crow distance etc). I went without for now, since most existing errors also simply state the server limit.


src/loki/polygon_search.cc
```diff
@@ -109,7 +109,7 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Ring>& rings_p
     rings_length += bg::perimeter(ring_bg, Haversine());
   }
   if (rings_length > max_length) {
-    throw valhalla_exception_t(167, std::to_string(max_length));
```
nilsnolde, line 112: meter precision is enough IMO:)

src/loki/matrix_action.cc
```diff
@@ -35,7 +35,8 @@ void check_distance(const google::protobuf::RepeatedPtrField<valhalla::Location>
       }
 
       if (path_distance > matrix_max_distance) {
-        throw valhalla_exception_t{154};
+        throw valhalla_exception_t{154, std::to_string(static_cast<size_t>(matrix_max_distance)) +
```
nilsnolde, line 38: is there somewhere a function already that could replace `std::to_string(static_cast<size_t>())`? if not, would it make sense to add one to `midgard.util.h` or so?

src/worker.cc
```diff
@@ -1073,7 +1073,7 @@ valhalla_exception_t::valhalla_exception_t(unsigned code, const std::string& ext
     *this = code_itr->second;
   }
   if (!extra.empty())
-    message += ":" + extra;
+    message += ": " + extra;
```
nilsnolde, line 1076: IMO it's nicer with a space after a colon

kevinkreiser (approved):

nilsnolde:
can you re-approve @kevinkreiser ?

kevinkreiser (approved):

PR #3783, opened 2022-10-06
Potential integer underflow in file suffix generation

### LINKED ISSUE #3780: Unsigned integer underflow in graphtile.cc
We observed a bug involving an underflow of the size_t variable *ind*:
https://github.com/valhalla/valhalla/blob/d83be12170bb536a8c0c4f839d967c2748c1e63c/src/baldr/graphtile.cc#L471

It seems that *ind* can end up being 0 at the time the separator is assigned and is then reduced by 1. Afterwards we try to iterate up to *ind* which is now way larger than the length of the string:
https://github.com/valhalla/valhalla/blob/d83be12170bb536a8c0c4f839d967c2748c1e63c/src/baldr/graphtile.cc#L475

Philipp Voigt (<philipp.voigt@mercedes-benz.com>) on behalf of Mercedes-Benz Tech Innovation GmbH.

PhilippVoigt:
fixes #3780

For certain level configurations the maximum tile ID can be a 6-digit number, e.g. for 512-by-265 tiles. In this case the separator at index 0 is added to the file suffix and the unsigned variable *ind* is reduced by 1 leading to a large positive number which causes an out-of-bounds in the subsequent loop. The default level configuration does not lead to 6-digit tile IDs, therefore the changes were tested solely for our own use cases, which might differ from yours.

Philipp Voigt (<philipp.voigt@mercedes-benz.com>) on behalf of Mercedes-Benz Tech Innovation GmbH.

Licensed under MIT


nilsnolde (dismissed):
LGTM thanks. Only changelog fix left: order is chronologically descending, so newest comes last in the list.

nilsnolde:
> Sorry, beginner's mistake.

No worries, it's not very intuitive, literally every first time contributor to Valhalla does that (which would speak for a PR template overhaul..)

nilsnolde (approved):

PhilippVoigt:
> this is too perfect a test case to not actually add a test for it. i'm going to do so. perhaps we can also simplify the code a bit.

I actually spent some time trying to add a test case but there is no way to reproduce it with the constants returned by `TileHierarchy::levels()`. I wasn't too sure how this can be mocked for a test.

kevinkreiser:
Yeah its quite annoying because of the static use of tileheirarchy. I was able to do it locally though. i'll push an update to your branch with it after i confirm it fails on master

kevinkreiser:
@PhilippVoigt i added a test here but i found that even master successfully passed the test! would you be able to tweak the test to actually test for the case that your code fixes?

PhilippVoigt:
> @PhilippVoigt i added a test here but i found that even master successfully passed the test! would you be able to tweak the test to actually test for the case that your code fixes?

The problematic TileIDs are those that fill all 6 available digits, so with TileID 123456, we can make it fail on master. I added an ASSERT_EXIT to make sure the test actually fails since it otherwise gets interrupted by SIGSEGV and doesn't report it's outcome. Is that fine for you or are there better ways to catch SIGSEGV?

nilsnolde (changes requested):
Pretty much done now. Only remove that one test line and you need to
- merge master after you
- run `pre-commit`: just do a `./script/format.sh`, that'll install pre-commit and run `format.sh` which will fix the formatting

test/graphtile.cc
```diff
@@ -32,6 +32,12 @@ TEST(Graphtile, FileSuffix) {
   EXPECT_THROW(GraphTile::FileSuffix(GraphId(1036800, 2, 0)), std::runtime_error);
   EXPECT_THROW(GraphTile::FileSuffix(GraphId(4050, 0, 0)), std::runtime_error);
   EXPECT_THROW(GraphTile::FileSuffix(GraphId(1036800, 3, 0)), std::runtime_error);
+
+  TileLevel level{7, valhalla::baldr::RoadClass::kSecondary, "half_degree_is_a_multiple_of_3",
+                  Tiles<PointLL>{{{-180, -90}, {180, 90}}, .5, 1}};
+  EXPECT_EQ(GraphTile::FileSuffix(GraphId(1234, 7, 0), ".qux", false, &level), "7/001/234.qux");
+  ASSERT_EXIT((GraphTile::FileSuffix(GraphId(123456, 7, 0), ".qux", false, &level),exit(0)),::testing::ExitedWithCode(0),".*");
```
nilsnolde, line 39: You can remove this one I guess. The point was that the below ID should fail master, we can't verify that anymore on this branch since you fixed it. It's enough for us when you say "jep, the below test fails master". Just verified it myself, you can safely remove this test line.

nilsnolde (approved):
Thanks @PhilippVoigt LGTM. I’ll let Kevin merge.

PhilippVoigt:
@kevinkreiser Is there a chance we could merge this in time for release 3.2.0? We currently depend on this fix and it would help us to base our work on a specific release. Thanks!

nilsnolde:
Yeah we’re just having 2 weekly merge PR meetings. We go chronologically more or less. We won’t release before all the low hanging fruits like this PR aren’t merged, no worries:) I’ll add this one later to the list in the release issue

kevinkreiser (approved):

kevinkreiser (approved):

nilsnolde (approved):

PR #3785, opened 2022-10-07
status endpoint should not answer queries if its shutting down

kevinkreiser:
the point of the status endpoint is to not only give back the user some info about their running system but its also there to negotiate with external processes when a scaling event happens (ie when the process is being told to die with sigterm). back in #3008 we accidentally shifted the check down to the bottom of the function which essentially disables it. @dgearhart mentioned that @ktatterso noticed this, thanks guys for reporting it!


nilsnolde (approved):
Jep makes sense! I’d actually consider this a bug, should we add a changelog entry?

kevinkreiser:
yeah had the same thought will do.

nilsnolde (approved):

PR #3792, opened 2022-10-13
CostMatrix fixes in SetSources and SetTargets

### LINKED ISSUE #3790: No results returned in sources_to_targets for a point in the very center of Dublin
It is a consequence of a weird behavior and of #3605.
When requesting with following options, the result gives `null` and `null` for the distance and time.
```
{
    "costing": "auto",
    "sources": [
        {"lat":53.34978,"lon":-6.26035}
    ],
    "targets": [
         {"lat":53.34978,"lon":-6.26039}
    ]
}
```
_Source and target are a few meters away._

When raising up the max_matrix_distance to a value greater than 2000000, then a result comes back. But with really weird values, as `distance= 1410.183` and `time= 137217` (more than 38 hours).
Even worse, when using the following positions (only longitudes are different for last digit), the same result is returned.
{
    "costing": "auto",
    "sources": [
        {"lat":53.34980053737527,"lon":-6.26029641930372**8**}
    ],
    "targets": [
         {"lat":53.34980053737527,"lon":-6.26029641930372}
    ]
}
And actually, every request I've tried with that source, does return the same result.

dnesbitt61:
In SetSources, only exclude inbound edges if other edges exist. Similarly, in SetTargets only exclude inbound edges if other edges exist. It was possible to not add any edges for a source or target when the only correlated edge was inbound/ outbound.

fixes #3790

# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/costmatrix.cc
+++ b/src/thor/costmatrix.cc
@@ -737,9 +737,9 @@ void CostMatrix::SetSources(GraphReader& graphreader,
     // Only skip inbound edges if we have other options
     bool has_other_edges = false;
     std::for_each(origin.correlation().edges().begin(), origin.correlation().edges().end(),
-                [&has_other_edges](const valhalla::PathEdge& e) {
-                  has_other_edges = has_other_edges || !e.end_node();
-                });
+                  [&has_other_edges](const valhalla::PathEdge& e) {
+                    has_other_edges = has_other_edges || !e.end_node();
+                  });
 
     // Iterate through edges and add to adjacency list
     for (const auto& edge : origin.correlation().edges()) {
@@ -818,9 +818,9 @@ void CostMatrix::SetTargets(baldr::GraphReader& graphreader,
     // Only skip outbound edges if we have other options
     bool has_other_edges = false;
     std::for_each(dest.correlation().edges().begin(), dest.correlation().edges().end(),
-                [&has_other_edges](const valhalla::PathEdge& e) {
-                  has_other_edges = has_other_edges || !e.begin_node();
-                });
+                  [&has_other_edges](const valhalla::PathEdge& e) {
+                    has_other_edges = has_other_edges || !e.begin_node();
+                  });
 
     // Iterate through edges and add to adjacency list
     for (const auto& edge : dest.correlation().edges()) {
--- a/src/thor/costmatrix.cc
+++ b/src/thor/costmatrix.cc
@@ -423,6 +423,7 @@ void CostMatrix::ForwardSearch(const uint32_t index, const uint32_t n, GraphRead
 void CostMatrix::CheckForwardConnections(const uint32_t source,
                                          const BDEdgeLabel& pred,
                                          const uint32_t n) {
+
   // Disallow connections that are part of an uturn on an internal edge
   if (pred.internal_turn() != InternalTurn::kNoTurn) {
     return;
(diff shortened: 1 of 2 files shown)


src/thor/costmatrix.cc
```diff
@@ -734,10 +734,17 @@ void CostMatrix::SetSources(GraphReader& graphreader,
...
+    std::for_each(origin.correlation().edges().begin(), origin.correlation().edges().end(),
+                  [&has_other_edges](const valhalla::PathEdge& e) {
+                    has_other_edges = has_other_edges || !e.end_node();
+                  });
+
     // Iterate through edges and add to adjacency list
     for (const auto& edge : origin.correlation().edges()) {
       // If origin is at a node - skip any inbound edge (dist = 1)
-      if (edge.end_node()) {
+      if (has_other_edges && edge.end_node()) {
```
nilsnolde, line 747: Ah yeah, good point! In fact that is also missing from the `TimeDistanceMatrix` when setting the destinations and the origin for many:one. I added it in #3773 .

nilsnolde (dismissed):
Thanks, that was easier than expected:) Would be nice to add a test case. I think I'll try to add the matrix to gurka now when adding time info to the matrix, so that stuff gets easier.

nilsnolde:
ah and changelog;)

nilsnolde (approved):

nilsnolde:
We recently added matrix support to gurka with a dummy test: https://github.com/valhalla/valhalla/blob/master/test/gurka/test_matrix.cc

Since that test is super trivial, I just wanted to add a slightly more complex one. I then saw that this PR is likely not fixing (at least some of) the trivial-route cases. e.g. this test here will fail current master as well as the commit before merging this branch @dnesbitt61 :
```
#include "gurka.h"
#include <boost/format.hpp>
#include <gtest/gtest.h>

using namespace valhalla;

TEST(Matrix, MatrixSimple) {
  constexpr double gridsize = 100;

  // the grid below is 50 x 100 meters at 36 km/h, i.e. 10 m/s
    A-----B
    |     | \
    |     |  \
    C-----D---E
  )";

  const gurka::ways ways = {
      {"AB", {{"highway", "residential"}}},
      {"AC", {{"highway", "residential"}}},
      {"BD", {{"highway", "residential"}}},
      {"CD", {{"highway", "residential"}}},
      {"DE", {{"highway", "residential"}}},
      {"BE", {{"highway", "residential"}}},
  };

  const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize);
  const auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/matrix_simple");

  // clang-format off
  const std::vector<float> exp_dists = {0.f,  1.2f, 0.3f, 0.9f, 1.1f,
                                        0.6f, 0.f,  0.9f, 0.3f, 0.5f,
                                        0.3f, 0.9f, 0.f,  1.2f, 1.f,
                                        0.9f, 0.3f, 0.6f, 0.f,  0.4f,
                                        1.1f, 0.5f, 1.4f, 0.8f, 0.f};
  // clang-format on
  std::string res;
  const auto result = gurka::do_action(Options::sources_to_targets, map,
                                       {"A", "B", "C", "D", "E"},
                                       {"A", "B", "C", "D", "E"},
                                       "auto", {}, {}, &res);

  rapidjson::Document res_doc;
  res_doc.Parse(res.c_str());

  size_t i = 0;
  for (const auto& origin_row : res_doc["sources_to_targets"].GetArray()) {
    auto origin_td = origin_row.GetArray();
    for (const auto& v : origin_td) {
      std::string msg = "Problem at source " + std::to_string(i / origin_td.Size()) + " and target " +
                        std::to_string(i % origin_td.Size());
      EXPECT_EQ(v.GetObject()["distance"].GetFloat(), exp_dists[i]) << msg;
      i++;
    }
  }
}
``` 

However, it works nicely when selecting `bicycle` (and hence using `TimeDistanceMatrix`).

PR #3793, opened 2022-10-13
add matrix to gurka

### LINKED ISSUE #3776: Add matrix action to gurka

nilsnolde:
fixes #3776 

I think it'll be useful when dealing with adding time info to matrix.

### DIFF
--- a/test/gurka/gurka.cc
+++ b/test/gurka/gurka.cc
@@ -55,35 +55,40 @@ std::vector<midgard::PointLL> to_lls(const nodelayout& nodes,
 /**
  * build a valhalla json request body
  *
- * @param location_type  locations or shape
- * @param waypoints      sequence of pointlls representing the locations
+ * @param location_types locations or shape, sources, targets
+ * @param waypointss     sequence of pointlls representing the locations
  * @param costing        which costing name to use, defaults to auto
  * @param options        overrides parts of the request, supports rapidjson pointer semantics
  * @param stop_type      break, through, via, break_through
  * @return json string
  */
-std::string build_valhalla_request(const std::string& location_type,
-                                   const std::vector<midgard::PointLL>& waypoints,
+std::string build_valhalla_request(const std::vector<std::string>& location_types,
+                                   const std::vector<std::vector<midgard::PointLL>>& waypoints,
                                    const std::string& costing = "auto",
                                    const std::unordered_map<std::string, std::string>& options = {},
                                    const std::string& stop_type = "break") {
+  assert(location_types.size() == waypoints.size());
 
   rapidjson::Document doc;
   doc.SetObject();
   auto& allocator = doc.GetAllocator();
 
-  rapidjson::Value locations(rapidjson::kArrayType);
-  for (const auto& waypoint : waypoints) {
-    rapidjson::Value p(rapidjson::kObjectType);
-    p.AddMember("lon", waypoint.lng(), allocator);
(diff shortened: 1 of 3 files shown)


test/gurka/gurka.cc
```diff
@@ -55,35 +55,40 @@ std::vector<midgard::PointLL> to_lls(const nodelayout& nodes,
...
+ * @param waypoints      all pointll sequences for all location types
  * @param costing        which costing name to use, defaults to auto
  * @param options        overrides parts of the request, supports rapidjson pointer semantics
  * @param stop_type      break, through, via, break_through
  * @return json string
  */
-std::string build_valhalla_request(const std::string& location_type,
-                                   const std::vector<midgard::PointLL>& waypoints,
+std::string build_valhalla_request(const std::vector<std::string>& location_types,
+                                   const std::vector<std::vector<midgard::PointLL>>& waypoints,
```
nilsnolde, line 66: these are now vectors of their previous types as to not have to overload this internal function

test/gurka/gurka.cc
```diff
@@ -678,12 +683,10 @@ valhalla::Api do_action(const valhalla::Options::Action& action,
   std::cerr << "[          ] " << Options_Action_Enum_Name(action)
             << " with mjolnir.tile_dir = " << map.config.get<std::string>("mjolnir.tile_dir")
             << " with locations ";
-  bool first = true;
   for (const auto& waypoint : waypoints) {
-    if (!first)
-      std::cerr << ",";
+    if (waypoint != waypoints.front())
```
nilsnolde, line 687: bit more concise

test/gurka/gurka.cc
```diff
@@ -695,7 +698,48 @@ valhalla::Api do_action(const valhalla::Options::Action& action,
...
+// overload for /sources_to_targets
+valhalla::Api do_action(const valhalla::Options::Action& action,
+                        const map& map,
+                        const std::vector<std::string>& sources,
+                        const std::vector<std::string>& targets,
+                        const std::string& costing,
+                        const std::unordered_map<std::string, std::string>& options,
+                        std::shared_ptr<valhalla::baldr::GraphReader> reader,
+                        std::string* response,
+                        std::string* request_json) {
```
nilsnolde, line 714: another overload, unfortunately. I could've changed the signature of the `do_action` but it's public so I'd have to change every gurka test almost.

test/gurka/test_matrix.cc
```diff
@@ -0,0 +1,46 @@
...
+TEST(MatrixTest, MatrixSimple) {
+  constexpr double gridsize = 100;
+
+  // the grid below is 50 x 100 meters at 36 km/h, i.e. 10 m/s
+  const std::string ascii_map = R"(
+      A----B
+      |    |
+      |    |
+      C----D
+    )";
```
nilsnolde, line 16: Just a small dumb test to show it works. Unfortunately, there's nothing relevant inside the `Api` object that gets returned from `do_action` as we don't track stuff in PBF for matrix. We can only predict the `distance` but that could be fair enough with the right map

kevinkreiser (dismissed):

test/gurka/gurka.cc
kevinkreiser, line 687:
```suggestion
    if (&waypoint != &waypoints.front())
```

kevinkreiser (approved):

Issue #3789, opened 2022-10-12
Few improvements to handling date_time on locations in algorithms


nilsnolde:
This is more a review of how date_time is currently handled for the different actions and some proposals for improvements.

So I went down the rabbit hole to understand the entire flow for date_time objects from input to path finding in preparation of #3765, and had some thoughts. Hope they make sense!

## For `isochrone` & `centroid`

In `Dijkstras::Expand` we could use the location-dependent timeinfos, so each input location's time is tracked for the appropriate branch. To me it looks like this should be pretty easy. So far, we compute all time_infos but only use the first one we find and that one then for all locations:

https://github.com/valhalla/valhalla/blob/7d05fc8f2e4bbefb5b55442a7d0d6dd85b25a52f/src/thor/dijkstras.cc#L366-L370

We could just index into the `time_infos` vector with the `pred.path_id()`. But that's until now only implemented in `centroid`, `isochrone` doesn't track edge labels per input location yet. It might not be a big problem to allow `isochrones` to do the same, and there it would be more useful. So far we **don't** allow `arrive_by` type for isochrones, but we could lift that since we have a reverse Dijkstra for that (and a undocumented `reverse` isochrone parameter). So it'd make sense to use it. But I admit I didn't follow through with investigating more. I kind of think it'd be the same behavior of the Dijkstra as currently, just that the edgelabels/status know about their path_id which is irrelevant as long as we don't use it (and still the costing will be more accurate in case we provide location-wise `date_time` strings). The initial adjacency list, after adding the origin/destination locations, will not change, neither will the algorithm itself, we'd just additionally look for the right time info for each location which will alter the weights somewhat when traffic info is present, but that's exactly what it should do.

For `centroid` I think it's best to just consider all location-dependent `date_time` objects, which is not much different from what is done now, only that we'd in the future allow a forward expansion from each location at a different time.. that's maybe mostly relevant if the location input to `centroid` really lives in different timezones, then they'd technically want to put different `date_time`s for each location. Actually I think I'd almost call it a bug: `Dijkstra::ExpandInner` will get the `offset_time` potentially based on wrong timezone info, the node might be in a different timezone than the `time_info` object it's diff'd to:

https://github.com/valhalla/valhalla/blob/master/src/thor/dijkstras.cc/#L145-L150

I also don't think it'd hurt the `reach` use of Dijkstra, as that doesn't pass time infos.

## Matrix

[...]

PR #3795, opened 2022-10-16
Add time tracking to matrix

### LINKED ISSUE #3765: Add time info to TimeDistance matrix

nilsnolde:
fixes #3765 and parts of #3789 

This PR adds time-dependent routing to the `TimeDistanceMatrix` algo. Will conflict with #3773 a bit, but it's quickly done.

The way it works as of now:
- the server has a new config `max_timedep_distance_matrix` in meters, defaulting to 0
- a user can set `date_time` strings on the `sources` and/or `targets` **or** globally with the `date_time` object
- if a global date_time object is set, for `date_time.value` 0 & 1 we add the `date_time.value` to the `sources` and for 2 we add it to the `targets`
- if in the end there are more sources than targets (i.e. we want to do a reverse algo to save resources) and there's `date_time` strings set on any source, we disable time-dependent routing and fall back on CostMatrix. The other way around is also true: more targets than sources and we want to do a forward expansion we'll disable time and a message to the `warnings` array in the response
- the `matrix` action decides now to take the `timedistancematrix` if the config says `selectoptimal`, it's a motorized vehicle and the right locations have date_time set

TODO:
- [x] tests
- [x] benchmark with `master`, also for unsupported models like bike/ped


scripts/valhalla_build_config
```diff
@@ -512,6 +513,7 @@ help_text = {
         'max_reachability': 'Maximum reachability (number of nodes reachable) allowed on any one location',
         'max_radius': 'Maximum radius in meters allowed on any one location',
         'max_timedep_distance': 'Maximum b-line distance between locations to allow a time-dependent route',
+        'max_timedep_distance_matrix': 'Maximum b-line distance between 2 most distant locations in meters to allow a time-dependent matrix',
```
nilsnolde, line 516: if this is exceeded for a source or target, the whole expansion of that source or target won't track time

src/thor/bidirectional_astar.cc
```diff
@@ -257,6 +257,8 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
     // We can set is_dest incorrectly in the second case, but it is the rare case.
     // The result path will be correct, because there are cosing.Allowed calls inside recost_forward
     // function in second time.
+    // TODO(nils): Should we really check the time restrictions for invariant type? The time is most
+    // likely vastly outdated at this point when not tracking it with invariant
     if (!costing_->Allowed(meta.edge, false, pred, tile, meta.edge_id, localtime,
                            time_info.timezone_index, restriction_idx) ||
         costing_->Restricted(meta.edge, pred, edgelabels_forward_, tile, meta.edge_id, true,
```
nilsnolde, line 264: it seemed wrong to me to check Allowed & Restricted with a valid time along the expansion in the case of an `invariant` time tracking

src/thor/matrix_action.cc
```diff
@@ -42,10 +58,10 @@ std::string thor_worker_t::matrix(Api& request) {
                                       mode, max_matrix_distance.find(costing)->second);
   };
   auto timedistancematrix = [&]() {
-    return time_distance_matrix_.SourceToTarget(options.sources(), options.targets(), *reader,
-                                                mode_costing, mode,
-                                                max_matrix_distance.find(costing)->second,
-                                                options.matrix_locations());
+    return time_distance_matrix_.SourceToTarget(*options.mutable_sources(),
+                                                *options.mutable_targets(), *reader, mode_costing,
+                                                mode, max_matrix_distance.find(costing)->second,
```
nilsnolde, line 63: we need mutable locations here for the TimeInfo object to be created properly with `current` time

src/thor/matrix_action.cc
```diff
@@ -75,7 +91,12 @@ std::string thor_worker_t::matrix(Api& request) {
           time_distances = timedistancematrix();
           break;
         default:
-          time_distances = costmatrix();
+          // force timedistance if traffic is desired and allowed
+          if (has_time(options) && max_timedep_dist_matrix) {
+            time_distances = timedistancematrix();
```
nilsnolde, line 96: this is debatable I guess

src/thor/timedistancematrix.cc
```diff
@@ -67,12 +67,45 @@ void TimeDistanceMatrix::clear() {
...
+    for (const auto& destination : destinations) {
+      time_allowed = max_timedep_dist >
+                     origin_ll.Distance(PointLL(destination.ll().lng(), destination.ll().lat()));
+      if (!time_allowed) {
+        break;
+      }
+    }
+    infos.emplace_back(time_allowed ? TimeInfo::make(origin, reader, &tz_cache_)
+                                    : TimeInfo::invalid());
+  }
```
nilsnolde, line 96: here, we check if we're allowed to track time, based on the max distance between any pair. this sucks a bit.. in worst case this calculates origin_size x target_size distances.. 

the vector of `time_infos` has one element per origin (either source or target), if `max_timedep_dist` for that origin is fine.

instead of this "monster" we could also have a boolean in the config, "allow_time_dependent" matrix or so, and we wouldn't have to mess distances and just respect the normal `max_matrix_distance`

src/thor/timedistancematrix.cc
```diff
@@ -383,50 +425,57 @@ TimeDistanceMatrix::ManyToOne(const valhalla::Location& dest,
...
     const sif::travel_mode_t mode,
     const float max_matrix_distance,
+    const float max_timedep_dist,
     const uint32_t matrix_locations) {
   // Run a series of one to many calls and concatenate the results.
   std::vector<TimeDistance> many_to_many(source_location_list.size() * target_location_list.size());
   if (source_location_list.size() <= target_location_list.size()) {
+    // Get the time information for the origin location
+    auto time_infos =
+        SetTime(source_location_list, target_location_list, graphreader, max_timedep_dist);
```
nilsnolde, line 459: Here we set the time infos once

src/thor/timedistancematrix.cc
```diff
@@ -383,50 +425,57 @@ TimeDistanceMatrix::ManyToOne(const valhalla::Location& dest,
...
     for (size_t source_index = 0; source_index < source_location_list.size(); ++source_index) {
       const auto& origin = source_location_list[source_index];
       std::vector<TimeDistance> td =
           OneToMany(origin, target_location_list, graphreader, mode_costing, mode,
-                    max_matrix_distance, matrix_locations);
+                    max_matrix_distance, time_infos[source_index], matrix_locations);
```
nilsnolde, line 464: here we create the time infos once and pass the right one to the algo

src/thor/worker.cc
```diff
@@ -100,6 +100,7 @@ thor_worker_t::thor_worker_t(const boost::property_tree::ptree& config,
 
   max_timedep_distance =
       config.get<float>("service_limits.max_timedep_distance", kDefaultMaxTimeDependentDistance);
+  max_timedep_dist_matrix = config.get<float>("service_limits.max_timedep_distance_matrix", 0.0f);
```
nilsnolde, line 103: I always make those floats as well, and every time I wonder why not a `uint23_t` :)

src/worker.cc
```diff
@@ -193,6 +193,9 @@ void add_date_to_locations(Options& options,
       default:
         break;
     }
+  } else if (!options.date_time().empty() && !locations.empty()) {
+    for (auto& loc : locations)
+      loc.set_date_time(options.date_time());
```
nilsnolde, line 198: if a global `date_time` is set but nothing on the locations, then let's set them on the locations.

test/matrix.cc
```diff
@@ -236,8 +236,9 @@ TEST(Matrix, test_matrix) {
   EXPECT_EQ(found, 10) << " not the number of results as expected";
 
   TimeDistanceMatrix timedist_matrix;
-  results = timedist_matrix.SourceToTarget(request.options().sources(), request.options().targets(),
-                                           reader, mode_costing, sif::TravelMode::kDrive, 400000.0);
+  results = timedist_matrix.SourceToTarget(*request.mutable_options()->mutable_sources(),
```
nilsnolde, line 239: just switching to mutable objects in this file

docs/api/matrix/api-reference.md
```diff
@@ -31,9 +31,6 @@ For example, while at your office, you want to know the times and distances to w
 {"sources":[{"lat":40.744014,"lon":-73.990508},{"lat":40.739735,"lon":-73.979713},{"lat":40.752522,"lon":-73.985015},{"lat":40.750117,"lon":-73.983704},{"lat":40.750552,"lon":-73.993519}],"targets":[{"lat":40.744014,"lon":-73.990508},{"lat":40.739735,"lon":-73.979713},{"lat":40.752522,"lon":-73.985015},{"lat":40.750117,"lon":-73.983704},{"lat":40.750552,"lon":-73.993519}],"costing":"pedestrian"}&id=ManyToMany_NYC_work_dinner
 ```
 
-For one-to-many or many-to-one requests there is an option to specify the minimum number of locations that satisfy the request. The `matrix_locations` parameters defaults to all locations. However, when specified, this option allows a partial result to be returned. This is basically equivalent to "find the closest/best `matrix_locations` locations out of the full location set".
-There is an option to name your matrix request.  You can do this by appending the following to your request `&id=`.  The `id` is returned with the response so a user could match to the corresponding request.
-
```
nilsnolde, line 36: I put `matrix_locations` into the table below.

docs/api/matrix/api-reference.md
```diff
@@ -56,6 +57,9 @@ The Time-Distance Matrix service uses the `auto`, `bicycle`, `pedestrian` and `b
 | Options | Description |
 | :------------------ | :----------- |
 | `id` | Name your matrix request. If `id` is specified, the naming will be sent thru to the response. |
+| `matrix_locations` | For one-to-many or many-to-one requests this specifies the minimum number of locations that satisfy the request. However, when specified, this option allows a partial result to be returned. This is basically equivalent to "find the closest/best `matrix_locations` locations out of the full location set". |
+| `date_time` | The local date and time at the location. <li>`value` - the date and time specified in ISO 8601 format (YYYY-MM-DDThh:mm) in the local time zone of departure or arrival. For example, "2016-07-03T08:06"</li></ul> **Note**, `type` is not relevant for matrix.|
+
```
nilsnolde, line 62: we only need `date_time.value` for matrix, `type` is irrelevant

src/worker.cc
```diff
@@ -175,8 +175,8 @@ rapidjson::Document from_string(const std::string& json, const valhalla_exceptio
 
 void add_date_to_locations(Options& options,
                            google::protobuf::RepeatedPtrField<valhalla::Location>& locations) {
-  // otherwise we do what the person was asking for
-  if (options.has_date_time_case() && !locations.empty()) {
+  // /route needs special treatment
+  if (options.action() == Options::route && options.has_date_time_case() && !locations.empty()) {
     switch (options.date_time_type()) {
```
nilsnolde, line 180: `date_time.type` is only relevant for routing, all other services should default all locations to the global `date_time.value`

nilsnolde:
hmpf, `Mapmatch.test_edges_discontinuity_with_multi_routes` fails with e.g. 
```
/home/circleci/project/test/mapmatch.cc:479: Failure
Value of: std::get<2>(test_answers[i])
  Actual: false
Expected: true
Found a leg with a start time when it shouldnt have had one
```

I guess setting times on all locations by default messed smth up with map-matching. Will investigate tmrw.

test/gurka/test_matrix.cc
```diff
@@ -41,8 +76,46 @@ TEST(MatrixTest, MatrixSimple) {
...
+
+  size_t i = 0;
+  for (const auto& origin_row : res_doc["sources_to_targets"].GetArray()) {
+    auto origin_td = origin_row.GetArray();
+    for (const auto& v : origin_td) {
+      std::string msg = "Problem at source " + std::to_string(i / origin_td.Size()) + " and target " +
+                        std::to_string(i % origin_td.Size());
+      EXPECT_EQ(v.GetObject()["distance"].GetFloat(), exp_dists[i]) << msg;
       i++;
     }
```
nilsnolde, line 120: can put this in a function

src/worker.cc
```diff
@@ -175,8 +175,10 @@ rapidjson::Document from_string(const std::string& json, const valhalla_exceptio
 
 void add_date_to_locations(Options& options,
                            google::protobuf::RepeatedPtrField<valhalla::Location>& locations) {
-  // otherwise we do what the person was asking for
-  if (options.has_date_time_case() && !locations.empty()) {
+  // /(trace_)route needs special treatment
+  if (options.has_date_time_case() &&
+      (options.action() == Options::route || options.action() == Options::trace_route) &&
```
nilsnolde, line 180: I wasn't aware that `trace_route` cares about `date_time.type` settings.. though I see it doesn't do anything with type = 2 (arrive_by), which wouldn't make much sense. `trace_attributes` probably shouldn't get the global `date_time` string either if it's set, that hardly makes sense I think?

nilsnolde:
this should be good for review now

src/thor/timedistancematrix.cc
kevinkreiser: we already check all pairs distances in loki and we always do the service limits checks there anyway. can we move this to loki? it would be nice if all this did was initialize the timeinfo stuff. in loki we can unset the time stuff if the service constraint fails and then here we'll see there is no time to set anyway.

src/worker.cc
kevinkreiser: > which wouldn't make much sense
are you trying to say that it seems correct to you that it doesnt handle `arrive_by`? or are you saying it doesnt seem right to you that it doesnt handle it?

why are we changing this logic anyway? what was wrong with just setting it regardless of the action? sure some actions dont support it but they are blocked at other parts in the request processing

src/worker.cc
kevinkreiser: yeah this is strange i guess. the reasons we have a spot for date time on each location is so that for multipoint routes we can propogate the time at the end of each leg over the course of the route. essentially the date time on the location is an output field not an input field. now that we want to do it with matrix i can see you want per location, ie to change it to an input field. we should consider how this works and possible strive a bit to unconfuse it (maybe we deprecate the global one with a warning and ask that you always put it on your first or last location depending on what you want? i dont know didnt think about it much just yet (well since a year or something the last time i spent brain cells on it)

src/thor/timedistancematrix.cc
nilsnolde: > we already check all pairs distances in loki

oops, I didn't even look :sweat_smile: I just saw how the matrix uses `max_matrix_distance` where it aborts the expansion if it's exceeded and thought that's the only way we use it, which did seem weird haha good point!

src/worker.cc
nilsnolde: I guess that's answered by your comment below. Essentially `date_time.type` really only has a meaning when a route is requested with an origin & destination. Then we want to force set a time on first/last location or it's invariant (for bidir astar).

src/worker.cc
nilsnolde: Well, I don't think it's so strange once I thought about it a bit more. It holds true for any other action as well: any input location can start at/arrive by a different time. I think that makes perfect sense. That also goes for isochrones & centroid and maybe even route (see [here](https://github.com/valhalla/valhalla/issues/3789) for more review). I think the global `date_time` is still useful to shortcut setting date_time for e.g. `/route`. There the user would have to set a date_time on the first/last location if there was no global date_time object. It just seems easier UX. While it might seem confusing that you can do both, the code currently allows for it as well. This function is only called if there is no time on any location set, so even now users can do that.

src/worker.cc
nilsnolde: > are you trying to say that it seems correct to you that it doesnt handle arrive_by?

right, this one:) sorry, was a bit unclear, I see that now..

src/worker.cc
nilsnolde: > so that for multipoint routes we can propogate the time at the end of each leg over the course of the route.

btw, is anything speaking against allowing a time being set on a location e.g. if it's a `break`? not that it's toooo useful, but could be. wouldn't hurt performance AFAICT

nilsnolde: hm, what I currently fail to see: how do the tests even work.. I can't see how this would allow a `current` time for e.g. matrix. damn, didn't really think about that.. so the date_time.type is important after all. just in the case we want to set it for all locations, not just the first one..

I tried it in a different which fails the map-matcher tests. Let's talk about it tmrw if we have time.

src/worker.cc
```diff
@@ -175,22 +175,16 @@ rapidjson::Document from_string(const std::string& json, const valhalla_exceptio
...
+          loc.set_date_time("current");
         break;
-      case Options::invariant:
+      default:
         for (auto& loc : locations)
           loc.set_date_time(options.date_time());
-      default:
         break;
     }
   }
```
nilsnolde, line 190: I don't think it'd matter much for `route` if we set the date_time for all locations, no matter the `date_time.type`, AFAICT it's always implicit by the algorithm how a location's date_time is interpreted. The only one I'm unsure about is `trace_route`

scripts/list_tiles.py
```diff
@@ -72,8 +72,9 @@ def Col(self, x):
 
     def Digits(self, number):
         digits = 1 if (number < 0) else 0
+        number = number
         while number:
-            number /= 10
+            number = int(number / 10)
```
nilsnolde, line 77: I needed that in another context. And realized it's not python3 ported yet. python2 would've output an int by default, python3 outputs a float by default if it needs to.

nilsnolde:
So, some test runs with `valhalla_run_matrix -m 5` on this branch and master. It was 63 locations in Berlin with TomTom historical data, so quite dense coverage:
<details><summary>request</summary>
<pre>
{ "costing": "auto", "id": 1, "sources": [ { "lon": 13.378117, "lat": 52.522983 }, { "lon": 13.39204, "lat": 52.526599 }, { "lon": 13.419038, "lat": 52.523396 }, { "lon": 13.423283, "lat": 52.517094 }, { "lon": 13.421585, "lat": 52.506967 }, { "lon": 13.412755, "lat": 52.500559 }, { "lon": 13.400869, "lat": 52.499008 }, { "lon": 13.390172, "lat": 52.498698 }, { "lon": 13.378626, "lat": 52.499628 }, { "lon": 13.368268, "lat": 52.503143 }, { "lon": 13.364363, "lat": 52.508104 }, { "lon": 13.363174, "lat": 52.514614 }, { "lon": 13.36708, "lat": 52.52102 }, { "lon": 13.373192, "lat": 52.526496 }, { "lon": 13.381003, "lat": 52.528975 }, { "lon": 13.391021, "lat": 52.529801 }, { "lon": 13.399851, "lat": 52.530524 }, { "lon": 13.412246, "lat": 52.530421 }, { "lon": 13.421755, "lat": 52.528148 }, { "lon": 13.428207, "lat": 52.5235 }, { "lon": 13.429905, "lat": 52.518644 }, { "lon": 13.431942, "lat": 52.513684 }, { "lon": 13.430584, "lat": 52.507277 }, { "lon": 13.430414, "lat": 52.503246 }, { "lon": 13.426679, "lat": 52.499008 } ], "targets": [ { "lon": 13.378117, "lat": 52.522983 }, { "lon": 13.39204, "lat": 52.526599 }, { "lon": 13.419038, "lat": 52.523396 }, { "lon": 13.423283, "lat": 52.517094 }, { "lon": 13.421585, "lat": 52.506967 }, { "lon": 13.412755, "lat": 52.500559 }, { "lon": 13.400869, "lat": 52.499008 }, { "lon": 13.390172, "lat": 52.498698 }, { "lon": 13.378626, "lat": 52.499628 }, { "lon": 13.368268, "lat": 52.503143 }, { "lon": 13.364363, "lat": 52.508104 }, { "lon": 13.363174, "lat": 52.514614 }, { "lon": 13.36708, "lat": 52.52102 }, { "lon": 13.373192, "lat": 52.526496 }, { "lon": 13.381003, "lat": 52.528975 }, { "lon": 13.391021, "lat": 52.529801 }, { "lon": 13.399851, "lat": 52.530524 }, { "lon": 13.412246, "lat": 52.530421 }, { "lon": 13.421755, "lat": 52.528148 }, { "lon": 13.428207, "lat": 52.5235 }, { "lon": 13.429905, "lat": 52.518644 }, { "lon": 13.431942, "lat": 52.513684 }, { "lon": 13.430584, "lat": 52.507277 }, { "lon": 13.430414, "lat": 52.503246 }, { "lon": 13.426679, "lat": 52.499008 } ]}
</pre>
</details>

When `date_time` was passed, the following settings were taken: `"date_time": { "type": 1, "value": "2022-10-21T09:00" }`.

### Results

| Branch | Mode | Time? | TimeDistance avg. | Cost avg.  |
|--------|------|-------|-------------------|------------|
| PR     | auto | 𐄂     | 3.69 secs         | 0.97 secs  |
| PR     | bike | 𐄂     | 1.82 secs         | 0.69 secs  |
| master | auto | 𐄂     | 3.60 secs         | 0.93 secs  |
| master | bike | 𐄂     | 1.77 secs         | 0.657 secs |
| PR     | auto | ✓     | 4.58 secs         | 0.97 secs  |
| PR     | bike | ✓     | 1.86 secs         | 0.69 secs  |

Timing went a few % down in this PR, but this was just one run each. Adding traffic is not even that crazy much longer. The difference bike/auto is stronger. Is that only the `GetSpeed` function?

src/worker.cc
```diff
@@ -1111,6 +1146,16 @@ valhalla_exception_t::valhalla_exception_t(unsigned code, const std::string& ext
...
+// function to add warnings to proto info object
+void add_warning(valhalla::Api& api, unsigned code) {
+  auto message = warning_codes.find(code);
+  if (message != warning_codes.end()) {
+    auto* warning = api.mutable_info()->mutable_warnings()->Add();
+    warning->set_description(message->second);
+    warning->set_code(message->first);
+  }
+}
+
```
nilsnolde, line 1158: we made a mistake when adding warnings to the response in #3588 : `add_warnings` wasn't in the `valhalla` namespace, so the linker didn't know where to find the implementation. it didn't hit us before bcs we were only ever using it in this file, but now we also use it in other files.

src/worker.cc
```diff
@@ -145,19 +145,11 @@ const std::unordered_map<int, std::string> warning_codes = {
...
 };
-
-// function to add warnings to proto info object
-void add_warning(valhalla::Api& api, int code) {
-  auto message = warning_codes.find(code);
-  if (message != warning_codes.end()) {
-    auto* warning = api.mutable_info()->mutable_warnings()->Add();
-    warning->set_description(message->second);
-    warning->set_code(message->first);
-  }
```
nilsnolde, line 158: moved into the `valhalla` namespace where it belongs

src/worker.cc
```diff
@@ -173,27 +165,54 @@ rapidjson::Document from_string(const std::string& json, const valhalla_exceptio
...
+      switch (options.date_time_type()) {
+        case Options::current:
+        case Options::depart_at:
+          if (node == "sources") {
+            for (auto& loc : locations) {
+              loc.set_date_time(options.date_time_type() == Options::current ? "current"
+                                                                             : options.date_time());
+            }
+          }
+          break;
```
nilsnolde, line 199: this will eventually also be interesting for isochrones & centroid, not only matrix.

src/worker.cc
```diff
@@ -173,27 +165,54 @@ rapidjson::Document from_string(const std::string& json, const valhalla_exceptio
...
+        default:
+          break;
+      }
     }
   }
+
+  return std::find_if(locations.begin(), locations.end(), [](valhalla::Location loc) {
+           return !loc.date_time().empty();
+         }) != locations.end();
```
nilsnolde, line 215: return true if any location has a date_time in the end

test/gurka/test_matrix_time_dependent.cc
```diff
@@ -0,0 +1,227 @@
...
+  // reverse tree, no time_info
+  res.erase();
+  result = gurka::do_action(Options::sources_to_targets, allowed_map, {"E", "H"}, {"H"}, "auto",
+                            options, reader, &res);
+  res_doc.Parse(res.c_str());
+  // without time, we'd expect to take the longer, faster route
+  // TODO: be aware that it's actually 2.0f, but CostMatrix has a bug here
+  check_dist(res_doc, {1.6f, 0.0f});
+  ASSERT_EQ(result.info().warnings().size(), 1);
+  ASSERT_EQ(result.info().warnings().Get(0).code(), 401);
```
nilsnolde, line 145: here we have two origins with date_time but perform a reverse search, so we can't respect it. in that case it'll fall back to CostMatrix which should actually take the faster and longer route, but it doesn't. thankfully I've seen it in other tests already that there's a bug, e.g. https://github.com/valhalla/valhalla/pull/3792#issuecomment-1279963888. ref to issue: https://github.com/valhalla/valhalla/issues/3803

src/thor/timedistancematrix.cc
```diff
@@ -111,12 +118,11 @@ void TimeDistanceMatrix::Expand(GraphReader& graphreader,
...
     if (!FORWARD) {
       // Get opposing edge Id and end node tile
-      graph_tile_ptr t2 =
-          directededge->leaves_tile() ? graphreader.GetGraphTile(directededge->endnode()) : tile;
+      t2 = directededge->leaves_tile() ? graphreader.GetGraphTile(directededge->endnode()) : tile;
       if (t2 == nullptr) {
         continue;
       }
-      GraphId opp_edge_id = t2->GetOpposingEdgeId(directededge);
+      opp_edge_id = t2->GetOpposingEdgeId(directededge);
```
nilsnolde, line 125: bug.. was redeclaring t2 & opp_edge_id

src/thor/timedistancematrix.cc
```diff
@@ -387,23 +398,10 @@ void TimeDistanceMatrix::SetDestinations(
...
-    bool has_other_edges = false;
-    std::for_each(loc.correlation().edges().begin(), loc.correlation().edges().end(),
-                  [&has_other_edges](const valhalla::PathEdge& e) {
-                    has_other_edges = has_other_edges || (FORWARD ? !e.begin_node() : !e.end_node());
-                  });
     for (const auto& edge : loc.correlation().edges()) {
-      // If destination is at a node skip any outbound edges
-      if (has_other_edges && (FORWARD ? edge.begin_node() : edge.end_node())) {
-        continue;
-      }
```
nilsnolde, line 401: This was also a bug in the recent templatization.. I kind of forgot there to test TimeDistanceMatrix for trivial cases. And now it showed. We have to include the outbound edges regardless if we have alternatives or not. The `UpdateDestinations` won't be triggered otherwise when source & target is the exact same on line 252. 

FWIW, before templatizing this whole thing there wasn't any check like this on the destinations either. I included it mindlessly without really thinking about it.. Also added a test to see if we're still properly resolving trivial cases, we do:)

test/gurka/test_matrix_time_dependent.cc
```diff
@@ -0,0 +1,235 @@
...
+  // forward tree, date_time on the locations, 2nd location has pointless date_time
+  options = {{"/sources/0/date_time", "current"},
+             {"/sources/1/date_time", "2016-07-03T08:06"},
+             {"/costing_options/auto/speed_types/0", "current"}};
+  res.erase();
+  result = gurka::do_action(Options::sources_to_targets, map, {"E", "H"}, {"E", "H"}, "auto", options,
+                            nullptr, &res);
+  res_doc.Parse(res.c_str());
+  // the second origin can't respect time (no historical data)
+  check_dist(res_doc, {0.0f, 1.6f, 2.0f, 0.0f});
```
nilsnolde, line 126: I really like this test. It proves that date_time is location-dependent

src/thor/timedistancematrix.cc
```diff
@@ -182,23 +191,24 @@ void TimeDistanceMatrix::Expand(GraphReader& graphreader,
...
+    google::protobuf::RepeatedPtrField<valhalla::Location>& source_location_list,
+    google::protobuf::RepeatedPtrField<valhalla::Location>& target_location_list,
     baldr::GraphReader& graphreader,
     const float max_matrix_distance,
     const uint32_t matrix_locations) {
   // Run a series of one to many calls and concatenate the results.
-  const auto& origins = FORWARD ? source_location_list : target_location_list;
-  const auto& destinations = FORWARD ? target_location_list : source_location_list;
+  auto& origins = FORWARD ? source_location_list : target_location_list;
+  auto& destinations = FORWARD ? target_location_list : source_location_list;
```
nilsnolde, line 208: these two lines can be removed. instead `SourceToTarget` can pass in the right order itself

nilsnolde:
@dnesbitt61 it's already drowning a bit, but I did some perf comparison here: https://github.com/valhalla/valhalla/pull/3795#issuecomment-1287895054

Also realized I had a few bugs in the recent templatization. If you're interested, Kevin and me will go over the changes in detail on Wednesday at 3 pm your time: https://next.gis-ops.com/call/cyuox9by

kevinkreiser:
@nilsnolde isnt it at 9:15 am? @dnesbitt61 is in my timezone last time i checked :smile:

dnesbitt61:
ok, I cannot make that time. All booked up Wed morning

nilsnolde:
We'll also meet on Friday same time if that fits you better @dnesbitt61 . On Wednesday we could then cut the release after merging the remaining low-hanging fruits. You're right @kevinkreiser , let's do it before adding this PR, then people have the chance to test the feature before it ends up in a release.

dnesbitt61:
Friday works - send me an invite!

nilsnolde:
Sure, for asking so nicely I'll invite you to the bi-weekly hehe :stuck_out_tongue_winking_eye:

nilsnolde:
That's true. That'd be a quick win. That's mostly a performance optimization right? As I understand it, it's just doing a few too many rounds in rare-ish situations when not_thru edges are involved no? Eventually doing a u-turn out of the not_thru region and maybe even taking another path than the other branch, and hopefully (but maybe possibly?) not ending up being the best path to the destination, though I can see how that could be possible. That should be happening fairly rarely, if at all? Anyways, good point and not much effort.

nilsnolde:
It does even set the closure pruning in the labels but never prunes the search using it. Never mind the pred/edge's not_thru flags. Adding it to #3804 .

dnesbitt61:
An example where not_thru impacts the result:

http://localhost:8002/sources_to_targets?json={"sources":[{"lat":39.654713,"lon":-76.006799,"heading":0}],"targets":[{"lat":39.647804,"lon":-76.033476,"heading":310},{"lat":39.650894,"lon":-76.038346,"heading":310}],"costing":"bicycle","costing_options":{"bicycle":{"service_penalty":600}}}

vs.

http://localhost:8002/route?json={"locations":[{"lat":39.654713,"lon":-76.006799,"heading":0},{"lat":39.650894,"lon":-76.038346,"heading":310}],"costing":"bicycle","costing_options":{"bicycle":{"service_penalty":600}}}

With TimeDistanceMatrix it will do a Uturn in a cul-de-sac. This may be a bit contrived since a heading is used on the origin location, but this case has uses for navigation / rerouting.

dnesbitt61:
Pruning not_thru is mostly a performance optimization. Probably the reason it isn't used in TimeDistanceMatrix is for cases where a target location requires traversing a not_thru edge - e.g. the location is in a neighborhood with only 1 entrance/exit. Bidirectional route (and CostMatrix) cases work here since the reverse traversal will exit the not_thru region and meet with the forward traversal. Not sure how this could be supported in forward Dijkstra and TimeDistanceMatrix methods.

nilsnolde:
> With TimeDistanceMatrix it will do a Uturn in a cul-de-sac.

Yeah I see, I didn't think about cul-de-sacs, that's really not optimal. I looked at the code yesterday, but already forgot some of it.. does an edge when it's inside a not_thru region? Or is only the edge leading into a region having not_thru set? If the former, we could look at all correlated edges of all destinations, and if one of them is inside a not_thru region we could turn off pruning. For closures that might be harder, as they're dynamic. Currently we're not pruning there either in TD. I guess the unidirectional A* has some hints as how to deal with that.

nilsnolde:
Nice example, definitely not used to a landscape of cul-de-sacs, too little space over here :sweat_smile: 

Had a quick look into how not_thru edges are determined at build time. I'm still lacking a proper deep dive but: it looks pretty easy to mark the ones inside a not_thru region, just a few lines of code. Not sure if there's any downsides, maybe you do off the top of your head? In my mind that's smth that would benefit all unidirectional algos.. Unidirectional AStar also doesn't prune them. Maybe there's a good reason I can't think of.. Other than not being easily able to determine if a destination's correlated edges are inside a not_thru region.

nilsnolde:
I can't let this go somehow.. In bidir A* it seems to also only care about edges which are currently set as not_thru:

https://github.com/gis-ops/valhalla/blob/2b208b8c60bfefa5fc7b1f72fe34da5eef5f3f0e/src/thor/bidirectional_astar.cc#L1115-L1117

With the current logic, it would miss the edges contained inside a not_thru region and abort the search early. But I think that'd need a bit more of adjustment to the algo to support what I'm thinking about.

src/thor/matrix_action.cc
```diff
@@ -16,9 +16,39 @@ using namespace valhalla::thor;
...
+      }
+      had_valid_time = true;
+      break;
+    }
+  }
+
+  return had_valid_time;
 }
 
+constexpr double kMilePerMeter = 0.000621371;
```
nilsnolde, line 49: move to midgard/constants.h

src/thor/timedistancematrix.cc
```diff
@@ -86,6 +117,14 @@ void TimeDistanceMatrix::Expand(GraphReader& graphreader,
     return;
   }
 
+  // will be updated along the expansion
+  time_info_tracked =
+      from_transition ? time_info_tracked
+                      : (FORWARD ? time_info_tracked.forward(invariant ? 0.f : pred.cost().secs,
+                                                             static_cast<int>(nodeinfo->timezone()))
+                                 : time_info_tracked.reverse(invariant ? 0.f : pred.cost().secs,
+                                                             static_cast<int>(nodeinfo->timezone())));
```
nilsnolde, line 126: handling invariant date_time

src/thor/timedistancematrix.cc
```diff
@@ -541,10 +583,14 @@ bool TimeDistanceMatrix::UpdateDestinations(
...
-std::vector<TimeDistance> TimeDistanceMatrix::FormTimeDistanceMatrix() {
+std::vector<TimeDistance> TimeDistanceMatrix::FormTimeDistanceMatrix(GraphReader& reader,
+                                                                     const std::string origin_dt,
+                                                                     const uint64_t origin_tz,
+                                                                     const GraphId& pred_id) {
   std::vector<TimeDistance> td;
   for (auto& dest : destinations_) {
-    td.emplace_back(dest.best_cost.secs, dest.distance);
+    auto date_time = GetDateTime(origin_dt, origin_tz, pred_id, reader, dest.best_cost.secs);
+    td.emplace_back(dest.best_cost.secs, dest.distance, date_time);
```
nilsnolde, line 593: here I try to add the final `date_time` to the results here. few things worth noting:
1. I have historical traffic on some dataset I generated ages ago, but somehow all node_info's along the expansion have timezone_index of 0 (i.e. invalid), no idea how that can be.. the origin location has 297 (correct)
2. we only output `date_time` in the response if there actually way time tracking going on
3. the results are NOT consistent right now (i.e. `date_time` field is off compared to request and `time` field, e.g. this request has been done with `2022-10-28T09:00` and outputs:
```
"sources_to_targets": [
		[
			{
				"date_time": "2022-10-28T08:59",
				"distance": 0.000,
				"time": 0,
				"to_index": 0,
				"from_index": 0
			},
			{
				"date_time": "2022-10-28T09:01",
				"distance": 0.729,
				"time": 126,
				"to_index": 1,
				"from_index": 0
			},
			{
				"date_time": "2022-10-28T09:06",
				"distance": 2.521,
				"time": 349,
				"to_index": 2,
				"from_index": 0
			}
		],
		[
			{
				"date_time": "2022-10-28T09:01",
				"distance": 0.720,
				"time": 151,
				"to_index": 0,
				"from_index": 1
			},
			{
				"date_time": "2022-10-28T08:59",
				"distance": 0.000,
				"time": 0,
				"to_index": 1,
				"from_index": 1
			},
			{
				"date_time": "2022-10-28T09:06",
				"distance": 2.678,
				"time": 391,
				"to_index": 2,
				"from_index": 1
			}
		],
		[
			{
				"date_time": "2022-10-28T09:04",
				"distance": 2.579,
				"time": 319,
				"to_index": 0,
				"from_index": 2
			},
			{
				"date_time": "2022-10-28T09:06",
				"distance": 2.670,
				"time": 392,
				"to_index": 1,
				"from_index": 2
			},
			{
				"date_time": "2022-10-28T08:59",
				"distance": 0.000,
				"time": 0,
				"to_index": 2,
				"from_index": 2
			}
		]
	]
```

src/thor/timedistancematrix.cc
```diff
@@ -22,6 +22,35 @@ static bool IsTrivial(const uint64_t& edgeid,
...
+  else if (const auto* node = reader.nodeinfo(out_nodes.second, tile))
+    dest_tz = node->timezone();
+
+  auto in_epoch =
+      DateTime::seconds_since_epoch(origin_dt, DateTime::get_tz_db().from_index(origin_tz));
+  uint64_t out_epoch = in_epoch + offset;
+  // TODO: how can it be that node.time_info() is 0?
+  // Also for the same origin/destination, the response date_time is origin.date_time() - 1 minute?!
+  dest_tz = dest_tz || origin_tz;
+  std::string out_dt = DateTime::seconds_to_date(static_cast<uint64_t>(out_epoch),
```
nilsnolde, line 49: this should not return "" for 0 timezone index, e.g. without timezone db

src/baldr/datetime.cc
```diff
@@ -151,16 +151,8 @@ int timezone_diff(const uint64_t seconds,
...
-    const auto& dest_info = from_cache(dest, dest_tz, *cache);
-    return static_cast<int>(
-        std::chrono::duration_cast<std::chrono::seconds>(dest_info.offset - origin_info.offset)
-            .count());
-  }
-
-  const auto& origin_info = origin.get_info();
-  const auto& dest_info = dest.get_info();
+  const auto& origin_info = cache ? from_cache(origin, origin_tz, *cache) : origin.get_info();
+  const auto& dest_info = cache ? from_cache(dest, dest_tz, *cache) : dest.get_info();
```
nilsnolde, line 155: pure refactor to get rid of a few lines

src/thor/bidirectional_astar.cc
nilsnolde: more a personal preference sort of thing apparently

nilsnolde:
this is finally done!

kevinkreiser (approved):

.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
         files: ^(src/bindings/python/|scripts/)
 
   # Python linter
-  - repo: https://gitlab.com/pycqa/flake8
+  - repo: https://github.com/pycqa/flake8
```
nilsnolde, line 21: @kevinkreiser this will fix CI in master

CHANGELOG.md
```diff
@@ -139,8 +140,7 @@
    * CHANGED: Templatized TimeDistanceMatrix for forward/reverse search [#3773](https://github.com/valhalla/valhalla/pull/3773)
    * CHANGED: Templatized TimeDistanceBSSMatrix for forward/reverse search [#3778](https://github.com/valhalla/valhalla/pull/3778)
    * CHANGED: error code 154 shows distance limit in error message [#3779](https://github.com/valhalla/valhalla/pull/3779)
-   * ADDED: Added a higway tag if manoeuver uses a highway. [#3815](https://github.com/valhalla/valhalla/issues/3815)
-   * ADDED: Added has_toll, has_higway, has_ferry tags to summary field of a leg and route. [#3815](https://github.com/valhalla/valhalla/issues/3815)
+   * ADDED: Added has_toll, has_higway, has_ferry tags to summary field of a leg and route and a highway tag to a maneuver if it includes a highway. [#3815](https://github.com/valhalla/valhalla/issues/3815)
```
nilsnolde, line 143: merged the previous PR's entries here

kevinkreiser (approved):

PR #3800, opened 2022-10-20
fix invalid speed detection in GraphTile::GetSpeed

nilsnolde:
While working on #3795 I came across the case where one matrix location has no `date_time` set but some other location does. Valhalla took the historical data anyways for that location and assumed a constant Monday noon. This PR fixes that, I tested it ~~(though not in code :innocent: )~~ it actually is tested in code 

I guess that wasn't really a problem so far. With routing we know at this point that we have a valid date time if the `flow_mask` says "please consider traffic". So it'd only be a bug once #3795 is merged;)

### DIFF
--- a/valhalla/baldr/graphconstants.h
+++ b/valhalla/baldr/graphconstants.h
@@ -652,7 +652,8 @@ constexpr uint8_t kDefaultFlowMask =
     kFreeFlowMask | kConstrainedFlowMask | kPredictedFlowMask | kCurrentFlowMask;
 constexpr uint32_t kFreeFlowSecondOfDay = 60 * 60 * 0;         // midnight
 constexpr uint32_t kConstrainedFlowSecondOfDay = 60 * 60 * 12; // noon
-constexpr uint32_t kInvalidSecondsOfWeek = -1;
+constexpr uint64_t kInvalidSecondsOfWeek =
+    1048575; // invalid (20 bits - 1), Sunday 23:59:59 is 604799
 
 // There is only 1 bit to store these values, do not exceed the value 1.
 enum class HOVEdgeType : uint8_t { kHOV2 = 0, kHOV3 = 1 };
--- a/valhalla/baldr/graphconstants.h
+++ b/valhalla/baldr/graphconstants.h
@@ -652,8 +652,7 @@ constexpr uint8_t kDefaultFlowMask =
     kFreeFlowMask | kConstrainedFlowMask | kPredictedFlowMask | kCurrentFlowMask;
 constexpr uint32_t kFreeFlowSecondOfDay = 60 * 60 * 0;         // midnight
 constexpr uint32_t kConstrainedFlowSecondOfDay = 60 * 60 * 12; // noon
-constexpr uint32_t kInvalidSecondsOfWeek =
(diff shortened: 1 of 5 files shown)


valhalla/baldr/graphtile.h
```diff
@@ -700,6 +700,7 @@ class GraphTile {
 
     // fallback to constrained if time of week is within 7am to 7pm (or if no time was passed in) and
     // if the edge has constrained speed
+    // kInvalidSecondsOfWeek %= midgard::kSecondsPerDay = 12.1
     seconds %= midgard::kSecondsPerDay;
     auto is_daytime = (25200 < seconds && seconds < 68400);
     if ((invalid_time || is_daytime) && (flow_mask & kConstrainedFlowMask) &&
```
nilsnolde, line 706: this works out here, because even if this `seconds` of 12.1 in the case of invalid is not a day time it's still invalid, so it will look at constrained speed either way.

valhalla/baldr/graphconstants.h
```diff
@@ -652,7 +652,8 @@ constexpr uint8_t kDefaultFlowMask =
     kFreeFlowMask | kConstrainedFlowMask | kPredictedFlowMask | kCurrentFlowMask;
 constexpr uint32_t kFreeFlowSecondOfDay = 60 * 60 * 0;         // midnight
 constexpr uint32_t kConstrainedFlowSecondOfDay = 60 * 60 * 12; // noon
-constexpr uint32_t kInvalidSecondsOfWeek = -1;                 // invalid
+constexpr uint32_t kInvalidSecondsOfWeek =
```
nilsnolde, line 655: I checked, this constant is not used anywhere else where it could damage smth

kevinkreiser: lets go back to `-1` it will still get the correct value but it will be more obvious as a sentinal value and it will continue to work even if we change the number of bits we give to that field in `timeinfo`.

kevinkreiser (dismissed):

valhalla/baldr/graphconstants.h
nilsnolde: right, now I remember why I felt like I had to do this:
```
[build] /home/nilsnolde/dev/cpp/valhalla/valhalla/baldr/time_info.h: In static member function ‘static valhalla::baldr::TimeInfo valhalla::baldr::TimeInfo::invalid()’:
[build] /home/nilsnolde/dev/cpp/valhalla/valhalla/baldr/time_info.h:49:66: error: conversion from ‘uint64_t’ {aka ‘long unsigned int’} to ‘unsigned int:20’ changes value from ‘18446744073709551615’ to ‘1048575’ [-Werror=overflow]
[build]    49 |     return {false, 0, 0, kInvalidSecondsOfWeek, 0, false, nullptr};
[build]       |                                                                  ^
```
while when it's explicitly `1048575` it didn't complain with a warning (aka error).

nilsnolde: I guess it wasn't a problem before bcs nothing ever used that constant `kInvalidSecondsOfWeek` actually to create a `time_info` object.

kevinkreiser (approved):

Issue #3815, opened 2022-10-28
Highway, ferry and tracks tags missing in the response to route


monteirodom:
When requesting a route, one can ask to (_try to_) avoid **ferry**, **highways**, **tolls**.
It's a good behavior to have some information in the response, allowing to know if such segments have actually been used or not.
A tag `toll=true` is actually added to a segment on the response as [documented](https://valhalla.readthedocs.io/en/latest/api/turn-by-turn/api-reference/#trip-legs-and-maneuvers)
But contrary to the docs, `ferry=true` should also be added to a ferry segment in the response, but it is actually missing.

nilsnolde:
Just noticing, this was solved in #3816.

PR #3816, opened 2022-10-28
Dev/tag highway ferry

monteirodom:
#3815
#3819

Beyond the missing tag declared in the issue, I think the tag **highway** should also appear and be documented.
Furthermore, I feel like those tags should also appear on the summary field for each leg and route (like for time_restrictions).

## Tasklist

 - [x] Add tests


kevinkreiser:
this looks like a good start! we should be able to update the regular route gurka tests to check it is working end to end and then also update the docs accordingly. thanks @monteirodom

nilsnolde:
Jep, good catch. Best you install the pre-commit hook by running `./scripts/format.sh` so the formatting gets fixed before pushing.

nilsnolde:
let us know if you need some help/pointers to test this @monteirodom . can be confusing at first. in the end we just want to see that those tags also make it into the pbf & json output. I'm not sure if we can piggy-back on some other gurka map, it also seems those info thingies in the response are not tested yet, so it's about time (and likely the reason `ferry` slipped through). you can look at some easy ones, e.g. `test/gurka/test_top_speed.cc` and inspect both the `Api` pbf object and the json result (there's an overload of `gurka::do_action` that accepts a response string to be set).

monteirodom:
From the comment from @kevinkreiser, I thought you (?) would handle the tests for that one. I can try on my side, but I'm not familiar at all with the gurka tests.

kevinkreiser:
It's simple I can give it a go some time soon

docs/api/turn-by-turn/api-reference.md
```diff
@@ -341,6 +341,9 @@ The summary JSON object includes:
 | :---- | :----------- |
 | `time` | Estimated elapsed time to complete the trip. |
 | `length` | Distance traveled for the entire trip. Units are either miles or kilometers based on the input units specified. |
+| `as_toll`| Flag indicating if the the path uses one or more toll segments. |
```
nilsnolde, line 344: Typo, has_toll

nilsnolde (changes requested):
IMO the changes in this PR look fine. 

What I’d change is the tests. We shouldn’t add smth to gurka that can be extracted into a single test. One map should be enough to test all changes, you could probably very well parameterize it. There’s another open PR we didn’t get to yet, and had very similar changes but for costing: https://github.com/valhalla/valhalla/pull/3733/files. If we could restructure the test similar to that I’d be happy with this PR. Basically only one file and one map.

monteirodom:
> If we could restructure the test similar to that ...

@nilsnolde I can restructure in that way.
Actually, I _copied_ the way it was done for the already exiting test `.../test/gurka/test_use_highways.cc`.
So, should I merge that one also within the same (single file and single map) test? Grouping the tests for `highway`, `toll`, `ferry`? Like for example `.../test/gurka/test_use_feature.cc` ?

monteirodom:
And by the way, shouldn't the `hard_exclusions` be extended for highways and ferries?

nilsnolde:
> So, should I merge that one also within the same (single file and single map) test? Grouping the tests for highway, toll, ferry?

That’s what I’d do. Mainly we want to test the output here (what you added) and to avoid another public method in gurka that doesn’t have a lot applicability for other tests, I think a single test file makes most sense. 

Tbh, we don’t know yet if the hard exclusion PR will make the cut, needs some more perf testing. It does add quite a few extra conditions to the tight loop that are _very_ edge case-y, yet every request will have to pay that penalty. You’re right though, it’d be only consistent to add at least ferries (actually that’s the sanest exclusion flag out of the bunch IMO).

monteirodom:
Indeed there are currently other tests that appear to be similar, but within different files, like all the `test_use_xxx.cc`. I do understand people that prefer to distinguish the tests, to see clearly which one is for what. I thought it was the approach here. But I do agree with the approach to do similar things at the same place.
To be fully consistent, all the `test_use_xxx.cc` should be assembled, but that may be addressed on a proper PR. I'll limit myself to put together the cases of `ferries`, `highways` and `tolls`.

docs/api/turn-by-turn/api-reference.md
nilsnolde: still a typo here;)

test/gurka/test_use_feature.cc
```diff
@@ -0,0 +1,150 @@
...
+   *
+   * @param raw_result the result of a /route or /match request
+   * @param feature expected feature
+   * @param expected_value the expected value that the expected_value should be
+   */
+  void
+  expect_feature(valhalla::Api& raw_result, const std::string& feature, const bool expected_value) {
+    // std::string json = tyr::serializeDirections(raw_result);
+    // std::cout << json.c_str() << std::endl;
+    rapidjson::Document result = gurka::convert_to_json(raw_result, valhalla::Options_Format_json);
```
nilsnolde, line 74: ha, didn't know we had that!

nilsnolde (approved):
perfect thanks, tests look really good! just the one nitpick, then we can merge

PR #3830, opened 2022-11-17
initialize destinations only once in timedistance matrix

### LINKED ISSUE #3768: SetDestinations only once for all origins in TimeDistance matrix

nilsnolde:
fixes #3768 

I'm sure I could've done it nicer by rewriting lots of logic, this is more of a minimal impact PR code-wise.


src/thor/timedistancebssmatrix.cc
```diff
@@ -43,22 +43,6 @@ float TimeDistanceBSSMatrix::GetCostThreshold(const float max_matrix_distance) c
   return max_matrix_distance / (2.0f * kMPHtoMetersPerSec);
 }
 
-// Clear the temporary information generated during time + distance matrix
-// construction.
-void TimeDistanceBSSMatrix::clear() {
```
nilsnolde, line 48: moved to header and renamed to `reset` where it doesn't clear most destination info

src/thor/timedistancebssmatrix.cc
```diff
@@ -203,11 +187,15 @@ std::vector<TimeDistance> TimeDistanceBSSMatrix::ComputeMatrix(
   // Run a series of one to many calls and concatenate the results.
   const auto& origins = FORWARD ? source_location_list : target_location_list;
   const auto& destinations = FORWARD ? target_location_list : source_location_list;
+  destinations_.reserve(destinations.size());
```
nilsnolde, line 190: oops, don't need that I think

src/thor/timedistancebssmatrix.cc
```diff
@@ -387,14 +374,14 @@ void TimeDistanceBSSMatrix::SetOrigin(GraphReader& graphreader, const valhalla::
...
-void TimeDistanceBSSMatrix::SetDestinations(
+void TimeDistanceBSSMatrix::InitDestinations(
     GraphReader& graphreader,
     const google::protobuf::RepeatedPtrField<valhalla::Location>& locations) {
   // For each destination
   uint32_t idx = 0;
   for (const auto& loc : locations) {
     // Set up the destination - consider each possible location edge.
-    bool added = false;
+    bool first_edge = true;
```
nilsnolde, line 384: just a clearer (IMO) var name

src/thor/timedistancebssmatrix.cc
```diff
@@ -501,8 +487,8 @@ bool TimeDistanceBSSMatrix::UpdateDestinations(
 
     // Erase this edge from further consideration. Mark this destination as
     // settled if all edges have been found
-    dest.dest_edges.erase(dest_edge);
-    if (dest.dest_edges.empty()) {
+    dest.dest_edges_available.erase(pred.edgeid());
```
nilsnolde, line 490: the actual change is adding this `dest_edges_available` unordered_set and splitting it from the `percent_along` info. the former we'll reset on every origin, the latter will stay for the entire request.

valhalla/thor/timedistancebssmatrix.h
```diff
@@ -78,7 +78,11 @@ class TimeDistanceBSSMatrix {
    * Clear the temporary information generated during time+distance
    * matrix construction.
    */
-  void clear();
+  inline void clear() {
+    reset();
+    destinations_.clear();
+    dest_edges_.clear();
+  };
```
nilsnolde, line 85: won't be called by td matrix anymore, but is called when cleaning up after the request.

.pre-commit-config.yaml
```diff
@@ -10,7 +10,7 @@ repos:
 
   # Python formatter
   - repo: https://github.com/psf/black
-    rev: 22.6.0
+    rev: 22.10.0
```
nilsnolde, line 13: updating black to recent version. doesn't affect the formatting at all (updates never should with this formatter)

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 * **Removed**
 * **Bug Fix**
    * FIXED: valhalla_run_route was missing config logic.[#3824](https://github.com/valhalla/valhalla/pull/3824)
+   * FIXED: Added missing ferry tag if manoeuver uses a ferry. It's supposed to be there according to the docs. [#3815](https://github.com/valhalla/valhalla/issues/3815)
+   * FIXED: only set most destination information once for all origins in timedistancematrix [#3830](https://github.com/valhalla/valhalla/pull/3830)
 * **Enhancement**
+   * ADDED: Added has_toll, has_higway, has_ferry tags to summary field of a leg and route and a highway tag to a maneuver if it includes a highway. [#3815](https://github.com/valhalla/valhalla/issues/3815)
```
nilsnolde, line 8: oops, in #3816 we forgot that we released 3.2.0 :smile:

src/worker.cc
```diff
@@ -958,6 +958,8 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
 
   // Get the matrix_loctions option and set if sources or targets size is one
   // (option is only supported with one to many or many to one matrix requests)
+  // TODO(nils): Why is that? IMO this makes a lot of sense for a many:many call
+  // of timedistancematrix as well no?
```
nilsnolde, line 962: @dnesbitt61 shouldn't we allow to set the `matrix_locations` also for many:many? I think the same logic applies there as well, faster results if someone only wants a subset of destinations for each origin.

valhalla/thor/matrix_common.h
```diff
@@ -27,22 +27,30 @@ constexpr uint32_t kAllLocations = std::numeric_limits<uint32_t>::max();
...
+  float threshold;   // Threshold above current best cost where no longer
+                     // need to search for this destination.
+  // partial distance of correlated edges
+  std::unordered_map<uint64_t, float> dest_edges_percent_along;
 
   // Constructor - set best_cost to an absurdly high value so any new cost
   // will be lower.
-  Destination()
-      : settled(false), best_cost{kMaxCost, kMaxCost}, distance(0), threshold(0.0f), date_time("") {
+  Destination() : settled(false), best_cost{kMaxCost, kMaxCost}, distance(0), threshold(0.0f) {
```
nilsnolde, line 46: in #3795 I apparently left a redundant `date_time` parameter in this struct. that was from an earlier attempt and not being used anymore.

src/thor/timedistancematrix.cc
```diff
@@ -502,16 +489,15 @@ bool TimeDistanceMatrix::UpdateDestinations(
...
     // See if this edge is part of the destination
-    // TODO - it should always be, but protect against not finding it
-    auto dest_edge = dest.dest_edges.find(pred.edgeid());
-    if (dest_edge == dest.dest_edges.end()) {
-      // If the edge isn't there but the path is trivial, then that means the edge
-      // was removed towards the beginning which is not an error.
-      if (!IsTrivial(pred.edgeid(), origin, dest_loc)) {
+    // If the edge isn't there but the path is trivial, then that means the edge
+    // was removed towards the beginning which is not an error.
+    if (dest.dest_edges_available.find(pred.edgeid()) == dest.dest_edges_available.end()) {
```
nilsnolde, line 494: cache the result iterator for the erase below

src/thor/timedistancebssmatrix.cc
```diff
@@ -472,16 +458,15 @@ bool TimeDistanceBSSMatrix::UpdateDestinations(
...
 
     // See if this edge is part of the destination
-    // TODO - it should always be, but protect against not finding it
-    auto dest_edge = dest.dest_edges.find(pred.edgeid());
-    if (dest_edge == dest.dest_edges.end()) {
-      // If the edge isn't there but the path is trivial, then that means the edge
-      // was removed towards the beginning which is not an error.
+    // If the edge isn't there but the path is trivial, then that means the edge
+    // was removed towards the beginning which is not an error.
+    if ((dest.dest_edges_available.find(pred.edgeid())) == dest.dest_edges_available.end()) {
```
nilsnolde, line 463: cache for erase below

src/thor/timedistancematrix.cc
nilsnolde: done

src/thor/timedistancebssmatrix.cc
nilsnolde: done

nilsnolde:
@kevinkreiser should be done now, I fixed the iterator.

@dnesbitt61 in case you want to have a look before we merge, I'd appreciate it.

kevinkreiser (approved):

kevinkreiser (approved):

Issue #3761, opened 2022-09-25
collect warnings to implement


nilsnolde:
with #3588 landing sometime soon, let's collect things to warn on.

- [ ] `matrix_locations` is ineffective if CostMatrix is used

PR #3833, opened 2022-11-18
add more warnings

nilsnolde:
will fix part of #3761 

- [x] matrix_locations is ineffective if CostMatrix is used
- [x] out-of-range values for costing options where we clamp to the default
- ~~[ ] when loki uses filtered edges since it can't find any other edges~~ maybe later..
- [x] when time-dependent was requested, but the crow-distance exceeds the server limit of max_timedep_distance, we don't do proper time-dependence but instead the 80% correct way with bidir AStar
- [x] second pass for routing

### DIFF
--- a/src/mjolnir/adminbuilder.cc
+++ b/src/mjolnir/adminbuilder.cc
@@ -144,9 +144,7 @@ void buffer_polygon(const bg::polygon_ll_t& polygon, bg::multipolygon_ll_t& mult
   for (const auto& inner : polygon.inners()) {
     inner_rings.push_back(geos_helper_t::from_striped_container(inner));
   }
-  auto* geos_poly =
-      GEOSGeom_createPolygon(outer_ring, inner_rings.empty() ? nullptr : inner_rings.data(),
-                             inner_rings.size());
+  auto* geos_poly = GEOSGeom_createPolygon(outer_ring, &inner_rings.front(), inner_rings.size());
   auto* buffered = GEOSBuffer(geos_poly, 0, 8);
   GEOSNormalize(buffered);
   auto geom_type = GEOSGeomTypeId(buffered);


valhalla/midgard/util.h
```diff
@@ -27,13 +27,21 @@
...
+  // important when we clamp the value
+  char* option = "";
+
   // Returns the value snapped to the default if outside of the range
-  T operator()(const T& value) const {
+  T operator()(const T& value, char* option_name = "") const {
     if (value < min || value > max) {
+      clamped = true;
+      // TODO: remove leading "/"
+      option = option_name;
```
nilsnolde, line 44: Turns out warning the user when out-of-range costing options were clamped to defaults is actually quite a big change potentially. #

This is just a suggestion: in the `JSON_PBF_RANGED_DEFAULT` macro we pass the option name and in this struct we flag it as clamped if that's what we're doing. Then later in the processing we could go over all such `ranged_default_t` objects and generate a warning for each that has `clamped = true`. There's LOTS of these range objects, so it'll be a bit ugly, unless we collect them even more in smth like `BaseCostingOptionsConfig`.

src/sif/dynamiccost.cc
```diff
@@ -382,6 +382,9 @@ void ParseBaseCostOptions(const rapidjson::Value& json,
   // shortest
   JSON_PBF_DEFAULT(co, false, json, "/shortest", shortest);
 
+  // TODO: need to check all ranges if they were clamped or not and add a warning if so
+  // thing is, many costing models have another host of custom default ranges
```
nilsnolde, line 386: see if we can pass the warnings from the request pbf to the operator(). also only pointer to "clamped" bool to pass to macro and set warning there (no option name passing).

kevinkreiser: we'd need to pass into `JSON_PBF_RANGED_DEFAULT` the `warnings` part of the request so it can modify it

valhalla/midgard/util.h
```diff
@@ -27,13 +27,21 @@
...
   T min, def, max;
 
+  bool clamped = false;
+  // so we know which costing option this object refers to
+  // important when we clamp the value
+  char* option = "";
+
   // Returns the value snapped to the default if outside of the range
-  T operator()(const T& value) const {
+  T operator()(const T& value, char* option_name = "") const {
```
kevinkreiser, line 40: instead of passing the name we could pass an optional pointer to a boolean, and we set that boolean (if not null) to whether or not the value was clamped

valhalla/sif/dynamiccost.h
```diff
@@ -44,7 +44,8 @@
         range(rapidjson::get<decltype(range.def)>(json, json_key,                                    \
                                                   costing_options->has_##option_name##_case()        \
                                                       ? costing_options->option_name()               \
-                                                      : range.def)));                                \
+                                                      : range.def),                                  \
+              json_key));                                                                            \
```
kevinkreiser, line 48: then over here we check if that boolean was `true` meaning the range default clamped it, and if so we add a warning to whatever place we have to add warnings

src/sif/autocost.cc
```diff
@@ -674,23 +674,24 @@ Cost AutoCost::TransitionCostReverse(const uint32_t idx,
...
 
-  ParseBaseCostOptions(json, c, kBaseCostOptsConfig);
-  JSON_PBF_RANGED_DEFAULT(co, kAlleyFactorRange, json, "/alley_factor", alley_factor);
-  JSON_PBF_RANGED_DEFAULT(co, kUseHighwaysRange, json, "/use_highways", use_highways);
-  JSON_PBF_RANGED_DEFAULT(co, kUseTollsRange, json, "/use_tolls", use_tolls);
-  JSON_PBF_RANGED_DEFAULT(co, kUseDistanceRange, json, "/use_distance", use_distance);
-  JSON_PBF_RANGED_DEFAULT(co, kAutoHeightRange, json, "/height", height);
-  JSON_PBF_RANGED_DEFAULT(co, kAutoWidthRange, json, "/width", width);
+  ParseBaseCostOptions(json, c, kBaseCostOptsConfig, warnings);
+  JSON_PBF_RANGED_DEFAULT(co, kAlleyFactorRange, json, "/alley_factor", alley_factor, warnings);
```
nilsnolde, line 687: these are most of the changes: when the input exceeds the acceptable range and we clamp, we add a warning to that effect, indicating the name of the option and the clamped value

src/thor/matrix_action.cc
```diff
@@ -6,6 +6,7 @@
 #include "thor/timedistancematrix.h"
 #include "thor/worker.h"
 #include "tyr/serializers.h"
+#include <valhalla/worker.h>
```
nilsnolde, line 9: this makes me want to pull the warning stuff out of the worker.h. tons of unneeded stuff being compiled in here..

src/thor/route_action.cc
```diff
@@ -262,6 +263,8 @@ thor::PathAlgorithm* thor_worker_t::get_path_algorithm(const std::string& routet
     PointLL ll2(destination.ll().lng(), destination.ll().lat());
     if (ll1.Distance(ll2) < max_timedep_distance) {
       return &timedep_forward;
+    } else {
+      add_warning(request, 402);
```
nilsnolde, line 267: further down we could still choose unidir a* if it's trivial/connected pair, but in that case it hardly matters if a time was passed in

src/worker.cc
```diff
@@ -157,11 +157,17 @@ const std::unordered_map<int, std::string> warning_codes = {
...
   // 3xx is used when costing options were specified but we had to change them internally for some reason
   {300, R"(Many:Many CostMatrix was requested, but server only allows 1:Many TimeDistanceMatrix)"},
   {301, R"(1:Many TimeDistanceMatrix was requested, but server only allows Many:Many CostMatrix)"},
   // 4xx is used when we do sneaky important things the user should be aware of
-  {400, R"(CostMatrix turned off destination-only on a second pass for connections: )"}
+  {400, R"(CostMatrix turned off destination-only on a second pass for connections: )"},
+  {401, R"(Routing turned off destination-only and might have used filtered edges on a second pass)"},
+  {402, R"(Distance exceeded max_timedep_distance for depart_at, probably falling back to bidirectional A*)"},
+  // RESERVED: 500 is for clamped values, set in JSON_PBF_RANGED_DEFAULT* macros
+  // {500, R"(<option> was clamped to <value> )"},
```
nilsnolde, line 170: this one we set in the macros instead of calling `add_warning()`, but we shouldn't forget that 500 is reserved

src/worker.cc
```diff
@@ -1045,7 +1053,8 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
   auto matrix_locations = rapidjson::get_optional<int>(doc, "/matrix_locations");
   if (matrix_locations && (options.sources_size() == 1 || options.targets_size() == 1)) {
     options.set_matrix_locations(*matrix_locations);
-  } else if (!options.has_matrix_locations_case()) {
+  } else if (!options.has_matrix_locations_case() ||
+             (options.sources_size() != 1 && options.targets_size() != 1)) {
```
nilsnolde, line 1057: even if the pbf has this option we still have to check if it even makes sense.

test/gurka/test_matrix.cc
```diff
@@ -155,7 +155,7 @@ TEST_F(MatrixTrafficTest, TDMatrixWithLiveTraffic) {
   res_doc.Parse(res.c_str());
   // the second origin can't respect time (no historical data)
   check_matrix(res_doc, {0.0f, 2.8f, 3.2f, 0.0f}, false, Matrix::TimeDistanceMatrix);
-  ASSERT_EQ(result.info().warnings().size(), 0);
+  EXPECT_EQ(result.info().warnings().size(), 0);
```
nilsnolde, line 158: just changing a bunch of ASSERTs to EXPECTs

valhalla/midgard/util.h
```diff
@@ -27,13 +27,14 @@
...
-// Holds a range plus a default value for that range
+// Holds a named option, its range plus a default value for that range
 template <class T> struct ranged_default_t {
   T min, def, max;
 
   // Returns the value snapped to the default if outside of the range
-  T operator()(const T& value) const {
+  T operator()(const T& value, bool& clamped) const {
     if (value < min || value > max) {
+      clamped = true;
```
nilsnolde, line 37: it's basically following the suggestion here: https://github.com/valhalla/valhalla/pull/3833#discussion_r1030396723

nilsnolde:
blasts from the past :) finally got around to finishing it with the most important warnings (IMO).

src/worker.cc
nilsnolde: though to be honest, I think even a many:many matrix with the 1:many timedistancematrix algorithm could want the `matrix_locations` parameter. probably one request every 10 years will want that, but it can still make sense (e.g. show me closest hotels for all these locations). so I'd actually be in favor to remove that restriction to use `matrix_locations`. 

what do you think @dnesbitt61 ?

dnesbitt61: this could be added to more than one to many matrix requests. To me that was the most common use case (and easiest) to add. For many-to-many matrices the parameter needs further definition and documentation. Does it mean each source and target has to find that minimum number of matrix locations? I guess optimized route (traveling salesman) would not use this but if the results were used in an external optimizer that could cause issues. Anyway, I think documentation has to be explicit about how this works with many-to-many cases.

nilsnolde: > Does it mean each source and target has to find that minimum number of matrix locations?

I wouldn't say _has to_, but IIRC the same is true for today's 1:many, we're finished if either that number of target has been reached or all targets have been reached. in the end isn't it the exact same but just extending it to every origin (for fwd, every target for rev)? we're just using a loop there over every source or target which does the 1:many, I don't think it makes much of a difference conceptually. 

maybe there's a misunderstanding though: I'm not advocating to use this for CostMatrix, just for many:many timedistancematrix, because with CostMatrix we find bidirectional connections which makes the interpretation of such a parameter near impossible in my mind.

nilsnolde:
hmpf, I always forget about this one.. would be nice if one could pin PRs just like issues..

src/worker.cc
nilsnolde: try again within the macros

valhalla/sif/dynamiccost.h
```diff
@@ -34,17 +34,26 @@
...
                                                   costing_options->has_##option_name##_case()        \
                                                       ? costing_options->option_name()               \
-                                                      : range.def)));                                \
+                                                      : range.def),                                  \
+              clamped));                                                                             \
+    if (clamped) {                                                                                   \
+      auto warning = warnings.Add();                                                                 \
+      warning->set_description("'" + std::string(json_key) + "' has been clamped to " +              \
+                               std::to_string(range.def));                                           \
+                                                                                                     \
```
nilsnolde, line 54: try to call `add_warning`

nilsnolde: @kevinkreiser if we'd like to set this warning status inside the macros, we'd have to
- include the whole `worker.h` into every costing model (which is quite a bit extra to compile?)
- `add_warning` expects the full request, which means I'd have to refactor everything again and also touch the `ParseCosting` etc to not accept the warnings but the full request

so it'd even lead a bigger refactor than what it is currently, unless we're ok that functions accept a `CostingOptions` pbf on top of the full request (where `CostingOptions` comes from). 

are you ok if we just leave it like this and set the warning directly in the macros, i.e. leave the PR as it is?

src/exceptions.cc
```diff
@@ -148,13 +148,19 @@ const std::unordered_map<int, std::string> warning_codes = {
...
   {302, R"("search_filter.level" was specified without a custom "search_cutoff", setting default default cutoff to )"},
   {303, R"("search_cutoff" exceeds maximum allowed value due to "search_filter.level" being specified, clamping cutoff to )"},
   // 4xx is used when we do sneaky important things the user should be aware of
-  {400, R"(CostMatrix turned off destination-only on a second pass for connections: )"}
+  {400, R"(CostMatrix turned off destination-only on a second pass for connections: )"},
+  {401, R"(Routing failed on first pass, retrying with relaxed restrictions)"},
+  {402, R"(Distance exceeded max_timedep_distance for depart_at, falling back to bidirectional A*)"},
+  // RESERVED: 500 is for clamped values, set in JSON_PBF_RANGED_DEFAULT* macros
+  // {500, R"(<option> was clamped to <value> )"},
 };
```
chrstnbwnkl, line 164: just an idea/question: did you consider using the full range of 5xx to be more specific about which values were clamped? Probably a lot of work though for  not that much extra benefit

src/mjolnir/adminbuilder.cc
```diff
@@ -144,7 +144,9 @@ void buffer_polygon(const bg::polygon_ll_t& polygon, bg::multipolygon_ll_t& mult
   for (const auto& inner : polygon.inners()) {
     inner_rings.push_back(geos_helper_t::from_striped_container(inner));
   }
-  auto* geos_poly = GEOSGeom_createPolygon(outer_ring, &inner_rings.front(), inner_rings.size());
+  auto* geos_poly =
+      GEOSGeom_createPolygon(outer_ring, inner_rings.empty() ? nullptr : inner_rings.data(),
+                             inner_rings.size());
```
chrstnbwnkl: what's this change? :smile:

src/exceptions.cc
nilsnolde: yeah I think so (way back when). not sure anymore what the problem was but I think it was just substantially more complex or more refactor. we can do it in the future maybe, when someone needs it?

src/mjolnir/adminbuilder.cc
nilsnolde: huh, yeah, so strange., although I'm very sure I didn't do this myself and it's some sort of bad merge at some point in the 4 year history of this PR, I was almost inclined to keep it :sweat_smile: 

I reverted it anyways, nothing suggests this check is needed.

PR #3836, opened 2022-11-21
add 'available_actions' to /status response

### LINKED ISSUE #3835: Return available actions in /status
It could be nice if we'd return the config's available actions in the status response by default (i.e. without `verbose`). Clients could just query that endpoint to see which features are available on that server, instead of hitting each endpoint and inspecting the error if one is missing. 

Any objections?

nilsnolde:
fixes #3835 

pretty straight-forward, not much more to say:)


src/tyr/serializers.cc
```diff
@@ -102,6 +102,12 @@ std::string serializeStatus(Api& request) {
   status_doc.AddMember("tileset_last_modified",
                        rapidjson::Value().SetInt(request.status().tileset_last_modified()), alloc);
 
+  rapidjson::Value actions_list(rapidjson::kArrayType);
+  for (const auto& action : request.status().available_actions()) {
+    actions_list.GetArray().PushBack(rapidjson::Value{}.SetString(action.c_str(), alloc), alloc);
+  }
+  status_doc.AddMember("available_actions", actions_list, alloc);
```
nilsnolde, line 109: would've like to switch this whole serializer to the managed `writer` syntax, so much more readable. but the `bbox` kinda prevents that, or we'd have to live with stringified coord floats. or write a geojson abstraction for the writer type..

test/loki_service.cc
```diff
@@ -94,87 +94,88 @@ const std::vector<http_request_t> valhalla_requests{
...
-     R"({"error_code":131,"error":"Failed to parse source","status_code":400,"status":"Bad Request"})"},
-    {400,
-     R"({"error_code":132,"error":"Failed to parse target","status_code":400,"status":"Bad Request"})"},
-    {400,
-     R"({"error_code":157,"error":"Exceeded max avoid locations: 0","status_code":400,"status":"Bad Request"})"},
-    {400,
-     R"({"error_code":153,"error":"Too many shape points: (102). The best paths shape limit is 100","status_code":400,"status":"Bad Request"})"}};
+const std::vector<std::pair<uint16_t, std::string>> valhalla_responses{{200,
+                                                                        R"({"version":")" VALHALLA_VERSION
+                                                                        R"(","tileset_last_modified":0,"available_actions":["status","centroid","expansion","transit_available","trace_attributes","trace_route","isochrone","optimized_route","sources_to_targets","height","route","locate"]})"},
```
nilsnolde, line 99: the change is really just adding the `available_actions` field to some responses

test/loki_service.cc
```diff
@@ -94,87 +94,88 @@ const std::vector<http_request_t> valhalla_requests{
...
+                                                                       {400,
+                                                                        R"({"error_code":131,"error":"Failed to parse source","status_code":400,"status":"Bad Request"})"},
+                                                                       {400,
+                                                                        R"({"error_code":132,"error":"Failed to parse target","status_code":400,"status":"Bad Request"})"},
+                                                                       {400,
+                                                                        R"({"error_code":157,"error":"Exceeded max avoid locations: 0","status_code":400,"status":"Bad Request"})"},
+                                                                       {400,
+                                                                        R"({"error_code":153,"error":"Too many shape points: (102). The best paths shape limit is 100","status_code":400,"status":"Bad Request"})"}};
```
kevinkreiser, line 178: extra comma will fix whitespace change
```suggestion
                                                                        R"({"error_code":153,"error":"Too many shape points: (102). The best paths shape limit is 100","status_code":400,"status":"Bad Request"})"},};
```

kevinkreiser (approved):

PR #3842, opened 2022-11-24
Handle hexlifying strings with unsigned chars

dillona:
Observed on aarch64-linux-gnu

# Issue

Fixes #2923

PR #3845, opened 2022-11-24
Build all of Mjolnir with -Werror

dillona:


dillona:
I tested initially with Ubuntu 20.04 on both amd64 and aarch64. I see the failing macOS test now, so I will get that fixed.

kevinkreiser (dismissed):

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 * **Bug Fix**
    * FIXED: valhalla_run_route was missing config logic.[#3824](https://github.com/valhalla/valhalla/pull/3824)
    * FIXED: Handle hexlifying strings with unsigned chars [#3842](https://github.com/valhalla/valhalla/pull/3842)
-   * FIXED: Valhalla does not build on macOS with clang 14.0.0 [#3846](https://github.com/valhalla/valhalla/issues/3846)
+   * FIXED: Fix warning as error on macOS with clang 14.0.0 [#3846](https://github.com/valhalla/valhalla/issues/3846)
```
dillona, line 6: I don't think this is a correct change, as the macOS problem was observed on master before I made any warning as error changes

dillona:
Rebased to latest master in order to get CI to run again (seemed to have some transient error earlier)

CHANGELOG.md
kevinkreiser: If I'm not mistaken the issue was that warnings were already treated as error and that on the arm platform the char was unsigned so the compiler warned that the check for `> 0` was meaningless and the build stopped because werror was already on. Are you saying this wasn't the cause? I remember others working around this issue before but i afmit it was just my assumption as i dont have an arm mac to verify. I found the message about mac builds being broken a bit vague and wanted to put a better description of the problem in the changelog

kevinkreiser (approved):

CHANGELOG.md
dillona: I believe you're describing #3842, which I was hitting on ARM Linux (Ubuntu 20.04).

For #3846, I was seeing it on x86 macOS. I believe `logging.cc` was already using `-Werror`, and the updated compiler issued a new warning for uses of `sprintf`, which then broke the build. I believe the only reason this hadn't show up in the x86 CI is that CI isn't using the 14.x compiler yet.

So I suppose the updated changelog text was correct; it was just different warning as error changes than the ones I have been working on.

kevinkreiser: Oops, no worries I'll fix it up in a little bit. Apologies for the mix up!

Issue #3825, opened 2022-11-11
support traffic without timezone


nilsnolde:
ref https://github.com/valhalla/valhalla/pull/3795#discussion_r1010803231

When traffic is available but the data was built without timezones, it seems like we lose date_time information on the destinations.

nilsnolde:
just came across the case where a gurka map wasn't build with timezone indices and exactly this failed.

PR #3849, opened 2022-11-26
add "waiting time" to locations

### LINKED ISSUE #3832: Add `waiting_time` to locations for traffic-enabled routing
A requirement from a client, quoting from the email:

_in the turn-by-turn API, in the input locations array, is it possible to add a parameter "wait_time_after" or something equivalent? Use case - we want to calculate a traffic-aware ETA on a path of ~5-10 locations, but at each location, we want to stop for a variable number of minutes (as is the case with delivery services). Currently we either have to serialize all calls, where we get the ETA between the first two locations, add the wait_time_after on our application side, and then request data for the next segment, or, we have to simply ignore the wait times and get more inaccurate results._

I think that's a very reasonable request, e.g. in the context of optimizing a delivery fleet with traffic-enabled matrix calls (they funded #3795 as well), a vehicle has to spend some service time at a location which should be taken into account when generating the final vehicle route with a narrative. Ideally Valhalla would do 2 things here:
- add the `waiting_time` to the `time_info` object at the end of the previous leg, so it'll assume more relevant predictive traffic for starting the next leg
- add a few things to the `summary` objects: `total_waiting_time` for the overall route and a `waiting_time` to each leg's `summary`

Obviously only relevant for `break`/`break_through` location types ; we can just ignore it (and add a warning) when using other types? What do you think @kevinkreiser (and others of course)?

nilsnolde:
fixes #3832 

How it works:
- `waiting` input in seconds as `float`
- `src/worker.cc` checks if a location is intermediate **and** `break` when `waiting` is set, otherwise emits a warning
- `thor/route_action.cc` deals with juggling the logic for `depart_at` & `arrive_by`. We always add the waiting time after a leg is finished to the next origin's `date_time` object. 
- in the JSON output we add a `waiting` field to the `locations` mirroring the input and a location's `date_time` field of intermediate waypoints are adjusted for the `waiting` time, so it's understood as a `depart_at` time, I think that makes most sense.

`/optimized` should get it for free, but not sure it's so desirable.

What I wanna discuss:
- name.. `waiting`, `waiting_secs`, `waiting_time`?
- unit: seconds or minutes (then e.g. `waiting_mins`)
- currently in the response, an intermediate's `date_time` field is regarded as `depart_at` if it has a `waiting` time, i.e. we add that time to the `date_time`. I think that's fine. Or should we add another field to those locations, e.g. `date_time_depart` and have `date_time` as arrival? I think the way it is it's fine.
- currently the `summary.time` fields **do not** include the `waiting` time. Both is appealing somehow. I feel like it kind of belongs in the overall summary, but not in each leg's summary, but I can't convince my fingers to execute such an inconsistency.. What do you think @kevinkreiser ?

A request looks e.g. like this, adding 10 mins `waiting` time to the intermediate location:
```json
{
	"costing": "truck",
	"locations": [
		{
			"lon": 13.44350292,
			"lat": 52.46437987
		},
		{
			"lon": 13.44564241,
			"lat": 52.47644020,
			"waiting": 600
		},
		{
			"lon": 13.42095168,
			"lat": 52.48738710
		}
	],
	"date_time": {
		"value": "2022-11-25T09:00",
		"type": 1
	}
}
``` 

A response like this, where it actually arrives at the second location at `2022-11-25T09:09` but leaves at `2022-11-25T09:19`:
```json
{
	"trip": {
		"locations": [
			{
				"type": "break",
				"lat": 52.464379,
				"lon": 13.443502,
				"date_time": "2022-11-25T09:00",
				"original_index": 0
			},
			{
				"type": "break",
				"lat": 52.47644,
				"lon": 13.445642,
				"date_time": "2022-11-25T09:19",
				"waiting": 600,
				"original_index": 1
			},
			{
				"type": "break",
				"lat": 52.487387,
				"lon": 13.420951,
				"date_time": "2022-11-25T09:25",
				"original_index": 2
			}
		],
...
}
````

### DIFF
--- a/proto/common.proto
+++ b/proto/common.proto
@@ -123,6 +123,7 @@ message Location {
   oneof has_preferred_layer {
     int32 preferred_layer = 28;
   }
+  float waiting_secs = 29;                   // waiting period before a new leg starts, e.g. for servicing/loading goods
 
   // This information will be ignored if provided in the request. Instead it will be filled in as the request is handled
   Correlation correlation = 90;
--- a/src/baldr/datetime.cc
+++ b/src/baldr/datetime.cc
@@ -126,13 +126,10 @@ std::string iso_date_time(const date::time_zone* time_zone) {
 
 // Get the seconds since epoch time is already adjusted based on TZ
 uint64_t seconds_since_epoch(const std::string& date_time, const date::time_zone* time_zone) {
-  if (date_time.empty()) {
+  if (date_time.empty() || !time_zone) {
     return 0;
   }
   const auto d = get_formatted_date(date_time);
-  if (!time_zone) {
-    return static_cast<uint64_t>(d.time_since_epoch().count());
-  }
   const auto utc = date::to_utc_time(get_ldt(d, time_zone).get_sys_time()); // supports leap sec.
   return static_cast<uint64_t>(utc.time_since_epoch().count());
 }
@@ -165,22 +162,19 @@ std::string
 seconds_to_date(const uint64_t seconds, const date::time_zone* time_zone, bool tz_format) {
 
   std::string iso_date;
-  if (seconds == 0) {
+  if (seconds == 0 || !time_zone) {
     return iso_date;
   }
 
-  std::ostringstream iso_date_time;
-  const char* format = tz_format ? "%FT%R%z" : "%FT%R";
   std::chrono::seconds dur(seconds);
   std::chrono::time_point<std::chrono::system_clock> tp(dur);
+  const auto date = date::make_zoned(time_zone, tp);
 
-  if (time_zone) {
-    const auto date = date::make_zoned(time_zone, tp);
-    iso_date_time << date::format(format, date);
-  } else {
-    iso_date_time << date::format(format, tp);
-  }
-
+  std::ostringstream iso_date_time;
+  if (tz_format)
+    iso_date_time << date::format("%FT%R%z", date);
+  else
+    iso_date_time << date::format("%FT%R", date);
   iso_date = iso_date_time.str();
   if (tz_format)
     iso_date.insert(19, 1, ':');
--- a/src/baldr/datetime.cc
+++ b/src/baldr/datetime.cc
@@ -126,10 +126,13 @@ std::string iso_date_time(const date::time_zone* time_zone) {
 
 // Get the seconds since epoch time is already adjusted based on TZ
 uint64_t seconds_since_epoch(const std::string& date_time, const date::time_zone* time_zone) {
-  if (date_time.empty() || !time_zone) {
+  if (date_time.empty()) {
     return 0;
   }
   const auto d = get_formatted_date(date_time);
+  if (!time_zone) {
+    return static_cast<uint64_t>(d.time_since_epoch().count());
+  }
   const auto utc = date::to_utc_time(get_ldt(d, time_zone).get_sys_time()); // supports leap sec.
   return static_cast<uint64_t>(utc.time_since_epoch().count());
 }
@@ -162,19 +165,22 @@ std::string
 seconds_to_date(const uint64_t seconds, const date::time_zone* time_zone, bool tz_format) {
 
   std::string iso_date;
-  if (seconds == 0 || !time_zone) {
+  if (seconds == 0) {
     return iso_date;
   }
 
+  std::ostringstream iso_date_time;
+  const char* format = tz_format ? "%FT%R%z" : "%FT%R";
   std::chrono::seconds dur(seconds);
   std::chrono::time_point<std::chrono::system_clock> tp(dur);
-  const auto date = date::make_zoned(time_zone, tp);
 
-  std::ostringstream iso_date_time;
-  if (tz_format)
-    iso_date_time << date::format("%FT%R%z", date);
-  else
-    iso_date_time << date::format("%FT%R", date);
+  if (time_zone) {
+    const auto date = date::make_zoned(time_zone, tp);
+    iso_date_time << date::format(format, date);
+  } else {
+    iso_date_time << date::format(format, tp);
+  }
+
   iso_date = iso_date_time.str();
   if (tz_format)
     iso_date.insert(19, 1, ':');
--- a/src/thor/route_action.cc
+++ b/src/thor/route_action.cc
@@ -481,11 +481,6 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
   auto origin = ++correlated.rbegin();
   while (origin != correlated.rend()) {
     auto destination = std::prev(origin);
-
-    std::cerr << "Processing origin/destination " +
-                     std::to_string(std::distance(correlated.rend(), origin)) + " / " +
-                     std::to_string(std::distance(correlated.rend(), destination));
-
     if (!route_two_locations(origin, destination)) {
       // if routing failed because an intermediate waypoint was snapped to the low reachability road
       // (such road lies in a small connectivity component that is not connected to other locations)
--- a/src/thor/route_action.cc
+++ b/src/thor/route_action.cc
@@ -388,6 +388,14 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
         origin->set_date_time(origin_dt);
       }
 
+      // add waiting_secs again from the final destination's datetime, so we output the departing time
+      // at intermediate locations, not the arrival time
+      if (destination->waiting_secs() && !destination->date_time().empty()) {
+        auto dest_dt =
+            offset_date(*reader, destination->date_time(), {}, destination->waiting_secs(), {});
+        destination->set_date_time(dest_dt);
+      }
+
       first_edge = temp_path.front().edgeid;
       temp_path.swap(path); // so we can append to path instead of prepend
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

File: proto/common.proto [hunk 1]
```diff
@@ -123,6 +123,7 @@ message Location {
   oneof has_preferred_layer {
     int32 preferred_layer = 28;
   }
+  float waiting_secs = 29;                   // waiting period before a new leg starts, e.g. for servicing/loading goods
```
nilsnolde, line 126: no `oneof`, default 0 is perfect

src/baldr/datetime.cc
```diff
@@ -126,10 +126,13 @@ std::string iso_date_time(const date::time_zone* time_zone) {
...
 // Get the seconds since epoch time is already adjusted based on TZ
 uint64_t seconds_since_epoch(const std::string& date_time, const date::time_zone* time_zone) {
-  if (date_time.empty() || !time_zone) {
+  if (date_time.empty()) {
     return 0;
   }
   const auto d = get_formatted_date(date_time);
+  if (!time_zone) {
+    return static_cast<uint64_t>(d.time_since_epoch().count());
+  }
```
nilsnolde, line 135: Ah right, almost forgot, this fixes #3825 at least partly. Still needs a bit more work in other functions here. this was just added so this PR works. The rest is still as broken as before when no timeinfo is present on the nodes.

src/baldr/datetime.cc
```diff
@@ -162,19 +165,22 @@ std::string
...
-  if (tz_format)
-    iso_date_time << date::format("%FT%R%z", date);
-  else
-    iso_date_time << date::format("%FT%R", date);
+  if (time_zone) {
+    const auto date = date::make_zoned(time_zone, tp);
+    iso_date_time << date::format(format, date);
+  } else {
+    iso_date_time << date::format(format, tp);
+  }
```
nilsnolde, line 182: same here, just give back the seconds without doing timezone stuff if there's none

src/thor/route_action.cc
```diff
@@ -388,6 +388,14 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
       }
 
+      // add waiting_secs again from the final destination's datetime, so we output the departing time
+      // at intermediate locations, not the arrival time
+      if (destination->waiting_secs() && !destination->date_time().empty()) {
+        auto dest_dt =
+            offset_date(*reader, destination->date_time(), {}, destination->waiting_secs(), {});
+        destination->set_date_time(dest_dt);
+      }
+
```
nilsnolde, line 398: So, now it gets a bit more complex.. We're inside the `route_two_locations` lambda of the `arrive_by` algo and go through each alternate. We already set this round's `origin.date_time` to that of the destination + elapsed secs for this path. 

Remember, on the first round we have here `origin` = input intermediate and `destination` = input destination. So there's no `waiting_secs` defined as they're not allowed on input origin/destination. After that, the current `destination` can only be an intermediate input location and potentially have `waiting_secs`. And if there were, it's `date_time` is including the `waiting_secs`, which for the reverse path means it would be serialized in the response to the `arrive_by` time at that intermediate location. However, we want the `depart_at` time there, so we add the `waiting` secs again to the current time. The only other place a location's date_time is looked at again is in the response serializer.

src/thor/route_action.cc
```diff
@@ -446,6 +454,14 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
                               *origin, *destination, leg, algorithms, interrupt, edge_trimming,
                               intermediates);
+
+        // advance the time for the next destination (i.e. algo origin) by the waiting_secs
+        // of this origin (i.e. algo destination)
+        if (origin->waiting_secs()) {
+          // no need for timezones, we'll wait at this location
+          auto origin_dt = offset_date(*reader, origin->date_time(), {}, -origin->waiting_secs(), {});
+          origin->set_date_time(origin_dt);
+        }
```
nilsnolde, line 464: Still inside the same loop through all alternates, just further down after we built this path's `TripLeg` . After the first leg was built, this will advance the `origin`'s time by `waiting` secs, because the `origin` will become the next leg's destination. And in the reverse A* `destination` & `origin` are flipped, so this makes sure that the next reverse expansion starts with the correct time.

src/thor/route_action.cc
```diff
@@ -540,7 +561,8 @@ void thor_worker_t::path_depart_at(Api& api, const std::string& costing) {
       if (!origin->date_time().empty() && options.date_time_type() != valhalla::Options::invariant) {
         auto destination_dt =
             offset_date(*reader, origin->date_time(), temp_path.front().edgeid,
-                        temp_path.back().elapsed_cost.secs, temp_path.back().edgeid);
+                        temp_path.back().elapsed_cost.secs + destination->waiting_secs(),
+                        temp_path.back().edgeid);
```
nilsnolde, line 565: Here we're in the `depart_at`'s `route_two_location` lamdba and things are much easier since we're going forward, not reverse. We just have to add `waiting_secs` to the elapsed time. No other special treatment necessary.

src/thor/route_action.cc
```diff
@@ -663,22 +686,24 @@ std::string thor_worker_t::offset_date(GraphReader& reader,
...
-    in_tz = node->timezone();
-
-  // get the timezone of the output location
-  auto out_nodes = reader.GetDirectedEdgeNodes(out_edge, tile);
   uint32_t out_tz = 0;
-  if (const auto* node = reader.nodeinfo(out_nodes.first, tile))
-    out_tz = node->timezone();
-  else if (const auto* node = reader.nodeinfo(out_nodes.second, tile))
-    out_tz = node->timezone();
+  if (in_edge.Is_Valid() && out_edge.Is_Valid()) {
```
nilsnolde, line 691: We just try to get the timezones and use them if we actually passed nodes.

src/worker.cc
```diff
@@ -258,8 +259,9 @@ void parse_ring(ring_pbf_t* ring, const rapidjson::Value& coord_array) {
 
 void parse_location(valhalla::Location* location,
                     const rapidjson::Value& r_loc,
-                    Options& options,
-                    const boost::optional<bool>& ignore_closures) {
+                    Api& request,
```
nilsnolde, line 262: Need to pass the `request` now, we also need access to the `warnings`

src/worker.cc
```diff
@@ -429,6 +431,23 @@ void parse_location(valhalla::Location* location,
...
+
+  float waiting_secs = rapidjson::get<float>(r_loc, "/waiting", 0.f);
+  switch (location->type()) {
+    case Location_Type_kBreak:
+    case Location_Type_kBreakThrough:
+      // set waiting_time to 0 on origin/destination
+      {
+        auto loc_idx = location->correlation().original_index();
+        // TODO: waiting time can be less than 0
+        location->set_waiting_secs(loc_idx == 0 || is_last_loc || waiting_secs < 0.f ? 0.f
```
nilsnolde, line 443: We only set an actual `waiting_secs` time when it's
- not a origin
- not a destination
- and not negative

src/worker.cc
```diff
@@ -429,6 +431,23 @@ void parse_location(valhalla::Location* location,
...
+                                                                                     : waiting_secs);
+        break;
+      }
+    default:
+      if (waiting_secs)
+        add_warning(request, 203);
```
nilsnolde, line 449: We also warn if someone had a `through` or `via` location with `waiting` set.

src/worker.cc
```diff
@@ -472,20 +492,24 @@ void parse_locations(const rapidjson::Document& doc,
...
         rapidjson::get_optional<rapidjson::Value::ConstArray>(doc, std::string("/" + node).c_str());
     if (request_locations) {
+      uint32_t last_loc_idx = request_locations->Size() - 1;
       for (const auto& r_loc : *request_locations) {
         auto* loc = locations->Add();
-        loc->mutable_correlation()->set_original_index(locations->size() - 1);
-        parse_location(loc, r_loc, options, ignore_closures);
+        auto loc_idx = locations->size() - 1;
+        loc->mutable_correlation()->set_original_index(loc_idx);
+        parse_location(loc, r_loc, request, ignore_closures, loc_idx == last_loc_idx);
```
nilsnolde, line 500: We need to pass to `parse_location` the info whether this location is the last one.

test/gurka/test_time_tracking.cc
```diff
@@ -174,86 +174,82 @@ TEST(TimeTracking, reverse) {
...
     for (const auto& leg : route.legs()) {
       for (const auto& node : leg.node()) {
         times.push_back(node.cost().elapsed_cost().seconds());
       }
     }
   }
-  const std::vector<double> expected = {0,        17.1429,  34.2857,  51.4286,
-                                        193.9319, 245.4273, 269.4273, 293.4273};
-  ASSERT_THAT(times, testing::Pointwise(testing::DoubleNear(0.0001), expected));
+  test_result(times, result);
```
nilsnolde, line 218: I tested all things here:
- that it's outputting the right `waiting` time
- it's ignoring `through` and
- issues a warning

Only JSON output is not tested I guess.

kevinkreiser (approved):

PR #3856, opened 2022-12-01
add more options to valhalla_build_extract

### LINKED ISSUE #3858: Use arbitrary polygons to archive tiles
We’re about to add functionality to limit taring tiles with a bbox in #3856. Now I’d like to have GeoJSON functionality as well, to select tiles with arbitrary polygons, similar to what we’re already doing with valhalla_build_elevation.

nilsnolde:
So we can archive a subset of tiles from a directory giving either:
- bbox string or
- directory path with `.geojson` files (analogous to `valhalla_build_elevation`, fixes #3858)

### DIFF
--- a/scripts/valhalla_build_extract
+++ b/scripts/valhalla_build_extract
@@ -7,14 +7,13 @@ from io import BytesIO
 import json
 from math import floor
 import logging
-import os
 from pathlib import Path
 import struct
 import sys
 import tarfile
 from tarfile import BLOCKSIZE
 from time import time
-from typing import List, Tuple, Optional
+from typing import List, Tuple, Union, Set
 
 # "<" prefix means little-endian and no alignment
 # order is important! if uint64_t is not first, c++ will use padding bytes to unpack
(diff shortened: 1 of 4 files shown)


scripts/valhalla_build_extract
```diff
@@ -69,6 +81,39 @@ handler.setFormatter(logging.Formatter("%(asctime)s %(levelname)5s: %(message)s"
 LOGGER.addHandler(handler)
 
 
+def tile_intersects_bbox(tile_path: str, in_bounds: Bbox) -> bool:
+    """Check if a tile intersects the bbox"""
+    if not in_bounds:
+        return True
```
nilsnolde, line 87: make bbox optional

test/scripts/test_valhalla_build_extract.py
```diff
@@ -12,13 +17,54 @@
...
+
+        bbox = Bbox(-20, -59.2, -10.2, -53.9)
+        for input_tuple in (
+            (-20, -59, 2),  # barely intersecting in the lower left
+            (-12, -54, 1),  # contained in bbox
+            (-12, -62, 0),
+        ):
+            tile_path = tile_base_to_path(*input_tuple)
+            self.assertTrue(valhalla_build_extract.tile_intersects_bbox(tile_path, bbox), f"Tuple {input_tuple} is failing.")
+
```
nilsnolde, line 61: add failure case for tiles it can't intersect

kevinkreiser (approved):

PR #3860, opened 2022-12-07
make hgv-only edges also driveable to include in the reclassification logic

### LINKED ISSUE #3731: Truck - weird ferry issue
There's been an issue with truck routing where it takes the ferry or not depending on whether a waypoint is placed before or after a certain graph node. What's hard to understand here is that this node connect two pretty identical edges (in terms of FRC & name), so there's nothing hierarchy related going on.

takes ferry: https://valhalla.openstreetmap.de/directions?profile=truck&wps=12.199952602386476%2C54.12032571468474%2C11.806182861328127%2C55.45160500981926
image

doesn't take ferry: https://valhalla.openstreetmap.de/directions?profile=truck&wps=12.200242280960083%2C54.12061181812585%2C11.806182861328127%2C55.45160500981926
image

Any clues what might happen here? Note, with `auto` it works fine in both cases.

EDIT: in fact it takes a ferry in all cases, but in the latter case it takes an unreasonable detour.

nilsnolde:
fixes #3731 

I'm not entirely sure if this is fine for all downstream logic.. it seems not too much depends on this directly or indirectly. If someone has a concern, please voice it.

Also, again, IMO including truck here is enough, no need to respect more modes like motorbike.


gknisely (dismissed):

valhalla/mjolnir/node_expander.h
```diff
@@ -97,12 +97,13 @@ struct Edge {
       e.attributes.driveableforward = false;
       e.attributes.driveablereverse = false;
     } else {
-      e.attributes.driveableforward = way.auto_forward();
-      e.attributes.driveablereverse = way.auto_backward();
+      e.attributes.driveableforward = way.auto_forward() || way.truck_forward();
```
kevinkreiser, line 100: This doesn't seem right. If we have motorcar or hgv only ways the || will mark them as usable by both for the purposes of reclassification but in the end its possible they are exclusive. That's why I said in my comment the fix will be tricky.

kevinkreiser (changes requested):
Maybe I'm wrong but I think this is too broad. Let's discuss

nilsnolde:
This is finished. I'll comment more in the PR.

src/mjolnir/ferry_connections.cc
```diff
@@ -42,7 +44,8 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
                       sequence<Edge>& edges,
                       sequence<Node>& nodes,
                       const bool inbound,
-                      const uint32_t rc) {
+                      const uint32_t rc,
+                      uint32_t& mode_count) {
```
nilsnolde, line 48: this counts the total number of expansion we had to do over all nodes in the tileset

src/mjolnir/ferry_connections.cc
```diff
@@ -54,154 +55,185 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
...
+
+  // will be checked during expansion, start off with all access and set to missing modes after first
+  // iteration
+  uint32_t overall_access = 0;
+  uint32_t expansion_count = 0;
+  while (overall_access != baldr::kVehicularAccess || expansion_count < vehicle_modes) {
+    expansion_count += 1;
+
+    // which modes do we still have to find a path for?
+    uint32_t access_filter = baldr::kVehicularAccess ^ overall_access;
```
nilsnolde, line 72: we try to find new paths while we don't have an `overall_access` of `kVehicularAccess` or we had more tries than we have vehicular modes. We also define `access_filter` to prune the expansion on edges which don't fulfill our missing mode accesses.

nilsnolde: Lots of the other changes are just because there is an extra outer `while` loop now

nilsnolde: I did rename a few variables, but only for clarity while I was understanding the code myself.

src/mjolnir/ferry_connections.cc
```diff
@@ -54,154 +55,185 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
...
+        // Skip non-driveable edges and ones which are not accessible for the current access_filter (based on inbound flag)
+        bool forward = (edge.sourcenode_ == expand_node_idx);
+        if (forward) {
+          if ((inbound && !edge.attributes.driveablereverse) ||
+              (inbound && !(access_filter & edge.rev_access)) ||
+              (!inbound && !edge.attributes.driveableforward) ||
+              (!inbound && !(access_filter & edge.fwd_access))) {
+            continue;
+          }
+        } else {
```
nilsnolde, line 153: Prune the expansion if we don't have the right access

src/mjolnir/ferry_connections.cc
```diff
@@ -54,154 +55,185 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
...
+      for (auto& edge : bundle2.node_edges) {
+        bool forward = edge.first.sourcenode_ == pred_node;
+        if (forward || edge.first.targetnode_ == pred_node) {
+          sequence<Edge>::iterator element = edges[edge.second];
+          auto update_edge = *element;
+          if ((forward && inbound) || (!forward && !inbound)) {
+            path_access &= update_edge.rev_access;
+          } else if ((forward && !inbound) || (!forward && inbound)) {
+            path_access &= update_edge.fwd_access;
+          }
```
nilsnolde, line 216: remember the cumulative accesses for all vehicular modes along the found path with `path_access`

src/mjolnir/ferry_connections.cc
```diff
@@ -54,154 +55,185 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
...
-      break;
+      // Get the predecessor - break once we have found the start node
+      if (pred_node == node_idx) {
+        break;
+      }
+      label_idx = node_status[pred_node].index;
     }
-    index = node_status[pred_node].index;
+    // update the overall mode access with the previous path
+    overall_access |= path_access;
```
nilsnolde, line 233: Update the `overall_access` with what mode access we found and maybe stop the `while` loop early

src/worker.cc
```diff
@@ -189,25 +190,10 @@ bool add_date_to_locations(Options& options,
...
-              loc.set_date_time(options.date_time());
-            }
-          }
-          break;
-        default:
-          break;
+      if (node == (dt == Options::arrive_by ? "targets" : "sources")) {
+        for (auto& loc : locations) {
+          loc.set_date_time(dt == Options::current ? "current" : options.date_time());
+        }
```
nilsnolde, line 196: just removing a few lines

test/gurka/test_ferry_connections.cc
```diff
@@ -3,53 +3,75 @@
...
 
-std::pair<baldr::RoadClass, baldr::RoadClass>
-ReclassifyFerryConnectionEdge(std::map<std::string, std::string> const& way_description) {
-  constexpr double gridsize_metres = 1000;
-
-  const std::string ascii_map = R"(
-          A--B--b--C----D--E
-          F--G--g--H----I--J
+class FerryTest : public ::testing::TestWithParam<std::string> {
+protected:
```
nilsnolde, line 7: I made this a parameterized test class with all vehicular modes as parameters

test/gurka/test_ferry_connections.cc
```diff
@@ -128,44 +150,35 @@ TEST(Standalone, TruckFerryDuration) {
...
+                                                        "unclassified",   "service",
+                                                        "secondary_link", "tertiary_link"};
+  for (const auto& cls : reclassifiable_ways) {
+    EXPECT_TRUE(edges_were_reclassified({{"highway", cls}}, GetParam()));
+  }
+}
+
+INSTANTIATE_TEST_SUITE_P(FerryConnectionTest,
+                         FerryTest,
+                         ::testing::Values("motorcar", "hgv", "moped", "motorcycle", "taxi", "bus"));
```
nilsnolde, line 184: this new test is the only using the parameterization and tests all highway types and all vehicular modes

valhalla/mjolnir/node_expander.h
```diff
@@ -89,20 +94,27 @@ struct Edge {
                         const uint32_t llindex,
                         const OSMWay& way,
                         const bool infer_turn_channels) {
+    // TODO(nils): include a "motorvehicle_fwd/rev" in lua/OSMWay?
+    bool drive_fwd = way.auto_forward() || way.truck_forward() || way.bus_forward() ||
+                     way.moped_forward() || way.motorcycle_forward() || way.emergency_forward() ||
+                     way.hov_forward() || way.taxi_forward();
+    bool drive_rev = way.auto_backward() || way.truck_backward() || way.bus_backward() ||
+                     way.moped_backward() || way.motorcycle_backward() || way.emergency_backward() ||
+                     way.hov_backward() || way.taxi_backward();
```
nilsnolde, line 103: include all edges which allow some vehicular mode

nilsnolde:
I first committed the test with https://github.com/valhalla/valhalla/pull/3860/commits/79f6ea01c80c36cc929c0b7920308933b780fb9a, so it's easy to just run it with current master.

nilsnolde:
Hm.. Wanted to run it on the test data of #3731 but on a 200 MB PBF took > 5 mins reclassifying ferry edges before I canceled it.. The whole planet takes that long with `master`.. Must somehow get stuck in a loop, can't see where though.. We can look at it together tmrw.

valhalla/mjolnir/node_expander.h
```diff
@@ -70,6 +69,10 @@ struct Edge {
   // index of the target (end) node of the edge
   uint32_t targetnode_;
 
+  // to record the access of an edge
+  uint16_t fwd_access;
+  uint16_t rev_access;
```
nilsnolde, line 74: I removed the `driveableforward`  & `driveablereverse` booleans and replaced all logic depending on them with the access masks which are only set if there's no `Use::EmergencyAccess` and has access to a vehicular mode. 

I guess these would fit still into the `attributes`  bit field, now that there's 25 bits available, though then only 1 would be left (access has 12 bits in directed edges) and we'd need to re-write to this if we needed more attributes. So I'd leave it this way.

src/mjolnir/node_expander.cc
```diff
@@ -14,18 +14,23 @@ node_bundle collect_node_edges(const sequence<Node>::iterator& node_itr,
   for (; itr != nodes.end() && (node = *itr).node.osmid_ == bundle.node.osmid_; ++itr) {
     ++bundle.node_count;
     if (node.is_start()) {
-      auto edge_itr = edges[node.start_of];
-      auto edge = *edge_itr;
+      auto edge = *edges[node.start_of];
+      uint16_t auto_fwd_access = edge.fwd_access & baldr::kAutoAccess;
```
nilsnolde, line 18: change to bool

kevinkreiser (approved):

kevinkreiser (approved):

nilsnolde:
OSX failed on the `mapmatch` test for some reason.. can't reproduce locally, so I kicked CI again

PR #3864, opened 2022-12-09
fix valhalla_build_extract 

nilsnolde:
Should've done a full end to end test. Just wanted to use it and there was a bug, where it tried to derive the tile ID & level from the full path, rather than the relative tile path.


scripts/valhalla_build_extract
```diff
@@ -143,15 +147,16 @@ def get_tiles_with_geojson(all_tile_paths: List[Path], geojson_dir: Path) -> Set
     tile_paths_ = set()
     geojson_polys = get_outer_rings(geojson_dir)
     for tile_path in all_tile_paths:
-        tile_bbox = box(*get_tile_bbox(tile_path))
+        tile_path_id = str(tile_path.relative_to(tiles_dir_))
```
nilsnolde, line 150: this is the important line: just forward the relative tile path to the `get_tile_bbox` function

kevinkreiser (dismissed):

kevinkreiser (approved):

PR #3865, opened 2022-12-09
fix hazmat

### LINKED ISSUE #3861: Truck-Routing with hazardous material
I'm having some issues with truck routing. Here is an example case:

Route start: 6-8, Rombacher Hütte, Weitmar, Bochum-Mitte, Bochum, Nordrhein-Westfalen, 44795, Deutschland
Route end: Kohlenstraße, Weitmar, Bochum-Mitte, Bochum, Nordrhein-Westfalen, 44795, Deutschland

When using Valhalla and truck routing, there is no possible route found:
grafik

After enabling hazardous materials, a route can be found:
grafik

The destination of my route has the tag `hazmat=destination`. I think Valhalla is not dealing properly with the hazmat tag. 

I've already checked the code in the truckcost.cc file and found this:
```
bool TruckCost::ModeSpecificAllowed(const baldr::AccessRestriction& restriction) const {
  switch (restriction.type()) {
    case AccessType::kHazmat:
      if (hazmat_ != restriction.value()) {
        return false;
      }
      break;
...
```
I guess this part of the function should only return false, if there is hazardous material on the vehicle **and** the street has an hazmat tag. If `hazmat_` is false it should never return false.

I neither know if my assumptions are correct nor how to properly fix this. Would be great if somebody could look into it.

nilsnolde:
fixes #3861 

Will comment in the PR.

### DIFF
--- a/lua/graph.lua
+++ b/lua/graph.lua
@@ -281,8 +281,8 @@ hazmat = {
 ["designated"] = "true",
 ["yes"] = "true",
 ["no"] = "false",
-["destination"] = "true",
-["delivery"] = "true"
+["destination"] = "false",
+["delivery"] = "false"
 }
 
 shoulder = {
--- a/src/sif/truckcost.cc
+++ b/src/sif/truckcost.cc
@@ -381,7 +381,7 @@ bool TruckCost::AllowMultiPass() const {
 bool TruckCost::ModeSpecificAllowed(const baldr::AccessRestriction& restriction) const {
   switch (restriction.type()) {
     case AccessType::kHazmat:
-      if (hazmat_ != restriction.value()) {
+      if (hazmat_ && hazmat_ != restriction.value()) {
         return false;
       }
       break;
--- a/src/sif/truckcost.cc
+++ b/src/sif/truckcost.cc
@@ -381,7 +381,7 @@ bool TruckCost::AllowMultiPass() const {
 bool TruckCost::ModeSpecificAllowed(const baldr::AccessRestriction& restriction) const {
   switch (restriction.type()) {
     case AccessType::kHazmat:
-      if (hazmat_ && hazmat_ != restriction.value()) {
+      if (hazmat_ && !restriction.value()) {
         return false;
       }
       break;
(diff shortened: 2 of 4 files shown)


lua/graph.lua
```diff
@@ -281,8 +281,8 @@ hazmat = {
 ["designated"] = "true",
 ["yes"] = "true",
 ["no"] = "false",
-["destination"] = "true",
-["delivery"] = "true"
+["destination"] = "false",
+["delivery"] = "false"
```
nilsnolde, line 285: IMHO we shouldn't allow trucks on roads with `hazmat=destination`, as we always allow starting/ending on a `hazmat=no` road, so for us it actually has no meaning.

src/sif/truckcost.cc
```diff
@@ -381,7 +381,7 @@ bool TruckCost::AllowMultiPass() const {
 bool TruckCost::ModeSpecificAllowed(const baldr::AccessRestriction& restriction) const {
   switch (restriction.type()) {
     case AccessType::kHazmat:
-      if (hazmat_ != restriction.value()) {
+      if (hazmat_ && hazmat_ != restriction.value()) {
```
nilsnolde, line 384: first check if `hazmat` is even turned on in the request before comparing it, since otherwise would not allow a combo like `hazmat: false` in the request but `hazmat: true` on the road.

src/sif/truckcost.cc
```diff
@@ -381,7 +381,7 @@ bool TruckCost::AllowMultiPass() const {
 bool TruckCost::ModeSpecificAllowed(const baldr::AccessRestriction& restriction) const {
   switch (restriction.type()) {
     case AccessType::kHazmat:
-      if (hazmat_ != restriction.value()) {
+      if (hazmat_ && !restriction.value()) {
```
nilsnolde, line 384: first check if hazmat is even turned on in the request before comparing it, since otherwise would not allow a combo like hazmat: false in the request but hazmat: true on the road.

nilsnolde:
Eventually I'd like to be able to categorize a bit more the hazmat types (A - E). AFAICT it doesn't need any change to any structures (no bits added), only the lua & access restriction logic would need to change a bit. But I think it's fairly easily doable.

kevinkreiser:
We would have to store the types somewhere, presumably in the restriction, so that structure would need to change to accommodate it, but I agree we can trivially stuff it in the value as a mask. Iirc there are 9 classifications that the hundreds of placards can fall into. We have 64 bits so 9 easily fits. Then in request processing we could allow specifying the placard and we can turn that into a type/category. Anyway seems totally doable

kevinkreiser (approved):

PR #3873, opened 2022-12-22
add inline config to some more exe's

nilsnolde:
For scriptability it's nicer to have inline config rather than a static JSON.


kevinkreiser (approved):

Blagodarenko:
Hi! return of function ParseArguments() is always false (or EXIT_FAILURE).
@nilsnolde , seems you skip _return true_ after https://github.com/valhalla/valhalla/blob/master/src/mjolnir/valhalla_ways_to_edges.cc#L73 or need to change [_return false_](https://github.com/valhalla/valhalla/blob/master/src/mjolnir/valhalla_ways_to_edges.cc#L82) to _true_

nilsnolde:
Ooops :sweat_smile: It looks like you're right. I'm not using that executable since a longer time anymore. Do you want to PR?

Blagodarenko:
> Do you want to PR?
No, thanks 😄 

and please check [valhalla_build_admins](https://github.com/valhalla/valhalla/blob/master/src/mjolnir/valhalla_build_admins.cc#L61), seems for "config" returns _true_, but for "inline-config" - always _false_   (i didn't check it)

nilsnolde:
Haha ok, then you'll have to wait for someone to find the time to take care of this.

> and please check [valhalla_build_admins](https://github.com/valhalla/valhalla/blob/master/src/mjolnir/valhalla_build_admins.cc#L61),

doesn't seem to be a problem.

PR #3876, opened 2022-12-23
GraphReader with write-access to traffic extract

PhilippVoigt:
For the purpose of live-traffic updates, write-access to traffic tiles is required. Instead of relying on a completely separate implementation, the traffic update mechanism could benefit from all the optimizations available in GraphReader by allowing read-only mode to be disabled.

The approach was working well for our use case, I would be interested to hear your thoughts on this and potential problems this might lead to. Thanks!

Licensed under MIT


kevinkreiser:
I dont think it poses any real problem so long as the way you are using it in practice is from another process which keeps the file up to date separately from the running service which still maps it read only.

kevinkreiser (approved):
@PhilippVoigt can you add a changelog entry to the bottom of the enhancements list?

kevinkreiser (approved):

Issue #3877, opened 2022-12-26
Move to C++17


nilsnolde:
As discussed in length over in #2009, the consensus is to at least move to C++17, before moving on to C++20 (find arguments in the linked issue).

I went through the improvements of C++14 to C++17 and collected some obvious TODOs for us (well, rather me, never wrote anything in 17). Feel free to edit this issue if you’re allowed, or just comment, I’ll add it to this top level list to keep it comprehensive.

- [ ] (nils) `boost::optional` -> `std::optional`
- [ ] (nils) `std::filesystem` maybe to replace our home baked one (at least to some degree)
- [ ] (nils) not sure if we use variants? Don’t think so actually.. 
- [ ] (nils) there’s a new `std::clamp` which sounds like it could remove our default range struct for input param processing
- [ ] (nils) `scoped_lock` replaces `lock_guard` and is variadic now
- [ ] (molind) make use of `string_view` (eg for locales), can be good for constexpr. Generally keep using `const string&` if the arg is used for functions expecting std::string (expensive conversion) or C strings (null termination not guaranteed)

And a few things I noticed which we might want to watch out for:
- emplace for sequences now returns a reference to the new element
- string::data() can return mutable pointer to the char 
- `shared_ptr` for arrays now available

kevinkreiser:
We use variant for the homemade json serializer and the updated emplace semantics are a very welcome change we can take advantage of in many places

mloskot:
What is the plan?
How about starting with switching CI builds to C++17?

1. Switch CMake and CI configurations to C++17.
2. Make the CI builds green applying bare minimum fixes, if any necessary.
3. Move from there with porting the code base to C++17 features.

nilsnolde:
Sounds good. Had some train time (ie no internet, good ol’ Germany..), was doing most of the optional stuff already. I can try CI later of you didn’t by then.

PhilippVoigt:
Thanks for considering us in this matter. We are running fully in Docker, so anything will work on our end.
To be honest, I'm not familiar enough with C++ development to even be aware of the latest features, but `std::filesystem` is something we already stumbled across so I fully support this move!

PR #3878, opened 2022-12-26
CI C++17

nilsnolde:
start with #3877 , still wait for more feedback. more a quick test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-## Release Date: 2022-??-?? Valhalla 3.3.0
+## Release Date: 2022-??-?? Valhalla 3.3.1
+* **Removed**
+* **Bug Fix**
+* **Enhancement**
+
+## Release Date: 2022-01-03 Valhalla 3.3.0
 * **Removed**
 * **Bug Fix**
 * **Enhancement**


nilsnolde:
hm, of course some `-WError` kicks in. 

on linux, seems like the `date` library has an `uncaught_exception` warning which some of our source files catch onto. it was apparently [fixed for windows](https://github.com/HowardHinnant/date/commit/de6a03d337dcd617a121254f700979086889c0a4), but seems to be an issue for unix as well, as [someone says](https://github.com/HowardHinnant/date/issues/338#issuecomment-485915937) it should say `if __cplusplus > 201703` [here](https://github.com/HowardHinnant/date/blob/8a563041fa65146531249ed557a6839809488521/include/date/date.h#L136), at least that's what locally works. worth warming up probably and PR upstream, right?

on macos, there's a few others in meili mostly apparently, mostly `error: 'iterator<std::forward_iterator_tag, const valhalla::meili::Label>' is deprecated [-Werror,-Wdeprecated-declarations]`. at least that's in our control.. but also some in libosmium with `error: 'result_of<osmium::io::detail::SerializeBlob ()>' is deprecated [-Werror,-Wdeprecated-declarations]`, which doesn't seem to be reported anywhere upstream. kinda weird, considering c++17 is in wide use, I'd have expected someone else to build with `-Werror`. but then again, seems to be macos(/xcode?) issue, builds fine on linux locally with the quick fix to overcome the `date` issue, hope CI will agree.

nilsnolde:
ah good, the libosmium issue was already fixed in https://github.com/osmcode/libosmium/commit/e933414c191f2209e9869247d2b2eede02f77c86. and was release with [2.17.2](https://github.com/osmcode/libosmium/releases/tag/v2.17.2). we're on 2.16.0. so I guess moving up the version should solve that. might as well go all the way to 2.18.0 (lots of deprecated stuff was removed there, but 2.17.x should work for us).

nilsnolde:
Back on my office machine, it also failed with gcc 12.2.0 with the same std::iterator deprecations when building the tests. Strange it works on CI Linux machines. There's a nice post about the why: https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/. I (tentatively) fixed it quickly and I could build the test locally.

Also upgraded to the latest libosmium release which went pretty smooth.

valhalla/meili/routing.h
```diff
@@ -305,8 +305,14 @@ find_shortest_path(baldr::GraphReader& reader,
...
 
 // Route path iterator. Methods to assist recovering route paths from Labels.
-class RoutePathIterator : public std::iterator<std::forward_iterator_tag, const Label> {
+class RoutePathIterator {
 public:
+  using iterator_category = std::forward_iterator_tag;
+  using value_type = const Label;
+  using difference_type = const std::ptrdiff_t;
+  using pointer = const Label*;
+  using reference = const Label&;
```
nilsnolde, line 314: basically what `std::iterator` was doing under the hood (also the only thing it was really doing).

nilsnolde: let's reason over the constness together..

kevinkreiser (approved):

Issue #3797, opened 2022-10-19
Update README with more (and less) info


nilsnolde:
I'm thinking out loud a bit haha let me know if smth is off-limits.

The README is a bit overwhelming IMO, especially for new users. And some stuff could be added maybe. But overall I'd like to slim down the main README a bit:
- outsource the build instructions, also re-structure a bit, maybe one small document per platform (linux, mac, win)
- CONTRIBUTING document to outsource the Contributing section with a bit of a rewrite maybe
- adding a section on translations & transifex for better visibility
- adding a section about "Related Projects" would make sense to me. E.g. we (gis-ops) have quite a few libraries/projects that are entirely Valhalla specific (the packaged Python bindings, a docker abstraction, web app, soon an awesome QGIS Plugin:)). It's not too hard to google "valhalla python bindings", but it is if you don't know what to look for. The Valhalla repo is often the entrypoint for people I think and those (and others) projects add a lot of value for the main project and also its users. I'm not aware of (m)any other Valhalla specific projects tbh other than openstreetmapspeeds (we should highlight that more as well), honestly, I don't want to make this about us, let me know if there are others

nilsnolde:
there's another new valhalla-specific project for k8s: https://github.com/valhalla/valhalla/discussions/3802

PR #3880, opened 2022-12-28
update locales

nilsnolde:
Transifex is a bit of a pain..

Most languages still miss the indoor stuff, I added it to German & English. I also wonder what to do with the other Transifex resource `osrmti.json`.. Is that somehow integrated in the Valhalla locale files or OSRM maneuvers just don't get translated so far?


locales/README.md
```diff
@@ -55,6 +55,8 @@ Czech `aliases` entry example:
 
 #### Adding new instruction strings
 
+TODO(nils): I don't think that's how it's supposed to work (anymore?). If I understand correctly, we'd upload a new `en_US.json` to Transifex and that'll take care of copying the new stuff to existing translation files.
```
nilsnolde, line 58: Smth might've changed here too..

locales/README.md
```diff
@@ -92,7 +94,13 @@ To learn more about the Config command, visit https://docs.transifex.com/client/
 
 ### Pulling translation updates from Transifex
 
-Use this command to pull all changes from Transifex into the Valhalla repo.
+Pull acceptable translation files with 16 threads:
+
+```
+tx pull --minimum-perc 95 --workers 16
+```
```
nilsnolde, line 101: transifex api is soooo sloooowwww...

locales/bg-BG.json
```diff
@@ -1149,8 +1149,8 @@
       },
       "example_phrases": {
         "0": [
-          "Пътувайте за 1 спирка.",
-          "Пътувайте за 3 спирки."
+          "Travel 1 stop.",
+          "Travel 3 stops."
```
nilsnolde, line 1153: for some reason, some `example_phrases` were reverted to the original (as it should be). however, most of them in most files are also translated..

locales/de-DE.json
```diff
@@ -2175,8 +2175,8 @@
     },
     "elevator": {
       "phrases": {
-        "0": "Take the elevator.",
-        "1": "Take the elevator to <LEVEL>."
```
nilsnolde, line 2179: Isn't this missing the string "level"? As in `Take the elevator to level <LEVEL>.`

locales/de-DE.json
```diff
@@ -2203,8 +2203,8 @@
     },
     "escalator": {
       "phrases": {
-        "0": "Take the escalator.",
-        "1": "Take the escalator to <LEVEL>."
```
nilsnolde, line 2207: same here, ref https://github.com/valhalla/valhalla/pull/3880/files#r1058484184

nilsnolde:
Since some stuff was added that wasn't standard-compliant and tests failed, I went through the entire changeset and can say to 99.9% that there's no obvious structural errors (can't say much about the quality of most languages). Lots of good additions since the last update (10 months ago), esp Russian benefited a lot it seems.

nilsnolde:
I also feel like we should make it more prominent that we're hosting locales on transifex. Currently there's no place other than the `locales` directory's README. It should go fairly highly up in the root README IMHO, I'll add it to #3797 .

locales/de-DE.json
michaelkirk: It looks like maybe it's hardcoded in https://github.com/valhalla/valhalla/blob/master/src/odin/enhancedtrippath.cc#L535, though that doesn't look to be localized. I'm not very familiar with this codebase though.

locales/en-US-x-pirate.json
```diff
@@ -1120,7 +1120,7 @@
       ],
       "example_phrases": {
         "0": [
-          "Avast ye, merge."
```
michaelkirk, line 1123: 😮🏴‍☠️

locales/de-DE.json
nilsnolde: Thx for the code hint, was too lazy to look it up so far, just had a harder look. You're right, the string "Level" is not localized. I guess that's because it's not necessarily referring to a numbered level, but could be just a string with the `level:ref=<string>` OSM tag, instead of the usual `level=<int>` tag. Guess that could be added to the localization and processing it, but not sure it's really worth the effort? Anyways, had a bit deeper glimpse into the narration code, which is good:)

kevinkreiser: Yeah I think the general idea is that it could be any name not just a number or something so putting level there all the time might not always make sense. To handle both cases properly we would need to detect the presence of a number like name which we could do for sure at some point

locales/fr-FR.json
```diff
@@ -531,7 +531,7 @@
       },
       "ordinal_values": [
         "1er",
-        "2e",
+        "2nd",
```
kevinkreiser, line 534: revert

locales/pl-PL.json
```diff
@@ -175,11 +175,11 @@
...
         "ścieżka rowerowa",
         "górski szlak rowerowy",
-        "the crosswalk"
+        "przejście dla pieszych"
       ],
       "metric_lengths": [
         "<KILOMETERS> km",
-        "1 km",
+        "1 kilometr",
         "<METERS> m",
```
kevinkreiser, line 183: since t his is verbal we probably shouldnt have any abbreviations.. i guess we'll get it later

locales/ru-RU.json
```diff
@@ -415,22 +415,22 @@
     },
     "enter_roundabout": {
       "phrases": {
-        "0": "Выезжайте на кольцевую развязку.<instructions.enter_roundabout.phrases.1>Выезжайте на кольцевую развязку и сверните на <ORDINAL_VALUE> съезде.",
```
kevinkreiser, line 418: its interesting that this doesnt throw an error in the unit tests. we should make a new issue to harden the tests to this. no only should we find the tags in the phrase that we expect we should also not find additional tags in the phrase that shouldnt be there. in this case someone copy past failed two phrases together... but a stronger unit test would catch this.

locales/ru-RU.json
```diff
@@ -530,16 +530,16 @@
...
+        "первом",
+        "втором",
+        "третьем",
+        "четвёртом",
+        "пятом",
+        "шестом",
+        "седьмом",
+        "восьмом",
+        "девятом",
+        "десятом"
```
kevinkreiser, line 542: very good, this person gets that "verbal" means you spell it out so text to speech can read it properly

locales/fr-FR.json
nilsnolde: oops, I thought I caught them all. was changing them manually from transifex's output.

kevinkreiser (approved):

kevinkreiser:
lets get #3585 merged first though so we dont have to update that external forks master branch

locales/fr-FR.json
nilsnolde: done. also caught another "2nd" in a phrase, changed that too

kevinkreiser (approved):

molind:
I've found few mistakes in Russian and fixed them in Transifex. 

But also I noticed that hardly could be fixed now. In Russian nouns use different endings after different verb. E.g.
```
"exit_roundabout": {
  "phrases": {
    "1": "Exit the roundabout onto <STREET_NAMES>.",
    "2": "Exit the roundabout onto <BEGIN_STREET_NAMES>. Continue on <STREET_NAMES>.",

"exit_roundabout": {
  "phrases": {
    "1": "Сверните с кольцевой развязки на <STREET_NAMES>.",
    "2": "Сверните с кольцевой развязки на <BEGIN_STREET_NAMES>. Продолжите движение по <STREET_NAMES>.",
```
In that case verb Exit **Сверните**, a noun after it should be used with endings for question "where? на что?" - **на пешеходный переход** 
But in case of verb Continue **Продолжайте**, a noun after it should be used with endings for question "where? по чем?" - **по пешеходному переводу**.

It makes exit_roundabout difficult to translate correctly because same placeholder <STREET_NAMES> is used after different verbs.

molind:
Sometimes there should be more plural forms than in English. I'm not sure how to add them in Transifex. e.g.       
```
"transit_stop_count_labels": {
  "one": "stop",
  "other": "stops"
},
```
would be
```
"transit_stop_count_labels": {
  "one": "остановка",
  "few": "остановки",
  "many": "остановок",
  "other": "остановки"
},
```

molind:
Sure. I'm ready to help. Created #3936 and #3937

PR #3887, opened 2023-01-02
Added reverse lookup to shortcut recovery (edge -> shortcut)

PhilippVoigt:
We've been working on adding a reverse lookup to the existing shortcut recovery cache in order to quickly identify which shortcut an edge is part of. Initially, we were hoping for a significant performance boost by doing this at the cost of some extra memory. However, for our use case it did not provide much benefits.

I'd still like to share this because by testing the approach against the current implementation, I became quite convinced that something is not right with the way `GraphReader::GetShortcut` is currently implemented.
For my test, I iterate all edges in the entire graph and call `GetShortcut` for each of them, then collect the shortcuts in a set. It seems that the current implementation returns regular edges instead of shortcuts in many cases.
Don't mind the fact that I shifted the shortcut cache into the graph reader, I was trying to do minimal refactoring for now just to have the recovery cache as a non-static member of GraphReader for testing.

My PR is more of a workaround which we use for now, but not a fix for the issue. `shortcut_recovery_t::find_shortcut` is still implemented as before in case caching is disabled. Unfortunately, my understanding of the underlying graph structures is still a little limited so I'm unable to figure out if this is a bug or if my test is invalid.

Looking forward to some suggestions. Also if you would think such a cache could be beneficial in general. Thanks!


kevinkreiser:
> It seems that the current implementation returns regular edges instead of shortcuts in many cases.
it would be awesome to get a unit test showing this happening. i would add a test here https://github.com/valhalla/valhalla/blob/f378eee955a17441d077a3a6485511c131fe5899/test/recover_shortcut.cc that throws when it finds a regular edge as you have noted. with any luck utrecht will have such degenerate cases and we can debug the root problem

regarding caching the inverse mapping of edge to shortcut i can't immediately see why that would be useful but if its useful to you and we can properly unit test it then it'd be fine to have it in case others would need it as well. if you wouldnt mind explaining your use-case for it that would be great but if you can't thats also completely understandable.

PhilippVoigt:
@kevinkreiser I've written some unit tests using Utrecht where the problem becomes immediately obvious. I created a new issue #3888 for this.

As far as our use case goes, we are incorporating live edge speeds, a few at a time and hope to do so with minimal overhead. Since each edge update comes with a potential shortcut update, this was one puzzle piece to reduce iterations. In combination with other optimizations, it's not the biggest factor anymore so I would close this for now.

kevinkreiser:
ah yes that is a great use-case. in previous implementations ive just created a shared cache for the infrastructure that does the traffic association but keeping it in graphreader could be useful as long as its a threadsafe cache all threads can make use of of course (or its something we pre-populate on first access of the singleton). anyway lets go tackle that issue as well. thank you for creating these!

PR #3890, opened 2023-01-04
Replace boost::optional with std::optional

nilsnolde:
It's unfortunately not really a simple drop-in replacement, quite a few things differ. Apart from the namespace changes, the following patterns keep repeating in the diff:

- `boost::none` -> `std::nullopt` 
- `make_optional(bool, value)` ctor doesn't exist for std, so we need to check the value first and then assign to `std::nullopt` if it's not initialized
- `value.is_initialized()` doesn't exist in std, instead compare to `std::nullopt`
- EDIT: I should also mention that I cleaned up a few `#include` orders in some files (3 blocks in order: std > dependencies > valhalla), that might be confusing when looking at the diffs. I'd revert it if I could, just to make the diff more focused, but I can't, sorry for that..

Also `baldr/rapidjson_utils` still needs `boost::optional` actually for rapidjson, so json related things still keep using `boost::optional`, which makes 1) it somewhat confusing and 2) me sorta regret this whole PR..


kevinkreiser:
> Also baldr/rapidjson_utils still needs boost::optional actually for rapidjson

i think im misunderstanding you here. i thought the reason rapidjson_utils still needs `boost::optional` is because we used to use only `boost::property_tree::ptree` to handle json. because ptree is kind of not well-suited for json i ported a lot of the ptree-like functions we had all over the code to using rapidjson utils but i kept the interface the same, ie using boost::optional. i dont see why we cant actuall change this to std::optional. rapidjson itself doesnt use it it was just a way to make it "work like ptree" to avoid refactoring all of the code that relied on ptree.

src/odin/markup_formatter.cc
```diff
@@ -60,13 +60,13 @@ MarkupFormatter::FormatPhonemeElement(const std::unique_ptr<baldr::StreetName>&
           FormatPhonemeElement(street_name->value(), street_name->pronunciation());
 
       // If the markup string exists then return the street name with the phoneme
-      return boost::make_optional(!phoneme_markup_string.empty(), phoneme_markup_string);
+      return phoneme_markup_string.empty() ? std::nullopt : std::make_optional(phoneme_markup_string);
```
kevinkreiser, line 63: is it actually necessary to call `make_optional`? i think we can just return the string and it will be implicitly converted (shorter code for the win)

kevinkreiser: or maybe because its in a ternary it has to have the same type as the other part of the ternary.. mental compiler says probably that

src/worker.cc
```diff
@@ -17,6 +17,7 @@
 #include "thor/worker.h"
 #include "worker.h"
 
+#include <boost/optional.hpp>
```
kevinkreiser, line 20: whats with this? removing it elsewhere made us have to put it in here i guess?

kevinkreiser (approved):

PR #3892, opened 2023-01-05
add some more goodies to mkdocs navigation

nilsnolde:
Only for UX:
- Show sub-sections of top-level leafs in the sidebar
- show "Back to Top" button when scrolling up further down in the document
- TOC follows down in the page, if the TOC would be really long

~~It's already live on https://valhalla.github.io/valhalla/ in case you wanna have a look.~~ new commits in `master` overwrote that.


nilsnolde:
I pushed it again to the docs URL: https://valhalla.github.io/valhalla/. It'll stay as long as nothing is pushed to master.

What do you think? Worth merging?

nilsnolde:
Should we merge it @kevinkreiser ? The build failure is due to loki_service test timeout. I think that happens here and there. There's an `alarm(180)` on that test suite.. Just skimming the code, it doesn't seem to do smth that'd warrant 3 mins.. Is it because the service it starts in the background is somehow not stopped properly after the tests finished sometimes?

kevinkreiser (approved):

PR #3893, opened 2023-01-06
Parse lit tag

### LINKED ISSUE #3848: Support `lit` OSM key
For pedestrian safety.

Easy:
- parse boolean from OSM into the graph
- add one bit on directed edge for lit

Medium-easy:
- decide how to use it in EdgeCost of pedestrian, e.g.
  -  `use_lit` costing option and use it directly?
  - `safe` costing option and wrap the `lit` attribute with others such as road class penalties?
  - other ideas?

chrstnbwnkl:
fixes #3848 

This PR adds parsing of the `lit` tag:

  - parse the possible values and map them to a boolean value in `graph.lua` 
  - add `lit` as a member variable to `OSMWay`
  - add `lit` as a member variable to the `DirectedEdge` class, thereby allocating one of the remaining bits
  - add the `lit` tag and the values we parse to `taginfo.json`
  
I also added a short parameterized test going through the different possible values of `lit`, including its absence.

### DIFF
--- a/lua/graph.lua
+++ b/lua/graph.lua
@@ -408,6 +408,17 @@ toll = {
 ["snowmobile"] = "true"
 }
 
+lit = {
+  ["yes"] = "true",
+  ["no"] = "false",
+  ["24/7"] = "true",
+  ["automatic"] = "true",
+  ["limited"] = "false",
+  ["disused"] = "false",
+  ["dusk-dawn"] = "true",
+  ["sunset-sunrise"] = "true"
+}
+
 --node proc needs the same info as above but in the form of a mask so duplicate..
 motor_vehicle_node = {
 ["yes"] = 1,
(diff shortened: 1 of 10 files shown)


lua/graph.lua
```diff
@@ -408,6 +408,17 @@ toll = {
 ["snowmobile"] = "true"
 }
 
+lit = {
+  ["yes"] = "true",
+  ["no"] = "false",
+  ["24/7"] = "true",
+  ["automatic"] = "true",
+  ["limited"] = "false",
```
chrstnbwnkl, line 416: `lit=limited` is a special case that, according to the OSM wiki mainly occurs in Austria and Germany, where it signifies that a street light marked with a ["Laternenring"](https://de.wikipedia.org/wiki/Laternenring) is not turned on during the whole night, so I guess we cannot reliably set the marked edge as `lit=true`

nilsnolde:
Yay thanks! Got my approval, but someone else should review and approve:)

kevinkreiser (changes requested):
so other than the change to osmway.h this looks good, we'll ship it after we get that fixed up. thanks!

kevinkreiser (approved):

PR #3894, opened 2023-01-06
Slightly optimize CI builds

nilsnolde:
I just had a case where clang-tidy was failing twice in a row on the same line, I forgot to reference instead of copy on top applying `const`. Then I noticed that the clang-tidy runs after the whole build is already done. Might be accidental, but I think it should be run before the build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,13 +41,14 @@ jobs:
     resource_class: xlarge
     steps:
       - checkout
-      - install_linux_dependencies
       - run: |
             if ! ./scripts/needs_ci_run; then
                 echo "Changes in last commit do not need CI. Skipping step"
                 circleci-agent step halt
             fi
       - run: sudo python3 -m pip install pre-commit && pre-commit run --all-files && ./scripts/error_on_dirty.sh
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -41,13 +41,14 @@ jobs:
...
       - checkout
-      - install_linux_dependencies
       - run: |
             if ! ./scripts/needs_ci_run; then
                 echo "Changes in last commit do not need CI. Skipping step"
                 circleci-agent step halt
             fi
       - run: sudo python3 -m pip install pre-commit && pre-commit run --all-files && ./scripts/error_on_dirty.sh
+      - run: scripts/clang-tidy-only-diff.sh 4
+      - install_linux_dependencies
```
nilsnolde, line 51: installing deps can also be done after statically checking the code

nilsnolde:
I totally misjudged how clang-tidy works, it needs the build for the linting. I assumed it was text-based analysis I guess, but being on clang it makes sense :sweat_smile: 

Installing the dependencies after checking the formatting still makes sense though.

kevinkreiser (approved):

PR #3896, opened 2023-01-07
re-ordered the factors in EdgeCost

### LINKED ISSUE #3895: Logical order of adding & multiplying cost factors in EdgeCost
While thinking about `use_lit` one thing puzzled me: in `EdgeCost` of most models, we have various kinds of factors which we treat in distinct ways, either we **add** a factor (such as highway factor) to the existing factor or we **multiply** it with some other factor. I'm wondering if we do that in the right order.. E.g. in 

https://github.com/valhalla/valhalla/blob/a8024b58f08ed54e644b9b1dc46aeb23f540b3b3/src/sif/autocost.cc#L509-L548

we first add the highway/surface/etc factors to the base factor and after that we multiply that factor with the applicable `Use` factor of the edge. That multiplication also affects the impact of the previous addends such as the highway factor. I think it'd be better to do it the other way around no? First do all multiplications since order doesn't matter there, and only then add the others on top? I'll PR to show what I mean.

It doesn't have the biggest impact I guess, but it does seem more right to me.

nilsnolde:
fixes #3895 

Apart from putting the factor multiplications before the additions, I added a few ferry fixes where I thought it's appropriate.

### DIFF
--- a/src/sif/bicyclecost.cc
+++ b/src/sif/bicyclecost.cc
@@ -608,11 +608,13 @@ Cost BicycleCost::EdgeCost(const baldr::DirectedEdge* edge,
   }
 
   // Ferries are a special case - they use the ferry speed (stored on the edge)
-  if (edge->use() == Use::kFerry) {
+  if (edge->use() == Use::kFerry || edge->use() == Use::kRailFerry) {
     // Compute elapsed time based on speed. Modulate cost with weighting factors.
     assert(edge->speed() < speedfactor_.size());
     float sec = (edge->length() * speedfactor_[edge->speed()]);
-    return {shortest_ ? edge->length() : sec * ferry_factor_, sec};
+    return {shortest_ ? edge->length()
+                      : sec * (edge->use() == Use::kFerry ? ferry_factor_ : rail_ferry_factor_),
+            sec};
   }
 
   // Represents how stressful a roadway is without looking at grade or cycle accommodations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * **Bug Fix**
 * **Enhancement**
    * CHANGED: replace boost::optional with C++17's std::optional where possible [#3890](https://github.com/valhalla/valhalla/pull/3890)
+   * CHANGED: re-ordered the factors in EdgeCost of most costing models so that we first multiply than add factors [#3896](https://github.com/valhalla/valhalla/pull/3896)
 
 ## Release Date: 2022-01-03 Valhalla 3.3.0
 * **Removed**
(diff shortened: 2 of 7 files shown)


src/sif/bicyclecost.cc
```diff
@@ -608,11 +608,13 @@ Cost BicycleCost::EdgeCost(const baldr::DirectedEdge* edge,
   }
 
   // Ferries are a special case - they use the ferry speed (stored on the edge)
-  if (edge->use() == Use::kFerry) {
+  if (edge->use() == Use::kFerry || edge->use() == Use::kRailFerry) {
```
nilsnolde, line 611: Actually bike was mostly fine as far as I could make out, this is the only I'd propose

nilsnolde:
Some minor error while building with casting from uint to float. Though I'd possibly expect some tests to fail as well eventually. I'll wait with fixing that.

nilsnolde:
Talked about it with @kevinkreiser and he had a good point: the order of factor multiplication/addition is debatable and depends on how we want define the impact of costing options. A point could be made here, that the current situation is more of a hierarchical model where the `Use` of a road is weighed much higher than stuff like `use_highways`. My change would make it more "democratic", it's in the end only addends, no multiplicative scaling applied. One could weigh here too (e.g. `0.2 * use_highway_factor + 0.8 * EdgeUse`), but that'd need a lot of testing and fine-tuning I guess. 

So what's the opinion of the others? @dnesbitt61 I heard you're mostly the mastermind behind the general/early costing logic:)

There's also more fundamental improvements that could be made with costing, e.g. the aforementioned decision to not only avoid things, but also prefer things with costing, which is a lot more ineffective than avoiding things and we make the code liable to possibly breaking the admissibility of the A* heuristic by favoring too much in summation over all factors, e.g. @dnesbitt61 noted that on a recent issue with ramping off highways only to join them again immediately after.

I wonder if that's not material for a v4 mostly. Even though all could be done without breaking the major version, I feel like that's stuff that just needs a ton of RAD testing and reasoning about lots of complex interactions (and that twice, for both one separate RAD testing). On the other hand, even with a v4 we'd need to do that sort of testing and since it's not breaking, it's independent of v4 and could be a good chunk or work which we'd like to do anyway?

dnesbitt61:
for bicycle costing most of the dev. work was done by Alex Wilton, an intern at Mapzen. I have no opinion on order of operations but imagine that changes will cause many routes to change (whether for good, bad, or indifferent is TBD) which could be an issue for some. I imagine lots of testing will be required (as you note) to ensure the options work as they should. Could be a case for deferring until V4. Or perhaps retain legacy costing methods and introduce new ones/

nilsnolde:
We talked about this last week and we actually wanted to summarize our findings, so I tried it in #3912

PR #3904, opened 2023-01-12
fix mac builds

kevinkreiser:
since #3894 we haven't been running mac builds. the CI checking script relies on some python dependencies which are no longer installed on mac at the time the script is run. in looking into this problem i realized that it is masked, in that the `needs_ci_run` script when it fails an import (`requests` in this instance) will exit code 1, which is what the CI config checks for to signal no CI is needed haha. so i first refactored the CI check to use printing to signal to do or skip CI. i want to also make sure that the build fails when we screw up the running of the script and then i can go fix the missing dependency


scripts/needs_ci_run
```diff
@@ -35,14 +31,24 @@ def last_build_successful():
...
+        print('CIRCLE_API_TOKEN env not set! Unable to do CI skip check', file=sys.stderr)
+        print('needs CI')
+        return False, None
+
+    # If master, we run 4 jobs, if PR, we run 3 (no docker workflow)
+    # jobs defined in .circleci/config.yml
+    if branch == "master":
+        JOBS = ['build-docker-images', 'lint-build-debug', 'build-release', 'build-osx']
+    else:
+        JOBS = ['lint-build-debug', 'build-release', 'build-osx']
```
nilsnolde, line 51: sorry to barge in here like that, but I realized it's _still_ not skipping CI and seems the reason is that in #3683 we (finally) disabled the docker workflow for PRs but didn't update this file to reflect that, so it was always failing when it didn't get the `build-docker-images` workflow ID for PRs.

Also added some early `sys.exit` when relevant

kevinkreiser: oops just messed that up, didnt realize you added that, apologies!

nilsnolde:
so, after the commit I just pushed builds fine, we'll have to do another empty commit to see if this works, right? otherwise it'll fail with "last build not successful" I think

nilsnolde:
omg.. doesn't skip, says 

```
last build successful?, gitsha: True, c4ff4cab1b15552d51cbc74159b80a7ab5e7381a
needs CI: last commit doesnt exist anymore
CircleCI received exit code 0
```

nilsnolde:
the script is messed up. but mac runs again, so that problem is solved. it's been long-standing that it wouldn't skip irrelevant commits, I vote for merging as-is and take care of the other stuff later.

nilsnolde (approved):

nilsnolde:
ok, one last try from my side.. if it doesn't work, I'll revert to the previous state.

nilsnolde:
yeah that finally worked!! hope it'll finally skip CI when we're having only a changelog commit or so.

nilsnolde (approved):

PR #3906, opened 2023-01-12
fixed underflow of uint64_t cast for matrix time results

nilsnolde:
We had bug report per email from a client that sometimes the matrix (with traffic enabled) gives a unreasonable high `time` result of 4294967295. I looked and it's possible we're casting negative floats to uint64_t.

### DIFF
--- a/src/thor/timedistancematrix.cc
+++ b/src/thor/timedistancematrix.cc
@@ -274,7 +274,7 @@ std::vector<TimeDistance> TimeDistanceMatrix::ComputeMatrix(
         // have been settled or the requested amount of destinations has been found
         tile = graphreader.GetGraphTile(pred.edgeid());
         const DirectedEdge* edge = tile->directededge(pred.edgeid());
-        if (UpdateDestinations(origin, destinations, destedge->second, edge, tile, pred,
+        if (UpdateDestinations(origin, destinations, destedge->second, edge, tile, pred, time_info,
                                matrix_locations)) {
           one_to_many = FormTimeDistanceMatrix(graphreader, origin.date_time(),
                                                time_info.timezone_index, pred.edgeid());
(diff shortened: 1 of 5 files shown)


dnesbitt61 (dismissed):

nilsnolde:
Hmpf, now the traffic matrix test fails.. will investigate another time, it's not so tragic anyways.

src/thor/timedistancematrix.cc
```diff
@@ -498,6 +499,23 @@ bool TimeDistanceMatrix::UpdateDestinations(
...
+      dest.dest_edges_available.erase(dest_available);
+      if (dest.dest_edges_available.empty()) {
+        dest.settled = true;
+        settled_count_++;
+      }
+    };
+
+    if (origin.ll().lat() == dest_loc.ll().lat() && origin.ll().lng() == dest_loc.ll().lng()) {
+      dest.best_cost = Cost{0.f, 0.f};
+      dest.distance = 0;
```
nilsnolde, line 514: The distance also has to be reset, not only the cost

nilsnolde:
I fixed the test, can you have a quick look again @dnesbitt61 ? Thanks!

nilsnolde:
Windows has, what I hope to be, a random error when installing dependencies, `vcpkg` actually raises on 404 mirrors (guess smth down), but that doesn't stop CI from continuing (why..) and eventually fail in the CMake config step, when it can't find protobuf.

I looked into it real quick: it seems that the mirrors (the ones I checked) yanked the current version of `libtool` that our `vcpkg` wants.. It needs `2.4.6.9` but e.g. on https://repo.msys2.org/msys/x86_64/ (main mirror) only `2.4.6.10` and higher is available. I don't know why the hell.. I'll try bumping the vcpkg version again, though that's only a few months ago I last did that..

nilsnolde:
Well, yeah, looks like we should disable the win build for the time being (or ignore it failing) @kevinkreiser . There's an issue about it in the `vcpkg` repo: https://github.com/microsoft/vcpkg/issues/28896. And frankly I wouldn't really want to apply the proposed workaround.

dnesbitt61 (approved):

PR #3907, opened 2023-01-12
fix exclude_polygons bug

### LINKED ISSUE #3905: Empty polygon in avoid_polygons causes Valhalla to crash
Reference to issue outlined in [docker-valhalla](https://github.com/gis-ops/docker-valhalla), as requested: gis-ops/docker-valhalla#82

nilsnolde:
fixes #3905

### DIFF
--- a/src/loki/polygon_search.cc
+++ b/src/loki/polygon_search.cc
@@ -101,7 +101,7 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Ring>& rings_p
                float max_length) {
   // protect for bogus input
   if (rings_pbf.empty() || rings_pbf.Get(0).coords().empty() ||
-      !rings_pbf.Get(0).coords()[0].has_lat()) {
+      !rings_pbf.Get(0).coords()[0].has_lat_case() || !rings_pbf.Get(0).coords()[0].has_lng_case()) {
     return {};
   }
 
--- a/src/loki/polygon_search.cc
+++ b/src/loki/polygon_search.cc
@@ -99,6 +99,11 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Ring>& rings_p
                baldr::GraphReader& reader,
                const std::shared_ptr<sif::DynamicCost>& costing,
                float max_length) {
+  // protect for bogus input
+  if (rings_pbf.empty() || rings_pbf.Get(0).coords().empty() ||
+      !rings_pbf.Get(0).coords()[0].has_lat()) {
+    return {};
+  }
 
   // convert to bg object and check length restriction
   double rings_length = 0;
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -991,7 +991,6 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
       try {
         for (const auto& req_poly : rings_req->GetArray()) {
           if (!req_poly.IsArray() || (req_poly.IsArray() && req_poly.GetArray().Empty())) {
-            add_warning(api, 204);
             continue;
           }
           auto* ring = rings_pbf->Add();
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -986,6 +986,9 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
     auto* rings_pbf = options.mutable_exclude_polygons();
     try {
       for (const auto& req_poly : rings_req->GetArray()) {
+        if (!req_poly.GetArray().Size()) {
+          continue;
+        }
         auto* ring = rings_pbf->Add();
         parse_ring(ring, req_poly);
       }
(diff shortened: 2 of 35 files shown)


nilsnolde:
I should mention it might still be possible to force a segfault with `avoid_polygons`, but that people could hardly accidentally trigger at least. But probably what I'm thinking of causes a runtime_error or so and we'd catch that.

src/worker.cc
```diff
@@ -986,6 +986,9 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
     auto* rings_pbf = options.mutable_exclude_polygons();
     try {
       for (const auto& req_poly : rings_req->GetArray()) {
+        if (req_poly.GetArray().Empty()) {
+          continue;
+        }
```
nilsnolde, line 991: also check if it even is an array as per @kevinkreiser

src/worker.cc
```diff
@@ -983,13 +984,21 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
...
+        for (const auto& req_poly : rings_req->GetArray()) {
+          if (!req_poly.IsArray() || (req_poly.IsArray() && req_poly.GetArray().Empty())) {
+            add_warning(api, 204);
+            continue;
+          }
+          auto* ring = rings_pbf->Add();
+          parse_ring(ring, req_poly);
+        }
+      } catch (...) { throw valhalla_exception_t{137}; }
+    }
```
nilsnolde, line 1001: now we're checking if the outer and inner arrays are indeed arrays and that the inner arrays have a size. if not, we're adding a warning

src/worker.cc
```diff
@@ -983,13 +984,20 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
       rapidjson::get_child_optional(doc, doc.HasMember("avoid_polygons") ? "/avoid_polygons"
                                                                          : "/exclude_polygons");
   if (rings_req) {
-    auto* rings_pbf = options.mutable_exclude_polygons();
-    try {
-      for (const auto& req_poly : rings_req->GetArray()) {
-        auto* ring = rings_pbf->Add();
-        parse_ring(ring, req_poly);
```
kevinkreiser, line 990: what was wrong with this, it should never segfault. we purposely force rapidjson to throw exceptions instead of abort. to my knowledge the only thing in here we do is use rapidjson to parse stuff with like GetArray and GetDouble or whatever, those should all throw when the types are not correct and it should fall into the exception below that says dont send us junk input. if thats not happening either its not rapidjson parsing that is failing or we screwed up the whole making it throw instead of abort stuff in rapidjson utils at some point. if i revert this and run the new tests does it segfault?

nilsnolde: > if i revert this and run the new tests does it segfault?

Jep, it will. GTest is outputting
```
[build] make[3]: *** [test/gurka/CMakeFiles/run-gurka_avoids.dir/build.make:73: test/gurka/gurka_avoids.log] Error 139
[build] make[3]: *** Deleting file 'test/gurka/gurka_avoids.log'
[build] make[2]: *** [CMakeFiles/Makefile2:11115: test/gurka/CMakeFiles/run-gurka_avoids.dir/all] Error 2
[build] make[1]: *** [CMakeFiles/Makefile2:11122: test/gurka/CMakeFiles/run-gurka_avoids.dir/rule] Error 2
[build] make: *** [Makefile:4609: run-gurka_avoids] Error 2
```

It's the polygon_search that segfaults. In the loki worker we send the polygons to `polygon_search` if they're not empty:
https://github.com/valhalla/valhalla/blob/d182120b4312b2534d15927d734e4a4a7a688808/src/loki/worker.cc#L87-L89

Then in polygon_search it segfaults when it wants to correct a polygon: the uninitialized `new_ring` is passed into `correct_ring()` which doesn't have any elements and it segfaults on the `bool is_open = (ring.begin()->lat() != ring.rbegin()->lat() || ring.begin()->lng() != ring.rbegin()->lng());` line.

https://github.com/valhalla/valhalla/blob/d182120b4312b2534d15927d734e4a4a7a688808/src/loki/polygon_search.cc#L36-L57

IMO catching the problem already while parsing the input JSON is best, so it never gets here. But thinking about it again, it's probably a good idea to protect the polygon_search as well, as it's public API and could affect downstream programs in the exact same way. I'll quickly add that and we can discuss.

kevinkreiser: I agree the sooner we see bogus input and throw the better but protecting downstream library functions must also be fixed as you point out

src/loki/polygon_search.cc
```diff
@@ -99,6 +99,11 @@ edges_in_rings(const google::protobuf::RepeatedPtrField<valhalla::Ring>& rings_p
                baldr::GraphReader& reader,
                const std::shared_ptr<sif::DynamicCost>& costing,
                float max_length) {
+  // protect for bogus input
+  if (rings_pbf.empty() || rings_pbf.Get(0).coords().empty() ||
+      !rings_pbf.Get(0).coords()[0].has_lat()) {
+    return {};
+  }
```
nilsnolde, line 106: early abort if there's an obvious invalid input

nilsnolde: also threw in a test for it, which segfaults when these lines are removed

kevinkreiser (approved):

PR #3885, opened 2022-12-31
Build most of Baldr with -Werror

dillona:

PR #3910, opened 2023-01-15
Unify handling of single-file -Werror in all modules

dillona:
This implements unified handling and flags like I was thinking [here](https://github.com/valhalla/valhalla/pull/3885#discussion_r1059765940)


src/CMakeLists.txt
```diff
@@ -70,9 +70,18 @@ endfunction()
...
   cmake_parse_arguments(MODULE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
+  if (UNIX AND ENABLE_SINGLE_FILES_WERROR)
+    set_source_files_properties(
+            ${MODULE_SOURCES}
+            PROPERTIES
+            COMPILE_FLAGS
+            "-Werror -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers")
+  endif()
+  list(APPEND MODULE_SOURCES ${MODULE_SOURCES_WITH_WARNINGS})
```
nilsnolde, line 83: nit: I guess appending to `MODULE_SOURCES` can go into the `if`?

nilsnolde (dismissed):
Thanks for the cleanup @dillona ! Esp for finding and protecting more source files!

src/CMakeLists.txt
dillona: I do not think so, as we need to ensure all files are compiled, regardless of the Werror configuration

nilsnolde: Ah right, sorry, wasn't looking into the code that followed this :sweat_smile: Then all good!

kevinkreiser (approved):

kevinkreiser (approved):

PR #3915, opened 2023-01-17
update vcpkg commit for Azure pipelines to fix libtool mirrors 

### LINKED ISSUE #3914: Fix Azure Win build
Once this is merged, hopefully soon, we can either wait for a new release or point to the tip of master: https://github.com/microsoft/vcpkg/pull/28919

EDIT: And we should update Azure config to also run on `master` pushes.

I'm really not sure what releases mean with `vcpkg` tbh.. @mloskot isn't `vcpkg` always usable in `master`, or it's not really best practice? I could imagine that they only release for the changelog or release-obsessed downstream devs..

nilsnolde:
and build azure on master pushes

fixes #3914

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -29,12 +29,12 @@ pr:
   drafts: false
 
 jobs:
-  - job: VS2019
-    displayName: 'Windows 2019 | VS2019'
+  - job: VS2022
+    displayName: 'Windows 2022 | VS2022'
     timeoutInMinutes: 120
 
     pool:
-      vmImage: 'windows-2019'
+      vmImage: 'windows-2022'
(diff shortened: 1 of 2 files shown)


nilsnolde:
Yay, that worked. Only left to decide if we want to wait for a release or have this commit from last night (it's a few commits later than the PR fixing our problem), ref https://github.com/valhalla/valhalla/issues/3914#issue-1535053232

nilsnolde:
A follow-up PR could be to switch to `windows-2022` (currently we're using 2019), to modernize the build a bit. It's the version used for `windows-latest` since a year.

kevinkreiser (approved):

PR #3943, opened 2023-02-02
fix stitching errors in gtfs ingestion

kevinkreiser:
After merging #3700 i wrote a huge pr description with the work left to do. i will convert that into an issue and we can burn down the tasks but today i bring  you the first task already burned down. There were errors about not being able to find the origin/destination stops of some of the stop pairs (the connections between stations where you actually ride on public transit). this makes other things fail. thankfully the problem wasnt all that complex. i'll describe the fix in the diff, comments should as well

ok so the only sad thing about fixing the stitching error was that the error wasnt stitching at all. this means that in a separate pr we need to offset the graph a bit more so that a trip crosses tiles (forcing stitchign to have to happen). i can add this to the list of things we need to do

### DIFF
--- a/src/mjolnir/ingest_transit.cc
+++ b/src/mjolnir/ingest_transit.cc
@@ -234,7 +234,7 @@ std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_
 void setup_stops(Transit& tile,
                  const gtfs::Stop& tile_stop,
                  GraphId& node_id,
-                 std::unordered_map<feedObject, GraphId>& stop_graphIds,
+                 std::unordered_map<feedObject, GraphId>& platform_node_ids,
                  const std::string& feed_path,
                  NodeType node_type,
                  bool isGenerated,
@@ -257,18 +257,29 @@ void setup_stops(Transit& tile,
   auto onestop_id = isGenerated ? tile_stop.stop_id + "_" + to_string(node_type) : tile_stop.stop_id;
   node->set_onestop_id(onestop_id);
   if (node_type == NodeType::kMultiUseTransitPlatform) {
-    stop_graphIds[{onestop_id, feed_path}] = node_id;
+    // when this platform is not generated, we can use its actual given id verbatim because thats how
+    // other gtfs entities will refer to it. however when it is generated we must use its parents id
+    // and not the generated one because the references in the feed have no idea that we are doing
+    // the generation of new ideas for non-existant platforms
+    platform_node_ids[{tile_stop.stop_id, feed_path}] = node_id;
   }
   node_id++;
 }
 
+/**
+ * Writes all of the nodes of the graph in pbf tile
+ * @param tile       the transit pbf tile to modify
+ * @param tile_info  ephemeral info from the transit feeds which we write into the tile
+ * @return a map of string gtfs id to graph id (so node id) for every platform in the tile
+ *         later on we use these ids to connect platforms that reference each other in the schedule
+ */
 std::unordered_map<feedObject, GraphId> write_stops(Transit& tile, const tileTransitInfo& tile_info) {
   const auto& tile_stopIds = tile_info.tile_stops;
   const auto& tile_children = tile_info.tile_station_children;
   auto node_id = tile_info.graphId;
   feedCache stopFeeds;
 
-  std::unordered_map<feedObject, GraphId> stop_graphIds;
+  std::unordered_map<feedObject, GraphId> platform_node_ids;
   // loop through all stops inside the tile
   for (const feedObject& feed_stop : tile_stopIds) {
     const auto& feed = stopFeeds(feed_stop);
@@ -285,27 +296,27 @@ std::unordered_map<feedObject, GraphId> write_stops(Transit& tile, const tileTra
       gtfs::Stop currChild = *feed.get_stop(it->second);
 
       if (currChild.location_type == gtfs::StopLocationType::EntranceExit) {
-        setup_stops(tile, currChild, node_id, stop_graphIds, feed_stop.feed,
+        setup_stops(tile, currChild, node_id, platform_node_ids, feed_stop.feed,
                     get_node_type(currChild.location_type), false);
       }
     }
     // We require an in/egress so if we didnt add one we need to fake one
     if (tile.nodes_size() == node_count) {
-      setup_stops(tile, *tile_stop, node_id, stop_graphIds, feed_stop.feed, NodeType::kTransitEgress,
-                  true);
+      setup_stops(tile, *tile_stop, node_id, platform_node_ids, feed_stop.feed,
+                  NodeType::kTransitEgress, true);
     }
 
     GraphId prev_id(tile.nodes(node_count).graphid());
     auto node_type = get_node_type((*tile_stop).location_type);
     // Add the Station
     if (node_type == NodeType::kTransitStation) {
-      setup_stops(tile, *tile_stop, node_id, stop_graphIds, feed_stop.feed, node_type, false,
+      setup_stops(tile, *tile_stop, node_id, platform_node_ids, feed_stop.feed, node_type, false,
                   prev_id);
     } else {
       // where there is no station provided, the program will default to creating only stops
       // so, the Station must be added
-      setup_stops(tile, *tile_stop, node_id, stop_graphIds, feed_stop.feed, NodeType::kTransitStation,
-                  true, prev_id);
+      setup_stops(tile, *tile_stop, node_id, platform_node_ids, feed_stop.feed,
+                  NodeType::kTransitStation, true, prev_id);
     }
 
     // Add the Platform
@@ -316,17 +327,17 @@ std::unordered_map<feedObject, GraphId> write_stops(Transit& tile, const tileTra
       gtfs::Stop currChild = *feed.get_stop(it->second);
 
       if (currChild.location_type == gtfs::StopLocationType::StopOrPlatform) {
-        setup_stops(tile, currChild, node_id, stop_graphIds, feed_stop.feed,
+        setup_stops(tile, currChild, node_id, platform_node_ids, feed_stop.feed,
                     get_node_type(currChild.location_type), false, prev_id);
       }
     }
     // We require platforms as well so if we didnt add at least 1 we need to fake one now
     if (tile.nodes_size() == node_count) {
-      setup_stops(tile, *tile_stop, node_id, stop_graphIds, feed_stop.feed,
+      setup_stops(tile, *tile_stop, node_id, platform_node_ids, feed_stop.feed,
                   NodeType::kMultiUseTransitPlatform, true, prev_id);
     }
   }
-  return stop_graphIds;
+  return platform_node_ids;
 }
 
 // read feed data per stop, given shape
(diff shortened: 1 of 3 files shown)


src/mjolnir/ingest_transit.cc
```diff
@@ -234,7 +234,7 @@ std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_
 void setup_stops(Transit& tile,
                  const gtfs::Stop& tile_stop,
                  GraphId& node_id,
-                 std::unordered_map<feedObject, GraphId>& stop_graphIds,
+                 std::unordered_map<feedObject, GraphId>& platform_node_ids,
```
kevinkreiser, line 237: when we write the stops (nodes) into a transit pbf tile we keep track of the gtfs ids assigned to those node which are platforms, thats what this map does. we need that because later we will ingest the part of the feed that references those platforms to connected them with routes/trips at some sort of schedule.

src/mjolnir/ingest_transit.cc
```diff
@@ -257,18 +257,29 @@ void setup_stops(Transit& tile,
   auto onestop_id = isGenerated ? tile_stop.stop_id + "_" + to_string(node_type) : tile_stop.stop_id;
```
kevinkreiser, line 257: to simplify our model of transit we require that every station (even if it doesnt in real life) has:
1. at least one in/egress
2. exactl one station
3. at least one platform

this makes handling connecting the data somewhat easiser and it models the highest level of complexity we want to support (even though bus stops dont need this in almost all circumstances). anyway... on the line above though you'll notice that what we do when the gtfs feed doesnt come with one of these things (either in/egress or platform, as it always has a station) we fake one so that our simplified model is fulfilled. when we fake one we have to give it a unique id so we  take its parent stations id and just add the type onto it (assuming real ones one use the type as a name).

src/mjolnir/ingest_transit.cc
```diff
@@ -257,18 +257,29 @@ void setup_stops(Transit& tile,
...
   node->set_onestop_id(onestop_id);
   if (node_type == NodeType::kMultiUseTransitPlatform) {
-    stop_graphIds[{onestop_id, feed_path}] = node_id;
+    // when this platform is not generated, we can use its actual given id verbatim because thats how
+    // other gtfs entities will refer to it. however when it is generated we must use its parents id
+    // and not the generated one because the references in the feed have no idea that we are doing
+    // the generation of new ideas for non-existant platforms
+    platform_node_ids[{tile_stop.stop_id, feed_path}] = node_id;
```
kevinkreiser, line 264: the problem with giving these generated platforms a new id is that that id is not known to any other entities in the gtfs feed. any schedule information (trips) that connection stations together will connect them by the stations id since from the gtfs feeds point of view there is no platform, valhalla is making that up. so later, when we want to look up with graphid we gave to a platform that we made up by its id, ie when we are iterating over the gtfs feeds trips we need to remember those graphids by their parent station id not their new made up generated id, because of course the feeds raw data doesnt know about that, its a valhalla convention

src/mjolnir/ingest_transit.cc
```diff
@@ -257,18 +257,29 @@ void setup_stops(Transit& tile,
...
   }
   node_id++;
 }
 
+/**
```
kevinkreiser, line 269: from here down there  are pretty much no real important changes, its all me re-naming the generic "stop_graphids" to the specific "platform_node_ids". i think this variable name makes it much more clear what information is going in there even though for the generated ones the gtfs ids will be the parent station (but it is the one to use to get that platforms graphid)

src/mjolnir/ingest_transit.cc
```diff
@@ -425,13 +436,14 @@ bool write_stop_pairs(Transit& tile,
...
                                 : origin_stopId + "_" + to_string(NodeType::kMultiUseTransitPlatform);
         stop_pair->set_origin_onestop_id(origin_onestop_id);
+
         bool dest_is_generated = tile_info.tile_station_children.find({dest_stopId, currFeedPath}) !=
                                  tile_info.tile_station_children.end();
         auto dest_onestop_id =
             dest_is_generated ? dest_stopId
                               : dest_stopId + "_" + to_string(NodeType::kMultiUseTransitPlatform);
         stop_pair->set_destination_onestop_id(dest_onestop_id);
-        stop_pair->set_origin_graphid(kInvalidGraphId);
```
kevinkreiser, line 434: the first time i started debugging this problem i thought i could help myself by initializing these to invalid but what i forgot was that this is protobuf 2 so later on it easily just calls `has_origin_graphid` rather than checking for our sentinel value. same for destination below

test/gurka/test_gtfs.cc
```diff
@@ -121,19 +121,19 @@ TEST(GtfsExample, WriteGtfs) {
   };
   feed.add_stop(nemo);
   struct gtfs::Stop nemo_egress {
-    .stop_id = stopOneID + "_" + to_string(NodeType::kTransitEgress),
-    .stop_name = gtfs::Text("POINT NEMO"), .coordinates_present = true,
-    .stop_lat = station_one_ll->second.second, .stop_lon = station_one_ll->second.first,
-    .parent_station = stopOneID, .location_type = gtfs::StopLocationType::EntranceExit,
-    .stop_timezone = "America/Toronto", .wheelchair_boarding = "1",
+    .stop_id = stopOneID + "_rotating_door_eh", .stop_name = gtfs::Text("POINT NEMO"),
```
kevinkreiser, line 124: decided to use a custom id name rather than what valhalla would generate to increase hardness to input from real gtfs

test/gurka/test_gtfs.cc
```diff
@@ -121,19 +121,19 @@ TEST(GtfsExample, WriteGtfs) {
...
+    .wheelchair_boarding = "1",
   };
   feed.add_stop(nemo_egress);
   struct gtfs::Stop nemo_platform {
-    .stop_id = stopOneID + "_" + to_string(NodeType::kMultiUseTransitPlatform),
-    .stop_name = gtfs::Text("POINT NEMO"), .coordinates_present = true,
-    .stop_lat = station_one_ll->second.second, .stop_lon = station_one_ll->second.first,
-    .parent_station = stopOneID, .location_type = gtfs::StopLocationType::StopOrPlatform,
-    .stop_timezone = "America/Toronto", .wheelchair_boarding = "1",
+    .stop_id = stopOneID + "_ledge_to_the_train_bucko", .stop_name = gtfs::Text("POINT NEMO"),
```
kevinkreiser, line 132: same here, just allowing it to have a non standard id

test/gurka/test_gtfs.cc
```diff
@@ -152,6 +152,15 @@ TEST(GtfsExample, WriteGtfs) {
     .stop_timezone = "America/Toronto", .wheelchair_boarding = "1",
   };
   feed.add_stop(thirdStop);
+  struct gtfs::Stop thirdStopPlatform {
+    .stop_id = stopThreeID + "_walk_the_plank_pal", .stop_name = gtfs::Text("THIRD STOP"),
```
kevinkreiser, line 156: this makes sure to test the case that we can find the graph id for platforms who have actual ids rather than being generated (like the second stations platform above)

test/gurka/test_gtfs.cc
```diff
@@ -181,8 +190,11 @@ TEST(GtfsExample, WriteGtfs) {
   feed.write_trips(path_directory);
 
   // write stop_times.txt
-  std::string stopIds[3] = {stopOneID + "_" + to_string(NodeType::kMultiUseTransitPlatform),
-                            stopTwoID, stopThreeID};
+  std::string stopIds[3] = {
+      stopOneID + "_ledge_to_the_train_bucko",
+      stopTwoID,
+      stopThreeID + "_walk_the_plank_pal",
```
kevinkreiser, line 196: have to update the trips information to reference the proper stops, you'll see that we reference the two non generated platforms by name but the middle one is generated and will use the parent station id to look up its platform when connecting during pbf generation

test/gurka/test_gtfs.cc
```diff
@@ -257,7 +269,7 @@ TEST(GtfsExample, WriteGtfs) {
   EXPECT_EQ(trips[0].trip_id, tripOneID);
 
   const auto& stops = feed_reader.get_stops();
-  EXPECT_EQ(stops.size(), 5);
+  EXPECT_EQ(stops.size(), 6);
```
kevinkreiser, line 272: added a platform to the 3rd station above

test/gurka/test_gtfs.cc
```diff
@@ -328,11 +340,10 @@ TEST(GtfsExample, MakeProto) {
 
       // stop_pair info
       for (int i = 0; i < transit.stop_pairs_size(); i++) {
+        EXPECT_TRUE(transit.stop_pairs(i).has_origin_graphid());
+        EXPECT_TRUE(transit.stop_pairs(i).has_destination_graphid());
```
kevinkreiser, line 344: no need to check for bad values, just check that they are actually set

test/gurka/test_gtfs.cc
```diff
@@ -352,15 +363,17 @@ TEST(GtfsExample, MakeProto) {
...
   EXPECT_EQ(stops.size(), 9);
   std::string stopIds[3] = {stopOneID, stopTwoID, stopThreeID};
-  std::string stop_pair_ids[3] = {stopOneID + "_" + to_string(NodeType::kMultiUseTransitPlatform),
-                                  stopTwoID + "_" + to_string(NodeType::kMultiUseTransitPlatform),
-                                  stopThreeID + "_" + to_string(NodeType::kMultiUseTransitPlatform)};
+  std::string stop_pair_ids[3] = {
+      stopOneID + "_ledge_to_the_train_bucko",
+      stopTwoID + "_" + to_string(NodeType::kMultiUseTransitPlatform),
+      stopThreeID + "_walk_the_plank_pal",
+  };
```
kevinkreiser, line 370: have to look for the correct names now that they are different

test/gurka/test_gtfs.cc
```diff
@@ -352,15 +363,17 @@ TEST(GtfsExample, MakeProto) {
...
 
   for (const auto& stopID : stopIds) {
-    EXPECT_EQ(*stops.find(stopID), stopID);
+    EXPECT_TRUE(stops.find(stopID) != stops.end());
   }
   for (const auto& stop_pair_id : stop_pair_ids) {
-    EXPECT_EQ(*stops.find(stop_pair_id), stop_pair_id);
+    EXPECT_TRUE(stops.find(stop_pair_id) != stops.end());
```
kevinkreiser, line 376: this was fairly non-sensical

test/gurka/test_gtfs.cc
```diff
@@ -385,7 +398,7 @@ TEST(GtfsExample, MakeTile) {
 
   GraphReader reader(pt.get_child("mjolnir"));
   auto graphids = reader.GetTileSet();
-  int totalNodes = 0;
```
kevinkreiser, line 388: bad name made it hard to understand

test/gurka/test_gtfs.cc
```diff
@@ -451,14 +464,15 @@ TEST(GtfsExample, MakeTile) {
...
-  // EXPECT_EQ(uses[Use::kTransitConnection], 6);
+  // NOTE: there are 4 for every station (3 of those) because we connect to both ends of the closest
+  // edge to the station and the connections are bidirectional (as per usual)
+  EXPECT_EQ(uses[Use::kTransitConnection], 12);
   EXPECT_EQ(uses[Use::kPlatformConnection], 6);
   EXPECT_EQ(uses[Use::kEgressConnection], 6);
-  // TODO: it looks like convert transit skips putting any actual rail edges into level 3
-  // EXPECT_EQ(uses[Use::kRail], 4);
+  // TODO: this is the only time in the graph that we dont have opposing directed edges (should fix)
+  EXPECT_EQ(uses[Use::kRail], 2);
```
kevinkreiser, line 475: all of these pass now! that means good things!

nilsnolde (approved):

PR #3947, opened 2023-02-03
upgrade just_gtfs to master

nilsnolde:
Since we're on C++17 now as well, we can switch from the `just_gtfs` branch.  Thanks @mesozoic-drones for keeping it alive/creating it.

There are no releases published, so `master` it is.


kevinkreiser (approved):
interesting github now shows the file level diff of the submodule not just the commit hash diff of the gitmodules file

PR #3949, opened 2023-02-03
parsing logic needed implicit order of stations/egresses/platforms the GTFS feeds

nilsnolde:
Fixed a bug in `ingest_transit` which assumed that stations are always the first entities in a GTFS feed's `stops.txt`, which the spec doesn't mention and very sure isn't the case for real GTFS feeds.

Also refactored a lot to make the logic (and variable names) a bit saner;)

### DIFF
--- a/src/mjolnir/ingest_transit.cc
+++ b/src/mjolnir/ingest_transit.cc
@@ -85,7 +85,7 @@ struct tile_transit_info_t {
   std::unordered_map<feed_object_t, size_t> shapes;
 
   bool operator<(const tile_transit_info_t& t1) const {
-    // sort tile_transit_info_t by size
+    // the queue needs to be able to sort
     return stations.size() < t1.stations.size();
   }
 };
@@ -140,16 +140,10 @@ std::priority_queue<tile_transit_info_t> select_transit_tiles(const boost::prope
   // adds a tile_info for the tile_id to tile_map if there is none yet, and returns the tile_info
   // object
   auto get_tile_info = [&tile_map, &local_tiles](const gtfs::Stop& stop) -> tile_transit_info_t& {
-    tile_transit_info_t tile_info;
     auto graphid = GraphId(local_tiles.TileId(stop.stop_lat, stop.stop_lon),
                            TileHierarchy::GetTransitLevel().level, 0);
 
-    auto tile_inserted = tile_map.insert({graphid, tile_info});
-    if (tile_inserted.second) {
-      tile_inserted.first->second.graphid = graphid;
-    }
-
-    return tile_inserted.first->second;
+    return tile_map.insert({graphid, tile_transit_info_t{graphid}}).first->second;
   };
 
   filesystem::recursive_directory_iterator gtfs_feed_itr(path);
@@ -216,11 +210,11 @@ std::priority_queue<tile_transit_info_t> select_transit_tiles(const boost::prope
                feed_path);
     }
   }
-  std::priority_queue<tile_transit_info_t> prioritized;
+  std::priority_queue<tile_transit_info_t> queue;
   for (auto it = tile_map.begin(); it != tile_map.end(); it++) {
-    prioritized.push(it->second);
+    queue.push(it->second);
   }
-  return prioritized;
+  return queue;
 }
 
 void setup_stops(Transit& tile,
--- a/src/mjolnir/ingest_transit.cc
+++ b/src/mjolnir/ingest_transit.cc
@@ -49,20 +49,6 @@ using namespace valhalla::mjolnir;
 
 namespace {
 
-valhalla::baldr::NodeType get_node_type(gtfs::StopLocationType stop_type) {
-
-  static const std::unordered_map<gtfs::StopLocationType, valhalla::baldr::NodeType>
-      stop_types{{gtfs::StopLocationType::StopOrPlatform,
-                  valhalla::baldr::NodeType::kMultiUseTransitPlatform},
-                 {gtfs::StopLocationType::Station, valhalla::baldr::NodeType::kTransitStation},
-                 {gtfs::StopLocationType::EntranceExit, valhalla::baldr::NodeType::kTransitEgress}};
-  auto i = stop_types.find(stop_type);
-  if (i == stop_types.cend()) {
-    throw std::runtime_error("Unusable GTFS Stop Type");
-  }
-  return i->second;
-}
-
 struct feedObject {
   gtfs::Id id;
   std::string feed;
@@ -87,20 +73,20 @@ template <> struct equal_to<feedObject> {
 namespace {
 
 struct tileTransitInfo {
-  GraphId graphId;
+  GraphId graphid;
   // TODO: unordered multimap-string to a pair of strings that maps from station (parent_id) ->
   //  platform/egress (child) (Max 2 kinds / many could exists)
-  std::unordered_multimap<feedObject, gtfs::Id> tile_station_children;
-  std::unordered_set<feedObject> tile_stops;
-  std::unordered_set<feedObject> tile_trips;
-  std::unordered_map<feedObject, size_t> tile_routes;
-  std::unordered_set<feedObject> tile_services;
-  std::unordered_set<feedObject> tile_agencies;
-  std::unordered_map<feedObject, size_t> tile_shapes;
+  std::unordered_multimap<feedObject, gtfs::Id> station_children;
+  std::unordered_set<feedObject> stations;
+  std::unordered_set<feedObject> trips;
+  std::unordered_map<feedObject, size_t> routes;
+  std::unordered_set<feedObject> services;
+  std::unordered_set<feedObject> agencies;
+  std::unordered_map<feedObject, size_t> shapes;
 
   bool operator<(const tileTransitInfo& t1) const {
     // sort tileTransitInfo by size
-    return tile_stops.size() < t1.tile_stops.size();
+    return stations.size() < t1.stations.size();
   }
 };
 
@@ -148,8 +134,23 @@ std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_
   auto path = pt.get<std::string>("mjolnir.transit_feeds_dir");
 
   std::set<GraphId> tiles;
-  const auto& tile_level = TileHierarchy::levels().back();
+  const auto& local_tiles = TileHierarchy::levels().back().tiles;
   std::unordered_map<GraphId, tileTransitInfo> tile_map;
+
+  // adds a tile_info for the tile_id to tile_map if there is none yet, and returns the tile_info
+  // object
+  auto get_tile_info = [&tile_map, &local_tiles](const gtfs::Stop& stop) -> tileTransitInfo& {
+    tileTransitInfo tile_info;
+    auto graphid = GraphId(local_tiles.TileId(stop.stop_lat, stop.stop_lon),
+                           TileHierarchy::GetTransitLevel().level, 0);
+
+    auto tile_inserted = tile_map.insert({graphid, tile_info});
+    if (tile_inserted.second) {
+      tile_inserted.first->second.graphid = graphid;
+    }
+
+    return tile_inserted.first->second;
+  };
 
   filesystem::recursive_directory_iterator gtfs_feed_itr(path);
   filesystem::recursive_directory_iterator end_file_itr;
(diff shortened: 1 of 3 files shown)


src/mjolnir/ingest_transit.cc
```diff
@@ -87,20 +73,20 @@ template <> struct equal_to<feedObject> {
...
+  std::unordered_set<feedObject> trips;
+  std::unordered_map<feedObject, size_t> routes;
+  std::unordered_set<feedObject> services;
+  std::unordered_set<feedObject> agencies;
+  std::unordered_map<feedObject, size_t> shapes;
 
   bool operator<(const tileTransitInfo& t1) const {
     // sort tileTransitInfo by size
-    return tile_stops.size() < t1.tile_stops.size();
+    return stations.size() < t1.stations.size();
```
nilsnolde, line 89: renamed all this camelcase stuff (and redundant parts of names)

src/mjolnir/ingest_transit.cc
```diff
@@ -162,66 +163,57 @@ std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_
...
         }
-        gtfs::Id currStopId = stop.stop_id;
-        float x = stop.stop_lon;
-        float y = stop.stop_lat;
-
-        GraphId currId =
-            GraphId(tile_level.tiles.TileId(y, x), TileHierarchy::GetTransitLevel().level, 0);
-
-        tileTransitInfo currTile;
+      }
```
nilsnolde, line 172: this is the real change: we do a first pass over the GTFS stops to just add all stations so we can determine the proper children for them in the second pass below.

tbh, it could still be simplified quite a bit IMO, but I'll look at it again some other time.

src/mjolnir/ingest_transit.cc
```diff
@@ -162,66 +163,57 @@ std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_
...
-          currTile = tile_map[currId];
-        } else {
-          currTile.graphId = currId;
+      // 2nd pass to add the platforms/stops
+      for (const auto& stop : stops) {
+        // TODO: GenericNode & BoardingArea could be useful at some point
+        if (!(stop.location_type == gtfs::StopLocationType::StopOrPlatform) &&
+            !(stop.location_type == gtfs::StopLocationType::EntranceExit)) {
+          continue;
         }
```
nilsnolde, line 180: for now, we're not interested in anything else but platforms/egresses

src/mjolnir/ingest_transit.cc
```diff
@@ -162,66 +163,57 @@ std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_
...
+        if (station_in_tile != tile_info.stations.end()) {
+          tile_info.station_children.insert({{stop.parent_station, feed_path}, stop.stop_id});
         } else {
-          currTile.tile_stops.insert({stop.stop_id, feed_path});
+          // we don't have the parent station, if
+          // 1) this stop has none or 2) its parent station is in another tile
+          // TODO: need to handle the 2nd case somehow!
+          tile_info.stations.insert({stop.stop_id, feed_path});
+          tile_info.station_children.insert({{stop.stop_id, feed_path}, stop.stop_id});
         }
```
nilsnolde, line 195: if the station is in this tile, we can add this one to the children of it. if it doesn't have a parent station at all, we add it to both, the `stations` & its `children`. further down we will distinguish, it won't actually be added as a station.

however, we need to somehow handle the case where the parent station is not even in the same tile as a platform, which surely happens somewhere in the world.

src/mjolnir/ingest_transit.cc
```diff
@@ -87,20 +73,20 @@ template <> struct equal_to<feedObject> {
...
   //  platform/egress (child) (Max 2 kinds / many could exists)
-  std::unordered_multimap<feedObject, gtfs::Id> tile_station_children;
-  std::unordered_set<feedObject> tile_stops;
-  std::unordered_set<feedObject> tile_trips;
-  std::unordered_map<feedObject, size_t> tile_routes;
-  std::unordered_set<feedObject> tile_services;
-  std::unordered_set<feedObject> tile_agencies;
-  std::unordered_map<feedObject, size_t> tile_shapes;
+  std::unordered_multimap<feedObject, gtfs::Id> station_children;
+  std::unordered_set<feedObject> stations;
```
nilsnolde, line 80: this is really supposed to be stations, not generic stops, which go into the `station_children`.

test/gurka/test_gtfs.cc
```diff
@@ -136,6 +129,13 @@ TEST(GtfsExample, WriteGtfs) {
     .wheelchair_boarding = "1",
   };
   feed.add_stop(nemo_platform);
+  struct gtfs::Stop nemo {
+    .stop_id = stopOneID, .stop_name = gtfs::Text("POINT NEMO"), .coordinates_present = true,
+    .stop_lat = station_one_ll->second.second, .stop_lon = station_one_ll->second.first,
+    .parent_station = "", .location_type = gtfs::StopLocationType::Station,
+    .stop_timezone = "America/Toronto", .wheelchair_boarding = "1",
+  };
+  feed.add_stop(nemo);
```
nilsnolde, line 138: now we're adding the stations for nemo last, after the platform & egress. before this PR, this failed.

src/mjolnir/ingest_transit.cc
```diff
@@ -72,42 +58,42 @@ struct feedObject {
...
-    return tile_stops.size() < t1.tile_stops.size();
+  std::unordered_multimap<feed_object_t, gtfs::Id> station_children;
+  std::unordered_set<feed_object_t> stations;
+  std::unordered_set<feed_object_t> trips;
+  std::unordered_map<feed_object_t, size_t> routes;
+  std::unordered_set<feed_object_t> services;
+  std::unordered_set<feed_object_t> agencies;
+  std::unordered_map<feed_object_t, size_t> shapes;
+
+  bool operator<(const tile_transit_info_t& t1) const {
```
kevinkreiser, line 87: cool i guess we can delete this function since we dont use it in any ordered containers?

src/mjolnir/ingest_transit.cc
```diff
@@ -162,69 +163,60 @@ std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_
...
-        tile_map[currId] = currTile;
       }
 
-      LOG_INFO("Done Sorting through " + feed_path);
+      LOG_INFO("Done parsing " + std::to_string(tile_map.size()) + " transit tiles for GTFS feed " +
+               feed_path);
     }
   }
-  std::priority_queue<tileTransitInfo> prioritized;
+  std::priority_queue<tile_transit_info_t> prioritized;
```
kevinkreiser, line 219: while we are fixing naming:
```suggestion
  std::priority_queue<tile_transit_info_t> queue;
```

src/mjolnir/ingest_transit.cc
```diff
@@ -143,13 +129,28 @@ struct unique_transit_t {
...
+    tile_transit_info_t tile_info;
+    auto graphid = GraphId(local_tiles.TileId(stop.stop_lat, stop.stop_lon),
+                           TileHierarchy::GetTransitLevel().level, 0);
+
+    auto tile_inserted = tile_map.insert({graphid, tile_info});
+    if (tile_inserted.second) {
+      tile_inserted.first->second.graphid = graphid;
+    }
+
+    return tile_inserted.first->second;
```
kevinkreiser, line 152: nit: slightly verbose, the `::insert` method is made to do this work for you already
```suggestion
    return tile_map.insert({graphid, tile_transit_info_t{graphid}}).first->second;
```

src/mjolnir/ingest_transit.cc
kevinkreiser: by doing it this way you have, im pretty sure, created a bug. if a trip begins at at station (ie. `gtfs::StopLocationType::Station`) you will not record it below because you have pulled the loops apart. one options would be, when we load the feed we do so with a derived class and do a quick sort on the stops by their type. something like this:

```c++
class sortable_feed : public just_gtfs::feed {
public:
 void sort_stops() {
   // subtract to make station the smallest in the enum
   std::sort(stops.begin(), stops.end(), [](const auto& a, const auto& b){ return static_cast<uint8_t>(a.location_type - 1) < static_cast<uint8_t>(b.location_type - 1); }
 }
};
...
// above near the feed reading part just call sort once
feed.sort_stops();
...
// then everything can stay in the same for loop and we dont have to check the types and all that
```

or maybe you can think of a cleaner way to loop over the stations first? i guess you could make an outer loop with the types you wnat to look for in each iteration, that might be clean as well? up to you, but i think there is a new bug here if we dont handle trips from stations :smile:

src/mjolnir/ingest_transit.cc
kevinkreiser: yeah in the builder or some other part of the code we mention this and i did cover it with @chris-jpark. it just didnt seem like something we should worry about on the first go but definitely will hit it in a real feed. if we can detect its happening (which we can) we should log about it. also we should make a test case for it eventually too. but yeah first things first lets get it working haha

src/mjolnir/ingest_transit.cc
```diff
@@ -162,69 +163,60 @@ std::priority_queue<tileTransitInfo> select_transit_tiles(const boost::property_
...
-
-          currTile.tile_routes.insert(
-              {{currRoute->route_id, feed_path}, currTile.tile_routes.size()});
-          currTile.tile_agencies.insert({currRoute->agency_id, feed_path});
-          // add to the queue of tiles
+          tile_info.services.insert({trip->service_id, feed_path});
+          tile_info.trips.insert({trip->trip_id, feed_path});
+          tile_info.shapes.insert({{trip->shape_id, feed_path}, tile_info.shapes.size()});
+          tile_info.routes.insert({{route->route_id, feed_path}, tile_info.routes.size()});
+          tile_info.agencies.insert({route->agency_id, feed_path});
```
kevinkreiser, line 211: tidy, thank you

src/mjolnir/ingest_transit.cc
nilsnolde: I'm not really sure it's a bug.. Sounds more like GTFS is designed in a way that a station can't be referenced by trips by a `stop_id`: https://developers.google.com/transit/gtfs/reference#stop_timestxt.

> `stop_id`: [...] Referenced locations must be stops, not stations or station entrances. [...]

Though I'm not saying it's not going to be the case for some shoddy feed, but the spec sounds pretty clear or am I missing smth? Do you still want to protect for that?

src/mjolnir/ingest_transit.cc
nilsnolde: the priority queue still needs it

kevinkreiser: strange that code coverage said its not used or maybe I misunderstood the coverage annotations

src/mjolnir/ingest_transit.cc
kevinkreiser: it's possible I am misremebering. our previous interface to gtfs was transitland and I thought that the station always existed and the platform and egress were faked when not there. seems like the spec says platform is always there but station and egress may not be?

nilsnolde: I think it's more trying to force a feed to be reasonable? I was actually surprised that a station should directly be referenced in a trip, it didn't seem intuitive to me (same for egresses no?). From what I understood, a station might not be there (e.g. a simple bus stop), but every station that is there, needs to have at least one platform. Egresses are probably optional as well?

I don't know why transit.land would fake a platform, you're right, and we're still doing that. Think I'll add a log to that so we can see it in real GTFS feeds.

kevinkreiser: seems to me we s hould look at some feeds with bus schedules in them to see how they are modeled for real.

but at least gtfs spec says regarding `parent_id`
> • Optional for stops/platforms (location_type=0).

which i guess means our understanding is correct regardless of what transitlands model had. but does definitely mean we need to handle `parent_id` being empty string from just_gtfs when its a stop/platform

src/mjolnir/ingest_transit.cc
```diff
@@ -143,98 +129,104 @@ struct unique_transit_t {
...
-            GraphId(tile_level.tiles.TileId(y, x), TileHierarchy::GetTransitLevel().level, 0);
+        auto& tile_info = get_tile_info(stop);
 
-        tileTransitInfo currTile;
-
-        if (tile_map.find(currId) != tile_map.end()) {
-          currTile = tile_map[currId];
+        // if this station doesn't exist, we need to create it: we use the fact that this entry is
+        // not a station type to fake a station object in write_stops()
+        auto station_in_tile = tile_info.stations.find({stop.parent_station, feed_path});
```
kevinkreiser, line 179: the `parent_station` may be empty for stop/platform types, aparently. what this means is that this code will presumeably still work (because empty string wont occur as a real id for another station) but it would be clearer if we said something about it or updated the if to check for empty or something

src/mjolnir/ingest_transit.cc
```diff
@@ -143,98 +129,104 @@ struct unique_transit_t {
...
-            stop.location_type == gtfs::StopLocationType::EntranceExit) {
-          auto parent_in_tile = currTile.tile_stops.find({stop.parent_station, feed_path});
-          if (parent_in_tile != currTile.tile_stops.end()) {
-            currTile.tile_station_children.insert({{stop.parent_station, feed_path}, stop.stop_id});
-          } else {
-            currTile.tile_stops.insert({stop.stop_id, feed_path});
+          // we don't have the parent station, if
+          // 1) this stop has none or 2) its parent station is in another tile
+          // TODO: need to handle the 2nd case somehow!
+          auto parent_station = feed.get_stop(stop.parent_station);
```
kevinkreiser, line 186: if the `parent_station` is empty we should not emit this warning and should skip this check entirely

src/mjolnir/ingest_transit.cc
nilsnolde: > but does definitely mean we need to handle parent_id being empty string from just_gtfs when its a stop/platform

We already do handle that, see line 180 onwards in this PR.

src/mjolnir/ingest_transit.cc
nilsnolde: ah yeah, exactly here do distinguish already, albeit not in the most obvious manner..

src/mjolnir/ingest_transit.cc
nilsnolde: hm, then we'd need to do more changes as downstream logic currently relies on stops getting inserted into this `stations` map. I also just made this a bit more readable, didn't change too much of the overall logic.

kevinkreiser (dismissed):
just the nit about the log

kevinkreiser (approved):

PR #3954, opened 2023-02-07
write multiple PBFs if the protobuf object gets too big

nilsnolde:
We had a small bug where large PBF objects wouldn't be able to get serialized.

TODO: fix `convert_transit.cc` to handle `.pbf.x` file extensions.

### DIFF
--- a/src/mjolnir/ingest_transit.cc
+++ b/src/mjolnir/ingest_transit.cc
@@ -128,7 +128,13 @@ struct unique_transit_t {
   std::unordered_map<std::string, size_t> lines;
 };
 
-// Read from GTFS feed, sort data into the tiles they belong to
+std::string get_tile_path(const std::string tile_dir, const GraphId& tile_id) {
+  auto file_name = GraphTile::FileSuffix(tile_id);
+  file_name = file_name.substr(0, file_name.size() - 3) + "pbf";
+  return tile_dir + filesystem::path::preferred_separator + file_name;
+};
+
+// Read from GTFS feed, sort data into the unique tiles they belong to
 std::priority_queue<tile_transit_info_t> select_transit_tiles(const boost::property_tree::ptree& pt) {
 
   auto path = pt.get<std::string>("mjolnir.transit_feeds_dir");
@@ -616,28 +622,37 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
     queue.pop();
     uniques.lock.unlock();
 
+    // PBF objects throw when they're too big, so keep writing .pbf.x tiles until there's no more
     Transit tile;
-    auto file_name = GraphTile::FileSuffix(current.graphid, SUFFIX_NON_COMPRESSED);
-    file_name = file_name.substr(0, file_name.size() - 3) + "pbf";
-    filesystem::path transit_tile = pt.get<std::string>("mjolnir.transit_dir") +
-                                    filesystem::path::preferred_separator + file_name;
+    bool dangles = false;
+    uint16_t ext = 0;
 
-    // tiles are wrote out with .pbf or .pbf.n ext
-    const std::string& prefix = transit_tile.string();
-    LOG_INFO("Writing " + prefix);
+    const auto tile_path = get_tile_path(pt.get<std::string>("mjolnir.transit_dir"), current.graphid);
+    // first tile also gets an extension
+    auto current_path = tile_path;
+    do {
+      std::unordered_map<feed_object_t, GraphId> platform_node_ids = write_stops(tile, current);
+      dangles = write_stop_pairs(tile, current, platform_node_ids, uniques) || dangles;
+      write_routes(tile, current);
+      write_shapes(tile, current);
 
-    std::unordered_map<feed_object_t, GraphId> platform_node_ids = write_stops(tile, current);
-    bool dangles = write_stop_pairs(tile, current, platform_node_ids, uniques);
-    write_routes(tile, current);
-    write_shapes(tile, current);
+      if (tile.stop_pairs_size() > 500000) {
+        LOG_INFO("Writing " + current_path);
+        write_pbf(tile, current_path);
+        tile.Clear();
+        current_path = tile_path + "." + std::to_string(ext++);
+      }
 
-    // list of dangling tiles
+    } while (filesystem::exists(current_path));
+
     if (dangles) {
       dangling.emplace_back(current.graphid);
     }
 
+    // write the last tile
     if (tile.stop_pairs_size()) {
-      write_pbf(tile, transit_tile.string());
+      LOG_INFO("Writing " + current_path);
+      write_pbf(tile, current_path);
     }
   }
   promise.set_value(dangling);
@@ -649,12 +664,7 @@ void stitch_tiles(const boost::property_tree::ptree& pt,
                   std::list<GraphId>& tiles,
                   std::mutex& lock) {
   auto grid = TileHierarchy::levels().back().tiles;
-  auto tile_name = [&pt](const GraphId& id) {
-    auto file_name = GraphTile::FileSuffix(id);
-    file_name = file_name.substr(0, file_name.size() - 3) + "pbf";
-    return pt.get<std::string>("mjolnir.transit_dir") + filesystem::path::preferred_separator +
-           file_name;
-  };
+  auto transit_dir = pt.get<std::string>("mjolnir.transit_dir");
 
   // for each tile
   while (true) {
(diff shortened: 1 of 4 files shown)


src/mjolnir/ingest_transit.cc
```diff
@@ -128,7 +128,13 @@ struct unique_transit_t {
   std::unordered_map<std::string, size_t> lines;
 };
 
-// Read from GTFS feed, sort data into the tiles they belong to
+std::string get_tile_path(const std::string tile_dir, const GraphId& tile_id) {
+  auto file_name = GraphTile::FileSuffix(tile_id);
+  file_name = file_name.substr(0, file_name.size() - 3) + "pbf";
+  return tile_dir + filesystem::path::preferred_separator + file_name;
+};
```
nilsnolde, line 135: pulled this out into a separate function, it's needed in multiple other functions

src/mjolnir/ingest_transit.cc
```diff
@@ -616,28 +622,37 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
...
+        write_pbf(tile, current_path);
+        tile.Clear();
+        current_path = tile_path + "." + std::to_string(ext++);
+      }
 
-    std::unordered_map<feed_object_t, GraphId> platform_node_ids = write_stops(tile, current);
-    bool dangles = write_stop_pairs(tile, current, platform_node_ids, uniques);
-    write_routes(tile, current);
-    write_shapes(tile, current);
+    } while (filesystem::exists(current_path));
```
nilsnolde, line 646: hmpf, this was too quick: it'll be a forever loop once a tile has more than 500k stop pairs, since I didn't write any code that removes any objects from the current `transit_info`.. will look at it tmrw again..

src/mjolnir/ingest_transit.cc
```diff
@@ -524,6 +534,17 @@ bool write_stop_pairs(Transit& tile,
...
+    tile_itr = tile_tripIds.erase(tile_itr);
+
+#if GOOGLE_PROTOBUF_VERSION >= 3001000
+    auto size = tile.ByteSizeLong();
+#else
+    auto size = tile.ByteSize();
+#endif
+    if (size > 134217728) {
+      return dangles;
+    }
```
nilsnolde, line 547: alright, took a stab at limiting PBF size (for ref PBF can't have a byte size of > 2 GB, apparently because their internal size check is based on `int`, not `size_t`, wtf!!).

we'd need to do smth like this IMO. here in `write_stop_pairs` we add most of the PBF objects and is likely the place where it goes wrong. also the old code was only checking the stop pairs for PBF size. but the old code was basically paging through transit.land's API response of `get_stop_pairs` endpoint and never really had the full picture of the entire tile available. however, we do have all GTFS objects this entire tile contains all at once. 

we can remove the the trip from the `transit_info` here and check in the calling function if there's still trips left to process.

src/mjolnir/ingest_transit.cc
```diff
@@ -616,28 +637,36 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
...
     write_shapes(tile, current);
+    while (current.trips.size()) {
+
+      // writes every GB a new file with extension
+      dangles = write_stop_pairs(tile, current, platform_node_ids, uniques) || dangles;
+
+      LOG_INFO("Writing " + current_path);
+      write_pbf(tile, current_path);
+      tile.Clear();
+      current_path = tile_path + "." + std::to_string(ext++);
```
nilsnolde, line 659: we can write the stops, routes & shapes, but have to be careful when writing the stop pairs, to not exceed the stupid 2 GB limit of PBF.

nilsnolde:
I quickly tried one way of doing it, but I really really don't like it @kevinkreiser . If you have a better idea later, most welcome, otherwise I'll pore over it tmrw to make it nicer. 

I guess we could iterate over the trips in the `ingest_tiles` function and have the `write_stop_pairs` function only write a particular stop pair, then at least the whole PBF size check would be more visible. Looks kinda odd right now..

src/mjolnir/ingest_transit.cc
```diff
@@ -364,167 +378,153 @@ float add_stop_pair_shapes(const gtfs::Stop& stop_connect,
...
 
 // return dangling stop_pairs, write stop data from feed
-bool write_stop_pairs(Transit& tile,
-                      const tile_transit_info_t& tile_info,
-                      std::unordered_map<feed_object_t, GraphId> platform_node_ids,
-                      unique_transit_t& uniques) {
-
-  const auto& tile_tripIds = tile_info.trips;
-  feed_cache_t tripFeeds;
+bool write_stop_pair(Transit& tile,
```
nilsnolde, line 381: pull the loop over trips in the calling function `ingest_tiles`

src/mjolnir/ingest_transit.cc
```diff
@@ -617,27 +617,42 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
...
 
-    // list of dangling tiles
+    // we have to be careful with writing stop_pairs to not exceed PBF's stupid 2 GB limit
+    feed_cache_t feeds;
+    size_t trip_count = 0;
+    for (const auto& trip : current.trips) {
+      dangles =
+          write_stop_pair(tile, current, trip, feeds(trip), platform_node_ids, uniques) || dangles;
+
+      if (trip_count > 50000) {
```
nilsnolde, line 637: TODO: better heuristic

src/mjolnir/ingest_transit.cc
```diff
@@ -617,27 +617,42 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
...
     write_routes(tile, current);
     write_shapes(tile, current);
 
-    // list of dangling tiles
+    // we have to be careful with writing stop_pairs to not exceed PBF's stupid 2 GB limit
+    feed_cache_t feeds;
+    size_t trip_count = 0;
+    for (const auto& trip : current.trips) {
+      dangles =
+          write_stop_pair(tile, current, trip, feeds(trip), platform_node_ids, uniques) || dangles;
```
nilsnolde, line 635: easier to loop over trips right here..

scripts/valhalla_build_config
```diff
@@ -33,6 +33,7 @@ config = {
         'transit_dir': '/data/valhalla/transit',
         'transit_feeds_dir': '/data/valhalla/transit_feeds',
         'transit_bounding_box': Optional(str),
+        'transit_pbf_limit': 20000,
```
nilsnolde, line 36: those are representing gtfs trips. let's see what real feeds do. thanks @kevinkreiser for suggesting that.

src/mjolnir/ingest_transit.cc
```diff
@@ -839,8 +840,8 @@ Transit read_pbf(const std::string& file_name, std::mutex& lock) {
   lock.lock();
   std::fstream file(file_name, std::ios::in | std::ios::binary);
   if (!file) {
-    throw std::runtime_error("Couldn't load " + file_name);
     lock.unlock();
+    throw std::runtime_error("Couldn't load " + file_name);
```
nilsnolde, line 844: first unlock, then throw

test/gurka/test_gtfs.cc
```diff
@@ -323,6 +324,13 @@ TEST(GtfsExample, MakeProto) {
       std::string fname = transit_file_itr->path().string();
       mjolnir::Transit transit = mjolnir::read_pbf(fname);
 
+      if (std::isdigit(fname.back())) {
+        // we produce 2 pbf tiles on purpose, where the last one (xx.pbf.0) only has a bunch of stop
+        // pairs
+        EXPECT_NE(transit.stop_pairs_size(), 0);
+        continue;
+      }
```
nilsnolde, line 332: with `mjolnir.transit_pbf_limit: 1` we get two PBFs (one with `xx.pbf.0`, but the last one only has a few stop pairs, most of the objects are in the first one

valhalla/mjolnir/ingest_transit.h
```diff
@@ -47,16 +47,6 @@ void stitch_transit(const boost::property_tree::ptree& pt, std::list<baldr::Grap
...
 
-/**
- * @brief Get PBF transit data given a GraphId / tile
- *
- * @param id graphId of the tile
- * @param transit_dir directory where transit data is located
- * @param file_name  name of the file
- * @return transit data of the tile with the given GraphId
- */
-Transit read_pbf(const baldr::GraphId& id, const std::string& transit_dir, std::string& file_name);
```
nilsnolde, line 58: not needed overload

nilsnolde:
This should be done for now @kevinkreiser .

- added a `mjolnir.transit_pbf_limit` config option to limit the size of trips per feed
- made `convert_transit` compatible with extended transit PBF file paths
- fixed test

kevinkreiser (dismissed):

kevinkreiser (approved):

PR #3957, opened 2023-02-09
added use_lit costing option for pedestrian

chrstnbwnkl:
# Issue

Adds a `use_lit` costing option to pedestrian costing, see discussion in #3848.

The `use_lit` value is converted into an `avoid_unlit` factor using the following formula:

```cpp
 unlit_factor_ = use_lit < 0.5f ? 1 + (kMinLitFactor + 2.f * use_lit) : ((kMinLitFactor - 5) + 12.f * use_lit);
```

Which maps the range `[0, 1]` to the factor range `[1, 8]`, so that `use_lit=0` has no effect on the route, `0 < use_lit < 0.5` has some effect on the route and `0.5 ≤ use_lit ≤ 1` has a comparatively stronger effect on the route. 

grafik

`use_lit` is defined in `dynamiccost` so it could be used as well by e.g. bicycle costing, the exact implementation for which would need to be discussed (as an avid cyclist myself I could see a necessity for this, as the degree to which a cyclist can be seen by other road users greatly determines safety at night).

## Tasklist

 - [x] Add tests
 - [x ] Add #fixes with the issue number that this PR addresses

## Requirements / Relations


nilsnolde (approved):
Thanks, looks good to me! Just a tiny nit.

Next PR(s) can deal with
- adding additional penalties if the edge is a tunnel _and_ unlit
- adding `use_lit` to bicycle costing, what do you think @dnesbitt61 ?

I personally don't think RAD testing is very necessary, though it'll be called on most edges which are ped accessible. Also its impact is fairly well imaginable IMO.

proto/options.proto
```diff
@@ -303,12 +303,14 @@ message Costing {
...
       uint32 restriction_probability = 77;
     }
+
     repeated AvoidEdge exclude_edges = 78;
     oneof has_elevator_penalty {
       float elevator_penalty = 79;
     }
     uint32 fixed_speed = 80;
     uint32 axle_count = 81;
+    float use_lit = 82;
```
nilsnolde, line 313: looks like we have to do it a `oneof` here to satisfy the `JSON_PBF_RANGED_DEFAULT` macro @chrstnbwnkl , even though it wouldn't actually be necessary.

proto/options.proto
```diff
@@ -303,12 +303,17 @@ message Costing {
     oneof has_restriction_probability {
       uint32 restriction_probability = 77;
     }
+    oneof has_use_lit {
+      float use_lit = 82;
+    }
```
nilsnolde, line 308: Let's keep this at the end of it, to keep it a linear sequence of pbf indices

proto/options.proto
```diff
@@ -309,6 +309,9 @@ message Costing {
     }
     uint32 fixed_speed = 80;
     uint32 axle_count = 81;
+    oneof has_use_lit {
+      float use_lit = 82;
+    }
```
kevinkreiser, line 314: if the default is 0.f then you dont need to use `oneof` because both not being set and the default are one and the same value. i know that all the other ones, even the ones that default to 0 are using `oneof` but thats only because i didnt get a chance to go back and do these, you could be the guy to start the wave :D

src/sif/pedestriancost.cc
```diff
@@ -719,6 +722,9 @@ Cost PedestrianCost::EdgeCost(const baldr::DirectedEdge* edge,
   } else if (edge->roundabout()) {
     factor *= kRoundaboutFactor;
   }
+  if (!edge->lit()) {
+    factor *= unlit_factor_;
+  }
```
kevinkreiser, line 727: this may be a preoptimization however, it is possible to do this without a branch (`if`), you can cast the bool to float and multiply with it instead which may help performance somewhat, it certainly wont be slower than the branch statement.
```suggestion
  factor *= (!edge->lit()) * unlit_factor_;
```

kevinkreiser: this is bullcrap sorry, didnt think about hard enough, when it is lit we need the factor to still be 1, so we need some slightly more complicated math to avoid the `if` i'll respond later, you can ignore this for now haha

proto/options.proto
nilsnolde: went the way of a new macro and no `oneof`

proto/options.proto
nilsnolde: ha, funny, reverting actually pulls up my initial comment again!

nilsnolde:
The completely unrelated `astar_bss` test is failing.. Since I can't figure out what he's complaining about in-code and the VS debugger is a piece of **** (and I don't want to setup CLion right now or go back and forth with gdb's cheat sheet), this will have to wait until tmrw or so.

src/sif/pedestriancost.cc
kevinkreiser: ok this is my actual suggestion to remove the branch:

`factor *= edge->lit() + (!edge->lit() * unlit_factor_);`

kevinkreiser: (maybe just forget about it haha)

kevinkreiser (dismissed):

kevinkreiser:
is it possible we changed costing in the default case? that would make the bss path different for the pedestrian part potentially. i would have expected the code we added would simply be `factor *= 1`. but perhaps we are wrong about that

nilsnolde:
> is it possible we changed costing in the default case?

That's what I figured too, but I just can't see it happening in the code changes.. Still can't really debug, but hopefully on the weekend I'll have more time.

src/sif/dynamiccost.cc
```diff
@@ -354,6 +358,11 @@ void DynamicCost::set_use_living_streets(float use_living_streets) {
              2.f * (1.f - use_living_streets) * (1.f - kMinLivingStreetFactor));
 }
 
+void DynamicCost::set_use_lit(float use_lit) {
+  unlit_factor_ =
+      use_lit < 0.5f ? kMinLitFactor + 2.f * use_lit : ((kMinLitFactor - 5.f) + 12.f * use_lit);
```
nilsnolde, line 363: > but I just can't see it happening in the code changes.

OK, I'm just blind (again).. There was a tiny bug there @chrstnbwnkl which caused the default to be a `unlit_factor_` of 2 (as it must've been, not sure why I didn't see it..).

nilsnolde (approved):
This is done now. I also renamed the `astar_bss.cc` test file to `astar_bikeshare.cc` as that caused problems with VS Code's gdb implementation.

src/sif/pedestriancost.cc
nilsnolde: @chrstnbwnkl actually implemented your suggestion:)

nilsnolde (approved):

src/sif/dynamiccost.cc
```diff
@@ -92,6 +95,7 @@ constexpr float kDefaultUseFerry = 0.5f;         // Default preference of using
 constexpr float kDefaultUseRailFerry = 0.4f;     // Default preference of using a rail ferry 0-1
 constexpr float kDefaultUseTracks = 0.5f;        // Default preference of using tracks 0-1
 constexpr float kDefaultUseLivingStreets = 0.1f; // Default preference of using living streets 0-1
+constexpr float kDefaultUseLit = 0.f;            // Default preference of using lit ways 0-1
```
gknisely, line 98: @chrstnbwnkl why a default of 0.f vs 0.5f?  Meaning no preference by default.

src/sif/pedestriancost.cc
```diff
@@ -89,6 +89,8 @@ const std::string kDefaultPedestrianType = "foot";
 // hills) to 1 (take hills when they offer a more direct, less time, path).
 constexpr float kDefaultUseHills = 0.5f;
 
+constexpr float kDefaultUseLit = 0.f;
```
gknisely, line 92: @chrstnbwnkl same comment as above.  Why not 0.5f for the default.

gknisely: @chrstnbwnkl The reason I am asking about this is that I am wondering what the differences are between master and this branch when running pedestrian routes from here https://github.com/valhalla/valhalla/tree/master/test_requests  I am wondering if there could be a negative impact.  May need to add new tests with areas that have lit tag prevalent and areas where it is not.

src/sif/dynamiccost.cc
nilsnolde: The concept with `lit` is to _avoid_ unlit edges and not do any modulation of lit edges. Then 0 has no influence at all, so it's fully opt-in for clients.

src/sif/pedestriancost.cc
nilsnolde: If you're not changing the `use_lit` request option there shouldn't be any differences at all compared to master. Only when you start with `use_lit > 0` in costing options, there'll be differences. We thought we could avoid doing a RAD testing tbh :innocent: since the impact is very well imaginable IMO, see the graph in @chrstnbwnkl 's OP. Do let us know if you're somewhat concerned, maybe we can talk about it tmrw in the 8:30 review session if you have the time @gknisely ?

kevinkreiser (approved):

PR #3961, opened 2023-02-11
dont require precommit for proper formatting

kevinkreiser:
i was starting on adding a feature to the extract building python script and i hated that in order to format the code i need to run a precommit hook. i dont like to wait to do the formatting until im ready to commit (yes i know i can force a hook to run without committing but again, its annoying, the format script is for linting). i prefer to format on every save. my ide does this for me and it does it via the format script.

this pr moves the python linting out of pre-commit hooks and puts them into the format script where they belong. the format script is still run by the precommit hook so those who do like this workflow will be unchanged. but those of us that like to format often will now be able to do so in a less kludgy way.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 repos:
-  # C++
+  # liniting c++ and python
   - repo: local
     hooks:
     - id: format
@@ -7,27 +7,3 @@ repos:
       entry: bash ./scripts/format.sh
       language: system
       pass_filenames: false
-
-  # Python formatter
-  - repo: https://github.com/psf/black
-    rev: 22.10.0
-    hooks:
-      - id: black
-        language_version: python3
-        args: ['--line-length=105', '--skip-string-normalization']
-        files: ^(src/bindings/python/|scripts/)
-
-  # Python linter
-  - repo: https://github.com/pycqa/flake8
-    rev: 5.0.4
-    hooks:
-      - id: flake8
-        # E501: ignore line length (black auto-formats)
-        # E731: ignore replace-lambda-with-def warnings
-        # E203: ignore "whitespace before ':'"
-        args: [
-          '--max-line-length=105',
-          '--extend-ignore=E501,E731,E203',
-          '--extend-exclude=src/bindings/python/__init__.py'
-        ]
-        files: ^(src/bindings/python/|scripts/)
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,33 +1,9 @@
...
-      - id: flake8
-        # E501: ignore line length (black auto-formats)
-        # E731: ignore replace-lambda-with-def warnings
-        # E203: ignore "whitespace before ':'"
-        args: [
-          '--max-line-length=105',
-          '--extend-ignore=E501,E731,E203',
-          '--extend-exclude=src/bindings/python/__init__.py'
-        ]
-        files: ^(src/bindings/python/|scripts/)
```
kevinkreiser, line 33: each of these hooks is a 1-liner in the format script now, which is still called above so pre-commit still does the work if you like it that way

scripts/bash_utils.sh
```diff
@@ -28,18 +28,15 @@ function setup_mason {
 
 }
 
-function setup_pre_commit {
+function setup_python {
```
kevinkreiser, line 31: this function just finds the binary path for python now

scripts/format.sh
```diff
@@ -22,3 +21,15 @@ echo "Using clang-format $CLANG_FORMAT_VERSION from ${CLANG_FORMAT}"
 
 find src valhalla test bench -type f -name '*.h' -o -name '*.cc' \
   | xargs -I{} -P ${NPROC} ${CLANG_FORMAT} -i -style=file {}
+
+
+# Python setup
+py=$(setup_python)
+${py} -m pip install black==22.10.0 flake8==5.0.4
```
kevinkreiser, line 28: same versions from the hooks so now formatting diffs

scripts/format.sh
```diff
@@ -22,3 +21,15 @@ echo "Using clang-format $CLANG_FORMAT_VERSION from ${CLANG_FORMAT}"
...
+python_sources=$(LANG=C find scripts src/bindings/python -type f -exec file {} \; | grep -F "Python script" | sed 's/:.*//')
```
kevinkreiser, line 29: uses `type` to figure out if a file is a python script or not. we dont always use the file extensions because some of the scripts are part of the installation package

kevinkreiser:
i dont think a changelog is needed but let me know if anyone disagrees

nilsnolde (approved):
LGTM.

Just wondering: doesn't CLion save on a every key stroke? I have a similar processor as you, running `format.sh`  takes a good second or two for me.

Since nothing really changes in user experience and all is done under the hood, I also think changelog isn't needed.

Issue #3770, opened 2022-09-30
Return costing options used for processing


nilsnolde:
As @kevinkreiser [suggested](https://github.com/valhalla/valhalla/discussions/3767#discussioncomment-3756433), we could and probably should return the costing options we ended up using, e.g. in case nothing was specified and we use defaults which can be very confusing to users. Even "worse" when we use the default when a costing option was out of range, but that's actually a point for #3761 :smile:

debasis9658:
Hey @nilsnolde , I would like to work on this issue. It would be great if I can be assigned to this. Thank You

kevinkreiser:
the primary work to be done is adding code to the serializers in Tyr to make json out of the costing options object that is attached to the request then we have to decide where it fits in the rest of the response json and write unit tests. Mayne you could put your thoughts or questions here @debasis9658

debasis9658:
Sure @kevinkreiser . Thanks

debasis9658:
hey @kevinkreiser , Where can I find more information regarding the costing options object and what other dependencies I will need for costiong_option_serializer?

nilsnolde:
The costing options are defined in protobuf, you'll find it in https://github.com/valhalla/valhalla/blob/36bd2ea980c9fd7d51d323150dbda3b95256c5d4/proto/options.proto#L328

Beware it pulls in some other proto objects like `Costing`. The `costing_options_serializer` will likely be a huge function, but very mechanical. You can put it in the common `tyr/serializers.cc` file and call it from the individual action serializers. I didn't quite figure out in my mind yet
- whether we simply serialize the whole options object every time or limit it to the ones which make sense for the requested endpoint
- if we want to make it optional to return the options object in JSON since it's quite a lot of bytes added to the response (but on the other hand it'll be painful in the code and to maintain)
- other open questions?

Those can be answered once we're further into it IMO, it won't affect the general design too much. So you can start to even have anything output in the response relating to the options, then we'll discuss further. 99% of the work will just be done in that one file `serializers.cc` with your new function accepting a PBF options object reference.

nilsnolde:
Oh sorry, screw that.. We won't serialize that kind of Options object. Actually we only need the _costing_ options serialized, since there's the stuff we can mess with while processing the request:

https://github.com/valhalla/valhalla/blob/36bd2ea980c9fd7d51d323150dbda3b95256c5d4/proto/options.proto#L86

It's part of the top-level options object and defined here: 

[...]

PR #3962, opened 2023-02-11
Serialization of costOption demo

debasis9658:
#3770 

 Added serilaizeCostOption function which will serilaize the costOption.

This is a demo function as I have only implemented the json formatting for only two attributes.

Review and comment

### DIFF
--- a/src/tyr/serializers.cc
+++ b/src/tyr/serializers.cc
@@ -356,6 +356,14 @@ json::ArrayPtr intermediate_waypoints(const valhalla::TripLeg& leg) {
 }
 
 // Function for serialization of cost options
+void serializeCostOptions(const valhalla::Api& api, rapidjson::writer_wrapper_t& writer) {
+  writer.start_object();
+  writer("maneuver_penalty", api.options().maneuver_penalty());
+  writer("gate_cost", api.options().gate_cost());
+  writer.end_object();
+}
+
+// Function for serialization of cost options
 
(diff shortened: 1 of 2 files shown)


nilsnolde:
Jep, goes in the right direction. Now you'll have to call that function in the action serializers. Eventually we'll have to overload the function you created with the variant [you had before](https://github.com/valhalla/valhalla/issues/3770#issuecomment-1416707909), I kinda forgot that most serializers are still using the legacy way.

I suggest you start with `route_serializer_valhalla.cc` which passes the same serializer class I asked you to use. Then write a small test in `test/gurka` which proves that the costing object is indeed returned on a route request. Have a look at other gurka tests, it's pretty self-explanatory. Then we can go ahead with the other actions.

nilsnolde:
Also, you'll need to follow the dev guidelines: https://github.com/valhalla/valhalla/blob/master/CONTRIBUTING.md. As you can see CI is failing because the code is not formatted.

debasis9658:
image
@nilsnolde kindly have a look at the screenshot. I am not able to resolve the error "git failed". I am in the git directory but it is still showing the error. How can it be resolved. Tried everything.

nilsnolde:
You’re on Windows. Surprised you came this far. You’ll need Linux or Mac to develop Valhalla.

debasis9658:
Yeah, Actually I am using linux virtual environment. Figured out that issue. Thanks @nilsnolde

nilsnolde:
Are you still on this @debasis9658 ?

PR #3964, opened 2023-02-11
fixed segfault in TimeDistanceMatrix

### LINKED ISSUE #3963: Pedestrian matrix segfaults
One of our other projects uses Valhalla requests in unit testing and a ped matrix in Andorra segfaulted the docker container. Last we ran those test is a good 6 months ago, so some regression must've happened in between. Since it's only the TimeDistanceMatrix, likely I can blame myself when merging #3773 (although the FOSSGIS instance doesn't show that bug, and that's only a 2 weeks old deployment, could also be that it's only happening in extracts though). 

Will try to fix soon.. Here's the request:

```
{
	"costing": "pedestrian",
	"sources": [
		{
			"lon": 1.539283,
			"lat": 42.612798
		},
		{
			"lon": 1.543225,
			"lat": 42.523528
		},
		{
			"lon": 1.709114,
			"lat": 42.542892
		}
	],
	"targets": [
		{
			"lon": 1.539283,
			"lat": 42.612798
		},
		{
			"lon": 1.543225,
			"lat": 42.523528
		},
		{
			"lon": 1.709114,
			"lat": 42.542892
		}
	]
}
```

nilsnolde:
fixes #3963 

Hmpf, I messed up on #3849 it seems. We didn't check anymore if the label indices, coming out of the TimeDistanceMatrix adjacency list, are valid.

### DIFF
--- a/src/thor/timedistancematrix.cc
+++ b/src/thor/timedistancematrix.cc
@@ -35,13 +35,15 @@ std::string GetDateTime(const std::string& origin_dt,
     return origin_dt;
   }
   graph_tile_ptr tile = nullptr;
-  // get the timezone of the output location
-  auto out_nodes = reader.GetDirectedEdgeNodes(pred_id, tile);
   uint32_t dest_tz = 0;
-  if (const auto* node = reader.nodeinfo(out_nodes.first, tile))
-    dest_tz = node->timezone();
-  else if (const auto* node = reader.nodeinfo(out_nodes.second, tile))
-    dest_tz = node->timezone();
+  if (pred_id.Is_Valid()) {
+    // get the timezone of the output location
+    auto out_nodes = reader.GetDirectedEdgeNodes(pred_id, tile);
+    if (const auto* node = reader.nodeinfo(out_nodes.first, tile))
+      dest_tz = node->timezone();
+    else if (const auto* node = reader.nodeinfo(out_nodes.second, tile))
+      dest_tz = node->timezone();
+  }
 
   auto in_epoch =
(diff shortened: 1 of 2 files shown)


src/thor/timedistancematrix.cc
```diff
@@ -250,15 +252,16 @@ std::vector<TimeDistance> TimeDistanceMatrix::ComputeMatrix(
...
       if (predindex == kInvalidLabel) {
         // Can not expand any further...
         one_to_many = FormTimeDistanceMatrix(graphreader, origin.date_time(),
-                                             time_info.timezone_index, pred.edgeid());
+                                             time_info.timezone_index, GraphId{});
         break;
       }
 
+      // Copy the EdgeLabel for use in costing
+      EdgeLabel pred = edgelabels_[predindex];
```
nilsnolde, line 263: moved the copy below the index check again

src/thor/timedistancematrix.cc
```diff
@@ -35,13 +35,15 @@ std::string GetDateTime(const std::string& origin_dt,
...
   }
   graph_tile_ptr tile = nullptr;
-  // get the timezone of the output location
-  auto out_nodes = reader.GetDirectedEdgeNodes(pred_id, tile);
   uint32_t dest_tz = 0;
-  if (const auto* node = reader.nodeinfo(out_nodes.first, tile))
-    dest_tz = node->timezone();
-  else if (const auto* node = reader.nodeinfo(out_nodes.second, tile))
-    dest_tz = node->timezone();
+  if (pred_id.Is_Valid()) {
```
nilsnolde, line 39: now tile id might not be valid

dnesbitt61 (approved):

PR #3966, opened 2023-02-12
Add configurable request timeouts.

### LINKED ISSUE #3965: Cannot set request timeouts
Given the difficulty of estimating request "cost" based on derivative (to complexity) attributes such as distance or number of inputs, stable operation of Valhalla requires a request timeout to prevent overly complex requests from running indefinitely.

[prime_server already supports request timeouts](https://github.com/kevinkreiser/prime_server/blob/master/prime_server/prime_server.hpp#L77), but the [default](https://github.com/kevinkreiser/prime_server/blob/master/prime_server/prime_server.hpp#L57) is "infinite".

Valhalla should allow this to be configured.

lseelenbinder:
# Issue

Fixes #3965.

## Tasklist

 - [ ] Add tests
 - [ ] ~If you made changes to the lua files, update the [taginfo](taginfo.json) too.~ N/A

## Requirements / Relations
None.

## Followup

How should I test this @nilsnolde / @kevinkreiser?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
    * CHANGED: Build skadi with -Werror [#3935](https://github.com/valhalla/valhalla/pull/3935)
    * ADDED: Connect transit tiles to the graph [#3700](https://github.com/valhalla/valhalla/pull/3700)
    * CHANGED: switch to C++17 master branch of `just_gtfs` [#3947](https://github.com/valhalla/valhalla/pull/3947)
+   * ADDED: Support for configuring a universal request timeout [#3966](https://github.com/valhalla/valhalla/pull/3966)
 
 ## Release Date: 2023-01-03 Valhalla 3.3.0
 * **Removed**
--- a/scripts/valhalla_build_config
+++ b/scripts/valhalla_build_config
@@ -156,6 +156,7 @@ config = {
             'interrupt': 'ipc:///tmp/interrupt',
             'drain_seconds': 28,
             'shutdown_seconds': 1,
+            'timeout_seconds': -1,
         }
     },
     'service_limits': {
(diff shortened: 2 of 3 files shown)


kevinkreiser (approved):

kevinkreiser:
there is one caveat to this which might be confusing to some people. if you are running the valhalla service in multiprocess mode via supervisor (which i highly recommend for operational robustness) this config value will not be honored. to anyone who understands how valhalla is architected this is obvious however i would wager that list of people is pretty short (even with the projects popularity). i dont feel strongly about it but i wondered if you or anyone else thinks we need a disclaimer somewhere. probabliy people who run in multi-process mode are advanced enough to realize that running `prime_httpd` as its own process means you have the timeout to that binary to configure the time out. anyway just wanted to mention it!

kevinkreiser:
no test needed in this repo, its prime_server's responsibilty for the functionality to work properly and it already has unit testing around this functionality

lseelenbinder:
That makes sense @kevinkreiser. I think you're probably right, especially since it takes some digging to even see how to setup multi-process mode.

I think I could create an issue to support request timeouts in multi-process mode, though...that seems like something we'd want?

nilsnolde:
> I think I could create an issue to support request timeouts in multi-process mode

There's not much to do there, it's just an argument to `prime_httpd` exe: https://github.com/kevinkreiser/prime_server/blob/eaa396342417323d1a80fc649ddc66cc5a221dbc/src/prime_httpd.cpp#L15. 

You'd set up one of those, then a few `prime_proxyd` and a few loki, thor & odin workers, all in their own process. The stability aspect is that one process can die without problems, as others are still there to service requests. And supervisor will automatically restart failed processes. Better than a single process only threading as `valhalla_service`.

lseelenbinder:
@nilsnolde, oh, of course, good call. I didn't realize you could easily pass request timeouts to the server in that case (and didn't look further).

nilsnolde:
> i wondered if you or anyone else thinks we need a disclaimer somewhere

IMO it's not needed. 99.9% of people run valhalla_service AFAIK and `drain_seconds` etc is the same concept.

PR #3972, opened 2023-02-14
optionally add platforms for pedestrian access

### LINKED ISSUE #3691: No routing over ways with highway=platform
Pedestrian routing is not possible using ways tagged with highway=platform. This tag is often used when bus platform is located directly on sidewalk and the way with this tag creates a gap. Any ideas how to resolve this issue? As I remember, ways with highway=platform are not included in graph for pedestrian routing, is this correct?

image

https://valhalla.openstreetmap.de/directions?profile=pedestrian&wps=19.006070421201006%2C50.30281311002737%2C19.006421921221776%2C50.302628515537414

nilsnolde:
fixes #3691 

Adds a config option to include `highway=platform` edges in the graph, defaulting to `false` so no one gets taken by surprise (also no throwing if it's not in the config JSON). No other special treatment.

Didn't review the entire graph.lua if it has any weird side effects, but probably I should do that.. Does anyone have any concern off the top of their head?

Also not sure if it needs any special handling for narration @dgearhart or others?


lua/graph.lua
```diff
@@ -1474,6 +1475,8 @@ function filter_tags_generic(kv)
         use = 27
      elseif kv["highway"] == "pedestrian" then
         use = 28
+     elseif kv["highway"] == "platform" then
+        use = 35
```
nilsnolde, line 1479: this branch needs to be before the next one. in fact, I wonder if the `bridleway` is not too low in the hierarchy? if a bridleway only has `pedestrian_forward` at this stage it doesn't get a bridleway use.

szymone:
@nilsnolde Thanks a lot!

PR #3974, opened 2023-02-15
Fix stray NULLs in logs.

lseelenbinder:
# Issue

The current logging code results in a `NUL` character between the timestamp and the rest of the message, which throws some log consumers (such as journald) for a loop.

This fixes it, and also removes the suspect usage of `snprintf` with a non C-string.

## Tasklist

 - [ ] Add tests - Current tests appear to cover the correctness of the logs; happy to add a test for there being no NULL in them.
 - [ ] ~If you made changes to the lua files, update the [taginfo](taginfo.json) too.~

## Requirements / Relations


kevinkreiser (changes requested):
i wrote all this stuff this way on purpose :smile: 

my goal was to have logging have as little impact as possible on the runtime performance of the code. this change does two things that i think are against that goal:

1. allocating 400 bytes when we know the exact number of bytes that are needed is wasteful
2. allocating a c string just to reallocate space to copy it into a std::string is wasteful

can we instead tackle the problem you are facing, which is `\0` in your downstream consumer? looking at the code its not clear to me why there would be a null terminator in the final output but we can certainly test to find out why. let me check what `snprintf` says it does...

> A terminating null character is automatically appended after the content written.

ok so yeah that was my fault, i guess its putting the 5th decimal place there and then instead of the 6th it puts the `\0`. i think we should instead just leave space for the null terminator and pop it off

src/midgard/logging.cc
```diff
@@ -28,6 +28,9 @@ inline std::tm* get_gmtime(const std::time_t* time, std::tm* tm) {
 #endif
 }
 
+// Choose something high to avoid format-truncation warnings.
+const size_t TIMESTAMP_SIZE = 400;
+
```
kevinkreiser, line 33:
```suggestion
```

src/midgard/logging.cc
```diff
@@ -38,10 +41,10 @@ std::string TimeStamp() {
   std::chrono::duration<double> fractional_seconds =
       (tp - std::chrono::system_clock::from_time_t(tt)) + std::chrono::seconds(gmt.tm_sec);
   // format the string
-  std::string buffer("year/mo/dy hr:mn:sc.xxxxxx");
-  snprintf(&buffer.front(), buffer.length(), "%04d/%02d/%02d %02d:%02d:%09.6f", gmt.tm_year + 1900,
+  char buffer[TIMESTAMP_SIZE];
+  snprintf(buffer, TIMESTAMP_SIZE, "%04d/%02d/%02d %02d:%02d:%09.6f", gmt.tm_year + 1900,
```
kevinkreiser, line 45:
```suggestion
  std::string buffer("year/mo/dy hr:mn:sc.xxxxxx0");
  snprintf(&buffer.front(), buffer.length(), "%04d/%02d/%02d %02d:%02d:%09.6f", gmt.tm_year + 1900,
```

src/midgard/logging.cc
```diff
@@ -38,10 +41,10 @@ std::string TimeStamp() {
...
            gmt.tm_mon + 1, gmt.tm_mday, gmt.tm_hour, gmt.tm_min, fractional_seconds.count());
-  return buffer;
+  return {buffer};
```
kevinkreiser, line 47:
```suggestion
  buffer.pop_back();
  return buffer;
```

lseelenbinder:
Hey @kevinkreiser, all those thoughts crossed my mind. Understand where you're coming from.

The issue for me is:

1) The `\0` messing with log consumers.
2) It appears using `snprintf` isn't the most correct with a `std::string`. I'm not a deep C++ expert, so I'll leave that to your discretion though.

If we can do that (and avoid the compiler warning-turned-errors) without introducing performance annoyances and extra allocations, I prefer that, too!

kevinkreiser:
@lseelenbinder would you mind giving my suggestions a try? i would be very grateful to see if my reading of the problem addresses it! also thank you very much for looking into it!

lseelenbinder:
Thanks! I'll take a look in the morning. :smile:

...too bad we can't use C++ 20 with proper std::format!

kevinkreiser:
regarding:

> It appears using snprintf isn't the most correct with a std::string. I'm not a deep C++ expert, so I'll leave that to your discretion though.

i can see why that might be a bit confusing, its definitely in the realm of "trick"

`snprintf` just needs an array of `char`, its true that it pre-dates `std::string` but at the end of the day the `std::string` must conform to spec, which means it has to:

1. have contiguous allocation AND
2. use `char` for its template parameter `CharT`

in other words, `std::string` is required to be just a wrapper around `char[]`. so as long as we dont mess with its allocation (which we cant) we are free to do whatever we want with the bytes it owns, including overwriting them. after all that is what the `::operator[]` does. and indeed the `::front` method just gives us a reference to the first element which we can turn into a pointer to treat like the array that it comes from

lseelenbinder:
Ah, perfect. That's the deep C++ magic I haven't had to plumb the depths of. :) Thanks for unpacking it.

kevinkreiser:
actually, we want to go to c++20 next, we're just letting the dust settle a little bit on c++17 before going to c++20. im not sure about the performance of `std::format` (hopefully better than `boost::format` from which it was inspired) but its definitely more convenient for this and for all the stuff we do over in the narrative generation part of the library

nilsnolde:
Also worth noting that std::format, last I checked a few weeks back when we had the discussion how to do the transition, it was only gcc 13.x which had full support and was at the time not even properly released (probably is by now). I still don’t have it on arch at least. So I think we’re not yet missing out on too much:)

lseelenbinder:
Looks good @kevinkreiser! Thanks for the fix and explanations. :smile:

kevinkreiser (approved):

Issue #3975, opened 2023-02-16
timedistance matrix limit warning shows up always


nilsnolde:
Apparently the warning about exceeding the distance limit for time-aware matrices is showing on all requests, not only the ones with `date_time` enabled..

nilsnolde:
fixed

PR #3976, opened 2023-02-16
fix matrix warning about time-aware limit exceeded

nilsnolde:
quick fix for #3975 

neither test nor changelog ok?


kevinkreiser (approved):

kevinkreiser:
oops looks like the unit test needs updating too!

PR #3981, opened 2023-02-18
add source & target index to error/warning 

### LINKED ISSUE #3958: Respond with offending source/target for `max_matrix_distance` loki error
As discussed in https://github.com/valhalla/valhalla/issues/3898#issuecomment-1377637254, we don't want to be too lenient on clients sending matrices which exceed the distance limit on one or more pairs.

However, we could at least respond with the source/target index which triggered the exception. Of course it could be multiple source/target pairs, but I'd not enter that complication. In worst case, the client will fail multiple times in a row, it doesn't matter so much for server performance.

nilsnolde:
fixes #3958 

Lets a user know which source/target id is the problem. Though on second thought it's not that super informative; if one pair is off, there's very likely more. Boils down again to people doing the distance checks on their side. Anyways, now that the "work" is done, a merge doesn't hurt either?;) Also adds the ability to pass extra info to warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
    * ADDED: Support for configuring a universal request timeout [#3966](https://github.com/valhalla/valhalla/pull/3966)
    * ADDED: optionally include highway=platform edges for pedestrian access [#3971](https://github.com/valhalla/valhalla/pull/3971)
    * ADDED: `use_lit` costing option for pedestrian costing [#3957](https://github.com/valhalla/valhalla/pull/3957)
+   * ADDED: source & target index when exceeding max distance limit config [#3981](https://github.com/valhalla/valhalla/pull/3981)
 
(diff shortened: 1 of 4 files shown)


nilsnolde:
Just saw this one again, kinda forgot about it.. I'd still merge it, even though it's only semi-useful, it's slightly better than before.

nilsnolde:
I realized I didn't include any testing. Let me do that before..

javiercaste:
Hi, are there any plans to bump this PR? It's kinda hard right now to understand if a location is unreachable cause the whole calculation fails.

nilsnolde:
ha, we just went through a few older stale ones a few minutes ago, but didn't make it down to 3 years old ones :D I'll brush it up, totally agree

javiercaste:
> ha, we just went through a few older stale ones a few minutes ago, but didn't make it down to 3 years old ones :D I'll brush it up, totally agree

Thank you!

Issue #3969, opened 2023-02-13
Walking inclined paths


hungerburg:
There are several crude approximations for estimating time when walking non flat terrain.

E.g. for this walk from the train station to the hospital - https://valhalla.openstreetmap.de/directions?profile=pedestrian&wps=11.249715685844423%2C47.28104865174184%2C11.24449610710144%2C47.2835305523123 Valhalla estimates 9min20sec. Very sportive :) Walking speed set to 4km/h estimates 12 minutes, still not bad for this journey :)  For the record:

On https://en.wikipedia.org/wiki/Naismith%27s_rule Under Langmuir corrections it says: 4 km/h distance + 450 m/h of ascent, that makes

```
600 m => 1/40*6 => 0.15h
+ 1/450*90      => 0,2h
                   ----
                   0,35h => 21min

```
On https://www.alpenverein.at/portal/news/aktuelle_news_kurz/2018/2018_06_14_wie-berechnet-man-die-gehzeit-auf-wanderwegen.php it says: 300m/h ascent, 500m/h descent, 4km/h distance and a formula: The smaller value gets halved and added to the larger one:

```
90m elevation => 90/300   => 0.3h
600m distance => 1/400*60 => 0.15h
                             ----
0.3+0.15/2                => 0.375 => 22.5min
```

Estimates on guideposts usually have a granularity of quarters of an hour, so a guidepost there would read half an hour. This issue not a bug report, just an idea for enhancement. Though, if you care for hiking, this should get in, because wrong estimates on duration are in the top reasons for people calling emergency.

kevinkreiser:
not too long ago we did try to modulate the speed of the pedestrian costing by the grade (slope) of a given edge in the graph: #3234 

have a read of that and see how we ended up implementing it. basically what we store in the graph is an "average" slope per graph edge and then we have a look up table (the slopes are broken into only 16 values from 'very much downhill' to 'flat' to 'very much uphill') which matches a given slope to a given multiplier that modifies ones walking speed.

in the end it is possible for us to measure total ascent but we decided not to do it that way and instead relied on a  weighted average slope value. its possible that those values could be modified but they did have some scientific grounding. at any rate, i wanted to let you know that at least we do try to modify the speed based on the ascent or descent we think you are going to be doing.

[...]

PR #3982, opened 2023-02-20
More conservative estimates for walking slopes

hungerburg:
# Issue

https://github.com/valhalla/valhalla/issues/3969

## Tasklist

 - [ ] Add a reminder to re-evaluate, once slopes steeper than 15% are taken into account


nilsnolde:
It needs to get formatted properly. So you'll need to run `./scripts/format.sh`. Also the changelog entry I mentioned.

src/sif/pedestriancost.cc
```diff
@@ -165,29 +165,42 @@ BaseCostingOptionsConfig GetBaseCostOptsConfig() {
...
+    1.00f,  //   0.0% - 1.00
+    1.10f,  //   1.5% - 1.06
+    1.20f,  //   3.0% - 1.13
+    1.33f,  //   5.0% - 1.22
+    1.43f,  //   6.5% - 1.30
+    1.57f,  //   8.0% - 1.38
+    1.83f,  //  10.0% - 1.49
+    2.03f,  //  11.5% - 1.58
+    2.23f,  //  13.0% - 1.68
+    2.50f   //  15.0% - 1.82
```
kevinkreiser, line 203: i know this is a nitpick and im sorry in advance, what is the last number in the comment supposed to mean?

i had though it was the "effective speed in MPH" after the factor is applied. so for example in the old code `-10% = 3.16 / .67 = 4.71mph` but even looking at the old values it doesnt hold true for all of them!

and in your new comment i thought maybe it was just the reciprocal of the factor so like 1/1.06  = .94 but nearer to the ends of the scales this doesnt hold.

apologies if this is obvious, for whatever reason im just not getting it at the moment!

hungerburg: Sure, this is not prperly explained: The last number is the one from Mountain Tactical Institute formula, i.e. `math.exp(-0.04*slope)` ; Its written there by my script https://gist.github.com/hungerburg/4936fa430552d625a0c8677f46492d1e - it is very similar to what is there right now, and in ascent also very similar to the 400/800 variant of the DIN recipe.

kevinkreiser: thanks for that! do you mind adding that info in comment form (the formula)?

kevinkreiser (dismissed):
LGTM! thanks very much for this. i think if we can just do the one comment update and run the formatting script we'll be good to go. let me know if you want help running the script

nilsnolde:
And the changelog is important in this case:)

hungerburg:
Finally got what Changelog is about :) Sorry, I cannot run the script, would need to set up the whole build environment? I made the change completely in github web ui. I cannot amend from there, so before adding explanation of last column in table, can I do the formatting manually?

src/sif/pedestriancost.cc
nilsnolde: Don't worry, I'll do the format script for you, it's quickly done. Can you amend the comment before? Otherwise it'll likely get messed up again. Thanks!

kevinkreiser: it was just a single space between the comment and the value, just did it with the github editor

kevinkreiser (approved):

nilsnolde (approved):

PR #3988, opened 2023-02-24
pass GTFS routing test

nilsnolde:
Mostly fixes some erroneous stuff in the tests and some time related things to **finally pass the transit routing test**!

Things I noticed which need to be fixed or at least looked at at some point:

- [x] add a proper method to `TimeInfo` to advance the `second_of_week` properly with transfer cost in `multimodal.cc`
- [ ] have timeinfo know about the pivot date (probably have a singleton pull it once on first use) so we can serve day
- [x] ~~default a node's timezone to the agency's as most times stops don't have timezones specified (I guess mostly when the feed/agency is operating in a single timezone, so the vast majority)~~ too much work without much benefit. let's see how it goes without overriding a stop's (mostly empty) timezone info with the agency's (required field) and fall back to our own timezone detection. there could be a rogue feed where that breaks, but I wouldn't expect so.
- [x] check `ingest_transit.cc::add_stop_pair_shapes()` again: looks like it's always returning the full shape length as `dist_traveled` for a stop_pair, in fact it should be the distance from the trip's shape origin to the current stop being passed to that function
- [ ] adjust the test so time is not hard-coded but in some relation to the timepoint the test is running, otherwise it'll fail soon
- [ ] generally test much more: start with the route response!
- [x] in `ingest_transit` we don't assign a `shape_id` to a stop_pair. however, if there's shapes.txt, convert_transit would like to use that to set the right `dist_traveled` from the full shape, rather than our interpolated shapes for each transit edge
- [x] ~~we currently assume that there'll be a shape associated with a trip. but that's rarely the case, often shapes.txt is missing. in ingest_traffic we should instead build shapes stuff in the pbf tiles from the stops.~~ actually we do exactly that in convert_transit
- [ ] obviously refactor the entire `convert_transit.cc::AddToGraph` thing, at the minimum it should remove all the crazy variable shadowing 
- [ ] make `kTransitConnection` edges (i.e. osm<->transit connection edges) visible for loki::Search, currently they're missing in the bins; also investigate why they're not being seen for node snaps
- [x] ~~review what it's doing with the different `frequencies.txt`, not sure if that's proper, esp if there's `frequencies.txt` together with the exact stop_times.~~ 
- [ ] `frequencies.txt` and how it's handled currently: if a stop_pair's trip has a frequency, we prefer that and don't create an exact departure for that stop_pair, which is totally fine. However, we assume that stop_pair's `origin_departure_time` is when that stop_pair's departure is and I don't think that's really how it's meant to be. From what I understand (but need to reason over real feeds): if there's an entry in `frequencies.txt` for a trip, commonly there'll one full sequence of all stops in `stop_times.txt` as an example to show how stops are timewise connected, but doesn't necessarily represent the `start_time` for that frequency. That should still come from the frequencies.txt entry, so we need to add it to the PBF definition. We should also add the `exact_times` field: if true, the frequency's `start_time` is the exact start time for the frequency and just rolls with one headway after the other. However, when false, then we can pick an approximate `departure_time` within the algorithm. I think it's fairly trivial to support both and just needs to extend a `TransitDeparture::type_` to be an enum of 3 values (there 4 spare bits, so no problem IMO).
- [ ] currently we only test schedule-based departures: if a frequency is defined we prefer that over exact stop_times, as we should IMO, but currently both test trips have a frequency assigned, need to un-assign one of them and do another routing test for the other trip's valid time range.
- [ ] see if we can't come up with a way to limit the expansion of all those transit-related edges a bit more. Problem is they're so irrelevant, they hardly contribute cost, so they're encountered during the expansion super frequently but never settled.
- [ ] do we need a second pass for multimodal? not sure yet, what would be sensible to relax there, don't think there's anything?
- [x]  ~~transit schedules are only valid for 60 days after the transit tiles were created, why? There's a 64 bit mask, seems like the reason on first glance.~~ (reasoning was that tilesets should be updated more often than 60 days and lots of feeds have validity ranges less than that) So meaning, transit tiles will have to be regenerated in case the feed is valid for longer than that (which I guess is quite common..).
- [ ] implement `transfers.txt` for proper transfer times. Currently it allows for 30 seconds, otherwise it'll wait for the next departure.
- [ ] implement `pathways.txt` maybe, though super low priority, apparently it's hardly ever there and if it is, it's not really correct
- [ ] move the allowed check for continuing expansion based on amount of pedestrian distance walked back into costing. at the moment we pulled it out into the algorithms but now that we have it tracked separately from path distance in the edge label we can indeed check it in the costing again


src/mjolnir/ingest_transit.cc
```diff
@@ -263,6 +263,7 @@ void setup_stops(Transit& tile,
     node->set_traversability(static_cast<uint32_t>(Traversability::kBoth));
   }
   node->set_name(tile_stop.stop_name);
+  // TODO(nils): this is usually empty and should only override an agency's timezone
```
nilsnolde, line 266: this will lead to problems: an agency's timezone is commonly the source of truth. need to fix that too

src/mjolnir/ingest_transit.cc
```diff
@@ -389,33 +390,50 @@ bool write_stop_pair(Transit& tile,
...
 
   // already sorted by stop_sequence
   const auto tile_stopTimes = feed.get_stop_times_for_trip(tile_tripId);
 
   for (size_t stop_sequence = 0; stop_sequence < tile_stopTimes.size() - 1; stop_sequence++) {
-    gtfs::StopTime origin_stopTime = tile_stopTimes[stop_sequence];
-    gtfs::Id origin_stopId = origin_stopTime.stop_id;
-    auto origin_stop = feed.get_stop(origin_stopId);
+    const auto& origin_stopTime = tile_stopTimes[stop_sequence];
+    const auto& origin_stopId = origin_stopTime.stop_id;
```
nilsnolde, line 402: just avoiding a few copies

src/mjolnir/ingest_transit.cc
```diff
@@ -389,33 +390,50 @@ bool write_stop_pair(Transit& tile,
...
+    const auto dest_graphid_it = platform_node_ids.find({dest_stopId, currFeedPath});
     const bool origin_is_in_tile = origin_graphid_it != platform_node_ids.end();
     const bool dest_is_in_tile = dest_graphid_it != platform_node_ids.end();
 
     // check if this stop_pair (the origin of the pair) is inside the current tile
     if ((origin_is_in_tile || dest_is_in_tile) && origin_stopTime.trip_id == dest_stopTime.trip_id) {
+      auto* stop_pair = tile.add_stop_pairs();
 
-      dangles = dangles || !origin_is_in_tile || !dest_is_in_tile;
+      auto* service_dow = stop_pair->mutable_service_days_of_week();
```
nilsnolde, line 418: we weren't actually including the service days of week from the calendar to populate the day-of-week bitmask in `convert_transit.cc`

src/mjolnir/ingest_transit.cc
```diff
@@ -389,33 +390,50 @@ bool write_stop_pair(Transit& tile,
...
+      service_dow->Add(trip_calendar->wednesday == gtfs::CalendarAvailability::Available);
+      service_dow->Add(trip_calendar->thursday == gtfs::CalendarAvailability::Available);
+      service_dow->Add(trip_calendar->friday == gtfs::CalendarAvailability::Available);
+      service_dow->Add(trip_calendar->saturday == gtfs::CalendarAvailability::Available);
+      service_dow->Add(trip_calendar->sunday == gtfs::CalendarAvailability::Available);
+      // skip if no days are valid
+      if (!trip_calendar || !service_dow->size()) {
+        LOG_WARN("Service ID " + currTrip->service_id + " has no calendar entry, skip.");
+        continue;
+      }
```
nilsnolde, line 430: tentative: don't think we need to care about this stop pair when there's no valid days in the week. but gotta validate that this is fine.

src/mjolnir/ingest_transit.cc
```diff
@@ -389,33 +390,50 @@ bool write_stop_pair(Transit& tile,
   const auto& tile_tripId = feed_trip.id;
   const std::string currFeedPath = feed_trip.feed;
 
-  auto currTrip = feed.get_trip(tile_tripId);
+  const auto& currTrip = feed.get_trip(tile_tripId);
+  const auto& currShape = feed.get_shape(currTrip->shape_id);
+  const auto& trip_calendar = feed.get_calendar(currTrip->service_id);
```
nilsnolde, line 395: I moved these out of the stop_pair loop as they don't need to be computed every time

src/thor/multimodal.cc
```diff
@@ -367,8 +367,8 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
 
       // Look up the next departure along this edge
       const TransitDeparture* departure =
-          tile->GetNextDeparture(directededge->lineid(), offset_time.local_time, day_, dow_,
-                                 date_before_tile_, tc->wheelchair(), tc->bicycle());
+          tile->GetNextDeparture(directededge->lineid(), offset_time.local_time % kSecondsPerDay,
```
nilsnolde, line 370: we need the seconds in the week. again, gotta check this holds true for the weirdo sunday-first-day-of-week stuff

test/gurka/test_gtfs.cc
```diff
@@ -75,9 +75,11 @@ boost::property_tree::ptree get_config() {
                            });
 }
 
+// put the base somewhere in toronto for timezone stuff to work
 valhalla::gurka::nodelayout create_layout() {
   int gridsize_metres = 100;
-  auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize_metres, {-0.000001, -0.000001});
+  auto layout =
+      gurka::detail::map_to_coordinates(ascii_map, gridsize_metres, {-79.401798, 43.679187});
```
nilsnolde, line 82: offset the map to toronto, otherwise timezones are off

test/gurka/test_gtfs.cc
```diff
@@ -196,16 +198,48 @@ TEST(GtfsExample, WriteGtfs) {
...
-    stopTime.trip_id = (i < 3) ? tripOneID : tripTwoID;
-    feed.add_stop_time(stopTime);
-  }
+
+  struct StopTime trip_one_stop_one {
+    .trip_id = tripOneID, .stop_id = stopOneID + "_ledge_to_the_train_bucko", .stop_sequence = 0,
+    .arrival_time = Time("6:00:00"), .departure_time = Time("6:00:00"), .stop_headsign = "head",
+    .shape_dist_traveled = 0.0, .timepoint = gtfs::StopTimePoint::Exact,
+  };
+  feed.add_stop_time(trip_one_stop_one);
```
nilsnolde, line 207: I was being lazy doing this.. I create every stop_time by itself since it needs lots of kinda individual info it was lacking before, e.g. the `departure_time` actually needs to progress;)

nilsnolde: reminds me: check how the stop_time's `shape_dist_traveled` relates to the shapes.txt in valhalla

test/gurka/test_gtfs.cc
```diff
@@ -215,19 +249,19 @@ TEST(GtfsExample, WriteGtfs) {
     .tuesday = CalendarAvailability::Available, .wednesday = CalendarAvailability::Available,
     .thursday = CalendarAvailability::Available, .friday = CalendarAvailability::Available,
     .saturday = CalendarAvailability::Available, .sunday = CalendarAvailability::Available,
-    .start_date = Date(2022, 1, 31), .end_date = Date(2023, 1, 31),
+    .start_date = Date(2023, 1, 31), .end_date = Date(2024, 1, 31),
```
nilsnolde, line 252: should do smth smarter here too soon (well, before january 2024;)). get the `now` date and offset a bit. careful since the tile creation date is also significant for the validity checks

test/gurka/test_gtfs.cc
```diff
@@ -248,11 +283,11 @@ TEST(GtfsExample, WriteGtfs) {
 
   // write frequencies.txt
   struct Frequency freqBased {
-    .trip_id = tripOneID, .start_time = Time(0, 0, 0), .end_time = Time(2, 0, 0),
+    .trip_id = tripOneID, .start_time = Time(6, 0, 0), .end_time = Time(22, 0, 0),
```
nilsnolde, line 286: these are redundant for the routing test but still..

test/gurka/test_gtfs.cc
```diff
@@ -398,6 +433,12 @@ TEST(GtfsExample, MakeTile) {
   auto station_two_ll = layout.find("2");
   auto station_three_ll = layout.find("3");
 
+  auto dt = "2023-02-27T05:50";
+  auto dt_date = DateTime::get_formatted_date(dt);
+  auto dt_date_days = DateTime::days_from_pivot_date(dt_date);
+  auto dt_dow = DateTime::day_of_week_mask(dt);
+  bool date_before_tile = false;
```
nilsnolde, line 440: only added these to test the `tile->GetNextDeparture`. think I'll remove it again (or be smarter around choosing a date, see above comment)

test/gurka/test_gtfs.cc
```diff
@@ -494,7 +550,11 @@ TEST(GtfsExample, MakeTile) {
 // TODO: TEST THAT TRANSIT ROUTING IS FUNCTIONAL BY MULTIMOTDAL ROUTING THROUGH WAYPOINTS, CHECK THAT
 // THE TRIP TYPE IS A TRANSIT TYPE
 
-TEST(GtfsExample, DISABLED_testRouting) {
+TEST(GtfsExample, DISABLED_test_routing) {
+
+  // TODO: create a simpler map: currently somehow at edge label 8 or so, it connects back to edge
+  // label 3, which has the origin as predecessor.. smth is wrong with some directed edges.. possibly
+  // with kRail not having opposing edges?
```
nilsnolde, line 557: we're almost there. smth is messed up in the expansion, not sure yet what it is. I'll create a much simpler test map tmrw and iteratively make it more complex to see what's going on.

nilsnolde:
BTW @kevinkreiser , I'm also happy to continue working on this branch until it's done. Shouldn't be much more. In case you wanna have a first look, feel free.

nilsnolde:
Conan being a PITA again..

src/mjolnir/convert_transit.cc
```diff
@@ -100,13 +100,14 @@ struct StopEdges {
   std::vector<TransitLine> lines;                     // Set of unique route/stop pairs
 };
 
-// Get scheduled departures for a stop
+// Get scheduled departures for a stop; here we also look at the .pbf.x files,
+// as there can be only stop pairs in the extended ones
```
nilsnolde, line 104: I did what I didn't want to do: mix variable renaming with functional changes.. I just couldn't hold it, was really illegible and then I didn't want to drop that work either. You can ignore everything except the stuff I'll comment on specifically in this file.

src/mjolnir/convert_transit.cc
```diff
@@ -754,26 +755,21 @@ void AddToGraph(GraphTileBuilder& tilebuilder_transit,
             tilebuilder_transit.AddEdgeInfo(0, station_graphid, egress_graphid, 0, 0, 0, 0, shape,
                                             names, tagged_values, pronunciations, 0, added);
         directededge.set_edgeinfo_offset(edge_info_offset);
-        directededge.set_forward(true);
+        directededge.set_forward(false);
```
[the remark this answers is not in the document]
nilsnolde: here it's from station -> egress, which seems intuitive to have backwards

src/mjolnir/convert_transit.cc
```diff
@@ -886,7 +883,7 @@ void AddToGraph(GraphTileBuilder& tilebuilder_transit,
                                         names, tagged_values, pronunciations, 0, added);
 
     directededge.set_edgeinfo_offset(edge_info_offset);
-    directededge.set_forward(true);
+    directededge.set_forward(false);
```
nilsnolde, line 886: here it's from platform -> station, which seems intuitive to have backwards

src/mjolnir/ingest_transit.cc
nilsnolde: I think the protobuf definition of a bool array is kinda legacy. I filed a request with `just_gtfs` to give us a bit mask for this, which would save quite a few lines of code and make it easier to handle: https://github.com/mapsme/just_gtfs/issues/19

src/mjolnir/ingest_transit.cc
```diff
@@ -590,21 +602,6 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
...
-
-  auto database = pt.get_optional<std::string>("mjolnir.timezone");
-  auto path = pt.get<std::string>("mjolnir.transit_feeds_dir");
-  // Initialize the tz DB (if it exists)
-  sqlite3* tz_db_handle = database ? GetDBHandle(*database) : nullptr;
-  if (!database) {
-    LOG_WARN("Time zone db not found.  Not saving time zone information from db.");
-  } else if (!tz_db_handle) {
-    LOG_WARN("Time zone db " + *database + " not found.  Not saving time zone information from db.");
-  }
```
nilsnolde, line 606: don't need that here.

src/sif/pedestriancost.cc
```diff
@@ -647,8 +647,7 @@ bool PedestrianCost::Allowed(const baldr::DirectedEdge* edge,
                              const uint64_t current_time,
                              const uint32_t tz_index,
                              uint8_t& restriction_idx) const {
-  if (!IsAccessible(edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
-      (edge->surface() > minimal_allowed_surface_) || edge->is_shortcut() ||
```
nilsnolde, line 651: uturn detection was duplicated, I removed the redundant one. we need to check if the predecessor wasn't a transit edge.

src/sif/pedestriancost.cc
```diff
@@ -681,9 +680,9 @@ bool PedestrianCost::AllowedReverse(const baldr::DirectedEdge* edge,
   // Do not check max walking distance and assume we are not allowing
   // transit connections. Assume this method is never used in
   // multimodal routes).
-  if (!IsAccessible(opp_edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
-      (opp_edge->surface() > minimal_allowed_surface_) || opp_edge->is_shortcut() ||
-      IsUserAvoidEdge(opp_edgeid) || edge->sac_scale() > max_hiking_difficulty_ ||
+  if (!IsAccessible(opp_edge) || (opp_edge->surface() > minimal_allowed_surface_) ||
+      opp_edge->is_shortcut() || IsUserAvoidEdge(opp_edgeid) ||
```
nilsnolde, line 684: same here

src/thor/multimodal.cc
```diff
@@ -44,8 +44,8 @@ constexpr uint32_t kInitialEdgeLabelCount = 200000;
 MultiModalPathAlgorithm::MultiModalPathAlgorithm(const boost::property_tree::ptree& config)
     : PathAlgorithm(config.get<uint32_t>("max_reserved_labels_count", kInitialEdgeLabelCount),
                     config.get<bool>("clear_reserved_memory", false)),
-      walking_distance_(0), max_label_count_(std::numeric_limits<uint32_t>::max()),
-      mode_(travel_mode_t::kPedestrian), travel_type_(0) {
+      max_label_count_(std::numeric_limits<uint32_t>::max()), mode_(travel_mode_t::kPedestrian),
```
nilsnolde, line 47: `walking_distance_` was only used in `ExpandForward`, similar to `new_cost`, no need to have it as a class member, was kinda confusing at first.

src/thor/multimodal.cc
```diff
@@ -287,6 +287,9 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
     // any costs along paths and roads. We only do this once
     // so if its from a transition we don't need to do it again
     if (mode_ == travel_mode_t::kPedestrian && !from_transition) {
+      // TODO: this needs to update second_of_week: careful, needs to wrap around the
+      // end of the week and we also need to update the day_ and dow_ mask in that case
+      // maybe make it a TimeInfo function for convenience
```
nilsnolde, line 292: this is not right yet. I left a TODO in the main description of the PR. we're not looking at the `TimeInfo.local_time` because that's already corrected for timezone. we want the time without timezone. eventually it needs a new method in TimeInfo IMO

src/thor/multimodal.cc
```diff
@@ -346,15 +349,16 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
 
     // Reset cost and walking distance
     Cost newcost = pred.cost();
-    walking_distance_ = pred.path_distance();
+    uint32_t walking_distance_ = pred.path_distance();
```
nilsnolde, line 352: should remove the `_` at the end..

src/thor/multimodal.cc
```diff
@@ -365,10 +369,12 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
         continue;
       }
 
+      uint32_t day_seconds = static_cast<uint32_t>(offset_time.second_of_week) % kSecondsPerDay;
```
nilsnolde, line 372: use seconds of the day instead of timezone corrected epoch

src/thor/triplegbuilder.cc
```diff
@@ -920,8 +920,8 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
   }
 
   // Set speed if requested
-  if (controller(kEdgeSpeed)) {
-    // TODO: if this is a transit edge then the costing will throw
+  // TODO: what to do about transit edges?
+  if (controller(kEdgeSpeed) && !(mode == sif::TravelMode::kPublicTransit)) {
```
nilsnolde, line 924: not sure what to do here yet..

src/mjolnir/ingest_transit.cc
nilsnolde: remove entirely in favor of our own stuff

src/mjolnir/ingest_transit.cc
nilsnolde: check again the functions' output here

src/mjolnir/ingest_transit.cc
nilsnolde: default to agency, then override with stops.txt, then fallback to own timezone db

src/sif/pedestriancost.cc
nilsnolde: add a comment for the future

src/thor/multimodal.cc
nilsnolde: maybe use the `TimeInfo::forward` method here too

src/thor/triplegbuilder.cc
nilsnolde: switch to using transit's edgecost instead

test/gurka/test_gtfs.cc
```diff
@@ -475,13 +535,15 @@ TEST(GtfsExample, MakeTile) {
 
     for (uint32_t schedule_it = 0; schedule_it < tile->header()->schedulecount(); schedule_it++) {
       auto* tileSchedule = tile->GetTransitSchedule(schedule_it);
-      EXPECT_EQ(tileSchedule->days(), 0);
-      EXPECT_EQ(tileSchedule->end_day(), 60);
+      // TODO: get the right value here, should be first 60 bits set
+      // EXPECT_EQ(tileSchedule->days(), );
+      // TODO: why not 60 days? Why only 60 days schedule validity at all?
+      EXPECT_EQ(tileSchedule->end_day(), 59);
```
nilsnolde, line 541: why is it 59?

valhalla/sif/edgelabel.h
```diff
@@ -416,7 +416,8 @@ class EdgeLabel {
   // be part of a bit field make sure kInvalidLabel is changed.
   uint32_t predecessor_;
 
-  // path_distance_: Accumulated path distance in meters.
+  // path_distance_: Accumulated path distance in meters. Can have a different meaning per algorithm,
+  //   e.g. represents walking distance in sif/multimodal.cc
```
nilsnolde, line 420: update this to have all distances, including transit, not only walking and re-introduce the member variable for walking distance

valhalla/thor/multimodal.h
```diff
@@ -75,7 +75,6 @@ class MultiModalPathAlgorithm : public PathAlgorithm {
 
 protected:
   // Current walking distance.
-  uint32_t walking_distance_;
```
nilsnolde, line 78: revert

src/sif/pedestriancost.cc
```diff
@@ -647,13 +648,13 @@ bool PedestrianCost::Allowed(const baldr::DirectedEdge* edge,
...
-  if (!IsAccessible(edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
-      (edge->surface() > minimal_allowed_surface_) || edge->is_shortcut() ||
+  if (!IsAccessible(edge) || (edge->surface() > minimal_allowed_surface_) || edge->is_shortcut() ||
       IsUserAvoidEdge(edgeid) || edge->sac_scale() > max_hiking_difficulty_ ||
       (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx() &&
        pred.mode() == TravelMode::kPedestrian) ||
       //      (edge->max_up_slope() > max_grade_ || edge->max_down_slope() > max_grade_) ||
-      ((pred.path_distance() + edge->length()) > max_distance_)) {
+      // path_distance for multimodal is currently checked inside the algorithm
+      (!allow_transit_connections_ && pred.path_distance() + edge->length()) > max_distance_) {
```
nilsnolde, line 657: can't be used with multimodal since `path_distance` is really the whole distance, not only the walking distance. we need to check that in the algorithm unfortunately for now..

src/thor/dijkstras.cc
```diff
@@ -498,7 +498,8 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
 
     // Reset cost and walking distance
     Cost newcost = pred.cost();
-    uint32_t walking_distance = pred.path_distance();
+    uint32_t walking_distance = pred.walking_distance();
+    uint32_t path_distance = pred.path_distance();
```
nilsnolde, line 502: separate walking & path_distance so we can report the right full distance in the summary object

nilsnolde: unfortunately also means it has to be a new field in the MMEdgeLabel

src/thor/multimodal.cc
nilsnolde: we put it now in the MMEdgeLabel

src/thor/multimodal.cc
```diff
@@ -454,6 +445,18 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
...
+      // and the walking distance didn't exceed (can't check in Allowed())
+      if (!mode_costing[static_cast<uint32_t>(mode_)]->Allowed(directededge, is_dest, pred, tile,
+                                                               edgeid, 0, 0, restriction_idx) ||
+          walking_distance > max_walking_dist_) {
+        continue;
+      }
+
+      Cost c = mode_costing[static_cast<uint32_t>(mode_)]->EdgeCost(directededge, tile);
+      c.cost *= mode_costing[static_cast<uint32_t>(mode_)]->GetModeFactor();
+      newcost += c;
```
nilsnolde, line 459: the rest of the changes in this block are just re-ordering: do all the checks where we might skip this edge candidate before actually getting to the cost

src/thor/multimodal.cc
```diff
@@ -523,9 +525,16 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
     // Add edge label, add to the adjacency list and set edge status
     uint32_t idx = edgelabels_.size();
     *es = {EdgeSet::kTemporary, idx};
+
+    std::cout << "Pred: " << std::to_string(pred_idx) << ", Label: " << std::to_string(idx)
+              << ", from " << std::to_string(pred.endnode()) << ", to "
+              << std::to_string(directededge->endnode()) << ", walking "
+              << std::to_string(walking_distance) << ", use" << baldr::to_string(directededge->use())
+              << std::endl;
```
nilsnolde, line 533: will remove before merging. seeing the expansion might be necessary again;)

src/thor/route_action.cc
```diff
@@ -245,6 +245,9 @@ thor::PathAlgorithm* thor_worker_t::get_path_algorithm(const std::string& routet
 
   // Have to use multimodal for transit based routing
   if (routetype == "multimodal" || routetype == "transit") {
+    const auto& max_walking_dist =
+        options.costings().find(Costing::pedestrian)->second.options().max_distance();
+    multi_modal_astar.SetMaxWalkingDist(max_walking_dist);
```
nilsnolde, line 250: urgh.. not sure how to do it any better.. opinions?

src/thor/route_action.cc
```diff
@@ -245,6 +245,11 @@ thor::PathAlgorithm* thor_worker_t::get_path_algorithm(const std::string& routet
 
   // Have to use multimodal for transit based routing
   if (routetype == "multimodal" || routetype == "transit") {
+    const auto& max_walking_dist = options.costings()
+                                       .find(Costing::pedestrian)
```
nilsnolde, line 249: that was a bit confusing, as the docs said it'd belong to the transit options. but really it's parsed in the pedestrian costing, so I changed the docs; IMO it makes more sense on pedestrian.

nilsnolde:
This is ready for review now.. I'll use it on a real feed tmrw, maybe that'll shake some things loose as well. Depending on what it is, I'll solve that in another PR though. This one's big enough as it is.

src/mjolnir/ingest_transit.cc
nilsnolde: this is too much work to be worth it. we default to our own timezone anyways which really shouldn't disagree with an agency's timezone.. if it does, likely the operator messed up and should've overriden this stop's timezone

src/mjolnir/ingest_transit.cc
nilsnolde: these changes are all fine

src/mjolnir/ingest_transit.cc
```diff
@@ -624,15 +621,23 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
...
     auto current_path = tile_path;
 
-    std::unordered_map<feed_object_t, GraphId> platform_node_ids = write_stops(tile, current);
-    write_routes(tile, current);
-    write_shapes(tile, current);
+    // collect all the feeds in this tile
+    feed_cache_t feeds;
+    for (const auto& route : current.routes) {
+      feeds(route.first);
+    }
```
nilsnolde, line 628: only load the feeds we need once and pass it to the functions needing it.

src/thor/triplegbuilder.cc
nilsnolde: replacing it with 
```

  if (controller(kEdgeSpeed) && !(mode == sif::TravelMode::kPublicTransit)) {    
    sif::Cost edge_cost{};
    if (mode == sif::TravelMode::kPublicTransit) {
      const TransitDeparture* dep =
          graphtile->GetTransitDeparture(directededge->lineid(), trip_id,
                                        time_info.second_of_week % kSecondsPerDay);
      if (dep) {
        edge_cost = costing->EdgeCost(directededge, dep, time_info);
      }
    } else {
      uint8_t flow_sources;
      costing->EdgeCost(directededge, graphtile, time_info, flow_sources);
    }
    auto speed = directededge->length() / edge_cost.secs * 3.6;
    trip_edge->set_speed(speed);
  }
  ```

throws a 
```

  if (controller(kEdgeSpeed) && !(mode == sif::TravelMode::kPublicTransit)) {
    // TODO: could get better precision speed here by calling GraphTile::GetSpeed but we'd need to
    // know whether or not the costing actually cares about the speed of the edge. Perhaps a
    // refactor of costing to have a GetSpeed function which EdgeCost calls internally but which we
    // can also call externally
    
    sif::Cost edge_cost{};
    if (mode == sif::TravelMode::kPublicTransit) {
      const TransitDeparture* dep =
          graphtile->GetTransitDeparture(directededge->lineid(), trip_id,
                                        time_info.second_of_week % kSecondsPerDay);
      if (dep) {
        edge_cost = costing->EdgeCost(directededge, dep, time_info);
      }
    } else {
      uint8_t flow_sources;
      costing->EdgeCost(directededge, graphtile, time_info, flow_sources);
    }
    auto speed = directededge->length() / edge_cost.secs * 3.6;
    trip_edge->set_speed(speed);
  }
  ```

src/mjolnir/ingest_transit.cc
```diff
@@ -427,19 +447,18 @@ bool write_stop_pair(Transit& tile,
...
+                              : origin_stopId;
       stop_pair->set_origin_onestop_id(origin_onestop_id);
 
       bool dest_is_generated = tile_info.station_children.find({dest_stopId, currFeedPath}) !=
                                tile_info.station_children.end();
       auto dest_onestop_id = dest_is_generated
-                                 ? dest_stopId
-                                 : dest_stopId + "_" + to_string(NodeType::kMultiUseTransitPlatform);
+                                 ? dest_stopId + "_" + to_string(NodeType::kMultiUseTransitPlatform)
+                                 : dest_stopId;
```
nilsnolde, line 458: these were in the wrong order for the ternary, which broke the stitcher

src/mjolnir/ingest_transit.cc
```diff
@@ -699,20 +700,20 @@ void stitch_tiles(const boost::property_tree::ptree& pt,
       // do while we have more to find and arent sick of searching
       std::set<GraphId, dist_sort_t> unchecked(all_tiles.cbegin(), all_tiles.cend(),
                                                dist_sort_t(current, grid));
+      // the first one will have to be the same tile as current, so remove it
+      unchecked.erase(unchecked.begin());
```
nilsnolde, line 704: the `current` tile will always be the first one, we can remove it here instead of adding a conditional further down

nilsnolde:
But NOW it's done :D

gknisely:
@nilsnolde @kevinkreiser I can review and discuss on Thursday.  Let me know if that works.

src/mjolnir/ingest_transit.cc
```diff
@@ -497,6 +517,7 @@ bool write_stop_pair(Transit& tile,
 
       // get frequency info
       if (!feed.get_frequencies(currTrip->trip_id).empty()) {
+        // TODO(nils): is this correct? a trip has a unique frequency?
```
nilsnolde, line 520: assumption nope

src/mjolnir/ingest_transit.cc
```diff
@@ -741,7 +751,9 @@ void stitch_tiles(const boost::property_tree::ptree& pt,
           // else{ TODO: we could delete this stop pair }
         }
       }
+      lock.lock();
       write_pbf(tile, current_path);
+      lock.unlock();
```
nilsnolde, line 756: read needs to lock too

src/thor/multimodal.cc
```diff
@@ -44,8 +44,8 @@ constexpr uint32_t kInitialEdgeLabelCount = 200000;
 MultiModalPathAlgorithm::MultiModalPathAlgorithm(const boost::property_tree::ptree& config)
     : PathAlgorithm(config.get<uint32_t>("max_reserved_labels_count", kInitialEdgeLabelCount),
                     config.get<bool>("clear_reserved_memory", false)),
-      walking_distance_(0), max_label_count_(std::numeric_limits<uint32_t>::max()),
```
nilsnolde, line 47: maybe initialize again

src/thor/triplegbuilder.cc
nilsnolde, line 924: no edgecost needed for transit edges

kevinkreiser:
@gknisely we just reviewed. from my side its good. the changes are actually pretty small there are just a lot of renames that really make it difficult to review. we could make a summary of the changes at a high level, that might help just for posterity? i'll let it up to @nilsnolde if he wants to wait until thursday for further review

nilsnolde:
I can summarize tmrw, just had a way too long review-addressing session before realizing my mistake.. I'm ok with thursday @kevinkreiser @gknisely

src/sif/pedestriancost.cc
```diff
@@ -257,17 +257,6 @@ class PedestrianCost : public DynamicCost {
...
   }
 
-  /**
-   * This method overrides the max_distance with the max_distance_mm per segment
-   * distance. An example is a pure walking route may have a max distance of
-   * 10000 meters (10km) but for a multi-modal route a lower limit of 5000
-   * meters per segment (e.g. from origin to a transit stop or from the last
-   * transit stop to the destination).
-   */
-  virtual void UseMaxMultiModalDistance() override {
```
gknisely, line 267: I think we need this as we would fail vs just walking.  We saw that we would prefer long walking routes vs taking transit.  

Also, I think we did this for transfers from one line to another.  Think about taking the PATH and walking for awhile then taking the MTA in NYC.  At that time they were separate operators/agencies.  Not sure if the feed is like that now as MTA owns PATH now, but need to test that case again.  It would be more desirable to just transfer at a station that has stops for both operators/agencies.

src/mjolnir/ingest_transit.cc
```diff
@@ -685,7 +698,7 @@ void stitch_tiles(const boost::property_tree::ptree& pt,
     do {
 
       // open tile make a hash of missing stop to invalid graphid
-      auto tile = read_pbf(current_path, lock);
+      auto tile = read_pbf(current_path);
```
nilsnolde, line 701: check again..

src/thor/triplegbuilder.cc
```diff
@@ -920,15 +920,20 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
...
     // TODO: could get better precision speed here by calling GraphTile::GetSpeed but we'd need to
     // know whether or not the costing actually cares about the speed of the edge. Perhaps a
     // refactor of costing to have a GetSpeed function which EdgeCost calls internally but which we
     // can also call externally
-    uint8_t flow_sources;
-    auto speed = directededge->length() /
-                 costing->EdgeCost(directededge, graphtile, time_info, flow_sources).secs * 3.6;
+    double speed = 0;
+    if (mode == sif::TravelMode::kPublicTransit) {
+      speed = directededge->speed();
```
nilsnolde, line 931: look into pathinfo elapsed secs

src/mjolnir/ingest_transit.cc
nilsnolde: this is actually not true: even the Google example feed has multiple frequencies for the same trip id. I'll re-add the comment

src/mjolnir/ingest_transit.cc
nilsnolde: no, this one is fine to not lock @kevinkreiser , just couldn't think of it in the rush of the review. 

this is the tile this thread will process and no other thread will ever write to (only maybe read); the current thread will write this tile only. there's another `read_pbf` further down, which needs locking (where it's trying to find missing nodes in other tiles) and that one is locked.

src/thor/triplegbuilder.cc
nilsnolde: fixed

kevinkreiser (approved):
one thing i regret about our pairing up to do code review is that i dont write any of my review comments in the pr, i just give them to you verbally. so i think it might be harder for people to follow along. i know you usually put a note though while we are reviewing. i think i'll try harder to actually write my review comments in the review while we are going over it.

the reason i say this is because we have reviewed this code together at least 3 or 4 times, and now im putting a ship on it but it looks like i didnt even review it once. i need to improve my workflow here. bear with me :smile:

nilsnolde:
haha yeah, maybe it's more explicit if you write the comments during review.

Issue #3991, opened 2023-02-25
Build LOGGING_LEVEL=ALL failed


noktoborus:
cmake -B build -DCMAKE_BUILD_TYPE=Debug -DLOGGING_LEVEL=ALL
make -C build -j$(nproc)

```
make[2]: Leaving directory '/home/noktoborus/source/valhalla/build'
[ 25%] Built target gmock_main
[ 25%] Building CXX object src/odin/CMakeFiles/valhalla-odin.dir/narrativebuilder.cc.o
[ 25%] Building CXX object src/odin/CMakeFiles/valhalla-odin.dir/util.cc.o
[ 29%] Building CXX object src/baldr/CMakeFiles/valhalla-baldr.dir/curler.cc.o
[ 29%] Building CXX object src/baldr/CMakeFiles/valhalla-baldr.dir/datetime.cc.o
/home/noktoborus/source/valhalla/src/odin/maneuver.cc: In member function ‘std::string valhalla::odin::Maneuver::ToParameterString() const’:
/home/noktoborus/source/valhalla/src/odin/maneuver.cc:1401:29: error: ‘class valhalla::baldr::StreetNames’ has no member named ‘ToParameterString’
 1401 |   man_str += street_names_->ToParameterString();
      |                             ^~~~~~~~~~~~~~~~~
/home/noktoborus/source/valhalla/src/odin/maneuver.cc:1404:35: error: ‘class valhalla::baldr::StreetNames’ has no member named ‘ToParameterString’
 1404 |   man_str += begin_street_names_->ToParameterString();
      |                                   ^~~~~~~~~~~~~~~~~
/home/noktoborus/source/valhalla/src/odin/maneuver.cc:1407:35: error: ‘class valhalla::baldr::StreetNames’ has no member named ‘ToParameterString’
 1407 |   man_str += cross_street_names_->ToParameterString();
      |                                   ^~~~~~~~~~~~~~~~~
make[2]: *** [src/odin/CMakeFiles/valhalla-odin.dir/build.make:192: src/odin/CMakeFiles/valhalla-odin.dir/maneuver.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[ 29%] Building CXX object src/baldr/CMakeFiles/valhalla-baldr.dir/directededge.cc.o
[ 29%] Building CXX object src/baldr/CMakeFiles/valhalla-baldr.dir/edgeinfo.cc.o
[...]

PR #3992, opened 2023-02-25
fix build with LOGGING_LEVEL=ALL

noktoborus:
streetnames.h should be included after mjolnir/logging.h, otherwise:

src/baldr/streetnames.cc:67:13: error: no declaration matches ‘std::string valhalla::baldr::StreetNames::ToParameterString() const’
   67 | std::string StreetNames::ToParameterString() const {
...

Additionally, linking errors are fixed:

/usr/bin/ld: src/libvalhalla.a(maneuver.cc.o): in function `valhalla::odin::Maneuver::ToParameterString[abi:cxx11]() const': maneuver.cc:(.text+0x4e8e): undefined reference to `valhalla::odin::Signs::ToParameterString[abi:cxx11]() const' /usr/bin/ld: src/libvalhalla.a(maneuversbuilder.cc.o): in function `valhalla::odin::ManeuversBuilder::InitializeManeuver(valhalla::odin::Maneuver&, int)': maneuversbuilder.cc:(.text+0xb6ad): undefined reference to `valhalla::odin::TransitRouteInfo::ToParameterString[abi:cxx11]() const' ...

# Issue

https://github.com/valhalla/valhalla/issues/3991

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
    * ADDED: `use_lit` costing option for pedestrian costing [#3957](https://github.com/valhalla/valhalla/pull/3957)
    * CHANGED: Removed stray NULL values in log output[#3974](https://github.com/valhalla/valhalla/pull/3974)
    * CHANGED: More conservative estimates for cost of walking slopes [#3982](https://github.com/valhalla/valhalla/pull/3982)
+   * CHANGED: fix build with LOGGING_LEVEL=ALL [#3992](https://github.com/valhalla/valhalla/pull/3992)
 
 ## Release Date: 2023-01-03 Valhalla 3.3.0
 * **Removed**
--- a/src/baldr/streetnames.cc
+++ b/src/baldr/streetnames.cc
@@ -2,6 +2,8 @@
 #include <optional>
 #include <vector>
 
+#include "midgard/logging.h"
+
 #include "baldr/streetname.h"
 #include "baldr/streetnames.h"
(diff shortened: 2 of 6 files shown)


nilsnolde (dismissed):

kevinkreiser:
while this does work I think it's probably more of a mistake that the headers use the preprocessor directive without including the logging header themselves that defines them. like why doesn't streetnames.h include logging.h if it depends on it? even more broadly perhaps we should avoid even having preprocessor stuff in the headers like this. doing so can invite a mismatch to happen in downstream projects that link valhalla. i suppose we can clean it up properly another time. opinions?

kevinkreiser (approved):
i would prefer a larger effort to fix the use of logging directives in headers in the first place (we should hide all these methods in the source anyway if we can, maybe a private header or source that is just free functions to serialize this crap) or alternatively we could stop ifdefing them at all except in the code paths where the library actually calls them and then we s hould add tests for them so we keep the code coverage high. anyway cleanup problems for another day. sorry for the delay on this

noktoborus:
oh, sorry, I missed the notifycation. You are right: current implementation of crap's serialization is not correct because .h files does not include the logging defines and that may adds to the annoyance to use in another projects.

I think better way is:
> we could stop ifdefing them (in headers) ... (keep ifdefing) in the code  ... add tests for them

I think I can do this for a start when I get back to valhalla.

(my english is bad, this is an automatic translation in most :( )

PR #3997, opened 2023-03-01
Fix matrix_locations when used in pbf request

### LINKED ISSUE #3996: matrix_locations overwritten when set in pbf requests
matrix_locations does not default to 0 so it should be one_of in the options.proto file. When setting matrix_locations in the Options pbf in worker.cc - make sure the options was not set in the pbf request (otherwise it will default to max int and try to find all locations).

dnesbitt61:
# Issue

fixes #3996

Needed to use one_of in options.proto and check has_matrix_locations_case() in the else case so we do not overwrite a properly set matrix_locations value in the incoming pbf request.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


nilsnolde (dismissed):
Ah nice, you're the first known (to me) to use the PBF interface!

I'm indifferent to changelog.

kevinkreiser:
if its not too much trouble i think bug fixes are very good candidates for the changelog.

pbf support is a first class citizen, i can tell you that huge numbers of people rely on it daily.

kevinkreiser (approved):

Issue #3955, opened 2023-02-08
Disable hierarchy pruning as costing option


nilsnolde:
As agreed on in #3851 , we should add a costing option to disable hierarchy pruning. It's fairly easy:
- add a `max_distance_disable_hierarchy_culling` (long, but exact) to the JSON config generator under `service_limits` and default to 0
- add a `max_distance_disable_hierarchy_culling` to `thor/worker.h` and assign in `thor/worker.cc` when parsing the config JSON
- add a `disable_hierarchy_pruning` (?) to the costing options
- somewhere we'll need to check that the arc distance(s) between locations are not above the `max_distance_disable_hierarchy_culling` config option. IMO we can add a function to `thor_worker_t` which the actions, which should support this, can call it. I think it doesn't make sense for isochrones, since the algo doesn't use hierarchy limits, and for matrix we can call that new `thor_worker_t` method on all source/target pairs. That new function should set the `disable_hierarchy_pruning` costing option to false. That makes me think.. Maybe we shouldn't do this check at all in thor, but rather in loki, that's where most config & distance checks are done. Then you can safely simply replace thor with loki in the above description everywhere. What do you think @kevinkreiser @dnesbitt61 ?
- parse the `disable_hierarchy_pruning` costing option in `DynamicCost::ParseBaseCostOptions` and set all `max_up_transitions` to `kUnlimitedTransitions` if the value is `true`

Then the routing algorithms don't cull down transitions anymore, which will be kinda disastrous for longer routes (hence the JSON config option, actually pretty much overlapping with the `max_distance` service limit for bike/ped). And when used with `shortest: true` it'd return a true shortest route for drive types. In fact, I just realized that isochrones always returned true shortest, since there isn't any hierarchy limits at play, interesting..

vesperlou:
Hi Nils, this is Junzhen. I think I am capable of solving this issue. Could you please assign this issue to me? Thanks.

kevinkreiser:
@vesperlou thanks for volunteering, let us know if you need any help apart from the excellent description

PR #3998, opened 2023-03-01
Disable hierarchy pruning as costing option 

vesperlou:
# Issue
Disable hierarchy pruning as costing option #3955 

## Tasklist
- add a max_distance_disable_hierarchy_culling to the JSON config generator under service_limits and default to 0
- add a max_distance_disable_hierarchy_culling to loki/worker.h and assign in loki/worker.cc when parsing the config JSON
- add a disable_hierarchy_pruning to the costing options
- add a new function to check the arc distance between locations in loki_worker_t, and set the disable_hierarchy_pruning 
- parse the disable_hierarchy_pruning costing option in DynamicCost::ParseBaseCostOptions and set all max_up_transitions to kUnlimitedTransitions if the value is true

@nilsnolde @kevinkreiser Please review.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -118,7 +118,7 @@ jobs:
         displayName: 'Download Ninja'
 
       - script: |
-          pip install conan
+          pip install "conan<2.0.0"
         displayName: Install conan
 
       - script: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ commands:
     steps:
       - run: brew install protobuf cmake ccache libtool libspatialite pkg-config luajit curl wget czmq lz4 spatialite-tools unzip
       - run: /usr/bin/python3 -m pip install --user requests
-      - run: pip3 install shapely conan
+      - run: pip3 install shapely "conan<2.0.0"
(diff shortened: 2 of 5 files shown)


nilsnolde:
Thanks, looking good on first glance, I'll give it a quick review soon.

I'd also like to know a few things, and don't worry about the answers, I'm just curious:
- did you build valhalla successfully and used its HTTP interfaces a bit?
- do you have a rough idea what you were contributing? (it's kinda of an unfair question tbh, it's pretty complex stuff and I don't expect you do)

2 more things already we'll need before we can merge this:
- we need to write tests. I suggest to look at other PR's that did smth similar (add a JSON config & costing option). I'll let you figure out how to find a PR like that for now, let us know if you need pointers though.
- you'll need to follow https://github.com/valhalla/valhalla/blob/master/CONTRIBUTING.md

nilsnolde:
Just wanted to check in, but some stuff happened apparently :smile: As a tip, it's best to not name your own branch `master` as well when you PR smth, it can be confusing handling the branches locally.

vesperlou:
Ack! Thanks for the tip. Let's discuss in the new PR

PR #4000, opened 2023-03-01
Disable hierarchy pruning as costing option

vesperlou:
# Issue
Disable hierarchy pruning as costing option #3955 

Previous PR: https://github.com/valhalla/valhalla/pull/3998

## Tasklist
- add a max_distance_disable_hierarchy_culling to the JSON config generator under service_limits and default to 0
- add a max_distance_disable_hierarchy_culling to loki/worker.h and assign in loki/worker.cc when parsing the config JSON
- add a disable_hierarchy_pruning to the costing options
- add a new function to check the arc distance between locations in loki_worker_t, and set the disable_hierarchy_pruning
- parse the disable_hierarchy_pruning costing option in DynamicCost::ParseBaseCostOptions and set all max_up_transitions to kUnlimitedTransitions if the value is true

@nilsnolde I accidentally messed up with the previous pr, sorry. 
Re your comment https://github.com/valhalla/valhalla/pull/3998#issuecomment-1450810365:
- Yes, I have successfully built Valhalla locally and played around the HTTP interfaces a bit
- Yes, it took me some time to understand the codebase, but I think I understand what I am contributing :)
- Sure, I'll write some tests tomorrow or later this week
- I followed the contributing guide to format my code


proto/options.proto
```diff
@@ -310,6 +310,9 @@ message Costing {
     uint32 fixed_speed = 80;
     uint32 axle_count = 81;
     float use_lit = 82;
+    oneof has_disable_hierarchy_pruning {
+      bool disable_hierarchy_pruning = 83;
+    }
```
nilsnolde, line 315: doesn't need to be a oneof, since the default will be false

src/loki/isochrone_action.cc
```diff
@@ -8,9 +8,6 @@ using namespace valhalla;
 using namespace valhalla::baldr;
 
 namespace {
-midgard::PointLL to_ll(const valhalla::Location& l) {
-  return midgard::PointLL{l.ll().lng(), l.ll().lat()};
-}
```
nilsnolde, line 13: good call!

src/loki/matrix_action.cc
```diff
@@ -105,6 +103,15 @@ void loki_worker_t::matrix(Api& request) {
...
   const auto& costing_name = Costing_Enum_Name(options.costing_type());
 
+  // Validate distance if disable hiererchy pruning is true
+  auto costing_options = options.mutable_costings()->find(options.costing_type());
+  if (costing_options != options.costings().end() &&
+      costing_options->second.options().disable_hierarchy_pruning() &&
+      !check_hierarchy_distance(options.locations())) {
+    add_warning(request, 205);
+    costing_options->second.mutable_options()->set_disable_hierarchy_pruning(false);
+  }
```
nilsnolde, line 113: at this point in the processing, you already went through the `ParseBaseCostingOptions` so the costing options are set, no need to check that again here. also you can put this whole function into the body of the function you added to loki::worker, right now it's not DRY and you repeat the same thing in all actions that need to respect it.

src/loki/worker.cc
```diff
@@ -262,6 +267,21 @@ void loki_worker_t::set_interrupt(const std::function<void()>* interrupt_functio
...
 
+bool loki_worker_t::check_hierarchy_distance(
+    const google::protobuf::RepeatedPtrField<valhalla::Location>& locations) {
+  // Check every location pair
+  for (auto source = locations.begin(); source != locations.end() - 1; ++source) {
+    for (auto target = source + 1; target != locations.end(); ++target) {
+      // Check if arc distance exceeds max distance limit for disabled hierarchy pruning
+      auto arc_distance = to_ll(*source).Distance(to_ll(*target));
+      if (arc_distance > max_distance_disable_hierarchy_culling) {
+        return false;
```
nilsnolde, line 278: I'd prefer to check the distance against the entire route distance, not only every pair. If someone does 20 locations and each 200 kilometers that'll be crazy computation

nilsnolde: Though not sure how other distance checks handle this, e.g. bike's max_distance_limit, which is pretty much the same as this one.

src/sif/dynamiccost.cc
```diff
@@ -168,6 +168,13 @@ DynamicCost::DynamicCost(const Costing& costing,
   for (auto& edge : costing.options().exclude_edges()) {
     user_exclude_edges_.insert({GraphId(edge.id()), edge.percent_along()});
   }
+
+  // set all max_up_transitions to kUnlimitedTransitions if disable_hierarchy_pruning is true
+  if (costing.options().disable_hierarchy_pruning()) {
+    for (auto& h : hierarchy_limits_) {
+      h.max_up_transitions = kUnlimitedTransitions;
+    }
+  }
```
nilsnolde, line 177: you can consolidate this with the hierarchy limit stuff a few lines up (line 163 currently)

src/worker.cc
```diff
@@ -152,7 +152,8 @@ const std::unordered_map<int, std::string> warning_codes = {
   {201, R"("sources" have date_time set, but "arrive_by" was requested, ignoring date_time)"},
   {202, R"("targets" have date_time set, but "depart_at" was requested, ignoring date_time)"},
   {203, R"("waiting_time" is set on a location of type "via" or "through", ignoring waiting_time)"},
-  {204, R"("exclude_polygons" received invalid input, ignoring exclude_polygons)"}
+  {204, R"("exclude_polygons" received invalid input, ignoring exclude_polygons)"},
+  {205, R"("disable_hierarchy_pruning" received invalid input, max distance exceeded, ignoring disable_hierarchy_pruning)"}
```
nilsnolde, line 156: good that you added a warning (it wasn't part of my description I think), but makes very much sense.

nit:
```suggestion
  {205, R"("disable_hierarchy_pruning" exceeded the max distance , ignoring disable_hierarchy_pruning)"}
```

nilsnolde (changes requested):
This looks good @vesperlou ! There's a few minor things, nothing major. Also, you'll need to add the new config option to the JSON producing script at `./scripts/valhalla_build_config` and give it a sensible default.

proto/options.proto
vesperlou: done!

src/loki/isochrone_action.cc
vesperlou: thanks!

src/loki/matrix_action.cc
vesperlou: Got it. Already put this whole function into loki::worker.

src/loki/worker.cc
vesperlou: Sure. I change to check the total distance of every pair of locations. For matrix computation, check every pair of source and target (As shown in the new commits). Three actions (matrix_action, route_action, and trace_route_action) call this function.

src/sif/dynamiccost.cc
vesperlou: done!

src/worker.cc
vesperlou: Thanks for the revision. done!

vesperlou:
Hi Nils @nilsnolde, I made some changes to address your helpful comments, and to solve the bugs in CI. Btw, I already added the new config option `max_distance_disable_hierarchy_culling` to `./scripts/valhalla_build_config` and set the default to zero. Not sure if you mean something else :) Plan to write tests tomorrow or later this week.
Please review and correct any errors. Thank you!

src/loki/worker.cc
```diff
@@ -262,6 +267,43 @@ void loki_worker_t::set_interrupt(const std::function<void()>* interrupt_functio
   reader->SetInterrupt(interrupt);
 }
 
+// Check if total arc distance exceeds the max distance limit for disable_hierarchy_pruning.
+// If true, add a warning and set the disable_hierarchy_pruning costing option to false.
+void loki_worker_t::check_hierarchy_distance(Api& request, bool is_matrix) {
```
nilsnolde, line 272: I'd rename this to `pair_wise` or so to decouple from the action

nilsnolde:
Jep great, looks good @vesperlou !

> Btw, I already added the new config option max_distance_disable_hierarchy_culling to ./scripts/valhalla_build_config and set the default to zero

Oops, yeah, my bad.

nilsnolde:
So things left to do:
- tests (no pressure, we still have time)
- update the docs, this one is enough: https://github.com/valhalla/valhalla/blob/master/docs/docs/api/turn-by-turn/api-reference.md.

src/sif/dynamiccost.cc
```diff
@@ -392,6 +397,10 @@ void ParseBaseCostOptions(const rapidjson::Value& json,
   // shortest
   JSON_PBF_DEFAULT(co, false, json, "/shortest", shortest);
 
+  // disable hierarchy pruning
+  co->set_disable_hierarchy_pruning(
+      rapidjson::get<bool>(json, "/disable_hierarchy_pruning", co->disable_hierarchy_pruning()));
```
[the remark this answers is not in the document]
vesperlou: Yeah, you are right!

src/loki/worker.cc
vesperlou: Done. Good suggestion.

vesperlou:
Hi @nilsnolde, thanks for the review. I've already updated the docs with reference to the `shortest` costing option. Plan to write tests later this week. 
Don't hesitate to tell me what else I can do.

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -88,7 +88,9 @@ Costing methods can have several options that can be adjusted to develop the rou
...
 
-A special costing option is `shortest`, which, when `true`, will solely use distance as cost and disregard all other costs, penalties and factors. It's available for all costing models except `multimodal` & `bikeshare`.
+A special costing option is `shortest`, which, when `true`, will solely use distance as cost and disregard all other costs, penalties and factors. It's available for all costing models except `multimodal` & `bikeshare`. However, setting `shortest` to `true` doesn't always result in the actual shortest path (although very close to the shortest). That's why `disable_hierarchy_pruning` costing option is introduced.
+
+When `disable_hierarchy_pruning` is `true` and arc distances between source and target are not above the max limit, the actual shortest path will be calculated at the expense of performance. Note that if arc distances between locations exceed the max limit, `disable_hierarchy_pruning` is `true` will not be applied. Likewise, this costing option available for all costing models except `multimodal` & `bikeshare`.
```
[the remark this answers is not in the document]
nilsnolde: Ah haha should've scrolled down before commenting! You already filled the tables:) Perfect description, thanks!

src/worker.cc
```diff
@@ -152,7 +152,8 @@ const std::unordered_map<int, std::string> warning_codes = {
   {201, R"("sources" have date_time set, but "arrive_by" was requested, ignoring date_time)"},
   {202, R"("targets" have date_time set, but "depart_at" was requested, ignoring date_time)"},
   {203, R"("waiting_time" is set on a location of type "via" or "through", ignoring waiting_time)"},
-  {204, R"("exclude_polygons" received invalid input, ignoring exclude_polygons)"}
+  {204, R"("exclude_polygons" received invalid input, ignoring exclude_polygons)"},
+  {205, R"("disable_hierarchy_pruning" exceeded the max distance , ignoring disable_hierarchy_pruning)"}
```
nilsnolde, line 156:
```suggestion
  {205, R"("disable_hierarchy_pruning" exceeded the max distance, ignoring disable_hierarchy_pruning)"}
```

docs/docs/api/turn-by-turn/api-reference.md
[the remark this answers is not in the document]
nilsnolde: small typo

```suggestion
When `disable_hierarchy_pruning` is `true` and arc distances between source and target are not above the max limit, the actual shortest path will be calculated at the expense of performance. Note that if arc distances between locations exceed the max limit, `disable_hierarchy_pruning` is `true` will not be applied. Likewise, this costing option available for all costing models except `multimodal` & `bikeshare`.
```

nilsnolde (commented):
Jep thanks, looks superb so far:) I think we're fully good after the tests.

docs/docs/api/turn-by-turn/api-reference.md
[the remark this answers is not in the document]
vesperlou: Can you point out where is the small typo? couldn't find it...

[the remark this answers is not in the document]
nilsnolde: Sorry, I forgot to actually change smth in the suggestion :sweat_smile:

nilsnolde:
Hm, it's hard to tell without seeing the commands and config you're using, possibly hidden inside your IDE. Generally the tests won't build on just `make`. They'll be configured to be built by CMake, unless you did a `-DENABLE_TESTS=OFF`, but they're not built by default. For that you'll need to build a specific test, e.g. `make -C build -j$(nproc) run-gurka_gtfs`. You can also do a `make check` but I wouldn't recommend that, as that's running ALL tests, which takes ages.

Ideally you use a C++ IDE that's capable of recognizing all targets which can be built, then it's just a click.

vesperlou:
Hi @nilsnolde , thanks for the tips! The command to build one test works :D

I just completed the tests for disable_hierarchy_pruning as a costing option. All worked well on my laptop. Additionally, I built the brand new Valhalla server locally and ran some tests on it,  everything was perfect. TBH disable_hierarchy_pruning is faster than I expected, even for a 400km tricky route in Switzerland the latency is tolerable :) 

The hardest part is to find a route of appropriate scale. Initially I implemented the tests with ascii map like everyone else, but the maps were too small and too simple to show the difference. Then I moved to the real maps and it really took me a while.

nilsnolde:
Hm, indeed a gurka test map will hardly be possible, as the effect of no hierarchies can only be tested if we can also exhaust the up transitions during the expansion (there's 2 conditions, one is maximum distance which is easy to trigger with an ascii map, but not the amount of up transitions..). 

What you tested is not really testing the hierarchies though. It might test `shortest=true` but IMO that has not much to do with the hierarchies directly, or at least it's too opaque right now. What we really want to test is whether the hierarchy limits are disabled when a request comes in with the costing option `disable_hierarchy_pruning`. I think we can do smth like in the astar test: https://github.com/valhalla/valhalla/blob/master/test/astar.cc#L1626-L1645. So we create a custom `mode_costing` from a custom `Options` object for one costing model (e.g. `auto_`) and then check its generated hierarchy limits are unlimited.

Let me know if that makes sense to you.

nilsnolde:
> But regarding the test results, the new route with disable_hierarchy_pruning = true is indeed shorter than the original one, I'd argue this helps to verify our new feature is functioning as expected :)

That's true, I didn't fully look into the logic. We can revert all those existing tests though once we got to test the hierarchies directly, at least IMHO. Also it looks like you're trying to re-build the Utrecht tiles. They're already built by CMake, we'd just need to reference their location (other tests do this too). 

WDYT @kevinkreiser?

vesperlou:
Hi @nilsnolde , I've completed the new tests to check the hierarchires directly, with reference to the astar test :D The old tests are removed from git. Feel free to check and comment anything.

test/gurka/test_disable_hierarchy.cc
```diff
@@ -0,0 +1,83 @@
...
+  co->set_disable_hierarchy_pruning(true);
+}
+
+// Test the hierarchy limits are actually disabled when disable_hierarchy_pruning = true
+void doTest(Costing::Type costing_type) {
+  // Set costing options
+  Options options;
+  set_disable_hierarchy_pruning(options, costing_type);
+  create_costing_options(options, costing_type);
+  auto travel_mode = valhalla::sif::TravelMode::kDrive;
```
nilsnolde, line 29: no need to assign, as `CreateModeCosting` will do that for you
```suggestion
  auto travel_mode;
```

test/gurka/test_disable_hierarchy.cc
```diff
@@ -0,0 +1,83 @@
...
+
+TEST(DisableHierarchyPruningTest, Auto) {
+  Costing::Type type = Costing::auto_;
+  doTest(type);
+}
+
+TEST(DisableHierarchyPruningTest, Bike) {
+  Costing::Type type = Costing::bicycle;
+  doTest(type);
+}
```
nilsnolde, line 48: bike, pedestrian & none you can remove, we're only looking for motorized vehicles as the others already have hierarchies disabled

extra points if we you want to make the others a parameterized test:) (e.g. gurka/test_avoids.cc does that)

nilsnolde (commented):
That looks great, thanks @vesperlou! 

Just a few comments:) No pressure towards parameterized test, it's already very concise now.

nilsnolde:
And also you'll need to add an entry to the changelog. 

FYI, tmrw afternoon we have our PR session and we probably merge right away if the comments are addressed. Actually you can join that session if you want? It's at 13:30 Berlin time, usually just me and @kevinkreiser to review each other's PRs. Then we could get to know each other as well and quickly talk about the proposal etc?

test/gurka/test_disable_hierarchy.cc
vesperlou: Oops I just wanted to have a try, forgot to remove them :D
Parameterized test is finished.

src/loki/worker.cc
```diff
@@ -245,6 +247,9 @@ loki_worker_t::loki_worker_t(const boost::property_tree::ptree& config,
   max_alternates = config.get<unsigned int>("service_limits.max_alternates");
   allow_verbose = config.get<bool>("service_limits.status.allow_verbose", false);
   max_timedep_dist_matrix = config.get<size_t>("service_limits.max_timedep_distance_matrix", 0);
+  // assign max_distance_disable_hierarchy_culling
+  max_distance_disable_hierarchy_culling =
+      config.get<float>("service_limits.max_distance_disable_hierarchy_culling");
```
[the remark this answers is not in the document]
vesperlou: Already changed the default value to 10000.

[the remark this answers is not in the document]
nilsnolde: 0 is good. Even though 10000 wouldn't hurt either, 0 ensures no one is negatively impacted.

vesperlou:
> And also you'll need to add an entry to the changelog.

Done.

> FYI, tmrw afternoon we have our PR session and we probably merge right away if the comments are addressed. Actually you can join that session if you want? It's at 13:30 Berlin time, usually just me and @kevinkreiser to review each other's PRs. Then we could get to know each other as well and quickly talk about the proposal etc?

We are in the same time zone. I'd love to attend to the session, but unfortunately I have a conflict meeting tomorrow. Could we reschedule it to Wednesday? Or later than 2pm is also good to me.

nilsnolde:
> Could we reschedule it to Wednesday? Or later than 2pm is also good to me.

How's Thursday 13:30 for you?

Don't forget to push :)

src/loki/worker.cc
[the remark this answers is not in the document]
vesperlou: Sure. Change back to 0 in the config :)

vesperlou:
> How's Thursday 13:30 for you?

It's perfect! Thanks.

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -113,6 +115,7 @@ These options are available for `auto`, `bus`, and `truck` costing methods.
 | `country_crossing_cost` | A cost applied when encountering an international border. This cost is added to the estimated and elapsed times. The default cost is 600 seconds. |
 | `country_crossing_penalty` | A penalty applied for a country crossing. This penalty can be used to create paths that avoid spanning country boundaries. The default penalty is 0. |
 | `shortest` | Changes the metric to quasi-shortest, i.e. purely distance-based costing. Note, this will disable all other costings & penalties. Also note, `shortest` will not disable hierarchy pruning, leading to potentially sub-optimal routes for some costing models. The default is `false`. |
+| `disable_hierarchy_pruning` | Disable hierarchy pruning to calculate the actual shortest distance. **Note:** This could be quite a performance drainer so there is a upper limit of distance. If the upper limit is exceeded, this option will always be `false`. The default is `false`. |
```
nilsnolde, line 118: also here, not only shortest, but rather actual optimal route. even though @kevinkreiser might not agree, maybe there's other stuff which prevents a real optimal path (note, with hierarchies it's in >95% also optimal, but there's lots of edge cases)

vesperlou: Ack.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
    * CHANGED: More conservative estimates for cost of walking slopes [#3982](https://github.com/valhalla/valhalla/pull/3982)
    * ADDED: An option to Slim down Matrix response [#3987](https://github.com/valhalla/valhalla/pull/3987)
    * CHANGED: Updated url for just_gtfs library [#3994](https://github.com/valhalla/valhalla/pull/3995)
+   * ADDED: `disable_hierarchy_pruning` costing option to find the true shortest route for motor-based costings [#4000](https://github.com/valhalla/valhalla/pull/4000)
```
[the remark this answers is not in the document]
vesperlou: Right, it makes a lot sense.

vesperlou:
Hi @nilsnolde , all changes to documentations are pushed.

src/loki/worker.cc
```diff
@@ -262,6 +267,43 @@ void loki_worker_t::set_interrupt(const std::function<void()>* interrupt_functio
...
+// If true, add a warning and set the disable_hierarchy_pruning costing option to false.
+void loki_worker_t::check_hierarchy_distance(Api& request, bool pair_wise) {
+  auto& options = *request.mutable_options();
+  auto costing_options = options.mutable_costings()->find(options.costing_type());
+  if (!costing_options->second.options().disable_hierarchy_pruning()) {
+    return;
+  }
+
+  bool max_distance_exceeded = false;
+  if (pair_wise) {
```
kevinkreiser, line 280: instead of passing the bool we can simply check `request.Options.action()` and see if its the `sources_to_targets` type

src/loki/worker.cc
kevinkreiser, line 252:
```suggestion
      config.get<float>("service_limits.max_distance_disable_hierarchy_culling", 0);
```

test/gurka/test_disable_hierarchy.cc
```diff
@@ -0,0 +1,54 @@
+#include "gurka.h"
```
kevinkreiser, line 1: this isnt a gurka test, it doesnt create osm data. we have another test for costing options: https://github.com/valhalla/valhalla/blob/master/test/parse_request.cc

i would suggest moving it there or to an other existing test. our build times are already long. the more tiny test files we have to compile the worse it gets (also gets harder to find stuff). long term we should organize the testing more but in the meantime we should avoid making our lives harder

vesperlou: Sure! Should I test the warning in gurka? It needs to create an ascii map for location and distance information, I think.

kevinkreiser: well the location comes from the request, no actual osm data or routing tiles are needed. we have tons of tests in loki_service that test all the errors that come from this sort of "bad request" checking. since this is only a warning, which is not testing in loki service yet, its a little bit more difficult to test because the warning actually doesnt stop the request from trying to route. still though you can do a more traditional unit test style. you can simply instantiate a loki worker and just call worker::route/matrix etc and see how it modified the request object. it either should or shouldnt have the warning added depending on your test case.

vesperlou: Hi Kevin, is there a way to avoid/bypass routing in `loki_worker::route` action? I tried initiating a loki worker and calling `route` action; by adding cheating code in `worker::route` I found the warning is functioning as expected. But as you said, it's just a warning so routing is not stopped. Errors of building tiles keep popping up, and they will cover up the warning.
Maybe the best way to test is to build an ascii map in gurka?

vesperlou: Thanks for the feedback! You are right about the exception. I'm working on it.

src/loki/worker.cc
vesperlou: done

src/loki/worker.cc
```diff
@@ -262,6 +267,43 @@ void loki_worker_t::set_interrupt(const std::function<void()>* interrupt_functio
   reader->SetInterrupt(interrupt);
 }
 
+// Check if total arc distance exceeds the max distance limit for disable_hierarchy_pruning.
+// If true, add a warning and set the disable_hierarchy_pruning costing option to false.
+void loki_worker_t::check_hierarchy_distance(Api& request, bool pair_wise) {
```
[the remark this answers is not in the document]
vesperlou: done

src/loki/worker.cc
vesperlou: done

vesperlou:
Hi @kevinkreiser @nilsnolde , all issues are addressed :) Ready for review now.

test/test.cc
```diff
@@ -309,6 +309,7 @@ boost::property_tree::ptree make_config(const std::string& path_prefix,
         "max_radius": 200,
         "max_reachability": 100,
         "max_timedep_distance": 500000,
+        "max_distance_disable_hierarchy_culling": 0,
```
nilsnolde, line 312: We _could_ remove this kind of change now (since we're also defaulting to `0.f` in the code, not only the config), but might also keep it. I don't have strong feelings around it, just commenting:)

nilsnolde (approved):
LGTM, thanks @vesperlou!

kevinkreiser (approved):

kevinkreiser (approved):

PR #4007, opened 2023-03-03
Workflow action ubuntu

### LINKED ISSUE #4006: Github action for ubuntu build
I propose:
* to have a workflow-dispatch action available for ubuntu 20-04
  * This action can serve as a guide to what needs to be installed based on the flags
  * Can help the GSoC students/mentors to verify the compilation of the code
* Not execute the gh-pages action on forks

cvvergara:
Closes #4006 

## Tasklist

 - [x] Add tests
   - This [gh-pages action](https://github.com/cvvergara/valhalla/actions/runs/4326343992) shows that docs are not been deployed on a branch
   - This [ubuntu-workflow action](https://github.com/cvvergara/valhalla/actions/runs/4326386945) was triggered manually 
 - [x] (Add fixes) Visually ignoring warnings not from Valhala (when using GCC)
   - Boost warnings are ignored
   - Warnings from `third_party/date` are ignored
   - Some warning remain, like for libprotobuf.
   - Dont think this is needed
   - @nilsnolde Please let me know if this is necessary for this PR

### DIFF
--- a/.github/workflows/ubuntu-build.yml
+++ b/.github/workflows/ubuntu-build.yml
@@ -1,0 +1,97 @@
+name: Valhalla Ubuntu Build
+on: [workflow_dispatch]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+  
+jobs:
+  build:
+    name: Ubuntu build
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+        fail-fast: false
+        matrix:
+          os: [ubuntu-20.04]
+          tools: [ON]
+          datatools: [ON]
+          services: [ON]
+          ccache: [ON]
+          tests: [ON]
+          benchmarks: [ON]
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Cache submodules
+        id: cache-third_party
+        uses: actions/cache@v3
+        with:
+          path: third_party
+          key: ${{ runner.os }}-third_party
+
+      - name: Checkout submodules
+        if: steps.cache-third_party.outputs.cache-hit != 'true'
+        shell: bash
+        run: |
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+          git submodule sync --recursive
+          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+
+      - name: Get dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            curl libcurl4 libcurl4-gnutls-dev \
+            libboost-all-dev \
+            libprotobuf-dev \
+            protobuf-compiler
+
+      - name: get datatools dependencies
+        if: ${{ matrix.datatools  == 'ON'}}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            libspatialite-dev \
+            libluajit-5.1-dev \
+            libgeos++-dev
+
+      - name: get benchmarks dependencies when datatools is ON
+        if: ${{ matrix.datatools  == 'ON' && matrix.benchmarks  == 'ON' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            spatialite-bin \
+            libspatialite-dev \
+            libluajit-5.1-dev \
+            libgeos++-dev
+
+      - name: get services dependencies
+        if: ${{ matrix.services  == 'ON' }}
+        run: |
+          sudo add-apt-repository ppa:kevinkreiser/prime-server
+          sudo apt-get update
+          sudo apt-get install libprime-server0 libprime-server-dev prime-server-bin
+
+      - name: Configure
+        run: |
+          mkdir -p build && cd build \
+          && cmake .. \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DENABLE_TOOLS=${{ matrix.tools }} \
+            -DENABLE_DATA_TOOLS=${{ matrix.datatools }} \
+            -DENABLE_SERVICES=${{ matrix.services }} \
+            -DENABLE_CCACHE=${{ matrix.ccache }} \
+            -DENABLE_TESTS=${{ matrix.tests }} \
+            -DENABLE_BENCHMARKS=${{ matrix.benchmarks }}
+
+      - name: get benchmark dependencies
+        if: ${{ matrix.benchmarks  == 'ON' && matrix.datatools  == 'ON'}}
+        run: |
+          mkdir -p build && cd build \
+          && make -j$(nproc) utrecht_tiles
+
+      - name: build
+        run: |
+          cd build \
+          && make -j$(nproc)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,8 @@ if (NOT Boost_FOUND)
 endif()
 find_package(Boost ${boost_VERSION} REQUIRED)
 add_definitions(-DBOOST_NO_CXX11_SCOPED_ENUMS)
+add_definitions(-DBOOST_ALLOW_DEPRECATED_HEADERS)
+add_definitions(-DBOOST_BIND_GLOBAL_PLACEHOLDERS)
 
 if(NOT TARGET CURL::CURL)
   add_library(CURL::CURL INTERFACE IMPORTED)
--- a/src/baldr/tz_alt.cpp
+++ b/src/baldr/tz_alt.cpp
@@ -96,7 +96,9 @@
 #  include <windows.h>
 #endif  // _WIN32
 
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #include "date/tz_private.h"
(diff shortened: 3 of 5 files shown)


kevinkreiser:
the reason we dont use github actions for the build currently is because it takes approximately 100 years to build on the underpowered free machines that github generiously provides. at least last time i did an experiment with it that was the case.

the fix for forks and ghpages makes good sense.

could you maybe tell a little bit more about your motivation here? from the linked issue i understood that it was easier for gsoc people to see how to build the code? for ubuntu we have a single script that essentially installs everything needed to build. its like:

```
scripts/install-linux-deps.sh
# regular cmake stuff
```

cvvergara:
Hi @kevinkreiser 
I've been a GSoC mentor for pgRouting since 2015.

The builds with the tests, it is clear to me that it will take ages to finish. So it is only compiling when all the flags are ON.
We (in pgRouting) have some that also take ages (with the tests), and we execute them only before releases for example
https://github.com/pgRouting/pgrouting/actions/runs/4316224087

This ubuntu build is just for compilation. and is on demand so it not executed on push or PR, its only executed when the committer chooses to (in their own fork) or when you guys choose to do it. So it is not invasive.

Basically gives the opportunity to the student to run it on their own fork if they see fit. (specially with students that use some other compiler that is not GCC) They can have a reference guide on their compilation status before making a PR.

I also focused on ignoring the non valhalla warnings, so that the students don't get confused about if they generated a problem because of their changes.

On the side note,  it also helps as a guide for valhalla ubuntu users of what needs to be installed based on their flags.

cvvergara:
Will fix for osx

kevinkreiser:
yep @cvvergara i'm not doubting your credentials, just trying to understand the intent!

so basically this allows people, in their fork to say hey github, please build this code for me without having a PR in progress in our repo (because obviously if they have a pr in progress it will already build as we have it configured that way). how does one opt into running this workflow? i am not familiar with `workflow_dispatch` in github actions. is it via commit message or via a button somewhere?

also thanks for the warnings clean up. on a side note, we really should have wrapped that date header in our own internal header so the warning suppression could be in one place.

kevinkreiser:
also, i agree that the workflow is nicely laid out for the sake of understanding (it used to be that way before actually) but would it be possible to modify our dependencies script and call that instead? that way we dont have to maintain two copies of what is essentially the same dependencies installation procedure.

cvvergara:
Sure thing

cvvergara:
About workflow dispatch, it's button on the action
image

third_party/About the date issue, yeah, there might be cleaner ways. (using cmake)
This is just a fast "fixup" because I need to understand your cmake a bit more to make the changes, the fixup does not affect the build the cmake, its work with delicate fingers

kevinkreiser:
yep yep, im not suggesting you fix the date header stuff, im merely expressing my embarrassment at the state of it ;)

cvvergara:
@kevinkreiser 
I am preparing a PR without the new workflow, only for the ignore warnings and do not depoly on forks change.
Do this kind of task need an update to the changelog?

kevinkreiser:
nah no changelog needed!

nilsnolde:
I wasn't sure what to think in the beginning, but after looking at your workflow config for Github Actions, I think it's a really nice move, also for contributors, not only GSoC, it's much faster than the CircleCI builds and it can run even before they raise a PR. I use the `dispatch_workflow` option on a lot of repos, I like it. I'm not the biggest fan of build-per-push, some people are just very "push-y".

People will have to be aware of it though, so should we add a section in https://github.com/valhalla/valhalla/blob/master/CONTRIBUTING.md#code-contributions?

Bit verbose with the warning stuff, but CMake solution would probably be less legible, if less verbose.. Hopefully this year still, with C++20 we can remove it again..

.github/workflows/ubuntu-build.yml
```diff
@@ -0,0 +1,97 @@
...
+            -DENABLE_SERVICES=${{ matrix.services }} \
+            -DENABLE_CCACHE=${{ matrix.ccache }} \
+            -DENABLE_TESTS=${{ matrix.tests }} \
+            -DENABLE_BENCHMARKS=${{ matrix.benchmarks }}
+
+      - name: get benchmark dependencies
+        if: ${{ matrix.benchmarks  == 'ON' && matrix.datatools  == 'ON'}}
+        run: |
+          mkdir -p build && cd build \
+          && make -j$(nproc) utrecht_tiles
```
nilsnolde, line 92: now that you build the utrecht_tiles target, you could even build `valhalla_service` on top, run it in the background with e.g. `./build/valhalla_service test/bindings/python/valhalla.json 1 &` and give it a quick curl for utrecht coordinates, then with very low effort/time at least > 50% of the code is executed (graph build and query). what do you think?

.github/workflows/ubuntu-build.yml
```diff
@@ -0,0 +1,97 @@
...
+            -DCMAKE_BUILD_TYPE=Release \
+            -DENABLE_TOOLS=${{ matrix.tools }} \
+            -DENABLE_DATA_TOOLS=${{ matrix.datatools }} \
+            -DENABLE_SERVICES=${{ matrix.services }} \
+            -DENABLE_CCACHE=${{ matrix.ccache }} \
+            -DENABLE_TESTS=${{ matrix.tests }} \
+            -DENABLE_BENCHMARKS=${{ matrix.benchmarks }}
+
+      - name: get benchmark dependencies
+        if: ${{ matrix.benchmarks  == 'ON' && matrix.datatools  == 'ON'}}
```
nilsnolde, line 89: I think we can ignore the benchmarks, they're also not executed in the current state.

nilsnolde (commented):
I like having the new workflow. I'm also rather leaning towards calling the installation script. The extra value for people knowing what to install isn't worth the extra headspace to keep maintaining a second place with the same commands. In worst case people install a few too many programs/libs for their use case.

kevinkreiser:
> it's much faster than the CircleCI builds

where did you see that? it looked like it took 10 minutes to get the deps and 5 minutes to do the build. the basic `build` step takes about 5 minutes on circle ci so it seems its reached the level of being equivalent (when actions came out i tested it then and it was absymally slow, though i completely understand, its free afterall!).

i feel like im still not fully understanding the final goal here. is this completely suplimental to what we have, ie. just to serve the use-case of a person who wants to check their code without a pr etc in their fork?

or do we want to do some larger changes like replacing linux builds with this? is it possible to disallow prs to be merged with out some checks but also to make to those checks manual? so like we require the linux build to pass but we have to actually click a button before it builds. is that a thing?

kevinkreiser:
> it looked like it took 10 minutes

ah it was actually more confusing than that. the 10 minutes was to build `valhalla_build_tiles` and then also build utrecht. so yeah its more like actions takes about 15 minutes to do what circle does in 5, which is obviously quite a bit slower (again though not complaining for something free!).

nilsnolde:
Well, the build completes in 5 mins yes, but the action still runs for > 20 mins.

For me, it's about
1. not wasting CI hours of any kind. If we also cache `ccache` it'll be a lot faster @cvvergara 
2. inclusivity. I doubt everyone who'd ever like to develop Valhalla has access to a machine with many cores. If a build takes 15 mins for them and they can't use their machine while doing so, it's probably pretty discouraging. This provides the means to offload that to a server. For me, it doesn't really matter where it's run.

kevinkreiser:
> but circleci still runs for > 20 mins.

thats because it runs all the tests, this new one doesnt yet have that enabled. apples to apples circle is like 3x faster, which is important. anyway i totally get that the differentiator is that actions lets people who arent us, run builds on github infra so the slowdown is worth it to have something. no need to sell me on it, i get the value!

> inclusivity

leaving the buzzwords aside, i think maybe my questions earlier weren't clear. i fully agree that people who fork should have the option to build the code remotely to check their work against our approved build procedure. thats great. im just asking is that all we are doing or did we have bigger plans regarding using actions or building on CI less automatically. from the comments here i was getting a vibe about too many wasted CI hours and so i just wanted to ask the question is some of this work done with a view toward solving that problem as well. and if that is the case the build time becomes an issue to some degree imho.

haha sorry my questions are all confused, im just trying to understand if there is a bigger picture here somethign more than just allowing forks to build. if so i just wanted to open the discussion up so i could understand what that was.  if not and its all just allowing forks to build, thats great.

nilsnolde:
Right, I wouldn't advocate for switching to GA completely, as you say, it's way too slow with the free version. I think it shouldn't run the tests either, building the tiles and running the service with a few curl is ok-ish if the idea is to properly compile & checking that there are no disastrous runtime errors. For me, it's really only about forks, not about thinking about replacing circleci.

nilsnolde:
From my side this is quite ready to merge @cvvergara. What do you think about adding:
- a cache step for `./.ccache` in the action?
- this one: https://github.com/valhalla/valhalla/pull/4007#discussion_r1125503467

Can also do it in a follow-up PR.

kevinkreiser:
I was still hoping we could do the dependencies installation via the existing script that's the only thing I'd ask for here before merging

PR #4011, opened 2023-03-07
baldr: Adjusting code to C++17

cvvergara:
# Issue

`baldr` part of #4018

# TODO

* [x] removing the superfluous try catch from the new edgeinfo function
* [x] correcting the docstrings for the edgeinfo functions
* [x] making a sentinel invalid function that is properly typed to replace uses of -1 (probably goes in midgard/util headers or somewhere everything can use it)
* [x] fixing the system include directive to work in the valhalla_module so we can keep date stuff in baldr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
    * CHANGED: Updated url for just_gtfs library [#3994](https://github.com/valhalla/valhalla/pull/3995)
    * ADDED: Docker image pushes to Github's docker registry [#4033](https://github.com/valhalla/valhalla/pull/4033)
    * ADDED: `disable_hierarchy_pruning` costing option to find the actual optimal route for motorized costing modes, i.e `auto`, `motorcycle`, `motor_scooter`, `bus`, `truck` & `taxi`. [#4000](https://github.com/valhalla/valhalla/pull/4000)
+   * CHANGED: baldr directory: remove warnings and C++17 adjustments [#4011](https://github.com/valhalla/valhalla/pull/4011)
 
 ## Release Date: 2023-01-03 Valhalla 3.3.0
 * **Removed**
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,7 @@ option(ENABLE_STATIC_LIBRARY_MODULES "If ON builds Valhalla modules as STATIC li
 set(LOGGING_LEVEL "" CACHE STRING "Logging level, default is INFO")
 set_property(CACHE LOGGING_LEVEL PROPERTY STRINGS "NONE;ALL;ERROR;WARN;INFO;DEBUG;TRACE")
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+include_directories(SYSTEM ${VALHALLA_SOURCE_DIR}/third_party/date/include)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
@@ -334,7 +335,7 @@ endif()
 foreach(script valhalla_build_config valhalla_build_elevation
   valhalla_build_extract valhalla_build_timezones)
   configure_file(${VALHALLA_SOURCE_DIR}/scripts/${script} ${CMAKE_BINARY_DIR}/${script} COPYONLY)
-  
+
   install(
     FILES
       scripts/${script}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,6 @@ option(ENABLE_STATIC_LIBRARY_MODULES "If ON builds Valhalla modules as STATIC li
 set(LOGGING_LEVEL "" CACHE STRING "Logging level, default is INFO")
 set_property(CACHE LOGGING_LEVEL PROPERTY STRINGS "NONE;ALL;ERROR;WARN;INFO;DEBUG;TRACE")
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
-include_directories(SYSTEM ${VALHALLA_SOURCE_DIR}/third_party/date/include)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
@@ -335,7 +334,7 @@ endif()
 foreach(script valhalla_build_config valhalla_build_elevation
   valhalla_build_extract valhalla_build_timezones)
   configure_file(${VALHALLA_SOURCE_DIR}/scripts/${script} ${CMAKE_BINARY_DIR}/${script} COPYONLY)
-
+  
   install(
     FILES
       scripts/${script}
--- a/src/baldr/CMakeLists.txt
+++ b/src/baldr/CMakeLists.txt
@@ -44,6 +44,9 @@ set(sources
     directededge.cc
     edgeinfo.cc
     graphid.cc
+    graphreader.cc
+    graphtile.cc
+    graphtileheader.cc
     incident_singleton.h
     edgetracker.cc
     nodeinfo.cc
@@ -63,17 +66,13 @@ set(sources
     transitroute.cc
     transitschedule.cc
     transittransfer.cc
+    tz_alt.cpp
     laneconnectivity.cc
     verbal_text_formatter.cc
     verbal_text_formatter_us.cc
     verbal_text_formatter_us_co.cc
     verbal_text_formatter_us_tx.cc
     verbal_text_formatter_factory.cc)
-
-set (sources_with_warnings
-    graphreader.cc
-    graphtile.cc
-    graphtileheader.cc)
 
 list(APPEND sources
     #basic timezone stuff
@@ -91,8 +90,6 @@ list(APPEND sources
     ${CMAKE_CURRENT_BINARY_DIR}/date_time_systemv.h
     ${CMAKE_CURRENT_BINARY_DIR}/date_time_windows_zones.h)
 
-list(APPEND sources_with_warnings
-    tz_alt.cpp)
 
 #ios we have more work to make use of system tzdb
 if(APPLE)
--- a/src/baldr/CMakeLists.txt
+++ b/src/baldr/CMakeLists.txt
@@ -33,6 +33,12 @@ set(includes
      ${CMAKE_CURRENT_BINARY_DIR}/src/baldr
      )
 
+# treat date library as system
+set(system_includes ${VALHALLA_SOURCE_DIR}/third_party/date/include)
+if(APPLE)
+  list(APPEND system_includes ${VALHALLA_SOURCE_DIR}/third_party/date/include/date)
+endif()
+
 set(sources
     accessrestriction.cc
     admin.cc
@@ -90,10 +96,11 @@ list(APPEND sources
     ${CMAKE_CURRENT_BINARY_DIR}/date_time_systemv.h
     ${CMAKE_CURRENT_BINARY_DIR}/date_time_windows_zones.h)
 
+list(APPEND sources_with_warnings
+    tz_alt.cpp)
 
 #ios we have more work to make use of system tzdb
 if(APPLE)
-  list(APPEND includes ${VALHALLA_SOURCE_DIR}/third_party/date/include/date)
   list(APPEND sources ${VALHALLA_SOURCE_DIR}/third_party/date/src/ios.mm)
 endif()
 
@@ -110,6 +117,9 @@ valhalla_module(NAME baldr
       ${includes}
     PRIVATE
       ${CMAKE_CURRENT_BINARY_DIR}
+  SYSTEM_INCLUDE_DIRECTORIES
+    PUBLIC
+      ${system_includes}
 
   DEPENDS
     valhalla::midgard
--- a/src/baldr/CMakeLists.txt
+++ b/src/baldr/CMakeLists.txt
@@ -30,7 +30,6 @@ set(includes
      ${VALHALLA_SOURCE_DIR}/valhalla
      $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/third_party/dirent/include>
      ${VALHALLA_SOURCE_DIR}/third_party/rapidjson/include
-     ${VALHALLA_SOURCE_DIR}/third_party/date/include
      ${CMAKE_CURRENT_BINARY_DIR}/src/baldr
      )
 
--- a/src/baldr/connectivity_map.cc
+++ b/src/baldr/connectivity_map.cc
@@ -76,7 +76,7 @@ json::MapPtr to_geometry(const polygon_t& polygon) {
 #else
       if (outer) {
         ring_coords->emplace_back(
-          json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
+            json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
       } else {
         ring_coords->emplace_front(
             json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
--- a/src/baldr/connectivity_map.cc
+++ b/src/baldr/connectivity_map.cc
@@ -70,16 +70,23 @@ json::MapPtr to_geometry(const polygon_t& polygon) {
   for (const auto& ring : polygon) {
     auto ring_coords = json::array({});
     for (const auto& coord : ring) {
-      // if (outer) {
+#if 1
       ring_coords->emplace_back(
           json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
-      /*} else {
+#else
+      if (outer) {
+        ring_coords->emplace_back(
+          json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
+      } else {
         ring_coords->emplace_front(
             json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
-      }*/
+      }
+#endif
     }
     coords->emplace_back(ring_coords);
-    // outer = false;
+#if 0
+    outer = false;
+#endif
   }
   return json::map({{"type", std::string("Polygon")}, {"coordinates", coords}});
 }
@@ -189,7 +196,7 @@ std::unordered_set<size_t> connectivity_map_t::get_colors(uint32_t hierarchy_lev
       AABB2<PointLL> bbox(Point2(ll.lng() - lngdeg, ll.lat() - latdeg),
                           Point2(ll.lng() + lngdeg, ll.lat() + latdeg));
       std::vector<int32_t> tilelist = tiles.TileList(bbox);
-      for (auto& id : tilelist) {
+      for (const auto& id : tilelist) {
         auto color = level->second.find(id);
         if (color != level->second.cend()) {
           result.emplace(color->second);
@@ -248,8 +255,7 @@ std::vector<size_t> connectivity_map_t::to_image(const uint32_t hierarchy_level)
   }
   const auto& level_tiles = TileHierarchy::levels()[tile_level];
 
-  std::vector<size_t> tiles(level_tiles.tiles.nrows() * level_tiles.tiles.ncolumns(),
-                            static_cast<uint32_t>(0));
+  std::vector<size_t> tiles(level_tiles.tiles.nrows() * level_tiles.tiles.ncolumns(), 0);
   auto level = colors.find(hierarchy_level);
   if (level != colors.cend()) {
     for (size_t i = 0; i < tiles.size(); ++i) {
--- a/src/baldr/connectivity_map.cc
+++ b/src/baldr/connectivity_map.cc
@@ -70,23 +70,10 @@ json::MapPtr to_geometry(const polygon_t& polygon) {
   for (const auto& ring : polygon) {
     auto ring_coords = json::array({});
     for (const auto& coord : ring) {
-#if 1
       ring_coords->emplace_back(
           json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
-#else
-      if (outer) {
-        ring_coords->emplace_back(
-            json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
-      } else {
-        ring_coords->emplace_front(
-            json::array({json::fixed_t{coord.first, 6}, json::fixed_t{coord.second, 6}}));
-      }
-#endif
     }
     coords->emplace_back(ring_coords);
-#if 0
-    outer = false;
-#endif
   }
   return json::map({{"type", std::string("Polygon")}, {"coordinates", coords}});
 }
--- a/src/baldr/datetime.cc
+++ b/src/baldr/datetime.cc
@@ -51,8 +51,9 @@ namespace DateTime {
 tz_db_t::tz_db_t() : db(date::get_tzdb()) {
   // NOTE: outside of this class 0 is reserved for invalid timezone
   // so we offset each index by 1 to get into the valid range 1-300 or so
-  for (size_t i = 0; i < db.zones.size(); ++i) {
-    names.emplace(db.zones[i].name(), i + 1);
+  size_t idx{0};
+  for (const auto& zone : db.zones) {
+    names.emplace(zone.name(), ++idx);
   }
 }
 
--- a/src/baldr/edgeinfo.cc
+++ b/src/baldr/edgeinfo.cc
@@ -93,7 +93,39 @@ std::vector<std::string> EdgeInfo::GetNames() const {
       throw std::runtime_error("GetNames: offset exceeds size of text list");
     }
   }
+  names.shrink_to_fit();
   return names;
+}
+
+std::vector<std::pair<std::string, bool>> EdgeInfo::GetNames(bool include_tagged_values) const {
+  // Get each name
+  std::vector<std::pair<std::string, bool>> name_type_pairs;
+  name_type_pairs.reserve(name_count());
+  const NameInfo* ni = name_info_list_;
+  for (uint32_t i = 0; i < name_count(); i++, ni++) {
+    // Skip any tagged names (FUTURE code may make use of them)
+    if (ni->tagged_ && !include_tagged_values) {
+      continue;
+    }
+    if (ni->tagged_) {
+      if (ni->name_offset_ < names_list_length_) {
+        std::string name = names_list_ + ni->name_offset_;
+        try {
+          if (IsNameTag(name[0])) {
+            name_type_pairs.push_back({name.substr(1), false});
+          }
+        } catch (const std::invalid_argument& arg) {
+          LOG_DEBUG("invalid_argument thrown for name: " + name);
+        }
+      } else
+        throw std::runtime_error("GetNamesAndTypes: offset exceeds size of text list");
+    } else if (ni->name_offset_ < names_list_length_) {
+      name_type_pairs.push_back({names_list_ + ni->name_offset_, ni->is_route_num_});
+    } else {
+      throw std::runtime_error("GetNamesAndTypes: offset exceeds size of text list");
+    }
+  }
+  return name_type_pairs;
 }
 
 // Get a list of tagged names
@@ -134,15 +166,16 @@ std::vector<std::string> EdgeInfo::GetTaggedValues(bool only_pronunciations) con
       throw std::runtime_error("GetTaggedNames: offset exceeds size of text list");
     }
   }
+  names.shrink_to_fit();
   return names;
 }
 
 // Get a list of names
-std::vector<std::pair<std::string, bool>>
-EdgeInfo::GetNamesAndTypes(std::vector<uint8_t>& types, bool include_tagged_values) const {
+std::vector<std::tuple<std::string, bool, uint8_t>>
+EdgeInfo::GetNamesAndTypes(bool include_tagged_values) const {
 
   // Get each name
-  std::vector<std::pair<std::string, bool>> name_type_pairs;
+  std::vector<std::tuple<std::string, bool, uint8_t>> name_type_pairs;
   name_type_pairs.reserve(name_count());
   const NameInfo* ni = name_info_list_;
   for (uint32_t i = 0; i < name_count(); i++, ni++) {
@@ -155,8 +188,7 @@ EdgeInfo::GetNamesAndTypes(std::vector<uint8_t>& types, bool include_tagged_valu
         std::string name = names_list_ + ni->name_offset_;
         try {
           if (IsNameTag(name[0])) {
-            name_type_pairs.push_back({name.substr(1), false});
-            types.push_back(static_cast<uint8_t>(name.at(0)));
+            name_type_pairs.push_back({name.substr(1), false, static_cast<uint8_t>(name.at(0))});
           }
         } catch (const std::invalid_argument& arg) {
           LOG_DEBUG("invalid_argument thrown for name: " + name);
@@ -164,12 +196,12 @@ EdgeInfo::GetNamesAndTypes(std::vector<uint8_t>& types, bool include_tagged_valu
       } else
         throw std::runtime_error("GetNamesAndTypes: offset exceeds size of text list");
     } else if (ni->name_offset_ < names_list_length_) {
-      name_type_pairs.push_back({names_list_ + ni->name_offset_, ni->is_route_num_});
-      types.push_back(0);
+      name_type_pairs.push_back({names_list_ + ni->name_offset_, ni->is_route_num_, 0});
     } else {
       throw std::runtime_error("GetNamesAndTypes: offset exceeds size of text list");
     }
   }
+  name_type_pairs.shrink_to_fit();
   return name_type_pairs;
 }
 
--- a/src/baldr/edgeinfo.cc
+++ b/src/baldr/edgeinfo.cc
@@ -165,7 +165,7 @@ std::vector<std::string> EdgeInfo::GetTaggedValues(bool only_pronunciations) con
   return names;
 }
 
-// Get a list of names with types
+// Get a list of names
 std::vector<std::tuple<std::string, bool, uint8_t>>
 EdgeInfo::GetNamesAndTypes(bool include_tagged_values) const {
 
--- a/src/baldr/edgeinfo.cc
+++ b/src/baldr/edgeinfo.cc
@@ -93,10 +93,10 @@ std::vector<std::string> EdgeInfo::GetNames() const {
       throw std::runtime_error("GetNames: offset exceeds size of text list");
     }
   }
-  names.shrink_to_fit();
   return names;
 }
 
+// Get a list of names tagged and/or untagged with tag status
 std::vector<std::pair<std::string, bool>> EdgeInfo::GetNames(bool include_tagged_values) const {
   // Get each name
   std::vector<std::pair<std::string, bool>> name_type_pairs;
@@ -118,11 +118,11 @@ std::vector<std::pair<std::string, bool>> EdgeInfo::GetNames(bool include_tagged
           LOG_DEBUG("invalid_argument thrown for name: " + name);
         }
       } else
-        throw std::runtime_error("GetNamesAndTypes: offset exceeds size of text list");
+        throw std::runtime_error("GetNames: offset exceeds size of text list");
     } else if (ni->name_offset_ < names_list_length_) {
       name_type_pairs.push_back({names_list_ + ni->name_offset_, ni->is_route_num_});
     } else {
-      throw std::runtime_error("GetNamesAndTypes: offset exceeds size of text list");
+      throw std::runtime_error("GetNames: offset exceeds size of text list");
     }
   }
   return name_type_pairs;
@@ -166,11 +166,10 @@ std::vector<std::string> EdgeInfo::GetTaggedValues(bool only_pronunciations) con
       throw std::runtime_error("GetTaggedNames: offset exceeds size of text list");
     }
   }
-  names.shrink_to_fit();
   return names;
 }
 
-// Get a list of names
+// Get a list of names with types
 std::vector<std::tuple<std::string, bool, uint8_t>>
 EdgeInfo::GetNamesAndTypes(bool include_tagged_values) const {
 
@@ -201,7 +200,6 @@ EdgeInfo::GetNamesAndTypes(bool include_tagged_values) const {
       throw std::runtime_error("GetNamesAndTypes: offset exceeds size of text list");
     }
   }
-  name_type_pairs.shrink_to_fit();
   return name_type_pairs;
 }
 
--- a/src/baldr/edgeinfo.cc
+++ b/src/baldr/edgeinfo.cc
@@ -110,12 +110,8 @@ std::vector<std::pair<std::string, bool>> EdgeInfo::GetNames(bool include_tagged
     if (ni->tagged_) {
       if (ni->name_offset_ < names_list_length_) {
         std::string name = names_list_ + ni->name_offset_;
-        try {
-          if (IsNameTag(name[0])) {
-            name_type_pairs.push_back({name.substr(1), false});
-          }
-        } catch (const std::invalid_argument& arg) {
-          LOG_DEBUG("invalid_argument thrown for name: " + name);
+        if (IsNameTag(name[0])) {
+          name_type_pairs.push_back({name.substr(1), false});
         }
       } else
         throw std::runtime_error("GetNames: offset exceeds size of text list");
@@ -185,12 +181,8 @@ EdgeInfo::GetNamesAndTypes(bool include_tagged_values) const {
     if (ni->tagged_) {
       if (ni->name_offset_ < names_list_length_) {
         std::string name = names_list_ + ni->name_offset_;
-        try {
-          if (IsNameTag(name[0])) {
-            name_type_pairs.push_back({name.substr(1), false, static_cast<uint8_t>(name.at(0))});
-          }
-        } catch (const std::invalid_argument& arg) {
-          LOG_DEBUG("invalid_argument thrown for name: " + name);
+        if (IsNameTag(name[0])) {
+          name_type_pairs.push_back({name.substr(1), false, static_cast<uint8_t>(name.at(0))});
         }
       } else
         throw std::runtime_error("GetNamesAndTypes: offset exceeds size of text list");
--- a/src/baldr/graphreader.cc
+++ b/src/baldr/graphreader.cc
@@ -183,6 +183,11 @@ void FlatTileCache::Reserve(size_t tile_size) {
 
 // Checks if tile exists in the cache.
 bool FlatTileCache::Contains(const GraphId& graphid) const {
+  // TODO there is something wrong with this check:
+  // get_index returns an uint32_t which is never negative
+  // therefore it allways different from -1
+  // warning: comparison of integers of different signs: 'uint32_t' (aka 'unsigned int') and 'int'
+  // [-Wsign-compare]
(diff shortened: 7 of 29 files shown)


src/baldr/connectivity_map.cc
```diff
@@ -70,16 +70,23 @@ json::MapPtr to_geometry(const polygon_t& polygon) {
   for (const auto& ring : polygon) {
     auto ring_coords = json::array({});
     for (const auto& coord : ring) {
-      // if (outer) {
+#if 1
```
kevinkreiser, line 73: you left the declaration of `outer` commented out above, that said, im cool if we just delete this unused code. im sure i just had it in there for testing something and im not sure why i didnt remove it when i was done.. im sorry you're finding all the cobwebs :smile:

src/baldr/datetime.cc
```diff
@@ -51,8 +51,9 @@ namespace DateTime {
 tz_db_t::tz_db_t() : db(date::get_tzdb()) {
   // NOTE: outside of this class 0 is reserved for invalid timezone
   // so we offset each index by 1 to get into the valid range 1-300 or so
-  for (size_t i = 0; i < db.zones.size(); ++i) {
-    names.emplace(db.zones[i].name(), i + 1);
+  size_t idx{0};
```
kevinkreiser, line 54: im not sure how i feel about this. sometimes i do this too but most times i question it. i know we like the `foreach` because it reads easier but we need the index incremented anyway. then my mind says if you are in-the-know enough to understand the difference between pre and post increment operators you can probably handle an index based for loop. anyway i could go either way on it, but imho its not a necessary change

valhalla/baldr/graphreader.h
```diff
@@ -157,6 +157,7 @@ class FlatTileCache : public TileCache {
   }
   inline uint32_t get_index(const GraphId& graphid) const {
     auto offset = get_offset(graphid);
+    // TODO a negative value can not be stored on an uint32_t, anything can happen
```
kevinkreiser, line 160: we do this a lot in the code making use of automatic conversion to set max value by wrapping around with the negative value on an unsigned type. we often use the max value as a sentinel value on unsigned types to signify invalid

cvvergara:
In this commit:
https://github.com/valhalla/valhalla/pull/4011/commits/9440737386795ab8526d9f1d84292fe5cd828ace
all files on `sources_with_warnings` are moved to `sources`
No warning on `baldr` should appear on the CI

CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ option(ENABLE_STATIC_LIBRARY_MODULES "If ON builds Valhalla modules as STATIC li
 set(LOGGING_LEVEL "" CACHE STRING "Logging level, default is INFO")
 set_property(CACHE LOGGING_LEVEL PROPERTY STRINGS "NONE;ALL;ERROR;WARN;INFO;DEBUG;TRACE")
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+include_directories(SYSTEM ${VALHALLA_SOURCE_DIR}/third_party/date/include)
```
kevinkreiser, line 43: why move this out to the top level? design-wise we were really hoping to get all date functionality wrapped by baldr (in helper functions mostly).

a little background on this, a long time ago (https://github.com/valhalla/valhalla/pull/1512/files#r215660149) we started using the date library but it has, on different platforms, different ways of getting at timezone information. for us, the timezones have to be in a stable order (because we use indices to reference them from the data that we build). this is why tz_alt exists. i copied tz.cpp from the date library and modified it to use static data. because this code isnt owned by us (nor should it be really) i dont think we need to bother about fixing warnings. in fact i'd really have prefered that we make as little changes to that source as possible so it was clear what we did there vs what comes standard in that thirdparty library.

src/baldr/tz_alt.cpp
```diff
@@ -439,7 +429,10 @@ get_tz_dir()
 static_assert(min_year <= max_year, "Configuration error");
 #endif
 
+#if 0
 static std::unique_ptr<tzdb> init_tzdb();
+#endif
```
kevinkreiser, line 434: yeah this is the crux of the change i made here when we took a copy of this and patched it. im open to better options here but none really occured to me at the time

src/odin/enhancedtrippath.cc
```diff
@@ -27,6 +27,7 @@ constexpr int kIsStraightestBuffer = 10;                   // Buffer between str
 constexpr uint32_t kBackwardTurnDegreeLowerBound = 124;
 constexpr uint32_t kBackwardTurnDegreeUpperBound = 236;
 
+#ifdef LOGGING_LEVEL_TRACE
```
kevinkreiser, line 30: we have lots of stuff like this in `odin` as well and i hate it. logging directives seeping out into header files and everything which is terrible. the bulk of this stuff is used for debug printing when analysing routes

src/baldr/graphtileheader.cc
```diff
@@ -12,16 +13,23 @@ namespace valhalla {
...
-  version_[kMaxVersionSize - 1] = 0;
-  tile_size_ = 0;
+GraphTileHeader::GraphTileHeader()
+    : // initialization of bitfields done here in c++20 can be done in the class definition
+      graphid_(0), density_(0), name_quality_(0), speed_quality_(0), exit_quality_(0),
+      has_elevation_(0), has_ext_directededge_(0), nodecount_(0), directededgecount_(0),
+      predictedspeeds_count_(0), spare1_(0), transitioncount_(0), spare3_(0), turnlane_count_(0),
+      spare4_(0), transfercount_(0), spare2_(0), departurecount_(0), stopcount_(0), spare5_(0),
+      routecount_(0), schedulecount_(0), signcount_(0), spare6_(0), access_restriction_count_(0),
+      admincount_(0), spare7_(0) {
```
kevinkreiser, line 23: why isnt `tile_size_` in here?

src/baldr/edgeinfo.cc
```diff
@@ -93,9 +93,41 @@ std::vector<std::string> EdgeInfo::GetNames() const {
       throw std::runtime_error("GetNames: offset exceeds size of text list");
     }
   }
+  names.shrink_to_fit();
```
kevinkreiser, line 96: this has performance implications, we should revert it as that type of thing isnt in the scope of this pr anyway

valhalla/baldr/graphreader.h
cvvergara: Ok, so I changed to `std::numeric_limits<uint32_t>::max();`

CMakeLists.txt
cvvergara: Treating the third party header as system header, the compiler wont generate warnings from that header.
By having it at top level, will help me out on fixing the other modules

src/baldr/CMakeLists.txt
```diff
@@ -92,8 +90,6 @@ list(APPEND sources
     ${CMAKE_CURRENT_BINARY_DIR}/date_time_systemv.h
     ${CMAKE_CURRENT_BINARY_DIR}/date_time_windows_zones.h)
 
-list(APPEND sources_with_warnings
-    tz_alt.cpp)
```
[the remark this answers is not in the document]
cvvergara: Yep, but its not generating warnings now :-)

src/baldr/tz_alt.cpp
cvvergara: Keeping this unused code as a reminder of the "looking for better options"

src/baldr/graphtileheader.cc
cvvergara: Everything that was not a bitfield is initialized on the header.
bitfield can be initialized on the header starting from std 20

src/odin/enhancedtrippath.cc
cvvergara: For the moment less intrusive to mark when to compile the loggings with the directive.
I will keep in mind when I see the other directories to think about that strategy

cvvergara:
@kevinkreiser I think this is ready.

kevinkreiser:
sounds good! ill take another look hopefully this evening, sorry again for the delay and thanks again for grinding th rough it!

CMakeLists.txt
kevinkreiser: i think we should patch the valhalla module function to work with the SYSTEM header directive then we can keep this inclusion inside of baldr with the rest of the date time stuff

src/baldr/datetime.cc
kevinkreiser: sorry i didn't explain myself very well. my point was that range loops are syntax sugar for readability when you dont need the index. i am perfectly aware of the pre-increment operator im just saying that it is another detail to know about which is just as mentally demanding as reading an index-based for loop. i agree that in this case the final result might be faster because there wont be any calls to `operator[]` (though maybe the compiler can figure out how to do this properly with pointer arithmetic, ie the iterator version which would be equivalent).

anyway its a fine change, i was just trying to explain my hesitation to a change that seems unrelated to warnings and modernization. regarding the latter, index based for loops arent deprecated or anything so we dont need to eg. remove all of them from the code base

src/baldr/edgeinfo.cc
```diff
@@ -93,9 +93,41 @@ std::vector<std::string> EdgeInfo::GetNames() const {
...
+    }
+    if (ni->tagged_) {
+      if (ni->name_offset_ < names_list_length_) {
+        std::string name = names_list_ + ni->name_offset_;
+        try {
+          if (IsNameTag(name[0])) {
+            name_type_pairs.push_back({name.substr(1), false});
+          }
+        } catch (const std::invalid_argument& arg) {
+          LOG_DEBUG("invalid_argument thrown for name: " + name);
```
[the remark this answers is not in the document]
kevinkreiser: i dont actually understand what this is meant to `catch`. the only think i can think of here is the indices you are using with `name` but neither of those throw `invalid_argument`. also we have a strong gaurantee that a name entry will have a length of 1 or more and if it is a tagged entry a length of 2 or more. we need not to add try catch for these items with strong gaurantees in favor of upstream code testing on the gaurantees themselves

valhalla/baldr/edgeinfo.h
```diff
@@ -165,8 +172,8 @@ class EdgeInfo {
    *
    * @return   Returns a list (vector) of name/route number pairs.
    */
-  std::vector<std::pair<std::string, bool>> GetNamesAndTypes(std::vector<uint8_t>& types,
-                                                             bool include_tagged_names = false) const;
+  std::vector<std::tuple<std::string, bool, uint8_t>>
+  GetNamesAndTypes(bool include_tagged_names = false) const;
```
kevinkreiser, line 176: the change in behavior should be reflected in the docstring above

valhalla/baldr/admin.h
```diff
@@ -62,11 +64,11 @@ class Admin {
...
 protected:
-  uint32_t country_offset_;         // country name offset
-  uint32_t state_offset_;           // state name offset
-  char country_iso_[kCountryIso]{}; // country ISO3166-1
-  char state_iso_[kStateIso]{};     // state ISO3166-2
-  char spare_[3]{};                 // spare for byte alignment
+  uint32_t country_offset_;                     // country name offset
+  uint32_t state_offset_;                       // state name offset
+  std::array<char, kCountryIso> country_iso_{}; // country ISO3166-1
+  std::array<char, kStateIso> state_iso_{};     // state ISO3166-2
```
kevinkreiser, line 70: this is another one i dont quite like the feel of. arrays are definitely nice but this is a binary structure that we write to disk and read in from disk. `array` will totally work here but changing it from a c-looking thing to an stl-looking thing gives me pause.

@dnesbitt61 how do you feel about it?

src/baldr/graphtile.cc
```diff
@@ -449,12 +449,14 @@ std::string GraphTile::FileSuffix(const GraphId& graphid,
                                    : TileHierarchy::levels()[graphid.level()]);
 
   // figure out how many digits in tile-id
-  const auto max_id = level.tiles.ncolumns() * level.tiles.nrows() - 1;
+  const uint32_t max_id = level.tiles.ncolumns() > 0 && level.tiles.nrows() > 0
+                              ? level.tiles.ncolumns() * level.tiles.nrows() - 1
+                              : 0;
```
kevinkreiser, line 454: this is not needed we do not allow for negative numbers in the dimensions of the tiles datastructure. in fact i forget why we even use signed integers in that class. at any rate, to avoid the warning use a `static_cast`

valhalla/baldr/admin.h
cvvergara: This is not a binary structure.
This C way of writing arrays:
```
char country_iso_[kCountryIso];
```
The size is kCountryIso of contiguous memory of type char.
The C++ way of writing arrays:
```
 std::array<char, kCountryIso> country_iso_{};
 ```
 The size is kCountryIso of contiguous memory of type char.
 with the benefits of having all std::array member function.
That is why the tests like [this](https://github.com/valhalla/valhalla/blob/master/test/admin.cc#L17) pass, because the data did not change size.

File: src/baldr/graphtile.cc (hunk 13 above)
Comment (cvvergara) [reply src/baldr/graphtile.cc:454]: Its not negative number on `ncolumns` or `nrows` the issue here,
its what happens when any of those is 0, the max_id gets a -1, but its a uint32_t.

I will change to `static_cast`.

File: src/baldr/graphtile.cc [hunk 14]
```diff
@@ -558,8 +559,10 @@ GraphId GraphTile::GetTileId(const std::string& fname) {
                                : TileHierarchy::levels()[level];
 
   // get the number of sub directories that we should have
-  auto max_id = tile_level.tiles.ncolumns() * tile_level.tiles.nrows() - 1;
-  size_t parts = static_cast<size_t>(std::log10(std::max(1, max_id))) + 1;
+  uint32_t max_id = tile_level.tiles.ncolumns() > 0 && tile_level.tiles.nrows() > 0
+                        ? tile_level.tiles.ncolumns() * tile_level.tiles.nrows() - 1
+                        : 0;
```
kevinkreiser, line 564: same casting here for same reason

src/baldr/graphtileheader.cc
```diff
@@ -12,16 +13,23 @@ namespace valhalla {
...
+  set_version(PACKAGE_VERSION);
 }
 
 // Set the version string.
 void GraphTileHeader::set_version(const std::string& version) {
-  strncpy(version_, version.c_str(), kMaxVersionSize);
+  // reinitializing the version array before copying
+  version_ = {};
+  std::copy(version.begin(), version.begin() + std::min(kMaxVersionSize, sizeof(version.c_str())),
```
kevinkreiser, line 31: the previous code was a bug because someone could screw with the version string and make it shorter than `kMaxVersionSize` and `strncpy` would have read past the end, so its good you are fixing it however i think you have a new bug here, this should be:

```suggestion
  std::copy(version.begin(), version.begin() + std::min(kMaxVersionSize, version.size()),
```

otherwise it will only ever copy 8 bytes because `c_str()` is a pointer which is 64bits wide (these days)

CMakeLists.txt
cvvergara: Added a task on the #4018 will work once things compile without warnings

src/baldr/edgeinfo.cc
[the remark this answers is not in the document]
cvvergara: was because the CI was complaining that the catch code wasn't been tested

src/baldr/graphtileheader.cc
cvvergara: nice catch :-)

cvvergara:
@kevinkreiser made the changes

valhalla/baldr/admin.h
kevinkreiser: sorry i think you misunderstand me. it is a structure, which we serialize as a binary object straight to disk/mem and back again by moving bytes around and casting: https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/graphtile.h#L834

i get that the tests still pass, my comment isnt about correctness, this is why i said:

> array will totally work here

im trying to say that when you switch to using stl containers it no longer looks like a POD and that makes me hesitate to use it. serializing and deserializing PODs is trivial you literally just copy/move bytes around and cast their address to the type you want. with non PODs you have to write actual logic.

in this case, after some research i can be sure that `array` (as expected) is a POD. however, by using old style arrays i dont even have to think about it, or worse look it up, to figure out if it is a POD or not.

its important that we use PODs which guarantee standard layout for data compatibility across versions and platforms

src/thor/triplegbuilder.cc
```diff
@@ -821,23 +821,33 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
...
         // Skip the tagged names
         name_index++;
         continue;
       }
+#else
+      if (std::get<2>(name_and_type) != 0) {
+        // Skip the tagged names
+        continue;
+      }
+#endif
```
kevinkreiser, line 845: im a little confused about these changes, im fine with the edge info name getting refactor to return a tuple instead of pass by reference. we can just delete the old way, or was there a good reason to keep it around?

valhalla/baldr/edgeinfo.h
```diff
@@ -147,10 +148,19 @@ class EdgeInfo {
...
+   * @return   Returns a list (vector) of untagged names.
    */
   std::vector<std::string> GetNames() const;
 
+  /** get the name & is_tagged flag
+   *
+   *  This one does not calculate the types but has the is_tagged flag
+   *  Like GetNamesAndTypes but without using memory for the types
+   *
+   *  @return iReturns a list (vector) (name, is_tagged) pairs
```
kevinkreiser, line 160: are the docstrings correct? to me the bools that come back in both cases are to indicate whether or not the name is a route number, not just if its generically tagged. and then the thing that it returns as the third piece in the tuple below isnt the route number itself but rather the tag type, ie `TaggedValue`

valhalla/baldr/graphreader.h
kevinkreiser: i feel like that hurts readibilty compared to `-1`, maybe we should make something like

```c++
template <typename numeric_t> invalid() { return std::numeric_limits<numeric_t>::max(); }
```

then in calling code at least it would read nicely:

```c++
return offset < cache_indices_.size() ? cache_indices_[offset] : invalid<uint32_t>();
```

kevinkreiser:
ok i made it to the bottom of the diff! the only changes i would request are:

1. removing the superfluous `try catch` from the new edgeinfo function
2. correcting the docstrings for the edgeinfo functions
3. making a sentinel `invalid` function that is properly typed to replace uses of `-1` (probably goes in midgard/util headers or somewhere everything can use it)
4. fixing the system include directive to work in the `valhalla_module` so we can keep date stuff in baldr

kevinkreiser:
regarding number 4, i tried it out myself and it just worked for me, this was my diff with master:

```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4632fdbcf..0889a357a 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -70,7 +70,7 @@ endfunction()
 
 function(valhalla_module)
   set(oneValueArgs NAME)
-  set(multiValueArgs SOURCES SOURCES_WITH_WARNINGS HEADERS DEPENDS INCLUDE_DIRECTORIES)
+  set(multiValueArgs SOURCES SOURCES_WITH_WARNINGS HEADERS DEPENDS INCLUDE_DIRECTORIES SYSTEM_INCLUDE_DIRECTORIES)
   cmake_parse_arguments(MODULE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
   if (UNIX AND ENABLE_SINGLE_FILES_WERROR)
diff --git a/src/baldr/CMakeLists.txt b/src/baldr/CMakeLists.txt
index 1ce31ea6d..306646f26 100644
--- a/src/baldr/CMakeLists.txt
+++ b/src/baldr/CMakeLists.txt
@@ -30,7 +30,6 @@ set(includes
      ${VALHALLA_SOURCE_DIR}/valhalla
      $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/third_party/dirent/include>
      ${VALHALLA_SOURCE_DIR}/third_party/rapidjson/include
-     ${VALHALLA_SOURCE_DIR}/third_party/date/include
      ${CMAKE_CURRENT_BINARY_DIR}/src/baldr
      )
 
@@ -114,6 +113,8 @@ valhalla_module(NAME baldr
       ${includes}
     PRIVATE
       ${CMAKE_CURRENT_BINARY_DIR}
+  SYSTEM
+      ${VALHALLA_SOURCE_DIR}/third_party/date/include
 
   DEPENDS
     valhalla::midgard
```

now in reality it should be slightly more complicated because of the ios includes i guess but only very slightly more complicated. also, did you know that clion now lets you debug cmake scripts!? its awesome you can litterally step through a cmake run in the debugger and watch varaibles and general flow through the script, pretty awesome!

valhalla/baldr/graphreader.h
cvvergara: Added this change to the list of tasks, on #4018 
When encountering with this again, need to figure out the best place to put the template.
Will do that on a later PR

valhalla/baldr/admin.h
cvvergara: I agree I dont get you, but I did my research as follows:
```
#include <iostream>
#include <array>
#include <type_traits>

constexpr size_t kCountryIso = 2;

class Admin {                                                                                                           
public: 
std::array<char, kCountryIso> country_iso_{};
};


int main()
{
    std::cout << std::boolalpha;
    std::cout << std::is_pod<std::array<char, kCountryIso>>::value << '\n';
    std::cout << std::is_pod<Admin>::value << '\n';
}
```
give as result:
```
true
false
```
Conclusion array is POD, but admin is not.

src/thor/triplegbuilder.cc
cvvergara: right, forgot to delete the old code,
(its in thor, and i unadvertedly added it to this PR)

valhalla/baldr/edgeinfo.h
cvvergara: I dont know, if the docstrings are correct,
To me this does not make sense,
The vector has name on the std::string and route number on the bool?
```
* @return   Returns a list (vector) of name/route number pairs.
*/
 std::vector<std::pair<std::string, bool>> GetNamesAndTypes(std::vector<uint8_t>& types,
                                                             bool include_tagged_names = false) const;
```

valhalla/baldr/graphreader.h
cvvergara: ah, I just saw the comment of where to place it

cvvergara:
> regarding number 4, i tried it out myself and it just worked for me, this was my diff with master:
> 
> ```diff
> diff --git a/src/baldr/CMakeLists.txt b/src/baldr/CMakeLists.txt
> index 1ce31ea6d..306646f26 100644
> --- a/src/baldr/CMakeLists.txt
> +++ b/src/baldr/CMakeLists.txt
> @@ -30,7 +30,6 @@ set(includes
>       ${VALHALLA_SOURCE_DIR}/valhalla
>       $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/third_party/dirent/include>
>       ${VALHALLA_SOURCE_DIR}/third_party/rapidjson/include
> -     ${VALHALLA_SOURCE_DIR}/third_party/date/include
>       ${CMAKE_CURRENT_BINARY_DIR}/src/baldr
>       )
>  
> @@ -114,6 +113,8 @@ valhalla_module(NAME baldr
>        ${includes}
>      PRIVATE
>        ${CMAKE_CURRENT_BINARY_DIR}
> +  SYSTEM
> +      ${VALHALLA_SOURCE_DIR}/third_party/date/include
>  
>    DEPENDS
>      valhalla::midgard
> ```
> 
> now in reality it should be slightly more complicated because of the ios includes i guess but only very slightly more complicated. also, did you know that clion now lets you debug cmake scripts!? its awesome you can litterally step through a cmake run in the debugger and watch varaibles and general flow through the script, pretty awesome!

Doing that is leading to this in my computer (doing `make VERBOSE=ON`):
```
-I/home/vicky/valhalla/cvvergara/src/baldr/SYSTEM -I/home/vicky/valhalla/cvvergara/third_party/date/include
```
instead of:
```
-isystem /home/vicky/valhalla/cvvergara/third_party/date/include
```

src/baldr/edgeinfo.cc
```diff
@@ -96,6 +96,38 @@ std::vector<std::string> EdgeInfo::GetNames() const {
...
+  const NameInfo* ni = name_info_list_;
+  for (uint32_t i = 0; i < name_count(); i++, ni++) {
+    // Skip any tagged names (FUTURE code may make use of them)
+    if (ni->tagged_ && !include_tagged_values) {
+      continue;
+    }
+    if (ni->tagged_) {
+      if (ni->name_offset_ < names_list_length_) {
+        std::string name = names_list_ + ni->name_offset_;
+        try {
```
cvvergara, line 113: @kevinkreiser 
Just making sure that this is the try & catch you want me to remove, even that is similar to the one in line 189 bellow, which then leads me to ask why do you have it in the first place on GetNamesAndTypes.
This new function is like GetNamesAndTypes without filling up the types. (hence the new name I used)

valhalla/baldr/edgeinfo.h
cvvergara: For me, tags come from osm: like: highway=motorway, boundary=yes  etc.
So I am very very confused:

Here is about the function I based the code on: (The new function I made was based on this one)
https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/edgeinfo.h#L162

From the docstrings:

> Returns a list (vector) of name/route number pairs.

which is std::vector<std::pair<std::string, bool>>

Therefore the bool value is a route number.

Which is not, it is just an indicator that:
for the ith value on the returned vector when the bool value is true then the value on the ith element on the `types` vector is valid.

But then you mention:

> so the boolean  to me the bools that come back in both cases are to indicate whether or not the name is a route number, 

what do you mean?
like [route 66](https://en.wikipedia.org/wiki/U.S._Route_66)? 
Then the name is '66' and that name will be returned only when the function is called with `true`

Maybe you mean:
so the boolean  to me the bools that come back in both cases are to indicate whether or not the value in `type` is a route number

Definitively the function that uses this one does not use the `types` (aka route number)

--------

In the new function I made, there is no `types` (wich I think is your route number)
So I am just keeping the indicator that the name has a `type`.

```
@return Returns a list (vector) (name, is_tagged) pairs 
```
Maybe you want
```
@return Returns a list (vector) (name, has_tag) pairs 
```

cvvergara:
> ok i made it to the bottom of the diff! the only changes i would request are:
> 
>     1. removing the superfluous `try catch` from the new edgeinfo function
> 

The new function I based on [this one:](https://github.com/valhalla/valhalla/blob/master/src/baldr/edgeinfo.cc#L140)
Please confirm that even that the GetNamesAndTypes have the try/catch, it is not needed on the new one.

>     2. correcting the docstrings for the edgeinfo functions
> 
I am confused, see [comment](https://github.com/valhalla/valhalla/pull/4011#discussion_r1141466715)

>     3. making a sentinel `invalid` function that is properly typed to replace uses of `-1` (probably goes in midgard/util headers or somewhere everything can use it)
> 
Working on it

>     4. fixing the system include directive to work in the `valhalla_module` so we can keep date stuff in baldr
Maybe I missed something, its not working for me: [comment](https://github.com/valhalla/valhalla/pull/4011#issuecomment-1475381863)

kevinkreiser:
ah regarding the cmake suggestion, sorry, i completely forgot to check the `VERBOSE` and just assumed it would fail to compile it didnt work (forgot that warnings werent treated as error on master). but great news with just a tiny little bit more work it properly functions. the main trick is that you have to call `target_include_directories` twice! once for regular stuff (`-I`) and again for anything you want to mark as system (`-isystem`):

```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4632fdbcf..94646600e 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -70,7 +70,7 @@ endfunction()
 
 function(valhalla_module)
   set(oneValueArgs NAME)
-  set(multiValueArgs SOURCES SOURCES_WITH_WARNINGS HEADERS DEPENDS INCLUDE_DIRECTORIES)
+  set(multiValueArgs SOURCES SOURCES_WITH_WARNINGS HEADERS DEPENDS INCLUDE_DIRECTORIES SYSTEM_INCLUDE_DIRECTORIES)
   cmake_parse_arguments(MODULE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
   if (UNIX AND ENABLE_SINGLE_FILES_WERROR)
@@ -107,6 +107,9 @@ function(valhalla_module)
      PRIVATE
        ${VALHALLA_BINARY_DIR}
        ${VALHALLA_BINARY_DIR}/valhalla)
+  if (MODULE_SYSTEM_INCLUDE_DIRECTORIES)
+    target_include_directories(${library} SYSTEM ${MODULE_SYSTEM_INCLUDE_DIRECTORIES})
+  endif()
   
   target_link_libraries(${library} Boost::boost)
 
diff --git a/src/baldr/CMakeLists.txt b/src/baldr/CMakeLists.txt
index 1ce31ea6d..948b81429 100644
--- a/src/baldr/CMakeLists.txt
+++ b/src/baldr/CMakeLists.txt
@@ -30,10 +30,15 @@ set(includes
      ${VALHALLA_SOURCE_DIR}/valhalla
      $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/third_party/dirent/include>
      ${VALHALLA_SOURCE_DIR}/third_party/rapidjson/include
-     ${VALHALLA_SOURCE_DIR}/third_party/date/include
      ${CMAKE_CURRENT_BINARY_DIR}/src/baldr
      )
 
+# treat date library as system
+set(system_includes ${VALHALLA_SOURCE_DIR}/third_party/date/include)
+if(APPLE)
+  list(APPEND system_includes ${VALHALLA_SOURCE_DIR}/third_party/date/include/date)
+endif()
+
 set(sources
     accessrestriction.cc
     admin.cc
@@ -97,7 +102,6 @@ list(APPEND sources_with_warnings
 
 #ios we have more work to make use of system tzdb
 if(APPLE)
-  list(APPEND includes ${VALHALLA_SOURCE_DIR}/third_party/date/include/date)
   list(APPEND sources ${VALHALLA_SOURCE_DIR}/third_party/date/src/ios.mm)
 endif()
 
@@ -109,11 +113,16 @@ valhalla_module(NAME baldr
 
   HEADERS
     ${headers}
+
   INCLUDE_DIRECTORIES
     PUBLIC
       ${includes}
     PRIVATE
-      ${CMAKE_CURRENT_BINARY_DIR}
+        ${CMAKE_CURRENT_BINARY_DIR}
+
+  SYSTEM_INCLUDE_DIRECTORIES
+    PUBLIC
+      ${VALHALLA_SOURCE_DIR}/third_party/date/include
 
   DEPENDS
     valhalla::midgard

```

if i just added `SYSTEM` to the begininng of `INCLUDE_DIRECTORIES` then everything woudl be marked as system and when i added it to the end of `INCLUDE_DIRECTORIES` (like i first suggested) then it was still marked as a project include. anyway having two separate calls makes it work!

anyway the ouptput of one of the targets shows:

```
cd /home/kkreiser/sandbox/valhalla/cmake-build-debug/src/baldr && ccache /usr/bin/c++ -DAUTO_DOWNLOAD=0 -DBOOST_ALLOW_DEPRECATED_HEADERS -DBOOST_ALL_NO_LIB -DBOOST_BIND_GLOBAL_PLACEHOLDERS -DBOOST_NO_CXX11_SCOPED_ENUMS -DCURL_STATICLIB -DDATA_TOOLS -DHAS_REMOTE_API=0 -DLOGGING_LEVEL_TRACE -I/home/kkreiser/sandbox/valhalla -I/home/kkreiser/sandbox/valhalla/valhalla -I/home/kkreiser/sandbox/valhalla/third_party/rapidjson/include -I/home/kkreiser/sandbox/valhalla/cmake-build-debug/src/baldr/src/baldr -I/home/kkreiser/sandbox/valhalla/cmake-build-debug/src/baldr -I/home/kkreiser/sandbox/valhalla/cmake-build-debug/src -I/home/kkreiser/sandbox/valhalla/cmake-build-debug/src/valhalla -isystem /home/kkreiser/sandbox/valhalla/third_party/date/include -fno-omit-frame-pointer -fdiagnostics-color=auto -fsanitize=address -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=undefined -fno-sanitize=vptr -g -fPIC -fdiagnostics-color=always -mfpmath=sse -msse2 -std=gnu++17 -Werror -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -MD -MT src/baldr/CMakeFiles/valhalla-baldr.dir/verbal_text_formatter.cc.o -MF CMakeFiles/valhalla-baldr.dir/verbal_text_formatter.cc.o.d -o CMakeFiles/valhalla-baldr.dir/verbal_text_formatter.cc.o -c /home/kkreiser/sandbox/valhalla/src/baldr/verbal_text_formatter.cc
```

i also added the small detail about ios builds

src/baldr/graphreader.cc
```diff
@@ -183,7 +183,8 @@ void FlatTileCache::Reserve(size_t tile_size) {
 
 // Checks if tile exists in the cache.
 bool FlatTileCache::Contains(const GraphId& graphid) const {
-  return get_index(graphid) != -1;
+  // max value means invalid
+  return get_index(graphid) != std::numeric_limits<uint32_t>::max();
```
kevinkreiser, line 187: use `invalid`

src/baldr/graphreader.cc
```diff
@@ -206,7 +207,8 @@ void FlatTileCache::Clear() {
 // Get a pointer to a graph tile object given a GraphId.
 graph_tile_ptr FlatTileCache::Get(const GraphId& graphid) const {
   auto index = get_index(graphid);
-  if (index == -1)
+  // max value means invalid
+  if (index == std::numeric_limits<uint32_t>::max())
```
kevinkreiser, line 211: same as above

valhalla/baldr/edgeinfo.h
cvvergara: @kevinkreiser  Need feedback here

src/baldr/edgeinfo.cc
cvvergara: @kevinkreiser
Need feed back here

kevinkreiser: yes this is the one im talking about. i dont know why its there at all! if i had to guess id say it had a purpose, the stuff inside the `try` got refactored and then it became vestigial. we should just remove both this one and the other

valhalla/baldr/admin.h
kevinkreiser: sorry, i think we are still talking past each other. i agree that `std::array` is a POD, i said as much above:

> that array (as expected) is a POD

what im really trying to say is that adding stl containers to a struct makes it "look" or "seem" like it's not a POD, even though it still functions as one in the ways that are important to us. also, i know that admin is technically not a POD already but thats only because it has its own constructor which we could easily ditch for a static make_admin method.

by adding an stl container to it though, you make it even less of a POD (because a POD cannot be composed of anything but primitive types). but the main reason i dont like the change here is that it reads as if its not a POD type anymore. meaning when people look at it it doesnt appear to be trivial. thats what i meant when i said:

>  it no longer looks like a POD

anyway id really appreciate it if we could keep this as close to a POD as possible, id even be for refactoring the constructor to be a `static Admin make_admin(...)` method but definitely a different pr for that

valhalla/baldr/edgeinfo.h
kevinkreiser: ok so we store a bunch of variable sized stuff in `edgeinfo` that makes it into the text list. some of those things are like what you'd expect, such as `main street` but some of the records are also marked with a boolean flag, roads like `I-81` or as you mentioned `US-66`. for the latter 2 that flag tells us that its not just a regular road name, its a shielded road. shielded roads get special treatment in the forming of directions and narrative (the textual instructions). if you have a look over at the comment in edgeinfo.h it tells you basically what im trying to explain here: https://github.com/valhalla/valhalla/blob/15dedaac1b26730d78bc1f4ae988ad88f4c95264/valhalla/baldr/edgeinfo.h#L37-L38

so yeah the boolean in both cases is `is_route_num` ie whether or not the name is the route relations name.

now we also have this concept of "tagged" values. this is not like osm exactly. we have space for an enumeration (1 byte worth), and then can attach with that 1 byte a bunch of data, whatever it may be in the same text record where we store normal textual stuff.

the code looks at the single bit (is_tagged) on the nameinfo object and if that is the case it says ok the first byte of the bytes in that record is the value of the enum for the tag. in this case the enum is here: https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/graphconstants.h#L378

so its not completely free form like OSM but rather we pick and choose which things we want ot support, give them a number in the enum, that becomes the first byte of the text record and then depending on which enum it was we would have different run time parsing of the bytes that followed.

this is a convenient way for us to randomly add non performance critical metadata to the dataset. especially data that is sparse (meaning it doesnt occur on all edge in the graph)

hopefully that explains what you are seeing here in edgeinfo :smile:

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
    * ADDED: An option to Slim down Matrix response [#3987](https://github.com/valhalla/valhalla/pull/3987)
    * CHANGED: Updated url for just_gtfs library [#3994](https://github.com/valhalla/valhalla/pull/3995)
    * ADDED: Docker image pushes to Github's docker registry [#4033](https://github.com/valhalla/valhalla/pull/4033)
+   * CHANGED: baldr directory: remove warnings and C++17 adjustments [#4011](https://github.com/valhalla/valhalla/pull/4011)
```
kevinkreiser, line 46: we need to move this down one slot

valhalla/baldr/admin.h
kevinkreiser: can you please revert the `std::array` change?

valhalla/baldr/edgeinfo.h
```diff
@@ -163,10 +173,10 @@ class EdgeInfo {
    * Convenience method to get the names and route number flags for an edge.
    * @param  include_tagged_values  Bool indicating whether or not to return the tagged values too
    *
-   * @return   Returns a list (vector) of name/route number pairs.
+   * @return   Returns a list (vector) of (name, is_tagged, route number) tuple.
```
kevinkreiser, line 176:
```suggestion
   * @return   Returns a vector tuples where the elements of the tuple are name/data, is_route_num flag and the TaggedValue enum if the entry was a tagged value
```

valhalla/baldr/edgeinfo.h
kevinkreiser, line 160:
```suggestion
   *  @return Returns a vector of pairs where the elements of the pair are name/data and the is_route_num flag respectively
```

valhalla/baldr/edgeinfo.h
```diff
@@ -147,10 +148,19 @@ class EdgeInfo {
...
    * Convenience method to get the names for an edge
-   * @return   Returns a list (vector) of names.
+   * @return   Returns a list (vector) of untagged names.
    */
   std::vector<std::string> GetNames() const;
 
+  /** get the name & is_tagged flag
+   *
+   *  This one does not calculate the types but has the is_tagged flag
+   *  Like GetNamesAndTypes but without using memory for the types
```
kevinkreiser, line 158:
```suggestion
  /** Get the name/data and the is_route_num flag
   *
   *  This variation does not return the tagged value type for those
   *  entries which are tagged and saves a bit of memory that way
```

valhalla/baldr/edgeinfo.h
```diff
@@ -163,10 +173,10 @@ class EdgeInfo {
    * Convenience method to get the names and route number flags for an edge.
    * @param  include_tagged_values  Bool indicating whether or not to return the tagged values too
```
kevinkreiser, line 174:
```suggestion
   * Convenience method to get the names, route number flags and tag value type for an edge.
   * @param  include_tagged_values  Bool indicating whether or not to return the tagged values entries too
```

valhalla/baldr/edgeinfo.h
```diff
@@ -147,10 +148,19 @@ class EdgeInfo {
...
   std::vector<std::string> GetNames() const;
 
+  /** get the name & is_tagged flag
+   *
+   *  This one does not calculate the types but has the is_tagged flag
+   *  Like GetNamesAndTypes but without using memory for the types
+   *
+   *  @return iReturns a list (vector) (name, is_tagged) pairs
+   */
+  std::vector<std::pair<std::string, bool>> GetNames(bool) const;
```
kevinkreiser, line 162:
```suggestion
  std::vector<std::pair<std::string, bool>> GetNames(bool include_tagged_names = false) const;
```

valhalla/baldr/graphtileheader.h
```diff
@@ -601,13 +603,13 @@ class GraphTileHeader {
 
   // TODO: in v4, don't store this its superfluous information, the graphid has all we need
   // Base lon, lat of the tile
-  std::pair<float, float> base_ll_;
+  std::pair<float, float> base_ll_ = {0, 0};
 
   // baldr version.
-  char version_[kMaxVersionSize];
+  std::array<char, kMaxVersionSize> version_ = {};
```
kevinkreiser, line 609: similar to admin id prefer not to put `std::array` in here (base_ll_ is also marked for deletion)

CHANGELOG.md
cvvergara: I am doing a rebase on master and placing at the end

valhalla/baldr/admin.h
cvvergara: Q. verifying, do you want only the ones I converted from C arrays to std::array back to C arrays, or also the std::arrays that are already there? like:
https://github.com/valhalla/valhalla/search?q=std%3A%3Aarray

kevinkreiser: i think i've still not made myself clear. im not opposed to using `std::array` in the code base, `std::array` is totally fine to use. i am opposed to using `std::array` or any other non-primitive, in structs that we serialize to disk. these structs rely on/exploit the fact that they will have the same memory layout across platforms.

but now that i think about it i might have an outdated understanding of how to guarantee that memory layout! i had thought that one needed to use PODs to do it but it turns out the struct needs only to fulfill the property of having a "standard layout." in other words `std::is_standard_layout<T>` must be `true` for the given type.

having `std::array` here and in other places where we serialize these things to disk definitely makes them no longer `std::is_trivial<T>` (one part of being a POD) but i think it doesn't stop them from being `std::is_standard_layout` (the other part of being POD). which means in the end that im wrong about needing to remove these, you can feel free to use `std::array` here and in the `graphtileheader`

whew!

cvvergara: Only removed the ones I added

cvvergara: Added them back

cvvergara:
> ok i left a fresh round of review comments after going over the whole thing again. all of them are trivial so i suspect we can merge this soon?

Yes

src/baldr/edgeinfo.cc
cvvergara: done

kevinkreiser (approved):

kevinkreiser (approved):

kevinkreiser:
we missed the one update to the docstring but commited it and since its only a comment i will just merge

PR #4019, opened 2023-03-10
Improvements and Bugfixes to GraphReader::GetShortcut

### LINKED ISSUE #3888: GraphReader::GetShortcut returns non-shortcut edge
As briefly discussed in #3887 we have noticed that `GraphReader::GetShortcut` doesn't always return the correct result. I've prepared simple unit tests to showcase the issue: [graph_reader_get_shortcut_bug](https://github.com/PhilippVoigt/valhalla/blob/graph_reader_get_shortcut_bug/test/recover_shortcut.cc).

At first glance it seems to be an index off-by-one kind of error, the returned edge always has ID one higher than the expected shortcut. I will keep debugging myself to deepen my understanding of the data structures, but the root cause is probably immediately obvious to some of your more experienced eyes 😉

PhilippVoigt:
# Issue

This resolves #3888 and fixes 3 issues within `GetShortcut` to make it more reliable:
1. **Align conditions for valid shortcut member edges with shortcut builder**
There was a slight mismatch between the basic requirements an edge must fulfill to become part of a shortcut. During shortcut builder, edges with `bss_connection` or the use `kConstruction` were excluded. These conditions were not checked during shortcut resolution, leading to more ambiguity and thereby to false negatives. The conditions were extracted to a member function of `DirectedEdge` to stay in sync.
2. **Resolve shortcut ID correctly in case of > 2 shortcuts at one node**
Once the correct superseded edge is found inside `GetShortcut`, it's superseded mask is converted to an edge index at the corresponding node to determine the shortcut's ID. The previous implementation worked for nodes with <= 2 shortcuts but lead to wrong GraphIds if a node has > 2 shortcuts. This lead to off-by-one errors in certain rare cases.
3. **Prevent traversal into unrelated shortcut for edges that are not superseded**
As suggested here 3f73e05d41090911a5ab999ef79ee1fe58ee5df5 by @kevinkreiser an edge can only really be a member of a shortcut, if the first node at which the traversal encounters any shortcut is also the node at which the shortcut we're looking for starts. I.e. if we encounter a node that has shortcuts but do not reach that node on the opposing edge of a superseded edge, we did not start the traversal within a shortcut. This validation prevents false positives as visualized in #3888.

Unit tests were added to check both false negatives (cases in which a valid shortcut was not found) and false positives (cases in which a regular edge or an unrelated shortcut were found). The tests were catered solely to our own use cases, which might differ from yours.

Licensed under MIT


valhalla/baldr/directededge.h
```diff
@@ -564,6 +564,17 @@ class DirectedEdge {
     return static_cast<Use>(use_);
   }
 
+  /**
+   * Evaluates a basic set of conditions to determine if this directed edge is a valid potential
+   * member of a shortcut. This is used while forming and resolving shortcuts.
+   * @return true if the edge is not a shortcut, not related to transit and not under construction
+   */
+  bool can_form_shortcut() const {
```
kevinkreiser, line 572: excellent

valhalla/baldr/directededge.h
```diff
@@ -1076,6 +1087,15 @@ class DirectedEdge {
     return superseded_;
   }
 
+  /**
+   * Unlike superseded(), this does not return the raw mask but the shortcut index
+   * that was originally passed to set_superseded().
+   * @return  Returns the index of the set bit in the superseded mask.
+   */
+  uint32_t superseded_idx() const {
+    return log2(superseded_) + 1;
```
kevinkreiser, line 1096: `log2` does some floating point stuff because it assumes we are trying to do a purely mathematical operation here. what we actually want to know is the position of the first non zero bit. i've never used it before but it seems a function called `ffs` (find first set) does this for us and usually has hardware optimization in modern architectures

src/mjolnir/shortcutbuilder.cc
```diff
@@ -386,10 +386,7 @@ uint32_t AddShortcutEdges(GraphReader& reader,
   for (uint32_t i = 0; i < edge_count; i++, ++edge_id) {
     // Skip transit connection edges.
     const DirectedEdge* directededge = tile->directededge(edge_id);
-    if (directededge->use() == Use::kTransitConnection ||
-        directededge->use() == Use::kEgressConnection ||
-        directededge->use() == Use::kPlatformConnection || directededge->bss_connection() ||
-        directededge->use() == Use::kConstruction) {
+    if (!directededge->can_form_shortcut()) {
```
kevinkreiser, line 389: there is another similar piece of code above that i think can make use of this function, its up on line 200

kevinkreiser: also in that method `CanContract` there are a bunch of stipulations that the node has to follow as well, it would be nice to add that as a function to nodeinfo and then maybe we need to use it in `GetShortcut` and `RecoverShortcut` as well?

valhalla/baldr/directededge.h
PhilippVoigt: Good point, I found a few builtins but did not come across `ffs`.

PhilippVoigt: Apparently this is not available on Windows, there is this [port of strings.h](https://github.com/win32ports/strings_h) or we implement it ourselves?

kevinkreiser: i had wondered about windows but didnt search for it. that sucks.. and also its in c++20 but we aren't there yet... how annoying. yeah i guess i would use strings.h in a "not windows" ifdef and then down in the function in an ifdef for when it is windows just implement the while loop and put a TODO that we'll use `std::countr_zero + 1` when we move to c++20. what do you prefer here?

src/mjolnir/shortcutbuilder.cc
kevinkreiser: definitely, i noticed that too and hoped you would as well :smile:

src/baldr/graphreader.cc
```diff
@@ -803,9 +804,16 @@ GraphId GraphReader::GetShortcut(const GraphId& id) {
     uint32_t idx = node->edge_index() + cont_de->opp_index();
     edgeid = {endnode.tileid(), endnode.level(), idx};
     directededge = tile->directededge(edgeid);
+
+    // If this edge is itself not the beginning of a shortcut, but we encountered another shortcut
+    // it means we must have started the traversal outside a shortcuts internal edges
+    if (!directededge->superseded() && shortcut_at_node) {
```
kevinkreiser, line 810: i think this works in most cases but it might be overly conservative. i need to review the shortcutbuilder but maybe its possible to have a graph like this


```
graph nodes: a, b, c, d, x
regular edges: .........
shortcut edges: ,,,,,,,,,

                                c
                                .,
                                .,
                                .,
                                .,
                                .,
 ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,.,
a.............x..........x......b,
                                .,
                                .,
                                .,
                                .,
                                x,
                                .,
                                .,
                                .,
                                .,
                                .,
                                d
```

so here we have a shortcut from `a` to `b` and another one from `c` to `d`. if you take an edge from inside of the latter shortcut and call `GetShortcut` it might find the former when it gets to node `b` and terminate.

now im not sure if this T intersection can happen with shortcuts passing by each other but if it can i think this means we will get a false negative.

PhilippVoigt: I could easily be wrong but the way I understand the ShortcutBuilder, a node is only eligible for contraction if it has exactly 2 eligible outbound edges, i.e. it is not a junction of 3 or more eligible roads.

In your example that would mean
- either node `b` cannot be contracted, because it has 3 eligible neighbors, so there might be two shortcuts `a->b`and `b->d` but `b` itself would not be part of one,
- or the outbound edge from `b` towards `a` is a transit edge or under construction, then there would be no shortcut `a->b` and the shortcut `c->d` could safely be resolved by `GetShortcut`.

Either way I'd love if you could confirm these assumptions 😉

PhilippVoigt:
>     1. the macos build is failing the shortcut tests hopefully its something trivial
Seems like the changes broke more than it fixed 😞 Some of the test failures seem quite unrelated to the changes...

src/baldr/graphreader.cc
kevinkreiser: you've looked at it more recently than i have so i trust your judgement :smile: i mean in any case what we are doing here is better than what we had before so it can only become more accurate.

kevinkreiser:
@PhilippVoigt did you want help looking at build failures? at first glace your last changes look perfect to me. just let me know if you want some help looking into what is going wrong with the build!

PhilippVoigt:
@kevinkreiser Thanks for offering help! After some digging it seems there are no shortcuts added to the graph anymore, most test failures are related to that fact. I'm now moving closer to the root cause, I'll let you know if I get stuck 😄

valhalla/baldr/nodeinfo.h
```diff
@@ -206,7 +206,7 @@ class NodeInfo {
   bool can_contract() const {
     return edge_count() >= 2 && intersection() != IntersectionType::kFork &&
            type() != NodeType::kGate && type() != NodeType::kTollBooth &&
-           type() != NodeType::kTollGantry && type() == NodeType::kSumpBuster;
```
kevinkreiser: i was sure it was going to be here but after looking at it 4 times i still didnt see it :smile:

PhilippVoigt: Not sure why the OSX build is still failing though. I'm developing on Mac myself so that is where I wrote the test in the first place 🤔

kevinkreiser:
with the mac failure being flakey the primary thing we should check is our use of opposing edges and graphtile. its very easy to mess up and use an edge with a tile to which it does not belong. because our edges can cross from one tile to another tile and we keep a copy of a tile around to re-use we can sometimes have a bug where the tile we use with the edge is not correct. this is what i would look for first

kevinkreiser:
@PhilippVoigt i wanted to lend a hand on this this weekend but didnt get a chance. i wanted to ask you if you tried with the cmake option `-DENABLE_SANITIZERS=ON`. that should help us find out if anything strange happens memory-access-wise during the test (was meine verherige vermutung war)

kevinkreiser:
i dont get why clang-tidy didnt tell us that we were using an uninitialized variable! these kinds of simple mistakes should all be caught by that..

kevinkreiser:
this passed all the tests in the last round, i simply fixed the conflict with the changelog. merging now and thanks again for sticking with this, its very much appreciated!

kevinkreiser:
i know its linked above but its hard to see. so just for full coverage i wanted to quote what you said and what i suspected would continue to be an issue for us (from this commit comment: https://github.com/valhalla/valhalla/commit/f54017797469b4f55049291b9f29a452328fe586#comments):

> Still when I run it on a larger graph I get false negatives in some cases where continuing_edge cannot decide between two equally valid edges

i remember debugging exactly this when i first wrote the shortcut cache singleton. i remember looking at one in saudi arabia or iraq or somewhere in that vicinity. anyway perhaps its just an iteration order thing? that doesnt seem intuitive because its all just looping over all edges at a node but there has to be something where we dont consider them in the right order or one loop bails in the loop when the other waits until the end of the loop and takes the last choice. anyway just need to look at the code for the 10th time to reason about it :smile:

PhilippVoigt:
@kevinkreiser Thanks for the merge! I know this is now closed but before opening something new I might just ask here.

I have isolated an example in Luxembourg and debugged both ShortcutBuilder and GetShortcut. It turns out that we ensure eligible edges fulfill the conditions `!start_restriction() && !end_restriction()` in both places but during shortcut generation there is a restriction and later in the final graph there is none.

The only place that I found removing restrictions is here: [graphvalidator.cc#L406](https://github.com/valhalla/valhalla/blob/606b035874cec0b1a072638a3a93ccc6040331bc/src/mjolnir/graphvalidator.cc#L406)
```
if (de->end_restriction()) {
     uint32_t modes = 0;
         for (uint32_t mode = 1; mode < kAllAccess; mode *= 2) {
             if ((de->end_restriction() & mode) &&
                 tile->GetRestrictions(true, edgeid, mode).size() > 0) {
                     modes |= mode;
         }
     }
     directededge.set_end_restriction(modes);
}
```

Now I'm wondering is it wrong to exclude edges that are part of complex restrictions from becoming part of a shortcut or is the it wrong the remove restrictions here. I'd be glad if you could share some ideas on this  👼

kevinkreiser:
@PhilippVoigt noch mal grossen dank für die ausführliche untersuchung!

my gut tells me we should not create shortcuts over complex restrictions... but let me see if i can reason about why...

if we did want to allow shortcuts the touched complex restrictions what would the rule have to be? i think it would have to be two fold:

1. you may include both `start_restriction` and `end_restriction` edges in a shortcut
2. an `end_restriction` edge may not appear after a `start_restriction` edge in the constructed shortcut

this basically means shortcut building and inclusion of edges has to be stateful, we have to remember if we ever used a `start_restriction` edge while building. the methods we used to filter candidates don't currently hold this state but for the sake of shortcut building could be made to do so. finding the shortcut that any one edge belongs to though also becomes a bit tougher, we'd need to, when walking backwards remember if we passed an `end_restriction` and if so, we would need to filter any `start_restriction` edges from the traversal. again this is totally doable.

by disallowing both we are a little more conservative than we need to be, which is sad because shortcuts help performance. at the same time it feels like a mild can of worms to have to worry about this stateful restriction traversal. so i guess i dont feel strongly about this point, if i had to pick one id rather more complexity for the sake of more shortcuts.

so then down to your other point about why the heck are we changing restrictions in validator. that one will require a bit more thought on my part. from a high level i agree with you, it would be far better to get it right the first time than to have code to come back and patch it. i would say we should ask the author of this but git blame says its me! i have no recollection of this but i can look at it a bit and reason about it. perhaps @gknisely or @dnesbitt61 remember why we revalidate/modify the modes of the complex restrictions in validator.

PhilippVoigt:
@kevinkreiser Danke für die vielen Details, das hilft mir sehr.

I have created a workaround that I feel should be valid without fully understanding what the graph validator does:
[Move complex turn restriction check out of can_form_shortcut() #4047](https://github.com/valhalla/valhalla/pull/4047)

Let me know what you think of it. I have tested it for our needs and it appears to be working.
At this point I feel that I can finally let this topic rest ☺️

PR #4020, opened 2023-03-11
Fix transit stitcher

nilsnolde:
When we were ingesting GTFS feeds, we were not stitching properly because we had a wrong way of determining whether a platform was generated.

I processed a real feed from Jena, Germany, and that was the first thing popping up. Then I kept fixing things until it worked, so it's again a bit overall patchy, sorry for that. I'll prepare well for the review session, since I unfortunately couldn't re-do the history and commit a bit nicer. But for sure I won't include variable renaming this time;)

I'm just building the full graph. Hope it'll work. If not, I'll see if it makes sense to include in this PR or if another bug is coming up for some reason when connecting the graphs.


src/baldr/graphtile.cc
```diff
@@ -980,38 +980,28 @@ const TransitDeparture* GraphTile::GetNextDeparture(const uint32_t lineid,
...
-          (!wheelchair || departures_[found].wheelchair_accessible()) &&
-          (!bicycle || departures_[found].bicycle_accessible())) {
-        return &departures_[found];
-      }
+    // Make sure it falls within the schedule and departure props are valid
+    const auto& d = departures_[found];
+    if (!GetTransitSchedule(d.schedule_index())->IsValid(day, dow, date_before_tile) ||
+        (wheelchair && !d.wheelchair_accessible()) || (bicycle && !d.bicycle_accessible())) {
+      continue;
+    }
```
nilsnolde, line 988: yeah, it's a refactor, but I find it much more legible. there's no functional changes and this stuff is fairly well tested

src/mjolnir/convert_transit.cc
```diff
@@ -47,14 +47,16 @@ namespace {
 
 // Struct to hold stats information during each threads work
 struct builder_stats {
-  uint32_t no_dir_edge_count;
-  uint32_t dep_count;
-  uint32_t midnight_dep_count;
+  uint32_t no_dir_edge_count = 0;
+  uint32_t dep_count = 0;
+  uint32_t midnight_dep_count = 0;
+  uint32_t invalid_service_dates = 0;
```
nilsnolde, line 53: report when there's calendar.txt entries which have 0 for all days and only add calendar_dates.txt entries which are either too far in the future or are in the past, compared with the tile creation date.

src/mjolnir/convert_transit.cc
```diff
@@ -184,39 +186,8 @@ ProcessStopPairs(GraphTileBuilder& transit_tilebuilder,
...
-            stop_no_access[dep.orig_pbf_graphid] |= kWheelchairAccess;
-            stop_no_access[dep.dest_pbf_graphid] |= kWheelchairAccess;
-          }
-
-          dep.bicycle_accessible = stop_pair.bikes_allowed();
-          dep.wheelchair_accessible = stop_pair.wheelchair_accessible();
-
-          // Compute days of week mask
+          // Compute the valid days
+          // set the bits based on the dow.// Compute days of week mask
```
nilsnolde, line 190: I moved the service date check right here so we abort early if there's no valid date range for a stop_pair's service

src/mjolnir/convert_transit.cc
```diff
@@ -264,17 +232,6 @@ ProcessStopPairs(GraphTileBuilder& transit_tilebuilder,
...
           }
 
-          // if dep.days == 0 then feed either starts after the end_date or tile_header_date >
-          // end_date
-          if (days == 0 && !stop_pair.service_added_dates_size()) {
-            LOG_ERROR("Feed rejected: " + pbf_fp +
-                      "!  Start date: " + to_iso_extended_string(start_date) +
-                      " End date: " + to_iso_extended_string(end_date));
-            continue;
-          }
```
nilsnolde, line 274: this was crazy verbose! in the feed I have, this happens frequently, lots of added dates to empty schedules are in january.

src/mjolnir/convert_transit.cc
```diff
@@ -264,17 +232,6 @@ ProcessStopPairs(GraphTileBuilder& transit_tilebuilder,
...
-
-          dep.headsign_offset = transit_tilebuilder.AddName(stop_pair.trip_headsign());
```
nilsnolde, line 276: moved this to a more proper place further down

src/mjolnir/convert_transit.cc
```diff
@@ -294,6 +251,46 @@ ProcessStopPairs(GraphTileBuilder& transit_tilebuilder,
             days = add_service_day(days, end_date, tile_date, add_date);
           }
 
+          // skip if no days are valid; this can happen a lot when there's a calendar.txt entry with
+          // no valid days but some calendar_dates.txt entries which add a date lying in the past
+          if (!days) {
+            stats.invalid_service_dates++;
+            continue;
+          }
```
nilsnolde, line 259: instead of having the log I removed, we stat invalid days

src/mjolnir/convert_transit.cc
```diff
@@ -413,9 +410,7 @@ std::list<PointLL> GetShape(const PointLL& stop_ll,
                             const float orig_dist_traveled,
                             const float dest_dist_traveled,
                             const std::vector<PointLL>& trip_shape,
-                            const std::vector<float>& distances,
-                            const std::string& origin_id,
-                            const std::string& dest_id) {
```
nilsnolde, line 418: these aren't used in this function (which looks proper). these are the onestop_id for origin/dest and this was the only time onestop_ids were used in convert_transit. it seems onestop_ids for the stop_pairs are only relevant for ingest_transit for the purpose of stitching.

src/mjolnir/convert_transit.cc
```diff
@@ -569,7 +558,6 @@ void AddToGraph(GraphTileBuilder& tilebuilder_transit,
     // Get the platform information
     GraphId platform_graphid = stop_edges.second.origin_pbf_graphid;
     const Transit_Node& platform = tile_pbf.nodes(platform_graphid.id());
-    const std::string& origin_id = platform.onestop_id();
```
nilsnolde, line 572: the next removed lines are all due to removing the 2 arguments to the `GetShape()` function

src/mjolnir/convert_transit.cc
```diff
@@ -1005,9 +990,6 @@ void build_tiles(const boost::property_tree::ptree& pt,
                  std::promise<builder_stats>& results) {
 
   builder_stats stats;
-  stats.no_dir_edge_count = 0;
-  stats.dep_count = 0;
-  stats.midnight_dep_count = 0;
```
nilsnolde, line 1010: defaulting to 0 in the struct

src/mjolnir/ingest_transit.cc
```diff
@@ -90,14 +90,18 @@ struct tile_transit_info_t {
 
 struct feed_cache_t {
   std::unordered_map<std::string, gtfs::Feed> cache;
+  std::string gtfs_dir;
+
+  feed_cache_t(const std::string& gtfs_dir) : gtfs_dir(gtfs_dir) {
+  }
```
nilsnolde, line 96: so this is the first sign for a bigger change: when parsing all feeds and collect their info per tile, we added uniqueness of the stop/trip etc objects by adding a feed's _absolute_ directory path. that's crazy wasteful, so I changed it to be only the directory name, i.e. instead of `{stop_id, "/home/nilsnolde/dev/cpp/valhalla/site/transit_feeds/Jena"`}`, it's now `{stop_id, "Jena"}`.

so here, we have to get the parent directory into the cache constructor, so we can still load the feeds properly.

src/mjolnir/ingest_transit.cc
```diff
@@ -129,7 +133,7 @@ struct unique_transit_t {
 std::string get_tile_path(const std::string& tile_dir, const GraphId& tile_id) {
   auto file_name = GraphTile::FileSuffix(tile_id);
   file_name = file_name.substr(0, file_name.size() - 3) + "pbf";
-  return tile_dir + filesystem::path::preferred_separator + file_name;
+  return tile_dir + file_name;
```
nilsnolde, line 136: we make sure `tile_dir` ends on a separator in the calling function

src/mjolnir/ingest_transit.cc
```diff
@@ -140,10 +144,12 @@ uint32_t to_local_epoch_sec(const std::string& dt) {
   return static_cast<uint32_t>(tp.time_since_epoch().count());
 };
 
-// Read from GTFS feed, sort data into the unique tiles they belong to
-std::priority_queue<tile_transit_info_t> select_transit_tiles(const boost::property_tree::ptree& pt) {
+std::string get_origin_id_base(const std::string& stop_id, const std::string feed_name) {
+  return feed_name + "_" + stop_id;
+}
```
nilsnolde, line 149: I also changed the definition of `onestop_id` to be e.g. `"Jena_stopid1" (plus the node_type extension if it was generated, but here we just get the base)

src/mjolnir/ingest_transit.cc
```diff
@@ -140,10 +144,12 @@ uint32_t to_local_epoch_sec(const std::string& dt) {
...
 
-  auto path = pt.get<std::string>("mjolnir.transit_feeds_dir");
+// Read from GTFS feed, sort data into the unique tiles they belong to
+std::priority_queue<tile_transit_info_t> select_transit_tiles(const std::string& gtfs_path) {
```
nilsnolde, line 152: we pass in the path to the gtfs directory, since we had to get that anyways in the calling function. also ensures to have a trailing separator, making things a bit safer.

src/mjolnir/ingest_transit.cc
```diff
@@ -158,22 +164,25 @@ std::priority_queue<tile_transit_info_t> select_transit_tiles(const boost::prope
...
   filesystem::recursive_directory_iterator end_file_itr;
   for (; gtfs_feed_itr != end_file_itr; ++gtfs_feed_itr) {
-    if (filesystem::is_directory(gtfs_feed_itr->path())) {
-      auto feed_path = gtfs_feed_itr->path().string();
-      LOG_INFO("Loading " + feed_path);
-      gtfs::Feed feed(feed_path);
+    const auto& feed_path = gtfs_feed_itr->path();
+    if (filesystem::is_directory(feed_path)) {
+      // feed_path has a trailing separator
+      const auto feed_name = feed_path.filename().string();
```
nilsnolde, line 173: this is the change where I take the feed name, not its full path, to create the tile_info objects

src/mjolnir/ingest_transit.cc
```diff
@@ -268,14 +277,14 @@ void setup_stops(Transit& tile,
   bool wheelchair_accessible = (tile_stop.wheelchair_boarding == "1");
   node->set_wheelchair_boarding(wheelchair_accessible);
   node->set_generated(isGenerated);
-  auto onestop_id = isGenerated ? tile_stop.stop_id + "_" + to_string(node_type) : tile_stop.stop_id;
-  node->set_onestop_id(onestop_id);
+  const auto& onestop_base = get_origin_id_base(tile_stop.stop_id, feed_name);
+  node->set_onestop_id(isGenerated ? onestop_base + "_" + to_string(node_type) : onestop_base);
```
nilsnolde, line 281: just a bit extended onestop_id

src/mjolnir/ingest_transit.cc
```diff
@@ -394,20 +406,30 @@ float get_stop_pair_dist(const gtfs::Stop& stop_connect,
...
-                     const std::unordered_map<feed_object_t, GraphId>& platform_node_ids,
-                     unique_transit_t& uniques) {
+bool write_stop_pair(
+    Transit& tile,
+    const tile_transit_info_t& tile_info,
+    const feed_object_t& feed_trip,
+    const gtfs::Feed& feed,
+    const std::unordered_map<feed_object_t, GraphId>& platform_node_ids,
+    unique_transit_t& uniques,
+    const google::protobuf::RepeatedPtrField<valhalla::mjolnir::Transit_Node>& tile_nodes) {
```
nilsnolde, line 416: just adding another argument for this tile's node objects

strange that clang-format decided to align the args differently

src/mjolnir/ingest_transit.cc
```diff
@@ -394,20 +406,30 @@ float get_stop_pair_dist(const gtfs::Stop& stop_connect,
...
   const auto& trip_calendar = feed.get_calendar(currTrip->service_id);
+  if (!currTrip || !trip_calendar) {
+    LOG_ERROR("Feed " + feed_trip.feed + ", trip ID" + tile_tripId +
+              " can't be found or has no calendar.txt entry, skipping...");
+    return false;
+  }
+
+  // get the gtfs shape and our pbf shape_id if present
+  const auto& currShape = feed.get_shape(currTrip->shape_id);
+  auto pbf_shape_it = tile_info.shapes.find({currTrip->shape_id, feed_trip.feed});
```
nilsnolde, line 432: so we can get its tile-specific shape_id

nilsnolde: I'm very proud of myself BTW that I didn't rename `currShape` at this point hahaha

src/mjolnir/ingest_transit.cc
```diff
@@ -426,10 +448,26 @@ bool write_stop_pair(Transit& tile,
...
 
+    // if it's not in the tile, we can't do anything else than take it's gtfs stop_id
+    // we check further down when stitching and adjust if it's was generated
+    std::string origin_onestop_id = origin_is_in_tile
+                                        ? tile_nodes.Get(origin_graphid_it->second.id()).onestop_id()
+                                        : get_origin_id_base(origin_stopId, feed_trip.feed);
+    std::string dest_onestop_id = dest_is_in_tile
+                                      ? tile_nodes.Get(dest_graphid_it->second.id()).onestop_id()
+                                      : get_origin_id_base(dest_stopId, feed_trip.feed);
+
```
nilsnolde, line 459: onestop_id equals e.g. `"Jena_stopid1"` if the origin/dest is not in this tile (note: this is actually wrong when it's generated, as it'd have its node_type appended). It gets its real onestop_id (including its node_type if generated) if it is in the tile.

src/mjolnir/ingest_transit.cc
```diff
@@ -426,10 +448,26 @@ bool write_stop_pair(Transit& tile,
...
+    // we don't use this value unless the origin is in the tile, so it's fine to set it false
+    bool origin_is_generated =
+        origin_is_in_tile ? tile_nodes.Get(origin_graphid_it->second.id()).generated() : false;
+    bool dest_is_generated =
+        dest_is_in_tile ? tile_nodes.Get(dest_graphid_it->second.id()).generated() : false;
```
nilsnolde, line 464: we can only determine if origin/dest is generated if they're inside the tiles, otherwise we assume `false` (it doesn't matter though, this variable is only used once if the location is in the tile).

src/mjolnir/ingest_transit.cc
```diff
@@ -438,12 +476,34 @@ bool write_stop_pair(Transit& tile,
...
+        } else
+          stop_pair->add_service_except_dates(d);
+      }
+
+      // this shouldn't happen, but let's make sure it doesn't
+      // in convert_transit we'll check if there was a valid date for this service and skip if not
+      if (!service_dow->size() && !had_added_date) {
+        LOG_WARN("Service ID " + currTrip->service_id +
+                 " has no valid calendar or calendar_dates entry, skipping...");
+        tile.mutable_stop_pairs()->RemoveLast();
```
nilsnolde, line 496: if there's no valid date range, remove this service's stop_pairs entirely

src/mjolnir/ingest_transit.cc
```diff
@@ -438,12 +476,34 @@ bool write_stop_pair(Transit& tile,
...
         continue;
       }
 
+      // test this, but careful, we might have to adjust the test's dist_shape_traveled or whatever
+      // for the test shapes to be a bit more realistic with the actual map where it travels much
+      // further than the GTFS objects indicate
+      if (!currShape.empty()) {
+        stop_pair->set_shape_id(pbf_shape_it->second);
+      }
```
nilsnolde, line 505: only set a shape_id when it's not empty. the shape is being used by convert_transit only if there's a `shape_id` set

src/mjolnir/ingest_transit.cc
```diff
@@ -457,58 +517,45 @@ bool write_stop_pair(Transit& tile,
...
-
-      bool dest_is_generated = tile_info.station_children.find({dest_stopId, currFeedPath}) !=
-                               tile_info.station_children.end();
-      auto dest_onestop_id = dest_is_generated
-                                 ? dest_stopId + "_" + to_string(NodeType::kMultiUseTransitPlatform)
-                                 : dest_stopId;
       stop_pair->set_destination_onestop_id(dest_onestop_id);
 
+      stop_pair->set_destination_arrival_time(dest_stopTime.arrival_time.get_total_seconds());
+      stop_pair->set_origin_departure_time(origin_stopTime.departure_time.get_total_seconds());
```
nilsnolde, line 524: this was also a bug: we have to set these attributes regardless if the locations are inside the tile or not. otherwise they wouldn't be set at all for some pairs which breaks lots of stuff

src/mjolnir/ingest_transit.cc
```diff
@@ -457,58 +517,45 @@ bool write_stop_pair(Transit& tile,
...
                                       static_cast<uint64_t>(origin_is_generated));
 
-        // call function to set shape
-        float dist = get_stop_pair_dist(*origin_stop, currShape, origin_stopTime);
-        stop_pair->set_origin_dist_traveled(dist);
+        // call function to set shape, only if it has a shape_id, otherwise no need (see
+        // convert_transit)
+        if (stop_pair->has_shape_id()) {
+          float dist = get_stop_pair_dist(*origin_stop, currShape, origin_stopTime);
+          stop_pair->set_origin_dist_traveled(dist);
```
nilsnolde, line 540: also the dist_traveled is only interesting for convert_transit if there's a shape_id

src/mjolnir/ingest_transit.cc
```diff
@@ -523,6 +570,8 @@ bool write_stop_pair(Transit& tile,
       if (!feed.get_frequencies(currTrip->trip_id).empty()) {
         const auto& currFrequencies = feed.get_frequencies(currTrip->trip_id);
         if (currFrequencies.size() > 1) {
+          // TODO(nils): this should be properly handled as 1 trip id can have
+          // multiple frequencies, e.g. the example Google feed does
```
nilsnolde, line 574: yeah I confirmed there can be more than 1 frequency per trip id, see https://developers.google.com/static/transit/gtfs/examples/sample-feed.zip

src/mjolnir/ingest_transit.cc
```diff
@@ -607,7 +656,9 @@ void write_shapes(Transit& tile, const tile_transit_info_t& tile_info, feed_cach
 }
 
 // pre-processes feed data and writes to the pbfs (calls the 'write' functions)
-void ingest_tiles(const boost::property_tree::ptree& pt,
+void ingest_tiles(const std::string& gtfs_dir,
+                  const std::string& transit_dir,
+                  const uint32_t pbf_trip_limit,
```
nilsnolde, line 661: pass those into the threads, they had to be generated in the calling function anyways

src/mjolnir/ingest_transit.cc
```diff
@@ -672,12 +727,19 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
...
   auto grid = TileHierarchy::GetTransitLevel().tiles;
-  auto transit_dir = pt.get<std::string>("mjolnir.transit_dir");
+
+  // for a missing stop_pair member's onestop_id we had to take the stop's gtfs stop_id
+  // but generated nodes actually have their node_type appended on the onestop_id, so here we
+  // need to remove the node_type before matching a candidate tile's nodes with the missing ones
+  const auto stop_id_from_onestop_id = [](const auto& n_oid, NodeType node_type) {
+    const auto type_found = n_oid.find("_" + to_string(node_type));
+    return type_found != n_oid.npos ? n_oid.substr(0, n_oid.length() - type_found) : n_oid;
+  };
```
nilsnolde, line 742: this lambda will accept a tile node's onestop_id and its node_type. if it finds the node_type in the onestop_id, it'll strip it and return the "base" onestop_id, e.g. in: `"Jena_stopid1_transit_multi_platorm"`, out: `"Jena_stopid1"`.

because when looking for missing stop_pair members, we only have the "base" onestop_id. but it's safe to use, since there can't possibly be both inside a single tile: a non-generated `"Jena_stopid1"` **and** a `"Jena_stopid1_transit_multi_platorm"`, which would lead to random results.

src/mjolnir/ingest_transit.cc
```diff
@@ -722,7 +784,11 @@ void stitch_tiles(const boost::property_tree::ptree& pt,
         auto neighbor_file_name = get_tile_path(transit_dir, neighbor_id);
         auto neighbor = read_pbf(neighbor_file_name, lock);
         for (const auto& node : neighbor.nodes()) {
-          auto platform_itr = needed.find(node.onestop_id());
+          const auto node_type = static_cast<NodeType>(node.type());
+          if (node_type != NodeType::kMultiUseTransitPlatform) {
+            continue;
+          }
```
nilsnolde, line 790: don't look for stations/egresses

test/gurka/test_gtfs.cc
```diff
@@ -121,7 +121,17 @@ TEST(GtfsExample, WriteGtfs) {
...
+  // 1st has all stop objects, egress/station/platform
+  // 2nd has only platform
+  // 3rd has only station
+  struct gtfs::Stop first_stop_station {
+    .stop_id = stopOneID, .stop_name = gtfs::Text("POINT NEMO"), .coordinates_present = true,
+    .stop_lat = station_one_ll->second.second, .stop_lon = station_one_ll->second.first,
+    .parent_station = "", .location_type = gtfs::StopLocationType::Station,
+    .wheelchair_boarding = "1",
+  };
+  feed.add_stop(first_stop_station);
```
nilsnolde, line 134: only moved this

test/gurka/test_gtfs.cc
```diff
@@ -136,37 +146,22 @@ TEST(GtfsExample, WriteGtfs) {
...
-  struct gtfs::Stop second_stop_station {
+  struct gtfs::Stop second_stop_platform {
     .stop_id = stopTwoID, .stop_name = gtfs::Text("SECOND STOP"), .coordinates_present = true,
     .stop_lat = station_two_ll->second.second, .stop_lon = station_two_ll->second.first,
-    .parent_station = "", .location_type = gtfs::StopLocationType::Station,
+    .parent_station = "", .location_type = gtfs::StopLocationType::StopOrPlatform,
     .stop_timezone = "America/Toronto", .wheelchair_boarding = "1",
   };
-  feed.add_stop(second_stop_station);
+  feed.add_stop(second_stop_platform);
```
nilsnolde, line 156: with stop 2 we only add a platform but no station and no egress

test/gurka/test_gtfs.cc
```diff
@@ -136,37 +146,22 @@ TEST(GtfsExample, WriteGtfs) {
...
   };
   feed.add_stop(third_stop_station);
-  struct gtfs::Stop third_stop_platform {
-    .stop_id = stopThreeID + "_walk_the_plank_pal", .stop_name = gtfs::Text("THIRD STOP"),
-    .coordinates_present = true, .stop_lat = station_three_ll->second.second,
-    .stop_lon = station_three_ll->second.first, .parent_station = stopThreeID,
-    .location_type = gtfs::StopLocationType::StopOrPlatform, .stop_timezone = "America/Toronto",
-    .wheelchair_boarding = "1",
-  };
-  feed.add_stop(third_stop_platform);
```
nilsnolde, line 169: also no station for the last platform

nilsnolde:
It's still a draft, but I don't understand why OSX/clang is failing with 
```
/Users/distiller/project/valhalla/midgard/point2.h:204:28: error: implicit instantiation of undefined template 'std::basic_string<char>'
  inline const std::string to_string() const {
```

(I added a `to_string()` method to `Point2` for convenience when printing coords)

src/mjolnir/convert_transit.cc
nilsnolde: (actually they were used, but only for printing when smth went wrong in `GetShape()`, I replaced that with both stop's lon/lat)

nilsnolde:
I ran a feed and looks good :partying_face: (there was no shapes.txt entries, so it's all direct lines connecting the stops)

image

src/mjolnir/ingest_transit.cc
```diff
@@ -140,10 +144,12 @@ uint32_t to_local_epoch_sec(const std::string& dt) {
   return static_cast<uint32_t>(tp.time_since_epoch().count());
 };
 
-// Read from GTFS feed, sort data into the unique tiles they belong to
-std::priority_queue<tile_transit_info_t> select_transit_tiles(const boost::property_tree::ptree& pt) {
+std::string get_origin_id_base(const std::string& stop_id, const std::string& feed_name) {
```
nilsnolde, line 147: rename to onestop_id

src/mjolnir/ingest_transit.cc
```diff
@@ -340,9 +349,12 @@ write_stops(Transit& tile, const tile_transit_info_t& tile_info, feed_cache_t& f
...
       }
     }
-    // We require platforms as well so if we didnt add at least 1 we need to fake one now
+    // This really shouldn't happen, as platform IDs are referenced by the
+    // stop_times.txt and the whole downstream logic would get messed up
+    // can only happen if there was an entrance and a platform at the same station
+    // without that station being referenced by their parent_station atttributes
     if (tile.nodes_size() == node_count) {
-      LOG_WARN("Generated platform for station " + station_as_stop->stop_id);
+      LOG_ERROR("Generated platform for station " + station_as_stop->stop_id);
```
kevinkreiser, line 357: we should put a TODO  here that peels back the changes above as well so that we can completely skip this stop. now it will also be possible that other things refer to this in which case we will have more errors.. but as best we can we should try to be robust to it. maybe after more imports it will be obvious the best way to recover from such a screwed up feed

src/mjolnir/ingest_transit.cc
nilsnolde: pull out the stop_time's dist_traveled

src/mjolnir/ingest_transit.cc
```diff
@@ -672,12 +727,18 @@ void ingest_tiles(const boost::property_tree::ptree& pt,
...
                   const std::unordered_set<GraphId>& all_tiles,
                   std::list<GraphId>& tiles,
                   std::mutex& lock) {
   auto grid = TileHierarchy::GetTransitLevel().tiles;
-  auto transit_dir = pt.get<std::string>("mjolnir.transit_dir");
+
+  // for a missing stop_pair member's onestop_id we had to take the stop's gtfs stop_id
+  // but generated nodes actually have their node_type appended on the onestop_id, so here we
+  // need to remove the node_type before matching a candidate tile's nodes with the missing ones
+  const auto stop_id_from_onestop_id = [](const auto& n_oid, NodeType node_type, bool is_generated) {
```
kevinkreiser, line 739: so one interesting question i had about this was that when we do the stitching itself and we are robust to this name not having the _type on the end, we dont actually patch the onestop_id when we find out if it is actually generated in the other tile. we talked about it and it sounds like we never use the id again so its safe but itd be good to write that in this comment because i think its a pretty tricky but important caveat. in other words if we ever decided we needed the one stop id again (for deduplication for example, it would be important to update the id as well to match its generated counter part, maybe, maybe it still wont be important but good to know about)

test/gurka/test_gtfs.cc
```diff
@@ -136,36 +146,21 @@ TEST(GtfsExample, WriteGtfs) {
...
-    .wheelchair_boarding = "1",
-  };
-  feed.add_stop(first_stop_station);
 
-  struct gtfs::Stop second_stop_station {
+  struct gtfs::Stop second_stop_platform {
     .stop_id = stopTwoID, .stop_name = gtfs::Text("SECOND STOP"), .coordinates_present = true,
     .stop_lat = station_two_ll->second.second, .stop_lon = station_two_ll->second.first,
-    .parent_station = "", .location_type = gtfs::StopLocationType::Station,
+    .parent_station = "", .location_type = gtfs::StopLocationType::StopOrPlatform,
```
kevinkreiser, line 153: i was hoping to cover all the cases in the test, a stop that has all 3 node types (egress/station/platform), a stop that has 2 of them (station/platform) and a stop that has just platform. this way we cover all permutations of generated not generated stuff

nilsnolde:
@kevinkreiser 

addressed the review:
https://github.com/valhalla/valhalla/pull/4020/commits/d28091cb4d00276ac1e1c515cbee230c627d4fd4: the trivial stuff
https://github.com/valhalla/valhalla/pull/4020/commits/f12cabb99ea7afd4d3ef5cbf15f34e0973e4910d: refactor how origin/dest dist_travelled is determined, also pulling it out of the `origin_is_in_tile` condition as we need to set it regardless

test/gurka/test_gtfs.cc
```diff
@@ -122,23 +122,23 @@ TEST(GtfsExample, WriteGtfs) {
   feed.write_agencies(path_directory);
 
   // write stops.txt:
-  // 1st has all stop objects, egress/station/platform
-  // 2nd has only platform
+  // 1st has all stop objects, egress/station/platform, but egress has no parent
+  // 2nd has only platform, station, but the platform has not parent
   // 3rd has only station
```
kevinkreiser: nit: maybe im reading the diff wrong but it seems like the comments here arent exactly right and the 2nd stop is deliberately incorrect? what i see is:

* 1st has all stop types, egress/station/platform (the egress and platform do point to the station as the correct parent parent)
* 2nd has station/platform only (the platform does not point to the parent station, so the station is orphaned, why do we test this? shouldnt we point the platform to the parent station?)
* 3rd has only a platform

kevinkreiser (approved):
i think the test stops could use some clean up but its ok to get it in the next round

test/gurka/test_gtfs.cc
nilsnolde: > so the station is orphaned

Right, but the comment is actually saying that.

> why do we test this?

I realized a bit too late that this is bs actually. What would you like to see here? I had only a platform before without a station for the second one, but you wanted to see more when we reviewed. Nothing really comes to my mind that would make more sense tbh. An egress without a parent station could be, but that's pretty much the same as an orphaned station.

kevinkreiser (approved):

PR #4022, opened 2023-03-13
try to bump osmpbf to v1.5.0

### LINKED ISSUE #4021: Update OSMPBF repo
We could (try to) update our OSM proto sub repo. Of course there hasn't been much going on since you included it, but one thing has been added (7 years ago :smile: ): `syntax=proto2`.

nilsnolde:
closes #4021


nilsnolde:
I realized I should at least give a reason why I closed this, for future reference.

@kevinkreiser realized linking is failing because OSM removed the protobuf-lite statement for their PBF definition. So we'd have to patch it. Not a problem but also why..

PR #4032, opened 2023-03-20
fix multimodal isochrones

nilsnolde:
Overhauled multimodal isochrones and made it consistent with multimodal routing. It's not that easy to check, even manually, that it did the right thing, but it for now it looks good:)

In the next round I'll update the tests for both routing and isochrones a bit.


src/loki/reach.cc
```diff
@@ -221,8 +221,11 @@ void Reach::ExpandingNode(baldr::GraphReader&,
 }
 
 // when the main loop is looking to continue expanding we tell it to terminate here
-thor::ExpansionRecommendation
-Reach::ShouldExpand(baldr::GraphReader&, const sif::EdgeLabel&, const thor::ExpansionType) {
+thor::ExpansionRecommendation Reach::ShouldExpand(baldr::GraphReader&,
+                                                  const sif::EdgeLabel&,
+                                                  const thor::ExpansionType,
+                                                  const float /*time*/,
+                                                  const uint32_t /*dist*/) {
```
nilsnolde, line 228: I added those, since the `Isochrone::ShouldExpand` was previously looking at the edge labels, but with multimodal we use `MMEdgeLabel`, not `BDEdgeLabel` so it's either this or templating the `ShouldExpand` function.

src/thor/dijkstras.cc
```diff
@@ -444,21 +460,13 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
...
 
-    if (processed_tiles_.find(tile->id().tileid()) == processed_tiles_.end()) {
-      tc->AddToExcludeList(tile);
-      processed_tiles_.emplace(tile->id().tileid());
-    }
-
-    // check if excluded.
-    if (tc->IsExcluded(tile, nodeinfo)) {
-      return;
-    }
```
nilsnolde, line 455: moved this further up to make it consistent with routing

src/thor/dijkstras.cc
```diff
@@ -581,32 +589,29 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
...
 
       // Regular edge - use the appropriate costing and check if access
       // is allowed. If mode is pedestrian this will validate walking
       // distance has not been exceeded.
-      if (!mode_costing[static_cast<uint32_t>(mode_)]->Allowed(directededge, false, pred, tile,
-                                                               edgeid, 0, 0, restriction_idx)) {
+      if (!pc->Allowed(directededge, false, pred, tile, edgeid, 0, 0, restriction_idx) ||
+          walking_distance > max_walking_dist_) {
         continue;
       }
```
nilsnolde, line 610: the only thing that really changed here is that we're checking against the `max_walking_dist_` here. it didn't do that before and I'm not 100% we really want to do that for isochrones. IMO we should, but it's open for discussion.

src/thor/dijkstras.cc
```diff
@@ -622,31 +627,20 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
...
       continue;
     }
 
     // Test if exceeding maximum transfer walking distance
-    if (directededge->use() == Use::kTransitConnection && pred.prior_stopid().Is_Valid() &&
+    // TODO: transfer distance != walking distance! (one more label member?)
+    if (directededge->use() == Use::kPlatformConnection && pred.prior_stopid().Is_Valid() &&
         walking_distance > max_transfer_distance_) {
       continue;
     }
```
nilsnolde, line 642: these ones are not really proper IMO, same situation as in the routing algo (where I updated it too with comments). let's reason over this together tmrw

src/thor/dijkstras.cc
```diff
@@ -622,31 +627,20 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
...
-                           walking_distance, tripid,      prior_stop,      blockid,
-                           operator_id,      has_transit, transition_cost, restriction_idx,
-                           pred.path_id()};
-
-    // See if this is even worth expanding
-    auto maybe_expand = ShouldExpand(graphreader, edge_label, ExpansionType::multimodal);
-    if (maybe_expand == ExpansionRecommendation::prune_expansion ||
-        maybe_expand == ExpansionRecommendation::stop_expansion) {
-      continue;
-    }
```
nilsnolde, line 648: we already do this in the outer loop. true, we wouldn't have to add it to the queue & labels in the first place, but I did find it silly to have this check twice.

src/thor/dijkstras.cc
```diff
@@ -665,7 +659,10 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
     // Add edge label, add to the adjacency list and set edge status
     uint32_t idx = mmedgelabels_.size();
     *es = {EdgeSet::kTemporary, idx};
-    mmedgelabels_.emplace_back(edge_label);
+    mmedgelabels_.emplace_back(pred_idx, edgeid, directededge, newcost, newcost.cost, 0.0f, mode_,
+                               path_distance, walking_distance, tripid, prior_stop, blockid,
+                               operator_id, has_transit, transition_cost, restriction_idx,
+                               pred.path_id());
```
nilsnolde, line 665: avoid the copy

src/thor/dijkstras.cc
```diff
@@ -684,7 +681,15 @@ void Dijkstras::ComputeMultiModal(
...
     const sif::mode_costing_t& mode_costing,
-    const travel_mode_t mode) {
+    const travel_mode_t mode,
+    const Options& options) {
+
+  // For now the date_time must be set on the origin.
+  if (origin_locations.Get(0).date_time().empty()) {
+    LOG_ERROR("No date time set on the origin location");
+    return;
+  }
```
nilsnolde, line 691: copied from further down to early exit

src/thor/dijkstras.cc
```diff
@@ -694,29 +699,22 @@ void Dijkstras::ComputeMultiModal(
...
-
-  // For now the date_time must be set on the origin.
-  if (origin_locations.Get(0).date_time().empty()) {
-    LOG_ERROR("No date time set on the origin location");
-    return;
-  }
+  // Get maximum transfer and walking distance
+  max_transfer_distance_ = pc->GetMaxTransferDistanceMM();
+  max_walking_dist_ =
+      options.costings().find(Costing::pedestrian)->second.options().transit_start_end_max_distance();
```
nilsnolde, line 705: not so sure about these two.. these checks were basically disabled before. I still do think that also with isochrones we might want to respect the relevant costing options. WDYT @gknisely ?

nilsnolde:
The commit history is mostly from another branch, should've rebased before..

src/thor/dijkstras.cc
```diff
@@ -419,8 +424,19 @@ void Dijkstras::ExpandForwardMultiModal(GraphReader& graphreader,
...
+    if (processed_tiles_.find(tile->id().tileid()) == processed_tiles_.end()) {
+      tc->AddToExcludeList(tile);
+      processed_tiles_.emplace(tile->id().tileid());
+    }
+
+    // check if excluded.
+    if (tc->IsExcluded(tile, nodeinfo)) {
+      return;
+    }
+  } else if (!mode_costing[static_cast<uint8_t>(mode_)]->Allowed(nodeinfo)) {
```
nilsnolde, line 439: remove the `else`

test/gurka/test_gtfs.cc
```diff
@@ -619,16 +619,26 @@ TEST(GtfsExample, MakeTile) {
...
+      gurka::do_action(valhalla::Options::isochrone, map, {"C"}, "multimodal",
+                       {{"/date_time/type", "1"},
+                        {"/date_time/value", "2023-02-27T05:58"},
+                        {"/contours/0/time", "20"},
+                        {"/costing_options/pedestrian/transit_start_end_max_distance", "20000"}},
+                       {}, &res_string);
+
+  rapidjson::Document doc;
+  doc.Parse(res_string.c_str());
+  EXPECT_TRUE(doc.HasMember("features"));
```
nilsnolde, line 643: put todo for future testing stuff

nilsnolde:
@kevinkreiser addressed the review:
- 6054d3bfa98b47eda10514920bf9b854dc9ef66b: turn else/if into a separate if
- c763514b474fbf0a8b96cdb5afa8ecc27871dfd8: leave a TODO for isochrone test
- f38654a315bef563606e6db98c667e27df6e1dc8: revert the time/dist args and push it all back into the isochrone implementation of `ShouldExpand`

kevinkreiser (approved):
thanks for the walkthrough and other than the simple fix, im happy for all the cleanups (eg. candidate penalty, transfer cap)

PR #4033, opened 2023-03-20
replace dockerhub with github docker registry

### LINKED ISSUE #4029: migrate docker builds to github
docker is nuking all free accounts, which is cool with me. one less place to care about. good news is github supports docker builds directly already. so yeah lets just migrate to githubs docker container registry: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

nilsnolde:
closes #4029 

Let's try with Github Actions, so much easier. On CircleCI it was running with a default machine as well, not the big ones. So probably similar performance. As temp I enabled to build this branch.

Also removes the `run-` prefix as that's not applicable anymore and since people will have to update the URL they're pulling from anyways, we can break this as well IMO.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,25 +19,10 @@ commands:
     steps:
       - run: sudo bash ./scripts/install-linux-deps.sh
 
-jobs:    
-  build-docker-images:
-    machine:
-      image: ubuntu-2004:current
-    steps:
-      - checkout
-      - run:
-          name: Build and push Valhalla Docker image
-          command: |
-            if [[ "${CIRCLE_BRANCH}" == "master" ]] || [[ ! -z "${CIRCLE_TAG}" ]]; then
-              BUILD_TAG="${CIRCLE_TAG:-latest}"
-              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
-              docker build --pull --tag valhalla/valhalla:run-${BUILD_TAG} .
-              docker push valhalla/valhalla:run-${BUILD_TAG}
-            fi
-
+jobs:
   lint-build-debug:
     machine:
-      image: ubuntu-2004:current
+      image: ubuntu-2204:current
     resource_class: xlarge
     steps:
       - checkout
@@ -86,7 +71,7 @@ jobs:
 
   build-release:
     machine:
-      image: ubuntu-2004:current
+      image: ubuntu-2204:current
     resource_class: xlarge
     steps:
       - checkout
@@ -159,12 +144,6 @@ workflows:
   version: 2
   build_test_publish:
     jobs:
-      - build-docker-images:
-          filters:
-            tags:
-              only: /.*/
-            branches:
-              only: master
       - lint-build-debug:
           filters:
             tags:
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -16,6 +16,8 @@ jobs:
     steps:
       - name: Check out the repo
         uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
       - name: Extract branch name
         if: startsWith(github.ref, 'refs/tags/')
         run: echo "##[set-output name=branch;]${GITHUB_REF#refs/heads/}"
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install newer git
-      - run: |
+        run: |
           sudo apt-get install -y software-properties-common \
           && sudo apt-get update \
           && sudo add-apt-repository -y ppa:git-core/ppa \
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -22,7 +22,7 @@ jobs:
           && sudo apt-get update \
           && sudo apt-get install -y git
       - name: Check out the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
         with:
           submodules: 'recursive'
           fetch-depth: 0
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,43 @@
+name: Publish Docker image
+on:
+  push:
+    branches:
+      - '*'
+    tags:
+      - '*'
+    paths-ignore:
+      - '.gitignore'
+      - '**.md'
+      - 'test/'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+      - name: Extract branch name
+        if: startsWith(github.ref, 'refs/tags/')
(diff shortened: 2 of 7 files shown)


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,52 @@
...
+      - '.gitignore'
+      - '**.md'
+      - 'test/'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install newer git
+      - run: |
```
michaelkirk:
```suggestion
        run: |
```

michaelkirk: Typo here I think?

michaelkirk: whoops. jinx.

nilsnolde: it's doing smth weird when building the image with docker's action..

```
/usr/bin/docker buildx build --iidfile /tmp/docker-build-push-HO6wTt/iidfile --secret id=GIT_AUTH_TOKEN,src=/tmp/docker-build-push-HO6wTt/tmp-4947-OT445lVOThVI --tag ghcr.io/valhalla/valhalla:latest --metadata-file /tmp/docker-build-push-HO6wTt/metadata-file https://github.com/valhalla/valhalla.git#823f1c924454555873d28fcb09bcac4ddf266d6e
#1 [internal] load git source https://github.com/valhalla/valhalla.git#823f1c924454555873d28fcb09bcac4ddf266d6e
#0 0.025 Initialized empty Git repository in /var/lib/docker/overlay2/vfkb97p0dg7afufrjfuyb3cie/diff/
#0 0.029 fatal: Not a valid object name 823f1c924454555873d28fcb09bcac4ddf266d6e^{commit}
#1 12.00 From https://github.com/valhalla/valhalla
```

.circleci/config.yml
```diff
@@ -19,25 +19,10 @@ commands:
...
-              echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
-              docker build --pull --tag valhalla/valhalla:run-${BUILD_TAG} .
-              docker push valhalla/valhalla:run-${BUILD_TAG}
-            fi
-
+jobs:
   lint-build-debug:
     machine:
-      image: ubuntu-2004:current
+      image: ubuntu-2204:current
```
nilsnolde, line 25: maybe we also want to keep building on 20.04 to keep some lower versions of deps? osx is already pretty up-to-date testing deps versions

.github/workflows/docker-build.yml
michaelkirk: Wild. Maybe that's https://github.com/marketplace/actions/build-and-push-docker-images#git-context?

Dockerfile
```diff
@@ -14,8 +14,8 @@ COPY ./scripts/install-linux-deps.sh /usr/local/src/valhalla/scripts/install-lin
 RUN bash /usr/local/src/valhalla/scripts/install-linux-deps.sh
 
 # get the code into the right place and prepare to build it
-ADD . /usr/local/src/valhalla
-RUN ls
+ADD . .
+RUN ls -la
```
nilsnolde, line 18: this is a remnant from testing when it didn't work, but doesn't hurt to keep either

.github/workflows/docker-build.yml
nilsnolde: Oh yeah, looks very relevant, thanks! Only skimmed it, already using manual approach, it's even less lines:) seems to work finally!

nilsnolde:
BTW, I built the image in https://github.com/valhalla/valhalla/actions/runs/4473531846/jobs/7861023336, but I canceled the build or it'd have pushed this branch's image as `latest` (would've been fine actually, but still). Tagged ones will also work, I use that in other repos.

michaelkirk (approved):
The docker command line is already so minimal that it indeed seems like a good idea to forego the "fancy" docker GH action.

Everything else (also) looks reasonable to me.

.circleci/config.yml
kevinkreiser: nah 22.04 is fine with me! is circle on 22.04 though? it would be sad if docker didnt match CI because we would merge when circle is good on 20.04 but then docker could fail. mismatch isnt good but 22.04 is fine with me across the board if that makes sense?

Dockerfile
kevinkreiser: oh i see the workdir is already set there

.circleci/config.yml
kevinkreiser: ah sorry was confused about what this was in passing, i see you are making circle match docker, this is great imho!

nilsnolde: > mismatch isnt good but 22.04 is fine with me across the board if that makes sense

I’m not sure it does to me tbh 😄What’s the relation between the docker workflow and the testing environment on circleci? For me they’re quite independent.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+          fetch-depth: 0
+      - name: Extract branch name
+        if: startsWith(github.ref, 'refs/tags/')
+        run: echo "##[set-output name=branch;]${GITHUB_REF#refs/heads/}"
```
nilsnolde, line 24: oops, this is getting the wrong thing, gotta look at `refs/tags` and set the tag, not the branch name

nilsnolde:
Jep that worked, 2 brand new images :)

https://github.com/valhalla/valhalla/pkgs/container/valhalla

Let me revert the test stuff before we merge.

nilsnolde:
this is also done @kevinkreiser . The 2 PRs should merge cleanly without conflict

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build tagged image
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          docker build -t ghcr.io/valhalla/valhalla:${{ steps.extract_branch.outputs.tag }} .
+          docker push ghcr.io/valhalla/valhalla:${{ steps.extract_branch.outputs.tag }}
```
kevinkreiser, line 36: i just noticed you removed the `run-` prefixes here and below, now that we dont have a "build" container that we actually publish i think this is fine (our build container is literally just run that deps script)

.circleci/config.yml
nilsnolde: Right, got your point. I always forget that the docker build only runs after we merged.. Makes sense.

.github/workflows/docker-build.yml
nilsnolde: Jep, I also mentioned that in the PR description. I also feel it’s ok to break that now, with a different URL

nilsnolde:
I’ll also remove the test tag

PR #4035, opened 2023-03-21
update multimodal docs

nilsnolde:
Remove stale scripts and references referring to the deprecated transitland method to get transit data into Valhalla and replace with raw GTFS references.


docs/docs/api/isochrone/api-reference.md
```diff
@@ -35,13 +35,13 @@ Refer to the [route location documentation](../turn-by-turn/api-reference.md#loc
...
 
 ### Other request parameters
 
 | Parameter | Description |
 | :------------------ | :----------- |
-| `date_time` | The local date and time at the location. These parameters apply only for multimodal requests and are not used with other costing methods.<ul><li>`type`<ul><li>0 - Current departure time for multimodal requests.</li><li>1 - Specified departure time for multimodal requests.</li><li>2 - Specified arrival time. Note: This is not yet implemented.</li></ul></li><li>`value` - the date and time specified in ISO 8601 format (YYYY-MM-DDThh:mm) in the local time zone of departure or arrival. For example, "2016-07-03T08:06"</li></ul> |
+| `date_time` | The local date and time at the location. <ul><li>`type`<ul><li>0 - Current departure time.</li><li>1 - Specified departure time.</li><li>2 - Specified arrival time. Note: This is not yet implemented for `multimodal`.</li></ul></li><li>`value` - the date and time specified in ISO 8601 format (YYYY-MM-DDThh:mm) in the local time zone of departure or arrival. For example, "2016-07-03T08:06"</li></ul> |
```
nilsnolde, line 44: isochrones can also be used for trafffic or time restrictions

docs/docs/api/isochrone/api-reference.md
```diff
@@ -65,10 +65,6 @@ When making a map, drawing the isochrone contours as lines is more straightforwa
 
 The Valhalla team has plans to improving the polygon isochrone output and rendering capabilities, including by demoting some rings to be inners of other rings and removing potential self-intersections in polygon geometries.
 
-## Isochrone demonstration in Mobility Explorer
```
nilsnolde, line 68: mobility explorer seems to be archived

docs/docs/api/isochrone/api-reference.md
```diff
@@ -77,9 +73,9 @@ Several other options are being considered as future service enhancements. These
 
 * ~~Using distance rather than time for each unit.~~
 * Generating outer contours or contours with interior holes for regions that cannot be accessed within the specified time, including with options to control the minimum size of interior holes.
-* Removing self intersections from polygonal contours.
-* Allowing multiple locations to compute the region reachable from any of the locations within a specified time.
-* Generating contours with reverse access logic to see the region that can reach a specific location within the specified time.
+* ~~Removing self intersections from polygonal contours.~~
+* ~~Allowing multiple locations to compute the region reachable from any of the locations within a specified time.~~
+* ~~Generating contours with reverse access logic to see the region that can reach a specific location within the specified time.~~
```
nilsnolde, line 78: looks like a few things were crossed off the list over the years :)

docs/docs/api/matrix/api-reference.md
```diff
@@ -59,7 +59,7 @@ The Time-Distance Matrix service uses the `auto`, `bicycle`, `pedestrian` and `b
 | :------------------ | :----------- |
 | `id` | Name your matrix request. If `id` is specified, the naming will be sent thru to the response. |
 | `matrix_locations` | For one-to-many or many-to-one requests this specifies the minimum number of locations that satisfy the request. However, when specified, this option allows a partial result to be returned. This is basically equivalent to "find the closest/best `matrix_locations` locations out of the full location set". |
-| `date_time` | This is the local date and time at the location.<ul><li>`type`<ul><li>0 - Current departure time.</li><li>1 - Specified departure time</li><li>2 - Specified arrival time. Not yet implemented for multimodal costing method.</li></ul></li><li>`value` - the date and time is specified in ISO 8601 format (YYYY-MM-DDThh:mm) in the local time zone of departure or arrival.  For example "2016-07-03T08:06"</li></ul><br>|
```
nilsnolde, line 62: there's no transit for matrix yet anyways

src/mjolnir/ingest_transit.cc
```diff
@@ -570,8 +570,8 @@ void write_routes(Transit& tile, const tile_transit_info_t& tile_info, feed_cach
     auto currRoute = feed.get_route(tile_routeId);
 
     route->set_name(currRoute->route_short_name);
-    route->set_onestop_id(currRoute->route_id);
-    route->set_operated_by_onestop_id(currRoute->agency_id);
+    route->set_onestop_id(feed_route.first + "_" + currRoute->route_id);
+    route->set_operated_by_onestop_id(feed_route.first + "_" + currRoute->agency_id);
```
nilsnolde, line 574: I was about to describe the way we determine onestop_ids for the filtering, and realized we don't do that yet. But I also kinda forgot that we should actually add it in #4032 . shouldn't have it added it here, but now it happened  :sweat_smile:

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -225,15 +225,8 @@ These options are available for transit costing when the multimodal costing mode
...
-| `filters` | A way to filter for one or more `stops`, `routes`, or `operators`. Filters must contain a list of [Onestop IDs](https://transit.land/documentation/onestop-id-scheme/), which is a unique identifier for Transitland data, and an `action`. <ul><li>`ids`: any number of Onestop IDs (such as o-9q9-bart)</li><li>`action`: either `exclude` to exclude all of the `ids` listed in the filter or `include` to include only the `ids` listed in the filter</li></ul>
+| `filters` | A way to filter for one or more ~~`stops`~~ (TODO: need to re-enable), `routes`, or `operators`. Filters must contain a list of so-called Onestop IDs, which is (supposed to be) a unique identifier for GTFS data, and an `action`. The OneStop ID is simply the feeds's directory name and the object's GTFS ID separated by an underscore, i.e. a route with `route_id: AUR` in `routes.txt` from the feed `NYC` would have the OneStop ID `NYC_AUR`, similar with operators/agencies. <ul><li>`ids`: any number of Onestop IDs</li><li>`action`: either `exclude` to exclude all of the `ids` listed in the filter or `include` to include only the `ids` listed in the filter</li></ul>
```
nilsnolde, line 228: yeah so remember how I was saying we don't use the onestop ids for stops when filtering.. you were right, they are supposed to be used but we lost that along the way in `convert_transit` and should re-enable I guess eventually. not that super high prio IMO.

src/mjolnir/ingest_transit.cc
```diff
@@ -614,8 +614,9 @@ void write_routes(Transit& tile, const tile_transit_info_t& tile_info, feed_cach
     auto currRoute = feed.get_route(tile_routeId);
 
     route->set_name(currRoute->route_short_name);
-    route->set_onestop_id(currRoute->route_id);
-    route->set_operated_by_onestop_id(currRoute->agency_id);
+    route->set_onestop_id(get_onestop_id_base(currRoute->route_id, feed_route.first.feed));
+    route->set_operated_by_onestop_id(
+        get_onestop_id_base(currRoute->agency_id, feed_route.first.feed));
```
nilsnolde, line 619: so the docs match the code:)

docs/docs/api/isochrone/api-reference.md
kevinkreiser: true tholugh i think self intersections is not one, you can still use a generalization configuration that will cause self intersection iirc

docs/docs/api/matrix/api-reference.md
kevinkreiser: yep says that in bold above which is good enough. this here was copy pasta

docs/docs/api/turn-by-turn/api-reference.md
kevinkreiser: its an important feature, we need to at least add it to the list of things to do for transit support

proto/common.proto
```diff
@@ -130,13 +130,13 @@ message Location {
 }
 
 message TransitEgressInfo {
-  string onestop_id = 1;        // Unique transitland ID
+  string onestop_id = 1;        // Unique ID
```
kevinkreiser, line 133: imho we should rename this when we do the work of coming up with our predictable id

kevinkreiser (approved):

PR #4040, opened 2023-03-25
Remove Arch-Type from strip command

### LINKED ISSUE #3950: docker image fails building for arm64
Was raised on the osm dev list (and is more a hunch as of now, but makes sense I think):

It might be not be possible to build our image on e.g. Apple Silicon. The error is caused by the name of the python bindings when we try to `strip` it:

https://github.com/valhalla/valhalla/blob/3fac9705794f902dceb9b031d04c2f4fa9ada109/Dockerfile#L45

I'd guess it builds actually `python_valhalla.cpython-310-darwin.so` or so.

homtec:
Fixes #3950

# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * **Removed**
    * REMOVED: Docker image pushes to Dockerhub [#4033](https://github.com/valhalla/valhalla/pull/4033)
 * **Bug Fix**
+   * FIXED: docker image fails building for arm64 [#3950](https://github.com/valhalla/valhalla/issues/3950)
    * FIXED: underflow of uint64_t cast for matrix time results [#3906](https://github.com/valhalla/valhalla/pull/3906)
    * FIXED: update vcpkg commit for Azure pipelines to fix libtool mirrors [#3915](https://github.com/valhalla/valhalla/pull/3915)
    * FIXED: fix CHANGELOG release year (2022->2023) [#3927](https://github.com/valhalla/valhalla/pull/3927)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 * **Removed**
    * REMOVED: Docker image pushes to Dockerhub [#4033](https://github.com/valhalla/valhalla/pull/4033)
 * **Bug Fix**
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -42,7 +42,7 @@ RUN tar -cvf valhalla.debug.tar valhalla_*.debug && gzip -9 valhalla.debug.tar
 RUN rm -f valhalla_*.debug
 RUN strip --strip-debug --strip-unneeded valhalla_* || true
 RUN strip /usr/local/lib/libvalhalla.a
-RUN strip /usr/lib/python3/dist-packages/valhalla/python_valhalla.cpython-310-x86_64-linux-gnu.so
+RUN strip /usr/lib/python3/dist-packages/valhalla/python_valhalla.cpython-310-*-gnu.so
```
nilsnolde, line 45: tbh, not so sure about the `-gnu` either, but should be fine? anyways, better than before for sure.

nilsnolde (commented):
Thanks @homtec ! Can you add a CHANGELOG entry?

Dockerfile
kevinkreiser: i agree, we can probably just do to be sure of ourselves

```suggestion
RUN strip /usr/lib/python3/dist-packages/valhalla/python_valhalla*.so
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 * **Removed**
    * REMOVED: Docker image pushes to Dockerhub [#4033](https://github.com/valhalla/valhalla/pull/4033)
 * **Bug Fix**
+   * FIXED: docker image fails building for arm64 [#3950](https://github.com/valhalla/valhalla/issues/3950)
```
nilsnolde, line 5: we organize our changelog in a date-descending manner, i.e. newest comes last. also we refer to the pull request, not the issue.

nilsnolde (changes requested):

nilsnolde (approved):
thx:)

CHANGELOG.md
nilsnolde: np;)

nilsnolde (approved):

kevinkreiser (approved):

PR #4042, opened 2023-03-27
throw on duplicated names for recostings

### LINKED ISSUE #4041: `recostings` can produce kinda invalid JSON
When multiple `recostings` for the same costing model are requested, the resulting `summary` objects have duplicated keys, e.g. for 2 `auto` recostings, it'd contain 2 `time_auto` keys. Against my own inuition, it's not directly _invalid_ apparently according to [this, quoting the spec](https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object). 

But consuming that JSON with JS or Python (or even `jq`) will lead to a last-one-wins situation, and in Java it seems to actually throw. I only discovered this because of `insomnia`, which thankfully is according to JSON spec (also surprisingly, it's an Electron app..). But `rapidjson` must allow it, it's probably even tested..

EDIT: quickly skimming the tests, it's actually not testing that situation. But I did learn that there's a `name` property one can pass with each `recosting` (hmpf, it's not documented at all..), which makes it unique with user input, e.g. `{"costing": "auto", "name": "1"}` will produce a `summary.time_1` key. So, either we:
1. throw if any of the names are dups, or there's no names for multiple recostings with the same costing
2. we make the output fields unique ourselves, e.g. with the index of a recosting within the `recostings` array

I strongly prefer 1. because it's more explicit and easy to handle for a client.

E.g. a request like 

```shell
curl -s --request POST \
  --url https://valhalla1.openstreetmap.de/route \
  --data '{
        "costing": "auto",
        "narrative": false,
        "id": 1,
        "locations": [
                {
                        "lon": 11.291481,
                        "lat": 52.500908
                },
                {
                        "lon": 10.899534,
                        "lat": 50.851427
                }
        ],
        "recostings": [
                {
                        "costing": "auto",
                        "fixed_speed": 20
                },
                {
                        "costing": "auto",
                        "fixed_speed": 50
                }
        ]
}'
``` 

would return a `summary` object of

```json
{
  "cost": 10775.333,
  "has_ferry": false,
  "has_highway": false,
  "has_time_restrictions": false,
  "has_toll": false,
  "length": 245.308,
  "max_lat": 52.54152,
  "max_lon": 11.553142,
  "min_lat": 50.852282,
  "min_lon": 10.889726,
  "time": 10114.933,			
  "time_auto": 44258.226,
  "time_auto": 17764.945
}
```

nilsnolde:
fixes #4041 

Chose to throw when there'd be dups.

### DIFF
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -68,7 +68,8 @@ const std::unordered_map<unsigned, valhalla::valhalla_exception_t> error_codes{
     {124, {124, "No edge/node costing provided", 400, HTTP_400, OSRM_INVALID_OPTIONS, "costing_required"}},
     {125, {125, "No costing method found", 400, HTTP_400, OSRM_INVALID_OPTIONS, "wrong_costing"}},
     {126, {126, "No shape provided", 400, HTTP_400, OSRM_INVALID_OPTIONS, "shape_required"}},
-    {127, {127, "Recostings require both name and costing parameters", 400, HTTP_400, OSRM_INVALID_OPTIONS, "recosting_parse_failed"}},
+    {127, {127, "Recostings require a valid costing parameter", 400, HTTP_400, OSRM_INVALID_OPTIONS, "recosting_parse_failed"}},
+    {128, {128, "Recostings require a unique 'name' field for each recosting", 400, HTTP_400, OSRM_INVALID_OPTIONS, "no_recosting_duplicate_names"}},
     {130, {130, "Failed to parse location", 400, HTTP_400, OSRM_INVALID_VALUE, "location_parse_failed"}},
     {131, {131, "Failed to parse source", 400, HTTP_400, OSRM_INVALID_VALUE, "source_parse_failed"}},
     {132, {132, "Failed to parse target", 400, HTTP_400, OSRM_INVALID_VALUE, "target_parse_failed"}},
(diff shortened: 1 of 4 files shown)


src/worker.cc
```diff
@@ -937,12 +938,16 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
   // parse any named costings for re-costing a given path
   auto recostings = rapidjson::get_child_optional(doc, "/recostings");
   if (recostings && recostings->IsArray()) {
+    // make sure we only have unique recosting names in the end
+    std::unordered_set<std::string> names;
```
kevinkreiser, line 942: this is incredibly nitpicky feel free to ignore:
```suggestion
    std::unordered_set<std::string> names;
    names.reserve(recostings->GetArray().Size());
```

nilsnolde: haha it's fine:)

test/gurka/test_recost.cc
```diff
@@ -520,10 +532,11 @@ TEST(recosting, api) {
                           std::to_string(map.nodes["7"].lng()) + R"(,"lat":)" +
                           std::to_string(map.nodes["7"].lat()) + "}";
 
-  // lets also do a bunch of costings
+  // lets also do a bunch of costings, note that one recosting omits the optional "name" parameter
+  // in which case it'll take the costing name as "name" field
   Api api;
   auto json = actor.route(R"({"costing":"auto","locations":[)" + locations + R"(],"recostings":[
-      {"costing":"auto","name":"same"},
```
kevinkreiser, line 526: just a note. i know that the base costing doesnt serialize its name in the attributes that come out but if you do name it the same as a recosting it is kind of strange.

kevinkreiser (approved):

nilsnolde:
I don’t trust codecov anymore.. why is it reporting from inside test files? Doesn’t make much sense to me.. Often it also seems to be pretty erratic.. anyways, I know it’s not easy to do.

nilsnolde:
Or is it a config issue to some degree?

PR #4045, opened 2023-03-29
bus=permit doesnt not override access=no

kevinkreiser:
fixes #4044

it was noticed that `access=no` wins over `highway=busway`. that does not seem to make sense so we probably have to fix up the lua. it is likely this tagging is meant to tell us no one is allow on it... except buses.

EDIT:

as per usual i am thorougly confused after reading the osm wiki. it is clear that overriding access with `bus=yes` is what is normally used, but its not clear that `highway=busway` should function the same. the way our interpretation seems to work is we get default access from the `highway` tag then we turn it off or let it on depending on what the `access` tag says, then finally we let that access for various modes get overriden by things like `bus=yes` or `bicycle=yes` etc.

so im thinking perhaps this particular bridge is a tagging issue. adding `bus=yes` to it allows the test pass.

EDIT:

lua was missing overrides for `bus=permit` and `taxi=permit` on ways. node access was also missing and for motorcycle as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
    * FIXED: duplicated recosting names should throw [#4042](https://github.com/valhalla/valhalla/pull/4042)
    * FIXED: Remove arch specificity from strip command of Python bindings to make it more compatible with other archs [#4040](https://github.com/valhalla/valhalla/pull/4040)
    * FIXED: GraphReader::GetShortcut no longer returns false positives or false negatives [#4019](https://github.com/valhalla/valhalla/pull/4019)
+   * FIXED: Tagging with bus=permit or taxi=permit did not override access=no [#4045](https://github.com/valhalla/valhalla/pull/4045)
 * **Enhancement**
    * CHANGED: replace boost::optional with C++17's std::optional where possible [#3890](https://github.com/valhalla/valhalla/pull/3890)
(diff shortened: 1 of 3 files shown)


harto:
OK, I think that could explain it, since I believe the code change in #3413 was at some point working as intended with various busways around Brisbane.

I noticed that the bridge (and the various other parts of the Brisbane busways) are tagged `bus=permit`. Are you saying that changing that to `bus=yes` would likely solve the issue?

(I tried searching the web for more information about `bus=permit` but didn't have much luck finding anything...)

kevinkreiser:
no this discussion absolutely belongs here.

https://wiki.openstreetmap.org/wiki/Tag:access%3Dpermit

seems like for our purposes `permit` is equivalent to `yes` would you agree @gknisely @nilsnolde  @dnesbitt61  the wiki basically says it's just there to tell you you can use it but you need to get an easily obtained permit

nilsnolde:
jep totally agreed

kevinkreiser:
ok cool, since we agree ill add it to this pr!

kevinkreiser:
ah i see we already had `permit` on many other modes, just forgot bus and taxi

nilsnolde (approved):

gknisely (approved):

kevinkreiser:
something is wrong with github reporting build statuses here. the windows build succeeded. so did the changelog validation.. im just going to merge rather than try to rebuild

harto:
Thank you for the quick fix! 🙏

PR #4046, opened 2023-03-29
docker tweaks

kevinkreiser:
~trying to get the packages thing to show up on the valhalla repo and not just at the org level~ fixed that with a check box on the repo landing page

the changes here are just meta data that get displayed on the pages github page


nilsnolde (approved):
even though it worked already, why not:)

Issue #4049, opened 2023-03-30
predictedspeeds test loads misaligned address


ianthetechie:
I'm not sure if this is a problem affecting "real" code paths or if it's just a funky test fixture, but I noticed this when running the sanitizer. Screenshot below (sorry, I love my IDE :D).

Relevant source permalinks (though the first one links to a comment for some reason; probably the next line):
* https://github.com/valhalla/valhalla/blob/4d9d106608b354d60f4d77b4e3501d6692bbe514/src/baldr/predictedspeeds.cc#L76
* https://github.com/valhalla/valhalla/blob/4d9d106608b354d60f4d77b4e3501d6692bbe514/test/predictedspeeds.cc#L211

<img width="1624" alt="image" src="https://user-images.githubusercontent.com/2140440/228761524-1dbd5556-953e-4f2b-860f-4d4837ecab80.png">

PR #4051, opened 2023-03-30
Upgrade RapidJSON

### LINKED ISSUE #4050: UB in RapidJSON
Ok, this one actually have a solution for, so bear with me ;)

Running the reproducible tile build test (ironically) triggers undefined behavior in RapidJSON.

RapidJSON has not had a release since 2016; I guess they just YOLO rolling releases now?

Upgrading the RapidJSON submodule to the latest commit (083f359f5c36198accc2b9360ce1e32a333231d9) _does_ fix the behavior, and all the tests pass. I'm not sure if there's a better solution available (different library? Is this the de facto standard in modern C++ codebases?), but it does appear to work, all tests pass, etc.

I'll open a PR for discussion.

ianthetechie:
# Issue

Fixes #4050.

## Tasklist

 - [ ] Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
    * FIXED: Remove arch specificity from strip command of Python bindings to make it more compatible with other archs [#4040](https://github.com/valhalla/valhalla/pull/4040)
    * FIXED: GraphReader::GetShortcut no longer returns false positives or false negatives [#4019](https://github.com/valhalla/valhalla/pull/4019)
    * FIXED: Tagging with bus=permit or taxi=permit did not override access=no [#4045](https://github.com/valhalla/valhalla/pull/4045)
+   * FIXED: Upgrade RapidJSON to address undefined behavior [#4051](https://github.com/valhalla/valhalla/pull/4051)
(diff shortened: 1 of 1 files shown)


ianthetechie:
This does not currently add any new tests, but it does fix UB in an existing test. We are currently permissive in our sanitized runs and do not throw an error.

I would suggest building with `-fno-sanitize-recover` eventually as the "test case" after https://github.com/valhalla/valhalla/issues/4049 is addressed.

nilsnolde:
Always a tough decision when a project isn't releasing anymore.. 7 years and counting.. Though the project is far from being dead apparently. @mloskot [summarizes it well](https://github.com/Tencent/rapidjson/issues/1006#issuecomment-1417332011) :smile: 

Since there's maintenance going on, I'd personally be in favor of upgrading to the latest commit, seeing how we're vendoring it anyways. Curiously they only seem to test windows though, never seen that on a C++ repo haha

mloskot (approved):
Apart from the generic "ain't broke, don't touch", I don't see any reason to not to upgrade

PR #4052, opened 2023-03-30
fixed time handling for transit service

nilsnolde:
Adjusted the transit service time things to always be relative to our own pivot date (01.01.2014), after a quick session with @kevinkreiser (thanks for that!)

The tests did pass before, but that was entirely random. I also tried a few feeds, but again, must've been entirely random successes.


kevinkreiser (approved):
no problem on the session, pairing on code is always more fun than review and planning!

nilsnolde:
Sorry, need to resolve changelog conflict. Can you push the button again @kevinkreiser ?

kevinkreiser:
build was fine before and this doesnt need to build again so i just merged it

PR #4054, opened 2023-03-30
make hierarchies robust to transit connection edges

nilsnolde:
Was pretty trivial in the end.

### DIFF
--- a/src/mjolnir/hierarchybuilder.cc
+++ b/src/mjolnir/hierarchybuilder.cc
@@ -308,8 +308,7 @@ void FormTilesInNewLevel(GraphReader& reader,
       // Do we need to force adding edgeinfo (opposing edge could have diff names)?
       // If end node is in the same tile and there is no opposing edge with matching
       // edge_info_offset).
-      bool diff_names = directededge->use() != Use::kTransitConnection &&
-                        directededge->endnode().tileid() == base_edge_id.tileid() &&
+      bool diff_names = directededge->endnode().tile_value() == base_edge_id.tile_value() &&
                         !OpposingEdgeInfoMatches(tile, directededge);
 
       // Get edge info, shape, and names from the old tile and add to the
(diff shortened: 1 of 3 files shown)


src/mjolnir/hierarchybuilder.cc
```diff
@@ -308,7 +308,8 @@ void FormTilesInNewLevel(GraphReader& reader,
       // Do we need to force adding edgeinfo (opposing edge could have diff names)?
       // If end node is in the same tile and there is no opposing edge with matching
       // edge_info_offset).
-      bool diff_names = directededge->endnode().tileid() == base_edge_id.tileid() &&
+      bool diff_names = directededge->use() != Use::kTransitConnection &&
```
nilsnolde, line 311: The `OpposingEdgeInfoMatches` was checking the `tile` (note, on local hierarchy!) for the `directededge`'s `endnode` which can be a transit node and then it fails..

I'm often getting confused if `tileid()` also comprises the level or not. But well, it doesn't..

nilsnolde: Any transit connection (currently? might change that?) has only one name for both directions anyways (AFAICT).

src/mjolnir/hierarchybuilder.cc
```diff
@@ -308,7 +308,7 @@ void FormTilesInNewLevel(GraphReader& reader,
       // Do we need to force adding edgeinfo (opposing edge could have diff names)?
       // If end node is in the same tile and there is no opposing edge with matching
       // edge_info_offset).
-      bool diff_names = directededge->endnode().tileid() == base_edge_id.tileid() &&
+      bool diff_names = directededge->endnode().tile_value() == base_edge_id.tile_value() &&
```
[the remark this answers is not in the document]
nilsnolde: I'm wondering why `tile_value` isn't used more often, seems to be safer in many situations than `tile_id` generally (and more explicit).

nilsnolde:
let's close this one, I have the same changes in #4055

nilsnolde:
Haha I promise we'll get there again after this GTFS hell hole :laughing:

PR #4062, opened 2023-04-06
add has_transit_tiles and osm_changeset ID to verbose /status response

### LINKED ISSUE #4053: Return more info on /status
I think these would be helpful:
- is there _any_ transit loaded?
- latest OSM changeset. Caveat: more complicated if we ever allow for incremental tile builds (also a few other /status members though), but until then it'd be useful to have.

nilsnolde:
fixes #4053 

Is that small enough?;)

I'll add a changelog later to see if the CI script works again.


test/loki_service.cc
```diff
@@ -94,90 +94,90 @@ const std::vector<http_request_t> valhalla_requests{
...
-     R"({"error_code":153,"error":"Too many shape points: (102). The best paths shape limit is 100","status_code":400,"status":"Bad Request"})"},
-};
+const std::vector<std::pair<uint16_t, std::string>>
+    valhalla_responses{
+        {200,
+         R"({"version":")" VALHALLA_VERSION
+         R"(","tileset_last_modified":0,"available_actions":["status","centroid","expansion","transit_available","trace_attributes","trace_route","isochrone","optimized_route","sources_to_targets","height","route","locate"]})"},
+        {200,
+         R"({"version":")" VALHALLA_VERSION
+         R"(","tileset_last_modified":0,"available_actions":["status","centroid","expansion","transit_available","trace_attributes","trace_route","isochrone","optimized_route","sources_to_targets","height","route","locate"],"has_tiles":false,"has_admins":false,"has_timezones":false,"has_live_traffic":false,"has_transit_tiles":false,"osm_changeset":0,"bbox":{"features":[],"type":"FeatureCollection"}})"},
```
nilsnolde, line 104: oops.. not that small after all.. obviously I only changed this line

nilsnolde:
Jep, re-generating the circleci token seems to have worked @kevinkreiser

proto/status.proto
```diff
@@ -18,11 +19,13 @@ message Status {
   oneof has_bbox {
     string bbox = 5;
   }
-  oneof has_version {
-    string version = 6;
+  string version = 6;
+  uint32 tileset_last_modified = 7;
+  repeated string available_actions = 8;
```
nilsnolde, line 24: these didn't need a `oneof` actually

src/loki/status_action.cc
```diff
@@ -64,6 +68,10 @@ void loki_worker_t::status(Api& request) const {
   status->set_has_admins(tile && tile->header()->admincount() > 0);
   status->set_has_timezones(tile && tile->node(0)->timezone() > 0);
   status->set_has_live_traffic(reader->HasLiveTraffic());
+
+  status->set_has_transit_tiles(filesystem::exists(transit_dir) &&
+                                filesystem::is_directory(transit_dir));
+  status->set_osm_changeset(tile && tile->header()->dataset_id());
```
nilsnolde, line 74: either connectivity_map or get_tileset and cache the result. problem is for tarred data

src/loki/status_action.cc
```diff
@@ -55,15 +57,23 @@ void loki_worker_t::status(Api& request) const {
...
+  // TODO: get this from the connectivity_map once transit is implemented there
 
   if (connectivity_map) {
     status->set_bbox(connectivity_map->to_geojson(2));
+    const bool has_transit_tiles =
+        connectivity_map->level_color_exists(TileHierarchy::GetTransitLevel().level);
+    status->set_has_transit_tiles(has_transit_tiles);
+  } else {
+    const static bool has_transit_tiles = !reader->GetTileSet(3).empty();
+    status->set_has_transit_tiles(has_transit_tiles);
```
nilsnolde, line 69: if we have a connectivity map, great, we'll use the newly added transit level in there via a new tiny function

src/loki/status_action.cc
```diff
@@ -55,15 +57,23 @@ void loki_worker_t::status(Api& request) const {
 
   // get _some_ tile
   const static baldr::graph_tile_ptr tile = get_graphtile(reader);
+  // TODO: get this from the connectivity_map once transit is implemented there
```
nilsnolde, line 60: remove this, it's done

src/loki/status_action.cc
```diff
@@ -55,15 +57,23 @@ void loki_worker_t::status(Api& request) const {
...
+  } else {
+    const static bool has_transit_tiles = !reader->GetTileSet(3).empty();
+    status->set_has_transit_tiles(has_transit_tiles);
   }
 
   status->set_has_tiles(static_cast<bool>(tile));
   status->set_has_admins(tile && tile->header()->admincount() > 0);
   status->set_has_timezones(tile && tile->node(0)->timezone() > 0);
   status->set_has_live_traffic(reader->HasLiveTraffic());
+  status->set_osm_changeset(tile && tile->header()->dataset_id());
```
kevinkreiser, line 76: this is no good. the result is a boolean

proto/status.proto
kevinkreiser: neither do the new ones

src/loki/status_action.cc
nilsnolde: argh, good catch!

test/loki_service.cc
nilsnolde: wouldn't show in the test of course..

proto/status.proto
nilsnolde: Ah not @kevinkreiser they're needed, just as the other ones which are only available via `"verbose": true`: in the serializer we're using their `has_` method to check if you should set it in the output JSON or leave it out. I guess that wouldn't necessarily need to be handled with `oneof`s, one could also look at the `verbose()` in the serializer, but that way it'd be tightly coupled to `verbose()` while now we're more flexible and could disallow including them in the JSON based on other logic.

src/loki/status_action.cc
```diff
@@ -58,12 +60,19 @@ void loki_worker_t::status(Api& request) const {
...
+  } else {
+    const static bool has_transit_tiles = !reader->GetTileSet(3).empty();
+    status->set_has_transit_tiles(has_transit_tiles);
   }
 
   status->set_has_tiles(static_cast<bool>(tile));
   status->set_has_admins(tile && tile->header()->admincount() > 0);
   status->set_has_timezones(tile && tile->node(0)->timezone() > 0);
   status->set_has_live_traffic(reader->HasLiveTraffic());
+  status->set_osm_changeset(tile ? tile->header()->dataset_id() : 0);
```
nilsnolde, line 75: we set the changeset here, but don't set it in the serializer's output JSON if it was 0

src/tyr/serializers.cc
```diff
@@ -120,6 +120,14 @@ std::string serializeStatus(Api& request) {
   if (request.status().has_has_live_traffic_case())
     status_doc.AddMember("has_live_traffic",
                          rapidjson::Value().SetBool(request.status().has_live_traffic()), alloc);
+  if (request.status().has_has_transit_tiles_case())
+    status_doc.AddMember("has_transit_tiles",
+                         rapidjson::Value().SetBool(request.status().has_transit_tiles()), alloc);
+  // a 0 changeset indicates there's none, so don't write in the output
+  // TODO: currently this can't be tested as gurka isn't adding changeset IDs to OSM objects (yet)
+  if (request.status().has_osm_changeset_case() && request.status().osm_changeset())
+    status_doc.AddMember("osm_changeset",
```
nilsnolde, line 129: since we only set the changeset when there was one, this code path isn't tested. gurka doesn't add changeset IDs to objects. I left a todo

kevinkreiser (approved):

PR #4066, opened 2023-04-10
Update vcpkg & FindLuaJIT.cmake to find Windows artifacts

nilsnolde:
`libiconv` wasn't building with the old commit on a new Windows system. It does build when pointing to the most recent commit. Couple of related questions:

- should we just point at master instead? This is happening regularly, likely Windows/MSVC updates causing that and we don't catch in CI because it's caching dependencies
- ~~the build finished fine, both Debug & Release, but I can't execute anything.. I get `The program '[12004] valhalla_build_tiles.exe' has exited with code -1073741515 (0xc0000135)` right.. Apparently missing _some_ `.dll`, but which?! CI executes graph builds fine for a Release build. Since the build finishes properly I guess I'm missing some Windows runtime `.dll`. I installed VS build tools on top of VS Code (+ the relevant extensions), which should include the relevant C++ runtime environment, right? Do you have maybe any idea where to start debugging @mloskot ? (I know it's hard to suggest anything, but is there any way that it tells which DLL I'm missing?)~~

### DIFF
--- a/cmake/FindLuaJIT.cmake
+++ b/cmake/FindLuaJIT.cmake
@@ -10,7 +10,7 @@
 find_path(LUA_INCLUDE_DIR luajit.h
   HINTS
     ENV LUA_DIR
-  PATH_SUFFIXES include/luajit-2.0 include/luajit-2.1 include
+  PATH_SUFFIXES include/luajit-2.0 include/luajit-2.1 include/luajit include
   PATHS
   ~/Library/Frameworks
   /Library/Frameworks
@@ -21,7 +21,7 @@ find_path(LUA_INCLUDE_DIR luajit.h
 )
 
 find_library(LUA_LIBRARY
-  NAMES luajit-5.1
+  NAMES luajit-5.1 lua51
   HINTS
     ENV LUA_DIR
   PATH_SUFFIXES lib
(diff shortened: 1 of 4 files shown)


nilsnolde:
Oh boy.. Forget about the execution problems.. If I do `.\build\Release\valhalla_build_tiles.exe --config .\test\win\valhalla.json .\test\data\utrecht_netherlands.osm.pbf` it works fine.. 

I kinda remember that crap: the JSON `valhalla_build_config` spits out is apparently not working for Windows/Valhalla combo.. It's just a `json.dumps`. Does that ring a bell? Even copy/pasting the working `./test/win/valhalla.json` into the JSON file doesn't work. 

The only thing that works is creating a new valhalla.json file and pasting the contents. But not with the Python generated file from `valhalla_build_config` WTF?! Anyways, happy it works again..

nilsnolde:
The way I created the valhalla.json initially was with `python .\scripts\valhalla_build_config > .\site\valhalla.json`. That's valid PowerShell though right? Something about that fails `valhalla_build_tiles` silently..

cmake/FindLuaJIT.cmake
```diff
@@ -10,7 +10,7 @@
 find_path(LUA_INCLUDE_DIR luajit.h
   HINTS
     ENV LUA_DIR
-  PATH_SUFFIXES include/luajit-2.0 include/luajit-2.1 include
+  PATH_SUFFIXES include/luajit-2.0 include/luajit-2.1 include/luajit include
```
nilsnolde, line 13: also decided to finally take care of this. previously it wouldn't find luajit and we'd need to tell cmake where to find it explicitly. adjusted all documentation and CI stuff accordingly.

kevinkreiser (dismissed):

kevinkreiser (approved):

Issue #4018, opened 2023-03-09
Adjusting code to C++17


cvvergara:
Default c++ version was set to C++17 on #3000

# Work to be done:

The main objective for the moment: **Remove warnings**
While working on the current main objective,  when encountered, also do the following:
* Use std::array instead of C style array
* Use range loops
* Wrap existing unused code with `#if 0 ... #endif` 
  * Alternatively remove that code
* Maybe more things found

# What to be adjusted
(maybe I am missing a directory)

## src (high priority)
* [ ] baldr
* [ ] loki
* [ ] meili
* [ ] midgard
* [ ] mjolnir
* [ ] odin
* [ ] sif
* [ ] skadi
* [ ] thor
* [ ] tyr
* [ ] local files

## test (low priority)
* [ ] local files
* [ ] gurka

kevinkreiser:
i know you've already made tons of progress on this, but im finding it quite difficult/tedious to review. is there any way we could break it up even further? the codebase is big so tackling modernizations and warnings simultanously makes the scope of a given pr (even when just in one namespace) quite large. would it be possible to decuople the warnings work from the modernization cleanup? thanks again for working on this, its the stuff that always gets deprioritized so we really apreciate you taking the time on it.

nikhil061102:
Hello @kevinkreiser sir
I am Nikhil and eager to contribute to world of open source. I am very enthusiastic about this and would like to work on this issue with you. We can divide the work amongst ourselves and finish the job. This will also be a good experience for me as I want to apply in GSoC 2023 this year :)

PR #4068, opened 2023-04-12
Issue 4018 odin

cvvergara:
# Issue
baldr part of #4018

## Tasklist

 - [x] Removal of warnings
 - [x] Not compiling functions with ambiguity with proto functions
 - [x] Avoid future ambiguities using `namespace detail`
 - [ ] Decide: 
   * Remove code of functions with ambiguity with proto functions
   * Use the the functions defined in `namespace detail`

## Requirements / Relations


cvvergara:
@kevinkreiser 
This is almost done.
There are some functions that I marked with `#if 0` because they are ambiguous with some functions with proto, and the compiler is using the ones defined in proto.
My opinion is to delete the ambiguous functions basically they do the same thing, but in proto: vector is been used, and the ambiguous functions use the unordered map.

So, please let me know what to do.

PR #4069, opened 2023-04-12
enable manual docker images

nilsnolde:
It'd be cool to have manually triggered docker image builds. Could be useful for:
- debugging purposes
- for previews of features during development
- easier for us to do previews with our gis-ops valhalla image :innocent:


.github/workflows/docker_build_pr.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Log in to GitHub Docker Registry
+      uses: docker/login-action@v2
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build latest image
+      run: |
+        docker build -t ghcr.io/valhalla/valhalla:manually_triggered_build .
```
nilsnolde, line 30: The image name will be `manually_triggered_build` and will naturally be overwritten by the next build. Personally I don't think the manual build will be used much, so that shouldn't negatively interfere IMO. Quite opposite actually, it cleans up after itself. If we'd name it after the branch it's originating from I'd fear we wouldn't be very diligent deleting stuff when unused. 

I'll likely use it now and then for client previews (e.g. the recent GTFS one, now the traffic matrix), but that's like once a month.

Issue #4038, opened 2023-03-23
time-dependent cost matrix


nilsnolde:
Since we implemented #3792 , the need arose from other clients to have CostMatrix care about time as well. That'd enable the faster matrix algo to respect any time-dependent attributes, such as traffic.

Caveat: CostMatrix is a bidirectional algorithm, hence full time support **in costing isn't possible**, i.e. only one direction can track time during the expansion: for a `depart_at` request it's the forward branches, for `arrive_by` it's the reverse branches. That also triggers another important caveat: for the direction which isn't tracking time, we can't be sure if we're not actually expanding over closed roads at that time, e.g. due to time restrictions in OSM (there'd be a convoluted way to deal with that, but meeh..).

~~In fact, looking at the `recost` algorithm: what happens when there's live traffic and at the time of recosting one road segment of the final path is closed all of a sudden? Can happen if the recosting happens at a different live traffic interval than the expansion (i.e. different speed values, it's not that unlikely), or if the time-unaware expansion direction went over an edge which is actually closed. I can't see code that's catching that case so it seems it'd create a gigantic ETA/cost? But maybe it's handled somewhere in the tripedgebuilder or so? Or not handled (yet, possibly only not in the public repo)? Just a side question, probably for @kevinkreiser :)~~ jep, it's handled in `GetSpeed` I was told.

However, we can still compute the correct ETA by recosting the "path". (note to self: obviously there's no path (yet), need to generate one probably when there's time requested)

So, to be crystal clear: in terms of optimality, it'll only be half optimal wrt time/traffic, but the final ETA will be as accurate as it can be for the final path.

It's quite a bit more complicated in general than TimeDistanceMatrix and there's conceptual questions on how to handle `date_time` strings on sources/targets for the many:many case. In the 1:many it was easy, as we only expand from source _or_ target, but now we do both at the same time, and right now that breaks my head somewhat. Will get back to that when I gathered my thoughts on it.

nilsnolde:
During implementing this, I realized we should (for now) follow the same approach as bidir A* and only allow/respect time on origins, i.e. only depart_at/current date_time.

Both bidir A* & matrix could in the future be updated to allow arrive_by as well, but that'd require quite a few changes, yet very doable I think.

Issue #3920, opened 2023-01-19
Visualize matrix expansion & paths


nilsnolde:
It'd be great to visualize the matrix expansion. Obviously it could be a crazy payload, but I wouldn't expect to ever debug more than 2-3 locations, everything after that is only a conceptual repetition I think. So in worst case, it's like 3 times the isochrone's expansion output for the one:many algo.

We could also hide a debug function in a trace block or so, which can trace back the chosen path (so it's not dependent on a `/expansion`  request). I'd find that invaluable to validate and compare with our other algos, which I'd love to do at some point.

kevinkreiser:
i always wanted the expansion stuff to be available from all relevant algorithms just didnt have time to do it. as you say it can be very useful for low level debugging without stepping through the debugger

PR #4071, opened 2023-04-13
add traffic to CostMatrix

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4038

Extending the work done in https://github.com/valhalla/valhalla/pull/3795, I added time awareness to the CostMatrix algo as well.

A few high-level things to mention:
- currently we **only** support adding `depart_at`/`current` time spec, i.e. only on origin locations. I left a TODO for `arrive_by` which could be done (also for bidir A*), but not a priority for me at least and also quite a bit more work/complexity
- the whole decision model in `thor/matrix_action` changed a bit. It still preserves the previous way we decided on the matrix algo, so nothing will change for Valhalla deployments which don't have any time info
- unless `prioritize_bidirectional: true` is set (and of course plus a valid time on the sources/targets), we'll favor the more accurate (albeit much slower) TimeDistance matrix algo
- to trigger a CostMatrix with time awareness:
  - add `"prioritize_bidirectional": true` in the root of the request
  - add at least one `date_time` string to an origin location or set the global `date_time.type = 1/0`


nilsnolde:
I'll leave it here for now. Next week I'll be on vacation, I'll comment more after. Feel free to have a glance of course:)

src/sif/recost.cc
```diff
@@ -98,9 +98,9 @@ void recost_forward(baldr::GraphReader& reader,
     // evaluate time restrictions
     const auto next_id = edge_cb();
     if (predecessor != baldr::kInvalidLabel &&
-        (!costing.Allowed(edge, !next_id.Is_Valid(), label, tile, edge_id, localtime,
-                          offset_time.timezone_index, time_restrictions_TODO) &&
-         !ignore_access)) {
+        (!ignore_access &&
+         !costing.Allowed(edge, !next_id.Is_Valid(), label, tile, edge_id, localtime,
+                          offset_time.timezone_index, time_restrictions_TODO))) {
```
nilsnolde, line 103: tiny nit: just changed around the order of calls, ignore_accees is much cheaper

src/thor/bidirectional_astar.cc
```diff
@@ -1276,6 +1276,9 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
     }
 
     // recost edges in final path; ignore access restrictions
+    // TODO: actually we should not ignore access restrictions: if the reverse path
+    //   traversed a closed edge due to time restrictions, we could do a mini traversal
+    //   to circumvent the closed edge(s)
```
nilsnolde, line 1281: it's not the nicest patch, but still nicer than navigating over an edge we know is closed

src/thor/costmatrix.cc
```diff
@@ -108,12 +117,14 @@ void CostMatrix::clear() {
...
-    GraphReader& graphreader,
+    google::protobuf::RepeatedPtrField<valhalla::Location>& source_location_list,
+    google::protobuf::RepeatedPtrField<valhalla::Location>& target_location_list,
+    baldr::GraphReader& graphreader,
     const sif::mode_costing_t& mode_costing,
-    const travel_mode_t mode,
-    const float max_matrix_distance) {
+    const sif::travel_mode_t mode,
+    const float max_matrix_distance,
+    const bool has_time,
```
nilsnolde, line 126: we tell it when there's a valid time. we compute it in `matrix_action` where we already check if any time is valid

src/thor/costmatrix.cc
```diff
@@ -890,5 +930,111 @@ void CostMatrix::SetTargets(baldr::GraphReader& graphreader,
...
+        path_edges.emplace_back(std::move(opp_edge_id));
+    }
+
+    auto edge_itr = path_edges.begin();
+    const auto edge_cb = [&edge_itr, &path_edges]() {
+      return (edge_itr == path_edges.end()) ? GraphId{} : (*edge_itr++);
+    };
+
+    Cost new_cost{0.f, 0.f};
+    const auto label_cb = [&new_cost](const EdgeLabel& label) { new_cost = label.cost(); };
```
nilsnolde, line 1010: we only need to update the cost when recosting the path, not path necessary. once we want to tackle https://github.com/valhalla/valhalla/issues/3920, we can do it similar to bidir a*

src/thor/costmatrix.cc
```diff
@@ -890,5 +930,111 @@ void CostMatrix::SetTargets(baldr::GraphReader& graphreader,
...
+        recovered_inner_edges.insert(superseded.begin() + 1, superseded.end());
+        std::move(superseded.begin(), superseded.end(), std::back_inserter(path_edges));
+      } else
+        path_edges.emplace_back(std::move(opp_edge_id));
+    }
+
+    auto edge_itr = path_edges.begin();
+    const auto edge_cb = [&edge_itr, &path_edges]() {
+      return (edge_itr == path_edges.end()) ? GraphId{} : (*edge_itr++);
+    };
```
nilsnolde, line 1007: same as bidir a*, only the label callback changes

src/thor/matrix_action.cc
```diff
@@ -14,39 +14,6 @@ using namespace valhalla::baldr;
 using namespace valhalla::sif;
 using namespace valhalla::thor;
 
-namespace {
-
-// return true if any location had a time set
-// also disable time if it doesn't make sense computationally
-bool has_time(Api& request) {
```
nilsnolde, line 21: changed it to `check_matrix_time` and moved it to `matrix_common.h`

src/thor/matrix_action.cc
```diff
@@ -93,33 +63,40 @@ std::string thor_worker_t::matrix(Api& request) {
...
       break;
   }
-  return tyr::serializeMatrix(request, time_distances, distance_scale);
+
+  // similar to routing: prefer the exact unidirectional algo if not requested otherwise
+  // don't use matrix_type, we only need it to set the right warnings for what will be used
+  bool has_time =
+      check_matrix_time(request,
+                        options.prioritize_bidirectional() ? MatrixType::Cost : MatrixType::TimeDist);
+  if (has_time && !options.prioritize_bidirectional() && !force_costmatrix) {
```
nilsnolde, line 91: do a timedistance matrix, no matter the costing model, if:
- there's time
- `prioritize_bidirectional` is false
- the server doesn't have `COST_MATRIX` defined

src/thor/matrix_action.cc
```diff
@@ -93,33 +63,40 @@ std::string thor_worker_t::matrix(Api& request) {
...
+    return tyr::serializeMatrix(request, timedistancematrix(), distance_scale, MatrixType::TimeDist);
+  } else if (has_time && options.prioritize_bidirectional()) {
```
nilsnolde, line 93: otherwise, if there's a valid time AND `prioritize_bidirectional` is true, we do a CostMatrix

src/thor/matrix_action.cc
```diff
@@ -93,33 +63,40 @@ std::string thor_worker_t::matrix(Api& request) {
...
+    return tyr::serializeMatrix(request, costmatrix(has_time), distance_scale, MatrixType::Cost);
+  } else if (matrix_type == MatrixType::Cost) {
+    return tyr::serializeMatrix(request, costmatrix(has_time), distance_scale, matrix_type);
+  } else {
+    return tyr::serializeMatrix(request, timedistancematrix(), distance_scale, matrix_type);
```
nilsnolde, line 98: these two are the same as before

src/thor/optimized_route_action.cc
```diff
@@ -28,9 +28,11 @@ void thor_worker_t::optimized_route(Api& request) {
...
   // Use CostMatrix to find costs from each location to every other location
   CostMatrix costmatrix;
-  std::vector<thor::TimeDistance> td =
-      costmatrix.SourceToTarget(options.sources(), options.targets(), *reader, mode_costing, mode,
-                                max_matrix_distance.find(costing)->second);
+  std::vector<TimeDistance> td =
+      costmatrix.SourceToTarget(*options.mutable_sources(), *options.mutable_targets(), *reader,
+                                mode_costing, mode, max_matrix_distance.find(costing)->second,
+                                check_matrix_time(request, MatrixType::Cost),
+                                options.date_time_type() == Options::invariant);
```
nilsnolde, line 35: also enable time awareness for `/optimized`

src/thor/timedistancematrix.cc
```diff
@@ -22,37 +22,6 @@ static bool IsTrivial(const uint64_t& edgeid,
   }
   return false;
 }
-
-// Will return a destination's date_time string
-std::string GetDateTime(const std::string& origin_dt,
```
nilsnolde, line 27: moved this to `matrix_common.h` as well so we can use it in both

src/tyr/matrix_serializer.cc
```diff
@@ -2,7 +2,7 @@
 
 #include "baldr/json.h"
 #include "proto_conversions.h"
-#include "thor/costmatrix.h"
+#include "thor/matrix_common.h"
```
nilsnolde, line 5: I moved most common things, which were actually previously defined in `costmatrix.h` to `matrix_common.h` as the import pattern was a bit weird otherwise (timedistance.h matrix would import costmatrix.h..)

src/tyr/matrix_serializer.cc
```diff
@@ -71,6 +72,8 @@ json::MapPtr serialize(const Api& request,
   }
   json->emplace("durations", time);
   json->emplace("distances", distance);
+  json->emplace("algorithm",
+                std::string(matrix_type == MatrixType::Cost ? "costmatrix" : "timedistancematrix"));
```
nilsnolde, line 76: add a `algorithm` field to the response

valhalla/thor/costmatrix.h
```diff
@@ -27,24 +29,6 @@ constexpr float kCostThresholdBicycleDivisor =
     56.0f; // 200 km distance threshold will result in a cost threshold of ~3600 (1 hour)
 constexpr float kCostThresholdPedestrianDivisor =
     28.0f; // 200 km distance threshold will result in a cost threshold of ~7200 (2 hours)
-constexpr float kMaxCost = 99999999.9999f;
-
-// Time and Distance structure
-struct TimeDistance {
```
nilsnolde, line 33: moved these to `matrix_common.h` as well

valhalla/thor/matrix_common.h
```diff
@@ -15,15 +15,18 @@
 #include <valhalla/sif/dynamiccost.h>
 #include <valhalla/sif/edgelabel.h>
 #include <valhalla/thor/astarheuristic.h>
-#include <valhalla/thor/costmatrix.h>
```
nilsnolde, line 18: this was strange :smile:

src/thor/matrix_action.cc
```diff
@@ -93,33 +63,40 @@ std::string thor_worker_t::matrix(Api& request) {
         case travel_mode_t::kBicycle:
           // Use CostMatrix if number of sources and number of targets
           // exceeds some threshold
-          if (options.sources().size() > kCostMatrixThreshold &&
-              options.targets().size() > kCostMatrixThreshold) {
-            time_distances = costmatrix();
-          } else {
-            time_distances = timedistancematrix();
+          if (options.sources().size() < kCostMatrixThreshold ||
+              options.targets().size() < kCostMatrixThreshold) {
```
nilsnolde, line 67: <= instead of <

src/thor/costmatrix.cc
```diff
@@ -890,5 +929,111 @@ void CostMatrix::SetTargets(baldr::GraphReader& graphreader,
   }
 }
 
+// Form the path from the adjacency list.
+void CostMatrix::RecostPaths(GraphReader& graphreader,
```
kevinkreiser, line 933: it would be cool to move the inner loop of recosting a single src and tgt to a common area like path info or somethign else that makes sense so that this and bidir a* can share it

valhalla/thor/costmatrix.h
```diff
@@ -247,6 +241,42 @@ class CostMatrix {
...
+  /**
+   * Sets the date_time on the origin locations.
+   *
+   * @param origins            the origins (sources or targets)
+   * @param reader             the reader for looking up timezone information
+   * @returns                  time info for each location
+   */
+  std::vector<baldr::TimeInfo>
+  SetOriginTimes(google::protobuf::RepeatedPtrField<valhalla::Location>& origins,
+                 baldr::GraphReader& reader) {
```
nilsnolde, line 269: make it const function

nilsnolde: move to anon namespace in the src

nilsnolde: ah I put it in the header bcs it needs access to `tz_cache_` member.

src/thor/costmatrix.cc
nilsnolde: left a todo

nilsnolde:
Addressed the review comments from yesterday @kevinkreiser :

- https://github.com/valhalla/valhalla/pull/4071/commits/9419e4beb8c322e3b17da4549de2455722dccf82: re-arranged the decision model for which algo to use in `matrix_action` a little
- https://github.com/valhalla/valhalla/pull/4071/commits/ada29d708ead4e1d48f24312c5a3ade654ea683d: Also added warnings if the config forces the algorithm but the request im-/explicitly suggested a different algorithm.
- https://github.com/valhalla/valhalla/pull/4071/commits/49ff04b3049d38cac669a80e4d4ca147ab29cc9a: minor things like leaving a todo & changing `<` to `<=` in one place

nilsnolde:
There was a minor conflict in costmatrix.cc. After CI is done, this is good to merge from my side @kevinkreiser :)

kevinkreiser:
looks like setsources is missing the timeinfo arg

nilsnolde:
Yeah there were a few code merge conflicts and I shouldn't have tried solving those in the Github UI, it made a bit of a mess. That function was called twice, old & new way, so it threw.

I fixed it.

src/thor/costmatrix.cc
```diff
@@ -750,7 +787,13 @@ void CostMatrix::BackwardSearch(const uint32_t index, GraphReader& graphreader)
 // Sets the source/origin locations. Search expands forward from these
 // locations.
 void CostMatrix::SetSources(GraphReader& graphreader,
-                            const google::protobuf::RepeatedPtrField<valhalla::Location>& sources) {
+                            const google::protobuf::RepeatedPtrField<valhalla::Location>& sources,
+                            const std::vector<baldr::TimeInfo>& time_infos) {
+  // Allocate edge labels and edge status
+  source_count_ = sources.size();
+  source_edgelabel_.resize(source_count_);
+  source_edgestatus_.resize(source_count_);
```
kevinkreiser, line 795: arent these now redundant with `Initialize`?

kevinkreiser (approved):

kevinkreiser (approved):

PR #4075, opened 2023-04-15
clear matrix after request and reserve some space

### LINKED ISSUE #4064: Matrix should clear temp structures, but pre-reserve some space
The matrix algorithms don't reserve edge labels after they're done with one request, instead they're shrinking the structures to 0. Also seems like we should `reuse()` the adjacency lists.

I'd even keep more labels than the routing algos, 1 Mio is not a lot for the Dijkstras IMO (it's "just" a few MBs per thread). Either e.g. double the `max_reserved_labels_count` config for the matrix algos or introduce a new config? What do you think @dnesbitt61 @kevinkreiser ?

nilsnolde:
fixes #4064 

- clears all matrix algorithms' structures after a computation
- introduces `max_reserved_labels_count_matrix` as that's quite a lot more demanding than A*
- clean up some of the `kInitialEdgeLabelCount` as that seemed fairly wild :)


src/thor/bidirectional_astar.cc
```diff
@@ -15,8 +15,6 @@ using namespace valhalla::sif;
 
 namespace {
 
-constexpr uint32_t kInitialEdgeLabelCountBD = 1000000;
```
nilsnolde, line 18: I moved them all to `edgelabel.h`

src/thor/costmatrix.cc
```diff
@@ -224,20 +228,50 @@ std::vector<TimeDistance> CostMatrix::SourceToTarget(
...
-  const uint32_t kMaxThreshold = std::numeric_limits<int>::max();
+  source_count_ = source_locations.size();
+  target_count_ = target_locations.size();
+
+  // Add initial sources status
+  source_status_.reserve(source_count_);
+  source_hierarchy_limits_.reserve(source_count_);
+  source_adjacency_.reserve(source_count_);
+  source_edgestatus_.resize(source_count_);
+  source_edgelabel_.resize(source_count_);
```
nilsnolde, line 239: pulled all initializations here

src/thor/costmatrix.cc
```diff
@@ -224,20 +228,50 @@ std::vector<TimeDistance> CostMatrix::SourceToTarget(
...
   for (uint32_t i = 0; i < source_count_; i++) {
+    // Allocate the adjacency list and hierarchy limits for this source.
+    // Use the cost threshold to size the adjacency list.
+    source_edgelabel_[i].reserve(
+        std::min(max_reserved_labels_count_, kInitialEdgeLabelCountBDMatrix));
```
nilsnolde, line 244: this is the main change

src/thor/matrix_action.cc
```diff
@@ -66,15 +66,21 @@ std::string thor_worker_t::matrix(Api& request) {
   // lambdas to do the real work
   std::vector<TimeDistance> time_distances;
   auto costmatrix = [&]() {
-    return costmatrix_.SourceToTarget(options.sources(), options.targets(), *reader, mode_costing,
-                                      mode, max_matrix_distance.find(costing)->second);
+    time_distances =
+        costmatrix_.SourceToTarget(options.sources(), options.targets(), *reader, mode_costing, mode,
+                                   max_matrix_distance.find(costing)->second);
+    costmatrix_.clear();
```
nilsnolde, line 72: clear the algos

valhalla/sif/edgelabel.h
```diff
@@ -12,6 +12,12 @@
 namespace valhalla {
 namespace sif {
 
+constexpr uint32_t kInitialEdgeLabelCount = 1000000;
+constexpr uint32_t kInitialEdgeLabelCountMatrix = 4000000;
+constexpr uint32_t kInitialEdgeLabelCountBD = 500000;
+constexpr uint32_t kInitialEdgeLabelCountBDMatrix = 2000000;
+constexpr uint32_t kInitialEdgeLabelCountMM = 200000;
```
nilsnolde, line 19: gather them all here

nilsnolde:
There's tons of ambiguous statements now in the tests, which I don't really understand where it's coming from..

valhalla/sif/edgelabel.h
kevinkreiser: having them all here is great but for the ones that we changed we shouldnt lower them with this pr just for the sake of avoiding performance issues hidden with the change. its ok to increase some but especially bidir astar we should set back to its higher limit

src/thor/matrix_action.cc
kevinkreiser: we want to do these over in worker.cc in the clear() method that way it happens not in the request path and rather only after returning the response to the requestor

src/worker.cc
```diff
@@ -282,7 +282,7 @@ void parse_location(valhalla::Location* location,
     location->set_type(valhalla::Location::kVia);
   } // other actions let you specify whatever type of stop you want
   else if (stop_type_json) {
-    Location::Type type = Location::kBreak;
+    valhalla::Location::Type type = valhalla::Location::kBreak;
```
nilsnolde, line 285: remove those

valhalla/thor/timedistancematrix.h
```diff
@@ -120,8 +122,6 @@ class TimeDistanceMatrix {
    * Reset all origin-specific information
    */
   inline void reset() {
-    // Clear the edge labels and destination list
-    edgelabels_.clear();
```
nilsnolde, line 124: review this again, maybe we want to clear the labels?

nilsnolde: this is being called at the end of every for loop where we expand from each origin. so we want to clear here, but we also need to reserve some space in the beginning of that for loop in the cc file.

valhalla/sif/edgelabel.h
nilsnolde: done

src/thor/matrix_action.cc
nilsnolde: done

src/thor/astar_bss.cc
```diff
@@ -84,7 +82,7 @@ void AStarBSSAlgorithm::Init(const midgard::PointLL& origll, const midgard::Poin
   // Reserve size for edge labels - do this here rather than in constructor so
   // to limit how much extra memory is used for persistent objects.
   // TODO - reserve based on estimate based on distance and route type.
-  edgelabels_.reserve(std::min(kInitialEdgeLabelCount, max_reserved_labels_count_));
+  edgelabels_.reserve(max_reserved_labels_count_);
```
nilsnolde, line 85: @kevinkreiser I checked in the PR that introduced this config option and I can't see any reasoning why it's not allowing to get past the hard-coded constants using `std::min`. I'd prefer to change it to respect the config option "blindly". WDYT?

nilsnolde: here the PR: https://github.com/valhalla/valhalla/pull/2719/files#diff-e7d2f8280a1f34d1175d125309288c693ea745ff14f88c3565378c77ed907bdd

kevinkreiser (approved):

danceb:
Just to clarify for me:
The paramters `max_reserved_labels_count_matrix` and `clear_reserved_memory_matrix` from https://github.com/valhalla/valhalla/pull/4075/commits/76b33d882cc3e6421a07d72d4e1e2f793c5ba080#diff-b3846614879a01a4589cc7be405a97326b3c79e8a9e64944d4f94a41c5d478e6
didn't made it into the master branch: https://github.com/valhalla/valhalla/commit/894659334e433ed334d001e24e2482d9b62593aa#diff-b3846614879a01a4589cc7be405a97326b3c79e8a9e64944d4f94a41c5d478e6.
Is there a reason for? Like is the memory cleaning now made in another way? 
Because I came across this issue as others, and there is no memory cleaning with valhalla 3.4.0 in my case.

nilsnolde:
Not sure what you're diffing there, but yes it's in master:) https://github.com/valhalla/valhalla/blob/05ac3423c9484362dcb4ef70854d5c18a6ecb9e1/scripts/valhalla_build_config#L116-L119.

In fact, there was never a real problem with clearing RAM after requests. Quite the opposite: it cleared way too much, impeding performance. Now it's reserving a lot more RAM after a request. That shouldn't be more than 50 MB or so per thread. What you're describing sounds like you're running valhalla on a plain tile directory instead of the .tar archive. Then each thread will have its own tile cache (default 1 GB). There's been a lot of confusion around that in the past (also myself).

danceb:
> What you're describing sounds like you're running valhalla on a plain tile directory instead of the .tar archive. Then each thread will have its own tile cache (default 1 GB)

That exactly sound like my issue. I am using 4 threads, the first matrix calculation loads around 4 GB into memory and the second calculation another 4 GB on top. 
But actually to my configuration the tar file should be used instead of the plain tile files. But this is another thing and nothing for this threet here.

PR #4082, opened 2023-04-22
added transit level to connectivity map

nilsnolde:
To compare colors for transit, we look at level 3 colors but within `transit_start_end_max_distance` meters of correlated edges, which are still missing the transit connection edges. We wanted to eventually add support for  that for multiple reasons anyways: start the journey on such an edge and also to return departure info for `/locate`. For the connectivity map it's not so bad, at least with the default `transit_start_end_max_distance`.

Also lets us determine whether transit tiles are loaded with less resources than currently in #4062.

### DIFF
--- a/src/loki/route_action.cc
+++ b/src/loki/route_action.cc
@@ -72,7 +72,7 @@ void loki_worker_t::route(Api& request) {
   check_hierarchy_distance(request);
 
   auto& ped_opts = *options.mutable_costings()->find(Costing::pedestrian)->second.mutable_options();
-  auto connectivity_level = TileHierarchy::levels().back().level;
+  auto connectivity_level = TileHierarchy::levels().back();
   uint32_t connectivity_radius = 0;
   // Validate walking distances (make sure they are in the accepted range)
   if (costing_name == "multimodal" || costing_name == "transit") {
@@ -94,7 +94,7 @@ void loki_worker_t::route(Api& request) {
       throw valhalla_exception_t{156, " Min: " + std::to_string(min_transit_walking_dis) + " Max: " +
                                           std::to_string(max_transit_walking_dis) + " (Meters)"};
     }
-    connectivity_level = TileHierarchy::GetTransitLevel().level;
+    connectivity_level = TileHierarchy::GetTransitLevel();
     connectivity_radius = ped_opts.transit_start_end_max_distance();
   }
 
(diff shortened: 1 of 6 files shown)


src/loki/route_action.cc
```diff
@@ -102,12 +106,10 @@ void loki_worker_t::route(Api& request) {
     for (size_t i = 0; i < locations.size(); ++i) {
       const auto& correlated = projections.at(locations[i]);
       PathLocation::toPBF(correlated, options.mutable_locations(i), *reader);
-      // TODO: get transit level for transit costing
-      // TODO: if transit send a non zero radius
       if (!connectivity_map) {
         continue;
       }
-      auto colors = connectivity_map->get_colors(TileHierarchy::levels().back().level, correlated, 0);
+      auto colors = connectivity_map->get_colors(connectivity_level, correlated, connectivity_radius);
```
nilsnolde, line 112: pass the full `TileLevel`, we need access to level 2 or 3 tiles and need to know which level it was

test/graphreader.cc
```diff
@@ -78,52 +78,60 @@ TEST(ConnectivityMap, Basic) {
...
-  EXPECT_EQ(conn.get_color({a0, 2, 0}), conn.get_color({a1, 2, 0})) << "a's should be connected";
-  EXPECT_EQ(conn.get_color({a0, 2, 0}), conn.get_color({a2, 2, 0})) << "a's should be connected";
-  EXPECT_EQ(conn.get_color({a1, 2, 0}), conn.get_color({a2, 2, 0})) << "a's should be connected";
-  EXPECT_EQ(conn.get_color({d0, 2, 0}), conn.get_color({d1, 2, 0})) << "d's should be connected";
-  EXPECT_NE(conn.get_color({c0, 2, 0}), conn.get_color({a1, 2, 0})) << "c is disjoint";
-  EXPECT_NE(conn.get_color({b0, 2, 0}), conn.get_color({a0, 2, 0})) << "b is disjoint";
-  EXPECT_NE(conn.get_color({a2, 2, 0}), conn.get_color({d0, 2, 0})) << "a is disjoint from d";
-
-  filesystem::remove_all(tile_dir);
+  for (const auto& level : {TileHierarchy::levels()[2], TileHierarchy::GetTransitLevel()}) {
```
nilsnolde, line 81: just added the for loop, nothing else

nilsnolde: ah and passing the level to the `touch_tile` function

kevinkreiser (approved):

PR #4084, opened 2023-04-25
dont remove sources for people just installing deps on their own

kevinkreiser:
our install docs say to use the linux deps scripts but it removes all sources in ubuntu. this makes sense in docker i guess but for general use it will screw up peoples systems


nilsnolde (approved):
Oops, good catch!

nilsnolde (approved):
:ship:

PR #4089, opened 2023-04-27
fix test when day is past 27

nilsnolde:
when the day is past 27 (could be 28 really), it would create a month string of e.g. `5` instead of `05`, so the test would fail.

### DIFF
--- a/test/gurka/test_gtfs.cc
+++ b/test/gurka/test_gtfs.cc
@@ -862,6 +862,7 @@ TEST(GtfsExample, route_trip1) {
   // wrap around the next month if it's getting critical
   if (tmrw_int > 27) {
     auto new_month = std::to_string(std::stoul(std::string(&req_time[5], &req_time[7])) + 1);
+    new_month = std::string(2 - std::min(2UL, new_month.length()), '0') + new_month;
     req_time.replace(req_time.find('T') - 5, 2, new_month);
     tmrw_int -= 20;
   }


test/gurka/test_gtfs.cc
```diff
@@ -862,6 +862,7 @@ TEST(GtfsExample, route_trip1) {
   // wrap around the next month if it's getting critical
   if (tmrw_int > 27) {
     auto new_month = std::to_string(std::stoul(std::string(&req_time[5], &req_time[7])) + 1);
+    new_month = std::string(2 - std::min(2UL, new_month.length()), '0') + new_month;
```
nilsnolde, line 865: fill the month string with a leading zero

test/gurka/test_gtfs.cc
```diff
@@ -862,6 +862,7 @@ TEST(GtfsExample, route_trip1) {
   // wrap around the next month if it's getting critical
   if (tmrw_int > 27) {
     auto new_month = std::to_string(std::stoul(std::string(&req_time[5], &req_time[7])) + 1);
```
kevinkreiser, line 864: and when its december? shouldnt we let the date time library handle this. it has support for addition and subtraction but requires a little bit more work and temporary values of course..

kevinkreiser (approved):
this will work for now, but when its december its going to  be very hard to build the test in the 13th month of the year :smile:

test/gurka/test_gtfs.cc
nilsnolde: argh.. right.. I'll write an issue about it. I'm sure I'll get to it before 2024;)

nilsnolde:
feel free to force push (no need for changelog IMO). I left an issue to fix this within the next 7 months:)

PR #4099, opened 2023-05-10
fail early when building admins

nilsnolde:
I was having a wrong path when building admins and it took minutes before it failed, because it couldn't create the parent dir for the sqlite file:
```
2023/05/10 08:51:22.538025 [INFO] Parsing files: routing_admins_test.pbf
2023/05/10 08:51:22.538080 [INFO] Parsing relations...
2023/05/10 08:52:03.124591 [INFO] Finished with 68 admin polygons comprised of 1039 ways
2023/05/10 08:52:03.124706 [INFO] Parsing ways...
2023/05/10 08:52:58.579642 [INFO] Finished with 1039 ways comprised of 1504341 nodes
2023/05/10 08:52:58.579678 [INFO] Parsing nodes...
2023/05/10 08:52:59.116406 [INFO] Processed 500000 nodes on ways
2023/05/10 08:53:40.567164 [INFO] Finished with 752148 nodes
2023/05/10 08:53:40.572887 [INFO] Admin directory not found. Admins will not be created.
```

I changed 2 things:
- check the directory etc before actually parsing the OSM data
- fail the command when smth went wrong


src/mjolnir/adminbuilder.cc
```diff
@@ -422,7 +419,7 @@ void BuildAdminFromPBF(const boost::property_tree::ptree& pt,
     LOG_ERROR("Error: " + std::string(err_msg));
     sqlite3_free(err_msg);
     sqlite3_close(db_handle);
-    return;
+    return false;
```
kevinkreiser, line 422: the number of repetitions of this block of code has reached a sinful level. not saying you should fix it, but who ever did this should be flogged!

kevinkreiser (dismissed):
one tiny caveat, we have tests which use the function `BuildAdminFromPBF` should the return value be asserted, i dont feel strongly about it

nilsnolde:
> BuildAdminFromPBF should the return value be asserted

yeah why not, added it

src/mjolnir/adminbuilder.cc
nilsnolde: haha yeah same thought. left a todo

kevinkreiser (approved):

PR #4104, opened 2023-05-12
Filesystem with clang tidy problem naive fix

### LINKED ISSUE #4109: valhalla/filesystem.h  -> systems's filesystem
[filesystem](https://en.cppreference.com/w/cpp/filesystem) started from C++17, Before, Valhalla was using a c++ standard less than 17, but the functionality of `filesystem` was needed it has a "copy" on `valhalla/filesystem.h`.
With the change done recently (#3877) started using C++17, using the packaged filesystem is needed.

Now, on #4098 where files that use: `#include valahalla/filesystem.h` 
The following warning in cling-tidy is happening:
```
/home/circleci/project/valhalla/filesystem.h:214:7: note: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119
warning: argument unused during compilation: '-fuse-ld=lld' [clang-diagnostic-unused-command-line-argument]
/home/circleci/project/valhalla/filesystem.h:214:7: warning: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119 [clang-analyzer-security.insecureAPI.strcpy]
      strcpy(entry_->d_name, filename.c_str());
```

I suggest 2 steps for moving from the "copy" to the system's filesystem, that is:
`#include valahalla/filesystem.h` -> `#include <filesystem.h>`
 

- [ ] #4104 
  - This "quick fix" will remove the warnings from clang-tidy making it easier to detect changes generated from the next step
  - With this change, #4098 should not generate warnings from `valhalla/filesystem.h` with clang-tidy
- [ ] #4105   (To be changed to the final PR number)
  - Using actually the `#include <filesystem.h>`
  - Currently its a POC, as I noticed after I started making changes that the "copy" was modified adding some convenient functions for valhalla, but gives the idea of what needs to be changed
  - [ ] Create final PR and change number in this issue

cvvergara:
#4109 

## Tasklist

 - [x] Modify a file that uses "valhalla/filesystem.h"
 - [x] Change code to avoid warning
 - [x] Make things work
 - [x] Create the issue

## Requirements / Relations


valhalla/filesystem.h
```diff
@@ -211,7 +211,8 @@ class directory_entry {
       // entry_->d_reclen =
       // entry_->d_off =
       entry_->d_ino = s.st_ino;
-      strcpy(entry_->d_name, filename.c_str());
+      /* Saving a char* into a std::string */
+      filename = entry_->d_name;
```
kevinkreiser, line 215: FYI this is backwards we are trying to put the filename into the directory entries name field

cvvergara:
First commit  (#4106)  does this change [with clang tidy](https://app.circleci.com/pipelines/github/valhalla/valhalla/10667/workflows/5475998e-e90c-4cb6-9ce0-059ca794863b/jobs/57937?invite=true#step-116-11)
Not happening [now](https://app.circleci.com/pipelines/github/valhalla/valhalla/10669/workflows/ebae0163-2f76-4499-be7e-ecf75b8ae115/jobs/57942)

TODO: rebase before merge

nilsnolde (approved):

kevinkreiser:
i know this pr is simple but some things spring to mind:

1. we should really spell out what this pr is doing in the description of the pr. the linked issue has it but why not have it in the actual pr? im guessing its because of the stupid PR template? i'm going to delete that thing, it has about 0 use :smile: sorry for the confusion there
2. i see that we are trying to truncate the filename to fit it into the space restriction we have on dirent. did you notice though on the next line in the `ifdef` that we need to set the size of the name? if we truncate we cant use the original size. it is true though that `d_namlen` isnt used elsewhere in the code so we should probably just remove the other fields which may or may not be present depending on your platform, hopefully we'll move to std::filesystem soonish..
3. big picture.. i get that the compiler is telling us we should consider whether or not the strings contents will fit in `d_name`, its a good point, but we have to think about the consequences when that happens and what our work around will do. ive not seen any discussion on this in the pr which makes it difficult to review. here is my take:

`directory_entry` has essentially 2 use-cases:
1. a bunch of free functions which do `stat` operations on it
2. directory crawling

for the use-case where we ask `exists` `is_directory` etc truncating in the constructor wont matter because we throw the object away immediately after the stat operation so that is safe

in directory crawling, if you give it something that isnt a directory it will truncate the filename and then fail to iterate because it realized it wasnt a directory. so no harm there really the truncated value goes unused.

when it actually is a directory, the constructor returns early before looking at the filename part of the path so that is safe no truncation happens there (its a directory there is no file name part to truncate anyway).

this leaves me with the following conclusion, the only purpose of setting the dirent inside of `directory_entry`s constructor is so we can store the information from `stat` and use it in the free functions. when the actual directory crawling stuff happens we dont use this info at all, instead we get that from `readdir` directly.

knowing this i wonder if that constructor should look more like this:

```c++
  directory_entry(const filesystem::path& path, bool iterate)
      : dir_(nullptr), entry_(nullptr), path_(path) {
    // stat it first in case its not a directory
    struct stat s;
    if (stat(path_.c_str(), &s) == 0) {
      // if it is a directory and we are going to iterate over it
      if (S_ISDIR(s.st_mode) && iterate) {
        dir_.reset(opendir(path_.c_str()), [](DIR* d) {
          if (d)
            closedir(d);
        });
        return;
      }
      // keep basic stat info (inode and size), path_ has everything else we make use of
      entry_.reset(new dirent);
      entry_->d_ino = s.st_ino;
      entry_->d_type = mode_to_type(s.st_mode);
      file_size_ = s.st_size;
    }
  }
```

Issue #3942, opened 2023-01-30
Handle destonly ways/edges around ferry terminals


nilsnolde:
ref https://github.com/valhalla/valhalla/issues/3941#issuecomment-1407713742.

Even though it's more of a mapping error, it seems it's quite pervasively done when mapping access to/from ferries terminals. Our client did a quick analysis, and around 400 ferry terminals world-wide have `access=customers/private/destination` tagged on the land-side connection edge, so they'd not really be considered by the bidir A* on the first pass, leading to horrendously long detours.

Maybe we could think about handling this case better in whatever way? I know it's a slippery slope to start handling mapping errors, but in this case IMO it could make sense.

There's a few approaches that come to my mind, but only viable one I think. Ideally it's done on the data, not the request, level, e.g. when reclassifying ferry edges we could "untag" `access` for all the path edges we find whenever the ferry connection edge is `dest_only`, and we'd stop untagging as soon as the next expanded edge is not `dest_only` anymore (similar stuff to what we're doing with `not_thru` when enhancing the graph). That way we could make sure to not open up legit `dest_only` along the way to the next `primary` road. 

That still leaves all the ferry terminals kinda "unsupported" where the connection edge is not treated as `dest_only` but only the edge further "inward". We wouldn't catch that, but gotta stop at some point supporting non-sense;)

We can discuss briefly tmrw..

kevinkreiser:
this was the same thought I had. we already remove destonly when reclassifying (Kristen and I did that a while ago). to me we are doing this reclassification exactly to make sure the ferry stays connected so it doesn't make sense to waste that work by keeping it closed unless the tagging isn't likely

kevinkreiser:
(sorry about the close and reopen, accidently tapped the wrong button)

nilsnolde:
ok cool. should we then not care if the connection is dest_only or not and simply remove dest_only from all edges which are part of the final upclassed paths for all modes? 

on the one hand there could be legit dest_only which we don't want to touch but on the other hand we'd miss out on some cases where only the connection edge is not weirdly tagged but all others in the area are.

kevinkreiser:
@nilsnolde i really thought we did that already in #1905

nilsnolde:
[...]

PR #1905, opened 2019-08-08
Removed destination_only when reclassifying ferries

kdiluca:
# Issue

This PR fixes #1889 

## Tasklist

 - [ ] Add tests
 - [ ] Review - you must request approval to merge any PR to master
 - [ ] Add #fixes with the issue number that this PR addresses
 - [ ] Generally use squash merge to rebase and clean comments before merging
 - [ ] Update the [changelog](CHANGELOG.md)
 - [ ] Update relevant [documentation](docs/README.md)

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #4118, opened 2023-05-19
ignore destonly edges around ferries

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/3942

In the end #1905 was mostly missing to unset `destonly` on a ferry-connecting edge. I threw in some more bug fixes around the reclassification and a bit more logic to not penalize destonly edges if the ferry-connecting edge was destonly. If the ferry-connecting edge wasn't destonly, we will always penalize any destonly edges while expanding for reclassification.

### DIFF
--- a/src/mjolnir/directededgebuilder.cc
+++ b/src/mjolnir/directededgebuilder.cc
@@ -53,7 +53,7 @@ DirectedEdgeBuilder::DirectedEdgeBuilder(const OSMWay& way,
   set_truck_route(way.truck_route());
 
   // Set destination only to true if the remove_destonly is set to false and either destination only
-  // or no thru traffic is set. remove_destonly can be set to true when reclassifying ferries.
+  // or no thru traffic is set. remove_destonly is set for reclassified paths due to ferries.
   set_dest_only(!remove_destonly && (way.destination_only() || way.no_thru_traffic()));
   if (remove_destonly && (way.destination_only() || way.no_thru_traffic())) {
     LOG_DEBUG("Overriding dest_only attribution to false for ferry.");
--- a/src/mjolnir/directededgebuilder.cc
+++ b/src/mjolnir/directededgebuilder.cc
@@ -26,7 +26,7 @@ DirectedEdgeBuilder::DirectedEdgeBuilder(const OSMWay& way,
                                          const bool minor,
                                          const uint32_t restrictions,
                                          const uint32_t bike_network,
-                                         const bool reclass_ferry)
+                                         const bool remove_destonly)
     : DirectedEdge() {
   set_endnode(endnode);
   set_use(use);
@@ -52,11 +52,10 @@ DirectedEdgeBuilder::DirectedEdgeBuilder(const OSMWay& way,
 
   set_truck_route(way.truck_route());
 
-  // Set destination only to true if the reclass_ferry is set to false and either destination only or
-  // no thru traffic is set. Adding the reclass_ferry check allows us to know if we should override
-  // the destination only attribution
-  set_dest_only(!reclass_ferry && (way.destination_only() || way.no_thru_traffic()));
-  if (reclass_ferry && (way.destination_only() || way.no_thru_traffic())) {
+  // Set destination only to true if the remove_destonly is set to false and either destination only
+  // or no thru traffic is set. remove_destonly can be set to true when reclassifying ferries.
+  set_dest_only(!remove_destonly && (way.destination_only() || way.no_thru_traffic()));
+  if (remove_destonly && (way.destination_only() || way.no_thru_traffic())) {
     LOG_DEBUG("Overriding dest_only attribution to false for ferry.");
   }
   set_dismount(way.dismount());
--- a/src/mjolnir/ferry_connections.cc
+++ b/src/mjolnir/ferry_connections.cc
@@ -61,7 +61,9 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
   uint16_t overall_access_before = baldr::kVehicularAccess;
   uint16_t overall_access_after = 0;
   // find accessible paths for as many modes as we can but stop if we can't find any more
-  while (overall_access_before != overall_access_after) {
+  // or found all modes in the first round
+  while ((overall_access_after != baldr::kVehicularAccess) &&
+         (overall_access_before != overall_access_after)) {
     overall_access_before = overall_access_after;
 
     // which modes do we still have to find a path for?
@@ -80,7 +82,7 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
     // Add node to list of node labels, set the node status and add
     // to the adjacency set
     uint32_t nodelabel_index = 0;
-    node_labels.emplace_back(0.0f, node_idx, node_idx, first_edge_destonly, first_edge_destonly);
+    node_labels.emplace_back(0.0f, node_idx, node_idx, first_edge_destonly);
     node_status[node_idx] = {kTemporary, nodelabel_index};
     adjset.push({0.0f, nodelabel_index});
     nodelabel_index++;
@@ -89,6 +91,7 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
     // is reached
     uint32_t n = 0;
     uint32_t label_idx = 0;
+    uint32_t last_label_idx = 0;
     while (!adjset.empty()) {
       // Get the next node from the adjacency list/priority queue. Gets its
       // current cost and index
@@ -180,11 +183,11 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
           }
         }
 
+        // Get the last label index to form the path
+        last_label_idx = nodelabel_index;
+
         // Add to the node labels and adjacency set. Skip if this is a loop.
-        const bool remove_destonly =
-            first_edge_destonly ? (pred_destonly && w.destination_only()) : false;
-        node_labels.emplace_back(cost, endnode, expand_node_idx, w.destination_only(),
-                                 remove_destonly);
+        node_labels.emplace_back(cost, endnode, expand_node_idx, w.destination_only());
         node_status[endnode] = {kTemporary, nodelabel_index};
         adjset.push({cost, nodelabel_index});
         nodelabel_index++;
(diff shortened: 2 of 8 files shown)


valhalla/mjolnir/node_expander.h
```diff
@@ -54,12 +54,14 @@ struct Edge {
...
-    uint64_t turn_channel : 1;  // Link edge should be a turn channel
-    uint64_t way_begin : 1;     // True if first edge of way
-    uint64_t way_end : 1;       // True if last edge of way
-    uint64_t spare : 25;
+    uint64_t reclass_ferry : 1;   // Has edge been reclassified due to
+                                  // ferry connection
+    uint64_t turn_channel : 1;    // Link edge should be a turn channel
+    uint64_t way_begin : 1;       // True if first edge of way
+    uint64_t way_end : 1;         // True if last edge of way
+    uint64_t remove_destonly : 1; // During ferry reclassification we decide to remove dest_only
```
nilsnolde, line 62: add a property telling the directed edge builder to ignore destonly

valhalla/mjolnir/ferry_connections.h
```diff
@@ -32,14 +32,18 @@ constexpr uint32_t kPermanent = 1;
 // be "adjacent" to an node that is permanently labeled.
 constexpr uint32_t kTemporary = 2;
 
+constexpr uint32_t kFerryUpClass = static_cast<uint32_t>(baldr::RoadClass::kPrimary);
+
```
nilsnolde, line 36: the road class was already a constant, but was passed through the whole function chain only to end up where it's needed. rather make it a real constanct

src/mjolnir/graphbuilder.cc
```diff
@@ -1238,15 +1238,8 @@ void GraphBuilder::Build(const boost::property_tree::ptree& pt,
     LOG_WARN("Not reclassifying link graph edges");
   }
 
-  // Reclassify ferry connection edges - use the highway classification cutoff
-  baldr::RoadClass rc = baldr::RoadClass::kPrimary;
-  for (auto& level : TileHierarchy::levels()) {
-    if (level.name == "highway") {
-      rc = level.importance;
-    }
-  }
```
nilsnolde, line 1247: this was unnecessary, replaced by a constant in `ferry_connections.h`

valhalla/mjolnir/ferry_connections.h
```diff
@@ -32,14 +32,18 @@ constexpr uint32_t kPermanent = 1;
...
 // NodeLabel - for simple shortest path
 struct NodeLabel {
   float cost;
   uint32_t node_index;
   uint32_t pred_node_index;
+  bool dest_only;
+  bool remove_dest_only;
```
nilsnolde, line 43: two new properties for the label:
- `dest_only` is to keep track 
- `remove_dest_only` is only set if we can be sure we want to remove it eventually

src/mjolnir/ferry_connections.cc
```diff
@@ -42,7 +42,7 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
                       sequence<Edge>& edges,
                       sequence<Node>& nodes,
                       const bool inbound,
-                      const uint32_t rc) {
+                      const bool first_edge_destonly) {
```
nilsnolde, line 45: this is the boolean dictating if we want to remove destonly or not

src/mjolnir/ferry_connections.cc
```diff
@@ -162,9 +163,12 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
...
         }
 
-        // Get cost - need the length and speed of the edge; Use a penalty if an edge is
-        // destination_only and calculate it in the cost
-        float penalty = w.destination_only() ? 300 : 0;
+        // if the first edge was destonly and the previous edge was as well, add no penalty
+        float penalty =
+            (first_edge_destonly ? (!pred_destonly && w.destination_only()) : w.destination_only())
+                ? 300
+                : 0;
```
nilsnolde, line 170: add a penalty only if destonly should be removed AND the pred was not destonly, or if the connecting edge wasn't even destonly

src/mjolnir/ferry_connections.cc
```diff
@@ -177,7 +181,10 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
         }
 
         // Add to the node labels and adjacency set. Skip if this is a loop.
-        node_labels.emplace_back(cost, endnode, expand_node_idx);
+        const bool remove_destonly =
+            first_edge_destonly ? (pred_destonly && w.destination_only()) : false;
```
nilsnolde, line 185: similarly here: only remove destonly if the pred had it too. it's still a bit too broad, but can't it any other way I think.

src/mjolnir/ferry_connections.cc
```diff
@@ -337,23 +350,24 @@ void ReclassifyFerryConnections(const std::string& ways_file,
...
         }
         ferry_endpoint_count++;
 
         // Reclassify the first/start edge. Do this AFTER finding shortest path so
         // we do not immediately determine we hit the specified classification
         sequence<Edge>::iterator element = edges[edge.second];
         auto update_edge = *element;
-        update_edge.attributes.importance = rc;
+        update_edge.attributes.importance = kFerryUpClass;
+        update_edge.attributes.remove_destonly = remove_destonly;
```
nilsnolde, line 370: to solve the attached issue, it'd have been actually enough to say `update_edge.reclass_ferry = true` right here, which is what #1905 was missing. I still think the new logic, while a bit more complex, is more robust.

test/gurka/test_ferry_connections.cc
```diff
@@ -226,6 +226,78 @@ TEST_P(FerryTest, ReclassifyFerryConnectionPerMode) {
...
+TEST(Standalone, ReclassifyFerryUntagDestOnly) {
+  // access=customers should be untagged if it's present on connecting edge(s)
+  // see https://github.com/valhalla/valhalla/issues/3941#issuecomment-1407713742
+  // EF is destonly -> should untag DE, EF, but not BC and take the faster path
+  // GH is not destonly -> shouldn't untag any destonly and take the detour
+  const std::string ascii_map = R"(
+    A--B--C--D--E--F------G--H--I--J--K
+             |  |            |  |
+             L--M            N--O
+  )";
```
nilsnolde, line 238: we test both cases, where a connection does have destonly, and one where it doesn't.

src/mjolnir/directededgebuilder.cc
```diff
@@ -26,7 +26,7 @@ DirectedEdgeBuilder::DirectedEdgeBuilder(const OSMWay& way,
                                          const bool minor,
                                          const uint32_t restrictions,
                                          const uint32_t bike_network,
-                                         const bool reclass_ferry)
+                                         const bool remove_destonly)
```
nilsnolde, line 29: the whole change in this file is only renaming the arg and update the comment, no logic changed

src/mjolnir/ferry_connections.cc
```diff
@@ -61,7 +61,9 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
   uint16_t overall_access_before = baldr::kVehicularAccess;
   uint16_t overall_access_after = 0;
   // find accessible paths for as many modes as we can but stop if we can't find any more
-  while (overall_access_before != overall_access_after) {
+  // or found all modes in the first round
+  while ((overall_access_after != baldr::kVehicularAccess) &&
+         (overall_access_before != overall_access_after)) {
```
nilsnolde, line 66: tiny optimization: if we actually found all modes in the first expansion, we can stop. before it would do another round, only to find the exact same path.

src/mjolnir/ferry_connections.cc
```diff
@@ -89,13 +91,15 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
     // is reached
     uint32_t n = 0;
     uint32_t label_idx = 0;
+    uint32_t last_label_idx = 0;
```
nilsnolde, line 94: we use this to initialize the label when forming the path

src/mjolnir/ferry_connections.cc
```diff
@@ -196,8 +206,9 @@ uint32_t ShortestPath(const uint32_t start_node_idx,
     uint16_t path_access = baldr::kVehicularAccess;
     while (true) {
       // Get the edge between this node and the predecessor
-      uint32_t idx = node_labels[label_idx].node_index;
-      uint32_t pred_node = node_labels[label_idx].pred_node_index;
+      const NodeLabel& node_lab = node_labels[last_label_idx];
```
nilsnolde, line 209: there was bug before where we were taking `label_idx` which was left over from the expansion. however, `label_idx` could be pointing to some random stuff in the adjacency list, just before the expansion was finished, and that label wouldn't even be part of the path.. so here we initialized the path forming with the wrong label

src/mjolnir/ferry_connections.cc
```diff
@@ -337,23 +353,24 @@ void ReclassifyFerryConnections(const std::string& ways_file,
...
         }
         ferry_endpoint_count++;
 
         // Reclassify the first/start edge. Do this AFTER finding shortest path so
         // we do not immediately determine we hit the specified classification
         sequence<Edge>::iterator element = edges[edge.second];
         auto update_edge = *element;
-        update_edge.attributes.importance = rc;
+        update_edge.attributes.importance = kFerryUpClass;
+        update_edge.attributes.reclass_ferry = remove_destonly;
```
nilsnolde, line 373: #1905 was missing this: the connecting edge itself also needs its destonly removed if there was one.

test/gurka/test_ferry_connections.cc
```diff
@@ -226,6 +225,82 @@ TEST_P(FerryTest, ReclassifyFerryConnectionPerMode) {
...
+  // see https://github.com/valhalla/valhalla/issues/3941#issuecomment-1407713742
+  // CD is destonly -> should untag BC, CD and take the faster path
+  // EF is not destonly -> shouldn't untag FG and take the detour
+
+  // the X is bcs of https://github.com/valhalla/valhalla/issues/4164
+  const std::string ascii_map = R"(
+    A--B--C--D------E--F-X-G--H--M
+       |  |            |   |
+       I--J            K---L
+  )";
```
nilsnolde, line 239: this was VERY annoying to find out what the problem was.. see #4164

kevinkreiser (approved):

PR #4121, opened 2023-05-21
matrix pbf output

### LINKED ISSUE #3867: Matrix output in PBF
Also a way to "slim down" a matrix request as proposed in #3866 . But also generally good to have for protobuf consumers.

Side question: few others are missing as well in protobuf output. What else would make sense? I don't know off the top of my head which endpoints' output is included. IMO isochrones would also make sense, mostly only for general protobuf consumers, as the usual use case probably is displaying it via GeoJSON anyways.

### LINKED ISSUE #4153: Multiple sources to sources_to_targets only returns valid result for first source.
Something strange is happening to /sources_to_targets when giving multiple sources in combination with the bicycle profile. The first source is accurately compared to all the targets. However, the second source (even when giving the same source again) is (incorrectly) compared to only the first and second targets. The rest of the targets always return a distance and time of zero. 

Different sources
`curl -d '{"targets":[{"lon":5.1976415000000005,"lat":52.1787269},{"lon":5.2328787000000005,"lat":52.188281800000006},{"lon":5.2535733,"lat":52.201553000000004},{"lon":5.24305,"lat":52.214337},{"lon":5.2225318000000005,"lat":52.220196400000006}],"sources":[{"lon":5.0827256,"lat":52.2065365},{"lon":5.094541,"lat":52.231289}],"costing":"bicycle"}' https://valhalla1.openstreetmap.de/sources_to_targets | jq .`

The same source twice
`curl -d '{"targets":[{"lon":5.1976415000000005,"lat":52.1787269},{"lon":5.2328787000000005,"lat":52.188281800000006},{"lon":5.2535733,"lat":52.201553000000004},{"lon":5.24305,"lat":52.214337},{"lon":5.2225318000000005,"lat":52.220196400000006}],"sources":[{"lon":5.0827256,"lat":52.2065365},{"lon":5.0827256,"lat":52.2065365}],"costing":"bicycle"}' https://valhalla1.openstreetmap.de/sources_to_targets | jq .`

When I use the car profile the output seems to be correct and in line with what I expect to see.
`curl -d '{"targets":[{"lon":5.1976415000000005,"lat":52.1787269},{"lon":5.2328787000000005,"lat":52.188281800000006},{"lon":5.2535733,"lat":52.201553000000004},{"lon":5.24305,"lat":52.214337},{"lon":5.2225318000000005,"lat":52.220196400000006}],"sources":[{"lon":5.0827256,"lat":52.2065365},{"lon":5.0827256,"lat":52.2065365}],"costing":"auto"}' https://valhalla1.openstreetmap.de/sources_to_targets | jq .`

nilsnolde:
fixes #3867 
fixes #4153

Next up: isochrones.


src/thor/matrix_action.cc
```diff
@@ -27,33 +27,30 @@ std::string thor_worker_t::matrix(Api& request) {
...
-    return tyr::serializeMatrix(request, time_distances, distance_scale, MatrixType::TimeDist);
+    request.mutable_matrix()->set_algorithm(Matrix::TimeDistanceMatrix);
+    time_distance_bss_matrix_.SourceToTarget(request, *reader, mode_costing, mode,
+                                             max_matrix_distance.find(costing)->second,
+                                             options.matrix_locations());
+    return tyr::serializeMatrix(request);
   }
 
-  MatrixType matrix_type = MatrixType::Cost;
+  Matrix::Algorithm matrix_algo = Matrix::CostMatrix;
```
nilsnolde, line 53: I removed the MatrixType enum we kept around and replaced it with the PBF's one.

src/thor/timedistancebssmatrix.cc
```diff
@@ -538,12 +535,16 @@ bool TimeDistanceBSSMatrix::UpdateDestinations(
...
+  uint32_t idx = 0;
   for (auto& dest : destinations_) {
-    td.emplace_back(dest.best_cost.secs, dest.distance);
+    Matrix::TimeDistance& td = *request.mutable_matrix()->mutable_time_distances()->Add();
+    td.set_from_index(idx / static_cast<uint32_t>(request.options().targets().size()));
+    td.set_to_index(idx % static_cast<uint32_t>(request.options().targets().size()));
+    td.set_distance(dest.distance);
+    td.set_time(dest.best_cost.secs);
+    idx++;
   }
```
nilsnolde, line 547: could also turn this into a templated function in `matrix_common.h` for all matrix algos to share

valhalla/thor/costmatrix.h
```diff
@@ -280,12 +277,6 @@ class CostMatrix {
     return infos;
   };
 
-  /**
-   * Form a time/distance matrix from the results.
-   * @return  Returns a time distance matrix among locations.
-   */
-  std::vector<TimeDistance> FormTimeDistanceMatrix();
```
nilsnolde, line 287: it never had a function like this implemented

proto/matrix.proto
```diff
@@ -0,0 +1,22 @@
...
+package valhalla;
+import public "common.proto";
+
+message Matrix {
+  enum Algorithm {
+    TimeDistanceMatrix = 0;
+    CostMatrix = 1;
+  }
+
+  message TimeDistance {
```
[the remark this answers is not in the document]
nilsnolde: hey bypasser, appreciate the feedback:) good point to flatten it so proto can pack them, didn't think of that.

nilsnolde:
I took @SiarheiFedartsou 's suggestion and changed the proto definition to be "flat" vectors of the matrix attributes (times, dists etc), instead of a TimeDistance message "wrapper", so the int vectors can be packed.

I also finished the forward vs reverse output order.

src/thor/timedistancematrix.cc
```diff
@@ -556,17 +553,33 @@ bool TimeDistanceMatrix::UpdateDestinations(
...
+    matrix.mutable_from_indices()->Set(pbf_idx, forward ? origin_index : i);
+    matrix.mutable_to_indices()->Set(pbf_idx, forward ? i : origin_index);
+    matrix.mutable_distances()->Set(pbf_idx, dest.distance);
+    matrix.mutable_times()->Set(pbf_idx, time);
+
+    // this logic doesn't work with string repeated fields, gotta collect them
+    // and process them later
+    auto date_time =
+        get_date_time(origin_dt, origin_tz, pred_id, reader, static_cast<uint64_t>(time));
+    out_date_times[pbf_idx] = date_time;
```
nilsnolde, line 581: this pisses me off a bit.. we can't use the `Resize()` and `Set()` syntax for the string repeated fields, protobuf has no such concept for those.. I didn't really know how to do  it any better, so:
- before we expand the sources or targets, we initialize an empty `std::vector<std::string>`, which we pass into this function for every source or target forming
- here we set the date_time string on the right index
- after we're all done with all expansions, we go over this vector and populate the PBF's `date_times`

Any suggestion on how to make it less ugly are welcome.

test/matrix.cc
```diff
@@ -172,10 +172,12 @@ const auto test_request_osrm = R"({
...
 
-std::vector<TimeDistance> matrix_answers = {{28, 28},     {2027, 1837}, {2403, 2213}, {4163, 3838},
-                                            {1519, 1398}, {1808, 1638}, {2061, 1951}, {3944, 3639},
-                                            {2311, 2111}, {701, 641},   {0, 0},       {2821, 2626},
-                                            {5562, 5177}, {3952, 3707}, {4367, 4107}, {1825, 1680}};
+// clang-format off
+std::vector<std::vector<uint32_t>> matrix_answers = {{28, 28},     {2027, 1837}, {2403, 2213}, {4163, 3838}, 
+                                                     {1519, 1398}, {1808, 1638}, {2061, 1951}, {3944, 3639},
+                                                     {2311, 2111}, {701, 641},   {0, 0},       {2821, 2626},
+                                                     {5562, 5177}, {3952, 3707}, {4367, 4107}, {1825, 1680}};
```
nilsnolde, line 179: some human-healthier formatting :)

test/matrix.cc
```diff
@@ -198,64 +200,113 @@ TEST(Matrix, test_matrix) {
...
+      {"lat":52.103105,"lon":5.081005}
+    ],
+    "targets":[
+      {"lat":52.106126,"lon":5.101497},
+      {"lat":52.100469,"lon":5.087099},
+      {"lat":52.103105,"lon":5.081005},
+      {"lat":52.094273,"lon":5.075254}
+    ],
+    "costing":"auto"
+  })";
```
nilsnolde, line 273: added a test for #4153

test/matrix.cc
```diff
@@ -286,25 +337,24 @@ TEST(Matrix, test_timedistancematrix_results_sequence) {
...
-               "'s time is not equal to the expected value for TimeDistMatrix";
+  timedist_matrix.SourceToTarget(request, reader, mode_costing, sif::TravelMode::kDrive, 400000.0);
+  auto& matrix = request.matrix();
+
+  // expected results are the same as `matrix_answers`, but without the last target
+  // clang-format off
+  std::vector<std::vector<uint32_t>> expected_results = {{28, 28},     {2027, 1837}, {2403, 2213}, 
+                                                      {1519, 1398}, {1808, 1638}, {2061, 1951},
+                                                      {2311, 2111}, {701, 641},   {0, 0},
+                                                      {5562, 5177}, {3952, 3707}, {4367, 4107}};
```
nilsnolde, line 348: only changed formatting

test/matrix.cc
wieringen: The bug in #4153 only occurred with the bicycle profile, the auto profile worked fine. Does that matter for your test/fix?

nilsnolde: Jep that's fine. Over the API we decide internally which matrix algo to use. In your case that was TDMatrix, for auto it's always CostMatrix. This test uses TDMatrix deliberately. Good point though!

test/matrix.cc
```diff
@@ -286,30 +355,28 @@ TEST(Matrix, test_timedistancematrix_results_sequence) {
...
+               " to expected value for TDMatrix";
+
+    EXPECT_NEAR(matrix.times()[i], expected_results[i][0], kThreshold)
+        << "result " + std::to_string(i) + "'s time is not equal" + " to expected value for TDMatrix";
   }
 }
 
-// TODO: it was commented before. Why?
-TEST(Matrix, DISABLED_test_matrix_osrm) {
+TEST(Matrix, test_matrix_osrm) {
```
nilsnolde, line 379: to increase coverage, I test the OSRM matrix format properly. I guess the reason why it commented was because it wasn't set up to actually test OSRM format, it just looked at the raw time/distances, not at the json output

kevinkreiser (approved):

kevinkreiser (approved):

PR #4130, opened 2023-05-27
Build Tyr with -Werror

dillona:


src/tyr/CMakeLists.txt
```diff
@@ -4,16 +4,16 @@ set(sources
...
+  route_serializer.cc
   serializers.cc
+  trace_serializer.cc
   transit_available_serializer.cc)
 
 set(sources_with_warnings
-  locate_serializer.cc
-  matrix_serializer.cc
-  route_serializer.cc
   route_serializer_osrm.cc
```
dillona, line 15: I could use feedback on these last two files.

The error is an unused function in each file. I don't want to delete them unless that's desired by the project.

nilsnolde: Which function is it @dillona ? Can you quickly point us there so no one has to run it? Thanks:)

valhalla/meili/state.h
```diff
@@ -53,7 +53,11 @@ class State {
       ++dest;
     }
     labelset_ = labelset;
+#ifdef LOGGING_LEVEL_TRACE
     LOG_TRACE("Found " + std::to_string(found) + " destinations out of " + std::to_string(dest - 1));
+#else
+    (void)found;
+#endif
```
kevinkreiser, line 60: instead of this i would use the `[[maybe_unused]]` qualifier on the declaration of found. we are on c++17 now so we can totally use that feature

src/tyr/CMakeLists.txt
dillona: Yes of course.

It is the serialize function. See https://github.com/valhalla/valhalla/blob/d44cd26f756e3861f0fc01b86e2a3b8bd2435f6c/src/tyr/route_serializer_osrm.cc#L1662

valhalla/meili/state.h
dillona: Thanks for the tip! I hadn't heard of this functionality before.

I made this change in 51d23a74c098612202f345a7f071e0c7d2ce110b

PR #4151, opened 2023-06-06
Changes to concurrency config

nilsnolde:
- removed `-j` from `valhalla_add_predicted_traffic` and respect the `mjolnir.concurrency` config as all other executables
- made the `mjolnir.concurrency` config non-optional and defaults to all threads, IMHO it's better people can easier see it in the config


kevinkreiser:
this kind of gets at the heart of what should and shouldnt be in the config. i think this is fine but honestly i would find it annoying if we read this value for `valhalla_service` from config instead of from the command line. to me, in some executables there are args that i want to be able to quickly change without editing a file. concurrency does seem like one. for tile building and data jobs reading it from config makes more sense becasue that isnt typically a process that you play around with a lot. after you find a configuration that works you stick with it. anyway those are teh thoughts this pr evoked for me :D

scripts/valhalla_build_config
```diff
@@ -288,7 +289,7 @@ help_text = {
         'user_agent': 'User-Agent http header to request single tiles',
         'tile_url': 'Http location to read tiles from if they are not found in the tile_dir, e.g.: http://your_valhalla_tile_server_host:8000/some/Optional/path/{tilePath}?some=Optional&query=params. Valhalla will look for the {tilePath} portion of the url and fill this out with a given tile path when it make a request for that tile',
         'tile_url_gz': 'Whether or not to request for compressed tiles',
-        'concurrency': 'How many threads to use in the concurrent parts of tile building',
+        'concurrency': 'How many threads to use in the concurrently running executables or parts of executables. Default all.',
```
kevinkreiser, line 292: depending on a persons familiarity with valhalla this might be misleading. i think it only applies to executables that are for data buildings tools, at present right? i mean its inside the mjolnir part of the config but im just saying, its not ALL executables.

scripts/valhalla_build_config
```diff
@@ -30,7 +31,7 @@ config = {
         'user_agent': Optional(str),
         'tile_url': Optional(str),
         'tile_url_gz': Optional(bool),
-        'concurrency': Optional(int),
+        'concurrency': mp.cpu_count(),
```
kevinkreiser, line 34: i assume the main benefit of doing this is to demystify what the code is doing when you leave this out, right? i cant argue with that!

kevinkreiser (approved):

nilsnolde:
Yeah, honestly, right after I was done with this, I thought I'd prefer to have every executable ask specifically for concurrency, rather than having it in a config. I sometimes want to do thing in parallel on the same machine and it can be necessary to limit threads then, e.g. `valhalla_build_tiles` (until `build` stage) & `valhalla_ingest_transit` can be run in parallel as they both have a long-ish single-thread thing going on before fanning out, and there I'd like to give the transit one more cores than the tile builder. 

So, in the end, should we go that way and I close here? Also great for "demystifying". And adding a `-j 6` is certainly less annoying than editing a file and wondering if that particular executable even does smth on more than one thread.

nilsnolde:
Should we just add concurrency to all affected executables, keeping the config as it was before (ie optional), and prefer command line arg if it was specified? Hardly extra code and should make everyone happy?

kevinkreiser:
that's sounds great, ship it

nilsnolde:
Ok, I changed it to the way I proposed. Was a lot more changes than I anticipated, but I pulled through.. It's super mechanical, no need to explain anything, will do when reviewing

src/mjolnir/valhalla_add_elevation.cc
```diff
@@ -6,19 +6,22 @@
...
 namespace opt = cxxopts;
 
 using namespace valhalla::baldr;
 using namespace valhalla::midgard;
 using namespace valhalla::mjolnir;
 
-enum Input { CONFIG, TILES };
+boost::property_tree::ptree config;
+std::vector<std::string> tiles;
+uint32_t num_threads = 0;
```
nilsnolde, line 24: I'm kinda sorry for this whole refactor, it wasn't really necessary, but I felt a need for consistency with the other executables. If we have boilerplate code, I'd personally prefer it to be actual boilerplate:)

valhalla/mjolnir/valhalla_add_elevation_utils.h
```diff
@@ -1,46 +0,0 @@
...
-#include "baldr/graphid.h"
-#include "baldr/graphreader.h"
-#include "baldr/graphtile.h"
-#include "filesystem.h"
-#include "midgard/logging.h"
-
-namespace valhalla {
-namespace mjolnir {
-
-std::deque<baldr::GraphId> get_tile_ids(const boost::property_tree::ptree& pt,
```
nilsnolde, line 18: this header was a bit unnecessary, was only used in valhalla_add_elevation (as the header name suggests..)

scripts/valhalla_build_config
nilsnolde: `Default all` refers to all threads, not all executables

nilsnolde:
What has been done consistently throughout all concurrently running executables:
- add `-j` as arg, defaulting to 0
- pass the `num_threads` down to the concurrent parts
- if `-j` is 0 (i.e. not specified on command line), it'll default to all threads

I kept `mjolnir.concurrency = mp.cpu_count()`, it's still a good change IMO.

nilsnolde:
Kevin had a very good point.. rather overwrite the config's `mjolnir.concurrency` with `-j` so the exes will be the only thing to change.. jeez, so much annoying work for nothing haha will close here and open a new one

PR #4152, opened 2023-06-07
Update the street name and sign data processing include language

gknisely:
# Issue

Update the street name and sign data processing include language.

## Data processing updates
The heart of this work involves processing `name:<lg>` and `destination:<lg>` tags. This includes all variations of each of those tags (e.g., `name:left:<lg>, name:right<lg>, official_name:<lg>, destination:street:lang:<lg>, destination:ref:lang:<lg>, etc.`) Within the tag the lg stands for the language. 

### Language updates
In order to determine the languages for an LL we utilized the [default_language](https://wiki.openstreetmap.org/wiki/Key:default_language) that is defined for a country or providence/state. The problem with this is the fact that some areas in the world are multi-lingual and don't have one language (e.g., in Belgium they support Dutch, French, and German and in Switzerland they speak German, French, Romanish, and Italian). However, the value of the default_language would usually contain only one language.  In order to resolve this issue, the administrative builder was updated to handle these special cases like Brussels via processing relations with `boundary=political` with `political_division=linguistic_community` to get areas that are bilingual. However, this logic was still not enough as other areas did not have these special polygons. Therefore, we added an "override" for the languages. So, in the future if we determine that an area is multilingual and we want to support additional language tags in this area. All we have to do is add them to the `supported languages` list. Our list currently consists of the following:

Wales = cy
United Kingdom = en
Ireland = ga
Northern Ireland = ga
Japan = ja and en
Canada = en and fr
Belarus = ru and be
Singapore = en, zh, ms, and ta
Saudi Arabia = ar and en

### Using the languages for the keys
The processing of the new keys is based on the languages that were determined to be "good" for that country or area. OSM users will typically add languages for names and destinations in all parts of the world even though that language may not be spoken in that country. For example, [5th Avenue in NYC](https://www.openstreetmap.org/way/511699185) has name:ru=5-я авеню. Obviously, we do not want to process the Russian name here. Therefore, using the default languages we can toss the tags with languages that we don't want to support. Moreover, we can create a hierarchy for our languages. For instance, Canada supports both English and French. However, in Ottawa, English will be first and in the Québec province French will be first.

### Edge Cases
We will now support names where they differ depending on which side of the street you are driving on. When combined with the multi-languages for some areas, it gets very complex. In this [example](https://www.openstreetmap.org/way/845098764), the official Dutch name differs depending on the municipality. Basically, the border of the towns runs down the middle of the road and on the right side the Dutch name of the street differs from the left side of the street. In part, this leads to the bizarre situation that the street on the Molenbeek side is called Steenweg op Gent and on the Koekelberg side Gentsesteenweg. However, the French name of the street does not change at all.

#### Chaussée de Gand - Steenweg op Gent/Gentsesteenweg Example
Notice that all dashes are removed and processed correctly.   

Driving left to right the Dutch name should be Steenweg op Gent and the French street name(Chaussée de Gand) does not change.  Notice that before we used to return the name tag with dashes and did not have the French and Dutch street names split up.

##### Before
Screenshot from 2022-02-21 15-13-21

##### After 
Screenshot from 2022-02-21 15-13-34

Driving right to left the Dutch name should be Gentsesteenweg and the French street name(Chaussée de Gand) does not change.
##### Before
Screenshot from 2022-02-21 15-14-36

##### After
Screenshot from 2022-02-21 15-14-26

#### name:forward and name:backward is now processed correctly.  
Notice in this [example](https://www.openstreetmap.org/way/58744373) we have name:forward and name:backward tags set; however, before we would just process the name tag.

##### Before - Waltonville Road and Quarry Road returned regardless of direction  

TBD - new image needed

Screenshot from 2022-02-21 15-35-34

##### After - Quarry Road correctly returned 
Screenshot from 2022-02-21 15-36-03

##### Before - Waltonville Road returned regardless of direction
Screenshot from 2022-02-21 15-36-26

##### After - Waltonville Road correctly returned.
Screenshot from 2022-02-21 15-36-38

#### Multilingual names are now processed.
Notice in this [example](https://www.openstreetmap.org/way/24165304) the name tag has both Welsh and English set.  Since we are in Wales we allow both of these languages and process them both.

##### Before - The name and name:en tag are both returned for the street:  Stryd y Castell / Castle Street/Castle Street 
Screenshot from 2022-02-21 15-46-01

##### After - Stryd y Castell and Castle Street both processed correctly and a language of cy is set for Stryd y Castell and en for Castle Street 
Screenshot from 2022-02-21 15-46-19

Contributors @gknisely @dgearhart


gknisely:
do not delete branch yet

PR #4167, opened 2023-06-16
speed up transit ingestion

kevinkreiser:
At the moment `valhalla_ingest_transit` is slow. After looking at the code for a few minutes i noticed that `just_gtfs` does a lot of scanning vectors to find stuff by gtfs id. This seemed slow to me, so i tried the profiler and it agreed.

I forked `just_gtfs` into the valhalla org and made a new branch that sorts the vectors (optionally for now but i think this should be the only mode really..).

In my test (pittsburgh) it brought the time to ingest the transit data from close to 6 minutes down to close to 2 minutes.

I'll probably just remove the ability to leave the data unsorted but wanted to throw this up here to show the progress.

@mesozoic-drones would you accept a pr? i see @nilsnolde has a pr that is unmerged for a while so i dont want to waste your time if you are not interested. thanks again!

EDIT:

Ok this is ready for second review, ive changed some more with just gtfs primarily:

1. no more copies of structs, everything returns `const &`
2. validity/not-found semantics now use `gtfs::valid` since `std::optional` cannot take `const &`
3. sorting is not optional anymore its too advantageous to not do it

the time is now down from 6min for the pittsburgh feed to 1m20s. thats just shy of an 80% reduction. the bulk of the time is now spent just doing the initial enumeration of all the stuff in the feed (the tiling part). this will quickly become the bottleneck once we start doing tons of feeds at once. we can parallelize it and do other optimizations when that time comes.


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
    * FIXED: when reclassifying ferry edges, remove destonly from ways only if the connecting way was destonly [#4118](https://github.com/valhalla/valhalla/pull/4118)
 * **Enhancement**
    * UPDATED: French translations, thanks to @xlqian [#4159](https://github.com/valhalla/valhalla/pull/4159)
+   * ENHANCED: sped up the transit gtfs ingestion process by sorting the feeds before querying them, forked just_gtfs to accomplish it [#4167](https://github.com/valhalla/valhalla/pull/4167)
```
nilsnolde, line 8: hard nit: technically we'd say `CHANGED` here

CHANGELOG.md
```diff
@@ -91,7 +92,7 @@
    * ADDED: Added has_toll, has_higway, has_ferry tags to summary field of a leg and route and a highway tag to a maneuver if it includes a highway. [#3815](https://github.com/valhalla/valhalla/issues/3815)
    * ADDED: Add time info to sources_to_targets [#3795](https://github.com/valhalla/valhalla/pull/3795)
    * ADDED: "available_actions" to the /status response [#3836](https://github.com/valhalla/valhalla/pull/3836)
-   * ADDED: "waiting" field on input/output intermediate break(_through) locations to respect services times [#3849](https://github.com/valhalla/valhalla/pull/3849)
+   * ADDED: "waiting" field on input/output intermediate break(\_through) locations to respect services times [#3849](https://github.com/valhalla/valhalla/pull/3849)
```
nilsnolde, line 95: oops haha did that turn the whole document after to italics?

nilsnolde (approved):

nilsnolde:
This is great, thanks for the effort! 

Though you forked from the wrong repo, earlier this year @mesozoic-drones shifted the dev version to her own repo: https://github.com/mesozoic-drones/just_gtfs. This is also the submodule we have here ever since. There my PR was  also merged:)

kevinkreiser:
ah ok I can make sure I have the changes from that repo. I just quickly googled and forgot about the backstory.

regarding 'changed' I looked through the changelog to see what we used because semantically changed is incredibly vague and to me seems worse than useless. actually i really wish we wouldn't use the all caps thing under the headings at all as it's just wasted bytes with 0 extra meaning. at times in the past we didn't use them but sadly we do now. so yeah I'll change it

nilsnolde:
"hard nit" was maybe lost in translation, I meant "I don't really care":) it's just a tiny consistency thing and I fully agree, it's really meaningless bytes

kevinkreiser:
Ok this is ready for second review, ive changed some more with just gtfs primarily:

1. no more copies of structs, everything returns `const &`
2. validity/not-found semantics now use `gtfs::valid` since `std::optional` cannot take `const &`
3. sorting is not optional anymore its too advantageous to not do it

kevinkreiser:
sadly still not done here. our transit tests fail and some of it is definitely due to the sorting. the first test where we write transit and then read it back to make sure its right doesnt realize that the data is sorted so it expects to be able to use certain indices to map to certain stops. i can change the test a bit to be robust to this (use the find methods rather than indices). 

further down it is also saying that the days mask on the tile says all 59 days have service. this again might be us relying on indices for a certain schedule but im not sure yet, didnt look too close. i'll check it later

kevinkreiser:
this is ready for final review

src/mjolnir/ingest_transit.cc
```diff
@@ -474,18 +477,18 @@ bool write_stop_pair(
...
+      service_dow->Add(gtfs::Thursday & dow_mask);
+      service_dow->Add(gtfs::Friday & dow_mask);
+      service_dow->Add(gtfs::Saturday & dow_mask);
+      service_dow->Add(gtfs::Sunday & dow_mask);
 
       bool had_added_date = false;
-      for (const auto& cal_date_item : trip_calDates) {
-        auto d = to_local_pivot_sec(cal_date_item.date.get_raw_date());
-        if (cal_date_item.exception_type == gtfs::CalendarDateException::Added) {
+      for (auto cal_itr = trip_calDates.first; cal_itr != trip_calDates.second; ++cal_itr) {
```
kevinkreiser, line 489: i had a bug here in the first draft of this pr where i didnt make a copy of the first iterator in the range and instead modified the original. this mean that for other stop pairs that use teh same trip they wouldnt get these exceptions which causes the tests to fail, so thanks to the tests this was caught!

nilsnolde (approved):

kevinkreiser:
just for preservation purposes i have pr'd our fork of just_gtfs with the speedups to the currently maintained version: https://github.com/mesozoic-drones/just_gtfs/pull/3

PR #4168, opened 2023-06-17
-j flag for mjolnir exes

nilsnolde:
a much simpler version of #4151


valhalla/baldr/graph_utils.h
```diff
@@ -6,14 +6,14 @@
 
 #include <boost/property_tree/ptree.hpp>
 
-#include "baldr/graphid.h"
-#include "baldr/graphreader.h"
-#include "baldr/graphtile.h"
-#include "filesystem.h"
-#include "midgard/logging.h"
+#include <valhalla/baldr/graphid.h>
```
nilsnolde, line 9: moved this header from mjolnir to baldr and called it graph_utils.h. it was only used in valhalla_add_elevation.cc (& its test), but could be useful in other places and really shouldn't live inside mjolnir

src/mjolnir/valhalla_add_elevation.cc
```diff
@@ -30,90 +34,97 @@ enum Input { CONFIG, TILES };
...
-  options.add_options()("h,help",
-                        "Print usage")("c,config", "Path to the configuration file.",
-                                       opt::value<
-                                           std::string>())("t,tiles", "Tiles to add elevations to",
-                                                           opt::value<std::vector<std::string>>());
-
-  options.allow_unrecognised_options();
-  std::string config_file;
-  std::vector<std::string> tiles;
+std::deque<GraphId> get_tile_ids(const boost::property_tree::ptree& pt,
```
nilsnolde, line 37: this is just one bigger refactor I felt like doing, to make this exe somewhat consistent with all others.

in a follow up PR I'll actually break out all of this boilerplate arg parsing to a private header

src/mjolnir/valhalla_convert_transit.cc
```diff
@@ -63,11 +69,6 @@ bool ParseArguments(int argc, char* argv[]) {
       std::sort(onestoptests.begin(), onestoptests.end());
     }
 
-    if (result.count("target_directory")) {
-      pt.get_child("mjolnir").erase("tile_dir");
-      pt.add("mjolnir.tile_dir", result["target_directory"].as<std::string>());
-    }
```
nilsnolde, line 69: this was duplicated code, already exists a few lines up

kevinkreiser (dismissed):
looks good just would like to keep the elevation tile_ids things as an unordered_set

nilsnolde:
> looks good just would like to keep the elevation tile_ids things as an unordered_set

did that and merged it into the superseding PR. closing here

PR #4169, opened 2023-06-17
move argparse boilerplate code to private header

nilsnolde:
Removes a few lines of code by breaking out argparse stuff of the programs into a private header.

The PR is based on #4168


nilsnolde:
Apart from adding the utility function, what it generally does:
- adds a `i,--inline-config` wherever it made sense (i.e. where `-i` wasn't taken yet and a config is needed)
- made all programs parse the args consistently, e.g. how exceptions are handled, help strings etc.

kevinkreiser (dismissed):
ok so this supercedes the previous one but a couple things:

- can we put the interface for the elevation get_tileids back to unordered_set
- can we not use relative paths to include the private headers, looked like cmakelists already had the source dir has an include directory

src/mjolnir/valhalla_add_elevation.cc
```diff
@@ -30,95 +32,61 @@ enum Input { CONFIG, TILES };
...
-  options.add_options()("h,help",
-                        "Print usage")("c,config", "Path to the configuration file.",
-                                       opt::value<
-                                           std::string>())("t,tiles", "Tiles to add elevations to",
-                                                           opt::value<std::vector<std::string>>());
-
-  options.allow_unrecognised_options();
-  std::string config_file;
+int main(int argc, char** argv) {
+  const std::string program = "valhalla_add_elevation";
```
kevinkreiser, line 36: you can get this by messing around with `__FILE__` then you dont have to type every single one though i guess it has to be the name of the source file (which we do adhere to)

nilsnolde: jep, that worked, thanks! made it a function in `argparse_utils.h`

nilsnolde:
@kevinkreiser 

> can we put the interface for the elevation get_tileids back to unordered_set

done

> can we not use relative paths to include the private headers, looked like cmakelists already had the source dir has an include directory

had to adjust the CMakeLists.txt a bit and add the entire `./src` to `target_include_directories()`, the `VALHALLA_SOURCE_DIR` is actually root of the repo, threw me off too..

src/argparse_utils.h
```diff
@@ -0,0 +1,87 @@
...
+#include <valhalla/midgard/util.h>
+
+// returns a (const char*)'s path's filename without the file extension, e.g.
+// /home/valhalla/valhalla_add_elevation.cc -> valhalla_add_elevation
+std::string file_stem(const char* path) {
+  auto basename = filesystem::path(path).filename().string();
+  if (basename.length() <= 3) {
+    return "";
+  }
+  basename.erase(basename.length() - 3);
```
kevinkreiser, line 17: nit: are you sure you dont want to `rfind` `'.'`, i know its pedantic but its robust to us one day changing from stupid `.cc` for to the more prevalent `.cpp/.hpp`

valhalla/baldr/graph_utils.h
```diff
@@ -6,14 +6,14 @@
```
kevinkreiser, line 6: this whole file seems misplaced. could we just move the functions unto an existing utils header?

nilsnolde: I didn’t feel it’s fitting anywhere.. def not mjolnir, and it’s not really midgard material.

kevinkreiser: why dont we just delete this file entirely and then we can change the interface of elevationbuilder::build to take an unordered_set and make a dequeue out of it and check the filesystem etc if that is what it needs. just move the functionality into the main function, or, if it means testing is less verbose thats ok with me. rather have less random files hanging around tbh

test/elevation_builder.cc
```diff
@@ -228,6 +228,7 @@ TEST(ElevationBuilder, get_tile_ids) {
   // check if config contains tile_dir
   EXPECT_TRUE(valhalla::mjolnir::get_tile_ids(config, {"/0/003/196.gph"}).empty());
 
+  // TODO: I don't understand what the following lines have to do with anything?
```
kevinkreiser, line 231: above this uses the default config, i suppose if on your local computer you had elevation data in the default location it would pick that up and change the result of this test. to avoid that the person removes this value so that it cant happen

kevinkreiser:
this looks done to me, i woul just remove the extra file and push that logic into elevationbuilder::build and call it a day

kevinkreiser (approved):

src/valhalla_export_edges.cc
```diff
@@ -158,39 +163,23 @@ int main(int argc, char* argv[]) {
...
+              << "This is a bug, please report it at " PACKAGE_BUGREPORT << "\n";
     return EXIT_FAILURE;
   }
 
-  // parse the config
-  boost::property_tree::ptree pt;
-  rapidjson::read_json(config.c_str(), pt);
-
-  // configure logging
-  valhalla::midgard::logging::Configure({{"type", "std_err"}, {"color", "true"}});
```
kevinkreiser, line 192: opps we wanted to keep this stderr #4496

PR #4170, opened 2023-06-18
fix "Edit this page" links in documentation

not-my-profile:
The mkdocs files were moved to the docs/ directory in c0bac7cf7f1c7005194e3cb4e929e00335a74e28,
which broke the "Edit this page" links from the Material theme. This commit fixes these links by configuring edit_uri[1] accordingly.

[1]: https://www.mkdocs.org/user-guide/configuration/#edit_uri

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -50,6 +50,7 @@ markdown_extensions:
 
 site_name: Valhalla Docs
 repo_url: https://github.com/valhalla/valhalla/
+edit_uri: edit/master/docs/docs/
 site_url: https://valhalla.github.io/valhalla
 docs_dir: docs
 nav:


nilsnolde (approved):

nilsnolde:
Thanks!

not-my-profile:
You're welcome! Thanks for the quick merge :) Just a small heads-up that I prefer my commit messages to be preserved on merge (not really important in this case but something I care about in general).

Sidenote: What's up with the [indirect code coverage changes](https://app.codecov.io/gh/valhalla/valhalla/pull/4170/indirect-changes)?

nilsnolde:
For me the default is to remove the commit messages. In 99% of the PRs (here at least), it’s not very linear (ie noisy) and we, for the better or worse, rely on GitHub PRs to give more context for relevant changes. 

codecov reporting is still somewhat a mystery to me and I personally don’t have the drive to find out what’s the problem or if we could do smth about it. It’s usually failing for such PRs where it shouldn’t 🤷

not-my-profile:
Right, I'd like you to make an exception for my PRs ... because my commit messages are 100% carefully crafted ;)

I am afraid I also don't have a clue regarding the code coverage flukes.

PR #4175, opened 2023-06-22
fix typos

not-my-profile:
The typos were found with the typos[1] spell checker.  That spell
checker will be incorporated into the CI in a followup PR.

[1]: https://github.com/crate-ci/typos


nilsnolde:
Makes me a little bit nervous to have the automated approach without proper verification, see conan failure on CI. Just my 2c, WDYT @kevinkreiser ?

not-my-profile:
I reverted the typo fix in `conan.cmake` so that this PR can be merged with a passing CI. But we should of course look into why fixing `detected_setings` to `detected_settings` in `conan.cmake` breaks conan.

src/meili/match_route.cc
```diff
@@ -73,7 +73,7 @@ EdgeSegment::EdgeSegment(baldr::GraphId the_edgeid,
 
 /**
  * Here we return the vector of edge segments between the source and target states. If its a node to
- * node route (meaning no realy edge is traversed) then we use the target_result to say what edge the
+ * node route (meaning no relay edge is traversed) then we use the target_result to say what edge the
```
kevinkreiser, line 76: this is why I hate stuff like this. this is wrong and this tool will surely make other mistakes. personally I can live with imperfection for the sake of not living under the boot of another annoying CI bot. if we put this in CI it should not fail the build

kevinkreiser (changes requested):

src/meili/match_route.cc
not-my-profile: I did actually review all of the changes and they made sense to me. "realy" isn't a word as far as I can tell and "relay edge" somewhat makes sense to me. What would be the correct fix, or what does "realy" mean?

kevinkreiser: it's supposed to be 'real'

kevinkreiser:
I would be ok with this as a separate github action that if it fail doesn't block merging. what about you @nilsnolde ?

nilsnolde:
Hm, first we'd need to review all 122 files in here. As far as I'm concerned, not in the coming weeks. Eventually, sure, separate non-blocking CI action would be nice.

kevinkreiser:
I made it partially through at the very least it's almost all comments and not actual code

src/mjolnir/timeparsing.cc
```diff
@@ -37,7 +37,7 @@ FormatCondition(const std::string& source, const std::regex& regex, const std::s
   return std::regex_replace(source, regex, pattern);
 }
 
-// get the dow mask from user inputed string.  try to handle most inputs
+// get the dow mask from user inputted string.  try to handle most inputs
```
kevinkreiser, line 40: if we are going to change this it should be 
```suggestion
// get the dow mask from the user's input string.  try to handle most inputs
```

and same below. no one says inputted

not-my-profile:
The typos spell-checker only reports known typos ... i.e. it's blacklist as opposed to whitelist checking against a dictionary, so it really has low false positives, making it suitable for a blocking CI action IMHO.

src/mjolnir/timeparsing.cc
not-my-profile: I'd rather leave grammar fixes for a future PR ... this PR is already big enough ... and I think it makes sense to separate mere typo fixing from rewriting.

valhalla/mjolnir/osmnode.h
```diff
@@ -90,8 +90,8 @@ struct OSMNode {
   }
 
   /**
-   * Sets the OSM node Id. Ensures the Id does not exceed the maximum allowed based on
```
kevinkreiser, line 93: good catch

kevinkreiser (changes requested):
the 9nly thing left to do with this is stop using the word 'inputted' then we can ship it

not-my-profile:
I changed all instances of "inputted" to "provided".

src/meili/match_route.cc
kevinkreiser: it's because you don't know the structure of our graph. this edge is fake because it has no spatial component. it connects a node to a copy of itself on another level of the graph. our graph is divided into levels where we can separate the edges by their importance. this allows for search space pruning via something akin to data partitioning

nilsnolde:
clang-tidy wants code changes, can you pls have a look why that’d be?

not-my-profile:
The `scripts/clang-tidy-only-diff.sh` script only checks files that have been modified in the PR. Since typos were fixed in these files the script complains ... the issues it complains about are already present in the master branch.

kevinkreiser:
well thats kind of par for the course when you make this kind of non functional wide reaching change to the repo.

at the end of the day we aren't merging a failing build. maybe we can fix the tidy stuff in a separate pr buy until we do this pr will have to wait unfortunately

not-my-profile:
I fixed the clang tidy failure :) ... see the commit I added.

kevinkreiser (approved):

PR #4182, opened 2023-06-28
Side cutoff

Trietes:
# Issue / Description

This PR adds a new option to consider the preffered side only for bigger streets, since it is irrelevant for smaller streets in use cases like package delivery. See discussion #4176 for more details.
 
A new option for locations called `preferred_side_cutoff` is introduced. If preferred side is set to `same` or `opposite` for a location, but the related street class is smaller then the street class set in `preferred_side_cutoff`, the preferred side is treated at `either`.

## Tasklist

 - [x] Add tests


src/loki/search.cc
```diff
@@ -57,6 +57,13 @@ bool side_filter(const PathLocation::PathEdge& edge, const Location& location, G
   auto* node = reader.GetEndNode(opp, tile);
   if (!node)
     return false;
+
+  // nothing to filter if it is a minor road
+  // since motorway = 0 and service = 7, higher number means smaller road class
+  uint32_t road_class = static_cast<uint32_t>(opp->classification());
+  if (road_class > location.street_side_cutoff_)
+    return false;
```
nilsnolde, line 65: tiny nit: could put that right after getting the `opp`, since we don't need the node for this

test/gurka/test_search_filter.cc
```diff
@@ -91,6 +91,22 @@ TEST_F(SearchFilter, PreferredSide) {
   gurka::assert::osrm::expect_steps(result, {"AB", "AD", "CD", "BC"});
   gurka::assert::raw::expect_path(result, {"AB", "AD", "CD", "BC"});
 }
+TEST_F(SearchFilter, StreetSideCutoff) {
+  auto from = "7";
+  auto to = "8";
+
+  const std::string& request =
+      (boost::format(
+           R"({"locations":[{"lat":%s,"lon":%s},{"lat":%s,"lon":%s,"preferred_side":"same","street_side_cutoff":"primary"}],"costing":"auto"})") %
```
nilsnolde, line 100: this threw me off a little, since it seems it has no impact compared to the default result. looking at the test for `preferred_side` I see though it's the same request, just without the cutoff, and it takes the southbound path. it's the long-standing bug that without an admin db we default to left-side driving..

can you put a capital comment right above the `map` in `SetUpTestSuite`, since it impacts all tests and is pretty confusing on first sight?

nilsnolde (approved):
This is cool, thanks @Trietes ! Just a few tiny updates and LGTM!

test/gurka/test_search_filter.cc
Trietes: Yes, thats little confusing. I've added the comment now. 
Here I just copied the test of the preferred_side setting and demonstrated that its not using the preferred side, since the road class is residential, so a too small class to consider the correct side.

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -41,10 +41,11 @@ To build a route, you need to specify two `break` locations. In addition, you ca
...
+| `street_side_tolerance` | If your input coordinate is less than this tolerance away from the edge centerline then we set your side of street to none otherwise your side of street will be left or right depending on direction of travel. The default is 5 meters. |
+| `street_side_max_distance` | The max distance in meters that the input coordinates or display ll can be from the edge centerline for them to be used for determining the side of street. Beyond this distance the side of street is set to none. The default is 1000 meters. |
+| `street_side_cutoff` | Cutoff for which road classes the preferred side matters, when `preferred_side` is set to `same` or `opposite`. It should be a string containing one of the following values (from highest to lowest road-class): motorway, trunk, primary, secondary, tertiary, unclassified, residential, service_other. Ignores the preferred side for all road classes which are smaller than the given cutoff class. Default value is `service_other`, so the set preferred side matters for all road classes. |
```
kevinkreiser, line 48: i find this difficult to read. i think for non-experts they will have a difficult time of it. maybe something like is marginally more clear:

```suggestion
| `street_side_cutoff` | Disables the `preferred_side` when set to `same` or `opposite` if the edge has a road class less than that provided by `street_side_cutoff`. The road class must be one of the following strings: motorway, trunk, primary, secondary, tertiary, unclassified, residential, service_other.  The default value is `service_other` so that `preferred_side` will not be disabled for any edges. |
```

i don't know its hard to explain something so deep in the code :smile: 

also we should deprecate the name `preferred_side` and start parsing `street_side` just so it matches the rest.. who the hell did that.. (a pr for another time of course)

src/worker.cc
```diff
@@ -370,6 +370,12 @@ void parse_location(valhalla::Location* location,
   if (street_side_max_distance) {
     location->set_street_side_max_distance(*street_side_max_distance);
   }
+  auto street_side_cutoff =
+      rapidjson::get<std::string>(r_loc, "/street_side_cutoff", "service_other");
```
kevinkreiser, line 374: why not optional? i worry that we set ourselves up to rely on this always being set. for example, in the pbf-only version of this api we do not parse json and so its totally possible to not set it.

kevinkreiser (changes requested):
i had some tiny nitpicks, looks great otherwise!

docs/docs/api/turn-by-turn/api-reference.md
Trietes: I like your description more. So lets use it.
I agree to the street_side name too. Changing afterwards is ofc always annoying since people are already using it already. But it should be changed.

src/worker.cc
Trietes: I made it optional now

Trietes:
I made all the requested changes now and did a rebase. I'll give it another quick test and mark the PR as ready afterwards.

kevinkreiser (approved):

Issue #4134, opened 2023-05-30
Landmark Routing 1: Intermediate Storage for Parsed Landmarks


kevinkreiser:
- decide on a storage format for the landmarks we parse out of the osm pbf, sqlite is a good one but we could do something simpler and custom like we do for the regular graph data
- write the code to interact with this intermediate storage format (putting, getting and probably spatially indexing are needed)
- write tests to validate the code, you dont need an actual gurka map for this. you should be able to do the 3 operations, put a new record, index it, get back all records within a bounding box

vesperlou:
Hi, 

I just created a PR https://github.com/valhalla/valhalla/pull/4189 for my current progress. 
Have one critical problem: the bounding box select query`LandmarkDatabase::getLandmarksInBoundingBox()` doesn't select the correct rows but select all data in the database. 
Tried a few other select queries, I think the problem is incorrect data in the geometry column. May be wrong `geom` value when inserting landmarks in `LandmarkDatabase::insertLandmark()`, or incorrect geometry part of the select query in `LandmarkDatabase::getLandmarksInBoundingBox()`. 
Spent a whole night but still have nothing. Need some help :)

Junzhen

PR #4189, opened 2023-07-04
Landmark database

vesperlou:
# Issue

Implement a sqlite database as the intermediate storage for parsed landmarks.
[Landmark Routing 1: Intermediate Storage for Parsed Landmarks](https://github.com/valhalla/valhalla/issues/4134)

## Parent project

[GSOC Landmark-based navigation](https://github.com/orgs/valhalla/projects/3?pane=issue&itemId=29465920)


src/mjolnir/landmark_database_builder.cc
```diff
@@ -0,0 +1,189 @@
...
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/geometries.hpp>
+#include <boost/geometry/geometries/register/point.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include <geos_c.h>
+
+namespace valhalla {
+namespace mjolnir {
+bool LandmarkDatabase::openDatabase() {
```
kevinkreiser, line 20: this body of this function and the function below should both be put in the constructor for the class. it doesnt make a lot of sense to have the person who instantiates the db tell it to do this stuff that has to be done regardless, lets just do it automatically.

now it may be the case that we have to differentiate between opening for the first time and opening for reading/appending but i kind of think there is very little if any work to make that happen

kevinkreiser: indexing should also happen automatically

src/mjolnir/landmark_database_builder.cc
```diff
@@ -0,0 +1,189 @@
...
+    sqlite3_free(err_msg);
+    sqlite3_close(db);
+    return false;
+  }
+
+  LOG_INFO("Executed bounding box query");
+  return true;
+}
+
+void LandmarkDatabase::closeDatabase() {
```
kevinkreiser, line 182: this should be done in the destructor no special function needed

src/mjolnir/landmark_database_builder.cc
```diff
@@ -0,0 +1,189 @@
...
+    static_cast<std::vector<std::pair<std::string, std::string>>*>(data);
+
+  results->push_back(std::make_pair(argv[0], argv[1]));
+
+  return 0;
+}
+
+bool LandmarkDatabase::getLandmarksInBoundingBox(std::vector<std::pair<std::string, std::string>> *landmarks, 
+      const std::string& minLat, const std::string& minLong, const std::string& maxLat, const std::string& maxLong) {
+  sql  = "SELECT name, type FROM landmarks WHERE ST_Covers(BuildMbr(" +
```
kevinkreiser, line 166: this should also select and return the geom. you should use st_intersects to future proof this. to create the bounding box you need to use st_envelope and probably also st_setsrid to tell it that it's the lat lon projection

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,47 @@
...
+#include <boost/property_tree/ptree.hpp>
+
+#include <geos_c.h>
+
+namespace valhalla {
+namespace mjolnir {
+struct LandmarkDatabase {
+public:
+  LandmarkDatabase(std::string& db_name) : db(nullptr), database(db_name) {}
+  bool openDatabase();
```
kevinkreiser, line 26: nit: generally we use snake case for everything to kind of match the stl. there are some parts of the code that match the google style like you have here but havent been doing that for quite some time except in the case where there is already an established class. for new stuff snake case please

src/mjolnir/landmark_database_builder.cc
kevinkreiser: ah i see maybe `buildmbr` just needs you to also pass the srid 4326?

src/mjolnir/landmark_database_builder.cc
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  if (type != "") {
+    sqlite3_bind_text(stmt, 2, type.c_str(), type.length(), SQLITE_STATIC);
+  } else {
+    sqlite3_bind_null(stmt, 2);
+  }
+
+  sqlite3_bind_text(stmt, 3, longitude.c_str(), longitude.length(), SQLITE_STATIC);
+  sqlite3_bind_text(stmt, 4, latitude.c_str(), latitude.length(), SQLITE_STATIC);
```
kevinkreiser, line 95: use `sqlite3_bind_double` and pass these args as double from the caller

src/mjolnir/landmark_database_builder.cc
vesperlou: I tried to parse srid 4326 but didn't work. If not clear to you as well I can try to parse real wkt data and see if it works.

src/mjolnir/landmark_database_builder.cc
vesperlou: will have a try

src/mjolnir/landmark_database_builder.cc
kevinkreiser: it really shouldnt be wkt in the end. also the arguments should be doubles not strings. i cant play with it myself at the moment but if you make some of the changes suggested i can take a look maybe at night

vesperlou: no problem. whenerver you are free. I'm working on the code reviews and will make some changes but probably not all of them

vesperlou:
Just updated new commits to address the code reviews. But the geom column still doesn't work. 
I found executing statement for say `INSERT INTO landmarks (name, type, geom) VALUES ('Statue of Liberty', 'Monument', CastToPoint(MakePoint(-74.044548, 40.689253, 4326)))` doesn't insert data into geom column. The points are always NULL.
Any idea?

kevinkreiser:
@vesperlou are you doing it inside of a transation? if so you need to commit it every so often so that it actually makes it into the database. transations are a good idea they are a way to batch update the database (shoudl perform better). maybe your class can take have a batch size and then on every batch_size'th item it can commit. the destructor should also commit of course to make sure everything makes it into the database

vesperlou:
> @vesperlou are you doing it inside of a transation? if so you need to commit it every so often so that it actually makes it into the database. transations are a good idea they are a way to batch update the database (shoudl perform better). maybe your class can take have a batch size and then on every batch_size'th item it can commit. the destructor should also commit of course to make sure everything makes it into the database

No I am not using transactions. The current problem is name and type data can store in the database, only points can’t. All points are NULL. 
Seems to me I can use transactions but we should fix this first otherwise transactions wouldn’t work :)

kevinkreiser:
i dont understand how the insert query doesnt fail then. i haven't looked at the code all that closely for that specific part but i assume yuou check the return code about whether it succeeded or failed. to me this looks like it needs a pair programming debug session. alternatively you can just get on the command line msql client and try the queries directly on yoru database.

i would instrument the code to print out each query that is made when it is made and then do them in the same order by starting an sqlite session. you can start the session by running the program `sqlite3` in your terminal (you might need to install it)

kevinkreiser:
oh and also what platform are you on, are you on linux?

vesperlou:
Yes I’m on linux.
I have checked the documentation which says `CastToPoint` sometimes will give NULL values to be inserted to the database while the query still succeeds. 
Already printed out each query but found nothing. Downloaded a sqlite client to manipulate with my database as well, but didn’t add spatial extension in limit amount of time so functions like `MakePoint` don’t work yet. Will continue on that.

src/mjolnir/landmark_database_builder.cc
```diff
@@ -0,0 +1,273 @@
...
+    return false;
+  }
+  LOG_INFO("Created spatial index");
+  return true;
+}
+
+bool LandmarkDatabase::insert_landmark(const std::string& name, const std::string& type, 
+                                      const double longitude, const double latitude) {
+  sql = "INSERT INTO landmarks (name, type, geom) ";
+  sql += "VALUES (?, ?, CastToPoint(MakePoint(?, ?, 4326)))";
```
nilsnolde, line 109: `CastToPoint` seems like a no-op here, `MakePoint` should also create that exact geom. You can try `ST_Point`, which doesn't take the SRID (it's already defined on the column when you create the table)

But yeah, repeating the same thing on the command line is a good way to debug:)

vesperlou:
@nilsnolde @kevinkreiser Update: I managed to use DB Browser for SQLite with spatialite extension to check my database and sql statements. It turns out my database is working as expected, the `geom` column actually stores the correct data, and bounding box query can get the correct answers. But the same things just don't work within the project.

I suspect the problem is adding problematic spatialite extension. In the project I found many functions are not supported such as `ST_SetSRID` and `ST_MakePoint`. `ST_Point` even gets an error saying `error: landmarks.geom violates Geometry constraint [geom-type or SRID not allowed]`. While in DB Browser all those work properly. 

In DB Browser I manually added `mod_spatialite` by `select load_extension('mod_spatialite');`. But when I tried to do the same in the project I got `Not Authorized` error. Not many useful suggestions online. 
WDYT?

vesperlou:
@nilsnolde @kevinkreiser No problem anymore. 
I loaded `mod_spatialite` in the project and solved all issues. But I had to remove `make_spatialite_cache` which was used to load spatialite as an extension because it stops bounding box query from working.
Feel free to leave any comments or reviews.

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,91 @@
...
+                       const double longitude,
+                       const double latitude);
+
+  bool get_landmarks_in_bounding_box(std::vector<Landmark>* landmarks,
+                                     const double minLat,
+                                     const double minLong,
+                                     const double maxLat,
+                                     const double maxLong);
+  bool test_select_query();
+  bool test_select_all();
```
kevinkreiser, line 55: i assume these will be removed

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,91 @@
...
+
+  bool open_database();
+  bool create_landmarks_table();
+  bool create_spatial_index();
+  void close_database();
+  bool connect_database();
```
kevinkreiser, line 61: we should make these protected and do them automatically based on lifetime of the object

test/gurka/test_build_landmark_database.cc
```diff
@@ -0,0 +1,61 @@
+#include <gtest/gtest.h>
```
kevinkreiser, line 1: can we rename this file to `test_landmarks.cc` then we can put everything in here not just the builder

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,81 @@
...
+
+#include <geos_c.h>
+
+namespace valhalla {
+namespace mjolnir {
+enum class AccessMode { ReadWriteCreate, ReadOnly, ReadWrite };
+
+struct Landmark {
+  std::string name;
+  std::string type;
```
kevinkreiser, line 27: in the next pr where you add parsing real data this will need to be come an enum of type uint8_t

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,81 @@
...
+  ~LandmarkDatabase() {
+    if (did_inserts && !vacuum_analyze()) {
+      LOG_ERROR("cannot do vacuum and analyze");
+    }
+    close_database();
+  }
+
+  bool insert_landmark(const Landmark& landmark);
+
+  bool get_landmarks_in_bounding_box(std::vector<Landmark>* landmarks,
```
kevinkreiser, line 49: this interface is surprising, i would have expected:

```suggestion
  std::vector<Landmark> get_landmarks_in_bounding_box(
```

if its possible that there is a failure we can log error and throw an exception. which actually  brings me to another good point, these methods should all have javadoc on them. have a look at some of the existing classes where they describe the returns and what each argument is and what the function generally does.

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,81 @@
...
+#include "baldr/graphconstants.h"
+#include "filesystem.h"
+#include "mjolnir/util.h"
+
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/geometries.hpp>
+#include <boost/geometry/geometries/register/point.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include <geos_c.h>
```
kevinkreiser, line 19: all of these except for ptree are probably not even used. we should avoid including things that we do not use, its not free. google the "include what you use" to learn more about why its a bad thing to take more than you need

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,81 @@
...
+
+namespace valhalla {
+namespace mjolnir {
+enum class AccessMode { ReadWriteCreate, ReadOnly, ReadWrite };
```
kevinkreiser, line 23: this isnt needed. the constructor should take a single boolean that maybe like: `read_only = true`

then in the constructor we can figure out the right thing to do automatically:

```pseudo code
if !exists(file)
  if read_only
    throw invalid option
  open_db(SQLITE_CREATE_FLAG | SQLITE_READWRITE_FLAG)
else
  open_db(read_only ? SQLITE_READONLY_FLAG : SQLITE_READWRITE_FLAG)
```

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,81 @@
...
+  bool did_inserts = false;
+
+  bool open_database();
+  bool create_landmarks_table();
+  bool create_spatial_index();
+  void close_database();
+  bool connect_database();
+  bool prepare_insert_stmt();
+  bool prepare_bounding_box_stmt();
+  bool vacuum_analyze();
```
kevinkreiser, line 76: honestly we should take all this crap and make a pimpl out of it but we can do it some other time. maybe i can pr that in my free time so you can move on to the next task

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+  ~LandmarkDatabase() {
+    if (did_inserts && !vacuum_analyze()) {
+      LOG_ERROR("cannot do vacuum and analyze");
+    }
+    close_database();
+  }
+
+  bool insert_landmark(const Landmark& landmark);
```
kevinkreiser, line 47: i know it seems convenient but it is wasteful for the parser to have to allocate (copy) the parsed data into a new structure just so you can push it into the database. if the builder itself takes a long time this is a place we can look to speed it up. for now we can leave it alone

src/mjolnir/landmark_database_builder.cc
```diff
@@ -0,0 +1,254 @@
...
+      break;
+    case AccessMode::ReadWrite:
+      open_flags = SQLITE_OPEN_READWRITE;
+      break;
+    default:
+      LOG_ERROR("invalid access mode");
+      return false;
+  }
+
+  if (!LandmarkDatabase::open_database()) {
```
kevinkreiser, line 36:
```suggestion
  if (!open_database()) {
```

src/mjolnir/landmark_database_builder.cc
```diff
@@ -0,0 +1,254 @@
...
+
+  if (!prepare_insert_stmt()) {
+    return false;
+  }
+  if (!prepare_bounding_box_stmt()) {
+    return false;
+  }
+
+  return true;
+}
```
kevinkreiser, line 57: tons of error checking overall, basically all of this stuff is fatal. just log error and throw. no need to return booling and checkign and the bubble the boolean out to a higher level. just the let the exception roll up the call stack

src/mjolnir/landmark_database_builder.cc
```diff
@@ -0,0 +1,254 @@
...
+    return true;
+  }
+
+  LOG_ERROR("Error: " + std::string(sqlite3_errmsg(db)));
+  sqlite3_close(db);
+  return false;
+}
+
+void LandmarkDatabase::close_database() {
+  sqlite3_close(db);
```
kevinkreiser, line 224: i think you need to release your prepared statments here too, i forget the function call but i remember it exists

kevinkreiser: `sqlite3_finalize`

valhalla/mjolnir/landmark_database_builder.h
vesperlou: I tried to use `std::ifstream` to decide if file exists but it didn't work very well. Do you have a better suggestion?

valhalla/mjolnir/landmark_database_builder.h
vesperlou: Thanks for pointing it out. They were copy paste from other files... I have removed the unused imports. It turns out we don't even need p_tree for now but I just comment it in case we need it in the next PR.

kevinkreiser (approved):

PR #4196, opened 2023-07-10
write traffic tile headers in valhalla_build_extract

nilsnolde:
Not the first time this bites me: `valhalla_build_extract` doesn't write the traffic tile headers yet and if one doesn't think about it in the external application, a long time can pass until realizing what the problem is (it checks for the tile version for quick sanity before reading any traffic).. Most apps will want to at least update the header with `last_update` (I don't this time), but the rest is static information and can be written by `valhalla_build_extract`:


kevinkreiser (approved):
much more convenient

Issue #4037, opened 2023-03-23
Directions with Landmark Support


kevinkreiser:
ref: #3999

## What is it and why do it?

Simply put, landmarks can be used to enhance a route and its directions with information about the routes surroundings. Landmarks could be almost anything that exists in the physical world but typically they are points of interest (POIs) such as a restaurant or gas station or perhaps a monument or statue. The why we want to do it is simillarly simple. In certain circumstances it can be advantageous to describe the route in terms of nearby POIs instead of just streets names/exit numbers/compass directions etc.

What follows are some key tenets of landmark use in directions. They are not absolutes but are merely something to keep in mind as we build this functionality.

### Visibility

The best, or most useful, landmarks are those that are easily visible from the road/path. Visibility is important because landmarks are most useful in the context of giving directions. For example, someone might ask you for directions, and to make it easier on them, you could say: 

> go about 5 miles and turn left at Steve's gas station

It wouldn't make sense for you to say

> go about 5 miles and turn left at the drainage ditch

The ditch is not the most salient thing you notice as that particular intersection. Interestingly, you also left out the road name. Do you even remember the roads name? Maybe not, but you totally notice the giant gas station on that corner.

### Verbosity

In the example above you also took care to only mention what was needed, you could have, but didn't say:

> go about 5 miles, you'll have passed a barber shop, a haunted house, 2 bus stops, and a shoe factory before turning left at Steve's gas station

Dumping all the information about all the landmarks one might notice along the way would be highly confusing. Instead you want the person to be on the lookout for a single landmark near the next maneuver point. Keeping the mental burdon of following directions low, by reducing the number of things one has to remember, makes it easier to do the right thing at the decision points in the route.

### Specificity

Notice we tried to be specific with the landmark, we didn't say:

> go about 5 miles and turn left at a gas station

There may have been many gas stations along that 5 mile stretch. It was advantageous to be as specific about which one it was as possible. The whole reason we are using landmarks to describe the route is to add clarity, not further ambiguity. In this case we were doubly lucky though in that the name of the POI also had the type of POI embedded in it. They now know its called "Steve's" so that should be written somewhere but also they know what a gas station looks like which makes it easier for them to filter down what they are looking at by type. In other words they only need to consider looking at gas stations to see what name is written on them not every landmark they see.

## How do we enable it?

[...]

PR #4201, opened 2023-07-14
Landmark parser

vesperlou:
# Issue

Implement landmark parser to parse landmarks in pbf and store them in database.

## Requirements / Relations

[Landmark Routing 2: Parse Landmarks](https://github.com/valhalla/valhalla/issues/4133)


vesperlou:
Hi @kevinkreiser @nilsnolde , I just finished task 2 for parsing and storing landmarks. Tests all passed :)
Not sure if I should split these stuff to seperate files like `admin builder`. Now the database, parser, and general function are all in `landmark_database_builder.cc` and `.h`

vesperlou:
Another question is in my tests the landmark `type` are parsed directly to `string` not `uint_64`. Is real data different or did I miss some steps to convert `string` to `uint_64`?

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -66,6 +65,14 @@ struct LandmarkDatabase {
   void release_prepared_stmt();
 };
 
+// class?
+struct LandmarkParser {
+public:
+  static std::vector<Landmark> Parse(const std::vector<std::string>& input_files);
+};
```
kevinkreiser, line 72: you can move this to the implementation file, its not needed for the public api, only the build function is needed which will make use of this parser callback thing, similar to the admin processing

kevinkreiser: oh i see this is the class analog. so in this case because you dont need any state, you can totally just remove this and put the guts of the work into the build function! sorry for the misunderstanding

src/mjolnir/landmark_database_builder.cc
```diff
@@ -169,10 +200,16 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
   int ret = sqlite3_step(bounding_box_stmt);
   while (ret == SQLITE_ROW) {
     const char* name = reinterpret_cast<const char*>(sqlite3_column_text(bounding_box_stmt, 0));
-    const char* type = reinterpret_cast<const char*>(sqlite3_column_text(bounding_box_stmt, 1));
+
+    LandmarkType type = LandmarkType::NA;
+    if (sqlite3_column_type(bounding_box_stmt, 1) != SQLITE_NULL) {
+      type = static_cast<LandmarkType>(sqlite3_column_int(bounding_box_stmt, 1));
+    }
+
```
vesperlou, line 208: I updated landmark type to enum of uint8_t which introduces some data type conversions like here.
I wonder if this could be a problem.

src/mjolnir/landmark_database_builder.cc
```diff
@@ -107,8 +138,8 @@ bool LandmarkDatabase::insert_landmark(const Landmark& landmark) {
     sqlite3_bind_null(insert_stmt, 1);
   }
 
-  if (landmark.type != "") {
-    sqlite3_bind_text(insert_stmt, 2, landmark.type.c_str(), landmark.type.length(), SQLITE_STATIC);
+  if (landmark.type != LandmarkType::NA) {
+    sqlite3_bind_int(insert_stmt, 2, static_cast<int>(landmark.type));
```
vesperlou, line 142: and here too

valhalla/mjolnir/landmark_database_builder.h
vesperlou: done.

src/mjolnir/landmark_database_builder.cc
```diff
@@ -1,8 +1,39 @@
...
+#include "mjolnir/osmpbfparser.h"
+
 namespace valhalla {
 namespace mjolnir {
+const std::string landmark_db = "landmarks.db";
+
+static const std::unordered_map<std::string, LandmarkType> str_type_map = {
+    {"parking", LandmarkType::parking},
+    {"bench", LandmarkType::bench},
+    {"parking_space", LandmarkType::parking_space},
```
kevinkreiser, line 13: i dont think any of these make sense really. maybe the first one whic could be a parking lot, but in reality it could but just some spaces on the side of the road. remember we are trying to call out things that are important, things people wont have to think about to see to navigate by

src/mjolnir/landmark_database_builder.cc
```diff
@@ -1,8 +1,39 @@
...
+    {"place_of_worship", LandmarkType::place_of_worship},
+    {"restaurant", LandmarkType::restaurant},
+    {"waste_basket", LandmarkType::waste_basket},
```
kevinkreiser, line 16: yeah no way are we using a trash can to navigate :smile:

src/mjolnir/landmark_database_builder.cc
```diff
@@ -1,8 +1,39 @@
...
+    {"bicycle_parking", LandmarkType::bicycle_parking},
```
kevinkreiser, line 17: same for this, doesnt seem relevant. maybe you are thinking about pedestrian routing? i suppose its possible that on unnamed trails smaller things like these could be worth using for navigation points?

src/mjolnir/landmark_database_builder.cc
```diff
@@ -1,8 +1,39 @@
...
+    {"fast_food", LandmarkType::fast_food},
+    {"cafe", LandmarkType::cafe},
+    {"fuel", LandmarkType::fuel},
+    {"shelter", LandmarkType::shelter},
+    {"recycling", LandmarkType::recycling},
+    {"toilets", LandmarkType::toilets},
```
kevinkreiser, line 23: these ones dont make much sense either

kevinkreiser:
> Now the database, parser, and general function are all in landmark_database_builder.cc and .h

in my opinion this is good. id prefer to keep all the stuff in one place. we already have a huge project the more like things that can go in one source the better imho


> Or I should encapsulate enum with uint8_t like enum class LandmarkType : uint8_t { parking = 0, restaurant = 1, };
yes thats what you should do. to do that we have a bunch of similar conversions using unordered maps. here is an example: https://github.com/valhalla/valhalla/blob/c3089cb52a05a61f3312747f88c49453b5069937/valhalla/baldr/graphconstants.h#L135-L161
you want the conversion in both directions

kevinkreiser:
i think for now we can focus on landmark navigation for motor vehicle routing and just ingest the larger most obvious types of amenities, bars, pubs, libraries, schools, restaurants, gas stations, banks, mechanics shops, firestation, post office, hotel. basically big buildings that shoudl be obviously marked so that the person doesnt have to look for them to navigate. think of it as something that should be more obvious than the road name sign itself

src/mjolnir/landmark_database_builder.cc
```diff
@@ -220,6 +257,90 @@ void LandmarkDatabase::release_prepared_stmt() {
...
+                          static_cast<OSMPBF::Interest>(OSMPBF::Interest::NODES |
+                                                        OSMPBF::Interest::CHANGESETS),
+                          callback);
+  }
+
+  // store landmarks in database
+  LandmarkDatabase db(landmark_db, false);
+
+  for (const auto& landmark : landmarks) {
+    if (!db.insert_landmark(landmark)) {
```
kevinkreiser, line 335: do this inside the callback for each one you parse. there is no sense in allocating a giant vector when you can just push a landmark into the db immediately as soon as its parsed. this means the parser needs the landmarkdatabase as a member though of course

src/mjolnir/landmark_database_builder.cc
kevinkreiser: oh and i should say by any i mean the top 3 here. if you noticed github lets you mark multiple lines but it doesnt actually highlight them when viewing the commment, but see above how it says +11 to +13. im not sure if you were aware of this github gotcha :smile:

vesperlou: yes I can see the highlighted 3 lines. 
For the landmrks here I just chose the 15 most popular ones. it make sense to carefully choose some obvious and straightfoward things. but from the amenity website most of the obvious landmarks are not very popular. we dont consider that as an issue right now?

src/mjolnir/landmark_database_builder.cc
vesperlou: make sense haha. trash cans can move, not good either.

src/mjolnir/landmark_database_builder.cc
vesperlou: good idea. I separated parser and insert because I wanted to test if parser is correct. now I know it is so can put insert into callback directly.

vesperlou:
> > Now the database, parser, and general function are all in landmark_database_builder.cc and .h
> 
> in my opinion this is good. id prefer to keep all the stuff in one place. we already have a huge project the more like things that can go in one source the better imho
> 
> > Or I should encapsulate enum with uint8_t like enum class LandmarkType : uint8_t { parking = 0, restaurant = 1, };
> > yes thats what you should do. to do that we have a bunch of similar conversions using unordered maps. here is an example: https://github.com/valhalla/valhalla/blob/c3089cb52a05a61f3312747f88c49453b5069937/valhalla/baldr/graphconstants.h#L135-L161
> > 
> > you want the conversion in both directions

Thanks for the reference! I already implemented string_to_landmark in the implementation file. should I implement the two func in the header like this one? Or implementation file is fine for they are not publicly exposed from my view.

vesperlou:
> i think for now we can focus on landmark navigation for motor vehicle routing and just ingest the larger most obvious types of amenities, bars, pubs, libraries, schools, restaurants, gas stations, banks, mechanics shops, firestation, post office, hotel. basically big buildings that shoudl be obviously marked so that the person doesnt have to look for them to navigate. think of it as something that should be more obvious than the road name sign itself

Make sense. maybe later we can add things for pedastrian navigation if everything works well.

src/mjolnir/landmark_database_builder.cc
vesperlou: correct: some of the obvious landmarks are still popular like "restaurant". so popularity is not an issue :)

kevinkreiser:
> should I implement the two func in the header like this one
for now its ok in the source but we'll need the other direction if we want to use the string in our narrative code "turn left at the church" where "church" comes from your enum to string conversion

vesperlou:
> > should I implement the two func in the header like this one
> 
> for now its ok in the source but we'll need the other direction if we want to use the string in our narrative code "turn left at the church" where "church" comes from your enum to string conversion

No extra efforts to put it in the header then! done.

src/mjolnir/landmark_builder.cc
```diff
@@ -220,6 +228,80 @@ void LandmarkDatabase::release_prepared_stmt() {
...
+  }
+
+  virtual void relation_callback(const uint64_t /*osmid*/,
+                                 const OSMPBF::Tags& /*tags*/,
+                                 const std::vector<OSMPBF::Member>& /*members*/) override {
+    LOG_WARN("relation callback shouldn't be called!");
+  }
+
+  LandmarkDatabase& db_;
+};
```
vesperlou, line 275: In admin builder the callback struct is put in anonymous namespace. should I do the same here? don't know the benefits though

vesperlou:
@kevinkreiser @nilsnolde I checked the project roadmap and found there is still one thing to do: decide whether parsing landmarks should be in a separate executable and be run separately. WDYT?

src/mjolnir/landmark_builder.cc
nilsnolde: Jep we should do the same here. In this case there'd be hardly a benefit, other than making it clear that it's not part of the public interface and rather implementation. If it's inside our named namespaces, I'd expect the callback struct's definition to be inside the header file too so I could import it into other files. Since we only use that struct in this .cc file, we should put it into a anonymous namespace.

test/gurka/test_landmarks.cc
```diff
@@ -13,15 +13,46 @@ using namespace valhalla::mjolnir;
...
+namespace {
+valhalla::gurka::map BuildPBF(const std::string& workdir) {
+  const std::string ascii_map = R"(
+      A-------B------C
+    )";
+
+  const gurka::nodes nodes = {
+      {"A", {{"name", ""}, {"amenity", "university"}}},
+      {"B", {{"name", "hai di lao"}, {"amenity", "restaurant"}}},
+      {"C", {{"name", "ke ji lu"}, {"amenity", ""}}},
```
nilsnolde, line 25: IMO this should be a degenerate case we want to throw early, as it has no meaning without a type.

src/mjolnir/landmark_builder.cc
```diff
@@ -220,6 +228,80 @@ void LandmarkDatabase::release_prepared_stmt() {
...
+        landmark.type = string_to_landmark_type(tag.second);
+      }
+      if (tag.first == "name" && !tag.second.empty()) {
+        landmark.name = tag.second;
+      }
+      landmark.lng = lng;
+      landmark.lat = lat;
+    }
+    // insert parsed landmark directly into database
+    db_.insert_landmark(landmark);
```
nilsnolde, line 256: doesn't this save _all_ nodes right now?! 

Also I'd not handle the case where `amenity` has no value. I'm not even sure that can happen in actual OSM data (but surely can in custom OSM data). Even there is smth like that, it has no value for us IMO.

valhalla/mjolnir/landmark_builder.h
```diff
@@ -0,0 +1,164 @@
...
+  auto it = landmark_type_to_string.find(static_cast<uint8_t>(t));
+  if (it != landmark_type_to_string.cend()) {
+    return it->second;
+  }
+  return "null";
+}
+
+struct Landmark {
+  std::string name;
+  LandmarkType type;
```
nilsnolde, line 102: the way we're currently using this struct, we should default at least the name & type or you'll get random stuff in there if they're never set.

test/gurka/test_landmarks.cc
nilsnolde: let's also include a node that shouldn't be saved. Maybe we can make this more realistic by e.g.

```
      a-------b------c
          A    B     C
    )";
```
So A, B, C would be on the side of the road, and a, b, c are simply graph nodes without any tags (which shouldn't appear in the db, but I think do currently, see above).

src/mjolnir/landmark_builder.cc
nilsnolde: We could instead do a LOG_ERROR when there's an amenity node without a value.

src/mjolnir/landmark_builder.cc
kevinkreiser: unnamed namespace is best for two reasons:

1. this is not in the valhalla/mjolnir namespace because we want it to remain an implementation detail of the library and not something in the public interface
2. adding it to a namespace makes it local to that translation unit and as such avoids potential naming clashes when linking

src/mjolnir/landmark_builder.cc
kevinkreiser: agreed, it has to have a type otherwise its meaningless

kevinkreiser:
> I checked the project roadmap and found there is still one thing to do: decide whether parsing landmarks should be in a separate executable and be run separately. WDYT?

originally i thought it should be just like admin building that we run it once in a separate process but now im thinking it might be nice to just turn on for regular tile builds. i was thinking like this


```json
{
  "mjolnir": {
    ...
    "parse_landmarks": false, // if you turn this on the regular osm parsing callback parses the nodes for landmarks
    "landmarks_db": the/path/to/the/db.sqlite, // this would be the location it writes to if parse is true and it will be the location it reads from in the enhancer
    ...
  }
}
```

this would mean a couple changes for this pr, but essentially we'd just add a tiny bit of code to the existing pbf graphparser.

alternatively we could make another executable called `valhalla_build_landmarks` and have it run separately. i will say that the reason we do that for admins is because they dont change that often and actually we dont like to take changes from them too often because people will break them on occasion. with landmarks i think we'd always like to get the latest that exists in the data regardless which makes me think it should just happen (still optionally) in the parse stage with the rest of the graph data

@nilsnolde what do you think about that?

nilsnolde:
Hm I think our bigger idea was that eventually all these graph enhancements like landmarks/transit/elevation would be best to have as external executables rather than cramming it all into the regular build pipeline. Comes with an overhead (parsing PBF again in this case), but that’d be fine IMO. 

So I think I’d advocate to keep it separate and have it enrich an existing built graph if possible. But I might overlook smth crucial?

nilsnolde:
We could still have it in build pipeline as very last steps so people don’t get too confused on what they need to run in which order, but IMO rather not tightly coupled with the initial graph build.

kevinkreiser:
separate is totally fine but yeah if I were running the system I'd hate to spend a couple hours parsing data before being able to run the tile build. my suggestion is you can still make and bring your sqlite file from. wherever outside process but we could offer the option to build it along with the regular parsing step so that it's up to date and doest require extra time to make. allow both use cases in other words

test/gurka/test_landmarks.cc
vesperlou: good example I will update that.

src/mjolnir/landmark_builder.cc
vesperlou: thanks for pointing this out! I totally forgot to include landmark nodes only and made a premise that not all landmark nodes have amenity value which is incorrect. will fix that.

valhalla/mjolnir/landmark_builder.h
vesperlou: ah right I forgot default / initialization again. let's set default name as `"null"` and default type as `LandmarkType::null` :D

vesperlou:
ok I will allow both use cases. it makes sense to offer choices for various situations and is what I would do. 
just need to fidure out how. already have some clues :D

src/mjolnir/landmark_builder.cc
vesperlou: a further point: in the landamark database should we only store nodes that belong to `enum class LandmarkType`? Now the list contians 21 popular landmark types and others will be translated into `LandmarkType::null` which is useless in navigation. plus we can shorten execution time and save storage by storing "known" nodes only.

kevinkreiser: exactly right, we should skip anything that is useless, no point in even looking at it if we cant parse the string into one of the enums that we accept

valhalla/mjolnir/landmark_builder.h
kevinkreiser: i disagree here. i think we shouldnt even h ave this struct but rather for the insert we can take the parameters in the function (why make a copy when we dont need to) and then when returning we can just return a tuple. as far as defaults, name can either exist or be empty we dont need a name for all amenities the type is good enough for the majority. default location makes no sense, we should never have a landmark that doesnt have a location. again i think we can avoid this by simply removing this structure completely from the interface

src/mjolnir/landmark_builder.cc
vesperlou: got it. done.

valhalla/mjolnir/landmark_builder.h
vesperlou: I can remove the struct. it simply makes the code prettier and easier so no problem. As for default value of name I think having a default `"null" `or `""` or none at all are interchangeable to me. which one is better from your opinion?
but I don't know how to check validity of lat and lng. other implementation files also don't deal with that so I assume it is either checked in parser or just doesn't exist, or both. so for `lat` and `lng` maybe we can keep the current with no defaults. WDYT? @nilsnolde

kevinkreiser: in the database, it can be NULL but in code it has to be the empty string. therefor to make our lives easier it could also be empty string in the database

test/gurka/test_landmarks.cc
```diff
@@ -13,16 +13,58 @@ using namespace valhalla::mjolnir;
...
+  const std::string ascii_map = R"(
+      a-----b-----c---d
+      A   B   C   D   E
+    )";
+
+  const gurka::nodes nodes = {
+      {"A", {{"landmark", "yes"}, {"name", ""}, {"amenity", "university"}}},
+      {"B", {{"landmark", "yes"}, {"name", "hai di lao"}, {"amenity", "restaurant"}}},
+      {"C", {{"landmark", "yes"}, {"name", "ke ji lu"}, {"amenity", ""}}}, // no amenity, log error
+      {"D", {{"landmark", "yes"}, {"name", "wan da"}, {"amenity", "cinema"}}},
```
nilsnolde, line 27: also  don't forget to remove the `landmark` in the tests

test/gurka/test_landmarks.cc
```diff
@@ -13,16 +13,58 @@ using namespace valhalla::mjolnir;
...
+      {"a", {{"landmark", "no"}, {"name", "gong ce"}, {"amenity", "toilet"}}},
+      {"b", {{"landmark", ""}, {"name", ""}, {"amenity", "trash_can"}}},
+      {"c", {{"name", "hua yuan"}, {"amenity", ""}}},
+      {"d", {{"name", ""}, {"amenity", ""}}},
+  };
+
+  const gurka::ways ways = {
+      {"ab", {}},
+      {"bc", {}},
+      {"cd", {}},
```
nilsnolde, line 42: let's give it at least a `{"highway", "residential"}` or some other highway attribute

test/gurka/test_landmarks.cc
vesperlou: meeting ends in one hour. we can discuss here @kevinkreiser @nilsnolde 
I think the `landmark` key is used to select landmark nodes as you suggested earlier we should only parse landmarks nodes otherwise there are too many. without landmark key how can we achieve that?

nilsnolde: Only the nodes which have a "amenity" tag having a value of the map you specified in your new header (hospital, restaurant etc) are the nodes we want to save. earlier you were just saving _every_ node. however, the `landmark` tag is useless for us. the gist is, we want to define what we consider a "landmark" and not leave that definition to random OSM mappers. does that make sense now?

vesperlou: ah right I understand now. thanks!

src/mjolnir/landmark_builder.cc
```diff
@@ -1,10 +1,67 @@
...
+  landmark_callback(valhalla::mjolnir::LandmarkDatabase& db) : db_(db) {
+  }
+  virtual ~landmark_callback() {
+  }
+
+  virtual void
+  node_callback(const uint64_t osmid, double lng, double lat, const OSMPBF::Tags& tags) override {
+    // parse "landmark" nodes only
+    auto iter = tags.find("landmark");
+    if (iter != tags.cend() && iter->second == "yes") {
```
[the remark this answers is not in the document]
vesperlou: no problem very clear to me :D

test/gurka/test_landmarks.cc
vesperlou: done

vesperlou:
@nilsnolde @kevinkreiser I finished addressing all issues discussed today, feel free to check :D
seems like CI is down for a while, clicking details merely returns 421 Error. maybe we should rerun the failed CI task.

I searched tests for command line but didn't find any reference, so I did a manual test to proove it works as expected. should I write test for `valhalla_build_landmarks` cli? 
and I think the current tests look a bit messy. the first test is about building database and bounding box only so I don't want to mix its data with others (mixing could be a problem in design of bounding box query which is often used to check correctness). In fact, I don't want to mix data from any single test, but all tests should have database initialization and file removal. repeating code is ugly... is there a better way to write the tests?

src/mjolnir/landmark_builder.cc
```diff
@@ -1,10 +1,63 @@
...
+struct landmark_callback : public OSMPBF::Callback {
+public:
+  landmark_callback(const std::string& db_name) : db_(db_name, false) {
+  }
+  virtual ~landmark_callback() {
+  }
+
+  virtual void
+  node_callback(const uint64_t /*osmid*/, double lng, double lat, const OSMPBF::Tags& tags) override {
+    valhalla::mjolnir::Landmark landmark;
```
nilsnolde, line 17: this wasn't addressed yet: we don't need this intermediate struct at all. just put the properties into the `insert_landmark` call.

src/mjolnir/valhalla_build_landmarks.cc
```diff
@@ -0,0 +1,70 @@
...
+  // configure logging
+  auto logging_subtree = pt.get_child_optional("mjolnir.logging");
+  if (logging_subtree) {
+    auto logging_config =
+        valhalla::midgard::ToMap<const boost::property_tree::ptree&,
+                                 std::unordered_map<std::string, std::string>>(logging_subtree.get());
+    valhalla::midgard::logging::Configure(logging_config);
+  }
+
+  // if (!valhalla::mjolnir::BuildAdminFromPBF(pt.get_child("mjolnir"), input_files)) {
```
nilsnolde, line 64: remove the comment

nilsnolde (changes requested):
To expand a bit on the executable testing: we don't do that, though it can be done, but it's fairly much effort and not much point as long as 99% of the logic is outside the executable, which is why we advocate for that.

There's a few more things to address.

nilsnolde:
FYI, the OSX failure will be fixed with #4209

test/gurka/test_landmarks.cc
```diff
@@ -2,60 +2,142 @@
...
+      a-----b-----c---d
+      A   B   C   D   E
+    )";
+
+  const gurka::nodes nodes = {
+      {"A", {{"name", ""}, {"amenity", "university"}}},
+      {"B", {{"name", "hai di lao"}, {"amenity", "restaurant"}}},
+      {"C", {{"name", "ke ji lu"}, {"amenity", ""}}}, // no amenity, shouldn't be stored
+      {"D", {{"name", "wan da"}, {"amenity", "cinema"}}},
+      {"E", {{"name", "zhong lou"}, {"amenity", "monument"}}}, // not in list, shouldn't be stored
```
nilsnolde, line 29: (accidentally?) you gave a VERY good example of OSM stuff we'd also like to save apart from `amenity`. See [`historic` key](https://taginfo.openstreetmap.org/keys/historic#values) which also contains `monument`. If the name is descriptive enough that's the perfect guidance POI:)

But let's shelve that for a later day: https://github.com/valhalla/valhalla/issues/4210

vesperlou: you remind me it's not an accident :) I was discovering osm key-values and found `monument`, thought it could be interesting to have it as landmarks but forgot to discuss with you. no problem let's discuss later.

src/mjolnir/valhalla_build_landmarks.cc
vesperlou: done

src/mjolnir/landmark_builder.cc
nilsnolde: let's keep it unresolved until we can see it's done :) best to let the reviewers resolve comments, then they don't lose the overview of their own comments

vesperlou: no problem. thanks for the tip! 
and it's fixed now.

src/mjolnir/landmark_builder.cc
```diff
@@ -104,28 +157,21 @@ LandmarkDatabase::LandmarkDatabase(const std::string& db_name, bool read_only)
...
-  } else {
-    sqlite3_bind_null(insert_stmt, 2);
-  }
-
-  sqlite3_bind_double(insert_stmt, 3, landmark.lng);
-  sqlite3_bind_double(insert_stmt, 4, landmark.lat);
+  sqlite3_bind_text(insert_stmt, 1, name.c_str(), name.length(), SQLITE_STATIC);
+  sqlite3_bind_int(insert_stmt, 2, static_cast<int>(type));
+  sqlite3_bind_double(insert_stmt, 3, lng);
+  sqlite3_bind_double(insert_stmt, 4, lat);
```
vesperlou, line 174: I removed the empty checks for name and type here because in common use cases `insert_landmark` should only be called by `node_callback` where checks and initializations already happen.

src/mjolnir/landmark_builder.cc
```diff
@@ -153,10 +199,17 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
   int ret = sqlite3_step(bounding_box_stmt);
   while (ret == SQLITE_ROW) {
     const char* name = reinterpret_cast<const char*>(sqlite3_column_text(bounding_box_stmt, 0));
-    const char* type = reinterpret_cast<const char*>(sqlite3_column_text(bounding_box_stmt, 1));
+
+    LandmarkType type = LandmarkType::null;
+    if (sqlite3_column_type(bounding_box_stmt, 1) != SQLITE_NULL) {
+      type = static_cast<LandmarkType>(sqlite3_column_int(bounding_box_stmt, 1));
+    }
```
nilsnolde, line 206: it shouldn't happen that the type is empty. IIRC we throw away nodes that don't have an `amenity` type no? I'd consider that a bug if that happens, but pretty sure it can't

src/mjolnir/landmark_builder.cc
```diff
@@ -168,5 +221,44 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
...
+  if (!db) {
+    LOG_ERROR("Landmarks config info not found. Landmarks builder will not be created.");
+    return false;
+  }
+
+  const filesystem::path parent_dir = filesystem::path(*db).parent_path();
+  if (!filesystem::exists(parent_dir) && !filesystem::create_directories(parent_dir)) {
+    LOG_ERROR("Can't create parent directory " + parent_dir.string());
+    return false;
+  }
```
nilsnolde, line 237: we should remove the database right here if it exists

nilsnolde (changes requested):
Sorry, now with the new changes I somehow have quite a few more questions.

valhalla/mjolnir/landmark_builder.h
```diff
@@ -0,0 +1,129 @@
...
+#include <string>
+#include <unordered_map>
+
+namespace valhalla {
+namespace mjolnir {
+static const std::string default_landmark_name = "";
+
+// obvious landmark types for vehicle routing
+enum class LandmarkType : uint8_t {
+  null = 0,
```
kevinkreiser, line 17:
```suggestion
```

valhalla/mjolnir/landmark_builder.h
```diff
@@ -0,0 +1,129 @@
...
+       {"college", LandmarkType::college},
+       {"university", LandmarkType::university},
+       {"theatre", LandmarkType::theatre},
+       {"cinema", LandmarkType::cinema}};
+
+  auto it = string_to_landmark_type.find(s);
+  if (it != string_to_landmark_type.cend()) {
+    return it->second;
+  }
+  return LandmarkType::null;
```
kevinkreiser, line 71: we can just disallow conversion when its unknown
```suggestion
  if (it == string_to_landmark_type.cend())
    throw std::runtime_error("unknown landmark type");
  return it->second;
```

valhalla/mjolnir/landmark_builder.h
```diff
@@ -0,0 +1,129 @@
...
+       {static_cast<uint8_t>(LandmarkType::university), "university"},
+       {static_cast<uint8_t>(LandmarkType::theatre), "theatre"},
+       {static_cast<uint8_t>(LandmarkType::cinema), "cinema"}};
+
+  auto it = landmark_type_to_string.find(static_cast<uint8_t>(t));
+  if (it != landmark_type_to_string.cend()) {
+    return it->second;
+  }
+  return "null";
+}
```
kevinkreiser, line 104: i know i said we would need this eventually but we dont need it yet right? on second thought i think we will never need it because these things will need to be localized to be used in directions like "turn left at the gas station". we have a wholly other way of getting those types of things into the response so i think we can remove this unless there is some other use?
```suggestion
```

src/mjolnir/landmark_builder.cc
kevinkreiser: yeah i was trying to say before that if we open the database in non-read-only mode and its there we should just blow it away. that can happen inside the database construction imho, unless we see some future need to be able to append to the database?

src/mjolnir/landmark_builder.cc
kevinkreiser, line 206:
```suggestion
    int landmark_type = -1;
    if (sqlite3_column_type(bounding_box_stmt, 1) != SQLITE_NULL) {
      landmark_type = sqlite3_column_int(bounding_box_stmt, 1);
    }
```

kevinkreiser: agree, its probably not needed to check this

src/mjolnir/landmark_builder.cc
```diff
@@ -153,10 +199,17 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
...
+    LandmarkType type = LandmarkType::null;
+    if (sqlite3_column_type(bounding_box_stmt, 1) != SQLITE_NULL) {
+      type = static_cast<LandmarkType>(sqlite3_column_int(bounding_box_stmt, 1));
+    }
+
     double lng = sqlite3_column_double(bounding_box_stmt, 2);
     double lat = sqlite3_column_double(bounding_box_stmt, 3);
-    landmarks.emplace_back(Landmark{name, type, lng, lat});
+
+    landmarks.emplace_back(std::make_tuple(name, type, lng, lat));
```
kevinkreiser, line 211:
```suggestion
    landmarks.emplace_back(std::make_tuple(name, static_cast<LandmarkType>(landmark_type), lng, lat));
```

kevinkreiser:
this looks just about done, but we should very much vet the list of amenities before we call it finished because those will be settled for ever once we bake them into tiles. maybe we should shorten the list to just the ones we are super sure about and add ones as we encounter useful examples.

anyway other than some nitpicks its there

valhalla/mjolnir/landmark_builder.h
```diff
@@ -0,0 +1,129 @@
...
+  hospital = 10,
+  post_office = 11,
+  pub = 12,
+  clinic = 13,
+  police = 14,
+  fire_station = 15,
+  car_wash = 16,
+  charging_station = 17,
+  college = 18,
+  university = 19,
```
kevinkreiser, line 36: these things are probably too large to use to navigate, id drop them

src/mjolnir/landmark_builder.cc
vesperlou: agree. for now it makes sense to allow writing only when a database is created. I'll update the code.

valhalla/mjolnir/landmark_builder.h
vesperlou: you are right, we don't need it yet. will delete it. 
how do you plan to do the wholly other way?

valhalla/mjolnir/landmark_builder.h
vesperlou: make sense

valhalla/mjolnir/landmark_builder.h
vesperlou: I just realize this function is called in `node_callback` to check every node with amenity. There are many amenity values not included in the list so we shouldn't throw. WDYT? @kevinkreiser if we remove LandmarkType::null how to do the check?

vesperlou: ah we can use try catch

valhalla/mjolnir/landmark_builder.h
kevinkreiser: its a large topic, once we have this stuff in the tiles we can go over it. if you are interested now you can go look at https://github.com/valhalla/valhalla/blob/master/locales/README.md for a start

valhalla/mjolnir/landmark_builder.h
kevinkreiser: exactly

valhalla/mjolnir/landmark_builder.h
```diff
@@ -0,0 +1,90 @@
...
+  post_office = 10,
+  pub = 11,
+  clinic = 12,
+  police = 13,
+  fire_station = 14,
+  car_wash = 15,
+  theatre = 16,
+  cinema = 17,
+  casino = 18,
+};
```
kevinkreiser, line 35: i disagree that some of these will be obvious, in fact i think some of them require a name. id really like to stress that reading what i wrote in the initial issue #4037 regarding landmarks, specifically the visibility/verbosity/specificity sections. we aren't making a poi database we are trying to pick features that might stand out more than a road sign because at the decision point in the route you have a high pressure situation where you must decide quickly. if we are at all unsure about a specific amenity type then we should err on not using it. anyway here are some of my thoughts on this inline:

```suggestion
// obvious landmark types for vehicle routing
enum class LandmarkType : uint8_t {
  // these will almost always be obvious by their function and don't require a name
  fuel = 1,
  post_office = 2,
  police = 3,
  fire_station = 4,
  car_wash = 5,

  // could be obvious if its a chain but could also just be a building
  restaurant = 6,
  // this will always be a chain in which case the name is a required distinguisher (as they are often next to each other on either side of a street)
  fast_food = 7,
  // this could be a chain making it easier to distinguish (starbucks) but could be a one off cafe in which case the name will be required
  cafe = 8,
  // on the fence about this one, these are often obvious without a name, maybe should go in the upper list?
  bank = 10,
  // same with this one, in some countries these are obvious and have the little plus sign, in america its not the case and they can be different chains or could be part of a grocery store or anything... speaking of grocery store, why isnt that on the list? also why arent department stores on the list, surely stores would be a good landmark to use (again when they are named)
  pharmacy = 11,
  // again these could just be a random building with nothing other than a sign like "early years" or something, they better have a name or it will be ambiguous
  kindergarten = 12,
  // same thing, often just a building
  bar = 13,
  // this could be very very big like a college campus, i think we should remove this
  hospital = 14,
  // same as bar could be just a building need at least a name to have a chance of seeing it
  pub = 16,
  // generic, could be manyy many types of clinics also often there are multiple of these located in same "complex", you might have a dentist, orthodontist, ear nose throat and eye specialists all next to each other. 
  clinic = 17,
  // on the fence about this one, to me this would be in the same category as "museum" or something where its so specialized maybe it doesnt need a name. what keeps me from say ing that is you often cant tell from the outside what function these amenities serve until you read the name
  theatre = 16,
  // this could work without a name in my opinion, they often have a board out front that shows what movies are playing which would probably remove the need for a name
  cinema = 17,
  // these are usually quite large as well but not usually multiple buildings, how often are they located near one another though? if its often perhaps we need a name to differentiate between this casino and the next one down on the next block? maybe that argument is a good argument for everything has to have a name?
  casino = 18,
};
```

src/mjolnir/landmark_builder.cc
```diff
@@ -1,10 +1,61 @@
-#include "mjolnir/landmark_database_builder.h"
+#include "mjolnir/landmark_builder.h"
 #include "filesystem.h"
+
+#include "mjolnir/osmpbfparser.h"
 #include "mjolnir/util.h"
+#include <boost/property_tree/ptree.hpp>
```
kevinkreiser, line 6: you already have this in the header
```suggestion
```

src/mjolnir/landmark_builder.cc
```diff
@@ -1,10 +1,61 @@
...
+      } catch (const std::runtime_error& e) { return; }
+
+      std::string name = "";
+      auto it = tags.find("name");
+      if (it != tags.cend() && !it->second.empty()) {
+        name = it->second;
+      }
+
+      // insert parsed landmark directly into database
+      db_.insert_landmark(name, type, lng, lat);
```
kevinkreiser, line 34: you can avoid this clunkiness by doing the whole thing inside the try catch. also dont use the name `type` as a variable name its just too close to keywords and likely other things it could collide with.

```suggestion
      try {
        auto landmark_type = valhalla::mjolnir::string_to_landmark_type(iter->second);
      
        std::string name = "";
        auto it = tags.find("name");
        if (it != tags.cend() && !it->second.empty()) {
          name = it->second;
        }

        // insert parsed landmark directly into database
        db_.insert_landmark(name, landmark_type, lng, lat);
      } catch (const std::runtime_error& e) { return; }
```

src/mjolnir/landmark_builder.cc
```diff
@@ -1,10 +1,61 @@
...
+    }
+  }
+
+  virtual void changeset_callback(const uint64_t changeset_id) override {
```
kevinkreiser, line 38:
```suggestion
  virtual void changeset_callback(const uint64_t /*changeset_id*/) override {
```

src/mjolnir/landmark_builder.cc
```diff
@@ -168,5 +221,44 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
...
 }
 
+bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
+                          const std::vector<std::string>& input_files) {
+  // parse config to get landmark database
+  auto db = pt.get_optional<std::string>("landmarks");
+  if (!db) {
+    LOG_ERROR("Landmarks config info not found. Landmarks builder will not be created.");
+    return false;
+  }
```
kevinkreiser, line 231: its not bad to have this but its not needed exactly.. the config generator will always put a value here, one option would be to do:

```suggestion
  auto db = pt.get<std::string>("landmarks", "");
```

then when we got to make the db it should fail with an invalid path

src/mjolnir/landmark_builder.cc
kevinkreiser, line 237: ok so we have the removal of existing db in the constructor then we need this create the directory if it doesnt exist also in the constructor, it doesnt make sense for the two things to be decoupled

test/gurka/test_landmarks.cc
```diff
@@ -1,61 +1,128 @@
...
+      a-----b-----c---d
+      A   B   C   D   E
+    )";
 
-class LandmarkDatabaseTest : public ::testing::Test {
-protected:
-  static void SetUpTestSuite() {
-    LandmarkDatabase db(db_name, false);
+  const gurka::nodes nodes = {
+      {"A", {{"name", ""}, {"amenity", "bar"}}},
```
kevinkreiser, line 23: in real life osm will not store a tag without a value, so if you want to test this then we should just omit the name tag completely:

```suggestion
      {"A", {{"amenity", "bar"}}},
```

src/mjolnir/landmark_builder.cc
vesperlou: got it. thanks

src/mjolnir/landmark_builder.cc
vesperlou: nice suggestion thanks

src/mjolnir/landmark_builder.cc
vesperlou: much more concise

src/mjolnir/landmark_builder.cc
vesperlou: make sense. done.

test/gurka/test_landmarks.cc
vesperlou: good to know!

test/gurka/test_landmarks.cc
```diff
@@ -1,61 +1,128 @@
...
+      {"B", {{"name", "hai di lao"}, {"amenity", "restaurant"}}},
+      {"C", {{"name", "ke ji lu"}, {"amenity", ""}}}, // no amenity, shouldn't be stored
+      {"D", {{"name", "wan da"}, {"amenity", "cinema"}}},
+      {"E", {{"name", "zhong lou"}, {"amenity", "monument"}}}, // not in list, shouldn't be stored
+      // non-landmark nodes
+      {"a", {{"name", "gong ce"}, {"amenity", ""}}},
+      {"b", {{"name", "la ji tong"}}},
+      {"c", {{"name", "hua yuan"}}},
+      {"d", {{"name", ""}, {"amenity", ""}}},
```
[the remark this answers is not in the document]
vesperlou: sure it makes sense to have extra tags and thanks for adding those

valhalla/mjolnir/landmark_builder.h
vesperlou: I've added the remarks. I feel this is going to be a long term discussion and decision :)

valhalla/mjolnir/landmark_builder.h
```diff
@@ -0,0 +1,116 @@
...
+  // again these could just be a random building with nothing other than a sign like "early years" or
+  // something, they better have a name or it will be ambiguous
+  kindergarten = 12,
+  // same thing, often just a building
+  bar = 13,
+  // this could be very very big like a college campus, i think we should remove this
+  hospital = 14,
+  // same as bar could be just a building need at least a name to have a chance of seeing it
+  pub = 16,
+  // generic, could be manyy many types of clinics also often there are multiple of these located in
```
kevinkreiser, line 45:
```suggestion
  // generic, could be many many types of clinics also often there are multiple of these located in
```

src/mjolnir/landmark_builder.cc
```diff
@@ -16,12 +66,20 @@ struct LandmarkDatabase::db_pimpl {
   bool vacuum_analyze = false;
 
   db_pimpl(const std::string& db_name, bool read_only) : insert_stmt(nullptr) {
-    // figure out if we need to create it or can just open it up
-    auto flags = read_only ? SQLITE_OPEN_READONLY : SQLITE_OPEN_READWRITE;
+    // create parent directory if not exist
```
kevinkreiser, line 69:
```suggestion
    // create parent directory if it doesn't exist
```

src/mjolnir/landmark_builder.cc
```diff
@@ -16,12 +66,20 @@ struct LandmarkDatabase::db_pimpl {
...
+    }
+
+    // figure out if we need to create database or can just open it up
+    auto flags = read_only ? SQLITE_OPEN_READONLY : SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
     if (!filesystem::exists(db_name)) {
       if (read_only)
         throw std::logic_error("Cannot open sqlite database in read-only mode if it does not exist");
-      flags |= SQLITE_OPEN_CREATE;
+    } else if (!read_only) {
+      std::filesystem::remove(db_name);
```
kevinkreiser, line 81: long story but we only just went to c++17 which as non-experimental support for filesystem. before that we use valhalla::filesystem, our internal home grown version. we've not migrated to std::filesystem yet so lets use our own until we do so

```suggestion
      filesystem::remove(db_name);
```

src/mjolnir/landmark_builder.cc
```diff
@@ -1,10 +1,60 @@
-#include "mjolnir/landmark_database_builder.h"
+#include "mjolnir/landmark_builder.h"
 #include "filesystem.h"
+
+#include "mjolnir/osmpbfparser.h"
 #include "mjolnir/util.h"
+#include <filesystem>
```
kevinkreiser, line 6:
```suggestion
```

kevinkreiser (changes requested):
i had some tiny nit picks yet and then this is ready to go

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
    * ADDED: `source_percent_along` & `target_percent_along` to /trace_attributes JSON response [#4199](https://github.com/valhalla/valhalla/pull/4199)
    * ADDED: sqlite database to store landmarks along with interfaces of insert and bounding box queries [#4189](https://github.com/valhalla/valhalla/pull/4189)
    * CHANGED: refactor landmark database interface to use a pimpl [#4202](https://github.com/valhalla/valhalla/pull/4202)
+   * ADDED: add landmark parser to parse POI from pbf files and store them as landmarks in landmark database [#4201](https://github.com/valhalla/valhalla/pull/4201)
```
kevinkreiser, line 19:
```suggestion
   * ADDED: add `valhalla_build_landmarks` to parse POIs from osm pbfs and store them as landmarks in the landmark sqlite database [#4201](https://github.com/valhalla/valhalla/pull/4201)
```

kevinkreiser (approved):

nilsnolde (approved):

PR #4202, opened 2023-07-14
make landmark db operations a pimpl

kevinkreiser:
The pimpl idiom is a powerful one especially for api design. It allows the designer to offer up the api interface without exposing the details of how its implemented, effectively giving the api a PrivateIMPLementation.

I think this best suits pretty much all of our c++ apis but sadly not all of them adhere to this. At any rate this is an attempt to refactor the recent work in #4189 to model this.

In addition to the pimpl I cut down logging as much as possible, I only close the db in the destructor and instead of returning booleans I throw when there is a problem that is unrecoverable

By and large the code is pretty much the same just leaner and meaner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
    * CHANGED: write traffic tile headers in `valhalla_build_extract` [#4195](https://github.com/valhalla/valhalla/pull/4195)
    * ADDED: `source_percent_along` & `target_percent_along` to /trace_attributes JSON response [#4199](https://github.com/valhalla/valhalla/pull/4199)
    * ADDED: sqlite database to store landmarks along with interfaces of insert and bounding box queries [#4189](https://github.com/valhalla/valhalla/pull/4189)
+   * CHANGED: refactor landmark database interface to use a pimpl [#4202](https://github.com/valhalla/valhalla/pull/4202)
 
 ## Release Date: 2023-05-11 Valhalla 3.4.0
 * **Removed**
--- a/src/mjolnir/landmark_database_builder.cc
+++ b/src/mjolnir/landmark_database_builder.cc
@@ -1,7 +1,6 @@
 #include "mjolnir/landmark_database_builder.h"
-#include "mjolnir/util.h"
 #include "filesystem.h"
-
+#include "mjolnir/util.h"
 
 namespace valhalla {
 namespace mjolnir {
@@ -16,7 +15,7 @@ struct LandmarkDatabase::db_pimpl {
   std::shared_ptr<void> spatial_lite;
   bool vacuum_analyze = false;
 
-  db_pimpl(const std::string& db_name, bool read_only): insert_stmt(nullptr) {
+  db_pimpl(const std::string& db_name, bool read_only) : insert_stmt(nullptr) {
     // figure out if we need to create it or can just open it up
     auto flags = read_only ? SQLITE_OPEN_READONLY : SQLITE_OPEN_READWRITE;
     if (!filesystem::exists(db_name)) {
@@ -38,7 +37,8 @@ struct LandmarkDatabase::db_pimpl {
     char* err_msg = nullptr;
     if (flags & SQLITE_OPEN_CREATE) {
       // make the table
-      const char* table = "SELECT InitSpatialMetaData(1); CREATE TABLE IF NOT EXISTS landmarks (name TEXT, type TEXT)";
+      const char* table =
+          "SELECT InitSpatialMetaData(1); CREATE TABLE IF NOT EXISTS landmarks (name TEXT, type TEXT)";
       ret = sqlite3_exec(db, table, NULL, NULL, &err_msg);
       if (ret != SQLITE_OK) {
         sqlite3_free(err_msg);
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
    * CHANGED: write traffic tile headers in `valhalla_build_extract` [#4195](https://github.com/valhalla/valhalla/pull/4195)
    * ADDED: `source_percent_along` & `target_percent_along` to /trace_attributes JSON response [#4199](https://github.com/valhalla/valhalla/pull/4199)
    * ADDED: sqlite database to store landmarks along with interfaces of insert and bounding box queries [#4189](https://github.com/valhalla/valhalla/pull/4189)
+   * CHANGED: refactor landmark database interface to use a pimpl [#4202](https://github.com/valhalla/valhalla/pull/4202)
```
kevinkreiser, line 18: @vesperlou the diff for the source files below is large so its hard to read. i would suggest just looking at the files directly (in the 3 dots on the side you can click "view file")

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -15,55 +16,16 @@ struct Landmark {
...
-    }
-    release_prepared_stmt();
-    close_database();
-  }
-
-  bool insert_landmark(const Landmark& landmark);
-
+  LandmarkDatabase(const std::string& db_name, bool read_only);
+  void insert_landmark(const Landmark& landmark);
   std::vector<Landmark> get_landmarks_in_bounding_box(const double minLat,
```
kevinkreiser, line 21: the interface becomes just the functions that we want to publicly expose. the only thing the outside caller can see that is protected is a pimpl which gives no clue about its implementation

src/mjolnir/landmark_database_builder.cc
```diff
@@ -1,103 +1,114 @@
...
-  }
 
-  if (open_flags == (SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE)) {
-    if (!create_landmarks_table() || !create_spatial_index()) {
-      throw std::runtime_error("Cannot create landmarks table or spatial index");
-      return;
+// TODO: this can be a utility and be more generic with a few more options, we could make the prepared
+//  statements on the fly and retrievable by the caller, then anything in the code base that wants to
+//  use sqlite can make use of this utility class. for now its ok to be specific to landmarks though
+struct LandmarkDatabase::db_pimpl {
```
kevinkreiser, line 11: i think this class could get more generic and let users of it register prepared statements, retrieve them and run them. then we could use this class for any sqlite needs. for now i didnt go that far but if we do then this can go to a private header

src/mjolnir/landmark_database_builder.cc
```diff
@@ -1,103 +1,114 @@
...
+  sqlite3* db;
+  sqlite3_stmt* insert_stmt;
+  sqlite3_stmt* bounding_box_stmt;
+  std::shared_ptr<void> spatial_lite;
+  bool vacuum_analyze = false;
+
+  db_pimpl(const std::string& db_name, bool read_only) : insert_stmt(nullptr) {
```
kevinkreiser, line 18: the constructor does all the initialization stuff directly rather than having myriad functions. if anything goes wrong it throws. if something upstream wants to catch what is thrown and log it fine. i would expect the graph enhancer for example to catch, log that i couldnt open the landmarks database and that its skipping adding landmarks to the tiles. for the writer part i would just let the throw abort the process (ie dont catch)

src/mjolnir/landmark_database_builder.cc
```diff
@@ -164,7 +148,7 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
   sqlite3_bind_double(bounding_box_stmt, 3, maxLong);
   sqlite3_bind_double(bounding_box_stmt, 4, maxLat);
 
-  LOG_INFO(sqlite3_expanded_sql(bounding_box_stmt));
+  LOG_TRACE(sqlite3_expanded_sql(bounding_box_stmt));
```
kevinkreiser, line 151: i pretty much removed all logs except:

1. the destructor logs when vacuum analyze fails
2. insert/select log trace now because anything else is too verbose

CHANGELOG.md
vesperlou: ok, will take a look

kevinkreiser:
@vesperlou @nilsnolde do either of you have time to look this pr over?

vesperlou:
> @vesperlou @nilsnolde do either of you have time to look this pr over?

@kevinkreiser I reviewed yesterday but didn't know I should submit it ooops ;) left 2 question marks above.

kevinkreiser:
> but didn't know I should submit it ooops

yeah that happens all the time, its a stupid UI design on githubs part

src/mjolnir/landmark_database_builder.cc
```diff
@@ -173,53 +157,16 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
...
-
     ret = sqlite3_step(bounding_box_stmt);
   }
 
   if (ret != SQLITE_DONE && ret != SQLITE_OK) {
-    LOG_ERROR("Error: " + std::string(sqlite3_errmsg(db)));
-    sqlite3_close(db);
-    throw std::runtime_error("Failed to retrieve landmarks in bounding box");
+    throw std::runtime_error("Sqlite could not query landmarks in bounding box: " +
+                             pimpl->last_error());
```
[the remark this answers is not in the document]
kevinkreiser: nope, the connection will be closed when the object is destructed no need to close it ahead of time.

valhalla/mjolnir/landmark_database_builder.h
```diff
@@ -1,7 +1,8 @@
 #pragma once
 
-#include "filesystem.h"
-#include "mjolnir/util.h"
+#include <memory>
+#include <string>
+#include <vector>
```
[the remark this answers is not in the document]
kevinkreiser: if you look below these are the only things the interface makes use of, IWYU

kevinkreiser:
@vesperlou to finalize the review you can either select "approve" or "request" changes on the diff tab in the upper right hand side.

valhalla/mjolnir/landmark_database_builder.h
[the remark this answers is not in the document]
vesperlou: OK! got it

src/mjolnir/landmark_database_builder.cc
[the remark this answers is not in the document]
vesperlou: but we need to finalize statement before destroy the object?

nilsnolde (approved):
LGTM thanks

vesperlou (approved):

vesperlou:
> @vesperlou to finalize the review you can either select "approve" or "request changes" on the diff tab in the upper right hand side.

thanks! approved :)

src/mjolnir/landmark_database_builder.cc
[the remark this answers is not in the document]
kevinkreiser: yep the finalize statement is in the destructor above before the call to close the connection

src/mjolnir/landmark_database_builder.cc
```diff
@@ -1,103 +1,114 @@
...
 
-bool LandmarkDatabase::create_spatial_index() {
-  sql = "SELECT CreateSpatialIndex('landmarks', 'geom')";
-  ret = sqlite3_exec(db, sql.c_str(), NULL, NULL, &err_msg);
-  if (ret != SQLITE_OK) {
-    LOG_ERROR("Error: " + std::string(err_msg));
-    sqlite3_free(err_msg);
-    sqlite3_close(db);
-    return false;
+    sqlite3_finalize(insert_stmt);
```
kevinkreiser, line 94: here you can see we finalize the statements and then close the connection

vesperlou: sure. I am actually wondering why we should explicitly finalize statements in destructor, because I thought they are going to be cleaned automatically when the object is destroyed, like before throwing we dont have to close the connection.

kevinkreiser: like the db connection the prepared statements are pointing to some memory that was dynamically allocated. in order for us to not "leak" memory we do need to manually deallocate that memory. when we `throw` the call immediately bubbles that exception upwards however the rules about what happens to objects when their life time ends still apply. that means if the outside scope that caused the exception was the scope that owned the `LandmarkDatabase` object lets that object go out of scope, its destructor (here) will be called and make sure to clean up the connection and the prepared statements.

if you mean closing the connection should finalize any statements, i dont think so. prepared statemnts are not tied to any connection they are a wholly separate construct. from the manual:

> The application must finalize every prepared statement in order to avoid resource leaks.

vesperlou: oh I see. I mixed up some concepts earlier but now I understand everything.
thanks for the detailed explanation!

PR #4204, opened 2023-07-15
let `:forward` and `:backward` tags override access for supported modes

### LINKED ISSUE #4203: Add support for motor_vehicle:forward = no
Valhalla will happily route through motor_vehicle:forward and motor_vehicle:backward restrictions

See this routing segment for an example, and compare to OSRM and GraphHopper that does it right

https://www.openstreetmap.org/directions?engine=fossgis_valhalla_car&route=59.91829%2C10.76063%3B59.91870%2C10.75938#map=18/59.91854/10.76084

https://www.openstreetmap.org/way/38850700

kevinkreiser:
fixes #4203 

let `:forward` and `:backward` tags override access for supported modes


lua/graph.lua
```diff
@@ -873,6 +873,11 @@ function filter_tags_generic(kv)
     return 1
   end
 
+  --toss actual areas
```
kevinkreiser, line 876: this was way down at the bottom, if we are going to bail, best to do it early and save everyone some time :smile:

lua/graph.lua
```diff
@@ -991,7 +996,7 @@ function filter_tags_generic(kv)
     if kv["motorroad"] == "yes" then
       kv["motorroad_tag"] = "true"
     end
-
+  -- its not a highway type that we know of
```
kevinkreiser, line 999: just adding comments to be able to make sense of this monstrosity

lua/graph.lua
```diff
@@ -1256,6 +1262,10 @@ function filter_tags_generic(kv)
     else
       kv["pedestrian_backward"] = kv["pedestrian_forward"]
     end
+
+  -- what on earth is going on here? if there is no oneway tagging we say the way is bidirectional despite all the
+  -- backward/forward crap we did above? this has to be the most common case (no oneway tag) so why on earth does it
+  -- reset any of the other directional tagging that is parsed above?
   elseif oneway_norm == nil or oneway_norm == "false" then
```
kevinkreiser, line 1269: i had tried putting the `:forward` overrides just above this with some of the bike and pedestrian stuff but this bit of code immediately neuters that. i cant really make sense of it, seems like we are assumign we only processed forward so far and no oneway means we can copy it over to backward. that would make sense but just above it there is a bunch of crap that sets `_backward` so.. not sure..

lua/graph.lua
```diff
@@ -1305,6 +1315,40 @@ function filter_tags_generic(kv)
     kv["bus_backward"] = "true"
   end
 
+  --let all the :forward overrides through
```
kevinkreiser, line 1318: this is the new code for `:forward` and `:backward` it is certainly redundant for somethings but whatever, i liked have it all in a single spot. i know it will be a terrible chore but the sooner we port this thing to c++ and just drop lua completely, the better

lua/graph.lua
```diff
@@ -2167,6 +2205,20 @@ function nodes_proc (kv, nokeys)
   return 0, kv
 end
 
+-- useful for dumping the main kv table to see before and after tag filtering
```
kevinkreiser, line 2208: i used this a lot to see what the heck the filtering logic was actually doing

test/gurka/test_access.cc
```diff
@@ -0,0 +1,621 @@
...
+    auto result = gurka::do_action(valhalla::Options::route, map, {"A", "E"}, c);
+    gurka::assert::raw::expect_path(result, {"AB", "BE"});
+
+    result = gurka::do_action(valhalla::Options::route, map, {"A", "F"}, c);
+    gurka::assert::raw::expect_path(result, {"AB", "BC", "CF"});
+
+    result = gurka::do_action(valhalla::Options::route, map, {"A", "G"}, c);
+    gurka::assert::raw::expect_path(result, {"AB", "BC", "CD", "DG"});
+  }
+}
```
kevinkreiser, line 573: these were all existing tests. i hate that we have so many little one off tests. it makes the CI build slow and it also makes stuff hard to find. all of these tests are verifying stuff with access, which is a big subject area in the code base of course. to me, its better to just have all the access crap in one test. i know that now its a big test but i think its better than 7 different executables that all need to be run (when we already have hundreds)

test/gurka/test_access.cc
```diff
@@ -0,0 +1,621 @@
...
+                |   |
+                H---G
+    )";
+
+  const gurka::ways ways = {
+      {"ABCDE", {{"highway", "primary"}}},
+      {"CFH", {{"highway", "primary"}, {"motor_vehicle:forward", "no"}, {"motor_vehicle:backward", "yes"}, {"foot:forward", "no"}, {"foot:backward", "yes"}, {"bicycle:forward", "no"}, {"bicycle:backward", "yes"}}},
+      {"HG", {{"highway", "primary"}}},
+      {"EG", {{"highway", "primary"}, {"motor_vehicle:forward", "yes"}, {"motor_vehicle:backward", "no"}, {"foot:forward", "yes"}, {"foot:backward", "no"}, {"bicycle:forward", "yes"}, {"bicycle:backward", "no"}}},
+  };
```
kevinkreiser, line 592: so EG can only be traversed from E to G for all modes and CFH can only be traversed from H to C by all modes

test/gurka/test_access.cc
```diff
@@ -0,0 +1,621 @@
...
+      gurka::detail::map_to_coordinates(ascii_map, gridsize_metres, {5.1079374, 52.0887174});
+  auto map =
+      gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_access_motor_vehicle", build_config);
+
+  auto cfg = test::make_config("test/data/gurka_access_motor_vehicle", build_config);
+  baldr::GraphReader r(cfg.get_child("mjolnir"));
+  auto first = gurka::findEdgeByNodes(r, layout, "C", "E");
+  auto second = gurka::findEdgeByNodes(r, layout, "E", "G");
+
+  for (auto& c : costing) {
```
kevinkreiser, line 604: for all costings we check all permutations of backward/forward + no/yes

src/mjolnir/luatagtransform.cc
```diff
@@ -126,7 +126,6 @@ Tags LuaTagTransform::Transform(OSMType type, uint64_t osmid, const Tags& maptag
       result.clear();
     }
   } catch (std::exception& e) {
-    // ..gets sent back to the main thread
```
kevinkreiser, line 129: this was just an errant comment that i noticed while debugging. sometimes its quite annoying that we silence logs in gurka tests, especially because its the only way to see what is wrong with your lua code

lua/graph.lua
```diff
@@ -1305,6 +1314,40 @@ function filter_tags_generic(kv)
     kv["bus_backward"] = "true"
   end
 
+  --let all the :forward overrides through
+  local mv_forward = kv["motor_vehicle:forward"] or kv["vehicle:forward"]
```
nilsnolde, line 1318: hm, being strict `motor_vehicle` does not equal `vehicle`, I think the most important distinction for us is that `vehicle` includes bicycle (but don't ask about e-bikes haha). OSM definition somewhat disagrees with Wikipedia definition (if rail-based is vehicle or not), but I think we should include bike for `vehicle` at least.

nilsnolde (changes requested):
I just have one reservation towards `vehicle` also including bike access. Not going through the entire ludicrous lua code, it seems to agree that `vehicle` would only not affect ped for our costing models.

Other than that, thanks for the cleanup. Maybe we find someone for next year's GSoC to turn lua to C++ :wink:

kevinkreiser:
@nilsnolde yeah this thought crossed my mind but i knew i wasn't going to like the insane definition that the internet was going to have so i covered my eyes :smile: (more like held my nose)

to me you would never call a bike a vehicle that is absolute insanity. in modern parlance vehicle is simply shorthand for motor vehicle. vehicle in the generic sense could be literally anything that transports so the rail distinction also makes zero sense.

anyway i can include bikes no worries. i'll get it tonight.

nilsnolde:
> to me you would never call a bike a vehicle that is absolute insanity.

Haha to me who never personally owned a car and likely never will (:crossed_fingers:), that makes perfect sense. Rail-based definitely falls into that category for me too, but not sure if e.g. skis should though (but is apparently)

nilsnolde (approved):

Issue #4173, opened 2023-06-19
Link error macOS M1 because of abseil/protobuf


rochadt:
Following the installation instructions for mac (https://valhalla.github.io/valhalla/building/#building-from-source-macos), I'm able to build the `Release` version and everything seems to work well. However I'm running into the following error when building the `Debug` version:

```
[ 83%] Linking CXX executable valhalla_build_statistics
/usr/local/Cellar/cmake/3.26.4/bin/cmake -E cmake_link_script CMakeFiles/valhalla_build_statistics.dir/link.txt --verbose=1
ld: warning: Linking with PIE, -image_base will be ignored
ld: warning: non-standard -pagezero_size is deprecated when targeting macOS 13.0 or later
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++  -fcolor-diagnostics  -g -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk -mmacosx-version-min=13.1 -Wl,-search_paths_first -Wl,-headerpad_max_install_names CMakeFiles/valhalla_service.dir/src/valhalla_service.cc.o -o valhalla_service  -Wl,-rpath,/usr/local/lib src/libvalhalla.a /usr/local/lib/libprotobuf-lite.dylib /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk/usr/lib/libcurl.tbd /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk/usr/lib/libz.tbd /usr/local/lib/libprime_server.dylib /usr/local/lib/libspatialite.dylib /usr/local/opt/sqlite3/lib/libsqlite3.dylib /usr/local/lib/libluajit-5.1.dylib
[...]

PR #4207, opened 2023-07-17
fix up Cmake/protobuf disaster

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4173

Let's see what CI says.. 

The problem is:
- protobuf requires abseil since 3.22.x
- CMake's own `FindProtobuf` is not looking for abseil and hence not linking properly, see https://gitlab.kitware.com/cmake/cmake/-/issues/24321

Instead we use protobuf's own CMake config files with `find_package(Protobuf REQUIRED CONFIG)`. However, we also need to enable `protobuf_MODULE_COMPATIBLE=ON`, which protects seeing `protobuf_generate_cpp` since a few protobuf versions. I made no attempt to research what the new way of generating the src/headers is.

Also I should add, I have no clue if all package managers actually install the CMake config when installing protobuf.. I'd guess that depends if a package manager uses CMake to build it? We'll see what CI says..


nilsnolde:
To add to this (but didn't try): for anyone running into this mess, you could try to determine your local protobuf version, clone https://github.com/protocolbuffers/protobuf, check out the right commit, install it somewhere in your home and copy the `install_dir/lib/cmake/protobuf` files to the location CMake would look for config files on your system (on linux `/usr/lib/cmake/protobuf/`).

It's a bother, but sorry, can't offer any better solution at this point..

CMakeLists.txt
```diff
@@ -181,6 +181,33 @@ endif()
...
+  if(Protobuf_VERSION VERSION_GREATER_EQUAL 4.22.0)
+    add_definitions(-DPROTOBUF_USE_DLLS)
+    if(APPLE OR WIN32)
+      # CONFIG mode is ideal for new abseil protobuf builds, which uses the protobuf-provided
+      # CMake configuration instead of CMake internal one, as that's not looking for abseil:
+      # https://github.com/valhalla/valhalla/issues/4173#issuecomment-1638067950
+      # Newer protobuf versions require a protobuf_MODULE_COMPATIBLE to be able 
+      # to see the protobuf_generate_cpp() CMake function
+      set(protobuf_MODULE_COMPATIBLE ON CACHE BOOL "")
+      find_package(Protobuf REQUIRED CONFIG)
```
nilsnolde, line 198: we can shortcut it when we're not on linux

CMakeLists.txt
```diff
@@ -181,6 +181,33 @@ endif()
...
+    else()
+      # linux might not have CMake config installed (e.g. Debian derivatives)
+      # need to link protobuf to abseil manually
+      find_package(Protobuf REQUIRED)
+      find_package(absl REQUIRED)
+      if(TARGET protobuf::libprotobuf-lite)
+        target_link_libraries(protobuf::libprotobuf-lite INTERFACE "absl::absl_check;absl::absl_log;absl::algorithm;absl::base;absl::bind_front;absl::bits;absl::btree;absl::cleanup;absl::cord;absl::core_headers;absl::debugging;absl::die_if_null;absl::dynamic_annotations;absl::flags;absl::flat_hash_map;absl::flat_hash_set;absl::function_ref;absl::hash;absl::layout;absl::log_initialize;absl::log_severity;absl::memory;absl::node_hash_map;absl::node_hash_set;absl::optional;absl::span;absl::status;absl::statusor;absl::strings;absl::synchronization;absl::time;absl::type_traits;absl::utility;absl::variant")
```
nilsnolde, line 205: I took this list from my local protobuf config cmake file, it's how abseil is linked there. I'm sure we hardly need all of them, but I felt no desire singling them out as it pretty is irrelevant

CMakeLists.txt
```diff
@@ -181,6 +181,33 @@ endif()
 
 if(NOT Protobuf_FOUND)
   find_package(Protobuf REQUIRED)
+
+  # orchestrated effort of both protobuf & cmake are responsible for the next lines..
+
+  # TODO: monitor how protobuf Linux packages evolve. Eventually everyone should use
+  # the Apple configuration, but e.g. apt does not install CMake config with protobuf
+  if(Protobuf_VERSION VERSION_GREATER_EQUAL 4.22.0)
+    add_definitions(-DPROTOBUF_USE_DLLS)
```
kevinkreiser, line 190: what does this do? the way it reads is that it would try to link dynamically rather than statically. if that is the case, i'd like to say that we shouldnt be making this choice for the person building. on platforms such as android and ios it makes way more sense to statically link so we dont have to worry about what versions of libraries are lying around. is this needed?

nilsnolde: apparently it's needed for (some) version on apple: https://github.com/apache/arrow/issues/35987. not entirely sure it's really needed, no. IIRC I built without it just fine on M1. I'll comment it out for now with that arrow issue

CMakeLists.txt
nilsnolde: I should add that this is not necessarily super safe to use, that's the configuration for the exact version of protobuf that I have installed. I can't really tell if this also works for smth between 3.22.0 and my 3.23.4, and of course this might break for future protobuf versions. It might result for a longer while in a series of version ifs. It'd be nice to make valhalla compile again without local patches (which tend to get committed and pushed). I'm not sure what else to try.

@rochad can you try this branch if it gets rid of your build problems on OSX?

kevinkreiser: we can make it link just the ones that protobuf actually needs. like delete all of them and add them back one at a time until it doesnt have linker errors anymore. seems like to start with you might only need: `absl::absl_log`

CMakeLists.txt
```diff
@@ -181,6 +181,34 @@ endif()
...
+      # to see the protobuf_generate_cpp() CMake function
+      set(protobuf_MODULE_COMPATIBLE ON CACHE BOOL "")
+      find_package(Protobuf REQUIRED CONFIG)
+    else()
+      # linux might not have CMake config installed (e.g. Debian derivatives)
+      # need to link protobuf to abseil manually
+      find_package(Protobuf REQUIRED)
+      find_package(absl REQUIRED)
+      if(TARGET protobuf::libprotobuf-lite)
+        target_link_libraries(protobuf::libprotobuf-lite INTERFACE "absl::absl_log")
```
nilsnolde, line 206: @kevinkreiser jep, that was easier than expected.. this links fine for me

PR #4209, opened 2023-07-19
sudo install python dependencies on OSX

nilsnolde:
OSX fails because conan 1.x depends on `pyyaml<6.0` and apparently there's already pyyaml 6.x installed, which it can't uninstall bcs of lacking permissions, so use `sudo` to install python packages.


nilsnolde:
now it's working finally..

kevinkreiser (approved):

PR #4213, opened 2023-07-21
playing around with arm builds

kevinkreiser:
circle ci has support for arm. the docs say its 20.04 which is not what we target on x86 but at the very least i plan to get the code actually building on arm. locally ive built it on arm just fine but it segfaults during `valhalla_build_tiles`. we'll see what the matter is and see if we can patch it up


Asher-JH:
How's it going any luck so far [🤔](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwingIWg8M6AAxWJSWwGHfAECRgQFnoECBYQAQ&url=https%3A%2F%2Femojipedia.org%2Fthinking-face%2F&usg=AOvVaw3Wu41wpPz_mCuyzPUV60n4&opi=89978449)

nilsnolde:
This will not solve your problem, it’s „just“ about testing on arm64. Once GHA adds a M1/2 runner for public use, we can think about distributing Docker images for M1. The docker build already takes 30 mins which a bit annoying on master merges. Cross compiling would easily double that. 

GH is still has it on their roadmap for end of 2023.

kevinkreiser:
this does seem to reproduce many of the issues we have on m1s though. i got it building on arm linux but a bunch of tests fail. most of them are just that the expected shape of a route doesnt quite match (small differences in floating point numbers) but there are a few that are more substantial. isochrones test fails in a way that makes the outputs seem quite different and the gtfs tests do some boost polygon crap that doesnt seem to work at all (why do we even use that, we know already its not aware of spherical geom imho we should remove it).

kevinkreiser:
so i decided to focus most of my time on the isochrones issue since it seems to be the one with the most egregious output. what i did was build the `run-isochrone` test target on both aarch64 and amd64 platforms. i then took the data created on the aarach64 platform, copied it over to the amd64 platform and ran `test/isochrone` manually. here's the output:

```
kkreiser@L50X1P73:~/sandbox/valhalla/build$ test/isochrone 
[==========] Running 5 tests from 1 test suite.
[----------] 5 tests from Isochrones
[ RUN      ] Isochrones.Basic
/home/kkreiser/sandbox/valhalla/test/isochrone.cc:96: Failure
  actual_geom.Size()
    Which is: 271
  expected_geom.Size()
    Which is: 272
/home/kkreiser/sandbox/valhalla/test/isochrone.cc:96: Failure
  actual_geom.Size()
    Which is: 271
  expected_geom.Size()
    Which is: 272
[  FAILED  ] Isochrones.Basic (89 ms)
[ RUN      ] Isochrones.OriginEdge
[          ] generating map PBF at test/data/isochrones/origin_edge/map.pbf
[          ] building tiles in test/data/isochrones/origin_edge
[          ] isochrone with mjolnir.tile_dir = test/data/isochrones/origin_edge with locations b with costing pedestrian
[          ] Valhalla request is: {"locations":[{"lon":0.03593248178978409,"lat":0.0,"type":"break"}],"costing":"pedestrian","costing_options":{"pedestrian":{"speed_types":["freeflow","constrained","predicted"]}},"verbose":true,"contours":[{"time":"10"}],"shape_match":"map_snap"}
[       OK ] Isochrones.OriginEdge (394 ms)
[ RUN      ] Isochrones.LongEdge
[          ] generating map PBF at test/data/isochrones/long_edge/map.pbf
[          ] building tiles in test/data/isochrones/long_edge
[          ] isochrone with mjolnir.tile_dir = test/data/isochrones/long_edge with locations a with costing pedestrian
[          ] Valhalla request is: {"locations":[{"lon":0.0,"lat":-0.0017966240891925996,"type":"break"}],"costing":"pedestrian","costing_options":{"pedestrian":{"speed_types":["freeflow","constrained","predicted"]}},"verbose":true,"contours":[{"time":"15"}],"shape_match":"map_snap"}
[       OK ] Isochrones.LongEdge (393 ms)
[ RUN      ] Isochrones.test_clear_reserved_memory
[       OK ] Isochrones.test_clear_reserved_memory (0 ms)
[ RUN      ] Isochrones.test_max_reserved_labels_count
[       OK ] Isochrones.test_max_reserved_labels_count (0 ms)
[----------] 5 tests from Isochrones (877 ms total)

[==========] 5 tests from 1 test suite ran. (877 ms total)
[  PASSED  ] 4 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] Isochrones.Basic
```

I can check the actual output just to see what it looks like but based on the above log, it looks like the data that the arm machine created, when run through the algorithm on the amd machine works as we expect it to. This means that we can trust the whole data creation pipeline and can focus on where the algorithm might be screwing up on the arm platform. my bet is all the iterator unordered map shenanigans i did to turn the output of the conrec algorithm into actual polygons/lines.

in fact, i recently refactored the whole admins stuff to use geos' c api and in doing so i rewrote the code that pieces segments together into rings. i was very happy at the simplification of this algorithm (though geos has its own version of linemerging) and i thought why didnt it do it this way in the isochrone generation code. i might take a swag at that first as a solution.

kevinkreiser:
alright after reviewing the isochrones geometry generation code i added some informational printing to it. the good news about the algorithm is that its deterministic in its iteration and doesnt rely on unordered datastructures for that. its scans by contour line value and by grid cell so the output of the algorithms can be directly compared.

so what i first did was made it output whenever it generated a line segment from a particular gird cell and diffed the line segments generated on both platforms. as we've seen in other tests, they are only off by tiny amounts of floating point wiggle (0.000001):

image

yet the output of the above arm run suggests that the algorithm is unable to connect the segments to each other. let me see if i can prove that... what i will do is annotate each segment with what happens to it. there are a number of options for that:

- the segment completes a ring
- the segment bridges two other segments
- the segment is appended to another segment
- the segment is prepended to another segment
- the segment is an orphan (at least temporarily)

my hypothesis is that on arm many more segments stay orphans because somehow looking up connectable segments in the map doesnt work (it uses direct pointll equality). looking at the logs we get pretty convincing numbers regarding teh hypothesis:

image

we can look at the logs in detail and reason about why a segment becomes an orphan when it shouldnt... and there you see it right off the bat:

image

the second segment on the arm platform should be prepended to the first one as it is on the amd platform but it gets orphaned instead. i can only imagine this is because of floating point noise on the end of point which is very unfortunate...

kevinkreiser:
ok yep if i round the points on the segments to the nearest 7 decimal places the test still fails but its only off by one coordinate meaning its acceptable.

so a quick run down of whats left to do:

1. unbork CI so that all the builds run again including the arm one
2. arm builds need to disable warnings because the compiler does a lot more warning
3. add testing work arounds for all the tests where shape is ever so slightly different (or if its too annoying disable those tests for arm)

kevinkreiser:
well i thought i would for sure fix the isochrones part of the build with a very tight tolerance of 1m max distance between the two polygons in the failing tests but nope it says its more than that:

image

whats that you cant see it? look in the upper left:

image

looks like they are off by 20m at some spots so i'll need to add tolerance to the test for it to pass.

kevinkreiser:
ok landmarks and searchfilter are the only 2 tests still failing..

kevinkreiser:
ok i think i just got the last one... i'll make some comments to the pr and get it ready for review

.circleci/config.yml
```diff
@@ -9,14 +9,35 @@ executors:
...
+    steps:
+      - run: |
+            source venv/bin/activate
+            ./scripts/install-linux-deps.sh
+  linux_preflight:
+    steps:
+      - run: |
+            sudo apt-get update --assume-yes
+            env DEBIAN_FRONTEND=noninteractive sudo apt-get install --yes --quiet python3-pip python3-venv git curl
+            python3 -m venv venv
```
kevinkreiser, line 30: both mac and linux platforms now use venv to install python stuff. newer versions of python insist on this so rather than work around it (which you can) we just comply in this case

.circleci/config.yml
```diff
@@ -9,14 +9,35 @@ executors:
...
+  check_ci_lint:
+    steps:
+      - run: |
+            source venv/bin/activate
```
kevinkreiser, line 34: annoyingly we always have to get into our venv before we do any new step in CI now but whatever... also notice i broke up the stuff into more steps

.circleci/config.yml
```diff
@@ -25,28 +46,24 @@ jobs:
...
-            pip3 install requests
-            if [[ $(./scripts/needs_ci_run) == "skip CI" ]]; then
-                echo "Changes in last commit do not need CI. Skipping step"
-                circleci-agent step halt
-            fi
-      - run: sudo python3 -m pip install pre-commit && pre-commit run --all-files && ./scripts/error_on_dirty.sh
-      - install_linux_dependencies
+      - linux_preflight
+      - check_ci_lint
+      - linux_deps
```
kevinkreiser, line 51: these are steps now

.circleci/config.yml
```diff
@@ -25,28 +46,24 @@ jobs:
...
-      - run: mkdir build
       - run: |
+          mkdir build
           # NOTE: -Werror disabled in CI, as we currently have >4k warnings.
-          cd build \
-            && cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=On -DCPACK_GENERATOR=DEB \
-                -DENABLE_COMPILER_WARNINGS=On -DENABLE_WERROR=Off -DCMAKE_EXPORT_COMPILE_COMMANDS=On \
-                -DCMAKE_CXX_FLAGS="-fuse-ld=lld" \
-                -DLOGGING_LEVEL=INFO \
-                -DENABLE_PYTHON_BINDINGS=On
```
kevinkreiser, line 49: there was no real benefit to `&&`ing these together so i just put them on separate lines

.circleci/config.yml
```diff
@@ -107,26 +120,60 @@ jobs:
       - run: sudo make -C build install
       - run: make -C build package
 
+  build-arm-release:
```
kevinkreiser, line 123: and this is the new CI task for building on arm. notice it uses even newer ubuntu (which is what forced the issues with venv). we need newer ubuntu because it has better support for arm packages (and we dont want to have to build our deps from source because that is a huge pain)

.circleci/config.yml
```diff
@@ -107,26 +120,60 @@ jobs:
...
+    docker:
+      - image: arm64v8/ubuntu:23.04
+    resource_class: arm.xlarge
+    steps:
+      - checkout
+      - run: apt update && apt install sudo
```
kevinkreiser, line 129: this is one main difference, the basic ubuntu image doesnt come with sudo preinstalled

.circleci/config.yml
```diff
@@ -107,26 +120,60 @@ jobs:
...
+      - run: git submodule sync && git submodule update --init
+      - restore_cache:
+          keys:
+            - ccache-release-linux-arm_64-v3-{{ .Branch }}-{{ checksum "conanfile.txt" }}
+            - ccache-release-linux-arm_64-v3-{{ checksum "conanfile.txt" }}
+      - run: |
+          mkdir build
+          source venv/bin/activate
+          cd build
+          cmake .. -DENABLE_SINGLE_FILES_WERROR=Off -DBENCHMARK_ENABLE_WERROR=Off -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=On -DENABLE_PYTHON_BINDINGS=On \
```
kevinkreiser, line 142: two main differences for arm here: `-DENABLE_SINGLE_FILES_WERROR=Off -DBENCHMARK_ENABLE_WERROR=Off`

we cannot enable warnings as errors. for whatever reason the compiler is extra pedantic on arm and so we cant get through a build with werror turned on for pretty much any parts of the project

scripts/format.sh
```diff
@@ -10,6 +10,10 @@ set -o errexit -o pipefail -o nounset
 
 readonly CLANG_FORMAT_VERSION=11.0.0
 
+if [[ $(uname -i) == 'aarch64' ]]; then
+  echo 'Formatting is disabled on arm for the time being'
+  exit
+fi
```
kevinkreiser, line 16: we cannot get a version of clang format for arm that is locked to the same version as the linux/mac binaries we get from mason. we should stop using mason but because we are in this retarded mode of having to have all formatting look identical, and because the linter is software and goes through revisions, if we dont have matchign revisions across platforms we cant get the same output. this situation kind of makes you lose hope for humanity

scripts/install-linux-deps.sh
```diff
@@ -48,13 +48,13 @@ env DEBIAN_FRONTEND=noninteractive apt-get install --yes --quiet \
   
 # build prime_server from source
 # readonly primeserver_version=0.7.0
-readonly primeserver_dir=/usr/local/src/prime_server
+readonly primeserver_dir=/tmp/prime_server
```
kevinkreiser, line 51: on the regular linux builds we no longer run this script as sudo, instead using sudo inside of the script only where its needed. because of that `/tmp` is a better place for temporary stuff

src/midgard/polyline2.cc
```diff
@@ -314,6 +314,31 @@ void Polyline2<coord_t>::Generalize(container_t& polyline,
     DouglasPeucker<coord_t>(polyline, epsilon_m, exclusions);
 }
 
+template <typename coord_t>
+template <typename container_t>
+typename container_t::value_type::first_type
+Polyline2<coord_t>::HausdorffDistance(const container_t& l1, const container_t& l2) {
```
kevinkreiser, line 320: on very very large component of the failing tests on arm were floating point noise ever so slightly changing the results of tests from the expected result. this showed itself mostly in shape comparisions. to combat this ive implemented the hausdorff distance metric which lets us tell the maximum (of the minimum) distance (in meters) between two shapes

src/mjolnir/graphtilebuilder.cc
```diff
@@ -1295,10 +1295,12 @@ void GraphTileBuilder::AddLandmark(const GraphId& edge_id, const Landmark& landm
 
   if (eib == edgeinfo_offset_map_.end()) {
     throw std::runtime_error("Couldn't find edge info for the given edge: " +
-                             std::to_string(static_cast<int>(edge_id.id())));
+                             std::to_string(edge_id));
```
kevinkreiser, line 1298: more useful error when debugging :D

src/mjolnir/graphtilebuilder.cc
```diff
@@ -1295,10 +1295,12 @@ void GraphTileBuilder::AddLandmark(const GraphId& edge_id, const Landmark& landm
...
   }
 
+  // get the value and prepend the tag to it
   std::string tagged_value = landmark.to_str();
+  tagged_value.insert(tagged_value.begin(), static_cast<char>(baldr::TaggedValue::kLandmark));
```
kevinkreiser, line 1303: both the encoding of a landmark to binary and decoding of the landmark to binary do not include the preceeding "tag" character now. it was quite confusing when testing that one had it and the other didn't

scripts/format.sh
nilsnolde: Is there an alternative to mason? Would love to get rid of that too, being unmaintained etc. Conan & vcpkg are only doing libs, no exes (as projects). 

FWIW, I’m very much pro formatter on a project with uhh say heterogeneous styles😄python was an eyesore to me before I got it linted & formatted.

src/mjolnir/landmarks.cc
```diff
@@ -352,8 +354,10 @@ void FindLandmarkEdges(const boost::property_tree::ptree& pt,
       // get landmarks in the tile
       midgard::AABB2<PointLL> bbox = baldr::TileHierarchy::GetGraphIdBoundingBox(tileset[i]);
 
-      std::vector<Landmark> landmarks =
-          db.get_landmarks_by_bbox(bbox.minx(), bbox.miny(), bbox.maxx(), bbox.maxy());
+      std::vector<Landmark> landmarks = db.get_landmarks_by_bbox(bbox.minx() - kLandmarkQuerryBuffer,
+                                                                 bbox.miny() - kLandmarkQuerryBuffer,
+                                                                 bbox.maxx() + kLandmarkQuerryBuffer,
+                                                                 bbox.maxy() + kLandmarkQuerryBuffer);
```
kevinkreiser, line 360: on arm, apparently when locations lie directly on the bounding box boundary (in this case the equator and the prime meridian) `st_covers` was not returning the landmarks. you might think that its just floating point precision in the actual landmarks but if that were the case then we'd iterate over more 1x1 degree tiles and get the landmarks, however we dont, our test still ends up doing only 1 1x1 query. so i copped out and added a 1 meter buffer

nilsnolde: I wonder, where would one report spatialite bugs😅it’s so awkward to not have it on some git platform. Every time I think, that’s it, now it’s over, a new version comes out from somewhere..

test/astar.cc
```diff
@@ -956,7 +956,7 @@ void test_backtrack_complex_restriction(int date_time_type) {
     default:
       throw std::runtime_error("unhandled case");
   }
-  EXPECT_EQ(leg.shape(), correct_shape)
+  EXPECT_TRUE(test::encoded_shape_equality(leg.shape(), correct_shape))
```
kevinkreiser, line 959: all the shape related failures look like th is now calling this helper function which does hausdorff distance to compare how close in distance the shapes are to each other

test/gurka/test_landmarks.cc
```diff
@@ -288,6 +293,13 @@ TEST(LandmarkTest, TestParseLandmarks) {
 
   EXPECT_NO_THROW({ landmarks = db.get_landmarks_by_ids({3, 2, 1}); });
   EXPECT_EQ(landmarks.size(), 3);
+
+  // test round trip encoding
+  for (auto& expected : landmarks) {
+    expected.id = 0;
+    Landmark actual(expected.to_str());
+    EXPECT_TRUE(expected == actual);
+  }
```
kevinkreiser, line 302: the landmark binary encoding had a bug in the lat lon transforms. to prove that i added this test which failed on amd64 as well. then i completely rewrote the encoding/decoding steps. the equality operator i made above uses approximate equality because there is a rounding that takes place on the 7th decimal place of the lat lons

test/gurka/test_landmarks.cc
```diff
@@ -302,23 +314,18 @@ TEST(LandmarkTest, TestTileStoreLandmarks) {
...
-
-    Landmark landmark(edge_index, std::to_string(edge_index), ltype, point.first, point.second);
-
-    auto edge_id = tile_id;
-    edge_id.set_id(edge_index++);
-
+    auto ltype = static_cast<LandmarkType>((edge_id.id() + 1) % invalid_landmark);
+    Landmark landmark(0, std::to_string(edge_id), ltype, point.first, point.second);
     tb.AddLandmark(edge_id, landmark);
+    ++edge_id;
```
kevinkreiser, line 328: just made this more succinct

test/gurka/test_landmarks.cc
```diff
@@ -327,55 +334,61 @@ TEST(LandmarkTest, TestTileStoreLandmarks) {
   GraphReader gr(landmark_map.config.get_child("mjolnir"));
   auto tile = gr.GetGraphTile(tile_id);
 
+  auto check_landmark = [&](const auto& landmark, const auto& point) {
```
kevinkreiser, line 337: both of the checks below did the same thing so i pulled it out into a lambda and reused it

test/gurka/test_landmarks.cc
```diff
@@ -327,55 +334,61 @@ TEST(LandmarkTest, TestTileStoreLandmarks) {
...
+    // we dont store this in the tile
+    EXPECT_EQ(landmark.id, 0);
+    // opposing edges will have a copy of the edge we associated it to so we need to check it
+    bool skip = false;
+    if (landmark.name != std::to_string(edge_id)) {
+      auto opp = gr.GetOpposingEdgeId(edge_id);
+      ASSERT_EQ(landmark.name, std::to_string(opp)) << "Unexpected landmark name or edge/opp_edge";
+      skip = true;
+    }
+    // we dont do deep check on opposing edge records
```
kevinkreiser, line 350: previously this test was relying on luck to pass. in truth we will find the landmark both on its edge and its opposing edge (since they share edgeinfo), when its the opposing copy we skip the more difficult tests saving them only for the direction of the edge where we named it to match the edge id

test/gurka/test_landmarks.cc
```diff
@@ -327,55 +334,61 @@ TEST(LandmarkTest, TestTileStoreLandmarks) {
...
-
       Landmark landmark(value.second);
 
       // check data correctness
       std::vector<PointLL> shape = ei.shape();
       auto point = shape[shape.size() / 2];
-      auto ltype = static_cast<LandmarkType>((edge_index + 1) % invalid_landmark);
-
-      EXPECT_EQ(landmark.id, 0);
-      EXPECT_EQ(landmark.name, std::to_string(edge_index++));
```
kevinkreiser, line 352: there was a bug here, incrementin the edge index but only if it got past the continue above

test/gurka/test_search_filter.cc
```diff
@@ -261,7 +261,7 @@ class ExcludeClosuresOnWaypoints : public ::testing::TestWithParam<std::string>
             |
             H
 
-   L-4---5-M-6-N
+   L4----5-M-6-N
```
kevinkreiser, line 264: this was evil. we have a test that checks if trivial routes fail when the edge is closed. the trivial route is from `4` to `5`. we close the edge `LM`. on amd64 `4` snaps to `A` above and `5` snaps to `M`. since they are in disconnected components the route fails as you expect. on arm64 `4` snaps to `M` and since `5` also snaps to `M` the route passes and the test fails. this is because, on both platforms, `4` is equidistant to `M` and to `A` above. so it was just some floating point noise that made it pick one over the other. by moving `4` to the left one char, it becomes closer to `A` regardless of platform and floating point noise and so we get reliable results.

test/summary.cc
```diff
@@ -52,9 +52,11 @@ TEST(Summary, test_time_summary) {
...
+  EXPECT_TRUE(
+      test::
+          encoded_shape_equality(
```
kevinkreiser, line 57: this line is the change, using the hausdorff distance to check the shapes

test/test.h
```diff
@@ -53,6 +54,23 @@ make_config(const std::string& path_prefix,
...
+                                        const container_t& actual,
+                                        typename container_t::value_type::first_type tolerance = 1) {
+  auto hd = Polyline2<typename container_t::value_type>::HausdorffDistance(expected, actual);
+  if (hd > tolerance)
+    return testing::AssertionFailure() << "shape exceeds tolerance by " << hd - tolerance;
+  return testing::AssertionSuccess();
+}
+
+inline testing::AssertionResult
+encoded_shape_equality(const std::string& expected, const std::string& actual, double tolerance = 1) {
```
kevinkreiser, line 68: added these two helper functions for any time we need to compare shape in a robust way

valhalla/baldr/landmark.h
```diff
@@ -96,27 +97,13 @@ inline LandmarkType string_to_landmark_type(const std::string& s) {
...
-  return (((uint64_t(lng * 1e7) + uint64_t(180 * 1e7)) & ((1ull << 32) - 1)) << 31) |
-         ((uint64_t(lat * 1e7) + uint64_t(90 * 1e7)) & ((1ull << 31) - 1));
-}
-
-inline std::pair<double, double> decode_lnglat(uint64_t location) {
-  double lng = (int64_t((location >> 31) & ((1ull << 32) - 1)) - 180 * 1e7) * 1e-7;
-  double lat = (int64_t(location & ((1ull << 31) - 1)) - 90 * 1e7) * 1e-7;
-
-  return std::make_pair(lng, lat);
-}
```
kevinkreiser, line 109: didnt need to expose these in the header moved them into the body and simplified. here you can see there is clearly something wrong with the decoding step

scripts/format.sh
kevinkreiser: i know im alone in my heterodox opinion however you have to admit the irony that you want to enforce conformity but cant because the thing that does it is different on different platforms and changes randomly over time. its like saying lets adopt a standard because people are doing this way too many different ways and then when you implement the standard it allows for you to implement it in different ways. the only thing you gained was the pain of trying to adhere to the standard the core problem is unsolved.

as far as a real solution goes we could run clang_format via a docker image that locks the version. i think i proposed that before. i know its a bit clunky but it is robust to all the problems. it does mean that docker is a required dependency to lint the code which is a bit nuts but short of building clang format from source its the only real option that or some other magical formatter

gknisely (approved):

kevinkreiser:
@gknisely thanks for the review! im going to add one more commit so that when the docker build happens and publishes our docker images we'll get both platforms amd and arm

.circleci/config.yml
kevinkreiser: technically we dont have to do it this way, we can simply export some environment var updates to get the virtual env automatically.. i can update that later, dont want to rock the CI boat at the moment

peitili (approved):

.circleci/config.yml
nilsnolde: not very opinionated here, I'd have just `--break-system-packages` when installing with pip

nilsnolde: also needs to go into format.sh, I know. I'm fine with that. nothing changed in the world, pip is just getting parental for some reason. so functionally we'd still do the exact same thing as before. I think I already did that in another PR that's in the queue (doesn't mean anything, just underlines my position:))

kevinkreiser: if you like i can simplify all this by just doing the environment variables to use the virtual env. ill do that quickly :smile:

nilsnolde (commented):
Looks good, thanks! M1 users should be very grateful for this.

I'd just do the `--break-system-packages` and not comply with pip's attitude.

Dockerfile
```diff
@@ -1,12 +1,16 @@
 ####################################################################
-FROM ubuntu:22.04 as builder 
+FROM ubuntu:23.04 as builder
```
nilsnolde, line 2: is this needed?

Dockerfile
```diff
@@ -43,30 +47,35 @@ RUN tar -cvf valhalla.debug.tar valhalla_*.debug && gzip -9 valhalla.debug.tar
...
+ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
+ENV LD_LIBRARY_PATH=/usr/local/lib:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib32:/usr/lib32
+
 # github packaging niceties
 LABEL org.opencontainers.image.description = "Open Source Routing Engine for OpenStreetMap and Other Datasources"
 LABEL org.opencontainers.image.source = "https://github.com/valhalla/valhalla"
 
+# grab the builder stages artifacts
 COPY --from=builder /usr/local /usr/local
-COPY --from=builder /usr/lib/python3/dist-packages/valhalla/* /usr/lib/python3/dist-packages/valhalla/
```
nilsnolde, line 58: not entirely sure about this. is the installed package the exact same as the built package in /usr/local? maybe..

nilsnolde: Oh I see.. it's installed to /usr/local in the venv

Dockerfile
```diff
@@ -1,12 +1,16 @@
...
 ARG CONCURRENCY
 
 # set paths
-ENV PATH /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
-ENV LD_LIBRARY_PATH /usr/local/lib:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib32:/usr/lib32
+ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
+ENV LD_LIBRARY_PATH=/usr/local/lib:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib32:/usr/lib32
+# python stuff
+ENV VIRTUAL_ENV=/usr/local/venv
+RUN export DEBIAN_FRONTEND=noninteractive && apt update && apt install -y sudo python3-venv && python3 -m venv ${VIRTUAL_ENV}
```
nilsnolde, line 12: hm, a venv is quite heavy.. on my local it weighs 26 mb.

nilsnolde (commented):
sorry for being such a stickler with python stuff. I'm pretty opinionated there and IMO we shouldn't make the project heavier and so much more code (well not that much, just considering zero functional change) just bcs of some pip dev changing attitude

Dockerfile
kevinkreiser: yep, arm builds fail on older versions of ubuntu because of problems with dependencies i mentioned that is why im using this in CI as well

kevinkreiser: though now that i fixed everything, its possible i tricked myself when i was wrestling with `werror`

Dockerfile
kevinkreiser: yeah

Dockerfile
nilsnolde: Right, i was only thinking about our published image (and we can’t publish linux arm images through GitHub). It’s fine though of course, not a big deal

Dockerfile
kevinkreiser: i had the same feeling about the size of the sqlite db update in the phonetics pr

kevinkreiser:
ok, undoing venv. its only about 12 lines of change only pity is waiting on ci again

Dockerfile
kevinkreiser: why cant we? so long as the builder installs the right stuff it can definitely build arm images. i have done it myself locally on my amd64 machine. you just need to install qemu and run a few commands... do we not have permissions sufficient enough to do that on the github actions maybe?

nilsnolde: Sorry, I should take more time to answer.. disregard that, of course we can. How long did the cross compilation take?

nilsnolde: Relatively to compiling the amd image

kevinkreiser: oh it was quite a bit long on my local machine. amd was done in 10 minutes while the arm one took maybe double that? since im making the virtual env changes i'll get to test it again shortly

kevinkreiser:
so i did 99% of the stuff to undo venv and then i realized that to do that i need to add the --break-system-packages to the install depedencies script. imho that is just as egregious as requiring venv because the --break-system-packages flag errors the command if the version of pip is before the flag existed.. seems like we are damned either way!

so in the end we have two options:

1. assume who ever called the script has a pip that will let them do what they want (venv or an old version of pip)
2. write a bit of bash in the install script to detect which pip you have and do the right thing depending on which it is

in my opinion you are right about whoever maintains pip. they royally screwed this up. i think the better option above is 2. i hate that it forces us to now become nannies just like the pip maintainers though, just like sudo id rather push that burden up to the caller of the script rather than write logic around it inside the script

kevinkreiser:
ah just a bit more research and i have a workable solution, apparently you can also use an environment var to tell pip to do whatever you want to the system pip packages. i'll use that. i assume old versions of pip will ignore it. whew

nilsnolde (dismissed):

nilsnolde (approved):

kevinkreiser:
looks like the docker stuff should work its building fine in the github workflow but seems like it will be a while before it completes: https://github.com/valhalla/valhalla/actions/runs/6574607552/job/17859996388

im going to merge this in the mean time

PR #4214, opened 2023-07-21
make PointXY and GeoPoint non-virtual

dm-u:
Non-virtual classes smaller, non-virtual methods faster. There no dynamic casts and calls of virtual methods.


kevinkreiser (approved):
goodbye viable, thanks!

PR #4216, opened 2023-07-26
Remove some casts double to float and vice versa

dm-u:
Removed implicit casts to Point2 which is float when it used for consruction AABB2<PointLL> which is double.

### DIFF
--- a/src/baldr/connectivity_map.cc
+++ b/src/baldr/connectivity_map.cc
@@ -181,10 +181,9 @@ std::unordered_set<size_t> connectivity_map_t::get_colors(const baldr::TileLevel
       // Get a list of tiles required within the radius of the projected point
       const auto& ll = edge.projected;
       DistanceApproximator<PointLL> approximator(ll);
-      float latdeg = (radius / kMetersPerDegreeLat);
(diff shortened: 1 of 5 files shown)


kevinkreiser (approved):
nice clean up

kevinkreiser (changes requested):
need to fix ci by running the `format.sh` script

dm-u:
> WARNING: There was an error checking the latest version of pip.

nilsnolde:
The output is a admittedly pretty confusing, but that's not the thing it complains about, it's
```
format...................................................................Failed
- hook id: format
- files were modified by this hook
```

You'll need to run scripts/format.sh to format the c++ code.

dm-u:
> ```
> format...................................................................Failed
> - hook id: format
> - files were modified by this hook
> ```

The same code passed the checks successfully. A bit confusing.

nilsnolde:
Same code, but different _formatting_, see the first 2 file diffs removing one line each, that's likely format.sh/clang-format

dm-u:
[Got it](https://github.com/valhalla/valhalla/compare/1a08a81cd5993c7ecf01d2a36499a14fa435e207..85446e73d3a426462bf3766ac084fb9ec81eef76), thanks!

kevinkreiser (approved):

Issue #4177, opened 2023-06-23
Configurable live speed traffic fading time


janusz-anue:
It could be beneficial to have the live speed fading configurable in the valhalla.json. Currently it is hardcoded in the file `valhalla/baldr/graphtile.h` in the GetSpeed(...) function with:

```cpp
constexpr double LIVE_SPEED_FADE = 1. / 3600.;
```

I am exploring ways to read out the value from the config and pass it to the GetSpeed-Function in a backwards compatible and contributable way that doesn't uglify the codebase. But reading the config in the cost models currently is not possible as far as I see it. Are there ways to get this value in the cost model that I am missing?

Janusz Spatz on behalf of Mercedes-Benz Tech Innovation GmbH.
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

nilsnolde:
Yeah, not having access to the config in costing makes quite a few things very ugly/hard.. @kevinkreiser had the idea to make the config a singleton which gets created first thing on startup of any process and is then easily accessible to every file. Then we remove all the code passing down the config to every small function way downstream.

Is that smth you'd be interested in as well? The singleton itself isn't much work I think..

janusz-anue:
Thanks for the quick reply. The suggestion sounds useful and I could not think about any drawbacks to this singleton, it is just the bad rep that it gets and the many little rewrites of source and tests that makes me think about other ways to achieve the solution.

Maybe it would be enough if we find a way to make the config pt available in the dynamic cost class?

nilsnolde:
> if we find a way to make the config pt available in the dynamic cost class

I looked into that before, I don't think we'd like that as a project (correct me @kevinkreiser). We should really do the work for the singleton (which is totally fine IMO, if used wisely), instead of further digging that hole. 95% of the changes are mechanical I think.

kevinkreiser:
yeah passing the costing into every place we need costing is a refactor nightmare. we need to do the global config singleton idea and then remove config passing from vast swathes from the code base. i forget if i have an issue for this but its definitely the way to go

kevinkreiser:
its this one, just updated: https://github.com/valhalla/valhalla/issues/1158

janusz-anue:
Okay this will be something that we'll work on then

PR #4220, opened 2023-07-27
Config Singleton

### LINKED ISSUE #1158: Config Singleton
It would be really nice if libvalhalla had a config singleton. Something where we initialize it once and then can access it globally after that. Why would we want that...

Right now we have the annoying problem of having to pass a config to all of the top level elements. It would be nice at program start to just configure the singleton and let these objects get access to the config on the fly.

One problem this would solve directly would be that we have config values that get validated against for requests. Currently a loki worker will load these when its instantiated and then validate each request against them. This makes it so that some of the things we'd like to pull out to shared request object cannot be because the code that genrates this pbf object doesnt have access to the config only loki_worker.

Basically having a globally configured valhalla is just convenient :smile: 

more details:

1. singleton getter takes an optional config file so that on first call it gets configured (i think the python bindings used to be like this, we'd just move this idea into the library)
2. the config object knows whether or not its configured. if its not configured we throw in the getter or maybe even assert to force kill the application? this forces who ever is using it to make sure their use includes configuring the library first.
3. wrap the config boilerplate that every excecutable will have into the arg parsing boilerplate that we are adding in #4169

janusz-anue:
Fixes #1158 

Sorry for the long wait. I have finally gotten around tackling this PR. As discussed in #4177 and in #1158, this PR solves the first phase of implementing the configuration as a singleton and changes the top-level passings. This means that basically the arg_parse is changed and every class using it, as well as the valhalla_service.
Later in another phase the other passings can be removed too.

Before using config() the singleton should be configured with a config file or an inline config string.

I am open for suggestions and also nitpicks as I am looking to improve my c++ coding.

Janusz Spatz on behalf of Mercedes-Benz Tech Innovation GmbH.


nilsnolde:
Thanks a lot @janusz-anue ! We'll take a look in the next days.

FYI, ever since your last commit one more exe has been added in `mjolnir/valhalla_build_landmarks.cc` which fails the build currently.

valhalla/configuration.h
```diff
@@ -0,0 +1,57 @@
...
+  boost::property_tree::ptree config_;
+
+public:
+  Configuration(Configuration const&) = delete;
+  void operator=(const Configuration&) = delete;
+
+  static Configuration& instance(const std::string& config_file_or_inline = "") {
+    static Configuration instance(config_file_or_inline);
+    return instance;
+  }
```
nilsnolde, line 38: IMO that's a bit convoluted here. I also might not fully be aware of all constraints, but I'd prefer if this would immediately return the `config_`. I think you could remove the 2 free functions then, and the first call to this would need to specify a valid path to a valhalla.json or it'll throw in the constructor as it currently does. I think it's basically the same as your concept, just with a bit less code/fewer functions and all encapsulated in the class. Also, free functions with such generic names could be problematic at some point. You could call the function maybe rather `get_config` or so, it wouldn't be an instance anymore at that point. The important thing is that both function & `Configuration` are static. We could also return a `Configuration` instance from another function if we'd want to add more useful functions to `Configuration` class, which might be interesting.

Does that make sense?

nilsnolde (commented):
Thanks, looks good @janusz-anue. One point for discussion in the singleton design. WDYT?

nilsnolde:
You'll also need to format the code, see https://github.com/valhalla/valhalla/blob/master/CONTRIBUTING.md#code-contributions for pre-commit which makes it a bit easier IMO.

valhalla/configuration.h
nilsnolde: Another nit might be to call the class `config_singleton_t` instead of `Configuration`, following the existing pattern around singletons.

janusz-anue: I think I have an idea what you mean, I think it makes sense from an encapsulation point of view, will update.

nilsnolde: FYI, was just following up with @kevinkresier last week and he’d prefer to have a free function in the valhalla namespace, behaving similarly as I described above, so that we can just do valhalla::config(<optional_path>).

janusz-anue: I have adjusted the code a bit and taken out the configure-function, is it like you imagined?

nilsnolde: Jep, thanks a lot!

nilsnolde (dismissed):
LGTM! Only thing left is a CHANGELOG entry.

valhalla/configuration.h
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+public:
+  config_singleton_t(config_singleton_t const&) = delete;
+  void operator=(const config_singleton_t&) = delete;
+
+  static const boost::property_tree::ptree& get_config(const std::string& config_file_or_inline) {
+    static config_singleton_t instance(config_file_or_inline);
+    return instance.config_;
```
kevinkreiser, line 38: just move this down into the namespace level function below and remove this function. we dont need two ways to do the same thing, do we?

valhalla/configuration.h
```diff
@@ -0,0 +1,49 @@
+#ifndef VALHALLA_CONFIGURATION_H_
+#define VALHALLA_CONFIGURATION_H_
+
+#include <stdexcept>
+#include <string>
+
+#include <boost/property_tree/ptree.hpp>
+
+#include "baldr/rapidjson_utils.h"
+#include "filesystem.h"
```
kevinkreiser, line 10: since we are on c++17 now you can use std::filesystem instead. we should try to get away from using the our port over time.

valhalla/configuration.h
kevinkreiser: (you'd need to add the namespace function as a friend function of this class to make that work i guess)

valhalla/configuration.h
```diff
@@ -0,0 +1,49 @@
+#ifndef VALHALLA_CONFIGURATION_H_
```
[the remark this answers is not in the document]
janusz-anue: Yeah, after doing that I needed to resolve some naming collisions (config function and config variable in anonymous namespace) in some testifies. I resolved them by just renaming the var to cfg. Hope that's ok

valhalla/configuration.h
janusz-anue: Added the function as a friend, needed to declare the function before defining it, that's why I added the declaration above the singleton

janusz-anue:
When playing around with the new singleton I was noticing that when used with a long inline config the check to see if the argument is a file or inline config threw an exception. This is now mitigated in the new try-catch block in the constructor.

Additionally I was finding that when I read the singleton in the GraphTile for example, many tests will fail because the tests (that use the GraphTile in any way) logically don't initialise the singleton. So I thought to create a new exception for when the singleton is not initialised that can be caught and handled accordingly. I can expand on that in a new (draft) PR. Another solution could be to add a member var that is true when the config is initialised, otherwise false.

johannes-no:
Hi @kevinkreiser and/or @nilsnolde, could you please take another look at this PR so that we can use the config singleton to complete the original task [#4177](https://github.com/valhalla/valhalla/issues/4177) . For the speed fading the PR is prepared but i would wait to open it until the merge of this PR.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
    * ADDED: support for `:forward` and `:backward` for `motor_vehicle`, `vehicle`, `foot` and `bicycle` tag prefixes [#4204](https://github.com/valhalla/valhalla/pull/4204)
    * ADDED: add `valhalla_build_landmarks` to parse POIs from osm pbfs and store them as landmarks in the landmark sqlite database [#4201](https://github.com/valhalla/valhalla/pull/4201)
    * ADDED: add primary key in the landmark sqlite database and a method to retrieve landmarks via their primary keys [#4224](https://github.com/valhalla/valhalla/pull/4224)
+   * CHANGED: the boost property tree config is now read in to a singleton that doesn't need to be passed down anymore [#4220](https://github.com/valhalla/valhalla/pull/4220)
```
kevinkreiser, line 26: would you mind moving this to the bottom of the list

test/isochrone.cc
```diff
@@ -30,7 +30,7 @@ using rp = rapidjson::Pointer;
 
 namespace {
 
-const auto config = test::make_config("test/data/utrecht_tiles");
+const auto cfg = test::make_config("test/data/utrecht_tiles");
```
kevinkreiser, line 33: yeah this is an inconvenience. not being able to name something `config` when inside the valhalla namespace, the good news is that in a future pr we wont need to even keep the variable around we can just always call `config()` whereever we are in the code (in most scenarios)

valhalla/config.h
```diff
@@ -31,3 +37,57 @@
...
+
+namespace {
+struct config_singleton_t {
+protected:
+  boost::property_tree::ptree config_;
+
+  config_singleton_t() = delete;
+  config_singleton_t(const std::string& config_file_or_inline) {
+    if (config_file_or_inline.empty()) {
+      throw valhalla::ConfigUninitializedException();
```
kevinkreiser, line 60: why not just use `throw std::runtime_error("Config singleton was not initialized before usage");`? it must be because we have try catch blocks in things like the workers which key off the type of exception? perhaps the workers have a new catch clause that returns a specific error when not configured? or maybe it should be a valhalla_exception_t with a specific error code? fwiw this is fine, dont feel the need to change it, i more just wanted to discuss it/think about it a little.

CHANGELOG.md
kevinkreiser: ill just move this and merge so as not to hold this up any longer. very sorry for the wait

kevinkreiser (approved):

CHANGELOG.md
johannes-no: Thanks Kevin.

Issue #4135, opened 2023-05-30
Landmark Routing 3: Tile Storage of Associated Landmarks


kevinkreiser:
- decide how we will store landmarks associated to nodes/edges/otherwise
- make the changes to the datastructures, both builders and the regular "read only" ones
- write the tests hard to do without the association code but\
- maybe we can use the setters directly on the builder objects in the unit tests and then use the getters to check that the stuff is properly associated in the tiles

kevinkreiser:
so we want to make use of edgeinfo after discussing the original design document some loose steps to do this. the first thing we need is a new executable `valhalla_add_landmarks` all it takes is the config from there it gets the sqlite db and the location of the tiles to which it will add the landmarks. so gets the config loaded then calls `landmark_builder::add_landmarks(ptree)`. so then add_landmarks all it does is:



```python
for t in tiles:
  landmarks = db.query(t.bbox())
  tb = tile_builder(t)
  for l in landmarks:
    edges = t.get_edges_near(l, 10) # 10 meter radius?
    for e in edges:
      tb.add_tag_value(e, to_tag(l), to_value(l))
  tb.store_graph_tile()
```

this is way way simplified but the basic idea is there. the bulk of the work will be in changing tile_builder to add the "add_tag_value" method to it and to modify the tag value stuff to be aware of the landmarks. we can get into more detail once implementation starts.

i should also note that we dont want to actually loop over the tiles instead we want to make a thread pool and have them work on the tile list in parallel. to start we can skip threading if that helps

kevinkreiser:
first question we need to answer here is if we are going by tile and we need to associate a given landmark to edges across tile hierarchy levels then we will be searching for edges for the same landmarks multiple times. so we have a few options:

1. do the naive way above where we end up finding edges for the same landmarks over and over but filtering out the ones that dont match the current tile we are updating. its main benefit is its simplicity but it will waste a lot of time searching for and finding the same edges for the same landmarks
2. we could instead only use level 0 (largest area tiles) as jobs. then when we get edges on levels 1 and 0 we keep these in their respective lists and update those child tiles. this is still a pretty simple setup but the book keeping is kind of annoying. first in the worst case we will have 21 tiles to update from one job. the area is large which means each thread may need quite a decent amount of storage while its looking up edges for that whole patch.
[...]

PR #4224, opened 2023-07-27
Add primary key for landmark database and landmark getter via primary key

vesperlou:
# Issue
To faciliate landmark storage in tiles, primary key is added in landmark databse to enable getting a landmark via its primary key.
 
## Tasklist

 - [x] add primary key to the landmarks database 
 - [x] add a method to be able to retreive a landmark from the db via its primary key

## Requirements / Relations

[Landmark Routing 3: Tile Storage of Associated Landmarks](https://github.com/valhalla/valhalla/issues/4135)

### DIFF
--- a/src/mjolnir/landmark_builder.cc
+++ b/src/mjolnir/landmark_builder.cc
@@ -185,50 +185,50 @@ void LandmarkDatabase::insert_landmark(const std::string& name,
 }
 
 // get multiple landmarks by their ids
+/** TODO: Currently this function dynamically creates query statement based on
+ *  the number of provided primary keys.
+ *  In the future, we may consider implementing a fix-sized batch retrieval approach, where
+ *  multiple landmarks are retrieved in batches using a prepared SQL statement with a fixed
+ *  number of placeholders (e.g., 10 question marks) to be filled with corresponding inputs.
+ *  If the caller provides more than the fixed number of inputs, the function will automatically
+ *  perform multiple batch retrieves.
+ */
 std::vector<Landmark> LandmarkDatabase::get_landmarks_by_ids(const std::vector<int64_t>& pkeys) {
-  sqlite3_stmt* get_landmarks_by_ids_stmt = nullptr;
-
-  // prepare the sql statement with appropriate number of question marks for the ids
+  // create the sql statement with inputs
   std::string sql = "SELECT id, name, type, X(geom), Y(geom) FROM landmarks WHERE id IN (";
   for (size_t i = 0; i < pkeys.size(); ++i) {
     if (i > 0) {
       sql += ", ";
     }
-    sql += "?";
+    sql += std::to_string(static_cast<int>(pkeys[i]));
   }
   sql += ")";
 
-  int ret =
-      sqlite3_prepare_v2(pimpl->db, sql.c_str(), sql.length(), &get_landmarks_by_ids_stmt, NULL);
+  // callback for the sql query
+  auto populate_landmarks = [](void* data, int argc, char** argv, char** col_names) {
+    std::vector<Landmark>* landmarks = static_cast<std::vector<Landmark>*>(data);
+
+    int64_t landmark_id = static_cast<int64_t>(std::stoi(argv[0]));
+    const char* landmark_name = argv[1];
+    int landmark_type = std::stoi(argv[2]);
+    double lng = std::stod(argv[3]);
+    double lat = std::stod(argv[4]);
+
+    landmarks->emplace_back(landmark_id, landmark_name, static_cast<LandmarkType>(landmark_type), lng,
+                            lat);
+    return 0;
+  };
+
+  std::vector<Landmark> landmarks;
+  char* err_msg = nullptr;
+  // execute query
+  int ret = sqlite3_exec(pimpl->db, sql.c_str(), populate_landmarks, &landmarks, &err_msg);
+
+  // check for errors in the sql execution
   if (ret != SQLITE_OK) {
-    throw std::runtime_error("Sqlite prepared get landmarks by ids statement error: " +
-                             std::string(sqlite3_errmsg(pimpl->db)));
+    throw std::runtime_error("Sqlite execution error: " + std::string(err_msg));
   }
 
-  // bind ids to the prepared statement
-  for (size_t i = 0; i < pkeys.size(); ++i) {
-    sqlite3_bind_int64(get_landmarks_by_ids_stmt, static_cast<int64_t>(i + 1), pkeys[i]);
-  }
-
-  // execute the statement and fetch the results
-  std::vector<Landmark> landmarks{};
-  while ((ret = sqlite3_step(get_landmarks_by_ids_stmt)) == SQLITE_ROW) {
-    int64_t landmark_id = static_cast<int64_t>(sqlite3_column_int64(get_landmarks_by_ids_stmt, 0));
-    const char* name =
-        reinterpret_cast<const char*>(sqlite3_column_text(get_landmarks_by_ids_stmt, 1));
-    int landmark_type = sqlite3_column_int(get_landmarks_by_ids_stmt, 2);
-    double lng = sqlite3_column_double(get_landmarks_by_ids_stmt, 3);
-    double lat = sqlite3_column_double(get_landmarks_by_ids_stmt, 4);
-
-    landmarks.emplace_back(landmark_id, name, static_cast<LandmarkType>(landmark_type), lng, lat);
-  }
-
-  if (ret != SQLITE_DONE) {
-    throw std::runtime_error("Sqlite could not get landmarks with given keys: " +
-                             pimpl->last_error());
-  }
-
-  sqlite3_finalize(get_landmarks_by_ids_stmt);
   return landmarks;
 }
 
--- a/src/mjolnir/landmark_builder.cc
+++ b/src/mjolnir/landmark_builder.cc
@@ -61,12 +61,12 @@ struct LandmarkDatabase::db_pimpl {
   sqlite3* db;
   sqlite3_stmt* insert_stmt;
   sqlite3_stmt* bounding_box_stmt;
-  sqlite3_stmt* get_landmark_stmt;
+  sqlite3_stmt* get_landmark_by_id_stmt;
   std::shared_ptr<void> spatial_lite;
   bool vacuum_analyze = false;
 
   db_pimpl(const std::string& db_name, bool read_only)
-      : insert_stmt(nullptr), bounding_box_stmt(nullptr), get_landmark_stmt(nullptr) {
+      : insert_stmt(nullptr), bounding_box_stmt(nullptr), get_landmark_by_id_stmt(nullptr) {
     // create parent directory if it doesn't exist
     const filesystem::path parent_dir = filesystem::path(db_name).parent_path();
     if (!filesystem::exists(parent_dir) && !filesystem::create_directories(parent_dir)) {
@@ -139,8 +139,10 @@ struct LandmarkDatabase::db_pimpl {
     }
 
     // prep the landmark getter statement
-    const char* get_landmark = "SELECT id, name, type, X(geom), Y(geom) FROM landmarks WHERE id = ?";
-    ret = sqlite3_prepare_v2(db, get_landmark, strlen(get_landmark), &get_landmark_stmt, NULL);
+    const char* get_landmark_by_id =
+        "SELECT id, name, type, X(geom), Y(geom) FROM landmarks WHERE id = ?";
+    ret = sqlite3_prepare_v2(db, get_landmark_by_id, strlen(get_landmark_by_id),
+                             &get_landmark_by_id_stmt, NULL);
     if (ret != SQLITE_OK) {
       throw std::runtime_error("Sqlite prepared landmark getter statement error: " +
                                std::string(sqlite3_errmsg(db)));
@@ -160,7 +162,7 @@ struct LandmarkDatabase::db_pimpl {
 
     sqlite3_finalize(insert_stmt);
     sqlite3_finalize(bounding_box_stmt);
-    sqlite3_finalize(get_landmark_stmt);
+    sqlite3_finalize(get_landmark_by_id_stmt);
     sqlite3_close_v2(db);
   }
   std::string last_error() {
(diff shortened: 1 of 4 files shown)


src/mjolnir/landmark_builder.cc
```diff
@@ -129,12 +131,20 @@ struct LandmarkDatabase::db_pimpl {
...
-        "SELECT name, type, X(geom), Y(geom) FROM landmarks WHERE ST_Covers(BuildMbr(?, ?, ?, ?, 4326), geom)";
+        "SELECT id, name, type, X(geom), Y(geom) FROM landmarks WHERE ST_Covers(BuildMbr(?, ?, ?, ?, 4326), geom)";
     ret = sqlite3_prepare_v2(db, select, strlen(select), &bounding_box_stmt, NULL);
     if (ret != SQLITE_OK) {
       throw std::runtime_error("Sqlite prepared select statement error: " +
                                std::string(sqlite3_errmsg(db)));
     }
+
+    // prep the landmark getter statement
+    const char* get_landmark = "SELECT id, name, type, X(geom), Y(geom) FROM landmarks WHERE id = ?";
```
nilsnolde, line 142: for clarity I'd rename `get_landmark` to `get_landmark_by_id`  and everything bbox related to `get_landmarks_by_bbox`.

src/mjolnir/landmark_builder.cc
```diff
@@ -183,6 +194,38 @@ void LandmarkDatabase::insert_landmark(const std::string& name,
...
+  auto* get_landmark_stmt = pimpl->get_landmark_stmt;
+
+  sqlite3_reset(get_landmark_stmt);
+  sqlite3_clear_bindings(get_landmark_stmt);
+
+  sqlite3_bind_int(get_landmark_stmt, 1, pkey);
+  LOG_TRACE(sqlite3_expanded_sql(get_landmark_stmt));
+
+  int ret = sqlite3_step(get_landmark_stmt);
+  if (ret == SQLITE_ROW) {
```
nilsnolde, line 207: instead of if/else, consider doing `if (ret != SQLITE_ROW) { throw ...}`, then you don't need the else

nilsnolde: Ah I see there's an `else if`. Still, the comment holds, else is not necessary.

src/mjolnir/landmark_builder.cc
```diff
@@ -183,6 +194,38 @@ void LandmarkDatabase::insert_landmark(const std::string& name,
...
+    uint32_t landmark_id = static_cast<uint32_t>(sqlite3_column_int(get_landmark_stmt, 0));
+    const char* name = reinterpret_cast<const char*>(sqlite3_column_text(get_landmark_stmt, 1));
+
+    int landmark_type = -1;
+    if (sqlite3_column_type(get_landmark_stmt, 2) != SQLITE_NULL) {
+      landmark_type = sqlite3_column_int(get_landmark_stmt, 2);
+    }
```
nilsnolde, line 214: this can't happen can it? if so, we should throw I think. passing on a value of `-1` would do crazy things.

src/mjolnir/landmark_builder.cc
```diff
@@ -202,17 +245,18 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
...
   while (ret == SQLITE_ROW) {
-    const char* name = reinterpret_cast<const char*>(sqlite3_column_text(bounding_box_stmt, 0));
+    uint32_t landmark_id = static_cast<uint32_t>(sqlite3_column_int(bounding_box_stmt, 0));
+    const char* name = reinterpret_cast<const char*>(sqlite3_column_text(bounding_box_stmt, 1));
 
     int landmark_type = -1;
-    if (sqlite3_column_type(bounding_box_stmt, 1) != SQLITE_NULL) {
-      landmark_type = sqlite3_column_int(bounding_box_stmt, 1);
+    if (sqlite3_column_type(bounding_box_stmt, 2) != SQLITE_NULL) {
+      landmark_type = sqlite3_column_int(bounding_box_stmt, 2);
```
nilsnolde, line 253: same comment holds here

nilsnolde (changes requested):
some nits:)

src/mjolnir/landmark_builder.cc
kevinkreiser: you meant the function name in this case not the prepared statement right?

nilsnolde: both but primarily the (public) function name, right. the prepare statement and variable names could/should be changed similarly, but that's more of a nit

src/mjolnir/landmark_builder.cc
```diff
@@ -95,7 +97,7 @@ struct LandmarkDatabase::db_pimpl {
     if (flags & SQLITE_OPEN_CREATE) {
       // make the table
       const char* table =
-          "SELECT InitSpatialMetaData(1); CREATE TABLE IF NOT EXISTS landmarks (name TEXT, type TEXT)";
+          "SELECT InitSpatialMetaData(1); CREATE TABLE IF NOT EXISTS landmarks (id INTEGER PRIMARY KEY, name TEXT, type TEXT)";
```
kevinkreiser, line 100: i wonder if we need to create an index on the primary key. i read somewhere on the internet that if you declare a primary key in the way that you have here you won't because under the hood sqlite treats it as an alias of the `rowid` (ordinal row number) for the table.

src/mjolnir/landmark_builder.cc
```diff
@@ -202,17 +245,18 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const doub
 
   int ret = sqlite3_step(bounding_box_stmt);
   while (ret == SQLITE_ROW) {
-    const char* name = reinterpret_cast<const char*>(sqlite3_column_text(bounding_box_stmt, 0));
+    uint32_t landmark_id = static_cast<uint32_t>(sqlite3_column_int(bounding_box_stmt, 0));
```
kevinkreiser, line 248:
```suggestion
    uint64_t landmark_id = static_cast<uint64_t>(sqlite3_column_int64(bounding_box_stmt, 0));
```

valhalla/mjolnir/landmark_builder.h
```diff
@@ -60,7 +60,7 @@ enum class LandmarkType : uint8_t {
   casino = 18,
 };
 
-using Landmark = std::tuple<std::string, LandmarkType, double, double>;
+using Landmark = std::tuple<uint32_t, std::string, LandmarkType, double, double>;
```
kevinkreiser, line 63:
```suggestion
using Landmark = std::tuple<int64_t, std::string, LandmarkType, double, double>;
```

valhalla/mjolnir/landmark_builder.h
```diff
@@ -98,11 +98,14 @@ struct LandmarkDatabase {
                        const LandmarkType& type,
                        const double lng,
                        const double lat);
+
   std::vector<Landmark> get_landmarks_in_bounding_box(const double minLat,
                                                       const double minLong,
                                                       const double maxLat,
                                                       const double maxLong);
 
+  Landmark get_landmark(const uint32_t pkey);
```
kevinkreiser, line 107:
```suggestion
  Landmark get_landmark(const int64_t pkey);
```

src/mjolnir/landmark_builder.cc
```diff
@@ -183,6 +194,38 @@ void LandmarkDatabase::insert_landmark(const std::string& name,
   pimpl->vacuum_analyze = true;
 }
 
+Landmark LandmarkDatabase::get_landmark(const uint32_t pkey) {
```
kevinkreiser, line 197: i wonder if we should allow batching, this could improve efficiency when we are fetching these things:

```suggestion
std::vector<Landmark> LandmarkDatabase::get_landmark(const std::vector<int64_t>& pkeys) {
```

src/mjolnir/landmark_builder.cc
kevinkreiser: i agree, i think we can just skip the if completely and set the type without checking null. we dont ever put a null entry in the db so we cant ever get a null entry out

src/mjolnir/landmark_builder.cc
vesperlou: good idea. maybe we should allow both cases

src/mjolnir/landmark_builder.cc
vesperlou: got it

src/mjolnir/landmark_builder.cc
vesperlou: I searched online and it seems SQLite automatically does indexing for primary key.
https://stackoverflow.com/questions/3379292/is-an-index-needed-for-a-primary-key-in-sqlite

src/mjolnir/landmark_builder.cc
vesperlou: I've added `get_landmarks_by_ids` to flexibly fetch multiple landmarks by ids. the statement is not prepared in advance anymore and no min/max limit of landmark number is set. maybe we should set a max number?

test/gurka/test_landmarks.cc
```diff
@@ -108,21 +108,26 @@ TEST(LandmarkTest, TestParseLandmarks) {
...
+  EXPECT_TRUE(std::get<1>(landmarks[0]) == "A");
+  EXPECT_TRUE(std::get<2>(landmarks[1]) == LandmarkType::restaurant); // B
+  EXPECT_TRUE(std::get<1>(landmarks[1]) == "hai di lao");
+  EXPECT_TRUE(std::get<2>(landmarks[2]) == LandmarkType::cinema); // D
+  EXPECT_TRUE(std::get<1>(landmarks[2]) == "wan da");
+
+  // check getting multiple landmarks by ids
+  landmarks.clear();
+  EXPECT_NO_THROW({ landmarks = db.get_landmarks_by_ids({1, 2, 3, 4}); });
+  EXPECT_EQ(landmarks.size(), 3);
```
nilsnolde, line 132: we should talk about allowing to pass IDs which aren't actually in the DB. I guess we could do it like that and the caller should make sure the returned size is as expected, i.e. in this case `4` doesn't exist.

vesperlou: I've done a few tests to make sure the return size is as expected. since we are using `"SELECT id, name, type, X(geom), Y(geom) FROM landmarks WHERE id IN (?, ?, ?, ?)";`, adding non-existing ids to the parameter wouldn't impact on the existing ids and intended results. it won't throw an error either but I guess we don't need that :D

nilsnolde: > adding non-existing ids to the parameter wouldn't impact on the existing ids and intended results

Sure, there's no negative impact per se, but the caller needs to be aware that the return could be less than what it asked for (and that it does not throw an exception if it encounters a missing ID). 

That brings up another important point: before merging this, please make sure there's proper docstrings on the header function declarations. A note in the function's docstring about what I just mentioned would be good.

vesperlou: done. please take a look

src/mjolnir/landmark_builder.cc
```diff
@@ -183,10 +184,58 @@ void LandmarkDatabase::insert_landmark(const std::string& name,
...
+std::vector<Landmark> LandmarkDatabase::get_landmarks_by_ids(const std::vector<int64_t>& pkeys) {
+  sqlite3_stmt* get_landmarks_by_ids_stmt = nullptr;
+
+  // prepare the sql statement with appropriate number of question marks for the ids
+  std::string sql = "SELECT id, name, type, X(geom), Y(geom) FROM landmarks WHERE id IN (";
+  for (size_t i = 0; i < pkeys.size(); ++i) {
+    if (i > 0) {
+      sql += ", ";
+    }
+    sql += "?";
```
nilsnolde, line 197: we can put the actual value here instead of `?` and omit preparing the statement

src/mjolnir/landmark_builder.cc
```diff
@@ -183,10 +184,58 @@ void LandmarkDatabase::insert_landmark(const std::string& name,
   pimpl->vacuum_analyze = true;
 }
 
-std::vector<Landmark> LandmarkDatabase::get_landmarks_in_bounding_box(const double minLat,
-                                                                      const double minLong,
-                                                                      const double maxLat,
-                                                                      const double maxLong) {
+// get multiple landmarks by their ids
+std::vector<Landmark> LandmarkDatabase::get_landmarks_by_ids(const std::vector<int64_t>& pkeys) {
+  sqlite3_stmt* get_landmarks_by_ids_stmt = nullptr;
```
nilsnolde, line 189: we should put a TODO to test how this performs and see if we should prepare a default query (like 10-20 elements) and if more come in we extend the statement preparation

src/mjolnir/landmark_builder.cc
vesperlou: done. but i notice sqlite officially suggests to use preparing statement when query includes variables.

src/mjolnir/landmark_builder.cc
vesperlou: done.

src/mjolnir/landmark_builder.cc
nilsnolde: Likely because they’re doing some internal validation of the variable type or so. I think in our use case here it’s fine to do as it is now, it’s not a multi-variable statement, just the same many times.

valhalla/mjolnir/landmark_builder.h
```diff
@@ -60,7 +60,7 @@ enum class LandmarkType : uint8_t {
   casino = 18,
 };
 
-using Landmark = std::tuple<std::string, LandmarkType, double, double>;
+using Landmark = std::tuple<int64_t, std::string, LandmarkType, double, double>;
```
nilsnolde, line 63: Hm I just realized this: not that we really need it but why not uint64_t for the index?

kevinkreiser: because the primary key has to be a signed 64 bit integer to get a free index on that column in sqlite. why they picked signed i do not know but if we want to use unsigned that means we have to add our own index. anyway this is what i read on the internet. if we were going to do something like that we should forget the primary key and use the osmid directly and add an index manually (this could be useful for debugging at some point?) but i figured at the moment its not needed, it did cross my mind though!

valhalla/mjolnir/landmark_builder.h
```diff
@@ -92,22 +92,69 @@ inline LandmarkType string_to_landmark_type(const std::string& s) {
 
 struct LandmarkDatabase {
 public:
+  /**
+   * LandmarkDatabase constructor.
+   * Create a new LandmarkDatabase object that connects to the SQLite landmark database
+   * with the given `db_name`. The `read_only` parameter determines whether the
+   * database connection is read-only or read-write.
+   * @param db_name The file path of the SQLite database to connect to.
+   * @param read_only Set to true to open the database in read-only mode, false for read-write.
```
kevinkreiser, line 101: very nitpicky but we typically do this extra white space:

```suggestion
   * database connection is read-only or read-write.
   *
   * @param db_name    The file path of the SQLite database to connect to.
   * @param read_only  Set to true to open the database in read-only mode, false for read-write.
```

kevinkreiser (approved):

kevinkreiser:
@nilsnolde are you good with the changes? anything else needed here?

nilsnolde (approved):

valhalla/mjolnir/landmark_builder.h
vesperlou: done

vesperlou:
oops @nilsnolde @kevinkreiser I add a new commit to add extra space in docstrings after your approvals. can either of you review again? :)

nilsnolde (approved):
:ship:

PR #4231, opened 2023-08-02
update submodules which had new releases

nilsnolde:
Originally I just wanted to update pybind & googletest as they were generating ugly CMake warnings on their previous releases.

Then I thought why not have a full update round of all submodules. A few didn't have anything new (microtar, dirent, few others), `cxxopts` had a major version upgrade so I didn't touch that. I know new is not always better (did someone say protobuf?), but if CI agrees, I'd be fine personally.


kevinkreiser:
> is it common practice for C++ libs to move to the next major version if older C++ stdlib version support is dropped?

do you mean, for example, that valhalla would go to version 4 because it started targetting c++17? if one is following semver (which i do think is ubiquitous) and one actually makes use of c++17 features that it exposes in its public api then yes they would need to update their major version as code dependent on them wouldnt build otherwise (if they were using that changed API). if they upgraded what standard they targetted but their public api didnt change as a result one could still build dependent code bases without even targetting the new version themselves and this would remain ABI compatible and as such semver wouldnt require a major version change (patch only)

nilsnolde:
Should we merge this @kevinkreiser ?

kevinkreiser (approved):

PR #4233, opened 2023-08-02
Update graph tile to allow adding landmarks

vesperlou:
# Issue
Update the datastructures in the tile (edgeinfo tags, etc.) and the tilebuilder interface to allow for adding landmarks to an existing graph tile, with tests.

## Requirements / Relations
[Landmark Routing 3: Tile Storage of Associated Landmarks](https://github.com/valhalla/valhalla/issues/4135)


vesperlou:
@kevinkreiser @nilsnolde HELP!! the current code fails sanity checks in `graphtilebuilder.cc` and `graphtile.cc`. 

The code skeleton https://github.com/valhalla/valhalla/pull/4233/files#diff-8189d3cf78ffc8f6d7ee3ee746acf277aea5615c4a6da4d43f31f126472ef573R1303-R1304 try to add data to the `EdgeInfoBuilder` without updating its primary key, i.e. without changing `edgeinfo_offset_map_` and `edge_offset_map_`. I think the sanity checks fail bacause of that. I've been working on it for a day trying to make both work, but still can't solve it. 
Any idea?
Or should I rebuild/update `edgeinfo_offset_map_` and `edge_offset_map_` as well like in `AddEdgeInfo`?

vesperlou:
update:
compared with real tile size in memory, header.end_offset() is always smaller (for a fixed size). 
for example, one landmark added for each direction of an edge comes to 8 byte difference; for two edges it comes to 16. and the difference is not relevant to the size of added landmark itself.

One more day's digging still couldn't solve it completely. I feel I missed something in the code and need your help to solve it :)

src/mjolnir/graphtilebuilder.cc
```diff
@@ -11,6 +11,13 @@
 #include <set>
 #include <stdexcept>
 
+#include "baldr/directededge.h"
+#include "baldr/graphconstants.h"
+#include "baldr/graphid.h"
+#include "baldr/graphtileheader.h"
+#include "mjolnir/edgeinfobuilder.h"
+#include "mjolnir/landmark_builder.h"
```
kevinkreiser, line 19: i am certain some of these are redundant. in fact i'd be the last one is the only one you need

src/mjolnir/landmark_builder.cc
```diff
@@ -232,10 +234,10 @@ std::vector<Landmark> LandmarkDatabase::get_landmarks_by_ids(const std::vector<i
   return landmarks;
 }
 
-std::vector<Landmark> LandmarkDatabase::get_landmarks_by_bbox(const double minLat,
-                                                              const double minLong,
-                                                              const double maxLat,
-                                                              const double maxLong) {
+std::vector<Landmark> LandmarkDatabase::get_landmarks_by_bbox(const double minLong,
+                                                              const double minLat,
+                                                              const double maxLong,
```
kevinkreiser, line 239: id use either lon or lng `long` is an integer type so is confusing to be used here

src/mjolnir/graphtilebuilder.cc
```diff
@@ -1258,5 +1265,63 @@ void GraphTileBuilder::UpdatePredictedSpeeds(const std::vector<DirectedEdge>& di
...
+  // like:
+  std::string tagged_value = landmark.to_str();
+
+  // do we already have this record, if so we dont want a copy instead we just want the offset
+    auto name_offset = AddName(tagged_value);
+  
+  // so now we know where we are storing this variable sized string in the tile, we need to keep the
+  // record in the edge info of where that was
+  NameInfo ni{name_offset, 0, 0, 1};
+  eib->second->AddNameInfo(ni);
```
nilsnolde, line 1303: after you add landmarks to the edgeinfobuilders, you'll need to update all offsets of structures that come after the edgeinfos in the binary tile, including`header->set_end_offset()`. You can have a look a few lines up in `GraphTileBuilder::AddBins` @vesperlou

nilsnolde: ok so that wasn't it. of course the StoreTileData function is supposed to update the offset stuff. the problem is here: https://github.com/valhalla/valhalla/blob/75b5e818eb694a47d1b21c5cebf0dad45e476537/src/mjolnir/graphtilebuilder.cc#L362. That `edge_info_offset_` is weird.. it's not an offset, it's a size. IMO the code should be refactored to remove that variable entirely, it doesn't seem to serve any purpose. We can get the size of the edgeinfo's dynamically in StoreTileData. The problem is that var is never updated. Which wasn't a problem so far because no other data builder uses StoreTileData AND does anything to edge infos, so far it's always been the same size. 

Let me draft a PR that'll unblock you here. FWIW, this patch is the main change:
```diff
diff --git a/src/mjolnir/graphtilebuilder.cc b/src/mjolnir/graphtilebuilder.cc
index 4c1fbdb7e..2cd265dac 100644
--- a/src/mjolnir/graphtilebuilder.cc
+++ b/src/mjolnir/graphtilebuilder.cc
@@ -352,14 +352,16 @@ void GraphTileBuilder::StoreTileData() {
     }
 
     // Write the edge data
+    uint32_t edge_info_size = 0;
     header_builder_.set_edgeinfo_offset(header_builder_.complex_restriction_reverse_offset() +
                                         reverse_restriction_size);
     for (const auto& edgeinfo : edgeinfo_list_) {
       in_mem << edgeinfo;
+      edge_info_size += edgeinfo.SizeOf();
     }
 
     // Write the names
-    header_builder_.set_textlist_offset(header_builder_.edgeinfo_offset() + edge_info_offset_);
+    header_builder_.set_textlist_offset(header_builder_.edgeinfo_offset() + edge_info_size);
     for (const auto& text : textlistbuilder_) {
       in_mem << text << '\0';
     }
```

kevinkreiser: other places in the code update that when you add to the edgeinfos. we could so this in addlandmark

kkreiser-sc: here you could just call SizeOf before and SizeOf after adding the name info and add the difference to the edge_info_offset_ (which yes is in bytes)

kevinkreiser: here you could just call SizeOf before and SizeOf after adding the name info and add the difference to the `edge_info_offset_` (which yes is in bytes)

vesperlou: got it thanks

valhalla/baldr/landmark.h
```diff
@@ -0,0 +1,187 @@
...
+  return (((uint64_t(lng * 1e7) + uint64_t(180 * 1e7)) & ((1ull << 32) - 1)) << 31) |
+          ((uint64_t(lat * 1e7) + uint64_t(90 * 1e7)) & ((1ull << 31) - 1));
+}
+
+inline std::pair<double, double> decode_lnglat(uint64_t location) {
+  double lng = (int64_t((location >> 31) & ((1ull << 32) - 1)) - 180 * 1e7) * 1e-7;
+  double lat = (int64_t(location & ((1ull << 31) - 1)) - 90 * 1e7) * 1e-7;
+
+  return std::make_pair(lng, lat);
+}
```
vesperlou, line 105: @kevinkreiser @nilsnolde I took `pointll.h` as reference and updated the conversions of lng and lat and now they work as expected. But the returned string size is 8 bytes now. I didn't success on 4 bytes.

vesperlou:
@kevinkreiser @nilsnolde and now the critical problem is `EdgeInfo::GetNamesAndTypes()` cannot get correct name string for landmarks. 
Had a few debugging, the results show that `name_info_list`, `name_info_offset`, and `text_list` are all correct. Landmarks are stored in `edge_info` at the expected size and index. 
`EdgeInfo::GetNamesAndTypes()` can get to the right starting point of the tagged name, but it cannot read the full string.
When `kLandmark = 6` it just reads an empty string; when `kLandmark = static_cast<uint8_t>('3')` it can read the starting byte `kLandmark` but only this byte.
I upload a picture in our chat channel to better explain the situation.

Do you have any idea?

kevinkreiser:
@vesperlou check how the other ones like layer and level are encoded and make sure we are doing the same thing in the landmark::to_string method. we definitely do not want to use the char '3'

src/mjolnir/graphtilebuilder.cc
```diff
@@ -1258,5 +1265,114 @@ void GraphTileBuilder::UpdatePredictedSpeeds(const std::vector<DirectedEdge>& di
...
+  if (eib == edgeinfo_offset_map_.end()) {
+    throw std::runtime_error("Couldn't find edge info for the given edge: " +
+                             std::to_string(static_cast<int>(edge_id.id())));
+  }
+
+  // we need to use the edgeinfobuilder to add a new tagged value record to it that takes some work
+
+  // first we construct the string that makes up the record we want to store, ours is going to look
+  // like:
+  std::string tagged_value = landmark.to_str();
```
kevinkreiser, line 1335: need a check that this string doesnt include any null chars, throw std::logic_error if so, also need to add this to the AddEdgeInfo function in 3 places above

src/baldr/edgeinfo.cc
```diff
@@ -195,6 +186,38 @@ EdgeInfo::GetNamesAndTypes(bool include_tagged_values) const {
...
+      try {
+        // due to short sightedness on pronunciations we have a dichotomy here
+        // basically you can either get other stuff OR pronunciations not both..
+        TaggedValue tv = static_cast<baldr::TaggedValue>(value[0]);
+        if ((tv == baldr::TaggedValue::kPronunciation) != only_pronunciations) {
+          continue;
+        }
+        // add a per tag parser that returns 0 or more strings, parser skips tags it doesnt know
+        std::vector<std::string> contents = parse_tagged_value(value);
+        tagged_values.insert(tagged_values.end(), contents.begin(), contents.end());
```
nilsnolde, line 210: do a move here instead

kevinkreiser:
strange the same test is timing out consistently...

src/baldr/edgeinfo.cc
```diff
@@ -1,7 +1,7 @@
 #include "baldr/edgeinfo.h"
 #include "baldr/graphconstants.h"
 
-#include "midgard/encoded.h"
+#include "baldr/landmark.h"
```
kevinkreiser, line 4:
```suggestion
```

src/baldr/edgeinfo.cc
```diff
@@ -32,6 +32,36 @@ json::ArrayPtr names_json(const std::vector<std::string>& names) {
...
+        pos += 3;
+        names.emplace_back((std::string(reinterpret_cast<const char*>(&header), 3) +
+                            std::string((ptr + pos), header.length_)));
+
+        pos += header.length_;
+      }
+      return names;
+    }
+    case TaggedValue::kLandmark:
+      return {std::string(ptr, find_landmark_taggedvalue_end(ptr))};
```
kevinkreiser, line 59: just bring the function from the header over here

src/mjolnir/graphtilebuilder.cc
```diff
@@ -43,7 +46,7 @@ std::vector<ComplexRestrictionBuilder> DeserializeRestrictions(char* restriction
 } // namespace
 
 // Constructor given an existing tile. This is used to read in the tile
-// data and then add to it (e.g. adding node connections between hierarchy
+// data and then add to it (e.g. adding edge_idnode connections between hierarchy
```
kevinkreiser, line 49:
```suggestion
// data and then add to it (e.g. adding node connections between hierarchy
```

test/gurka/test_landmarks.cc
```diff
@@ -142,3 +141,101 @@ TEST(LandmarkTest, TestParseLandmarks) {
...
+                << static_cast<int>(landmark.type) << " " << landmark.lng << " " << landmark.lat
+                << std::endl;
+
+      // check data correctness of the fixed landmark
+      // EXPECT_EQ(landmark.id, 0);
+      // EXPECT_EQ(landmark.name, "fixed landmark");
+      // EXPECT_EQ(landmark.type, LandmarkType::casino);
+      // EXPECT_EQ(landmark.lng, lng);
+      // EXPECT_EQ(landmark.lat, lat);
+    }
```
kevinkreiser, line 239: this needs to be finished, printing doesnt count as tests ;)

valhalla/mjolnir/edgeinfobuilder.h
```diff
@@ -103,6 +103,20 @@ class EdgeInfoBuilder {
    */
   std::size_t SizeOf() const;
 
+  /**
+   * Check if the given name_info exists in this edge info.
+   * @param  offset  The offset of the name info.
+   * @return  Returns true if the name info exists, false otherwise.
+   */
+  bool has_name_info(uint32_t offset) const {
+    for (const auto& name : name_info_list_) {
```
kevinkreiser, line 112: if adding landmarks turns out to be slow this is one thing we could speed up with an unordered_set of offsets as they are added and when they are deserialized

src/baldr/edgeinfo.cc
vesperlou: oops sorry I forgot to do that. not good at remembering things...

kevinkreiser (changes requested):
there are still minor things to fix but the biggest one is the test is incomplete

test/gurka/test_landmarks.cc
vesperlou: ok will get it done

test/gurka/test_landmarks.cc
```diff
@@ -142,3 +148,128 @@ TEST(LandmarkTest, TestParseLandmarks) {
...
+  // make a config, though we dont need the landmarks db in there until the next test
+  landmark_map.config =
+      test::make_config(workdir, {{"mjolnir.landmarks_db", db_path}},
+                        {{"additional_data", "mjolnir.traffic_extract", "mjolnir.tile_extract"}});
+
+  // build regular graph tiles from the pbf that we have already made, there wont be landmarks in them
+  mjolnir::build_tile_set(landmark_map.config, {pbf_filename}, mjolnir::BuildStage::kInitialize,
+                          mjolnir::BuildStage::kValidate, false);
+
+  // load one of the graphtiles via the graphtilebuilder with the deserialize option turned on
```
nilsnolde, line 163: some comments are good, like this one, but lots of this file has comments which were instructions for you. please clean the comments up a bit and leave only those explaining what's happening

test/gurka/test_landmarks.cc
```diff
@@ -142,3 +148,128 @@ TEST(LandmarkTest, TestParseLandmarks) {
...
+      // check data correctness for flexible landmarks
+      std::vector<PointLL> shape = ei.shape();
+      auto point = shape[shape.size() / 2];
+      auto ltype = static_cast<LandmarkType>((edge_index + 1) % invalid_landmark);
+
+      EXPECT_EQ(landmark.id, 0);
+      EXPECT_EQ(landmark.name, std::to_string(edge_index++));
+      EXPECT_EQ(landmark.type, ltype);
+      EXPECT_EQ(round_to(landmark.lng), round_to(point.first));
+      EXPECT_EQ(round_to(landmark.lat), round_to(point.second));
```
nilsnolde, line 234: instead of this you can do EXPECT_NEAR

test/gurka/test_landmarks.cc
```diff
@@ -142,3 +148,128 @@ TEST(LandmarkTest, TestParseLandmarks) {
...
+
+      // check data correctness of the fixed landmark
+      // EXPECT_EQ(landmark.id, 0);
+      // EXPECT_EQ(landmark.name, "fixed landmark");
+      // EXPECT_EQ(landmark.type, LandmarkType::casino);
+      // EXPECT_EQ(round_to(landmark.lng), round_to(lng));
+      // EXPECT_EQ(round_to(landmark.lat), round_to(lat));
+    }
+  }
+}
```
nilsnolde, line 275: newline at end of file missing, I think there's still some unix tools which bark at this..

test/gurka/test_tagged_values.cc
```diff
@@ -31,6 +31,7 @@ class TaggedValues : public ::testing::Test {
                           H
     )";
 
+    // TODO: does the graphparser make sure we dont ever try to set layer to 0 since its the default
```
nilsnolde, line 34: can you elaborate?

nilsnolde (commented):
Just a bit of cleanup left in the test file.

test/gurka/test_landmarks.cc
vesperlou: good idea, done

test/gurka/test_landmarks.cc
vesperlou: done :)

test/gurka/test_landmarks.cc
vesperlou: done. I'm suprised pre-commit didn't find this

test/gurka/test_landmarks.cc
```diff
@@ -230,15 +208,16 @@ TEST(LandmarkTest, TestTileStoreLandmarks) {
...
 
       // check data correctness of the fixed landmark
+
       // EXPECT_EQ(landmark.id, 0);
       // EXPECT_EQ(landmark.name, "fixed landmark");
       // EXPECT_EQ(landmark.type, LandmarkType::casino);
-      // EXPECT_EQ(round_to(landmark.lng), round_to(lng));
-      // EXPECT_EQ(round_to(landmark.lat), round_to(lat));
+      // EXPECT_NEAR(landmark.lng, lng, rounding_error);
+      // EXPECT_NEAR(landmark.lat, lat, rounding_error);
```
nilsnolde: this should be part of the cleanup IMO. also further down. sorry for the nits, but without any comment why we keep commented code in tests, it only adds confusion. if there is a reason, best to leave to a comment about it

test/gurka/test_tagged_values.cc
vesperlou: I think kevin added the comment here. the gist is the parser currently suppose kLayer doesn't include any null bytes. if in any case kLayer is set to zero the parser will return an problematic empty string. So we'd better make sure kLayer is never set to 0, or use special parser for it which can parse null byte as the one for kLandmark.

test/gurka/test_landmarks.cc
vesperlou: https://github.com/valhalla/valhalla/pull/4233/files#diff-6b826629e0424f344205ddbff8872b3d9a1c3bd856ca21cece6d226656ab0aebR164
there is small comment on this - we can either try to test the fixed landmark or flexible ones. I tried to keep both but I guess we have the flexible landmarks and there is no need to keep the fixed one :) already cleaned up.

nilsnolde (approved):

kevinkreiser (approved):

PR #4255, opened 2023-08-15
add option to valhalla_build_extract to use existing tar file for lookup/copy

nilsnolde:
When running an instance of Valhalla, what one usually has is the tar file, and often no plain tile directory. Then `valhalla_build_extract` can't be used, which is a pity. At least for the public instance, I regularly have that "problem" and I don't want to un-tar those 80 GB before I can rip a few out.

This PR adds support to build extracts from tar files.


scripts/valhalla_build_extract
```diff
@@ -53,6 +53,38 @@ class TileHeader(ctypes.Structure):
     ]
 
 
+class TileResolver:
```
nilsnolde, line 56: a wrapper so we don't have to care about whether we're pulling tiles from a dir or tar

scripts/valhalla_build_extract
```diff
@@ -66,6 +98,19 @@ parser.add_argument(
     type=str,
     default='{}',
 )
+parser.add_argument(
+    "-e",
+    "--extract-tar",
+    help="If specified, will build an extract from an existing tar file at 'mjolnir.tile_extract' and save it to this specified path.",
+    type=str,
+    default=""
+)
```
nilsnolde, line 107: so if someone specifies e.g. `-e new_tar.tar`, we'd write the result to `./new_tar.tar` instead of what specified in the config. the tar specified in the config is instead taken as input.

scripts/valhalla_build_extract
```diff
@@ -342,27 +379,35 @@ if __name__ == '__main__':
...
+        sys.exit(1)
+    elif Path(tiles_extract_out).is_file() and not args.overwrite:
+        LOGGER.critical(f"File exists. Specify --overwrite to overwrite {tiles_extract_out}")
+        sys.exit(1)
+
+    # prefer the tar file to extract from, fall back to the tile dir
     tiles_dir: Path = Path(config["mjolnir"].get("tile_dir", '/dev/null'))
-    if not tiles_dir.is_dir():
+    # optionally use the tile extract to find the tiles
+    if args.extract_tar and (tiles_extract_in := Path(config["mjolnir"].get("tile_extract"))).is_file():
```
nilsnolde, line 394: first time ever I used the cute "walrus" operator haha it's python3.8 (3.7 is EOL by now)

nilsnolde:
There's still some small stuff failing, will look at it tmrw.

test/scripts/test_valhalla_build_extract.py
```diff
@@ -37,41 +37,48 @@ def tile_base_to_path(base_x: int, base_y: int, level: int, fake_dir: Path) -> P
...
 
-    return fake_dir.joinpath(path + ".gph")
+    return Path(path + ".gph")
 
 
 class TestBuildExtract(unittest.TestCase):
     def test_tile_intersects_bbox(self):
         # bogus tile dir
-        tile_dir = Path("/home/")
+        tile_dir = Path("/foo/")
```
nilsnolde, line 46: `/home` is far away from a "bogus dir" haha

scripts/valhalla_build_extract
```diff
@@ -173,11 +223,9 @@ def get_tiles_with_bbox(all_tile_paths: List[Path], bbox_str: str, tiles_dir_: P
         LOGGER.critical(f"Bbox invalid: {list(bbox)}")
         sys.exit(1)
 
-    tile_paths_ = set()
```
nilsnolde, line 176: this used to collect a `set` before. in python, that's an unordered structure, so really we were producing unordered tar files :roll_eyes: my bad.. it's a list now and we deduplicate just before writing it to the tar

nilsnolde: for reference: we can get duplicate tiles for `--geojson-dir` where multiple polygons intersect the same tile.

kevinkreiser (dismissed):

kevinkreiser (approved):

Issue #3564, opened 2022-03-09
 Integrate traffic data into Valhalla routing engine.


suryavanshikunal:
I want to integrate 5 min real time traffic data into my Valhalla routing engine.

[Here is an article](https://valhalla.readthedocs.io/en/latest/thor/simple_traffic/) which has great theory but absolutely has no practice. Besides that article and [this issue](https://github.com/valhalla/valhalla/issues/1284), https://github.com/valhalla/valhalla/issues/3414 nothing else is found.

How can I integrate traffic data into my Valhalla routing engine? 
how to generate the traffic data or what format it required.

kevinkreiser:
There is another issue where this is currently being asked. I'll answer that one and link it here I just haven't had the time of late.

suryavanshikunal:
@kevinkreiser Hope you answer the issue, can you please link that issue here.

vinayakkulkarni:
Few related issues:

1. https://github.com/valhalla/valhalla/issues/2097#issuecomment-907609865, 
2. https://github.com/valhalla/valhalla/issues/3180#issuecomment-907609739
3. https://github.com/valhalla/valhalla/issues/2083#issuecomment-891177939
4. https://github.com/valhalla/valhalla/issues/2840#issuecomment-787426224

nilsnolde:
fwiw, the last 2 don't have much to do with traffic integration

kevinkreiser:
what i'll do is write a readme about traffic integration in valhalla, the headings will be something like:

* historical traffic
    * data layout
    * adding it to the graph
* real time traffic and incidents
    * data layout
    * adding it to the graph
* association to the graph
    * via (OSM) id
    * via openlr
    
give me some time though, im quite busy at the moment

suryavanshikunal:
@kevinkreiser Thank you for your response.  this will really helpful.

vinayakkulkarni:
> what i'll do is write a readme about traffic integration in valhalla, the headings will be something like:
> 
> * historical traffic
>   
>   * data layout
>   * adding it to the graph
> * real time traffic and incidents
>   
>   * data layout
>   * adding it to the graph
> * association to the graph
>   
>   * via (OSM) id
>   * via openlr
> 
> give me some time though, im quite busy at the moment

Would love to help out over the weekend @kevinkreiser, do we have a Discord / Slack for Valhalla for real-time/live collab?

vinayakkulkarni:
[...]

Issue #3180, opened 2021-06-27
Unable to update live speed with add_predicted_traffic


kitindir:
Hi everyone, 
I went through a number of documentations, trying to update live speed with no luck. 

I ran the command `valhalla_ways_to_edges --config valhalla.json` to get a mapping between OSM way and Valhalla edge.
I then put some sample data in a csv file inside the `speeds` folder as follows:
<pre>
194313962073,1,1
63820776025,1,1
</pre>

In order to update live speed, I ran `valhalla_add_predicted_traffic -c valhalla.json -t speeds/` 
and below is the output I'm getting: 
<pre>
2021/06/27 20:17:27.653644 [INFO] Adding predicted traffic with 4 threads
2021/06/27 20:17:27.653689 [INFO] Parsing speeds from 0 tiles.
2021/06/27 20:17:27.653889 [INFO] Parsed 0 constrained traffic speeds.
2021/06/27 20:17:27.653905 [INFO] Parsed 0 free flow traffic speeds.
2021/06/27 20:17:27.653920 [INFO] Parsed 0 compressed records.
2021/06/27 20:17:27.653928 [INFO] Updated 0 directed edges.
2021/06/27 20:17:27.653933 [INFO] Duplicate count 0.
2021/06/27 20:17:27.653937 [INFO] Finished
</pre> 

How come live speeds are not updating? What am I missing? 
In the config I have the following:
<pre> 
    "tile_dir": "/root/valhalla/tiles/valhalla_tiles",
    "tile_extract": "/root/valhalla/tiles/valhalla_tiles.tar",
    "timezone": "/root/valhalla/tiles/valhalla_tiles/timezones.sqlite",
    "traffic_extract": "/data/valhalla/traffic.tar",
    "transit_dir": "/data/valhalla/transit",
</pre> 

I read in one of the existing issues that one needs to unset `tile_extract` so I tried both leaving it blank and removing it altogether with no luck. I'm really stuck and don't know what else to do.
Can anybody help please? 

I look forward to your assistance. 
Thanks

nilsnolde:
There's 2 modes of traffic Valhalla supports: live & historical traffic data. You were actually going for the workflow of the latter (predicted as in "what will traffic be in 2 days from now", you can only achieve that with fixed patterns like historical data).

I recently went through the process of creating predicted traffic tiles for TomTom data, so let me know if you need support on that.

[...]

Issue #2097, opened 2019-12-05
Add traffic data to Valhalla


YuriiNskyi:
I want to integrate traffic data into my Valhalla routing engine.

[Here is an article](https://valhalla.readthedocs.io/en/latest/thor/simple_traffic/) which has great theory but absolutely has no practice. Besides that article and [this issue](https://github.com/valhalla/valhalla/issues/1284), nothing else is found.

How can I integrate traffic data into my Valhalla routing engine?

cosmin-mogos:
Is the POC code publicly available?

TheTonda:
I have also been facing the same problem, any updates on this issue?? Kindly update us if it is being looked into or isn't currently worked on(temporarily abandoned).
Also if the predictive-traffic is fully functional, and if so kindly provide us the required documentation and/or set of commands to run valhalla with predictive-traffic(i.e. how to use valhalla_add_predicted_traffic command and the required files' information to run this )
Also if [this](https://valhalla.readthedocs.io/en/latest/speeds/) feature is completed.
Any help/update is highly appreciated thanks.

alinmindroc:
This should help on how to add traffic as predicted speeds: https://github.com/valhalla/valhalla/issues/3180#issuecomment-907609739

Issue #1284, opened 2018-05-22
Traffic Influenced Routing in Valhalla


dnesbitt61:
# Traffic Influenced Routing in Valhalla
Valhalla needs to incorporate logic for associating a speed to a road segment / edge.  This logic generally needs to use a hierarchy of speed data, where there are fallbacks to lower hierarchies where speed does not exist. This document discusses some design considerations for implementing speeds from traffic systems into route computation.

## Current Edge Cost Interface

Costing logic in *sif* currently has a single method to compute and return a cost to traverse an edge:

Cost DynamicCost::EdgeCost(const DirectedEdge* edge) const;

This method relies on all information required to compute the *Cost* (cost and elapsed time) being present within the *DirectedEdge*. This data structure currently contains a speed that can be used in absence of traffic information. However, this method needs to support several speed from several possible sources:

- Current speed from a real-time traffic source
- Time based predicted speed based on historical data. 
- Free flow speed and constrained flow speed
- Fallback speed - e.g. speed limit or speed based on road classification.

The hierarchy listed above presents the general order of speed lookup along an edge: use a real-time speed if it exists (and potentially within X minutes from the route origin), fall back to a predicted speed based on day of week and hour/minute of the day if one exists, fall back to either free flow or constrained speed (or a combination of the two), and finally fall back to a speed limit or speed based on road classification.

These speed sources will be contained in different parts of the Valhalla tiles or, in the case of real-time speed, within companion tiles that are frequently updated and designed to hold dynamic information about an edge.

## *DynamicTile*

The design for implementing traffic within Valhalla assumes that *GraphTile(s)* contain the routing graph topology and static attribution such as edge lengths along with all types of speed data with the exception of current/real-time speed.  Companion tiles, called *DynamicTile(s)*, contain edge information that may change frequently, such as current speed, presence of incidents, traffic camera locations, etc. Another characteristic of *DynamicTiles* is that the edge information may be sparse. In particular, *DirectedEdges* which do not allow driving (e.g. edges opposing oneway travel) will not generally contain any dynamic data. Current speed information may also not be present on even driveable *DirectedEdges* - especially for lower road classes. 

One consideration for the *DynamicTile* is that if *DirectedEdges* are sparsely populated then a quick method to determine if an edge has dynamic data is required. This method likely cannot use unordered_map to look up the presence of speed or dynamic data as this will likely be too slow. Creating an array (sized the same as the *DirectedEdge* list for the tile) with flags indicating 
[...]

PR #4259, opened 2023-08-18
Document traffic feature

Trietes:
# Issue

Currently there is no official documentation available about how to integrate traffic data in valhalla. See e.g. issues #3564 #3180 #2097 #1284. This PR adds a first documentation about it.

I don't have any experience with live traffic information, so I couldn't really do the documentation of it. Neither I did matching OpenLR data to valhalla. The documentation will only be about historical traffic and the mapping of OSM ids to the valhalla graph. I guess that is better then having no documentation at all.

## Tasklist


docs/docs/mjolnir/traffic_tiles.md
```diff
@@ -0,0 +1,73 @@
+# Traffic in Valhalla
+
+Traffic data for routing engines can be divided into two different types:
+
+- *Historical traffic* describes the typically observed speed on a road at a specific point in time. So e.g. the typical speed on a Monday morning at 8 o'clock.
+- *Live traffic* describes the currently observed speed on a road according to the actual traffic situation.
+
+Valhalla supports both types of traffic for all APIs, but there are some excluded use-cases. First, traffic is not supported for multimodal costing. Furthermore, traffic is currently not supported for `arrive_by` timestamps in the `source_to_targets` API. For more information about how to route with traffic data and time information, check the respective API documentation. In the following it will be explained how to integrate historical and live traffic data into the valhalla graph.
```
nilsnolde, line 8: > traffic is currently not supported for `arrive_by` timestamps in the `source_to_targets` API

That's only half true: using timedistance matrix (and less targets than sources), we do support `arrive_by`, so it's mostly a configuration item.

docs/docs/mjolnir/traffic_tiles.md
```diff
@@ -0,0 +1,73 @@
...
+```
+The format of each row is `<osm_way_id>,[<direction: 0 | 1>, <graph_id>]`. Accordingly, the `osm_way_id` is mapped to multiple `graph_id`'s. An `graph_id` can be converted to the required string format according to the `to_string` function of the class `valhalla::baldr::GraphId`, whose implementation can be found in the valhalla repository. The result is a string of the following form `level/tile_id/id` (e.g. `1/47701/130`).
+
+### With OpenLR
+
+TODO
+    
+## Historical Traffic
+
+To make use of historical traffic data in valhalla, the estimated speed (in km/h) of a road segment has to be given in 5 minute intervals covering a whole week. This leads to a total of 2016 speed values per road segment. This data has to be converted to a format which can be used by valhalla and has to be structured properly in a folder hierarchy similar to the valhalla tile hierarchy. Valhalla requires the speed information in following CSV file format:
```
nilsnolde, line 27: high-res speeds is an optional field, the others however are not

docs/docs/mjolnir/traffic_tiles.md
```diff
@@ -0,0 +1,73 @@
...
+```
+valhalla_tiles
+└───0
+│   │
+│   └───003
+│       │   015.gph
+│   
+└───1
+    └───047
+        │   701.gph
```
nilsnolde, line 49: you can check `tree` for linux which pretty prints directories

nilsnolde (commented):
Just a few comments. Also I think we could remove the stale "Simple traffic" document here in https://github.com/valhalla/valhalla/blob/master/docs/docs/thor/simple_traffic.md

nilsnolde:
And I'd call it "historical_traffic.md" and remove the references to live traffic/OpenLR for now.

Trietes:
Thanks a lot for the review @nilsnolde. I just gave it some updateds according to your review. I just need to add some information at which point in time the historical speeds do start (at sunday 00:00?), so there isn't any mess with the days of the week. Then I'll give it another proof read and it would be ready from my side.

nilsnolde:
> at which point in time the historical speeds do start (at sunday 00:00?)

Yep, the datetime library is using the stupid sunday-starts-the-week notion which even creeped into C++20: https://en.cppreference.com/w/cpp/chrono/weekday. Good point though, one of our clients had to find out that bug in our TomTom conversion pipeline where I assumed too much common sense :smile:

docs/mkdocs.yml
```diff
@@ -105,6 +105,7 @@ nav:
       - mjolnir/map_roulette.md
       - mjolnir/map_roulette_blog.md
       - mjolnir/tag_parsing.md
+      - mjolnir/traffic_tiles.md
```
nilsnolde, line 108: this'll also need to change:)

Trietes:
Its now ready from my side.

nilsnolde (approved):

nilsnolde:
Oh wow, didn't think it's more or less half the world starting the week on Sunday: https://vividmaps.com/first-day-of-week-in-different-countries/. Then I'm clearly biased :laughing: I thought it's similar to Celsius/Fahrenheit, maybe a few more countries.

Trietes:
Completely forgott about this one. Should we merge it?

RelxOff (approved):
thank you for adding 👍

Trietes:
I've just updated the Changelog. It should by in-sync with the master now

nilsnolde (dismissed):
I’ll let @kevinkreiser merge this (or anyone else), from my side this is good, thanks

PR #4260, opened 2023-08-19
fix: links on the optimized route API document

witPranav:
Some of the links on the optimized route API [page](https://valhalla.github.io/valhalla/api/optimized/api-reference/) redirect to wrong locations. This PR should fix it.


nilsnolde (approved):
Thanks for the chore. Actually no need for changelog, but you were just following our to-be-overhauled PR template:)

PR #4263, opened 2023-08-22
add sources_to_targets to expansion service

### LINKED ISSUE #3920: Visualize matrix expansion & paths

nilsnolde:
fixes #3920 

for now only the costmatrix. ~~will do the tdmatrix as well later.~~ or maybe rather in a separate PR, CostMatrix is more important.

### DIFF
--- a/src/loki/matrix_action.cc
+++ b/src/loki/matrix_action.cc
@@ -59,7 +59,7 @@ namespace loki {
 void loki_worker_t::init_matrix(Api& request) {
   // we require sources and targets
   auto& options = *request.mutable_options();
-  if (options.action() == Options::sources_to_targets) {
+  if (options.action() == Options::sources_to_targets || options.action() == Options::expansion) {
     parse_locations(options.mutable_sources(), valhalla_exception_t{112});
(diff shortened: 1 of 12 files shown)


src/loki/matrix_action.cc
```diff
@@ -59,7 +59,7 @@ namespace loki {
 void loki_worker_t::init_matrix(Api& request) {
   // we require sources and targets
   auto& options = *request.mutable_options();
-  if (options.action() == Options::sources_to_targets) {
+  if (options.action() == Options::sources_to_targets || options.action() == Options::expansion) {
```
nilsnolde, line 62: just tons of this plumbing. the only real change is adding the expansion_callback member and calling that during the matrix expansion

PR #4268, opened 2023-08-23
Update the street name and sign data processing include language

gknisely:
# Issue

Update the street name and sign data processing include language.

## Data processing updates
The heart of this work involves processing `name:<lg>` and `destination:<lg>` tags. This includes all variations of each of those tags (e.g., `name:left:<lg>, name:right<lg>, official_name:<lg>, destination:street:lang:<lg>, destination:ref:lang:<lg>, etc.`) Within the tag the lg stands for the language. 

### Language updates
In order to determine the languages for an LL we utilized the [default_language](https://wiki.openstreetmap.org/wiki/Key:default_language) that is defined for a country or providence/state. The problem with this is the fact that some areas in the world are multi-lingual and don't have one language (e.g., in Belgium they support Dutch, French, and German and in Switzerland they speak German, French, Romanish, and Italian). However, the value of the default_language would usually contain only one language.  In order to resolve this issue, the administrative builder was updated to handle these special cases like Brussels via processing relations with `boundary=political` with `political_division=linguistic_community` to get areas that are bilingual. However, this logic was still not enough as other areas did not have these special polygons. Therefore, we added an "override" for the languages. So, in the future if we determine that an area is multilingual and we want to support additional language tags in this area. All we have to do is add them to the `supported languages` list. Our list currently consists of the following:

Wales = cy
United Kingdom = en
Ireland = ga
Northern Ireland = ga
Japan = ja and en
Canada = en and fr
Belarus = ru and be
Singapore = en, zh, ms, and ta
Saudi Arabia = ar and en

### Using the languages for the keys
The processing of the new keys is based on the languages that were determined to be "good" for that country or area. OSM users will typically add languages for names and destinations in all parts of the world even though that language may not be spoken in that country. For example, [5th Avenue in NYC](https://www.openstreetmap.org/way/511699185) has name:ru=5-я авеню. Obviously, we do not want to process the Russian name here. Therefore, using the default languages we can toss the tags with languages that we don't want to support. Moreover, we can create a hierarchy for our languages. For instance, Canada supports both English and French. However, in Ottawa, English will be first and in the Québec province French will be first.

### Edge Cases
We will now support names where they differ depending on which side of the street you are driving on. When combined with the multi-languages for some areas, it gets very complex. In this [example](https://www.openstreetmap.org/way/845098764), the official Dutch name differs depending on the municipality. Basically, the border of the towns runs down the middle of the road and on the right side the Dutch name of the street differs from the left side of the street. In part, this leads to the bizarre situation that the street on the Molenbeek side is called Steenweg op Gent and on the Koekelberg side Gentsesteenweg. However, the French name of the street does not change at all.

#### Chaussée de Gand - Steenweg op Gent/Gentsesteenweg Example
Notice that all dashes are removed and processed correctly.   

Driving left to right the Dutch name should be Steenweg op Gent and the French street name(Chaussée de Gand) does not change.  Notice that before we used to return the name tag with dashes and did not have the French and Dutch street names split up.

##### Before
Screenshot from 2022-02-21 15-13-21

##### After 
Screenshot from 2022-02-21 15-13-34

Driving right to left the Dutch name should be Gentsesteenweg and the French street name(Chaussée de Gand) does not change.
##### Before
Screenshot from 2022-02-21 15-14-36

##### After
Screenshot from 2022-02-21 15-14-26

#### name:forward and name:backward is now processed correctly.  
Notice in this [example](https://www.openstreetmap.org/way/58744373) we have name:forward and name:backward tags set; however, before we would just process the name tag.

##### Before - Waltonville Road and Quarry Road returned regardless of direction  

TBD - new image needed

Screenshot from 2022-02-21 15-35-34

##### After - Quarry Road correctly returned 
Screenshot from 2022-02-21 15-36-03

##### Before - Waltonville Road returned regardless of direction
Screenshot from 2022-02-21 15-36-26

##### After - Waltonville Road correctly returned.
Screenshot from 2022-02-21 15-36-38

#### Multilingual names are now processed.
Notice in this [example](https://www.openstreetmap.org/way/24165304) the name tag has both Welsh and English set.  Since we are in Wales we allow both of these languages and process them both.

##### Before - The name and name:en tag are both returned for the street:  Stryd y Castell / Castle Street/Castle Street 
Screenshot from 2022-02-21 15-46-01

##### After - Stryd y Castell and Castle Street both processed correctly and a language of cy is set for Stryd y Castell and en for Castle Street 
Screenshot from 2022-02-21 15-46-19

Contributors @gknisely @dgearhart


lua/admin.lua
```diff
@@ -75,6 +75,7 @@ drive_on_right = {
 ["Turks and Caicos Islands"] = "false",
 ["Tuvalu"] = "false",
 ["Uganda"] = "false",
+["United Kingdom"] = "false",
```
kevinkreiser, line 78: iirc nils removed this because the admin name changed to like great Britain or something or no that each member of the UK became a country level admin poly

gknisely: @kevinkreiser Thank you for catching this issue.  I will merge his updates into mine.

kevinkreiser: https://github.com/valhalla/valhalla/pull/3619

src/baldr/edgeinfo.cc
```diff
@@ -200,12 +227,11 @@ std::vector<std::string> EdgeInfo::GetTaggedValues(bool only_pronunciations) con
     if (ni->name_offset_ < names_list_length_) {
       const char* value = names_list_ + ni->name_offset_;
       try {
-        // due to short sightedness on pronunciations we have a dichotomy here
-        // basically you can either get other stuff OR pronunciations not both..
```
kevinkreiser, line 204: something like this comment still stands,  you are explicitly skipping tagged values which are kLinguistic, the only difference now is you dont have the choice. i feel like this is a gotcha, is the javadoc for the function updated to say kLinguistic is skipped?

src/baldr/edgeinfo.cc
```diff
@@ -273,38 +300,49 @@ std::unordered_map<uint8_t, std::pair<uint8_t, std::string>> EdgeInfo::GetPronun
...
+            name_index = header.name_index_;
+
+            auto iter = index_linguistic_map.insert(std::make_pair(name_index, liguistic_attributes));
+
+            // Edge case.  Sometimes when phonemes exist but the language for that phoneme is not
+            // supported in that area, we toss the phoneme but add the default language for that
+            // name/destination key.  We only want to return the highest ranking phoneme type
+            // over the language.
+            if (!iter.second) {
+              if ((std::get<kLinguisticMapTuplePhoneticAlphabetIndex>(liguistic_attributes) >
```
kevinkreiser, line 326: nit: why have 2 separate `if`s here?

src/mjolnir/admin.cc
```diff
@@ -59,6 +59,70 @@ uint32_t GetMultiPolyId(const std::multimap<uint32_t, multi_polygon_type>& polys
   return index;
```
kevinkreiser, line 59: the changes in the database are breaking right, meaning old code can't use new database (or maybe it can because we just added a couple columns?) but new code certainly cant use old databases. i think this is ok because it doesnt mean compatibility is broken for routing but for data building. i just wanted to call it out. maybe worth putting in the pr description

src/mjolnir/graphenhancer.cc
```diff
@@ -1631,6 +1631,7 @@ void enhance(const boost::property_tree::ptree& pt,
         uint32_t ntrans = nodeinfo.local_edge_count();
         for (uint32_t k = 0; k < ntrans; k++) {
           DirectedEdge& fromedge = tilebuilder->directededge(nodeinfo.edge_index() + k);
+          std::vector<uint8_t> types;
```
kevinkreiser, line 1634: is this needed?

src/thor/triplegbuilder.cc
```diff
@@ -813,41 +889,37 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
...
-        pronunciation->set_alphabet(GetTripPronunciationAlphabet(
-            static_cast<valhalla::baldr::PronunciationAlphabet>((iter->second).first)));
-        pronunciation->set_value((iter->second).second);
+      switch (std::get<2>(name_and_type)) {
+        case kNotTagged: {
+          ProcessNonTaggedValue(trip_edge->mutable_name()->Add(), linguistics, name_and_type,
+                                name_index);
+          break;
+        }
+        case kTunnelTag: {
```
kevinkreiser, line 911: why does only tunnel ever get used, i wonder why bridge is left out

test/gurka/gurka.cc
```diff
@@ -487,8 +487,10 @@ map buildtiles(const nodelayout& layout,
  * @param way_name the way name you want a directed edge for
  * @param end_node the node that should be the target of the directed edge you want
  * @param tile_id optional tile_id to limit the search to
+ * @param way_id optional way_id to limit the search to
  * @return the directed edge that matches, or nullptr if there was no match
  */
+/*
```
kevinkreiser, line 493: what happened to this function?

src/baldr/edgeinfo.cc
```diff
@@ -6,10 +6,14 @@ using namespace valhalla::baldr;
...
-// do not return TaggedValue::kPronunciation
+// do not return TaggedValue::kLinguistic
 bool IsNameTag(char ch) {
   static const std::unordered_set<TaggedValue> kNameTags = {TaggedValue::kBridge,
-                                                            TaggedValue::kTunnel};
+                                                            TaggedValue::kTunnel,
+                                                            TaggedValue::kBssInfo,
+                                                            TaggedValue::kLayer,
+                                                            TaggedValue::kLevel,
+                                                            TaggedValue::kLevelRef};
```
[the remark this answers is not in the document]
gknisely: @kevinkreiser Yeah we just want the ones that are not Linguistic records.  Let me rename

src/thor/triplegbuilder.cc
gknisely: The mothership did not care about them yet.  Updating

test/gurka/gurka.cc
gknisely: my bad.  fixing

src/baldr/edgeinfo.cc
gknisely: updated

src/mjolnir/graphenhancer.cc
gknisely: nope

test/gurka/gurka.cc
gknisely: my bad

src/baldr/edgeinfo.cc
gknisely: updated the comments

kevinkreiser:
i guess mac builds are now broken project wide... i freaking hate CI. i know they have to change and update and stuff but im completely sick of it

kevinkreiser (changes requested):
ok 2 more changes, fix the order of the entries in changelog and undo the formatting of the taginfo.json (you switched from 2 spaces to 3)

nilsnolde:
I remember an email about a deprecated resource class and I kinda thought I/you PR'd that, but maybe I just mentioned it in some chat and forgot about it. Should be an easy fix. I'm more concerned over M1-only from Jan 24 on, but yeah, that pretty much falls in line with 

> i freaking hate CI

gknisely:
> ok 2 more changes, fix the order of the entries in changelog and undo the formatting of the taginfo.json (you switched from 2 spaces to 3)

@kevinkreiser Sorry about that....fixed

kevinkreiser (changes requested):
do we really need 4 copies of the same tests? 
image

in gtest its easy to do permutations where you just vary some component of the test for each run. in this case the pronunciation tag. i see you also use it for the enums from baldr and the directories but its really very trivial to make these test function generic and then call them with permutations. makes it much much easier to maintain, especially when we are talking about a test which is 3k lines

mandeepsandhu:
> do we really need 4 copies of the same tests? image
> 
> in gtest its easy to do permutations where you just vary some component of the test for each run. in this case the pronunciation tag. i see you also use it for the enums from baldr and the directories but its really very trivial to make these test function generic and then call them with permutations. makes it much much easier to maintain, especially when we are talking about a test which is 3k lines

You could use [parametrized fixtures](https://google.github.io/googletest/advanced.html#how-to-write-value-parameterized-tests) in gtest. Something like [this](https://github.com/valhalla/valhalla/blob/master/test/gurka/test_closure_penalty.cc#L230-L237) to call the tests with different languages.

gknisely:
> > do we really need 4 copies of the same tests? image
> > in gtest its easy to do permutations where you just vary some component of the test for each run. in this case the pronunciation tag. i see you also use it for the enums from baldr and the directories but its really very trivial to make these test function generic and then call them with permutations. makes it much much easier to maintain, especially when we are talking about a test which is 3k lines
> 
> You could use [parametrized fixtures](https://google.github.io/googletest/advanced.html#how-to-write-value-parameterized-tests) in gtest. Something like [this](https://github.com/valhalla/valhalla/blob/master/test/gurka/test_closure_penalty.cc#L230-L237) to call the tests with different languages.

@mandeepsandhu yep that was my plan.  ty

gknisely:
> do we really need 4 copies of the same tests? image
> 
> in gtest its easy to do permutations where you just vary some component of the test for each run. in this case the pronunciation tag. i see you also use it for the enums from baldr and the directories but its really very trivial to make these test function generic and then call them with permutations. makes it much much easier to maintain, especially when we are talking about a test which is 3k lines

@kevinkreiser done

PR #4275, opened 2023-08-27
Return LineString geometries on expansion

### LINKED ISSUE #4266: Expansion output to LineStrings
In the interest of response size/performance, we decided to output `MultiLineString`s for the expansion endpoint. That's fine for pure network visualization, however very inconvenient for more complex visualization or further analysis/debugging.

Should we allow a `verbose` output to create `LineString`s instead? Conveniently that can be turned off in the config. When doing that I think I'd like to add a GeoJSON utility, it's not the only place where we construct GeoJSON and it's pretty verbose.

nilsnolde:
fixes #4266 

- makes /expansion return LineStrings
- adds the field `pred_edge_id` to the response to figure expansion

The config limit is the same as before: it's derived from the expansion action. We could adapt that logic, but I'd do that in a follow-up PR if it's wanted.

Is based on #4263 , so let's merge that first.


src/proto_conversions.cc
```diff
@@ -373,7 +373,9 @@ bool Options_ExpansionProperties_Enum_Parse(const std::string& prop,
               {"durations", Options_ExpansionProperties_durations},
               {"distances", Options_ExpansionProperties_distances},
               {"statuses", Options_ExpansionProperties_statuses},
-              {"edge_ids", Options::ExpansionProperties::Options_ExpansionProperties_edge_ids}};
+              {"edge_ids", Options::ExpansionProperties::Options_ExpansionProperties_edge_ids},
```
nilsnolde, line 376: we might want to rename that? previously it was the same field name in the response, but it's the singular version i.e. `edge_id` instead of `edge_ids`. would be nicer to have them consistent IMO.

src/thor/bidirectional_astar.cc
```diff
@@ -357,8 +357,13 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
...
   if (expansion_callback_) {
-    expansion_callback_(graphreader, FORWARD ? meta.edge_id : opp_edge_id, "bidirectional_astar", "r",
-                        pred.cost().secs, pred.path_distance(), pred.cost().cost);
+    const auto& prev_pred =
+        pred.predecessor() == kInvalidLabel
+            ? GraphId{}
+            : (FORWARD ? edgelabels_forward_ : edgelabels_reverse_)[pred.predecessor()].edgeid();
+    expansion_callback_(graphreader, FORWARD ? meta.edge_id : opp_edge_id, prev_pred,
+                        "bidirectional_astar", "r", newcost.secs,
+                        pred.path_distance() + meta.edge->length(), newcost.cost);
```
nilsnolde, line 366: I also changed all callbacks to use the end of the current edge. Most were doing that already, but not the "reached" ones. That's really confusing when looking at the results.

src/proto_conversions.cc
nilsnolde: ok, let me get that in real quick. also the shorter version.

nilsnolde:
done @kevinkreiser , also renamed `status` to `edge_status` due to a proto enum clash with `actions::status`.

nilsnolde:
Can we merge here? I need to use matrix expansion on many PRs right now and it's painful to merge this branch around.

docs/docs/api/expansion/api-reference.md
```diff
@@ -16,34 +16,134 @@ Since this service wraps other services, the request format mostly follows the o
...
-| `costs`       | Returns the accumulated cost for each edge in order of graph traversal. | 
-| `edge_ids`   | Returns the internal edge IDs for each edge in order of graph traversal. Mostly interesting for debugging. | 
-| `statuses`   | Returns the edge states for each edge in order of graph traversal. Mostly interesting for debugging. Can be one of "r" (reached), "s" (settled), "c" (connected). |
+| `distance`   | Returns the accumulated distance in meters for each edge in order of graph traversal. | 
+| `duration`   | Returns the accumulated duration in seconds for each edge in order of graph traversal. | 
+| `cost`       | Returns the accumulated cost for each edge in order of graph traversal. | 
+| `edge_id`   | Returns the internal edge IDs for each edge in order of graph traversal. Mostly interesting for debugging. | 
+| `pred_edge_id` |  Returns the internal edge IDs of the predecessor for each edge in order of graph traversal. Mostly interesting for debugging. |
+| `edge_status`   | Returns the edge states for each edge in order of graph traversal. Mostly interesting for debugging. Can be one of "r" (reached), "s" (settled), "c" (connected). |
```
kevinkreiser, line 30: one day we'll have to make this no longer beta and lock down the api :smile:

kevinkreiser:
this looks fine to me!

kevinkreiser (approved):

docs/docs/api/expansion/api-reference.md
nilsnolde: I’d be fine with lifting the BETA status actually. I’ve used it now quite a bit. Apart from that one bug where random edges are included for isochrones, which I’d really like to fix first.

Issue #4136, opened 2023-05-30
Landmark Routing 4: Associate Landmarks and Add to Tiles


kevinkreiser:
- update the enhancer or maybe add a post process or so (to use locate for association) to use the intermediate storage to find landmarks in the tile its enhancing
- add some logic to the enhancer to decide if a landmark is worthy of associating
- and then some logic to determine to where the landmark should be associated
- call into the builder datastructure that allows making the association

PR #4278, opened 2023-08-29
Add Landmarks to Tiles

vesperlou:
# Issue

Enable identifying, evaluating, associating, and integrating landmarks from intermediate storage (landmark database) into graph tiles. Landmarks are associated with edges.

## Tasks / workflows
- [x]  find landmarks in a graph tile and retrieve them from the landmark database
- [x]  associate each of the landmarks with nearby edges (need some filtering on the edges)
- [x]  update the graph tile to store the associations of landmark and edge

## Reference
[Landmark Routing 4: Associate Landmarks and Add to Tiles#4136](https://github.com/valhalla/valhalla/issues/4136)


src/mjolnir/graphtilebuilder.cc
```diff
@@ -1319,14 +1319,16 @@ void GraphTileBuilder::AddLandmark(const GraphId& edge_id, const Landmark& landm
...
-      auto* eib = e.second;
-      uint32_t newkey = e.first + shift;
-      edgeinfo_offset_map_.erase(e.first);
-      edgeinfo_offset_map_.insert(std::make_pair(newkey, eib));
+      new_edgeinfo_offset_map_.insert(std::make_pair(e.first + shift, e.second));
+    } else {
+      new_edgeinfo_offset_map_.insert(e);
     }
   }
+  edgeinfo_offset_map_ = new_edgeinfo_offset_map_;
```
vesperlou, line 1331: @kevinkreiser @nilsnolde here is the edited part to update edgeinfo_offset_map. the previous code manipulated the map while reading it, which would cause segmentation fault.
I guess here is still optimization to do.

kevinkreiser: this doesnt look right to me. anything that was in there before is completely gone and will only contain the stuff you just added right?

looking at the old code i cannot understand why there would be a segfault, can you say more about what you were seeing with the old code?

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,219 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
                           callback);
   }
 
+  LOG_INFO("Successfully build landmark databse from PBF");
```
kevinkreiser, line 321:
```suggestion
  LOG_INFO("Successfully built landmark database from PBF");
```

src/mjolnir/graphtilebuilder.cc
kevinkreiser: ok after spending 5 minutes looking at it while not doing something else i can see the issue. we were iterating over it while updating it which will be a problem depending on what happens with reallocations. i would have thought it wouldnt do reallocations because we arent actually growing the size but it must be since you hit a segfault.

anyway the change here will work. i think in the long term we can optimize this another way.

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+  // join all the threads and collect the sequence file names
+  for (auto& thread : threads) {
+    thread->join();
+  }
+
+  std::vector<std::string> seq_names{};
+  for (std::promise<std::string>& s : sequence_file_names) {
+    seq_names.push_back(s.get_future().get());
+  }
+
```
kevinkreiser, line 492:
```suggestion

  LOG_INFO("Sorting landmark edge pairs by tile...");

```

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+  const std::string dbname = pt.get_child("mjolnir").get<std::string>("landmarks", "");
+
+  LandmarkDatabase db(dbname, true);
+  GraphReader reader(pt.get_child("mjolnir"));
+  // create the sequence file
+  std::string file_name = "landmark_dump_" + std::to_string(thread_number);
+  midgard::sequence<std::pair<GraphId, uint64_t>> seq_file(file_name, true);
+
+  for (size_t i = 0; i < tileset.size(); ++i) {
+    // assign tiles for threads
```
kevinkreiser, line 342: i find this comment a bit lacking for someone who doesnt already know what is going on
```suggestion
    // every i'th thread works on every i'th tile
```

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+      std::vector<Landmark> landmarks =
+          db.get_landmarks_by_bbox(bbox.minx(), bbox.miny(), bbox.maxx(), bbox.maxy());
+
+      // find and collect all nearby path locations for the landmarks
+      for (const auto& landmark : landmarks) {
+        baldr::Location landmark_location(midgard::PointLL{landmark.lng, landmark.lat},
+                                          baldr::Location::StopType::BREAK, 0, 0,
+                                          25); // a 25m radius, which allows us to only keep the close
+                                               // ones in the tight cities
+        landmark_location.search_cutoff_ = 75.; // a 75m search cutoff should allow us to get gas
```
kevinkreiser, line 356: these numbers should be constants at the top of the file

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+                                                // stations that are off the road a bit (for parking)
+
+        // call loki::Search to get nearby edges to each landmark
+        std::unordered_map<valhalla::baldr::Location, PathLocation> result =
+            loki::Search({landmark_location}, reader, sif::CreateNoCost({}));
+        if (result.size() > 1) {
+          throw std::logic_error("Error occurred in calling loki::Search. Result size is " +
+                                 std::to_string(result.size()) + ", but should be one or zero");
+        }
```
kevinkreiser, line 365: i dont think its this code's job to check for a logic error here. did this actually happen to you? unit tests should catch something so terrible as this
```suggestion
```

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+        if (result.size() == 0) {
+          continue;
+        }
```
kevinkreiser, line 368: can this actually happen? i thought it returns always the same number of locations passed in but if it doesnt find anything then the path edges are empty

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+        // for each edge insert edgeid - landmark_pkey pair into the sequence file
+        // TODO: maybe do some filtering and only keep some of the edges it finds?
+        for (const auto& edge : edges) {
+          seq_file.push_back(std::make_pair(edge.id, landmark.id));
+        }
+      }
+    }
+  }
+  // sort the sequence by tiles
+  seq_file.sort(sort_seq_file);
```
kevinkreiser, line 380: this is a waste of time as we sort the whole thing later
```suggestion
```

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+    if ((*it).first.Tile_Base() != last_tile) {
+      last_tile = (*it).first.Tile_Base();
+      tile_count++;
+    }
+    if (tile_count % total_threads != thread_number) {
+      continue;
+    }
+
+    if (!tile_builder_ptr ||
+        tile_builder_ptr->header_builder().graphid().Tile_Base() != (*it).first.Tile_Base()) {
```
kevinkreiser, line 416: we need some comments in here for sure. we want to lessen the mental burden of whoever reads this next

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+  }
+  // sort the sequence by tiles
+  seq_file.sort(sort_seq_file);
+
+  seq_file_name.set_value(file_name);
+}
+
+// Update tiles to associate landmarks, return some stats about the numbers of updated tiles, edges,
+// and landmarks
+void UpdateTiles(midgard::sequence<std::pair<GraphId, uint64_t>>& seq_file,
```
kevinkreiser, line 387: we should mention that seq_file is by ref but should not be modified by these threads. the other option is to pass teh file name and have each thread open the same file

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+        tile_builder_ptr->StoreTileData();
+        updated_tiles++;
+      }
+      // reset the tile builder ptr to the current tile
+      tile_builder_ptr.reset(new GraphTileBuilder(tile_dir, (*it).first.Tile_Base(), true));
+    }
+
+    // retrieve the landmark to be added
+    const std::vector<Landmark> landmark =
+        db.get_landmarks_by_ids({static_cast<int64_t>((*it).second)});
```
kevinkreiser, line 428: in the future we can do batches of ids, though it will complicate the code it will likely speed up the processing

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+      throw std::logic_error("Incorrect result size " + std::to_string(landmark.size()) +
+                             " of retrieved landmarks, which should be 1");
+    }
+    // add the landmark to the tile
+    GraphId edge_id = (*it).first;
+    tile_builder_ptr->AddLandmark(edge_id, landmark[0]);
+
+    // update the counting numbers
+    updated_landmarks++;
+    if (last_edge != (*it).first) {
```
kevinkreiser, line 439:
```suggestion
    // a single edge can have multiple landmarks
    if (last_edge != (*it).first) {
```

kevinkreiser (changes requested):
this is looking very close, just some nits and the clang tidy

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+
+  const size_t num_threads =
+      pt.get<size_t>("mjolnir.concurrency", std::thread::hardware_concurrency());
+  const std::string db_name = pt.get_child("mjolnir").get<std::string>("landmarks_db", "");
+
+  // get tile access
+  baldr::GraphReader reader(pt.get_child("mjolnir"));
+
+  // get all tile ids and sort the tiles in descending order by size to balance the threads
+  auto tileset = reader.GetTileSet(1);
```
kevinkreiser, line 466: it is possible in a global tileset that we have coverage only at level 2 for some places and we'd still like to get landmarks there. we'll probably need to fix this...

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+  }
+  seq_file.close();
+
+  midgard::sequence<std::pair<GraphId, uint64_t>> merged_sequence_file(merged_seq_file, false);
+  merged_sequence_file.sort(sort_seq_file);
+
+  LOG_INFO("Updating tiles...");
+
+  // open up a new pool thread to update tiles
+  std::vector<std::shared_ptr<std::thread>> threads_new(num_threads);
```
kevinkreiser, line 510:
```suggestion
```

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+  std::vector<std::promise<std::tuple<size_t, size_t, size_t>>> stats_info(
+      num_threads); // tiles, edges, landmarks
+
+  const std::string tile_dir = reader.tile_dir();
+  for (size_t i = 0; i < num_threads; ++i) {
+    // assume size doesn't affect performance a lot
+    threads_new[i].reset(new std::thread(UpdateTiles, std::ref(merged_sequence_file), tile_dir,
```
kevinkreiser, line 517:
```suggestion
    threads[i].reset(new std::thread(UpdateTiles, std::ref(merged_sequence_file), tile_dir,
```

src/mjolnir/landmarks.cc
```diff
@@ -296,6 +318,226 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
...
+    stats.push_back(s.get_future().get());
+  }
+
+  // log how many were written to how many tiles INFO
+  size_t tiles = 0, edges = 0, landmarks = 0;
+  for (const auto& s : stats) {
+    tiles += std::get<0>(s);
+    edges += std::get<1>(s);
+    landmarks += std::get<2>(s);
+  }
```
kevinkreiser, line 536: can we just m erge this and have one loop, we can add the stats up right there dont need to store them in a vector at all in the end right?

src/mjolnir/landmarks.cc
kevinkreiser: seems like not

src/mjolnir/landmarks.cc
nilsnolde: ah finally!!! seems github preserves comments in file view once they were addressed. was the most annoying UX oversight IMO..

nilsnolde (approved):
looks good now IMO, thanks @vesperlou

kevinkreiser (approved):

Issue #4138, opened 2023-05-30
Landmark Routing: Narrative Generation Updates


kevinkreiser:

kevinkreiser:
more detailed notes with links. for the scoping for this we want to get to just the first possible place we can start seeing landmarks in narrative. we'll scope to just simple turns (not even sharp or bear or keep) and we can expand to the other turn types once we have one of them working

- we'll want to start by expanding what we have here: https://github.com/valhalla/valhalla/blob/97cad0609c2085b98196286b6cb02b0e09128749/locales/en-US.json#L1948
- we'll add versions of the first 4 phrases that have 2 additional replacements, see below for example (might want to add the article "the" after the word "at", we can play with it to see how it sounds)
- the LANDMARK_TYPE replacement will need a translation in each language so we can do english and do the rest as contributors notice, see below again for example
- when you are happy with that dont forget to do the "verbal" variant of the phrases (they are in another phrase block)
- now go and do the "sharp" phrases and verbal phrases and the "bear" (which is slight) phrases and verbal phrases (should be simple copy paste)
- this should cause the locale tests to fail because all the languages have to have all the phrases (though maybe we fallback to english when missing). so we can just make the changes to the english locale and we can run https://github.com/valhalla/valhalla/blob/master/locales/merge-en.sh to get the new phrases into the other language files
- next we need to update the narrative building, to do that we need to find the function where we decide which phrase to use and set the landmark information
- somewhere around here: https://github.com/valhalla/valhalla/blob/97cad0609c2085b98196286b6cb02b0e09128749/src/odin/narrativebuilder.cc#L1165 we can say if the phrase id is less than 4 (its one of the ones we allow landmarks to be used in) AND we have landmarks to use, then
- we increment the phrase id by 6 to get to the landmark ones
- we set the landmark data on an empty landmark string name here: https://github.com/valhalla/valhalla/blob/97cad0609c2085b98196286b6cb02b0e09128749/src/odin/narrativebuilder.cc#L1145
[...]

PR #4301, opened 2023-09-19
Update narrative generation with landmarks

vesperlou:
# Issue

Update narrative generation process to support the use of landmarks in text guidance.

## Tasklist

 - [ ] Update the maneuver types in locales to support landmark replacements. (update the en-us one, and wait for translations)
 - [ ] Update the narrative generation to pick phrases that support the use of landmarks.
 - [ ] Update tests and the changelog accordingly.

## Requirements / Relations
[Landmark Routing 6: Narrative Generation Updates #4138](https://github.com/valhalla/valhalla/issues/4138)

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -39,8 +39,33 @@
         "2": "Bear <RELATIVE_DIRECTION> onto <BEGIN_STREET_NAMES>. Continue on <STREET_NAMES>.",
         "3": "Bear <RELATIVE_DIRECTION> to stay on <STREET_NAMES>.",
         "4": "Bear <RELATIVE_DIRECTION> at <JUNCTION_NAME>.",
-        "5": "Bear <RELATIVE_DIRECTION> toward <TOWARD_SIGN>."
+        "5": "Bear <RELATIVE_DIRECTION> toward <TOWARD_SIGN>.",
+        "6": "Bear <RELATIVE_DIRECTION> at the <LANDMARK_TYPE> <LANDMARK_NAME>.",
+        "7": "Bear <RELATIVE_DIRECTION> onto <STREET_NAMES> at the <LANDMARK_TYPE> <LANDMARK_NAME>.",
+        "8": "Bear <RELATIVE_DIRECTION> onto <BEGIN_STREET_NAMES> at the <LANDMARK_TYPE> <LANDMARK_NAME>. Continue on <STREET_NAMES>.",
+        "9": "Bear <RELATIVE_DIRECTION> at the <LANDMARK_TYPE> <LANDMARK_NAME> to stay on <STREET_NAMES>."
       },
+      "landmark_types": [
+        "unused",
+        "gas station",
+        "post office",
+        "police station",
+        "fire station",
+        "car wash",
+        "restaurant",
+        "fast food",
+        "cafe",
+        "bank",
+        "pharmacy",
+        "kindergarden",
+        "bar",
+        "hospital",
+        "pub",
+        "clinic",
+        "theatre",
+        "cinema",
+        "casino"
+      ],
       "empty_street_name_labels": [
         "the walkway",
         "the cycleway",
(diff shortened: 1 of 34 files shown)


nilsnolde:
Reminder to ourselves: since we're updating (possibly) stale translation files here, we should first pull whatever transifex has currently, merge to master, then merge master in this PR, to avoid tons of conflicts.

locales/en-US.json
```diff
@@ -2036,6 +2188,7 @@
         "6": "Make a <RELATIVE_DIRECTION> U-turn at <JUNCTION_NAME>.",
         "7": "Make a <RELATIVE_DIRECTION> U-turn toward <TOWARD_SIGN>."
       },
+      "landmark_types": [],
```
nilsnolde, line 2191: for u-turn we also want landmark narratives

src/odin/narrativebuilder.cc
```diff
@@ -4531,6 +4581,52 @@ NarrativeBuilder::FormRelativeThreeDirection(DirectionsLeg_Maneuver_Type type,
   }
 }
 
+std::string NarrativeBuilder::FormLandmarkType(RouteLandmark_Type type,
```
nilsnolde, line 4584: instead of this function, we can cast the `RouteLandmark_Type` to `uint8_t` and index into the phrases directly

src/odin/narrativebuilder.cc
```diff
@@ -36,6 +36,38 @@ constexpr float kVerbalAlertMergePriorManeuverMinimumLength = kVerbalPostMinimum
...
+constexpr double kLandmarkOppositeDirectionCompensation = 100.;
+
+// Select the most suitable landmark in a maneuver to be used in narrative generation.
+// TODO: the current logic is to select the nearest landmark to the maneuver and it should be
+// preferably on the correct side. However, if for say the maneuver is to turn right, the nearest
+// right handed landmark is far away from the maneuver but the nearest left handed landmark is very
+// close, than we pick the nearest left handed one.
+valhalla::RouteLandmark select_landmark(const std::vector<valhalla::RouteLandmark>& landmarks,
+                                        const std::string& relative_direction) {
+  bool direction = relative_direction == "right" ? 1 : 0;
```
nilsnolde, line 52:
```suggestion
  bool direction = relative_direction == "right";
```

src/odin/narrativebuilder.cc
```diff
@@ -36,6 +36,38 @@ constexpr float kVerbalAlertMergePriorManeuverMinimumLength = kVerbalPostMinimum
...
+    if (idx == -1) {
+      idx = i;
+      min_dist = dist;
+    } else if (dist < min_dist) {
+      idx = i;
+      min_dist = dist;
+    }
+  }
+  return landmarks[idx];
+}
```
nilsnolde, line 70:
```suggestion
  size_t idx = 0;
  double min_dist = std::numeric_limits<double>::max();
  for (auto i = 0; i < landmarks.size(); i++) {
    // adjust landmark distance with compensation based on their direction
    auto dist = landmarks[i].right() == direction
                    ? landmarks[i].distance()
                    : (landmarks[i].distance() + kLandmarkOppositeDirectionCompensation);
    if (dist < min_dist) {
      idx = i;
      min_dist = dist;
    }
  }
  return landmarks[idx];
}
```

src/odin/narrativebuilder.cc
```diff
@@ -1163,6 +1196,20 @@ std::string NarrativeBuilder::FormTurnInstruction(Maneuver& maneuver,
...
+    landmark =
+        select_landmark(maneuver.landmarks(),
+                        FormRelativeTwoDirection(maneuver.type(), subset->relative_directions));
+
+    // Make sure the landmark type is legal
+    if (static_cast<int>(landmark.type()) > 0 &&
+        static_cast<int>(landmark.type()) < subset->landmark_types.size()) {
+      // increment the phrase id by 6 to get the corresponding landmark phrase
+      phrase_id += 6;
+    }
```
nilsnolde, line 1210: IMO we don't need a check like this, bcs we're in full control of the landmark types and we do make sure in the data building pipeline that nothing enters that's outside of the type bounds. If we want to check this at this point, then let's throw an exception, bcs that'd be a very serious flaw.

nilsnolde:
One thing we noticed is that the overall graph size is shrinking after adding landmarks.

src/mjolnir/valhalla_add_landmarks.cc
```diff
@@ -51,7 +49,7 @@ int main(int argc, char** argv) {
     valhalla::midgard::logging::Configure(logging_config);
   }
 
-  if (!valhalla::mjolnir::AddLandmarks(pt.get_child("mjolnir"))) {
+  if (!valhalla::mjolnir::AddLandmarks(pt)) {
```
vesperlou, line 52:
```suggestion
  if (!valhalla::mjolnir::AddLandmarks(pt.get_child("mjolnir"))) {
```
I've edited AddLandmarks to handle the input as child mjolnir, maybe we want to change this back to get child mjolnir. it works for me :)

nilsnolde: Yeah I must not've seen that for some reason. Weird though that I could push at all if I didn't pull your changes before.. I'll revert this!

vesperlou: I guess you can push because there is no conflict caused by this commit

nilsnolde:
I started looking at this again, but didn't get very far yet. When I tried a route request, it wouldn't work, because we mess smth up when adding landmarks and make the tiles invalid somehow (as said before, tile size _shrinks_ when adding landmarks!). It's not much code, but still annoying to debug.

kevinkreiser:
yeah i was hoping we could patch it up to the point that it could somewhat work sometimes and be a thing we opt into. maybe in the coming weeks we can pair on it for an hour at a time a couple times a week

nilsnolde:
my shortcut PR to remove names/speed is just tripping over the landmarks stuff: @kevinkreiser reminder for us to check how shortcuts are behaving with landmarks.. might be very well the key..

kevinkreiser:
interesting hypothesis. imho we shouldn't put them on shortcuts anyway its a duplication with purpose

nilsnolde:
Yeah we don’t want that. But seems like we are, and while not being aware of it, maybe we’re forgetting about the base edges in the process and they get lost..

kevinkreiser:
also we use loki search iirc which doesn't return shortcuts

nilsnolde:
Right, I think that was a fluke.. the test was going over the tileset and exactly, didn’t expect to see shortcuts, so all good there

PR #4302, opened 2023-09-20
update grade based speed factor by using a modified Tobler's function

xlqian:
# Issue

By bumping Valhalla from 3.1 to 3.4, we realized that the average walk speed has dropped considerably, which is in connection with [this issue](https://github.com/valhalla/valhalla/issues/3969). Although a decrease in speed when hiking uphills holds true, a considerable drop in speed when walking on a slight downhill contradicts common sense, especially for city walkers walking on paved paths. 

We suggest in this PR to use [Tobler's function](https://en.wikipedia.org/wiki/Tobler%27s_hiking_function) to compensate for the loss when the angle is between **0% and -10 %**. 

We modified Tobler's function **in order not to** modify the previous behaviour(according to DIN) studied and validated by @hungerburg,  and only focus on **0% and -10 %**. 

image

(https://gist.github.com/xlqian/0b25c8db6f45fb2c8bf68494e1ea54f1)

## Tasklist

## Requirements / Relations

 https://github.com/valhalla/valhalla/pull/3982


xlqian:
> Does weighted_grade work well for you in most areas as an approximation of grade?

I don't know how to define `work well` in this context, but for us, the paths we have got with `elevation data` are consistent with what we would do in real life, so far so good.

> In those areas I suspect the grade approximation may be too high

hmmm, that is an interesting angle and worth being looked into... in any case, it is baffling for us to see the speed factor drop under 1 even when we're on a mildly tilted and well-paved path...

hungerburg:
You certainly are right, a mildly descending slope will not slow people down, rather do the opposite. Perhaps regardless of roughness of terrain, when compared to flat speed at same roughness. After all, walking is modelled as a kind of controlled falling, that is why it is so efficient. Only when terrain gets steeper than some limit, this fires backwards. From reading up on the literature, the limit is set somewhere between -6 and -12%.

The so-called DIN function looks to me a variation of https://en.wikipedia.org/wiki/Naismith%27s_rule - a rule of thumb to estimate ToA without a computer with pen and paper only. Unlike estimates from Naismith's rule the ones given by the DIN rule are easy to beat though. The Swiss have a more elaborate function, a polynomial with 15 degrees of freedom or so. They are nerds, and more sportive too!

When I pushed for the change, I feared that Valhalla results would get ridiculous, but they turned out rather fine. Maybe that was mostly due to the fact, that before the change, downhill speed did increase with no limit, so instead of walking you would be running? So in fact, what the change mostly did was smooth out noise in the elevation data?

Said all that, the turning point of where slowing starts rather be to the left of the zero degree mark. That for sure.

nilsnolde:
Any objections merging this @xlqian @hungerburg et al?

xlqian:
I'm ok with it

hungerburg:
The changes to kGradeBasedSpeedFactor look sensible.

nilsnolde (approved):
Ok cool, then let's do it I'd say:) I'll let someone else merge, @kevinkreiser @dnesbitt61 @gknisely might have an opinion about it..

nilsnolde (approved):

kevinkreiser (approved):
the reasoning makes sense and the values look fine to me

PR #4318, opened 2023-10-03
Add use_tunnels option

ADMUKHE:
# Add Tunnel  factor to Autocost

Presently valhalla cannot avoid roads if tunnel is present. These changes will add tunnel factor in Edge Cost Calculations. 

The program was tested solely for our own use cases, which might differ otherwise. I acknowledge that these changes have the potential to impact other use cases, and it is essential to exercise caution and thorough testing when merging them to the master. Your expert opinion is requested to help us review, test, and adapt these modifications as needed. I am also looking forward to other implementation ideas and discussions about how we can tackle this issue. 

## Tasklist

 - [ ] Add tests

Adrika Mukherjee adrika.mukherjee@mercedes-benz.com, Mercedes-Benz Tech Innovation GmbH

### DIFF
--- a/src/sif/autocost.cc
+++ b/src/sif/autocost.cc
@@ -32,6 +32,7 @@ constexpr float kDefaultServicePenalty = 75.0f; // Seconds
 // Other options
 constexpr float kDefaultUseHighways = 0.5f; // Default preference of using a motorway or trunk 0-1
 constexpr float kDefaultUseTolls = 0.5f;    // Default preference of using toll roads 0-1
+constexpr float kDefaultUseTunnels = 0.5f;  // Default preference of using tunnel roads 0-1
 constexpr float kDefaultUseTracks = 0.f;    // Default preference of using tracks 0-1
 constexpr float kDefaultUseDistance = 0.f;  // Default preference of using distance vs time 0-1
 constexpr uint32_t kDefaultRestrictionProbability = 100; // Default percentage of allowing probable
@@ -76,6 +77,7 @@ constexpr float kDefaultAutoWidth = 1.9f;  // Meters (74.8031 inches)
 constexpr ranged_default_t<float> kAlleyFactorRange{kMinFactor, kDefaultAlleyFactor, kMaxFactor};
 constexpr ranged_default_t<float> kUseHighwaysRange{0, kDefaultUseHighways, 1.0f};
 constexpr ranged_default_t<float> kUseTollsRange{0, kDefaultUseTolls, 1.0f};
+constexpr ranged_default_t<float> kUseTunnelsRange{0, kDefaultUseTunnels, 1.0f};
 constexpr ranged_default_t<float> kUseDistanceRange{0, kDefaultUseDistance, 1.0f};
 constexpr ranged_default_t<float> kAutoHeightRange{0, kDefaultAutoHeight, 10.0f};
 constexpr ranged_default_t<float> kAutoWidthRange{0, kDefaultAutoWidth, 10.0f};
@@ -330,6 +332,7 @@ public:
   float highway_factor_;      // Factor applied when road is a motorway or trunk
   float alley_factor_;        // Avoid alleys factor.
   float toll_factor_;         // Factor applied when road has a toll
+  float tunnel_factor_;       // Factor applied when road has a toll
   float surface_factor_;      // How much the surface factors are applied.
   float distance_factor_;     // How much distance factors in overall favorability
   float inv_distance_factor_; // How much time factors in overall favorability
@@ -398,6 +401,14 @@ AutoCost::AutoCost(const Costing& costing, uint32_t access_mask)
   float use_tolls = costing_options.use_tolls();
   toll_factor_ = use_tolls < 0.5f ? (4.0f - 8 * use_tolls) : // ranges from 4 to 0
                      (0.5f - use_tolls) * 0.03f;             // ranges from 0 to -0.15
+
+  // Preference to use tunnel roads. Sets a tunnel
+  // factor. A tunnel factor of 0 would indicate no adjustment to weighting for tunnel roads.
+  // use_tunnel = 1 would reduce weighting slightly (a negative delta) while
+  // use_tunnel = 0 would penalize (positive delta to weighting factor).
+  float use_tunnels = costing_options.use_tunnels();
+  tunnel_factor_ = use_tunnels < 0.5f ? (4.0f - 8 * use_tunnels) : // ranges from 4 to 0
+                       (0.5f - use_tunnels) * 0.03f;               // ranges from 0 to -0.15
 
   include_hot_ = costing_options.include_hot();
   include_hov2_ = costing_options.include_hov2();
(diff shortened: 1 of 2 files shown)


ADMUKHE:
The error "Job was rejected because resource class large, image xcode:13.4.1 is not a valid resource class" in the first check seems weird. I would really appreciate some help here.

johannes-no:
Since this is a soft avoidance of tunnels, there should be a boolean set like has_tunnel in the route summary indicating whether the route uses tunnels.

ADMUKHE:
@nilsnolde Can you pease have a quick look into this PR and leave your comments?

test/parse_request.cc
```diff
@@ -333,7 +334,7 @@ std::string get_request_str(const std::string& parent_key,
   return R"({")" + parent_key + R"(":{)" + get_kv_str(key, expected_value) + R"(}})";
 }
 
-std::string get_kv_str(const std::string& key, const std::vector<std::string>& values) {
+std::string get_kv_str(const std::string& key) {
```
nilsnolde, line 337: why this change?

test/parse_request.cc
```diff
@@ -457,8 +458,7 @@ void test_filter_action_parsing(const valhalla::FilterAction expected_value,
   validate(key, expected_value, true, request.options().filter_action());
 }
 
-void test_filter_attributes_parsing(const std::vector<std::string>& expected_values,
-                                    const Options::Action action = Options::trace_attributes) {
+void test_filter_attributes_parsing(const Options::Action action = Options::trace_attributes) {
```
nilsnolde, line 461: and why this change?

test/parse_request.cc
```diff
@@ -1537,7 +1554,6 @@ void test_bicycle_parsing(const Costing::Type costing_type,
 
 void test_filter_stop_parsing(const Costing::Type costing_type,
                               const valhalla::FilterAction filter_action,
-                              const std::vector<std::string>& filter_ids,
```
nilsnolde, line 1540: and this?

test/parse_request.cc
```diff
@@ -1555,7 +1571,6 @@ void test_filter_stop_parsing(const Costing::Type costing_type,
 
 void test_filter_route_parsing(const Costing::Type costing_type,
                                const valhalla::FilterAction filter_action,
-                               const std::vector<std::string>& filter_ids,
```
nilsnolde, line 1558: and this?

test/parse_request.cc
```diff
@@ -1573,7 +1588,6 @@ void test_filter_route_parsing(const Costing::Type costing_type,
 
 void test_filter_operator_parsing(const Costing::Type costing_type,
                                   const valhalla::FilterAction filter_action,
-                                  const std::vector<std::string>& filter_ids,
```
nilsnolde, line 1576: and this?

valhalla/proto_conversions.h
```diff
@@ -566,4 +566,4 @@ inline void from_ll(valhalla::Location* l, const midgard::PointLL& p) {
   l->mutable_ll()->set_lng(p.lng());
 }
 
-} // namespace valhalla
+} // namespace valhalla
```
nilsnolde, line 569: and this?

src/sif/autocost.cc
```diff
@@ -399,6 +402,14 @@ AutoCost::AutoCost(const Costing& costing, uint32_t access_mask)
   toll_factor_ = use_tolls < 0.5f ? (4.0f - 8 * use_tolls) : // ranges from 4 to 0
                      (0.5f - use_tolls) * 0.03f;             // ranges from 0 to -0.15
 
+  // Preference to use tunnel roads. Sets a tunnel
+  // factor. A tunnel factor of 0 would indicate no adjustment to weighting for tunnel roads.
```
nilsnolde, line 406:
```suggestion
  // factor. A tunnel factor of 0.5 would indicate no adjustment to weighting for tunnel roads.
```

nilsnolde (changes requested):
IMO it only makes sense to expose a `use_tunnel` option. PR looks generally fine, just LOTS of unrelated adverse changes in the test.

Also 
- I'd personally like to see a gurka test as well for the costing option, parameterized for all modes that use this. Ideally we'd not create an additional test file, but rather consolidate the other test_use_* into one single file for the sake of compilation speed (i.e. `test_use_lit.cc`, `test_use_living_streets.cc`, `test_use_tracks.cc`)
- an additional edge and check in gurka/test_use_feature.cc
- we should do this for all motorized vehicles I think? for ped (and bike?) we had a discussion before around safety issues, I think that'd need to be differently handled than the rest

nilsnolde:
I'd recommend you go through the Github view of the diff in https://github.com/valhalla/valhalla/pull/4318/files after you commit (ideally in your own IDE before you even push to not needlessly run CI with stuff that'll break for sure).

nilsnolde:
Bonus points for adding comments on your own PR to explain non-obvious changes, which saves tons of time for the reviewers. (e.g. here I personally would have noted that the costing change is a 100% copy of `use_tolls`)

ADMUKHE:
@nilsnolde for ped and bike then would you suggest to have a seperate PR?

nilsnolde:
If you’re interested in the functionality, I’d prefer a separate PR bcs the discussion would be very different and we’d first open an issue for that. But don’t feel obliged to take care of that too. It’s fine to not expose for those modes for the time being.

kevinkreiser:
there is a slight technical detail here that i thought someone mentioned but i dont see it mentioned here: 
https://github.com/valhalla/valhalla/blob/3806832a11fe250e08f6f21c9f55da0e052bfc80/src/mjolnir/shortcutbuilder.cc#L69C1-L71C28

we create shortcuts overtop of bridges and tunnels which means the costing will not avoid a tunnel if its part of a shortcut that the algorithm is using. given that tunnels arent as prevalant as bridges we could probably get away with adding tunnel as an attribute that has to match during shortcut building. bridges could be a pretty big problem though..

kevinkreiser:
i also agree with @nilsnolde that we should consider just adding `use_tunnels` to all motorized vehicles. we could move the parsing and factor to the base costing class and just use it from the children that make sense.

the rest of the pr looks great!

nilsnolde:
Just as a note: the only thing missing is a gurka test to verify the behavior of the costing option. Then it’s good to merge.

ADMUKHE:
@nilsnolde is it not covered in the test_use_feature test? can you give any reference to any other gurka test specific to costing option in case you are talking about something else?

nilsnolde:
No, that is only testing that it's flagged in the response, it's not testing that the _costing_ is actually doing anything with it. I already described it: https://github.com/valhalla/valhalla/pull/4318#pullrequestreview-1686407319. If you have specific questions, feel free to ask.

nilsnolde:
Oh, and an important note: if we use tunnels to break up shortcuts, we'll need to do a thorough perf test and see if it's fine. Ideally that's done by the committer, but it's poorly documented atm. I always wanted to document that a bit better. We might want to add some Switzerland routes for this:)

PR #4326, opened 2023-10-12
dont build shortcuts across access_restrictions

### LINKED ISSUE #4025: Question: route api and radius
Hello,

when I have a given set of coordinates and have to calculate the route, I will sometimes get the error msg:
{"error_code":442,"error":"No path could be found for input","status_code":400,"status":"Bad Request"}

The reason is, that some of the coordinates are in an area without street network or next street is too far away. For example inside a pedestrian zone. 

Here is an example
This request doesn't work with the default radius of 50.

`route?json={"locations":[{"lat":48.1375617980957,"lon":11.574857711792},{"lat":48.1582565307617,"lon":11.5032968521118}],"costing":"truck","costing_options":{"truck":{"top_speed":120,"height":3.45,"width":3.0,"length":8.0,"weight":40.0}},"units":"km","directions_type":"none","alternates":0}
`

To get it working, just change the radius for the first coordinate:
`route?json={"locations":[{"lat":48.1375617980957,"lon":11.574857711792,"radius":130},{"lat":48.1582565307617,"lon":11.5032968521118}],"costing":"truck","costing_options":{"truck":{"top_speed":120,"height":3.45,"width":3.0,"length":8.0,"weight":40.0}},"units":"km","directions_type":"none","alternates":0}`

Because I don't know, if the coordinate is inside a pedestrian area or not and I don't know what radius is needed to get a valid result, I could change the default value for radius high enough (e.g. 1000). But this will result in a much higher duration in cacluation time if you have about 20 coordinates.

A possible solution is to get a more detailed error description. If I would know which coordinate cannot be connected to the street network, I could higher the radius for this coordinate(s).

So my question:
Is there another parameter which I have to use, that a coordinate is always connected to the street network without raising computing time so much? 

Regards

Karsten

### LINKED ISSUE #4257: OSM maxwidth / maxweight are ignored on truck profile
While playing around with the truck profile I realized that on the infamous "Leverkusener Brücke" the maxwidth / maxweight restriction is ignored for the truck profile and Valhalla is happily routing over the bridge

260945888-8e4f80bf-15d8-4ed5-9763-af9a8b86240b

[Corresponding Route](https://valhalla.openstreetmap.de/directions?profile=truck&wps=6.949024200439454%2C51.059519698855276%2C6.939926147460938%2C51.01266308334306)
[Related Way](https://www.openstreetmap.org/way/133838845#map=14/51.0298/6.9628)

Interestingly @nilsnolde pointed out that a [much shorter route](https://valhalla.openstreetmap.de/directions?profile=truck&wps=6.965181775252557%2C51.03668505570838%2C6.956751282292306%2C51.031948745824195) does not work, while adding only a little more distance the bridge is routed again.

Potentially related to #3487 and #3377 although the width restrictions posed by the way referred to in #3377 seems to be respected which puzzles me even more...

image

[Corresponding Route](https://valhalla.openstreetmap.de/directions?profile=truck&wps=-4.260056018829347%2C55.849482539690996%2C-4.2652273178100595%2C55.8518652298043)
[Related way: Cook street](https://www.openstreetmap.org/way/972410214#map=17/55.85024/-4.26043)

If these restrictions are generally susceptible IMHO the truck profile is basically rendered not usable.

### LINKED ISSUE #4401: Hazmat ignored for truck routing
Hello, we've distilled an example of a truck route that does not adhere to hazmat restrictions.

On the route provided [https://valhalla.openstreetmap.de/directions?profile=truck&wps=15.272970199584963%2C49.57237520373628%2C14.960396289825441%2C49.75913196660811](url) there is a bridge with hazmat restrictions.

You can verify this using Overpass Turbo [https://overpass-turbo.eu/#](url) with the following script:

`[out:json][timeout:25];
// adjust the search radius (in meters) here
{{radius=1000}}
// gather results
nwr["hazmat"="no"](around:{{radius}},{{geocodeCoords:Vojslavický most}});
// print results
out geom;`

Even when we set up a route for a truck with hazardous materials, the route remains the same, as illustrated in the image below:

image

Is this somehow related to the fact that this route is possibly part of a shortcut, and thus the hazmat restriction is not being respected?

kevinkreiser:
fixes #4401 
fixes #4257 

this is a quick check to see if we can fix #4025 based on a hypothesis that we are building shortcuts across truck restrictions and then the bidirectional algorithm is getting screwed up with one search tree using the shortcut and one tree not


kevinkreiser:
seems to have a runtime error during CI so looks like my code is borked. i'll check it when i have time

src/mjolnir/shortcutbuilder.cc
```diff
@@ -81,21 +81,25 @@ bool EdgesMatch(const graph_tile_ptr& tile, const DirectedEdge* edge1, const Dir
...
+  if (edge1->access_restriction()) {
+    auto res1 = tile->GetAccessRestrictions(edge1 - tile->directededge(0), kVehicularAccess);
+    auto res2 = tile->GetAccessRestrictions(edge2 - tile->directededge(0), kVehicularAccess);
+    if (res1.size() != res2.size())
       return false;
+    for (size_t i = 0; i < res1.size(); ++i) {
+      res1[i].set_edgeindex(0);
+      res2[i].set_edgeindex(0);
+      if (res1[i].value() != res2[i].value())
+        return false;
```
nilsnolde, line 99: right, I kinda forgot you already had an active PR @kevinkreiser . since those restrictions are 90% truck specific (+ time-based restrictions), I wonder if we shouldn't make this somehow configurable? it seems like a worse impact than allowing tunnels to break up shortcuts..

kevinkreiser: on first instinct, i dont like the idea of making it configurable. if you can configure it to give you incorrect routes we just leave the door open to all kinds of comments about: hey my routes are broken. instead, id rather break shortcuts everywhere where its necessary and measure performance differences. if its a huge performance killer for sure then we should start working on an alternative. my 2 cents

kevinkreiser: an example alternative would be to stiil build the shortcuts for modes that could use it but exclude trucks from using the shortcut. havent thought it all the way through and it surely makes shortcut building suck but hey might be worth it if it keep auto costing perf up

nilsnolde:
In case you won't find the time for it in the next weeks, I'd take this over, after fixing some more matrix stuff. @radekvermirovsky and their company are actually a client. They've been funding a lot of my recent efforts and this is big no-go for compliant truck routing.

I'll try to come up with a strategy along your idea to add mode-specific shortcuts @kevinkreiser.

kevinkreiser:
i was going to do it yesterday but wanted to finish my chore first. i will at least get this compiling in the next couple days

kevinkreiser:
> I'll try to come up with a strategy along your idea to add mode-specific shortcuts

that sounds good. i was hoping to finish this simple idea in this pr and then yeah let you do the more complicated shortcut building that accounts for different modes as i suggested (as that will be quite a bit more time consuming) in a separate pr. is that what you are thinking?

nilsnolde:
Jep let’s do that.

nilsnolde:
Just to give a public update on our thoughts: the planet is currently building with this PR and we'll see how the performance is impacted. If it's not much (1-2%?), we might be fine leaving it as-is.

Personally I'd like to see that we don't break up shortcuts on most access restrictions and rather assign the most restrictive value of all base edges to the entire shortcut. It wouldn't be a lot of code changes, basically tracking those restrictions similar to total edge length etc, and it would restore those 1-2% or whatever almost fully for anything else than truck, and even truck would be less impacted.

nilsnolde:
Did some benchmarking on this on the 2 largest test files we have, with a running server and each file 4 complete runs for each branch:

`test_requests/de_benchmark_routes.txt`:
master: 132 secs
branch: 131 secs

`test_requests/random_routes.txt`:
master: 658 secs
branch: 659 secs

So it's fair to say, performance doesn't seem to be impacted at all :tada:

test/scripts/test_valhalla_build_extract.py
```diff
@@ -151,15 +151,15 @@ def test_create_extracts(self):
...
         self.check_tar(EXTRACT_PATH, exp_tuples, tile_count * INDEX_BIN_SIZE)
         # same for traffic.tar
-        exp_tuples = ((1536, 25568, 26416), (28672, 410441, 65552), (95232, 6549282, 604608))
+        exp_tuples = ((1536, 25568, 26416), (28672, 410441, 65312), (94720, 6549282, 604608))
         self.check_tar(TRAFFIC_PATH, exp_tuples, tile_count * INDEX_BIN_SIZE)
 
         # tests the implementation using the tile_dir
         new_tile_extract = TILE_PATH.joinpath("tiles2.tar")
-        exp_tuples = ((2560, 25568, 302544), (306688, 410441, 678640), (987136, 6549282, 6137896))
+        exp_tuples = ((2560, 25568, 302472), (306688, 410441, 676208), (984576, 6549282, 6137768))
```
nilsnolde, line 162: less edges since we avoid building some shortcuts, so the tiles gets smaller

run_route_scripts/run_with_server.py
```diff
@@ -120,3 +121,17 @@ def make_request(post_body):
...
+    duration = 0
+    for out_f in output_dir_path.iterdir():
+      if out_f.is_file() and out_f.suffix == ".json":
+        with open(out_f) as f:
+          try:
+              duration += json.load(f)["performance"]["response_time"]
+          except json.decoder.JSONDecodeError:
+              continue
+
+    print(f"Requests took {duration} seconds")
```
nilsnolde, line 137: accumulate all durations of all requests so we can see how long each run took

nilsnolde (approved):

run_route_scripts/run_with_server.py
kevinkreiser: I always run this with `time` but this is good

kevinkreiser:
really have to fix that tar index thing on osx

kevinkreiser:
I'm good to merge this if you are @nilsnolde

nilsnolde:
> really have to fix that tar index thing on osx

it's on arm too. not ever seen it on x86 linux though!

nilsnolde (dismissed):

src/mjolnir/shortcutbuilder.cc
```diff
@@ -81,21 +81,25 @@ bool EdgesMatch(const graph_tile_ptr& tile, const DirectedEdge* edge1, const Dir
...
+
+  // if they have access restrictions those must match (for modes that use shortcuts)
+  if (edge1->access_restriction()) {
+    auto res1 = tile->GetAccessRestrictions(edge1 - tile->directededge(0), kVehicularAccess);
+    auto res2 = tile->GetAccessRestrictions(edge2 - tile->directededge(0), kVehicularAccess);
+    if (res1.size() != res2.size())
       return false;
+    for (size_t i = 0; i < res1.size(); ++i) {
+      res1[i].set_edgeindex(0);
+      res2[i].set_edgeindex(0);
```
nilsnolde, line 97: just one last question: why these last 2 lines? they're modifying a copy of the restriction and seems irrelevant for the equal comparison. I'd get rid of them if that's the case.

kevinkreiser: restrictions dont have an equality operator implemented, so i thought i'll just compare using `.value()` rather than implement one. i cant just compare with `.value()` though because for two different edges, the edge indices will always be different, which means we will always return false below. by zeroing out the edge index of both edges' restrictions we can make sure to compare the only part we care about: the actual restriction meta data.

tldr we set them both to the same value exactly so that they dont matter in the comparison below (ie that part of the bitfield we dont care about is always matching)

nilsnolde: it does seem to me that the `edgeindex` is not relevant for the `value()` equality comparison, but maybe I'm blind to some concrete detail here? fwiw, the relevant tests pass if I remove those lines.

nilsnolde: I don't wanna be pedantic, but those things tend to confuse future-me and lead down to the same investigation I did now. I'd leave a comment once I understand why it's needed

kevinkreiser: if the test still passes then id be skeptical its working properly. i was going to ask you to prove it worked by removing the tagging for the restrictions from those middle ways and proving that one long shortcut is made over the whole thing. if that is the case then i cannot see how these lines would not be needed.

test/gurka/test_shortcut.cc
```diff
@@ -289,3 +289,55 @@ TEST(Shortcuts, ShortcutsInBins) {
...
+  A--B--C--D--E--F
+
+  G--H--I--J--K--L
+
+  M--N--O--P--Q--R
+  )";
+  const gurka::ways ways = {
+      {"AB", {{"highway", "motorway"}, {"name", "highway"}}},
+      {"BC", {{"highway", "motorway"}, {"name", "highway"}}},
+      {"CD", {{"highway", "motorway"}, {"name", "highway"}, {"maxweight", "3.5"}}},
```
kevinkreiser, line 305: nitpick: to be sure this is working properly we would need to also run this graph without the restrictions and prove that each line makes just 2 shortcuts instead of 4

src/mjolnir/shortcutbuilder.cc
nilsnolde: yes, I got also skeptical since codecov reports that it never got into the `return false` below. will add a quick test

kevinkreiser: oh man i misread the accessrestriction header! in fact we need to change this comparison, sorry!

kevinkreiser: pushed an update, sorry for the confusion

src/mjolnir/shortcutbuilder.cc
```diff
@@ -81,21 +81,24 @@ bool EdgesMatch(const graph_tile_ptr& tile, const DirectedEdge* edge1, const Dir
...
+
+  // if they have access restrictions those must match (for modes that use shortcuts)
+  if (edge1->access_restriction()) {
+    auto res1 = tile->GetAccessRestrictions(edge1 - tile->directededge(0), kVehicularAccess);
+    auto res2 = tile->GetAccessRestrictions(edge2 - tile->directededge(0), kVehicularAccess);
+    if (res1.size() != res2.size())
       return false;
+    for (size_t i = 0; i < res1.size(); ++i) {
+      if (res1[i].type() != res2[i].type() || res1[i].modes() != res2[i].modes() ||
+          res1[i].value() != res2[i].value())
```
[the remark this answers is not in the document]
kevinkreiser: i was going to put a comment on this too, i saw that `operator<` didnt sort on `type` that means that it could be that they are the same set of restrictions but one has hazmat first and the other has weight first but would otherwise be the same. i think the best way to fix this is in the operator, by adding type to the consideration. i started doing it in another tab but didnt commit it yet..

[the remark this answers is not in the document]
kevinkreiser: done

src/baldr/accessrestriction.cc
```diff
@@ -47,13 +47,14 @@ void AccessRestriction::set_value(const uint64_t v) {
 bool AccessRestriction::operator<(const AccessRestriction& other) const {
```
kevinkreiser, line 47: i removed the superfluous `else`s and added in `type` to the consideration. this makes it so that everything deterministically sorted

src/mjolnir/shortcutbuilder.cc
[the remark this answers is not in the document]
nilsnolde: I didn't find code that uses the operator though.. Probably I'm missing some implicit sorting somewhere? I guess at this point I'll need to add some more tests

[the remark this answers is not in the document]
kevinkreiser: i think all we need is to run the same graph but without the restrictions and assert that we only get one pair of shortcuts for each set of connected ways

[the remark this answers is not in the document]
kevinkreiser: this is where that operator is used: https://github.com/valhalla/valhalla/blob/ffe790ddfe08b89c02c419ec08cfec0bf715a64c/src/mjolnir/graphtilebuilder.cc#L287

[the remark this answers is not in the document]
nilsnolde: Great, thanks, missed that!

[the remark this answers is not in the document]
kevinkreiser: std::sort is templatized so the compiler looks for `operator<` on whatever thing is used in the template (techincally it looks first for the `operator*` to dereference the iterable thing and then it looks for the less than). in this case the template parameter is implicit as well :upside_down_face:

test/gurka/test_shortcut.cc
```diff
@@ -289,3 +289,81 @@ TEST(Shortcuts, ShortcutsInBins) {
...
+
+  G--H--I--J--K--L
+  )";
+
+  // the first line should produce only one shortcut, the second two
+  gurka::ways ways = {
+      {"AB", {{"highway", "motorway"}, {"name", "highway"}}},
+      {"BC", {{"highway", "motorway"}, {"name", "highway"}}},
+      {"CD", {{"highway", "motorway"}, {"name", "highway"}, {"maxweight", "3.5"}}},
+      {"DE", {{"highway", "motorway"}, {"name", "highway"}, {"maxweight", "8"}}},
```
nilsnolde, line 305: amended the test so it'll at least once get to the "return false" to prove we're breaking here too

test/gurka/test_shortcut.cc
```diff
@@ -289,3 +289,81 @@ TEST(Shortcuts, ShortcutsInBins) {
...
+      {"EF", {{"highway", "motorway"}, {"name", "highway"}}},
+
+      {"GH", {{"highway", "motorway"}, {"name", "highway"}}},
+      {"HI", {{"highway", "motorway"}, {"name", "highway"}, {"hazmat", "yes"}}},
+      {"IJ", {{"highway", "motorway"}, {"name", "highway"}, {"hazmat", "yes"}}},
```
nilsnolde, line 310: also that we're building shortcuts over consecutive edges with the same restriction. could've done many more permutations, but this should be ok

test/gurka/test_shortcut.cc
```diff
@@ -289,3 +289,81 @@ TEST(Shortcuts, ShortcutsInBins) {
...
+    EXPECT_NE(std::get<3>(edge)->superseded(), 0);
+  }
+
+  // test that without those restrictions we're still building all shortcuts
+
+  // remove those access restrictions 
+  ways["CD"].erase("maxweight");
+  ways["DE"].erase("maxweight");
+  ways["HI"].erase("hazmat");
+  ways["IJ"].erase("hazmat");
```
nilsnolde, line 347: do another graph without those restrictions and test that we're only creating 1 shortcut now

kevinkreiser:
yep its what i expected, good touch with the checking that matching access restringions allow shortcuts to be built

kevinkreiser:
feel free to put a sh ip on it again so we can merge once the testing shows basically no difference :smile:

nilsnolde (approved):

nilsnolde:
Why is codecov doing that? It's a 100% tested and still it reports the whole coverage is decreased.. sometimes I can understand if patch coverage is close to current project coverage and we remove quite a bit of code which might've been 100% tested, the overall coverage decreases, makes sense. But in this case it's clearly bullshit..

Issue #4274, opened 2023-08-27
CostMatrix should check for connections on reverse tree


nilsnolde:
If there's edges with a high cost (e.g. a ferry), it can happen that the forward tree of the CostMatrix finds a connection, sets a fairly low threshold and the ferry edge is never settled. However, on the backward tree the edge might've been settled already and if we'd check similar to bidir a*, we'd return the right cheapest path and would have more consistent results (**note**, routing doesn't have that problem).

E.g. these parameters cause the matrix to not take the much faster ferry for 1744 secs, instead it detours on land with 3352 secs:

```json
{
	"units": "kilometers",
	"sources": [
		{
                        "lon": 9.481841,
                        "lat": 47.652777
                      }
	],
	"targets": [
		{
                        "lon": 9.370105,
                        "lat": 47.557136
                      }
	],
	"costing": "auto",
	"costing_options": {
		"auto": {
			"use_ferry": 0.5
		}
	}
}
```

Same locations with routing return 1744 secs, so does increasing the thresholds or `use_ferry=1.0` for matrix. 

For reference, here's the route: https://valhalla.openstreetmap.de/directions?profile=car&wps=9.481780526645279%2C47.65277745225442%2C9.369959279164652%2C47.557331368647

nilsnolde:
try again with increased thresholds instead

PR #4329, opened 2023-10-13
enable bidir matrix connections on reverse tree

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4274

I'll comment more inline. It's similar to bidir A* now. A future PR could maybe shed some code there, templating a `FindConnection` function, both for bidir A* and matrix.


scripts/format.sh
```diff
@@ -25,7 +25,7 @@ find src valhalla test bench -type f -name '*.h' -o -name '*.cc' \
 
 # Python setup
 py=$(setup_python)
-${py} -m pip install black==22.10.0 flake8==5.0.4
+${py} -m pip install --break-system-packages black==22.10.0 flake8==5.0.4
```
nilsnolde, line 28: newer pip (I think) has adopted an annoying parenting attitude and wouldn't let people install packages to system python anymore, even sudo doesn't help.. needs this flag. still better than requiring people to have a venv.

src/thor/costmatrix.cc
```diff
@@ -41,13 +41,13 @@ inline float find_percent_along(const valhalla::Location& location, const GraphI
 namespace valhalla {
 namespace thor {
 
-class CostMatrix::TargetMap : public robin_hood::unordered_map<uint64_t, std::vector<uint32_t>> {};
+class CostMatrix::ReachedMap : public robin_hood::unordered_map<uint64_t, std::vector<uint32_t>> {};
```
nilsnolde, line 44: not only for targets anymore

src/thor/costmatrix.cc
```diff
@@ -41,13 +41,13 @@ inline float find_percent_along(const valhalla::Location& location, const GraphI
...
 
 // Constructor with cost threshold.
 CostMatrix::CostMatrix(const boost::property_tree::ptree& config)
     : mode_(travel_mode_t::kDrive), access_mode_(kAutoAccess), source_count_(0),
       remaining_sources_(0), target_count_(0), remaining_targets_(0),
-      current_cost_threshold_(0), targets_{new TargetMap},
+      current_cost_threshold_(0), targets_{new ReachedMap}, sources_{new ReachedMap},
```
nilsnolde, line 50: I didn't see another way than keeping the same huge data structure for `sources` as we have for `targets`. bit annoying but as you'll see in the tests (and by fixing the initial issue), maybe it's worth it

src/thor/costmatrix.cc
```diff
@@ -817,6 +829,102 @@ void CostMatrix::BackwardSearch(const uint32_t index, GraphReader& graphreader)
   }
 }
 
+// Check if the edge on the forward search connects to a reached edge
+// on the reverse search trees.
+void CostMatrix::CheckReverseConnections(const uint32_t target,
```
nilsnolde, line 834: won't comment here, it's basically like `CheckForwardConnection`

test/matrix.cc
```diff
@@ -233,7 +233,7 @@ TEST(Matrix, test_matrix) {
 
   CostMatrix cost_matrix_abort_source;
   cost_matrix_abort_source.SourceToTarget(request, reader, mode_costing, sif::TravelMode::kDrive,
-                                          100000.0);
+                                          90000.0);
```
nilsnolde, line 236: this test is testing that we abort the search for anything over 100 km and tests that 1 pair wasn't connected before expanding that far. so it takes less cycles to find the pair that the test tries to miss here, that's why the test failed. it's good that it's even showing in the test.

disclaimer: I tried to reproduce this test on master but couldn't for some reason. I changed the config with the 100k m max_matrix_distance, it still found all 16 pairs and there was no change to this PR. I have 0 idea why..

valhalla/thor/costmatrix.h
```diff
@@ -59,7 +59,8 @@ struct BestCandidate {
       : found(false), edgeid(e1), opp_edgeid(e2), cost(c), distance(d), threshold(0) {
   }
 
-  void Update(const baldr::GraphId& e1, baldr::GraphId& e2, const sif::Cost& c, const uint32_t d) {
+  void
+  Update(const baldr::GraphId& e1, const baldr::GraphId& e2, const sif::Cost& c, const uint32_t d) {
```
nilsnolde, line 63: just added a `const` to the 2nd argument and updated the docstring

nilsnolde:
fucked up CI:
- spell check: now triggering for totally unrelated files. I just realized we use the github action from their `master` branch, certainly not ideal.. apparently they updated `driveable` to `drivable` :roll_eyes: . I pinned the version to the date of the PR introducing it, now it's working again. 
- azure: some 404's for msys packages when installing vcpkg. it doesn't fail the step, then cmake fails with a cryptic error. I'll kick it again in the next days, should be only temp

test/matrix.cc
nilsnolde: fwiw, the time was identical with `route` for that pair

dnesbitt61:
What does this do to performance? I guess my question is whether the accuracy gained is worth the performance loss. I think when I implemented this I had read that connections were checked on the forward pass.

nilsnolde:
I can’t imagine more than a few percent, but let me try and report back.

nilsnolde:
And FWIW, conceptually it’s not just accuracy, it’s also performance gains (by finding a connection with less expansion), though admittedly just in some instances. Still, it even affected the test with 5 locations, indicating it’s not that super rare.

nilsnolde:
I ran `valhalla_run_matrix` on 27 and 51 locations in/around Detroit with 10 requests each to average out:

master (27 locs): 3.0116 secs
branch (27 locs): 3.1849 secs
~5.5% slowdown

master (51 locs): 4.639400 secs
branch (51 locs): 5.628601 secs
~18% slowdown

don't really understand that relationship.. I'd have expected a more or less constant impact relatively, e.g. 5%, no matter the size of the expansion.. I'll look into this again soon

scripts/format.sh
nilsnolde: will remove this again, now that we have another solution in master

src/thor/costmatrix.cc
```diff
@@ -817,6 +829,102 @@ void CostMatrix::BackwardSearch(const uint32_t index, GraphReader& graphreader)
...
+    return;
+  }
+
+  // Get the opposing edge. Get a list of source locations whose forward
+  // search has reached this edge.
+  GraphId fwd_edgeid = rev_pred.opp_edgeid();
+  auto sources = sources_->find(fwd_edgeid);
+  if (sources == sources_->end()) {
+    return;
+  }
```
nilsnolde, line 856: I ran the 51 locs example again multiple times each, the slowdown seems to be in the 10-15% range. 

I don't quite get it.. the code stops right here for the vast majority of reverse expanded edges, since it's only towards the end. So it's mostly "just" the lookup `sources_->find(fwd_edgeid)` (plus putting the edge/pair id into that map). I wouldn't have thought that'd correspond to ~ 10% of the whole execution time. 

\> 10% is quite a lot for a little improvement in quality. I'd also find it sad to compromise (even more) quality for performance.

WDYT @dnesbitt61 @kevinkreiser ?

kevinkreiser: some questions.. how are you testing this? are you running valhalla_service in one shot mode or are you making http requests while keeping the server running? i always test with: https://github.com/valhalla/valhalla/blob/5b3cd9704ffd699b992b08bc4a5bf60e34d318f5/run_route_scripts/run_with_server.py

it simulates a real world condition of a long running server process with http requests hitting it in parallel. doing multiple runs of this is the most trust worthy benchmark we have (imho).

if we think its not the look up, are you sure its not the extra allocations/deallocations assocated to the new map itself. i notice that we resize to 0 and shrink to fit, essentially forcing the container to deallocate and reallocate for the next request. imho it would be better to do like we do in other places which is to allow some amount of that container to keep its reservation (ie `resize(MAX_MATRIX_CONNECTIONS_RESERVATION)` so that modestly sized matrix requests can just reuse the slots we've already allocated for previous requests.

another optimization we could do is to not use two containers. edge_ids have plenty of unused bits. you could use a single container to hold both sets of potential connections by bitwise or'ing a single bit mask over the id to differentiate between edge ids that are keys for the sources vs keys for the targets. the only intuitive benefit of this would be that regardless of search path direction it would be looking at the same bit of memory when calling `.find` on the map which might have some kind of cache hit benefit

also could we do some sanity checks? can you simply disable the call to this new function and see if all the performance is recovered? if it is then it has to either be the call to `.find` or that it gets past this point more often than we would expect and does the looping below.

long story short, can we do a little bit more investigation to inform ourselves on why? also, its totally ppossible you did all this and i just dont know it, but in case yoou didnt these are my ideas :smile:

nilsnolde: I used `valhalla_run_matrix` which still has the cold start problem, but with 10 runs, it shouldn't hit that hard? I'll try the run_with_server.py..

Didn't I fix those non-allocations?! I'll look into it again.

> another optimization we could do is to not use two containers. edge_ids have plenty of unused bits.

Intuitively, without looking into it, we need to use the full 46 bits for this (or whatever it was, surely more than 32). What am I missing?

I'll sanity check more, thanks for the hints!

kevinkreiser: > Intuitively, without looking into it, we need to use the full 46 bits for this (or whatever it was, surely more than 32). What am I missing?

the key in the map is uint64_t, which is what graph_ids are. the edge ids only use up, like you said, 46 bits of that. that leaves 18 spare bits for you to do whatever you like with. to use the same container, you only need to differentiate whether you are asking the container for an edge on behalf of the forward or reverse search paths. which means you only need one bit designating forward or reverse. pick one of the directions and any time you ask for it, mask the edge id with one of the spare 18 bits flipped `lookup.find(edge_id | (1 << 47))`

src/thor/costmatrix.cc
```diff
@@ -78,7 +78,13 @@ float CostMatrix::GetCostThreshold(const float max_matrix_distance) {
 // Clear the temporary information generated during time + distance matrix
 // construction.
 void CostMatrix::clear() {
-  // Clear the target edge markings
+  // Clear the source/target edge markings
+  for (auto& iter : *sources_) {
+    iter.second.clear();
+    iter.second.resize(0);
+    iter.second.shrink_to_fit();
+  }
```
kevinkreiser, line 86: this, and the same version below for targets is definitely not free. here we pay the cost of dealloacting everything and then in the next request we have to do the allocations all over again. it would be interesting here to instead do something like:

```suggestion
  for (auto& iter : *sources_) {
    iter.second.resize(MATRIX_CONNECTION_ENTRIES_MAX_BUFFER_SIZE);
    iter.second.shrink_to_fit();
    iter.second.clear();
  }
```

and the same below for targets

nilsnolde: yeah, seems like I missed a spot there!

src/thor/costmatrix.cc
nilsnolde: right, thanks!

nilsnolde:
This will be one of the next things I'd work on, after #4372 is merged. I'd hate to lose all the wonderful speedups of that PR right here again. I don't think I'll be able to speed things up here considerably. Staring at the CostMatrix for the past 1-2 weeks, I now understand that this change will degrade performance more the more locations are involved, it makes sense.

However, this is still good to have for quality reasons. Basically, what happens right now, is that the forward tree needs to _settle_ the connecting edge _before_ the reverse tree sees its opposing. That must have negative implications on quality.

So, @dnesbitt61 @kevinkreiser what do you think about making this a configurable option? I think some people would be very interested in increasing solution quality for matrix calculations (e.g. for optimization), but it's already quite good so most could people would live without and enjoy the higher performance.

nilsnolde:
It's very low effort/complexity, just passing in another config option into CostMatrix ctor, among the others that are already there.

nilsnolde:
Maybe the misunderstanding is that I'm not proposing a runtime option, but a config option, similar to the label reservations for the algos.

kevinkreiser:
oh i meant the complexity of having the algorithm respect the option not how the option gets to the algorithm. maybe a few pr comments in the code about where the option would modify  the algorithms behavior would make it more obvious, ive not reviewed the code for a bit

nilsnolde:
I hacked it in quickly in this PR so you can see: https://github.com/valhalla/valhalla/pull/4329/commits/974c0594902fd001619dd076bfa1f357cc40c48e. 

FWIW, I already rebased on #4372 in a different branch, so it'd be ready as well, after that base one is merged.

nilsnolde:
superseded by #4418

nilsnolde:
Some more investigation on performance with 50 sources/targets (symmetrical) and 50 requests in Berlin area. master came out with 61 secs.

- only allocations (not actually checking connections): 65 secs, so around 6% degradation
- allocations + checking connections, i.e. full PR: 78 secs, so ~ 17% extra degradation, totaling -23%

We'd save some by using the existing robin hood hash, but the majority is with the connection check itself. We should still do that optimization as @kevinkreiser suggested, even though we surely wouldn't too much of the 6% degradation. But it could be in a next PR as well IMO. WDYT?

src/thor/costmatrix.cc
```diff
@@ -294,8 +298,6 @@ void CostMatrix::Initialize(
     const google::protobuf::RepeatedPtrField<valhalla::Location>& source_locations,
     const google::protobuf::RepeatedPtrField<valhalla::Location>& target_locations) {
 
-  targets_->reserve(max_reserved_labels_count_);
```
nilsnolde, line 297: removing the reservation from the robin hood hash is actually better for performance! I wonder why..

test/gurka/test_expansion.cc
```diff
@@ -99,11 +99,11 @@ TEST_P(ExpansionTest, RoutingNoOpposites) {
 }
 
 TEST_P(ExpansionTest, Matrix) {
-  check_result("sources_to_targets", {"E", "H"}, false, 48, GetParam());
+  check_result("sources_to_targets", {"E", "H"}, false, 54, GetParam());
```
nilsnolde, line 102: one test with the option turned on of course

test/gurka/test_matrix.cc
```diff
@@ -87,11 +87,13 @@ class MatrixTest : public ::testing::Test {
...
 
     const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize);
+    // also turn on the reverse connection search; there's no real test for it
     map = gurka::buildtiles(layout, ways, {}, {}, "test/data/matrix_traffic_allowed",
                             {{"service_limits.max_timedep_distance_matrix", "50000"},
                              {"mjolnir.traffic_extract",
                               "test/data/matrix_traffic_allowed/traffic.tar"},
-                             {"mjolnir.timezone", "test/data/tz.sqlite"}});
+                             {"mjolnir.timezone", "test/data/tz.sqlite"},
+                             {"thor.costmatrix_check_reverse_connection", "1"}});
```
nilsnolde, line 96: updated this test to turn on the new option. it'd be pretty tedious to come up with a test that shows the benefit of it. the request at the original issue in https://github.com/valhalla/valhalla/issues/4274 works now though

test/gurka/test_matrix.cc
```diff
@@ -53,7 +53,7 @@ void check_matrix(const rapidjson::Document& result,
 }
 } // namespace
 
-class MatrixTest : public ::testing::Test {
+class MatrixTrafficTest : public ::testing::Test {
```
nilsnolde, line 56: also renamed the test, it's not anymore only traffic specific. the class here is though.

nilsnolde:
fixed the tests @kevinkreiser . almost under 50 open PRs, not bad for 2 weeks of meetings:)

kevinkreiser (approved):

PR #4334, opened 2023-10-16
more sustainable way to work with protobuf in cmake

nilsnolde:
I committed my local patch way too often.


CMakeLists.txt
```diff
@@ -180,35 +180,15 @@ else()
 endif()
 
 if(NOT Protobuf_FOUND)
+  # prefer CONFIG mode over MODULE mode, there were a few CMake/protobuf fuckups
+  set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
```
nilsnolde, line 184: finally found a solution that should work: https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html. the intention was different for this, but it works for our case.

CMakeLists.txt
```diff
@@ -180,35 +180,15 @@ else()
...
-      find_package(absl REQUIRED)
-      if(TARGET protobuf::libprotobuf-lite)
-        target_link_libraries(protobuf::libprotobuf-lite INTERFACE "absl::absl_log")
-      elseif(TARGET protobuf::libprotobuf)
-        target_link_libraries(protobuf::libprotobuf INTERFACE "absl::absl_log")
-      endif()
-    endif()
-  endif()
+  # and turn it off again
+  set(CMAKE_FIND_PACKAGE_PREFER_CONFIG OFF)
```
nilsnolde, line 191: can't leave it on for the whole file. geos has problems finding the right target, which is almost surely due to our own FindGeos.cmake file which defined the target differently from the distributed geos-config.cmake file (if it's even installed). we could take this to make our Find* modules consistent with the *-config.cmake files, but meeh

nilsnolde:
Can we merge this after it finished building? It's so annoying for me when changing branches etc

kevinkreiser (approved):

PR #4339, opened 2023-10-17
remove conan from windows build instructions

### LINKED ISSUE #4331: windows builds failing on `conan --version`
```
Starting: Run CMake to configure build
==============================================================================
Task         : Command line
Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
Version      : 2.229.0
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
==============================================================================
Generating script.
========================== Starting Command Output ===========================
"C:\Windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\21a36795-3ed8-4bac-8314-973d04336fe7.cmd""
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.7.4
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'
cmake version 3.27.6

CMake suite maintained and supported by Kitware (kitware.com/cmake).
-- The CXX compiler identification is MSVC 19.35.32217.1
-- The C compiler identification is MSVC 19.35.32217.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.35.32215/bin/HostX64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.35.32215/bin/HostX64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found PkgConfig: C:/Strawberry/perl/bin/pkg-config.bat (found version "0.26") 
-- Configuring in release mode
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
-- Found Threads: TRUE  
-- Found ZLIB: D:/a/1/s/vcpkg/installed/x64-windows/lib/zlib.lib (found version "1.2.13")  
-- No compatible boost version detected, using conan...
-- Conan: Detected VS runtime: MD
-- Conan: checking conan executable
-- Conan: Found program C:/hostedtoolcache/windows/Python/3.12.0/x64/Scripts/conan.exe
CMake Error at cmake/conan.cmake:813 (message):
  Conan --version failed='1'
Call Stack (most recent call first):
  cmake/conan.cmake:538 (conan_check)
  CMakeLists.txt:153 (conan_cmake_install)
```

looking at the above it seems the command `conan --version` has a return code of 1. debugging this might be annoying but at the very least we can hack on `cmake/conan.cmake` to print some stuff out where its failing..

nilsnolde:
fixes #4331 

- removes conan from win builds in favor of vcpkg
- changes the way vcpkg works

didn't try on win, only on linux, probably needs some more work.

as follow-up I'd like to apply vcpkg to at least one linux CI build


.azure-pipelines.yml
```diff
@@ -76,26 +68,17 @@ jobs:
...
 
-      - task: UsePythonVersion@0
-        inputs:
-          versionSpec: '3.x' 
-          addToPath: true 
-          architecture: "$(TRIPLET)"
-
       - script: |
           git clone https://github.com/microsoft/vcpkg %VCPKG_DIR%
-          git -C %VCPKG_DIR% checkout %VCPKG_REF%
```
nilsnolde, line 87: we'll use master going forward

vcpkg.json
```diff
@@ -0,0 +1,21 @@
...
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
+  "dependencies": [
+    "boost-algorithm",
+    "boost-foreach",
+    "boost-format",
+    "boost-geometry",
+    "boost-optional",
+    "boost-property-tree",
+    "boost-range",
+    "boost-tokenizer",
```
nilsnolde, line 11: these are all header-only, but currently this installs quite a few compiled libraries as deep-endencies, like boost-regex. just fyi. some might be optional (if not all)

.azure-pipelines.yml
kevinkreiser: but if we do that then every time master is updated it could potentially update one of our depenencies, which would be bad for several reasons:

1. it blows away our vcpkg cache or at least it should right? the commit hash should be part of the circle cache key since new packaging code could mean new packages
2. blowing away the cache causes us to fetch the deps and build them from source which means very slow builds every time vcpkg/master is updated
3. what happens when a new version of a dependency is no longer compatible with our code. mainlining the latest stuff all the time can totally screw us into a bunch of work to bring our code up to date with whatever library, not to mention could pull in new bugs from those or slowdowns or whatever. its certainly better to pin the versions of our dependencies

i dont know much about vcpkg but these are my concerns, maybe im totally misunderstanding. its better for us to take updates when we want to rather than randomly via external factors. thats already a sucky part of CI we should seek to minimize that. every now and then we can update and validate that the update is good (like we do for thirdparty submodules)

nilsnolde: convinced:) my reasoning was that _sometimes_ details in the packaging changes (i.e. how they're built, e.g. URLs to sources or so). but that really only happened a few times in the past years.

.azure-pipelines.yml
```diff
@@ -76,26 +69,18 @@ jobs:
...
+          git checkout %VCPKG_REF%
```
nilsnolde, line 74: nooooo, ffs..  I built its master accidentally.. damn, will have to invalidate the cache once again.

kevinkreiser (approved):
:pray:

kevinkreiser:
@localizepic its not normal. in fact if they all use the same config i dont see how they would be able to start because they would all be trying to bind on the same ipcs and ports.

kevinkreiser:
kill them all and just start one of them :smile:

localizepic:
Already did it but even if I pass the argument for one core "valhalla_service ~/valhalla/scripts/conf/valhalla.json 1" it start with 16 processes.
Other thing, when I kill just one process it kill all, seem's they are cloned...strange

kevinkreiser:
im not sure how this is possible. that executable can only be a single process. if you kill them all and then do `ps aux | grep valhalla` you see nothing running right? and hten you simply run `valhalla_service` and you get 16 instances of it? are you sure you dont have supervisord automatically restarting processes or something?

localizepic:
Thank you very much for your explanations, btw I'm testing and running valhalla in a dedicated server with 128GB RAM, 12 CPU's. What are the recommended maximum number of cores should I pass as argument let's say for a massive usage?

kevinkreiser:
pass it a 12 and use all your cpus maxed out. that will end up as like 40 threads but should be fine

Issue #4281, opened 2023-08-31
Dirt roads for motorcycle navigation.


molind:
I've received a few feature requests from our users and I'm looking at how difficult it could be. Motorcyclists often look for adventures and are ready to ride on tracks and some dirt paths. We already use `useTrails`, which adds priority to secondary roads, while they are asking about `highway=track` roads. What do you think about it? Does it make sense to add an extra option that unlocks track roads for navigation?

nilsnolde:
Yeah motorcyclists often want some curious extras. IMO it’d make sense to allow preference of dirt roads (or rather avoidance of non-dirt roads?). 

Curvature is probably even more important? We have that available as edge property already IIRC and could be exposed. See https://kurviger.de for a Graphhopper based router.

bmapsnavigation:
I use curvature in a local valhalla instance. Currently. and if I remember correctly, it builds with tests turned off or else the build fails. Still working on fixing that, but it's a slow progress as I find programming in C quite the challenge. My solution is an slightly adapted motorcyclecost.cc and a couple of other things I found here on the valhalla github when looking for how to add a custom cost, and besides testing during build, that actually works fine as far as I can see. There will always be a bit of discussion whether someone would prefer road A over B, because "A has more curves than B" though.

I would be interested to see if some kind of off road navigation could be build-in in valhalla.

molind:
Here's the corrected version:

@nilsnolde Yes, I've seen it already. However, in this issue, I'd like to discuss the possibility of using `highway=track` for motorcycle routes. There are many tracks where driving motorcycles is legally permitted or not explicitly prohibited. We could probably use rules from https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions. Tracks shouldn't be allowed by default, but if someone needs to go there and it's legal, we should be able to build a route.

molind:
[Access restrictions](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions) mention Austria, Belarus, France and others. But that `use_tracks` definitely worth a try. Thanks for mentioning it!

kevinkreiser:
looks related to this issue #4336 too

nilsnolde:
I'll close here then, feel free to re-open if there's anything else

molind:
Seems I was too happy to learn about that option. It worked for car, but doesn't work for motorcycle. Tried again today and it builds car route, but won't build route for motorcycle.

```
[...]

PR #4348, opened 2023-10-21
Allow all roads for motorcycles.

molind:
# Issue

There are motorcycles that able to drive tracks and paths. They should be able to build route on that roads. While sport-bikes could omit them using `exclude_unpaved` option.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

As mentioned in #4281


molind:
#fixes #4281

kevinkreiser:
imho there could be an argument for this change being a breaking change if the intended default for motorcycle was a sport bike that couldnt go off road. by changing this, we change the default which is a pretty serious change in functionality.

after reading the documentation though i notice there is no mention of surface specifically wrt motorcycle costing. furthermore the `exclude_unpaved` flag defaults to false. this would lead one to believe motorcycle can use unpaved roads by default. so indeed, this change is a fix for broken functionality if we go by the docs (or lack thereof). i have a bit of a feeling this was on purpose though (thinking about how @dnesbitt61 wrote it and he's been a sport bike rider as far as i know).

that said im cool with this change, i just wanted to put some official thought into such an important change as changing the default way a given costing works.

nvcnvn:
I think the best is we can have minimum surface as an option. Keep default "dirt" level for now.

In Vietnam and most of South East Asia, the OSM map tagged with "motorcycle" but most vehicles we have is just better than a scooter, not a sport kind of motorcycle.
We drive scooter but follow motorcycle instructions.
It's very normal for a scooter/motorcycle driver to enter whatever the surface of the road.

nilsnolde:
@kevinkreiser is right, I only really thought about that impact when I read the OP's last sentence.. I agree with @nvcnvn, maybe should keep the current behavior and introduce a `minimum_allowed_surface` as [0,7] which defaults to 4 (`kDirt`)? 

The bike profile doesn't allow for that, instead it chooses a hard-coded `minimum_allowed_surface` based on the bike type. Doesn't make sense here. In the future there could be a `avoid_bad_surfaces` costing option as well.

WDYT @molind ?

molind:
What options do we have?
1. Add a `minimum_allowed_surface` option, similar to `max_hiking_difficulty`.
2. Invert `exclude_unpaved` for motorcycles and hardcode `minimum_allowed_surface` to `kImpassable`.

I don't believe that `minimum_allowed_surface` would be used with values other than `kDirt` and `kImpassable`. In that case, it's very similar to `exclude_unpaved`. Having two options that seem to solve the same task adds unnecessary complexity. 

If `exclude_unpaved` default value will be changed to `true` for motorcycles it will be safe to show that routes even to sport bikes.

kevinkreiser:
i was trying to say that although it is a change of a default, which is against the idea of not breaking or changing users experiences, the way the docs reads, it technically looks like a bug, and that by making this change we fix a long standing bug.

normally i would not be in favor of making such a change except in two circumstances:

1. we are fixing a bug, ie the code is not working the way it is intended
2. the change allows more routes succeed (maybe a special class of bug fix)

in this case i think we could argue that we meet these criteria but im somewhat ambivalent on this specific case. what tips me toward allowing it is that it avoids further complicating the api.

it should be noted that it is possible certain users are relying on not allowing dirt sections but at least their code will continue to work without them making a code change. they just might get back a route with a dirt road in it when they didnt want it. this is likely to be a minority of users.

imho we can make this change, but also its very important to have the discussion first. what do you guys say?

nilsnolde:
Good points. That's fine with me too.

molind:
Any news on this?

kevinkreiser:
as described im cool with merging it, can you bring your branch up to date @molind ?

molind:
@kevinkreiser done

nilsnolde (approved):

PR #4356, opened 2023-10-26
add --optimize & --log-details to valhalla_run_matrix

nilsnolde:
the logging it does clutters the terminal enormously and often I'm not interested in its output. and I made optimization optional per arg.


nilsnolde:
BTW, I can also negate the flags to preserve current behavior (i.e. `--no-optimize` etc), in case you rely on it.

dnesbitt61 (dismissed):

dnesbitt61 (approved):

PR #4359, opened 2023-10-27
motorcar:conditional should not apply to motorcycle and moped

### LINKED ISSUE #4355: "motorcar:conditional" should not restrict motorcycle
https://github.com/valhalla/valhalla/blob/master/src/mjolnir/pbfgraphparser.cc#L2390-L2393

Any particular reason to apply motorcar:conditional to motorcycle is a different kind?

nvcnvn:
# Issue

fixes #4355

## Tasklist

 - [ ] Add tests


src/mjolnir/pbfgraphparser.cc
```diff
@@ -2387,10 +2387,12 @@ struct graph_callback : public OSMPBF::Callback {
         if (tokens.size() == 2 && tmp.size()) {
 
           uint16_t mode = 0;
-          if (boost::algorithm::starts_with(tag_.first, "motorcar:conditional") ||
-              boost::algorithm::starts_with(tag_.first, "motor_vehicle:conditional")) {
+          if (boost::algorithm::starts_with(tag_.first, "motor_vehicle:conditional")) {
```
nilsnolde, line 2390: we have to respect `motorcar` as well though. I'm not sure what would be the best thing here, as that usually implies in some way that motorcycle & motorscooter are affected as well. If it's a `kTimedDenied` it usually doesn't count for motorcycles and they're still allowed, but if it's a `kTimedAllowed` it should also allow motorcycles. so maybe it `motorcar` needs to distinguish the type? That's what https://wiki.openstreetmap.org/wiki/Key:motorcar suggests.

src/mjolnir/pbfgraphparser.cc
```diff
@@ -2387,10 +2387,12 @@ struct graph_callback : public OSMPBF::Callback {
...
             mode = (kAutoAccess | kTruckAccess | kEmergencyAccess | kTaxiAccess | kBusAccess |
                     kHOVAccess | kMopedAccess | kMotorcycleAccess);
+          } else if (boost::algorithm::starts_with(tag_.first, "motorcar:conditional")) {
+            mode = (kAutoAccess | kTruckAccess | kEmergencyAccess | kTaxiAccess | kBusAccess |
+                    kHOVAccess);
```
nilsnolde, line 2395: Oh sorry, for some reason my eyes didn't see that you're doing that here.. Anyways, I think this is too broad. This is only valid for `kTimeDenied`. for `kTimeAllowed` we don't need to add anything I think:

```suggestion
          } else if (boost::algorithm::starts_with(tag_.first, "motorcar:conditional")) {
          if (type == kTimedDenie) {
             mode = (kAutoAccess | kTruckAccess | kEmergencyAccess | kTaxiAccess | kBusAccess |
                      kHOVAccess);
           }
```

nvcnvn: I will follow you suggestion. (I only know how to drive motorcycle LoL)

src/mjolnir/pbfgraphparser.cc
nilsnolde: @gknisely I personally only looked at the wiki definition. IMO it's pretty clear that both motorcar and motorcycle are a subset of motor_vehicle, but motorcycle can't be a subset of motorcar, it's clearly a car.. If any mapper doesn't think so, I personally don't care, I can only stomach a certain level of OSM craziness :smile:

src/mjolnir/pbfgraphparser.cc
```diff
@@ -2387,10 +2387,16 @@ struct graph_callback : public OSMPBF::Callback {
         if (tokens.size() == 2 && tmp.size()) {
 
           uint16_t mode = 0;
-          if (boost::algorithm::starts_with(tag_.first, "motorcar:conditional") ||
-              boost::algorithm::starts_with(tag_.first, "motor_vehicle:conditional")) {
+          if (boost::algorithm::starts_with(tag_.first, "motor_vehicle:conditional")) {
             mode = (kAutoAccess | kTruckAccess | kEmergencyAccess | kTaxiAccess | kBusAccess |
                     kHOVAccess | kMopedAccess | kMotorcycleAccess);
+          } else if (boost::algorithm::starts_with(tag_.first, "motorcar:conditional")) {
+            if (type == kTimedAllowed) {
```
nilsnolde, line 2394: the build fails @nvcnvn 
```suggestion
            if (type == AccessType::kTimedAllowed) {
```

nilsnolde (dismissed):

src/mjolnir/pbfgraphparser.cc
gknisely: fair enough

gknisely (dismissed):

nilsnolde (approved):

nilsnolde:
argh, now both arm & osx are failing with https://github.com/valhalla/valhalla/issues/4379

PR #4361, opened 2023-10-27
Ferry reclassification fix when multiple edges exist between a node pair

### LINKED ISSUE #4357: Ferry Approach Edge Reclassification - Recovers multiple edges from node on ShortestPath
I added a log statement in graphbuilder to identify driveways that are reclassified by the ferry_connections logic in mjolnir. I found that the ShortestPath method stores the node and its predecessor (and cost) along the ShortestPath from the ferry connection node. This can lead to multiple edges being reclassified when there are multiple edge between two nodes as in the case below:

image

This is on Long Island, NY - this driveway (https://www.openstreetmap.org/way/1003285180) ends up being reclassified. 

To fix this, we can add the edge wayindex to the node labels for the shortest path. Then when recovering the edges along the ShortestPath (edges to be reclassified) we can check for matching edge wayindex so we don't pick the higher cost path between 2 edges.

One point of discussion - ferry entrances with multiple "lanes" could benefit from this "bug" as it would reclassify more edges in these cases. However, in a route it would still construct shortest paths using only one lane so it likely doesn't matter if only one is reclassified and allows use in longer routes where road class hierarchy comes into play.

image

dnesbitt61:
fixes #4357 

When adding a node label, add the edge's wayindex. When recovering theshortest path, make sure the wayindex of the edge connecting the node with its predecessor node matches the wayindex. This ensures the correct edge on the shortest path is marked for reclassificiation. Prior to this, multiple edges could become reclassified when there is more than 1 edge between a node pair.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


dnesbitt61:
of course tests fail. My code changes were started on a fork that did not have all of the "mode" changes. I am not quite sure how to resolve this - I can look later but if anyone has any thoughts please let me know!

nilsnolde (commented):
I'm probably annoying with this, but a test would be nice;) there's already a gurka/test_ferry_connections.cc file. I can add that as well (and this time for real:)), just let me know.

nilsnolde:
> of course tests fail. My code changes were started on a fork that did not have all of the "mode" changes.

just seeing that. I think the only thing to do is to merge current upstream master in your branch and resolve the conflicts. but can't be much if any?

nilsnolde:
@dnesbitt61 , see the test here: https://github.com/valhalla/valhalla/compare/master...nn-ferry-test. luckily there was a fitting layout already:)

dnesbitt61:
I am not very familiar with gurka. Is it even possible to set 2 edges between a node-pair? 

A ------ B
 |          |
 ----------

If you add a node C (so you have A-C-B) that won't test this case. There has to be two distinct edges between A and B.

image

nilsnolde:
Did you look at my test? Isn't it doing the right thing?

nilsnolde:
Oh wait, that's wrong.. let me amend.

nilsnolde:
Now @dnesbitt61 , was too quick before, couple of typos

dnesbitt61:
In gurka, do each of the letters become a node in the graph (in your case, are I and J nodes in the graph)? I see you have BIJC as a way, so as long as I and J are not nodes in the graph then this works for me.

nilsnolde:
can't prove it with code, but yeah, AFAIU those "nodes" are only there for geometry, and really only the first & last vertex/letter become graph nodes. it's failing properly, so that must be the case:)

gknisely:
@dnesbitt61 you need to format / lint

nilsnolde (approved):
:ship:

nilsnolde:
phew that was close :D

kevinkreiser:
> can't prove it with code, but yeah, AFAIU those "nodes" are only there for geometry

thats right, they are OSM nodes, not graph nodes. they only get upgraded to graph nodes if mjolnir decides it :smile:

gknisely (approved):

PR #4362, opened 2023-10-27
More matrix clearing memory

nilsnolde:
fixes CostMatrix memory stuff ref https://github.com/valhalla/valhalla/pull/4329#discussion_r1373217995

I ran the same 51 location request as in https://github.com/valhalla/valhalla/pull/4329#issuecomment-1763687934. Those are the times for 10 of those requests on 4 threads (both server & `run_with_server.py`), run each 3 times with very little variance:

|        | run        | time  |
|--------|------------|-------|
| master | 1st        | 45.2s |
| master | subsequent | 40.7s |
| branch | 1st        | 36.7  |
| branch | subsequent | 28.5  |

That's around 20% speedup for the first, uninitialized request and ~30% speedup for subsequent requests. Quite good:) Of course this speedup degrades with more locations and/or more distance in between them. In this example I couldn't make out any noticeable extra memory.

### DIFF
--- a/run_route_scripts/run_with_server.py
+++ b/run_route_scripts/run_with_server.py
@@ -3,6 +3,7 @@ import argparse
 import multiprocessing
 import requests
 import json
+from pathlib import Path
 import os
 import shutil
 import datetime
@@ -59,11 +60,13 @@ def make_request(post_body):
 
       # summary json
       elif args.format == 'json':
-        out = {'routes':[{'legs':[]}]}
-        for leg in response['trip']['legs']:
-          out['routes'][-1]['legs'].append({'maneuvers':[]})
-          for man in leg['maneuvers']:
-            out['routes'][-1]['legs'][-1]['maneuvers'].append({'length': man['length'], 'time': man['time'], 'instruction': man['instruction']})
+        out = dict()
+        if args.url.endswith("route"):
+          out['routes'] = [{'legs':[]}]
+          for leg in response['trip']['legs']:
+            out['routes'][-1]['legs'].append({'maneuvers':[]})
+            for man in leg['maneuvers']:
+              out['routes'][-1]['legs'][-1]['maneuvers'].append({'length': man['length'], 'time': man['time'], 'instruction': man['instruction']})
         out['performance'] = {'response_time': elapsed}
         f.write('%s' % json.dumps(out, sort_keys=True, indent=2))
 
@@ -120,3 +123,13 @@ if __name__ == "__main__":
         progress = int(next_progress / increment)
     if progress != 100 / increment:
       print('100%')
+
+    # print total duration
+    if parsed_args.format == "json":
+      output_dir_path = Path(parsed_args.output_dir)
+      duration = 0
+      for out_f in output_dir_path.iterdir():
+        if out_f.is_file() and out_f.suffix == ".json":
+          with open(out_f) as f:
+            duration += json.load(f)["performance"]["response_time"]
+      print(f"Run took {duration} seconds")
--- a/src/thor/astar_bss.cc
+++ b/src/thor/astar_bss.cc
@@ -30,8 +30,7 @@ AStarBSSAlgorithm::AStarBSSAlgorithm(const boost::property_tree::ptree& config)
     : PathAlgorithm(config.get<uint32_t>("max_reserved_labels_count_astar",
                                          kInitialEdgeLabelCountAstar),
                     config.get<bool>("clear_reserved_memory", false)),
-      max_label_count_(std::numeric_limits<uint32_t>::max()), mode_(travel_mode_t::kDrive),
-      travel_type_(0) {
+      mode_(travel_mode_t::kDrive), travel_type_(0) {
 }
 
 // Destructor
@@ -287,11 +286,6 @@ AStarBSSAlgorithm::GetBestPath(valhalla::Location& origin,
       (*interrupt)();
     }
     total_labels = current_labels;
-
-    // Abort if max label count is exceeded
-    if (total_labels > max_label_count_) {
-      return {};
-    }
 
     // Get next element from adjacency list. Check that it is valid. An
     // invalid label indicates there are no edges that can be expanded.
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -107,8 +107,8 @@ void BidirectionalAStar::Init(const PointLL& origll, const PointLL& destll) {
 
   // Reserve size for edge labels - do this here rather than in constructor so
   // to limit how much extra memory is used for persistent objects
-  edgelabels_forward_.reserve(kInitialEdgeLabelCountBidirAstar);
-  edgelabels_reverse_.reserve(kInitialEdgeLabelCountBidirAstar);
+  edgelabels_forward_.reserve(max_reserved_labels_count_);
+  edgelabels_reverse_.reserve(max_reserved_labels_count_);
 
   // Construct adjacency list and initialize edge status lookup.
   // Set bucket size and cost range based on DynamicCost.
--- a/src/thor/costmatrix.cc
+++ b/src/thor/costmatrix.cc
@@ -279,7 +279,7 @@ void CostMatrix::Initialize(
   // Add initial sources status
   source_status_.reserve(source_count_);
   source_hierarchy_limits_.reserve(source_count_);
-  source_adjacency_.resize(source_count_);
+  source_adjacency_.reserve(source_count_);
   source_edgestatus_.resize(source_count_);
   source_edgelabel_.resize(source_count_);
   for (uint32_t i = 0; i < source_count_; i++) {
@@ -295,7 +295,7 @@ void CostMatrix::Initialize(
   // Add initial targets status
   target_status_.reserve(target_count_);
   target_hierarchy_limits_.reserve(target_count_);
-  target_adjacency_.resize(target_count_);
+  target_adjacency_.reserve(target_count_);
   target_edgestatus_.resize(target_count_);
   target_edgelabel_.resize(target_count_);
   for (uint32_t i = 0; i < target_count_; i++) {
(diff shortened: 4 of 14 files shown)


run_route_scripts/run_with_server.py
```diff
@@ -59,11 +60,13 @@ def make_request(post_body):
...
-          out['routes'][-1]['legs'].append({'maneuvers':[]})
-          for man in leg['maneuvers']:
-            out['routes'][-1]['legs'][-1]['maneuvers'].append({'length': man['length'], 'time': man['time'], 'instruction': man['instruction']})
+        out = dict()
+        if args.url.endswith("route"):
+          out['routes'] = [{'legs':[]}]
+          for leg in response['trip']['legs']:
+            out['routes'][-1]['legs'].append({'maneuvers':[]})
+            for man in leg['maneuvers']:
+              out['routes'][-1]['legs'][-1]['maneuvers'].append({'length': man['length'], 'time': man['time'], 'instruction': man['instruction']})
```
nilsnolde, line 69: allow for other endpoints

run_route_scripts/run_with_server.py
```diff
@@ -120,3 +123,13 @@ def make_request(post_body):
         progress = int(next_progress / increment)
     if progress != 100 / increment:
       print('100%')
+
+    # print total duration
+    if parsed_args.format == "json":
```
nilsnolde, line 128: just gimme the time

src/thor/bidirectional_astar.cc
```diff
@@ -107,8 +107,8 @@ void BidirectionalAStar::Init(const PointLL& origll, const PointLL& destll) {
 
   // Reserve size for edge labels - do this here rather than in constructor so
   // to limit how much extra memory is used for persistent objects
-  edgelabels_forward_.reserve(kInitialEdgeLabelCountBidirAstar);
-  edgelabels_reverse_.reserve(kInitialEdgeLabelCountBidirAstar);
+  edgelabels_forward_.reserve(max_reserved_labels_count_);
```
nilsnolde, line 110: I'd use the user provided default also in initialization

src/thor/costmatrix.cc
```diff
@@ -78,40 +80,49 @@ float CostMatrix::GetCostThreshold(const float max_matrix_distance) {
...
+  }
+  if (target_count_ > kMaxLocationCache) {
+    target_edgelabel_.resize(kMaxLocationCache);
+    target_edgelabel_.shrink_to_fit();
+    target_adjacency_.resize(kMaxLocationCache);
+    target_adjacency_.shrink_to_fit();
+    target_edgestatus_.resize(kMaxLocationCache);
+    target_edgestatus_.shrink_to_fit();
   }
-  targets_->clear();
```
nilsnolde, line 87: this is actually what we thought to rectify in https://github.com/valhalla/valhalla/pull/4329#discussion_r1373217995, but I realized it won't work: `targets_` is a robin_hood::unordered_map which didn't implement any `clear()` or anything else. I kept the `targets_.clear()` further down. however, what the old code cleared were the tiny vectors of recorded targets _for every edge_, which could've been a few million and in any case, calling `targets_.clear()` will destruct all those vectors just the line below AFAIK.

nilsnolde: btw, robin_hood is deprecated! yay..

src/thor/costmatrix.cc
```diff
@@ -78,40 +80,49 @@ float CostMatrix::GetCostThreshold(const float max_matrix_distance) {
...
+    source_edgestatus_.resize(kMaxLocationCache);
+    source_edgestatus_.shrink_to_fit();
+  }
+  if (target_count_ > kMaxLocationCache) {
+    target_edgelabel_.resize(kMaxLocationCache);
+    target_edgelabel_.shrink_to_fit();
+    target_adjacency_.resize(kMaxLocationCache);
+    target_adjacency_.shrink_to_fit();
+    target_edgestatus_.resize(kMaxLocationCache);
+    target_edgestatus_.shrink_to_fit();
```
nilsnolde, line 98: I chose to shrink all containers to 20 locations each after each request. so the server isn't severely hurt if someone decides to send off a 1000 location request.

src/thor/costmatrix.cc
```diff
@@ -78,40 +80,49 @@ float CostMatrix::GetCostThreshold(const float max_matrix_distance) {
...
-  target_edgelabel_.clear();
-  target_edgelabel_.resize(0);
-  target_edgelabel_.shrink_to_fit();
-  source_edgestatus_.clear();
-  source_edgestatus_.resize(0);
-  source_edgestatus_.shrink_to_fit();
-  target_edgestatus_.clear();
-  target_edgestatus_.resize(0);
-  target_edgestatus_.shrink_to_fit();
+  auto reservation = clear_reserved_memory_ ? 0U : max_reserved_labels_count_;
```
nilsnolde, line 101: and for those 20 locations, we keep a default of 2 Mio around per location. for labels that should be around 20MB per location.

src/thor/costmatrix.cc
```diff
@@ -78,40 +80,49 @@ float CostMatrix::GetCostThreshold(const float max_matrix_distance) {
...
+      iter.shrink_to_fit();
+    }
+  }
+
+  for (auto& iter : source_adjacency_) {
+    iter.clear();
+  }
+  for (auto& iter : target_adjacency_) {
+    iter.clear();
+  }
```
nilsnolde, line 122: and clear all adj lists, but don't destroy them

src/valhalla_run_matrix.cc
```diff
@@ -45,20 +45,23 @@ std::string GetFormattedTime(uint32_t secs) {
...
-  for (uint32_t i = 0; i < matrix.times().size(); i++) {
-    auto distance = matrix.distances().Get(i);
-    LOG_INFO(std::to_string(idx1) + "," + std::to_string(idx2) + ": Distance= " +
-             std::to_string(distance) + " Time= " + GetFormattedTime(matrix.times().Get(i)) +
-             " secs = " + std::to_string(distance));
-    idx2++;
-    if (idx2 == nlocs) {
-      idx2 = 0;
-      idx1++;
+                const valhalla::Matrix& matrix,
```
nilsnolde, line 48: this is from #4356 , never mind (or approve it here and I'll close the other one)

src/thor/costmatrix.cc
nilsnolde: 2 Mio is quite a lot.. but also it's dijkstra.. well, it's configurable for the server

src/thor/astar_bss.cc
```diff
@@ -30,8 +30,7 @@ AStarBSSAlgorithm::AStarBSSAlgorithm(const boost::property_tree::ptree& config)
     : PathAlgorithm(config.get<uint32_t>("max_reserved_labels_count_astar",
                                          kInitialEdgeLabelCountAstar),
                     config.get<bool>("clear_reserved_memory", false)),
-      max_label_count_(std::numeric_limits<uint32_t>::max()), mode_(travel_mode_t::kDrive),
-      travel_type_(0) {
+      mode_(travel_mode_t::kDrive), travel_type_(0) {
```
nilsnolde, line 33: this was pointless, it was never set to anything else than max(uint32_t)

src/thor/dijkstras.cc
```diff
@@ -78,7 +78,7 @@ void Dijkstras::Initialize(label_container_t& labels,
   uint32_t edge_label_reservation;
   uint32_t bucket_count;
   GetExpansionHints(bucket_count, edge_label_reservation);
-  labels.reserve(std::min(max_reserved_labels_count_, edge_label_reservation));
+  labels.reserve(max_reserved_labels_count_);
```
nilsnolde, line 81: forgot to remove the std::min here last time

src/thor/multimodal.cc
```diff
@@ -43,8 +43,7 @@ MultiModalPathAlgorithm::MultiModalPathAlgorithm(const boost::property_tree::ptr
     : PathAlgorithm(config.get<uint32_t>("max_reserved_labels_count_astar",
                                          kInitialEdgeLabelCountAstar),
                     config.get<bool>("clear_reserved_memory", false)),
-      max_walking_dist_(0), max_label_count_(std::numeric_limits<uint32_t>::max()),
-      mode_(travel_mode_t::kPedestrian), travel_type_(0) {
+      max_walking_dist_(0), mode_(travel_mode_t::kPedestrian), travel_type_(0) {
```
nilsnolde, line 46: same here, the max_label_count_ is pointless

src/thor/unidirectional_astar.cc
```diff
@@ -584,7 +584,7 @@ void UnidirectionalAStar<expansion_direction, FORWARD>::Init(const midgard::Poin
     astarheuristic_.Init(origll, costing_->AStarCostFactor());
     mincost = astarheuristic_.Get(destll);
   }
-  edgelabels_.reserve(std::min(max_reserved_labels_count_, kInitialEdgeLabelCountAstar));
+  edgelabels_.reserve(max_reserved_labels_count_);
```
nilsnolde, line 587: forgot to remove the std::min here last time

valhalla/thor/timedistancematrix.h
```diff
@@ -79,7 +79,7 @@ class TimeDistanceMatrix {
   inline void clear() {
     auto reservation = clear_reserved_memory_ ? 0 : max_reserved_labels_count_;
     if (edgelabels_.size() > reservation) {
-      edgelabels_.resize(max_reserved_labels_count_);
+      edgelabels_.resize(reservation);
```
nilsnolde, line 82: another typo

src/thor/costmatrix.cc
kevinkreiser: yeah saw that a while ago. the guy made a new project though that is he actively maintaining. honestly though who cares? unlike javascript which is subject ot loads of dependency injection rotting your code over time, a header only c++ library is unlikely to stop working any time soon. im not worried about it in other words :smile:

kevinkreiser: > calling targets_.clear() will destruct all those vectors just the line below AFAIK.

`clear` will simply set the current end pointer of the container back to the beginning AND call the destructor on all the elements which are now past the end pointer but still in the containers allocated space

src/thor/costmatrix.cc
```diff
@@ -78,40 +80,49 @@ float CostMatrix::GetCostThreshold(const float max_matrix_distance) {
...
+  targets_->clear();
```
nilsnolde, line 123: maybe do resize, check memory & perf

nilsnolde:
mostly superseded by #4372

Issue #4241, opened 2023-08-06
timezone info in response `date_time`


nilsnolde:
It'd be nicer IMO if we'd return the timezone info in the response's `date_time` strings. I didn't look into it too much, not sure at which point it'd be safe to add the tz info for /route output, for matrix it'd be fine. I know local time is good enough for navigation, but it's useful to be able to easily infer the timezone as well from the response and I can't see that it'd be a hassle to implement or any other disadvantage.

Any thoughts?

nilsnolde:
chatting with @kevinkreiser he'd prefer to have this as a separate response field. totally agree there.

PR #4365, opened 2023-10-28
refactor datetime output 

### LINKED ISSUE #4360: Datetime in result elements of sources to targets are in UTC not local as documentation states
The documentation [https://valhalla.github.io/valhalla/api/matrix/api-reference/](url) states that the date_time parameter in the result elements of sources_to_targets is in local time.

> (optional) If the date_time was valid for an origin, date_time will return the local time at the destination.

 However, the results appear to be in UTC. When I use the date_time specified on the source, it should indicate the departure time from that source in local time. In the example below, I use "2023-11-27T10:00,". The result element shows a time of 1662 seconds, which is equivalent to 27,7 minutes. So, I would expect the date_time value on element to be "2023-11-27T10:28," but it displays "2023-10-27T09:28." The coordinates timezone is UTC+1, which seems to cause a 1-hour shift. If we use same request for current month (UTC+2 daylight saving) result is "2023-10-27T08:28".

Request

```json
{
    "sources": [
        {
            "lat": 50.067278, 
            "lon": 14.546463,
            "date_time": "2023-11-27T10:00"
        }
    ],
    "targets": [
         {
            "lat": 50.095831,
            "lon": 14.345975
        }
    ],
    "verbose": true,
        "costing": "auto",
	"costing_options": {
	    "auto": {
	        "use_tolls": 1,
		"ignore_restrictions": true
	    }
	},
    "prioritize_bidirectional": true
}
```

Result
```json
{
    "algorithm": "costmatrix",
    "units": "kilometers",
    "sources": [
        [
            {
                "lon": 14.546463,
                "lat": 50.067278
            }
        ]
    ],
    "targets": [
        [
            {
                "lon": 14.345975,
                "lat": 50.095831
            }
        ]
    ],
    "sources_to_targets": [
        [
            {
                "date_time": "2023-11-27T09:28",
                "distance": 23.927,
                "time": 1662,
                "to_index": 0,
                "from_index": 0
            }
        ]
    ]
}
```

nilsnolde:
Route & matrix didn't share the code to generate the response's datetime strings. Can't remember why I didn't do that in https://github.com/valhalla/valhalla/pull/4071. I thought I'll be smarter with my PRs for once, and not cram refactors together with functional changes:) Comes from the experience of reviewing such PRs I guess (not pointing fingers!). 

Really, it's to prepare for outputting timezone-aware datetime strings: https://github.com/valhalla/valhalla/issues/4241. Last chance to tell me that no one wants that:) It's a breaking change the way I'd do it (i..e. just return `2023-02-27T22:50+02:00` instead of `2023-02-27T22:50`). Depending on the client code that could be a problem.. IMO it's acceptable.

EDIT: Also fixes #4360 as collateral damage..


src/baldr/datetime.cc
```diff
@@ -524,6 +524,14 @@ uint32_t second_of_week(uint32_t epoch_time, const date::time_zone* time_zone) {
...
 }
 
+std::string
+offset_date(const std::string& in_dt, const uint32_t in_tz, const uint32_t out_tz, float offset) {
+  // get the input UTC time, add the offset and translate to the out timezone
+  auto iepoch = DateTime::seconds_since_epoch(in_dt, DateTime::get_tz_db().from_index(in_tz));
+  auto oepoch =
+      static_cast<uint64_t>(static_cast<double>(iepoch) + static_cast<double>(offset + .5f));
+  return DateTime::seconds_to_date(oepoch, DateTime::get_tz_db().from_index(out_tz), false);
+}
```
nilsnolde, line 534: I broke it up a bit, so we can put this here in baldr/date_time without including GraphReader too. I added a function to GraphReader instead.

src/baldr/graphreader.cc
```diff
@@ -975,6 +975,16 @@ int GraphReader::GetTimezone(const baldr::GraphId& node, graph_tile_ptr& tile) {
...
 
+int GraphReader::GetTimezoneFromEdge(const baldr::GraphId& edge, graph_tile_ptr& tile) {
+  auto nodes = GetDirectedEdgeNodes(edge, tile);
+  if (const auto* node = nodeinfo(nodes.first, tile))
+    return node->timezone();
+  else if (const auto* node = nodeinfo(nodes.second, tile))
+    return node->timezone();
+
+  return 0;
+}
```
nilsnolde, line 986: analogous to GraphReader::GetTimezone for nodes

src/thor/costmatrix.cc
```diff
@@ -233,23 +234,31 @@ void CostMatrix::SourceToTarget(Api& request,
...
   reserve_pbf_arrays(matrix, best_connection_.size());
   for (const auto& connection : best_connection_) {
     uint32_t target_idx = count % target_location_list.size();
     uint32_t origin_idx = count / target_location_list.size();
-    float time = connection.cost.secs + .5f;
-    auto date_time = get_date_time(source_location_list[origin_idx].date_time(),
-                                   time_infos[origin_idx].timezone_index,
-                                   target_edgelabel_[target_idx].front().edgeid(), graphreader,
-                                   static_cast<uint64_t>(time));
+    float time = connection.cost.secs;
```
nilsnolde, line 244: adding `0.5f` makes no sense here

src/thor/costmatrix.cc
```diff
@@ -233,23 +234,31 @@ void CostMatrix::SourceToTarget(Api& request,
...
+    // no datetime for same locations or unfound connections
+    if (time && time < kMaxCost) {
```
nilsnolde, line 246: only add date_time if it's not 0 and the connection was found

src/thor/costmatrix.cc
```diff
@@ -233,23 +234,31 @@ void CostMatrix::SourceToTarget(Api& request,
...
     uint32_t target_idx = count % target_location_list.size();
     uint32_t origin_idx = count / target_location_list.size();
-    float time = connection.cost.secs + .5f;
-    auto date_time = get_date_time(source_location_list[origin_idx].date_time(),
-                                   time_infos[origin_idx].timezone_index,
-                                   target_edgelabel_[target_idx].front().edgeid(), graphreader,
-                                   static_cast<uint64_t>(time));
+    float time = connection.cost.secs;
+    // no datetime for unfound connections
+    if (time < kMaxCost) {
```
nilsnolde, line 246: we don't add date_time when there's no connection, but we do add it for origin == target. in the serializer we put `null` when there was no connection

src/tyr/matrix_serializer.cc
```diff
@@ -12,6 +12,24 @@ using namespace valhalla::thor;
 
 namespace {
 
+json::ArrayPtr serialize_datetime(const valhalla::Matrix& matrix,
```
nilsnolde, line 15: I realized there was no date_time array for a "slim" matrix. The change is contained only in this file (+ test), so I figured it's fine to cram in here

src/tyr/matrix_serializer.cc
```diff
@@ -150,16 +168,24 @@ std::string serialize(const Api& request, double distance_scale) {
...
                              options.targets_size(), source_index, 0, distance_scale));
+      date_time->emplace_back(serialize_datetime(request.matrix(),
+                                                 source_index * options.targets_size(),
+                                                 options.targets_size(), has_dt));
     }
     matrix->emplace("distances", distance);
     matrix->emplace("durations", time);
+    if (has_dt) {
+      matrix->emplace("date_times", date_time);
+    }
```
nilsnolde, line 188: we only put the date_times array if at least one connection had a date_time

src/thor/costmatrix.cc
nilsnolde: in fact this was a bug some clients were complaining about, where matrix had a one second offset to /route. but not always. now I know why haha

valhalla/thor/matrix_common.h
```diff
@@ -56,36 +56,6 @@ struct Destination {
...
-    return "";
-  } else if (!offset) {
-    return origin_dt;
-  }
-  graph_tile_ptr tile = nullptr;
-  uint32_t dest_tz = 0;
-  if (pred_id.Is_Valid()) {
-    // get the timezone of the output location
-    auto out_nodes = reader.GetDirectedEdgeNodes(pred_id, tile);
-    dest_tz = reader.GetTimezone(out_nodes.first, tile) || reader.GetTimezone(out_nodes.second, tile);
```
nilsnolde, line 75: I just realized, this PR will also fix #4360 . this line is not what python-eyes think it is.. `dest_tz` will be true or false, i.e. 1 or 0. And 1 must be UTC. The new logic doesn't have that flaw.

nilsnolde:
I drafted this bcs we agreed to output the timezone info as an extra response field. I’d just put the „+02:00“ in an extra field.

valhalla/baldr/graphreader.h
```diff
@@ -947,6 +947,15 @@ class GraphReader {
...
   int GetTimezone(const baldr::GraphId& node, graph_tile_ptr& tile);
 
+  /**
+   * Convenience method to get the timezone index from an edge. Preferably it returns
+   * the start's node's timezone.
+   * @param edge   GraphId of the edge to get the timezone index.
+   * @param tile   Current tile.
+   * @return Returns the timezone index. A value of 0 indicates an invalid timezone.
+   */
+  int GetTimezoneFromEdge(const baldr::GraphId& edge, graph_tile_ptr& tile);
```
kevinkreiser, line 957: just need to check and mention if the tile is modified. if its not modified we should pass it wihtout the reference if modified we should say whether its for the end or begin node of the edge

nilsnolde: you're right, in `nodeinfo()` the tile can change, so we could end up with the edge's end node's tile instead. is a comment in the docstring good enough for you or you want smth more elaborate?

kevinkreiser:
the history and subsequently the diff of this branch are screwed up.

nilsnolde:
fixed. btw, regarding your comment if the tile is being switched in the new `GetTimezoneFromEdge` function, I added that as info to the docstring. is that good enough for you?

nilsnolde:
@chrstnbwnkl will PR tmrw or so with the next bit to actually return the tz offset in the response:)

valhalla/baldr/graphreader.h
kevinkreiser: yeah just docstring update was all i wanted. i hope that its a gaurantee though that as a post condition it will be the endnodes tile.

nilsnolde: Fwiw, our own usage of this (so far) is safe, we don’t even use the tile, only for caching

kevinkreiser: yep yep its just to avoid mistakes

nilsnolde: > i hope that its a gaurantee though that as a post condition it will be the endnodes tile not just could be, i guess when the tile isnt found it could be, thats ok.

after this function is called, it's very unlikely that the tile is changed (unless you compiled the graph without timezone info, but maybe that's even handled further up in the processing chain). that only happens if the edge's start node has no timezone info.

kevinkreiser (approved):

PR #4335, opened 2023-10-16
respond with correlated edges for format=valhalla and matrix

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4333

based on #4334

the whole matrix serialization is quite a mess tbh..

PR #4366, opened 2023-10-28
few matrix serializer fixes

nilsnolde:
I should've walked away when I had the chance.. mostly I'm to blame for these fixes here:
- ~~add a `date_times` array to the response of a "slim" matrix (i.e. `verbose=false`) in case there was any datetime~~ I'll revert that for now. I had complaints about bloating the "slim" matrix from clients. 
- revert some of https://github.com/valhalla/valhalla/pull/4335: the `sources` & `targets` arrays in the response contained all correlated points of all edges and of course tons of duplicates, so here I change it to only return the first correlated edge


src/tyr/matrix_serializer.cc
```diff
@@ -12,6 +12,23 @@ using namespace valhalla::thor;
 
 namespace {
 
+json::ArrayPtr serialize_datetime(const valhalla::Matrix& matrix,
```
nilsnolde, line 15: serialize date times for the slim matrix

src/tyr/matrix_serializer.cc
```diff
@@ -89,10 +106,9 @@ json::ArrayPtr locations(const google::protobuf::RepeatedPtrField<valhalla::Loca
     if (location.correlation().edges().size() == 0) {
       input_locs->emplace_back(nullptr);
     } else {
-      for (const auto& corr_edge : location.correlation().edges()) {
-        input_locs->emplace_back(json::map({{"lat", json::fixed_t{corr_edge.ll().lat(), 6}},
-                                            {"lon", json::fixed_t{corr_edge.ll().lng(), 6}}}));
-      }
+      auto& corr_ll = location.correlation().edges(0).ll();
+      input_locs->emplace_back(json::map(
+          {{"lat", json::fixed_t{corr_ll.lat(), 6}}, {"lon", json::fixed_t{corr_ll.lng(), 6}}}));
```
nilsnolde, line 111: changing it to only one correlated edge in the `sources` & `targets` response arrays

PR #4372, opened 2023-10-31
make costmatrix handle uturns

### LINKED ISSUE #4371: CostMatrix can't handle deadends
I can't give an example, this popped up in a TomTom graph: with the default radius it can happen that we only correlate a single edge and that edge is connected to:
- an edge it can't go on because of a turn restriction
- another edge which eventually ends up in a deadend

The bidir A* does it right and allows a uturn at the deadend so it can find its way back and take the other edge, but the CostMatrix doesn't. Looking at it, I might do a refactor first, before handling deadends. Interestingly TDMatrix computes the route in question, even though I can't see anything handling deadends there either. Might point to another bug in TDMatrix. :rabbit: :hole:

nilsnolde:
fixes #4371 

only for myself for now..


src/thor/astar_bss.cc
```diff
@@ -29,9 +29,7 @@ constexpr uint32_t kMaxIterationsWithoutConvergence = 200000;
 AStarBSSAlgorithm::AStarBSSAlgorithm(const boost::property_tree::ptree& config)
     : PathAlgorithm(config.get<uint32_t>("max_reserved_labels_count_astar",
                                          kInitialEdgeLabelCountAstar),
-                    config.get<bool>("clear_reserved_memory", false)),
-      max_label_count_(std::numeric_limits<uint32_t>::max()), mode_(travel_mode_t::kDrive),
```
nilsnolde, line 33: this is a frequent change in this PR, `max_label_count_` must've been used at some point as termination criteria and was copy/pasted along. however, it's not used anywhere

src/thor/astar_bss.cc
```diff
@@ -278,20 +276,12 @@ AStarBSSAlgorithm::GetBestPath(valhalla::Location& origin,
   uint32_t nc = 0; // Count of iterations with no convergence
                    // towards destination
   std::pair<int32_t, float> best_path = std::make_pair(-1, 0.0f);
-  size_t total_labels = 0;
+  size_t n = 0;
   while (true) {
     // Allow this process to be aborted
-    size_t current_labels = edgelabels_.size();
```
nilsnolde, line 284: just a bit more succinct

src/thor/bidirectional_astar.cc
```diff
@@ -107,8 +107,8 @@ void BidirectionalAStar::Init(const PointLL& origll, const PointLL& destll) {
 
   // Reserve size for edge labels - do this here rather than in constructor so
   // to limit how much extra memory is used for persistent objects
-  edgelabels_forward_.reserve(kInitialEdgeLabelCountBidirAstar);
-  edgelabels_reverse_.reserve(kInitialEdgeLabelCountBidirAstar);
+  edgelabels_forward_.reserve(max_reserved_labels_count_);
```
nilsnolde, line 110: ze jusch

src/thor/bidirectional_astar.cc
```diff
@@ -136,23 +136,14 @@ void BidirectionalAStar::Init(const PointLL& origll, const PointLL& destll) {
...
-  // Set this flag to 'true' if we can expand edges at all hierarchy levels without limits
-  ignore_hierarchy_limits_ = ignore_forward_limits && ignore_reverse_limits;
+  auto& hierarchy_limits = costing_->GetHierarchyLimits();
+  ignore_hierarchy_limits_ = std::all_of(hierarchy_limits.begin() + 1,
+                                         hierarchy_limits.begin() + TileHierarchy::levels().size(),
+                                         [](const HierarchyLimits& limits) {
+                                           return limits.max_up_transitions == kUnlimitedTransitions;
+                                         });
+  hierarchy_limits_forward_ = hierarchy_limits;
+  hierarchy_limits_reverse_ = hierarchy_limits;
```
nilsnolde, line 146: `hierarchy_limits_forward_` and `hierarchy_limits_reverse_` are the exact same at this point, enough to do this whole thing once

src/thor/bidirectional_astar.cc
```diff
@@ -367,7 +358,7 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
 }
 
 template <const ExpansionType expansion_direction>
-bool BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
+void BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
```
nilsnolde, line 361: oversight I guess, it's not using the return value anywhere

src/thor/bidirectional_astar.cc
```diff
@@ -581,8 +569,7 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
         if (opp_status.set() == EdgeSet::kPermanent ||
             (opp_status.set() == EdgeSet::kTemporary &&
              edgelabels_reverse_[opp_status.index()].predecessor() == kInvalidLabel)) {
-          if (SetForwardConnection(graphreader, fwd_pred) &&
-              opp_status.set() == EdgeSet::kPermanent) {
+          if (SetForwardConnection(graphreader, fwd_pred)) {
```
nilsnolde, line 572: this seems like a bug to me. in the line above we're explicitly allowing `kTemporary` under the condition that the reverse tree is exhausted (which I think makes sense or at least made sense when I changed it 2 weeks ago), but then one line below we're again not allowing `kTemporary`. we should choose one way to do it, so this is just a best guess, I can also remove the 2-3 lines above instead

src/thor/bidirectional_astar.cc
```diff
@@ -677,6 +663,7 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
     // the origin and destination and provides valid conditions for the reach-based pruning.
     bool force_forward = false;
     bool force_reverse = false;
+    // TODO: this is forcing if _any_ level is different, it doesn't distinguish levels
```
nilsnolde, line 666: meeh, will remove, not sure anymore if this a todo or not and I'm a bit sick of thinking about algos for now haha

src/thor/costmatrix.cc
```diff
@@ -16,6 +16,7 @@ namespace {
 
 constexpr uint32_t kMaxMatrixIterations = 2000000;
 constexpr uint32_t kMaxThreshold = std::numeric_limits<int>::max();
+constexpr uint32_t kMaxLocationCache = 20;
```
nilsnolde, line 19: I'm actually thinking about making this configurable. It goes a loooong way to speed up the computations and for people who are using Valhalla for matrix computations mainly, it makes sense to reserve as much memory as possible.

src/thor/costmatrix.cc
```diff
@@ -75,48 +75,6 @@ float CostMatrix::GetCostThreshold(const float max_matrix_distance) {
   return cost_threshold * 2.0f;
 }
 
-// Clear the temporary information generated during time + distance matrix
-// construction.
-void CostMatrix::clear() {
```
nilsnolde, line 80: todo: put this back again at its previous place

src/thor/costmatrix.cc
```diff
@@ -260,559 +238,528 @@ void CostMatrix::SourceToTarget(Api& request,
...
     }
   }
-  remaining_targets_ = 0;
-  for (const auto& t : target_status_) {
-    if (!t.remaining_locations.empty()) {
-      remaining_targets_++;
+  locs_remaining_[MATRIX_REV] = 0;
+  for (const auto& t : locs_status_[MATRIX_REV]) {
+    if (!t.unfound_connections.empty()) {
+      locs_remaining_[MATRIX_REV]++;
```
nilsnolde, line 301: up to here it's mostly variable renaming, hardly any logic change

src/thor/costmatrix.cc
```diff
@@ -260,559 +238,528 @@ void CostMatrix::SourceToTarget(Api& request,
...
+    const bool is_uturn = pred.opp_local_idx() == meta.edge->localedgeidx();
+    uturn_meta = is_uturn ? meta : uturn_meta;
+
+    // Expand but only if this isnt the uturn, we'll try that later if nothing else works out
+    disable_uturn =
+        (!is_uturn &&
+         ExpandInner<expansion_direction>(graphreader, index, pred, opp_pred_edge, nodeinfo, pred_idx,
+                                          meta, shortcuts, tile, offset_time)) ||
+        disable_uturn;
+  }
```
nilsnolde, line 581: this is whole motivation for this PR: properly handle uturns. should look familiar :)

src/thor/costmatrix.cc
```diff
@@ -1065,5 +1015,49 @@ void CostMatrix::RecostPaths(GraphReader& graphreader,
...
+  for (const auto exp_dir : {MATRIX_FORW, MATRIX_REV}) {
+    // resize all relevant structures down to 20 locations
+    if (locs_count_[exp_dir] > kMaxLocationCache) {
+      edgelabel_[exp_dir].resize(kMaxLocationCache);
+      edgelabel_[exp_dir].shrink_to_fit();
+      adjacency_[exp_dir].resize(kMaxLocationCache);
+      adjacency_[exp_dir].shrink_to_fit();
+      edgestatus_[exp_dir].resize(kMaxLocationCache);
+      edgestatus_[exp_dir].shrink_to_fit();
+    }
```
nilsnolde, line 1041: and finally the reason why this PR improves performance to up to 30%

valhalla/thor/matrix_common.h
```diff
@@ -72,7 +72,9 @@ inline std::string get_date_time(const std::string& origin_dt,
   if (pred_id.Is_Valid()) {
     // get the timezone of the output location
     auto out_nodes = reader.GetDirectedEdgeNodes(pred_id, tile);
-    dest_tz = reader.GetTimezone(out_nodes.first, tile) || reader.GetTimezone(out_nodes.second, tile);
+    dest_tz = reader.GetTimezone(out_nodes.first, tile);
```
nilsnolde, line 75: ah right, this fixes #4360

nilsnolde: in my defense, python damaged my mind a bit

src/thor/astar_bss.cc
GrigoryDobrov: Maybe I don't understand smth but it was used here https://github.com/valhalla/valhalla/pull/4372/files/c322b77e3eab02d1b60a0dee449be36575ac8a22#diff-acd5a91be66009c2563642ad11eed627f663a49e0fab12b2ec2c8fbce77c1702L292

valhalla/thor/costmatrix.h
```diff
@@ -131,43 +146,37 @@ class CostMatrix {
...
-  std::vector<baldr::DoubleBucketQueue<sif::BDEdgeLabel>> target_adjacency_;
-  std::vector<std::vector<sif::BDEdgeLabel>> target_edgelabel_;
-  std::vector<EdgeStatus> target_edgestatus_;
+  std::array<std::vector<LocationStatus>, 2> locs_status_;
+
+  // Adjacency lists, EdgeLabels, EdgeStatus, and hierarchy limits for each location
+  std::array<std::vector<std::vector<sif::HierarchyLimits>>, 2> hierarchy_limits_;
+  std::array<std::vector<baldr::DoubleBucketQueue<sif::BDEdgeLabel>>, 2> adjacency_;
+  std::array<std::vector<std::vector<sif::BDEdgeLabel>>, 2> edgelabel_;
+  std::array<std::vector<EdgeStatus>, 2> edgestatus_;
```
GrigoryDobrov, line 165: I know Structure of arrays concept but is it profitable here?
Why not to create a struct like:
```
struct SearchWaveState {
  std::vector<LocationStatus> status_;
  std::vector<std::vector<sif::HierarchyLimits>> hierarchy_limits_;
  std::vector<baldr::DoubleBucketQueue<sif::BDEdgeLabel>> adjacency_;
 std::vector<EdgeStatus> edgestatus_;
};
SearchWaveState forward_wave_, reverse_wave_; //or if you prefer array<SearchWaveState,2>;
```
I think lots of accesses by index will be eliminated and it will simplify code. For example, in `CostMatrix::ExpandInner`  you can get SearchWaveState once in the beginning of the function.

src/thor/costmatrix.cc
```diff
@@ -239,10 +217,10 @@ void CostMatrix::SourceToTarget(Api& request,
   for (const auto& connection : best_connection_) {
     uint32_t target_idx = count % target_location_list.size();
     uint32_t origin_idx = count / target_location_list.size();
-    float time = connection.cost.secs + .5f;
+    float time = connection.cost.secs;
```
nilsnolde, line 220: this was another bug, we shouldn't add 0.5 at this point.. also reflected in the tests

valhalla/thor/costmatrix.h
```diff
@@ -131,43 +146,37 @@ class CostMatrix {
...
 
   // List of best connections found so far
   std::vector<BestCandidate> best_connection_;
 
+  bool ignore_hierarchy_limits_;
```
nilsnolde, line 170: we really have to be more strict when we implement smth for one algorithm, to not leave out all others. at the very least, open an issue or even better, create a "consistency" project or so.

src/thor/astar_bss.cc
nilsnolde: right, I should've said "not being effectively used anywhere". you can see your highlighted section is removed too, there was no point for this variable

valhalla/thor/costmatrix.h
nilsnolde: yeah, I did this pretty much towards the end of the whole refactor and I did consider your approach, albeit slightly differently than what you're doing (I like your way better though). if it's ok with everyone I'd leave a TODO with your comment and get to it another time, I'm currently pretty sick of this work right now and just want to get it merged ;) (and this is not at all critical)

nilsnolde:
ok so this should be final now from my side.

a few remarks (same as shared over slack):

### Quality

I generated 50 matrix requests with 20 (symmetrical) sources/targets each in Berlin area with randomized coords, which sums up to 20k routes. I diffed those results against master and each diff I compared against routing results, to decide whether it's a regression or an improvement. Eventually ended up with 11 diffs:
- 6 diffs were very good improvements due to deadends, e.g. 30 km (master) vs 83 km (this branch)
- 4 diffs were tiny differences (50 - 400 m/up to 30 seconds) which I deemed negligible; routing has the same problem with unidir/bidir A*
- 1 negative diff was due to a lower maximum expansion threshold, since we're now expanding a bit more with deadends; hence I increased that threshold by 3% and that diff went away

### Performance

On the Berlin dataset, contrary to my last night experiment where I thought I saw 30% performance gain, there's pretty much no difference compared to master. However, on bigger distances (i.e. more expansion), this PR should have a bigger impact (a client reported 6 - 7 %).

src/thor/costmatrix.cc
```diff
@@ -16,11 +16,12 @@ namespace {
 
 constexpr uint32_t kMaxMatrixIterations = 2000000;
 constexpr uint32_t kMaxThreshold = std::numeric_limits<int>::max();
+constexpr uint32_t kMaxLocationCache = 20;
 
 // Find a threshold to continue the search - should be based on
 // the max edge cost in the adjacency set?
 int GetThreshold(const travel_mode_t mode, const int n) {
-  return (mode == travel_mode_t::kDrive) ? std::min(2700, std::max(100, n / 3)) : 500;
+  return (mode == travel_mode_t::kDrive) ? std::min(2800, std::max(100, n / 3)) : 500;
```
nilsnolde, line 24: more expansion now, so we should increase the maximum a bit. at least I had a negative diff because of this. eventually we should do the TODO though maybe..

src/thor/astar_bss.cc
```diff
@@ -271,27 +270,19 @@ AStarBSSAlgorithm::GetBestPath(valhalla::Location& origin,
...
-  uint32_t density = SetDestination(graphreader, destination);
+  SetDestination(graphreader, destination);
   SetOrigin(graphreader, origin, destination);
 
   // Find shortest path
   uint32_t nc = 0; // Count of iterations with no convergence
                    // towards destination
   std::pair<int32_t, float> best_path = std::make_pair(-1, 0.0f);
-  size_t total_labels = 0;
+  size_t n = 0;
```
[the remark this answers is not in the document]
nilsnolde: @dnesbitt61 this is just for the `interrupt` actually, there's another `nc` variable that checks for max iterations and aborts with that.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
    * FIXED: recover proper shortest path to ferry connections (make sure correct label index is used) [#4378](https://github.com/valhalla/valhalla/pull/4378)
    * FIXED: Allow all roads for motorcycles [#4348](https://github.com/valhalla/valhalla/pull/4348)
    * FIXED: motorcar:conditional should not apply to motorcycle and moped [#4359](https://github.com/valhalla/valhalla/pull/4359)
+   * FIXED: lots of issues with CostMatrix (primarily deadend logic) with a complete refactor modeling things very close to bidir A*, also to prepare for a unification of the two [#4372](https://github.com/valhalla/valhalla/pull/4372)
```
kevinkreiser, line 19: nit: need to move this down one

run_route_scripts/run_with_server_matrix.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
```
kevinkreiser, line 1: do we actually need a separate script for this, the only thing that should really differ is the parsing of the response? the other script i think can let you run any file of requests so it shoudl work for the input part. and the output can be handled pretty easily by simply looking at the url to find which endpoint was hit route/sources_to_targets whatever and calling appropriate handler to parse the resulting json

run_route_scripts/run_with_server_matrix.py
```diff
@@ -0,0 +1,97 @@
...
+
+    return {"response": response, "line": post_body["id"], "perf": elapsed, "sources": post_body["sources"], "targets": post_body["targets"]}
+
+
+if __name__ == "__main__":
+    # parse some program arguments
+    parser = argparse.ArgumentParser(description="Runs payloads generated by gen_requests_matrix.py against a Valhalla server and writes the results to a single file for diffing purposes. Also prints out the total time.")
+    parser.add_argument('--test-file', type=Path, help='The file with the test requests', required=True)
+    parser.add_argument('--url', type=str, help='The url to which you want to POST the request bodies',
+                        default='http://localhost:8002/route')
```
kevinkreiser, line 59: case in point haha

run_route_scripts/run_with_server_matrix.py
kevinkreiser: i dont feel strongly about it though, but it feels to me like tech debt

src/thor/astar_bss.cc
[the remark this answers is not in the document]
kevinkreiser: hjaha both are aborts one is driving by the server side and one is driving by the client side!

src/thor/astar_bss.cc
```diff
@@ -271,27 +270,19 @@ AStarBSSAlgorithm::GetBestPath(valhalla::Location& origin,
...
-        total_labels / kInterruptIterationsInterval < current_labels / kInterruptIterationsInterval) {
+    if (interrupt && (++n % kInterruptIterationsInterval) == 0) {
       (*interrupt)();
     }
-    total_labels = current_labels;
-
-    // Abort if max label count is exceeded
-    if (total_labels > max_label_count_) {
-      return {};
-    }
```
kevinkreiser, line 286: but dont we still want this? as i was saying there are two types of aborting, using the interrupt which is telling us the client said hey i dont care about this request anymore, and then the other kind where we say ok we searched hard enough that we now give up

src/thor/bidirectional_astar.cc
kevinkreiser: just wondering why are you resolving these? if they add context to the pr review its better to see them

src/thor/bidirectional_astar.cc
kevinkreiser: so i think i implemented deadend detection before we did another refactor to add expandinner. this probably should ahve been done when expandinner became a thing

src/thor/bidirectional_astar.cc
```diff
@@ -403,11 +394,12 @@ bool BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
     // is labelled
     pred.set_deadend(true);
     // Check if edge is null before using it (can happen with regional data sets)
-    return opp_edge &&
-           ExpandInner<expansion_direction>(graphreader, pred, opp_pred_edge, nodeinfo, pred_idx,
-                                            {opp_edge, opp_edge_id,
-                                             edgestatus.GetPtr(opp_edge_id, tile)},
-                                            shortcuts, tile, offset_time);
+    if (opp_edge) {
```
kevinkreiser, line 397: it pains me to see this `if` but i guess its equivalent to logical and anyway

src/thor/bidirectional_astar.cc
kevinkreiser: with all the changes to this algorithm we need to also check it for regressions by running a befor and after route set.

i have messed with these lines too before and confused myself about what is the right thing

src/thor/multimodal.cc
```diff
@@ -42,9 +42,10 @@ namespace thor {
 MultiModalPathAlgorithm::MultiModalPathAlgorithm(const boost::property_tree::ptree& config)
     : PathAlgorithm(config.get<uint32_t>("max_reserved_labels_count_astar",
                                          kInitialEdgeLabelCountAstar),
-                    config.get<bool>("clear_reserved_memory", false)),
-      max_walking_dist_(0), max_label_count_(std::numeric_limits<uint32_t>::max()),
-      mode_(travel_mode_t::kPedestrian), travel_type_(0) {
+                    config.get<bool>("clear_reserved_memory", false)) {
+  mode_ = travel_mode_t::kPedestrian;
+  travel_type_ = 0;
+  max_walking_dist_ = 0;
```
kevinkreiser, line 48: why move them out of the initializer list? if it was a warning, just reorder them to shut the warning up

valhalla/thor/matrix_common.h
kevinkreiser: please dont resolve these, its helpful for other reviews :pray:

src/thor/costmatrix.cc
kevinkreiser: nit: cache is a poor name here, it should be something about "reservation". cache implies that it holds data that we want to use again, when in fact it simply holds an allocation that we want to re-use with new data

run_route_scripts/run_with_server_matrix.py
nilsnolde: tbh, the vast majority of python in this repo feels like tech debt (except for most installable scripts)😄 I should’ve done the work to clean this all up a bit more and not just copy paste. I’ll do that at some point I’m sure. If you want I can keep this locally instead of committing until I find the motivation to do that:)

run_route_scripts/run_with_server_matrix.py
nilsnolde: Na the other script is doing one file per request, for matrix that’s not cool

src/thor/astar_bss.cc
nilsnolde: I mentioned it somewhere: it’s not used at all. We still do what you’re saying, but it’s done with this variable. It’s this “nc” a bit further up.

src/thor/multimodal.cc
nilsnolde: That was more of a preference call, but I can revert. I find it more legible. My eyes have a hard time seeing all variables in long initializer lists

src/thor/dijkstras.cc
```diff
@@ -78,7 +78,7 @@ void Dijkstras::Initialize(label_container_t& labels,
   uint32_t edge_label_reservation;
   uint32_t bucket_count;
   GetExpansionHints(bucket_count, edge_label_reservation);
-  labels.reserve(std::min(max_reserved_labels_count_, edge_label_reservation));
```
kevinkreiser, line 81: this used have the effect of capping the reservation at `kInitialEdgeLabelCountDijkstras` now you can configure to allow for more. i guess thats ok, but the diff isnt enough to understand thats happening

valhalla/thor/costmatrix.h
```diff
@@ -131,43 +146,39 @@ class CostMatrix {
...
+
   // when doing timezone differencing a timezone cache speeds up the computation
   baldr::DateTime::tz_sys_info_cache_t tz_cache_;
 
   // for tracking the expansion of the Dijkstra
   expansion_callback_t expansion_callback_;
 
+  const std::function<void()>* interrupt_ = nullptr;
```
kevinkreiser, line 180: i appreciate you doing that. its not so hard to add so there is no good excuse why i havent done this for so long..

src/thor/costmatrix.cc
```diff
@@ -152,70 +150,87 @@ void CostMatrix::SourceToTarget(Api& request,
...
-              if (targets.empty() && source_status_[source].threshold > 0) {
-                source_status_[i].threshold = -1;
-                if (remaining_sources_ > 0) {
-                  remaining_sources_--;
+              // if there's no more target and the current source has not exhausted
+              // we update the source's threshold so that it doesn't enter its outer "if" statement
+              // anymore
+              if (targets.empty() && locs_status_[MATRIX_FORW][source].threshold > 0) {
+                // TODO(nils): shouldn't we extend the search here similar to bidir A*
+                //   i.e. if pruning was disabled we extend the search in the other direction
```
kevinkreiser, line 180: makes intuitive sense

src/thor/costmatrix.cc
```diff
@@ -152,70 +150,87 @@ void CostMatrix::SourceToTarget(Api& request,
...
-            auto& targets = source_status_[source].remaining_locations;
+    for (uint32_t i = 0; i < locs_count_[MATRIX_REV]; i++) {
+      if (locs_status_[MATRIX_REV][i].threshold > 0) {
+        locs_status_[MATRIX_REV][i].threshold--;
+        Expand<MatrixExpansionType::reverse>(i, n, graphreader);
+        // if we exhausted this search
+        if (locs_status_[MATRIX_REV][i].threshold == 0) {
+          for (uint32_t source = 0; source < locs_count_[MATRIX_FORW]; source++) {
+            // update the target for each source's remaining targets, if it still exists
+            auto& targets = locs_status_[MATRIX_FORW][source].unfound_connections;
```
kevinkreiser, line 170: haha cool yeah there you did do the alias thing i was suggesting and it indeed saved some small diffs haha. i still cant really read this without manually comparing each line though

src/thor/costmatrix.cc
```diff
@@ -152,70 +150,87 @@ void CostMatrix::SourceToTarget(Api& request,
...
+  if (!ignore_hierarchy_limits_)
+    ModifyHierarchyLimits();
+
   // Perform backward search from all target locations. Perform forward
   // search from all source locations. Connections between the 2 search
   // spaces is checked during the forward search.
-  int n = 0;
+  uint32_t n = 0;
   while (true) {
-    // Iterate all target locations in a backwards search
```
kevinkreiser, line 160: why remove things like this comment? i know its clear to you and me but its nice to have headings on code for newcomers

src/thor/costmatrix.cc
```diff
@@ -152,70 +150,87 @@ void CostMatrix::SourceToTarget(Api& request,
...
-          if (remaining_sources_ > 0) {
-            remaining_sources_--;
+          // in any case make sure this was the last time we looked at this source
+          locs_status_[MATRIX_FORW][i].threshold = -1;
+          if (locs_remaining_[MATRIX_FORW] > 0) {
+            locs_remaining_[MATRIX_FORW]--;
           }
         }
       }
     }
```
kevinkreiser, line 230: ok i got to here, am i correct that nothing really changed in either of these loops, was just renaming and comments

src/thor/bidirectional_astar.cc
nilsnolde: Ok will do a check with a bigger request file and diff the narrative

src/thor/dijkstras.cc
nilsnolde: I can confirm that is what's happening:)

src/thor/costmatrix.cc
kevinkreiser: yeah why would that be there? does it have something to do with trivial matrices and avoid a 0 seconds?

src/thor/costmatrix.cc
kevinkreiser: haha ah ok :pray:

src/thor/costmatrix.cc
```diff
@@ -45,11 +46,11 @@ class CostMatrix::TargetMap : public robin_hood::unordered_map<uint64_t, std::ve
...
-    : mode_(travel_mode_t::kDrive), access_mode_(kAutoAccess), source_count_(0),
-      remaining_sources_(0), target_count_(0), remaining_targets_(0),
-      current_cost_threshold_(0), targets_{new TargetMap},
-      max_reserved_labels_count_(config.get<uint32_t>("max_reserved_labels_count_bidir_dijkstras",
-                                                      kInitialEdgeLabelCountBidirDijkstra)) {
+    : max_reserved_labels_count_(config.get<uint32_t>("max_reserved_labels_count_bidir_dijkstras",
+                                                      kInitialEdgeLabelCountBidirDijkstra)),
+      clear_reserved_memory_(config.get<bool>("clear_reserved_memory", false)), targets_{
+                                                                                    new TargetMap} {
+  // Note, most things are being initialized in Initialize() or before
```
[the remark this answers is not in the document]
nilsnolde: ok, I'll do that. fwiw, IIRC most algorithms have this problem, not even half the stuff is usually initialized. I'll change it for costmatrix, but will not touch the others for now, the PR is big enough as it stands:)

src/thor/costmatrix.cc
nilsnolde: ok..

src/thor/costmatrix.cc
nilsnolde: correct. the whole costmatrix was 95% pure refactor, the remaining 5% the deadend logic, which is also the only thing I wrote additional tests for. I'd like to add some tests comparing the algos too, making sure they report the same duration/distance (+/- some minor threshold)

src/thor/costmatrix.cc
```diff
@@ -260,249 +278,390 @@ void CostMatrix::SourceToTarget(Api& request,
...
-        GraphId node = trans->endnode();
-        graph_tile_ptr endtile = graphreader.GetGraphTile(node);
-        if (endtile != nullptr) {
-          expand(endtile, node, endtile->node(node), pred, pred_idx, true, ti);
-        }
-      }
-    }
-  };
+    // TODO(nils): what if there is a shortcut that supersedes our u-turn? can that even be?
+    // We then need to decide if we should expand the shortcut or the non-shortcut edge...
```
kevinkreiser, line 656: this is a good question. yes we can make a uturn where a shortcut begins/ends. however, isnt it always safe to expand on the non shortcut? the worst scenario is that we were slightly less efficient in finding a solution because there as a shortcut that we didnt use. maybe thats all you're really saying here: "to be fully optimial we shoudl favor a shortcut uturn (when using shortcuts) as opposed to a superseded edge"

src/thor/bidirectional_astar.cc
kevinkreiser: also check the perf diff (can be done at the same time with the same script by running with `time`)

src/thor/costmatrix.cc
nilsnolde: no, it's there because I messed up on a previous PR and it wasn't caught in the review. I can only guess that I wanted to round, but it's for sure the wrong place for that:)

kevinkreiser:
ok im done looking it over! i had a lot of little nitpicks so i'll let you choose which ones you want to do, if you dont feel like doing them but also dont care that much let me know and i can do them.

but the only major thing i would say is we have to do the same perf test/diff for bidir a* since it got some refactoring and edge status changes (which you said you'd already look at).

let me know when done and ill put a ship!

valhalla/thor/costmatrix.h
```diff
@@ -295,6 +325,20 @@ class CostMatrix {
     return infos;
   };
 
+  void ModifyHierarchyLimits() {
+    // Distance threshold optimized for unidirectional search. For bidirectional case
+    // they can be lowered.
+    // Decrease distance thresholds only for arterial roads for now
+    for (size_t source = 0; source < locs_count_[MATRIX_FORW]; source++) {
```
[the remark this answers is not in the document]
nilsnolde: oh shit, I didn't actually see the 5.0! I saw a 2.0 (meaning I didn't even bother to look apparently)! haha oops, good catch!

nilsnolde:
@kevinkreiser I addressed your review in https://github.com/valhalla/valhalla/pull/4372/commits/94af58b59bd51f60a84faf0a444e10c82a06150f

run_route_scripts/run_with_server_matrix.py
kevinkreiser: nah its ok, i trust you to keep it on your list :smile:

run_route_scripts/run_with_server_matrix.py
kevinkreiser: i was focused on this part

> default='http://localhost:8002/**route**'

valhalla/thor/costmatrix.h
```diff
@@ -131,43 +146,39 @@ class CostMatrix {
...
-  std::vector<std::vector<sif::BDEdgeLabel>> source_edgelabel_;
-  std::vector<EdgeStatus> source_edgestatus_;
-
-  // Adjacency lists, EdgeLabels, EdgeStatus, and hierarchy limits for each
-  // target location (reverse traversal)
-  std::vector<std::vector<sif::HierarchyLimits>> target_hierarchy_limits_;
-  std::vector<baldr::DoubleBucketQueue<sif::BDEdgeLabel>> target_adjacency_;
-  std::vector<std::vector<sif::BDEdgeLabel>> target_edgelabel_;
-  std::vector<EdgeStatus> target_edgestatus_;
+  std::array<std::vector<LocationStatus>, 2> locs_status_;
```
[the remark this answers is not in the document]
kevinkreiser: all good no worries

src/thor/costmatrix.cc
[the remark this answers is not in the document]
kevinkreiser: agree. its like many things. over time we improve, when we touch stuff good to improve it in ways we have found deficient thus far

kevinkreiser (approved):

PR #4375, opened 2023-10-31
add --provenance=false --sbom=false to buildx command

### LINKED ISSUE #4374: Docker image can't be pulled!
I can't pull the latest image after https://github.com/valhalla/valhalla/pull/4346.. When pulling it says `no matching manifest for linux/amd64 in the manifest list entries`, and indeed, looking at our packages, there's no `linux/amd64` descriptor, just says `unknown/unknown`:
image

Seems to be a Github bug: https://github.com/orgs/community/discussions/45969#discussioncomment-4852744.

nilsnolde:
fixes #4374 :crossed_fingers:


nilsnolde:
Could you try cowboying it @kevinkreiser ? If it doesn't work I don't know how to resolve this tbh..

nilsnolde:
Comparing the logs of the last master docker image build, it _seems_ as if the `linux/amd64` isn't exporting the manifests which `linux/arm64` is doing. I don't know what that exactly means though..

nilsnolde:
Oh, maybe buildx isn't supposed to be used this way.. I mean having separate workflows for two archs, since they both push and if there's some sort of manifest file, that might be overwritten and cause other stuff.. Maybe that whole PR needs to be reverted..

nilsnolde:
reverted the main change of splitting the workflows.

nilsnolde:
the last arm64 run took just under 5.5 hours and amd64 takes just under 30 mins as  well. walking the line of github limits:)

nilsnolde:
should it go past github's limit, we can still do both builds in separate jobs (but same workflow) without pushing, and do  the push in a separate job after the build jobs finished. just in case..

ImreSamu:
> the last arm64 run took just under 5.5 hours and amd64 takes just under 30 mins as well. walking the line of github limits:)

If the current solution is not working, then a possible alternative could be to perform the ARM-based building entirely on the native ARM CircleCI.

What I also noticed is that the [PostGIS tests could only be run successfully in QEMU](https://github.com/postgis/docker-postgis/issues/216#issuecomment-1763373131) when JIT (Just-In-Time compilation) was turned off. Because of this, I am somewhat overly skeptical about using images generated under QEMU for production purposes.

nilsnolde:
wouldn't circleci suffer the same problem, i.e. the last image build will overwrite whatever was there previously? running it only on arm would need cross-compilation for amd. though there must be a way to "append" images to an already existing latest..

kevinkreiser:
running on circle wont take nearly as long because we can schedule it to happen on an arm cpu since circle now has native arm. overwriting whatever was there previously is ok right? basically we have latest and tags. all we have to do is port the docker github workflows back to circle :frowning_face: i'll have to add the right auth for ghcr but should be fine. there should be no need for cross compilation. we might need a build arg to tell the builder to pull the arm image from ubuntu though i kind of forget how that works

kevinkreiser:
oh maybe you want the ability to do a specific branch which circle then wouldnt support.. i see :frowning_face:

ImreSamu:
> wouldn't circleci suffer the same problem, i.e. the last image build will overwrite whatever was there previously?


Indeed, it's not an easy problem;
I've also encountered this issue with the new 'postgis/postgis' Docker images. 
However, I currently have a working prototype that shows promise, relying on the [manifest-tool](https://github.com/estesp/manifest-tool).

Now:
* internal amd64 only repo: https://hub.docker.com/r/imresamu/postgis-amd64/tags
  * generated via Github CI  + running manifest-tool [1] re-creating manifest.
* internal arm64 only repo: https://hub.docker.com/r/imresamu/postgis-arm64/tags
  * generated via CircleCI arm + running manifest-tool  [1] re-creating manifest.
* (amd64 + arm64) for the users: https://hub.docker.com/r/imresamu/postgis

image


[1]  https://github.com/estesp/manifest-tool
```
echo "manifest:  16-master-bookworm"
manifest-tool push from-args \
    --platforms linux/amd64,linux/arm64 \
    --template ${dockername}-ARCHVARIANT:16-master-bookworm \
    --target ${dockername}:16-master-bookworm || true
 ```

nilsnolde:
yeah this PR is basically reverting the most recent commit. I'd say, if this works for now, let's merge so master builds again. circleci can come after? 

> there should be no need for cross compilation

but you can't build on 2 machines and push two archs to the same repo (that's where our current problem comes from), so how would that work then? might be that docker supports that in some very convoluted way, but hard to google and couldn't find anything useful.

ImreSamu:
this is my code: https://github.com/ImreSamu/docker-postgis
and this is the CircleCI config : https://github.com/ImreSamu/docker-postgis/blob/master/.circleci/config.yml

nilsnolde:
thanks @ImreSamu and congrats for making it work! it'd be much nicer to have this done on proper machines at circleci.

should we merge here first and add an issue to migrate (again)?

nilsnolde:
a word of caution though: ghcr might be very differently implemented than dockerhub.

kevinkreiser:
> but you can't build on 2 machines and push two archs to the same repo (that's where our current problem comes from), so how would that work then? 

i see, i couldnt grok from this pr that that was the exact problem. we can suffix the image names like we for the manual builds and be done with it

kevinkreiser:
ok i think if we really want this to work we should make a small change to the build to speed it up. basically we should build only release and comment out the parts about stripping binaries and gzipping the debug symbols. i can add that here if you want or you can take a swag. if we do that i think the build will make the 6 hour deadline

kevinkreiser:
:cowboy_hat_face:

kevinkreiser:
@ImreSamu interesting comment about JIT and postgis. i have found that in the previous version of ubuntu 22.04 lua JIT is broken under qemu and therefor valhalla would not build on that version. after updating to ubutnu 23.04 lua JIT works and valhalla builds successfully. now this is lua and i dont know what compiler we are talking about for postgis (didnt know it had any JIT support) but still it made me remember that and so i thought id post about it before i lost it from memory again :smile:

ImreSamu:
> and i dont know what compiler we are talking about for postgis (didnt know it had any JIT support) 

Actually, PostgreSQL has SQL  JIT (Just-In-Time) support; however, since it's enabled by default, it also affects the PostGIS tests.

> "PostgreSQL has builtin support to perform JIT compilation using [LLVM](https://llvm.org/) when PostgreSQL is built with [--with-llvm](https://www.postgresql.org/docs/current/install-make.html#CONFIGURE-WITH-LLVM)."
> * https://www.postgresql.org/docs/current/jit-reason.html

PR #4378, opened 2023-11-02
Fix ferry shortest path recovery

### LINKED ISSUE #4358: Ferry Reclassification - Can reclassify nearly all edges on small islands
I found a case in NY (other cases may be likely) on Fishers Island, NY (south of New London, Connecticut) where all edges are reclassified when reclassifying edges connected to ferries. This island has no ways/edges that meet the target classification (primary) - so the algorithm continues to expand from the ferry and essentially reclassify nearly all edges. The algorithm continues until the adjacency queue is empty or:

      if (n > 400 && GetBestNonFerryClass(expanded_bundle.node_edges) <= kFerryUpClass) {
        break;
      }
the 2nd half of this condition is never met on Fisher Island.

I think we should:
1. Alter this condition to only increment n if the 2nd condition is met and break sooner (e.g. maybe after 32 nodes meet this criteria).
2. If after breaking out of the loop (or adj queue is empty which exits the loop) there are no connections to nodes with edges meeting the road classification target perhaps we should not reclassify any edges.

### LINKED ISSUE #4373: Ferry Approach Reclassification - recovers paths not ending in proper classification
When reclassifying edges connected to ferries I have found that the shortest path is often recovered from low classification edges (rather than an edge that meets to proper classification) when the priority queue becomes empty.

In code, 

```
        // Get the last label index to form the path
        last_label_idx = nodelabel_index;

```
is done for any edge when expanding. We should only do this when the proper classification is met: 

`        GetBestNonFerryClass(expanded_bundle.node_edges) <= kFerryUpClass`

so that when recovering the shortest path it starts on the proper classification edge. This will protect against starting on the wrong edge when the adjacency set is exhausted. 

The proper label_idx should be found using `label_idx = expand_node.second;` for each expanded node. Then when a node that meets the classification target is found, last_label_idx should be updated to label_idx. Currently, the shortest path always starts from the last label that was added rather than the label that was removed from the priority queue.

dnesbitt61:
fixes #4373
fixes #4358

Incorrect paths are frequently recovered from the ShortestPath method when reclassifying edges approaching ferries. This fixes the last_label_idx and ensures it is only set when the proper classification is reached or exceeded.

If no edges are encountered with proper classification, then no edges are reclassified. A log statement was added to identify lat,lng locations where no edges are reclassified.

For a test case of New York state I see the following for ferries that failed to find paths to the target road classification:
```
2023/11/03 14:58:22.177548 [INFO] Reclassification fails both directions to ferry at LL =40.693268,-74.015295
2023/11/03 14:58:22.306188 [INFO] Reclassification fails both directions to ferry at LL =41.172929,-72.204647
2023/11/03 14:58:22.317576 [INFO] Reclassification fails both directions to ferry at LL =41.256999,-72.031202
2023/11/03 14:58:22.448317 [INFO] Reclassification fails both directions to ferry at LL =43.911944,-74.919672
2023/11/03 14:58:22.456406 [INFO] Reclassification fails both directions to ferry at LL =44.553565,-73.855213
```
These all seem valid.

## Tasklist

 - [x] Add tests

## Requirements / Relations


dnesbitt61:
Looking for ideas on a better "termination" - identifying when "enough" edges of the proper classification are found.

nilsnolde:
Should we add a small test for #4358 ? i.e. a map which doesn't have any edge >= kPrimary and make sure we didn't reclassify anything?

dnesbitt61:
I added such a test. It looks like the first connecting edge is always reclassified (see comment close to the bottom of ferry_connections.cc).

nilsnolde:
> It looks like the first connecting edge is always reclassified

Hmpf, yeah, maybe we should do this [here](https://github.com/valhalla/valhalla/pull/4378/files#diff-954cf29d63a3640b506aec5b23307e99932ee1c3d4d607c75dae12d0cbeebde5L377):
```
        update_edge.attributes.importance = total_count > 0 ? kFerryUpClass : update_edge.attributes.importance;
``` 
?

nilsnolde (approved):
LGTM :ship:

Issue #3817, opened 2022-11-03
TZDB shapefile is out of date


ianthetechie:
Hey guys,

I noticed that the TZDB shapefile is now pretty far out of date. I *thought* updating this would be straightforward, but looking at the script I'm seeing what appears to be a set of manually defined aliases. This raised a few questions.

1. Could we document why / how these are used?
2. I noticed as well that there is a `USE_SYSTEM_TZ_DB` CMAKE option as well that's disabled by default, but wasn't sure how/if it related to this. Obviously the point of the sqlite tzdb is to determine which timezone a point is located in. I would assume system functions handle the rest, since this is basically just a shapefile and not the whole tzdb ruleset.

TL;DR any gotchas just doing an update to `2022f`?

kevinkreiser:
@ianthetechie this is a great question!

1. yes we could and we should document how this works
2. timezone information can be platform dependent but we build the ordering of timezones (as an integer index) into our tiles. and those indices are expected to match the timezones in the date_time code. if you use a different tz db with a different ordering than was used to create the tile data you will have problems

so long story short, i am not opposed to updating to newer data but we need to do the work of making sure the same indices are used for the same semantic timezones. eg. maybe there are new ones, these would have to go on the end. i havent worked out all the details and maybe its less tedious than i am imagining but we need to proceed cautiously to avoid breaking backward/forward compatibility. in other words, when we make a chance we either have to do one of two things:

1. make sure that both: old data paired with new code AND new data paired with old code will work after the change, OR
2. update the major version of valhalla

I guess you probably realized we are trying very very hard not to do the latter until we have sufficiently many important breaking changes as to warrant it.

nilsnolde:
This has gotten some new traction, see https://github.com/valhalla/valhalla/issues/4363. That's pretty bad, apart from the changes in geometry (not too little in the US, I visually compared the two releases), DST is probably the worst affected. 

I looked into it for a bit and this is my summary of the process:
1. We get the shapefile for the geometries and build our sqlite from that, containing the IANA timezone IDs (e.g. Europe/Berlin), but also various aliases for a bit of normalization (e.g. Europe/Vaduz -> Europe/Zurich). This is ONLY the timezone ID with geometries though, no rules.
2. The actual IANA timezone rules are in ./date_time with ASCII files and they need to match the shapefile release, e.g. [here](https://www.iana.org/time-zones). They're compiled to simple headers so we can include the info in baldr/tz_alt.cpp
[...]

PR #4382, opened 2023-11-05
remove timezone aliases from sql script

nilsnolde:
first step towards #3817 

I think it's that simple.


src/baldr/datetime.cc
```diff
@@ -55,6 +55,10 @@ tz_db_t::tz_db_t() : db(date::get_tzdb()) {
   for (const auto& zone : db.zones) {
     names.emplace(zone.name(), ++idx);
   }
+  // add the aliases of the links here with the index of their targets
+  for (const auto& link : db.links) {
+    names.emplace(link.name(), names.find(link.target())->second);
+  }
```
nilsnolde, line 61: this should be all we need. previously valhalla_build_timezones would alias old timezone names from the shapefile in the sqlite db. that adds a fair bit of burden. `date_time` has its own list of those "links" where the `name()` is the old tz name and the `target()` is the new alias, equivalent to what we were doing ourselves before. now the whole aliasing bound to the `date_time` library, which is so much nicer. and no one has to wonder about those aliases anymore.

test/datetime.cc
```diff
@@ -631,6 +631,24 @@ TEST(DateTime, TestDayOfWeek) {
   EXPECT_EQ(dow, 3) << "DateTime::day_of_week failed: 3 expected";
 }
 
+TEST(DateTime, TimezoneAliases) {
+  auto dt_db = DateTime::get_tz_db();
+  // map of alias and target names (taken from old hand curated list in valhalla_build_timezones)
+  std::vector<std::pair<std::string, std::string>> pairs = {{"Etc/Zulu", "Etc/UTC"},
```
nilsnolde, line 637: the test passes at least, I'm convinced:)

scripts/valhalla_build_timezones
```diff
@@ -23,225 +23,13 @@ rm -f ./timezones-with-oceans.shapefile.zip
 
 url="https://github.com/evansiroky/timezone-boundary-builder/releases/download/2018d/timezones-with-oceans.shapefile.zip"
```
gknisely, line 24: @nilsnolde should we use the latest release?  https://github.com/evansiroky/timezone-boundary-builder/releases

nilsnolde: Jep @gknisely, gut that'll take 1-2 more PRs. We need to also update the IANA files.

Next PR I want to make one more refactor so we can keep the current tz order to not break existing clients after we update tzdb.

src/baldr/datetime.cc
molind: I've merged changes from master into our micro branch and noticed that this change breaks US_OS_TZDB build option. It won't build because there is no links member inside tzdb, when US_OS_TZDB=1 is defined.

quote from data/tz.h
```
struct tzdb
{
    std::string                 version = "unknown";
    std::vector<time_zone>      zones;
#if !USE_OS_TZDB
    std::vector<time_zone_link> links;
#endif
```

Is it expected limitation?

kevinkreiser: iirc we never wanted to use the os tzdb as the ordering of stuff could change and therefore make data that is incompatible with our ordering of the tz names.

molind: I've seen in some other sources this define and thought it's supported.

molind: ok. I'll refactor our part to use embed tzdb from third_parties. Thanks for clarification.

nilsnolde: it'd be nice to be able to use system tzdbs. and I kinda figured we're able to do just that, now that we're hard-coding things even more.

> It won't build because there is no links member inside tzdb, when US_OS_TZDB=1 is defined.

that's a pity and I wonder how we'd get aliases in system dbs.

anyways, if anyone ever badly wants it, they can dive in:)

PR #4390, opened 2023-11-13
Avoid Issues with GraphReader::GetBeginNodeId

kevinkreiser:
this method, like all the other methods takes a graphtile pointer so that it can speed up lookups if the tile is (likely) already loaded. The problem is it takes this as a reference and then may mutate it to an adjacent tile if the edge in question leaves the tile. The problem with this is that its unintuitive because the begin node which you get back, by definition should be in the same tile as the edge you passed in. There should then, be no need to mutate the tile if it was already the right one. An argument could be made that we want to set it in case the person passed in an uninitialized one so. Perhaps its best to make it do this, especially since this would match all the rest of the functions that work like this. i'll do that

### DIFF
--- a/valhalla/baldr/graphreader.h
+++ b/valhalla/baldr/graphreader.h
@@ -669,7 +669,8 @@ public:
    * @param   edge2  GraphId of second directed edge.
    * @param   tile    Reference to a pointer to a const tile.
    * @return  Returns true if the directed edges are directly connected
-   *          at a node, false if not.
+   *          at a node, false if not. If successful the tile will be set
+   *          to the one containing the end node of edge1
    */
   bool AreEdgesConnectedForward(const GraphId& edge1, const GraphId& edge2, graph_tile_ptr& tile);
 
@@ -768,7 +769,8 @@ public:
    * @return Returns a pair of GraphIds: the first is the start node
    *         and the second is the end node. An invalid start node
    *         can occur in regional extracts (where the end node tile
-   *         is not available).
+   *         is not available). If successful tile will be set to
+   *         the one containing edgeid
    */
   std::pair<GraphId, GraphId> GetDirectedEdgeNodes(const GraphId& edgeid, graph_tile_ptr& tile) {
     if (tile && tile->id().Tile_Base() == edgeid.Tile_Base()) {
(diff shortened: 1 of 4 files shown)


kevinkreiser:
seems timedep tests are failing. with any lucky they were relying on faulty logic and need to be updated :smile:

valhalla/baldr/graphreader.h
```diff
@@ -637,16 +637,19 @@ class GraphReader {
   /**
    * Method to get the begin node of an edge by using its opposing edges end node
    * @param edge    the edge whose begin node you want
-   * @param tile    reference to a pointer to a const tile
+   * @param tile    reference to a pointer to a const tile containing the begin node
    * @return        returns GraphId of begin node of the edge (empty if couldn't find)
    */
   GraphId GetBeginNodeId(const DirectedEdge* edge, graph_tile_ptr& tile) {
```
danpat, line 643: If this isn't actually going to mutate the tile pointer any mroe, the signature should be changed `const graph_tile_ptr&`

danpat: Oh I see, I missed the added call to `GetGraphTile(opp_edge->endnode(), tile);`.

nilsnolde (approved):

danpat:
@kevinkreiser Some tests that verify the behaviour of this (and other similar graphptr-mutating functions) might be a good idea.  This is probably the 10th time I've seen a similar issue related to using the wrong tile in tile-boundary situations, it's a bit of a chronic issue.

valhalla/baldr/graphreader.h
```diff
@@ -637,16 +637,19 @@ class GraphReader {
...
-    if (!GetGraphTile(edge->endnode(), tile))
+    // grab the end node maybe in an adjacent tile
+    graph_tile_ptr maybe_other_tile = tile;
+    if (!GetGraphTile(edge->endnode(), maybe_other_tile))
       return {};
-    const auto* node = tile->node(edge->endnode());
-    // grab the opp edges end node
-    const auto* opp_edge = tile->directededge(node->edge_index() + edge->opp_index());
+    const auto* node = maybe_other_tile->node(edge->endnode());
+    // grab the opp edge also could be in this adjacent tile
```
julianskartor, line 649: I was wondering if we could write out "opposite" in the comment? Also maybe some punctuation is missing in the sentence? :thinking:

nilsnolde: maybe a comma missing, but "opp edge" should be readily understandable for someone coming this far in the code base. it's literally all over the comments & var names, can't avoid it

PR #4392, opened 2023-11-14
properly use defined source and build dirs for testing

kevinkreiser:
trivial one but in a lot of the tests we dont use directories in a CWD safe way which causes some tests to fail when you run them outside of their narrow working criteria. this pr simply fixes the tests that rely on paths to either source stuff or stuff we are building in the build directory

### DIFF
--- a/test/countryaccess.cc
+++ b/test/countryaccess.cc
@@ -20,16 +20,12 @@
 
 #include "test.h"
 
-#if !defined(VALHALLA_SOURCE_DIR)
-#define VALHALLA_SOURCE_DIR
-#endif
-
 using namespace valhalla::mjolnir;
 using namespace valhalla::baldr;
 
 namespace {
 
-const std::string config_file = "test/test_config_country";
+const std::string config_file = VALHALLA_BUILD_DIR "test/test_config_country";
 
 // Remove a temporary file if it exists
 void remove_temp_file(const std::string& fname) {
--- a/test/alternates.cc
+++ b/test/alternates.cc
@@ -21,7 +21,7 @@ using namespace valhalla::midgard;
 
 namespace {
 
-const auto conf = test::make_config("test/data/utrecht_tiles");
+const auto conf = test::make_config(VALHALLA_BUILD_DIR "test/data/utrecht_tiles");
 
 struct route_tester {
   route_tester()
--- a/test/astar.cc
+++ b/test/astar.cc
@@ -127,7 +127,7 @@ const gurka::relations relations3 = {{{gurka::relation_member{gurka::way_member,
 
 const vb::GraphId tile_id = vb::TileHierarchy::GetGraphId({.125, .125}, 2);
 gurka::nodelayout node_locations;
-const std::string test_dir = "test/data/fake_tiles_astar";
+const std::string test_dir = VALHALLA_BUILD_DIR "test/data/fake_tiles_astar";
 const auto fake_conf =
     test::make_config(test_dir,
                       {{"mjolnir.admin", VALHALLA_SOURCE_DIR "test/data/netherlands_admin.sqlite"},
--- a/test/astar_bikeshare.cc
+++ b/test/astar_bikeshare.cc
@@ -28,7 +28,7 @@ namespace {
 // avoid that we set a radius here to get both sets of edges and let the algorithm take the cheaper
 // one. this only worked before by luck
 const auto conf =
-    test::make_config("test/data/paris_bss_tiles", {{"loki.service_defaults.radius", "10"}});
+    test::make_config(VALHALLA_BUILD_DIR "test/data/paris_bss_tiles", {{"loki.service_defaults.radius", "10"}});
 
 struct route_tester {
   route_tester()
--- a/test/graphbuilder.cc
+++ b/test/graphbuilder.cc
@@ -22,14 +22,6 @@ using valhalla::baldr::GraphReader;
 using valhalla::mjolnir::build_tile_set;
 using valhalla::mjolnir::TileManifest;
 
-#if !defined(VALHALLA_SOURCE_DIR)
-#define VALHALLA_SOURCE_DIR
-#endif
-
-#if !defined(VALHALLA_BINARY_DIR)
-#define VALHALLA_BINARY_DIR
-#endif
-
 namespace {
 
 const std::string pbf_file = {VALHALLA_SOURCE_DIR "test/data/harrisburg.osm.pbf"};
--- a/test/graphparser.cc
+++ b/test/graphparser.cc
@@ -16,17 +16,13 @@
 #include "baldr/directededge.h"
 #include "baldr/graphconstants.h"
 
-#if !defined(VALHALLA_SOURCE_DIR)
-#define VALHALLA_SOURCE_DIR
-#endif
-
 using namespace valhalla::midgard;
 using namespace valhalla::mjolnir;
 using namespace valhalla::baldr;
 
 namespace {
 
-const std::string config_file = "test/test_config_gp";
+const std::string config_file = VALHALLA_BUILD_DIR "test/test_config_gp";
 
 const auto node_predicate = [](const OSMWayNode& a, const OSMWayNode& b) {
   return a.node.osmid_ < b.node.osmid_;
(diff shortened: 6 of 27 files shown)


nilsnolde:
ah thanks for the chore! never had the motivation so far, but it's so annoying to get all the test maps dumped in the wrong directory and messing up git (plus the inability to run them in some ways..)

kevinkreiser:
yeah was working on #4390 and it was annoying me so last night i decided i was going to remedy it. i didnt check the gurka directory yet but i can get that in another pass

nilsnolde:
Should we just add the `VALHALLA_BUILD_DIR` to `test::make_config` and be done with it? That'd give us all gurka tests for free

kevinkreiser:
i had the same th ought yep

test/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ target_include_directories(valhalla_test PUBLIC
   ${VALHALLA_SOURCE_DIR}/third_party/just_gtfs/include
   ${VALHALLA_SOURCE_DIR}/third_party/libosmium/include
   ${VALHALLA_SOURCE_DIR}/third_party/microtar/src)
+target_compile_definitions(valhalla_test PRIVATE
+  VALHALLA_SOURCE_DIR="${VALHALLA_SOURCE_DIR}/"
+  VALHALLA_BUILD_DIR="${VALHALLA_BUILD_DIR}/")
```
kevinkreiser, line 31: all test targets get these two dirs defined so that..

test/actor.cc
```diff
@@ -7,10 +7,6 @@
 
 #include "test.h"
 
-#if !defined(VALHALLA_SOURCE_DIR)
-#define VALHALLA_SOURCE_DIR
-#endif
-
```
kevinkreiser, line 13: we dont have to have this random hack in there and we always get the right value (i have no idea why this kind of thing was hacked in here..)

test/countryaccess.cc
```diff
@@ -20,16 +20,12 @@
...
-#define VALHALLA_SOURCE_DIR
-#endif
-
 using namespace valhalla::mjolnir;
 using namespace valhalla::baldr;
 
 namespace {
 
-const std::string config_file = "test/test_config_country";
+const std::string config_file = VALHALLA_BUILD_DIR "test/test_config_country";
```
kevinkreiser, line 28: in places where we dont use `make_config` we use the thing manually

test/gurka/test_gtfs.cc
```diff
@@ -151,16 +151,16 @@ const gurka::ways ways = {{"AB", {{"highway", "primary"}}}, {"BC", {{"highway",
 
 boost::property_tree::ptree get_config() {
 
-  return test::make_config(VALHALLA_BUILD_DIR "test/data/transit_tests",
+  return test::make_config("test/data/transit_tests",
```
kevinkreiser, line 154: in places where we do use `make_config` we rely on it to do the work for us

test/gurka/test_match.cc
```diff
@@ -334,7 +335,7 @@ TEST(MapMatchRoute, IgnoreRestrictions) {
                                        },
                                        {{"type", "restriction"}, {"restriction", "no_straight_on"}}}};
   const gurka::map map =
-      gurka::buildtiles(layout, ways, {}, relations, "test/data/mapmatch_restrictions",
+      gurka::buildtiles(layout, ways, {}, relations, VALHALLA_BUILD_DIR "test/data/mapmatch_restrictions",
                         {{"mjolnir.timezone", VALHALLA_BUILD_DIR "test/data/tz.sqlite"}});
```
kevinkreiser, line 339: i felt it made sense to do it manually in these cases where we use the config option overrides because if we rely on `make_config` doing the right thing by prefixing directories with `VALHALLA_BUILD_DIR`, but then we add a config override for timezone and we ourselves dont prefix it, then we've defeated ourselves.

alternatively i could have written code to see if any override value begins with "test/data" and then prefixed that but it felt a little too much? i dont feel stronly about it

test/test.cc
```diff
@@ -385,8 +385,14 @@ boost::property_tree::ptree make_config(const std::string& path_prefix,
...
 
+  // we should use absolute paths so the binaries work regardless of CWD
+  std::string absolute;
+  if (path_prefix.find("test/") == 0) {
+    absolute = VALHALLA_BUILD_DIR;
+  }
+
   // force the paths to be different
-  boost::replace_all(defaults, "%%", path_prefix);
+  boost::replace_all(defaults, "%%", absolute + path_prefix);
```
kevinkreiser, line 395: this is the important part, we detect if your data prefix doesnt have the absolute path on and add it if not. this lets the binaries be run from any working directory (which can be useful)

kevinkreiser: also note that `gurka::buildtiles` calls into this function so its covered too

kevinkreiser:
this is ready for review

test/gurka/test_gtfs.cc
```diff
@@ -151,16 +151,14 @@ const gurka::ways ways = {{"AB", {{"highway", "primary"}}}, {"BC", {{"highway",
...
-                             VALHALLA_BUILD_DIR "test/data/transit_tests/gtfs_feeds"},
-                            {"mjolnir.transit_dir",
-                             VALHALLA_BUILD_DIR "test/data/transit_tests/transit_tiles"},
+  return test::make_config("test/data/transit_tests",
+                           {{"mjolnir.transit_feeds_dir", "test/data/transit_tests/gtfs_feeds"},
+                            {"mjolnir.transit_dir", "test/data/transit_tests/transit_tiles"},
                             {"mjolnir.transit_pbf_limit",
                              "1"}, // so we create more than one file per tile
                             {"mjolnir.hierarchy", "1"},
-                            {"mjolnir.timezone", VALHALLA_BUILD_DIR "test/data/tz.sqlite"},
```
nilsnolde, line 162: hm, do we really want to remove it here? or maybe it's added somewhere else I didn't yet?

test/gurka/test_match.cc
nilsnolde: not sure I understand this: on line 205 you're removing the prefix for tzdb, on line 337 you're adding it for the graph build (which should be covered already by make_config) and on line 339 you left it again for tzdb. probably an oversight?

IMO it'd be more consistent that noone ever has to specify VALHALLA_BUILD_DIR manually, but it's also a much rarer thing to do, so I'm ok with it too.

test/gurka/test_probable_restrictions.cc
```diff
@@ -67,7 +67,8 @@ class ProbableRestrictions : public testing::TestWithParam<int> {
               {gurka::node_member, "I", "via"},
           },
           {{"type", "restriction"}, {"restriction:probable", "no_left_turn @ probability=0"}}}};
-    map = gurka::buildtiles(layout, ways, {}, relations, "test/data/gurka_time_probable_restrictions",
+    map = gurka::buildtiles(layout, ways, {}, relations,
+                            VALHALLA_BUILD_DIR "test/data/gurka_time_probable_restrictions",
```
nilsnolde, line 71: yeah I'm confused.. isn't this the graph directory? don't we set that now in make_config()?

nilsnolde: same in lots of other places in gurka

kevinkreiser: we do and I went through 2 rounds of this and missed this in the last round. it's surprisingly annoying to find all occurances at midnight after a long day at work hahaha

test/gurka/test_gtfs.cc
kevinkreiser: yeah this is the primary open issue at this point. the signature for `make_config` takes first the directory argument and second a map of key values replacements, some of which might be things that are directories some not.

it was easy for me to patch the first arg when it didnt have the absolute path because i knew it was a directory but for the key values i dont know which ones are and arent directories. this means detecting and patching those is harder.

there are two options to do it automatically:

1. something simple where i see the value begins with `"test/data"` this will not catch everythign but it should get most of everything
2. figure out which ones should be directories by looking at the default config for the path value that got substituted in everywhere and then patching only those that were substituted paths

i think ill take a try at the 2nd one because then everything can be completely automatic and i can remove any of these that exist

nilsnolde: that'd be much appreciated, albeit quite a sizable mechanical chore

Issue #4389, opened 2023-11-13
Allow distance check for hierarchy pruning in CostMatrix


nilsnolde:
The bidir a* checks for both `max_up_transitions` and the distance from the origin/destination. We should do the same for CostMatrix, currently it's aborting transitions way too soon which messes up quality for shortish routes.

nilsnolde:
yeah, this turned out to be a performance hole

PR #4406, opened 2023-11-23
costmatrix: add hierarchy limit distance check

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4389

based on #4372 

this is expected to drain performance somewhat, as we're expanding much more on lower levels, especially in urban environments.


nilsnolde:
Performance test with the same dataset and method as in https://github.com/valhalla/valhalla/pull/4372#issuecomment-1823289253 :

master: 40 secs
branch: 55 secs

Probably not worth the penalty and this is much uglier to make configurable, as opposed to https://github.com/valhalla/valhalla/pull/4329. It's kind of the worst dataset for this PR as it's all pretty short distances in a very urban environment, but it's also where the PR should have the highest impact on route quality. 

I don't think it's even worth to measure quality impact.

Issue #4164, opened 2023-06-13
ferry reclassification algorithm reclassifies a few too many edges


nilsnolde:
While fixing #4118 I realized quite a few more issues. Most I could fix I think, but one is remaining and is rooted in the basic node-based algorithm it uses. It's a serious edge case, but this map
```
    A--B--C--D------E--F--G--H--M
       |  |            |  |
       I--J            K--L
```

with DE as ferry and FG as `dest_only`, it should upclass FKLG, and **not** FG, but it does upclass both of them.

The problem is that both edges have the same start & end node.

I just leave it here if we ever get that edge case as issue.

PR #4416, opened 2023-11-28
cleanup ferry connection test after dave's recent fixes

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4164


test/gurka/test_ferry_connections.cc
```diff
@@ -261,9 +259,7 @@ TEST(Standalone, ReclassifyFerryUntagDestOnly) {
         {"motorcycle", "yes"},
         {"route", "ferry"}}},
       {"EF", not_destonly},
-      {"FX", destonly}, // should stay destonly and low-class bcs connecting edge isn't and this will
```
nilsnolde, line 264: before there was another edge necessary to not have 2 edges with the same start/end node. luckily https://github.com/valhalla/valhalla/pull/4361 fixes that

dnesbitt61 (approved):

PR #4418, opened 2023-11-28
V2: enable bidir matrix connections on reverse tree

nilsnolde:
it's the same as in #4329, just rebased on #4372


nilsnolde:
Some more investigation on performance with 50 sources/targets (symmetrical) and 50 requests in Berlin area. `master` came out with 61 secs.

- only allocations (not actually checking connections): 65 secs, so around 6% degradation
- allocations + checking connections, i.e. full PR: 78 secs, so ~ 17% extra degradation, totaling -23%

We'd save some by using the existing robin hood structure, but the majority is with the connection check itself. We should still do that optimization as @kevinkreiser suggested, but it could be in a next PR as well IMO. WDYT?

src/thor/costmatrix.cc
```diff
@@ -294,8 +298,6 @@ void CostMatrix::Initialize(
     const google::protobuf::RepeatedPtrField<valhalla::Location>& source_locations,
     const google::protobuf::RepeatedPtrField<valhalla::Location>& target_locations) {
 
-  targets_->reserve(max_reserved_labels_count_);
```
nilsnolde, line 297: funnily enough it's better for performance if we don't reserve anything in the robin hood hash (3-4% improvement)

kevinkreiser:
you could force push this branch over top of the old one so we can keep the discussions in one place on the same original pr

nilsnolde:
merged into the old branch, closing here..

PR #4423, opened 2023-11-30
remove needs_ci_run script

nilsnolde:
~~it's failing again because `requests` isn't installed. let's remove it completely to remove that headache. it's not even simplifying things in this script, so it's really just overhead.~~

instead I removed it, it's breaking all the time silently and has hardly any impact even when it's working..


scripts/needs_ci_run
```diff
@@ -41,17 +43,22 @@ def last_build_successful():
         return False, None
 
     # jobs defined in .circleci/config.yml
-    JOBS = ['lint-build-debug', 'build-release', 'build-osx']
+    JOBS = ['lint-build-debug', 'build-release']
```
nilsnolde, line 46: don't ask me why.. in fact.. should I change the PR to remove the damn thing? so sick of it.. if there's smth not needing any CI for sure, I'd rather quickly cancel those jobs and force merge

nilsnolde:
delete it..

nilsnolde:
deleted the script instead

kevinkreiser (approved):
i hate precommit hooks but since circle doesnt support a list of paths that should skip ci but it does support skipping ci via the commit message we could make a custom commit hook that fails if:

1. the only changes happening are on files in paths marked unimportant for ci AND
2. if the commit message doesnt have the `[skip ci]` string in it

this would at least trigger a person to think oh yeah i should skip ci here. problem is it wo uld only help people who use precommit which is probably not everyone.

i personally like it simple with no trickery regarding should or shouldnt build a commit

nilsnolde:
That’d be fine too. I once customized a got command to “git draft” which automatically prepended “skip ci” but I never remembered to use it. It’s more transparent than a commit hook, but we can’t do that for people, they’ll have to do that. A quick and simple commit hook would go a long way for myself at least.

PR #4426, opened 2023-12-01
Removed old traffic docs

Trietes:
#4259 broke the documentation pipeline. The simple_traffic.md was removed, but it was still listed in the make file of the docs.


nilsnolde:
great thanks, was about to do that:)

Issue #4427, opened 2023-12-02
elevation problems


nilsnolde:
I just realized that https://github.com/valhalla/valhalla/pull/4279 introduced code that crazily logs on a planet build. Here the problematic part: 
https://github.com/valhalla/valhalla/blob/0e5770eb93bfc47376a2369f99152f747532a3f2/src/mjolnir/elevationbuilder.cc#L58-L63

Basically it prints each elevation postmark _if_ there's an error. I looked a bit at the logs of the planet build, this is more a log of my observations in the next 20 mins or so:
- the error is only set true when there's an elevation difference of > 256 m between postmarks, ~~which (I think) are set in intervals of 30 m~~ (correction, we're recording intervals of 1/32the of the way length). Intuitively this shouldn't happen on many ways, maybe some crazy roads in the Himalayas or Andes or very very long ways in even less alpine terrain. 
- in the logs I see ways like this (that's the full entry for this way, it's short):
    ```
    2023/12/02 15:43:57.711462 [WARN] Elevation delta -389 exceeds limit!
    2023/12/02 15:43:57.711490 [INFO] edge elevation wayid= 709619819
    2023/12/02 15:43:57.711498 [INFO]    108.411952
    2023/12/02 15:43:57.711503 [INFO]    11.269739
    2023/12/02 15:43:57.711507 [INFO]    95.647790
    2023/12/02 15:43:57.711512 [INFO]    81.675021
    2023/12/02 15:43:57.711516 [INFO]    41.417263
    ```
    which is this way: https://www.openstreetmap.org/way/709619819. The way is on an island VERY close to the sea, doubtful there's those crazy elevation differences. And even the logs don't show differences > 256 m anywhere, smth is strange.

Not sure why it thinks there's a huge elevation difference on that short way, the data doesn't seem to agree.

Some others could be legit, e.g. https://www.openstreetmap.org/way/377609526. It's super long (so huge intervals) and somewhere on the outskirts of Himalayas.

PR #4428, opened 2023-12-02
declutter elevation builder logs

nilsnolde:
less logging when building elevation, ref https://github.com/valhalla/valhalla/issues/4427


src/mjolnir/elevationbuilder.cc
```diff
@@ -56,10 +56,14 @@ std::vector<int8_t> encode_edge_elevation(const std::unique_ptr<valhalla::skadi:
...
-    LOG_INFO("edge elevation wayid= " + std::to_string(wayid));
-    for (const auto h : heights) {
-      LOG_INFO("   " + std::to_string(h));
+    double diff = 0;
+    for (size_t i = 1; i <= heights.size(); i++) {
+      auto d = std::abs(heights[i] - heights[i - 1]);
+      diff = d < diff ? diff : d;
+      LOG_DEBUG("  " + std::to_string(heights[i - 1]));
     }
+    LOG_WARN("edge elevation wayid = " + std::to_string(wayid) + " exceeds difference with " +
```
nilsnolde, line 65: only this line will be printed by default. I have no feeling if this will happen a lot. it'd be nice to have the way id in the logs when this happens, since encoded elevations can be really really off for those ways. good in case issues roll in regarding that. 

I'll try one planet with this and see how noisy this is now.

valhalla/midgard/elevation_encoding.h
```diff
@@ -86,14 +86,12 @@ inline std::vector<int8_t> encode_elevation(const std::vector<double>& elevation
     auto delta = v - prior;
     if (delta > 127) {
       if (delta > 256) {
-        LOG_WARN("Elevation delta " + std::to_string(delta) + " exceeds limit!");
```
nilsnolde, line 89: this is printed now in the calling functions

kevinkreiser (approved):

nilsnolde:
forcing merge as it's just missing the recent spelling pr

nilsnolde:
I just built the planet, elevation builder still produces > 500kB of logs which is much better but still pretty noisy.. Also lots of `2023/12/15 18:03:07.028585 [WARN] edge elevation wayid = 0 exceeds difference with 78.788402 meters.`. Not sure where the wayid = 0 comes from, looks a bit fishy

PR #4430, opened 2023-12-03
attempt a cleanup of unused vehicle types

### LINKED ISSUE #4429: Cleanup vehicle types
I realized that we're having a concept of vehicle types that's not really being used and also isn't documented, e.g. a `type` request option which can take values of  `four_wheel_drive` etc but for car that can even be `motorcycle` which seems weird. Those travel types are hardly ever used anywhere and with the exception of `bicycle_type` it's not even documented that they exist.

#### Car

```
  const std::string& type = costing_options.transport_type();
  if (type == "motorcycle") {
    type_ = VehicleType::kMotorcycle;
    surface_factor_ = 1.0f;
  } else if (type == "bus") {
    type_ = VehicleType::kBus;
  } else if (type == "tractor_trailer") {
    type_ = VehicleType::kTractorTrailer;
  } else if (type == "four_wheel_drive") {
    type_ = VehicleType::kFourWheelDrive;
    surface_factor_ = 0.0f;
  } else {
    type_ = VehicleType::kCar;
  }
  ```

Here I'd like to comment out the whole thing, and just do `type_ = VehicleType::kCar`. `kTractorTrailer` is confusingly our truck type as well, and `kMotorcycle` is our motorcycle type, both shouldn't be available in AutoCosting. The `kFourWheelDrive` does do a tiny little bit, but that's also a bit too little to be called 4WD maybe?

##### Pedestrian

```
  const std::string& type = costing_options.transport_type();
  if (type == "wheelchair") {
    type_ = PedestrianType::kWheelchair;
  } else if (type == "segway") {
    type_ = PedestrianType::kSegway;
  } else {
    type_ = PedestrianType::kFoot;
  }
``` 

Again, nothing here is actually used other than `kFoot`. Also wheelchair not, other than for the access mask. For transit we expose a `wheelchair` bool request parameter, i.e. not even transit does anything with this wheelchair `transport_type`. A lot more should be done and I'd rather remove it than document it and have people use it, as it's really not remotely usable for any person in a wheelchair. Segway haha I'd also dump that;)

### PR DESCRIPTION
**Pull request by nilsnolde**
fixes #4429 

This is pretty opinionated, feel free to reject some of it.

### CODE REVIEW

File: src/sif/autocost.cc [hunk 1]
```diff
@@ -697,7 +684,7 @@ void ParseAutoCostOptions(const rapidjson::Document& doc,
   const auto& json = rapidjson::get_child(doc, costing_options_key.c_str(), dummy);
 
   ParseBaseCostOptions(json, c, kBaseCostOptsConfig);
-  JSON_PBF_DEFAULT(co, kDefaultAutoType, json, "/type", transport_type);
+  // JSON_PBF_DEFAULT(co, kDefaultAutoType, json, "/type", transport_type);
```
nilsnolde, line 687: I left this here, it's very likely we'll use it at some point. I might have case for RV routing soon-ish, though not sure if that wouldn't be better done inside truck costing.

proto/common.proto
```diff
@@ -361,15 +361,15 @@ enum VehicleType {
...
+  kTruck = 3;
   kMotorScooter = 4;
 }
 
 // TODO: review and update as needed
 enum PedestrianType {
   kFoot = 0;
   kWheelchair = 1;
-  kSegway = 2;
+  // kSegway = 2;  // deprecated
```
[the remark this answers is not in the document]
nilsnolde: delete

nilsnolde:
reviewed and agreed to remove some more from the code, instead of commenting

nilsnolde:
@kevinkreiser I cleaned up some more things as we talked yesterday, just removing the commented out code entirely

kevinkreiser (dismissed):

src/odin/enhancedtrippath.cc
```diff
@@ -122,7 +122,7 @@ const std::string& TripLeg_TravelMode_Name(int v) {
 
 const std::string& TripLeg_VehicleType_Name(int v) {
   static const std::unordered_map<int, std::string> values{
-      {0, "kCar"}, {1, "kMotorcycle"}, {2, "kAutoBus"}, {3, "kTractorTrailer"}, {4, "kMotorScooter"},
+      {0, "kCar"}, {1, "kMotorcycle"}, {2, "kAutoBus"}, {3, "kTruck"}, {4, "kMotorScooter"},
```
kevinkreiser, line 125: should update the docs to explain the defaults are in line with what is typically called a semi-trailer truck

nilsnolde: I guess that’s kinda implicit by the default values for dimensions etc, but wouldn’t hurt to spell it out

nilsnolde:
merged master, cowboying the commit

PR #4431, opened 2023-12-03
add access restrictions to /locate response

nilsnolde:
While debugging some issue I was missing the access restrictions in the `/locate` response. I added most, except the time-specific restrictions, there it's a `"value": null` in the response. Adding time domains would add a bit more code. I'd leave it out for now and add it in a separate PR if we want that too.


src/baldr/accessrestriction.cc
```diff
@@ -43,6 +61,37 @@ void AccessRestriction::set_value(const uint64_t v) {
...
+                        {"wheelchair", static_cast<bool>(modes_ & kWheelchairAccess)},
+                        {"moped", static_cast<bool>(modes_ & kMopedAccess)},
+                        {"motorcycle", static_cast<bool>(modes_ & kMotorcycleAccess)}});
+
+  switch (type()) {
+    case AccessType::kTimedAllowed:
+    case AccessType::kTimedDenied:
+    case AccessType::kDestinationAllowed:
+      // TODO(nils): turn the time domain into a map
+      map->emplace("value", nullptr);
```
kevinkreiser, line 83: i would wait on this. it would be better to either just do this now or let this out of the pr to avoid any api breakage questioning

nilsnolde: this is more work than anticipated.. basically we'd have to reverse the parsing process of OSM tags and output smth similar to it to make it (intuitively) usable
```
    // Mo-Fr 06:00-11:00,17:00-19:00; Sa 03:30-19:00
    // Apr-Sep: Mo-Fr 09:00-13:00,14:00-18:00; Apr-Sep: Sa 10:00-13:00
    // Mo,We,Th,Fr 12:00-18:00; Sa-Su 12:00-17:00
    // Feb#16-Oct#15 09:00-18:30; Oct#16-Nov#15: 09:00-17:30; Nov#16-Feb#15: 09:00-16:30
 ```
as you can imagine that'd be ugly code. and a lot of it. making it saner in the code would make it much less usable or very big as response.

but I'd honestly still like to know if there's a time-based restriction, even the value is not provided. if anyone would ever tackle this, it'll need to be a (nested) map for the value for sure. so IMHO providing an empty map right now would still be fine for the future. we could also leave out the `value` property, then those time-based ones would show up but not have that field at all (but all other restrictions would). I'd prefer the empty map, but not too strongly.

wdyt @kevinkreiser ?

Comment (gknisely) [reply src/baldr/accessrestriction.cc:83]: who wrote this....oh wait nm ;-)

Comment (nilsnolde) [reply src/baldr/accessrestriction.cc:83]: haha this time shit is just so disgusting to work with..

Comment (gknisely) [reply src/baldr/accessrestriction.cc:83]: yes it was not fun at all

Comment (gknisely) [reply src/baldr/accessrestriction.cc:83]: should be able to "build" a string based on the `ComplexRestriction` logic

Comment (kevinkreiser) [reply src/baldr/accessrestriction.cc:83]: i mean yes the ideal would be to somewhat recreate the human readable string that we have in osm but the whole point of the parsing it to  normalize the data into structures that always have the same shape regardless of the fugly tagging that we had to regex our way through.

in the end we get this struct right:
https://github.com/valhalla/valhalla/blob/d5338866e66759f8afa6f9b500e825eb818427a8/valhalla/baldr/timedomain.h#L314-L326

why not just serialize out each piece of that struct. yes its not super easy to read but it has all the info and would take about 10 minutes to write the serializer code. what am i missing?

Comment (nilsnolde) [reply src/baldr/accessrestriction.cc:83]: is there anything that serializes the time domain of complex restrictions? couldn't find anything quickly.

Comment (kevinkreiser) [reply src/baldr/accessrestriction.cc:83]: actually dont we have to turn this into a date_time object when we check it in dynamic cost? can we do the same thing that that does, turn begin and end into date times for the YMD case and just show the human strings for that.

for the day of the week case i guess we'd just want the day name for start and end and the time? surely there are clues in costing on how its interpreted but also we could just print out each field and call it done

Comment (kevinkreiser) [reply src/baldr/accessrestriction.cc:83]: a compromise would be to have a map with just the raw value in it:

```json
{
  "value": 23049872312
}
```

and then down the road we can add "begin" and "end" date time strings when we have the time (pun intended)

gknisely: Also, we should grab from `ComplexRestriction` because we do toss malformed input from the user.  We may not toss everything as parts could be correct.  We do also toss unsupported types like `PH` for public holidays.

nilsnolde: 10 mins is vastly exaggerated for any sort of half-usable output, but yeah prob not more than 2-3 hours with some fanciness added. I'm also fine with the compromise of returning the raw value. we can keep it forever as well, even when we add the actual map. let's do that then for now. the next weeks should be pretty calm, maybe I find the motivation to flesh it out a bit more.

kevinkreiser (dismissed):

src/baldr/accessrestriction.cc
```diff
@@ -1,6 +1,24 @@
 #include "baldr/accessrestriction.h"
+#include "baldr/timedomain.h"
 #include <string.h>
 
+namespace vb = valhalla::baldr;
+
+namespace {
+std::unordered_map<vb::AccessType, std::string> type_to_string = {
```
kevinkreiser, line 8:
```suggestion
const std::unordered_map<vb::AccessType, std::string> type_to_string = {
```

kevinkreiser (approved):

src/baldr/accessrestriction.cc
```diff
@@ -43,6 +61,37 @@ void AccessRestriction::set_value(const uint64_t v) {
   value_ = v;
 }
 
+const json::MapPtr AccessRestriction::json() const {
+  auto map = json::map({{"type", type_to_string[type()]},
```
kevinkreiser, line 65:
```suggestion
  auto maybe_found = type_to_string.find(type());
  std::string restriction_type = "unsupported";
  if (maybe_found != type_to_string.cend()) {
    restriction_type = maybe_found->second;
  }
  auto map = json::map({{"type", restriction_type},
```

kevinkreiser: this will happen when we add a new type and old code is accessing it

nilsnolde: good point! thanks!

kevinkreiser (approved):

PR #4432, opened 2023-12-04
add shape to matrix response

### LINKED ISSUE #4300: (optionally) return the matrix route geometries
Since we already have the `RecostPaths` for the bidirectional dijkstra, it'd be quite natural to also return the path geometries for all connections. It's hardly any computation needed and would go a long way making sense of the results together with the new-ish /expansion for matrix.

nilsnolde:
fixes #4300 

based on #4372 

implements a new `shape_format` request field for `/sources_to_targets` with all available formats (polyline5/6, geojson and by default `"none"`), but only for CostMatrix. `shape_format` will dictate what the `shape` response field contains, if not `"none"` in which case there is no `shape` response field.

**note**: trivial routes have [a bug](https://github.com/valhalla/valhalla/issues/4433) in master (and the PR this is based on), where it finds the connection on the wrong directed edge, consistently. I guess this went unnoticed for so long because it hardly matters for the matrix. the distance will always be the same, thought the duration can of course vary. I'll band-aid that for now.


proto/options.proto
```diff
@@ -444,7 +445,9 @@ message Options {
   oneof has_use_timestamps {
     bool use_timestamps = 37;                                      // Use timestamps to compute elapsed time for trace_route and trace_attributes [default = false]
   }
-  ShapeFormat shape_format = 38;                                   // Shape format (defaults to polyline6 encoding)
+  oneof has_shape_format {
```
nilsnolde, line 448: had to make a `oneof` so matrix can default to false

src/midgard/util.cc
```diff
@@ -128,6 +128,20 @@ void trim_shape(float start,
   }
 }
 
+// Generate leg shape in geojson format.
+baldr::json::MapPtr geojson_shape(const std::vector<PointLL> shape) {
```
nilsnolde, line 132: this was in the osrm serializer before. moved it here, though we might actually have it better in `baldr/json.h`?

src/thor/costmatrix.cc
```diff
@@ -133,6 +136,7 @@ void CostMatrix::SourceToTarget(Api& request,
   mode_ = mode;
   costing_ = mode_costing[static_cast<uint32_t>(mode_)];
   access_mode_ = costing_->access_mode();
+  has_time_ = has_time;
```
nilsnolde, line 139: we'll use this when recosting & forming paths

src/thor/costmatrix.cc
```diff
@@ -250,26 +254,30 @@ void CostMatrix::SourceToTarget(Api& request,
     }
   }
 
-  if (has_time) {
-    RecostPaths(graphreader, source_location_list, target_location_list, time_infos, invariant);
-  }
```
nilsnolde, line 255: moved this into the loop below

src/thor/costmatrix.cc
```diff
@@ -948,82 +956,93 @@ void CostMatrix::SetTargets(baldr::GraphReader& graphreader,
...
+                                       const bool invariant,
+                                       const ShapeFormat shape_format) {
+  // no need to look at source == target or missing connectivity
+  if ((!has_time_ && shape_format == no_shape) || connection.cost.secs == 0.f ||
+      connection.distance == kMaxCost) {
+    return "";
+  }
 
-    uint32_t source_idx = idx / static_cast<uint32_t>(targets.size());
-    uint32_t target_idx = idx % static_cast<uint32_t>(targets.size());
```
nilsnolde, line 968: this is a terrible diff: I changed no logic here. just removed the for loop (as the caller is doing that)

src/thor/costmatrix.cc
```diff
@@ -948,82 +956,93 @@ void CostMatrix::SetTargets(baldr::GraphReader& graphreader,
...
+  if (path_edges.size() == 1 && source_pct > target_pct) {
+    // it found the wrong direction, so let's turn that around
+    auto opp_id = graphreader.GetOpposingEdgeId(path_edges[0]);
+    path_edges.clear();
+    path_edges.emplace_back(opp_id);
+    source_pct = 1.f - source_pct;
+    target_pct = 1.f - target_pct;
+  }
+  // recost the path if this was a time-dependent expansion
+  if (has_time_) {
```
nilsnolde, line 1045: we only need to do the below if there was a valid time

src/thor/costmatrix.cc
```diff
@@ -948,82 +956,93 @@ void CostMatrix::SetTargets(baldr::GraphReader& graphreader,
...
+  // TODO(nils): bug with trivial routes https://github.com/valhalla/valhalla/issues/4433
+  // remove this whole block below once that's fixed
+  if (path_edges.size() == 1 && source_pct > target_pct) {
+    // it found the wrong direction, so let's turn that around
+    auto opp_id = graphreader.GetOpposingEdgeId(path_edges[0]);
+    path_edges.clear();
+    path_edges.emplace_back(opp_id);
+    source_pct = 1.f - source_pct;
+    target_pct = 1.f - target_pct;
+  }
```
nilsnolde, line 1043: this is quite messed up, see https://github.com/valhalla/valhalla/issues/4433. I think it'd be easy-ish to fix before here and I'll do that. but this works quite well, even though you'll see in the test that there's some tiny floating point issue apparently

src/thor/costmatrix.cc
```diff
@@ -1032,22 +1051,59 @@ void CostMatrix::RecostPaths(GraphReader& graphreader,
...
+  // return the shape of the path if requested
+  if (shape_format == no_shape)
+    return "";
+
+  auto source_vertex = PointLL{source_edge.ll().lng(), source_edge.ll().lat()};
+  auto target_vertex = PointLL{target_edge.ll().lng(), target_edge.ll().lat()};
+  std::vector<PointLL> points;
+  for (const auto& path_edge : path_edges) {
+    auto is_first_edge = path_edge == path_edges.front();
+    auto is_last_edge = path_edge == path_edges.back();
```
nilsnolde, line 1076: it's more or less the same logic as in triplegbuilder

src/thor/matrix_action.cc
```diff
@@ -36,7 +36,8 @@ std::string thor_worker_t::matrix(Api& request) {
   auto costmatrix = [&](const bool has_time) {
     return costmatrix_.SourceToTarget(request, *reader, mode_costing, mode,
                                       max_matrix_distance.find(costing)->second, has_time,
-                                      options.date_time_type() == Options::invariant);
+                                      options.date_time_type() == Options::invariant,
+                                      options.shape_format());
```
nilsnolde, line 40: actually why do we even pass these things here.. why also have the full `request` as argument, it's kinda unnecessary

src/tyr/matrix_serializer.cc
```diff
@@ -43,6 +43,32 @@ json::ArrayPtr serialize_distance(const valhalla::Matrix& matrix,
   }
   return distance;
 }
+
+json::ArrayPtr serialize_shape(const valhalla::Matrix& matrix,
```
nilsnolde, line 47: I tried to do this whole thing a bit nicer in the serializer, but the fact that geojson & polyline are different types (map vs string), I couldn't come up with smth nicer without a lot more changes. honestly this whole serializer needs a complete overhaul. it's just hard to come up with terse/elegant code for as many options as we provide for the response :)

src/tyr/route_serializer_osrm.cc
```diff
@@ -316,6 +302,10 @@ std::vector<PointLL> simplified_shape(const valhalla::DirectionsRoute& direction
 void route_geometry(json::MapPtr& route,
                     const valhalla::DirectionsRoute& directions,
                     const valhalla::Options& options) {
+  if (options.shape_format() == no_shape) {
+    return;
+  }
```
nilsnolde, line 307: keep or revert? it's not documented for /route anyways;)

test/gurka/test_matrix_time_dependent.cc
```diff
@@ -396,3 +398,123 @@ TEST(StandAlone, CostMatrixDeadends) {
     check_matrix(res_doc, {0.8f}, false, Matrix::CostMatrix);
   }
 }
+
+TEST(StandAlone, CostMatrixShapes) {
+  // keep the same order in the map.nodes for encoding easily
+  const std::string ascii_map = R"(
+    A-B-C-D-E-F-G-H-I-J-K-------L
```
nilsnolde, line 405: it's 3 ways but in order to keep the order in the `map.nodes`  `std::map` I made them contiguous, making it harder to see what's happening

test/gurka/test_matrix_time_dependent.cc
```diff
@@ -396,3 +398,123 @@ TEST(StandAlone, CostMatrixDeadends) {
...
+                           .GetArray()[0]
+                           .GetArray()[0]
+                           .GetObject()["shape"];
+  EXPECT_TRUE(gj_shp.IsObject());
+  EXPECT_EQ(gj_shp["coordinates"].GetArray().Size(), 12);
+  EXPECT_EQ(gj_shp["type"], "LineString");
+  res.erase();
+
+  // trivial route
+  // has a bug: https://github.com/valhalla/valhalla/issues/4433, but it's band-aided for now
```
nilsnolde, line 494: I'd hope once that bug is fixed, that this will resolve itself

proto/options.proto
```diff
@@ -41,6 +41,7 @@ enum ShapeFormat {
   polyline6 = 0;
   polyline5 = 1;
   geojson = 2;
+  no_shape = 3;  // to omit returning the geometry, e.g. default in matrix
```
kevinkreiser, line 44: nit: 
```suggestion
  none = 3;  // to omit returning the geometry, e.g. default in matrix
```

in worker.cc dont we have to protect against this value when its not a matrix action

src/midgard/util.cc
kevinkreiser: midgard should not have access to baldr, if it does, which this suggests it does then we have lost the plot :smile: 

since this is for serialization why not put it in the tyr serializers header?

src/thor/costmatrix.cc
```diff
@@ -29,10 +30,11 @@ bool equals(const valhalla::LatLng& a, const valhalla::LatLng& b) {
          (!a.has_lat_case() || a.lat() == b.lat()) && (!a.has_lng_case() || a.lng() == b.lng());
 }
 
-inline float find_percent_along(const valhalla::Location& location, const GraphId& edge_id) {
+inline valhalla::PathEdge find_correlated_edge(const valhalla::Location& location,
```
kevinkreiser, line 33:
```suggestion
inline const valhalla::PathEdge& find_correlated_edge(const valhalla::Location& location,
```

src/thor/costmatrix.cc
```diff
@@ -124,7 +126,8 @@ void CostMatrix::SourceToTarget(Api& request,
                                 const sif::travel_mode_t mode,
                                 const float max_matrix_distance,
                                 const bool has_time,
-                                const bool invariant) {
+                                const bool invariant,
+                                const ShapeFormat shape_format) {
```
kevinkreiser, line 130:
```suggestion
                                const ShapeFormat& shape_format) {
```

proto/options.proto
nilsnolde: let's see if that works. there's already a `none` enum member and the way protobuf works is they are siblings, pretty sure that'll clash. but let's see:)

nilsnolde: > in worker.cc dont we have to protect against this value when its not a matrix action

I did that for `/height` (you probably didn't get to worker.cc yet..), which is the only documented use of `shape_format`. We do use it for OSRM as well, but it's not documented and other than mapbox I'm sure no one knows. I even added a bit of code to use `no_shape` for OSRM as well. Eventually we should enable `shape_format` also for the default Valhalla format. I find it kinda useful to disable getting the geometry from a route call, same as the narrative, it's the second largest payload item and some might not be interested in the geometry.

nilsnolde: yeah that's what I meant @kevinkreiser :
```
[build] options.proto:44:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it
```

stupid protobuf..

should we just leave `no_shape`, it's short enough and very exclicit.

nilsnolde:
were you done with the review on this one @kevinkreiser? can't remember.. previous comments are addressed where possible.

proto/options.proto
kevinkreiser: yeah i thought this might be the case but it was worth asking.

shape_format=no_shape i just find redundant, but its fine.

kevinkreiser: though i do think we can change the nesting to accomplish this if we really want

src/thor/costmatrix.cc
kevinkreiser: :pray:

valhalla/thor/costmatrix.h
```diff
@@ -302,11 +308,15 @@ class CostMatrix {
    * @param   time_infos   The time info objects for the sources
    * @param   invariant    Whether time is invariant
```
kevinkreiser, line 309: forgot to update docstring to say that its returning the shape if it was requested

src/thor/costmatrix.cc
```diff
@@ -1035,22 +1054,59 @@ void CostMatrix::RecostPaths(GraphReader& graphreader,
...
-      continue;
+      return "";
     }
 
     // update the existing best_connection cost
-    best_connection->cost = new_cost;
+    connection.cost = new_cost;
+  }
+
+  // return the shape of the path if requested
```
kevinkreiser, line 1070: nit
```suggestion
  // bail if no shape was requested
```

src/thor/matrix_action.cc
kevinkreiser: im sure its a result of things slowy changing over time. i appreciate resisting the urge to refactor it as part of this though i agree its stupid :smile:

src/tyr/matrix_serializer.cc
```diff
@@ -43,6 +43,32 @@ json::ArrayPtr serialize_distance(const valhalla::Matrix& matrix,
...
+                               const size_t start_td,
+                               const size_t td_count,
+                               const ShapeFormat shape_format) {
+  // TODO(nils): shapes aren't implemented yet in TDMatrix
```
kevinkreiser, line 51: does this mean that we disable them further up in the code or that timedistance requests make it here and serialize an empty string?

src/tyr/matrix_serializer.cc
```diff
@@ -118,6 +145,17 @@ json::ArrayPtr serialize_row(const valhalla::Matrix& matrix,
       if (!date_time.empty()) {
         map->emplace("date_time", date_time);
       }
+      // TODO(nils): shapes aren't implemented yet in TDMatrix
+      if (matrix.shapes().size() && !matrix.shapes()[i].empty()) {
+        switch (shape_format) {
+          case geojson:
+            map->emplace("shape",
```
kevinkreiser, line 152: ok so for this serializer every entry now gets an optional shape output. i guess for failed pairs even if shape was requested we dont output anything?

kevinkreiser: i guess my point is that disagrees with the comment in the valhalla serializer about always having an entry null or not

src/tyr/matrix_serializer.cc
```diff
@@ -43,6 +43,32 @@ json::ArrayPtr serialize_distance(const valhalla::Matrix& matrix,
...
+    return shapes;
+
+  for (size_t i = start_td; i < start_td + td_count; ++i) {
+    switch (shape_format) {
+      // even if it source == target or no route found, we want to emplace an element
+      case geojson:
+        if (!matrix.shapes()[i].empty())
+          shapes->emplace_back(tyr::geojson_shape(decode<std::vector<PointLL>>(matrix.shapes()[i])));
+        else
+          shapes->emplace_back(json::map({}));
```
kevinkreiser, line 63: why not `null` i think its better than empty map:

```suggestion
          shapes->emplace_back(nullptr);
```

src/tyr/matrix_serializer.cc
```diff
@@ -118,6 +145,17 @@ json::ArrayPtr serialize_row(const valhalla::Matrix& matrix,
       if (!date_time.empty()) {
         map->emplace("date_time", date_time);
       }
+      // TODO(nils): shapes aren't implemented yet in TDMatrix
+      if (matrix.shapes().size() && !matrix.shapes()[i].empty()) {
```
kevinkreiser, line 149: i question this, why isnt:

```suggestion
      if (matrix.shapes().size()) {
```

and then further down we emplace a `nullptr` when its empty

kevinkreiser: my assumption is that size == 0 when you didnt want shape and that if you have an empty entry its because of a disconnected pair (which we still need column form in the row)

src/tyr/route_serializer_osrm.cc
kevinkreiser: i th ink its good to keep

valhalla/thor/matrix_common.h
```diff
@@ -126,6 +126,7 @@ inline void reserve_pbf_arrays(valhalla::Matrix& matrix, size_t size) {
   matrix.mutable_distances()->Resize(size, 0U);
   matrix.mutable_times()->Resize(size, 0U);
   matrix.mutable_date_times()->Reserve(size);
+  matrix.mutable_shapes()->Reserve(size);
```
kevinkreiser, line 129: good call!

kevinkreiser:
some tiny things to change otherwise looks done

proto/options.proto
nilsnolde: > shape_format=no_shape i just find redundant, but its fine.

it's necessary for omitting `shape` from a route response no? at least I don't think we should change the default for osrm format. or how else would you do it?

src/thor/matrix_action.cc
nilsnolde: this is my fail anyways:)

src/tyr/matrix_serializer.cc
nilsnolde: yeah let's do that

src/tyr/matrix_serializer.cc
nilsnolde: yes, for the verbose version (default), we don't even put a shape, also not for failed pairs. for the terser verbose=false, we do have to keep the `nullptr` in the response so it's properly parsable.

src/tyr/serializers.cc
```diff
@@ -249,6 +249,20 @@ std::string serializePbf(Api& request) {
...
+baldr::json::MapPtr geojson_shape(const std::vector<PointLL> shape) {
+  auto geojson = baldr::json::map({});
+  auto coords = baldr::json::array({});
+  coords->reserve(shape.size());
+  for (const auto& p : shape) {
+    coords->emplace_back(
+        baldr::json::array({baldr::json::fixed_t{p.lng(), 6}, baldr::json::fixed_t{p.lat(), 6}}));
+  }
+  geojson->emplace("type", std::string("LineString"));
+  geojson->emplace("coordinates", std::move(coords));
```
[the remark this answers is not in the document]
nilsnolde: right, I'll change that then

valhalla/midgard/util.h
```diff
@@ -15,6 +15,7 @@
 #include <utility>
 #include <vector>
 
+#include <valhalla/baldr/json.h>
```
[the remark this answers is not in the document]
nilsnolde: oops, must be a relict from before. I moved the function which needed it to tyr/serializers.h

proto/options.proto
kevinkreiser: oh jeez im sorry i just meant the naming is redundant i get that we need a way to say we dont want shape. `shape_format=shapity_shape_shape_shaperino` zum beispiel :wink: we could name it `disabled` or `off` or something but i dont care that much im just very into language stuff

kevinkreiser:
just a little clang tidy and then done:
image

nilsnolde:
oops done

kevinkreiser:
merge master and update changelog yet

src/worker.cc
```diff
@@ -806,10 +806,17 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
       options.set_shape_format(polyline5);
     } else if (*shape_format == "geojson") {
       options.set_shape_format(geojson);
+    } else if (*shape_format == "none") {
```
kevinkreiser, line 809: oh i didnt notice this in the json api it is still called `none`

src/worker.cc
```diff
@@ -806,10 +806,17 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
...
+      if (action == Options::height) {
+        throw valhalla_exception_t{164};
+      }
```
kevinkreiser, line 812: its not clear to me why we need this. the height action will never look at `options.shape_format()` right? i think we shouldnt throw errors for superfluous parameters. or what am i missing?

kevinkreiser (approved):
3 small things then merge:

1. changelog entry at bottom of list
2. merge master into branch
3. check if we can remove height action error response

src/worker.cc
nilsnolde: > the height action will never look at options.shape_format() right?

yes it does actually, to return either polyline5 or 6. in fact it's the only documented use of `shape_format` :)

src/worker.cc
nilsnolde: hm that was the case since the beginning. but you're right, it's unintuitive to have them differ in PBF/JSON formats, I'll change it

src/worker.cc
kevinkreiser: not to return though right! its the input format in that case hahaha, i had forgotten about that!

nilsnolde:
last things done, sorry for the hassle..

src/worker.cc
kevinkreiser: we probably have to block this then for mapmatching too right?

kevinkreiser: i shouldnt say i didnt check this but we should have a look quick while we are here

kevinkreiser: https://github.com/valhalla/valhalla/blob/253443ed2b3c7838a85be0d2b89e1876a7d542d4/src/worker.cc#L839-L852

kevinkreiser: looks like mapmatching forces to polyline6. kind of dumb, we should allow it but its a problem for another day

kevinkreiser (approved):

src/worker.cc
nilsnolde: for height as input, right.. for map matching not relevant, it's hard-coded to polyline6 it seems in worker.cc

nilsnolde:
hmpf.. tidy again and a test was broken after the recent changes.. I also remembered I'd like to add a warning when TDMatrix is used but shape_format was requested. since the approve is dismissed anyways, I'll add now.

don't hurry with this really, let's take it on monday real quick.

kevinkreiser (approved):

PR #4435, opened 2023-12-05
Remove warnings from changed files on nn-matrix-geometry

cvvergara:
# Issue

## Tasklist

 - [x] Remove warnings
 - [x] Header/code files separate into header and code files

## Requirements / Relations


valhalla/tyr/route_serializer_osrm.h
```diff
@@ -0,0 +1,44 @@
+#ifndef VALHALLA_TYR_ROUTE_SERIALIZER_OSRM_
+#define VALHALLA_TYR_ROUTE_SERIALIZER_OSRM_
+#pragma once
+
+#include "proto_conversions.h"
+
+namespace osrm_serializers {
```
nilsnolde, line 7: I can't say much to this, not sure why it was a private header/source file before. IMO it's good to separate it, but maybe there was some logic behind it initially @kevinkreiser ?

valhalla/tyr/route_summary_cache.h
```diff
@@ -0,0 +1,96 @@
...
+
+#include "proto/directions.pb.h"
+#include "proto/options.pb.h"
+#include "proto/trip.pb.h"
+#include "proto_conversions.h"
+
+namespace valhalla {
+namespace tyr {
+
+struct NamedSegment {
```
nilsnolde, line 12: here I'd be careful. when we have headers outside of the `valhalla` directory then that usually means they're considered private and only used for the compilation but not publicly installed/distributed.

src/tyr/route_serializer_osrm.cc
```diff
@@ -623,9 +592,9 @@ json::ArrayPtr intersections(const valhalla::DirectionsLeg::Maneuver& maneuver,
     // Add tunnel_name for tunnels
     if (!arrive_maneuver) {
       if (curr_edge->tunnel() && !curr_edge->tagged_value().empty()) {
-        for (uint32_t t = 0; t < curr_edge->tagged_value().size(); ++t) {
-          if (curr_edge->tagged_value().Get(t).type() == TaggedValue_Type_kTunnel) {
-            intersection->emplace("tunnel_name", curr_edge->tagged_value().Get(t).value());
+        for (const auto& e : curr_edge->tagged_value()) {
```
nilsnolde, line 595: yes, so much nicer to look at

valhalla/tyr/route_summary_cache.h
```diff
@@ -0,0 +1,96 @@
...
+// 1 (index 0)                        R2
+// 2 (index 1)                        R2, R3
+// 3 (index 2)                        R0, R2, R3
+// 4 (index 3)                        R0, R1, R2, R3
+//=============================================================================
+class route_summary_cache {
+  // vector 1: routes
+  // vector 2: legs for the route
+  // vector 3: named segments sorted by longest segment distance
+  std::vector<std::vector<std::vector<NamedSegment>>> route_leg_segs_by_dist;
```
cvvergara, line 73: This member use `NameSegment` so the definition must be on the header.

valhalla/tyr/route_summary_cache.h
cvvergara: Keeping on header, but as private member of route_summary_cache class

valhalla/tyr/route_summary_cache.h
cvvergara: In header but private to the class

valhalla/tyr/route_summary_cache.h
nilsnolde: hm, can you elaborate? I still see that this is in `./valhalla/tyr` directory, so it's still public as it'll be installed. what am I missing?

valhalla/tyr/route_serializer_osrm.h
nilsnolde: sometimes it's fine to not have a header I think, as e.g. `src/tyr/matrix_serializer.cc` which just implements smth from `valhalla/tyr/serializers.h`. what I really don't like is to **include** implementation files. but I might not see some hidden benefit of this practice?

PR #4439, opened 2023-12-07
Refactor EdgeLabel to reduce memory

### LINKED ISSUE #4437: EdgeLabel Size Issues
EdgeLabel (and its derived classes) has a couple of issues and places where its size can be slimmed down.

1. In the base EdgeLabel, one of the 64bit data fields has bit-field data members that add up to 65 bits (has_measured_speed_ puts this over 64 bits). This leads to an extra 8 byte word in the data. Fixing this can reduce the EdgeLabel size by 8 bytes and all derived classes benefit as well.
2. The base EdgeLabel class could be slimmed down by removing the transition_cost_ (only needed for recovering paths) and the distance (to destination) data member (only needed for A* based algorithms). Time-distance matrix, routing within map-matching, and a couple other places could benefit from moving these out of the base class.

dnesbitt61:
# Issue

#fixes #4437

Refactor EdgeLabel and derived classes to lower memory use. Fixed an overflow of one of the 64 bit data members (lowers memory use by 8 bytes in EdgeLabel). Moved transition_cost and distance out of base EdgeLabel class. This makes EdgeLabel smaller for use by time distance matrix, meili routing, and multimodal CanReachDestination method. Removed some unused Updated methods. The base EdgeLabel class was reduced to 40 bytes (was 56 bytes).
Added a new dervied class, PathEdgeLabel, for use in path recosting and LabelCallback.
Updated calls to EdgeLabel constructor and methods within thor's path algorithms and within path recosting.
Added unit tests to validate EdgeLabel size (and its derived classes). Hopefully this can catch cases where data members exceed the allocated bit size (e.g. bit fields within a 64 bit word data member for instance).
Updates astar_bss to use BDEdgeLabel (as do all other A* methods). 

Sizes:
EdgeLabel:         **40 bytes (was 56)**
PathEdgeLabel   48 bytes (new)
BDEdgeLabel     64 bytes (was 64) - though this now has 4 bytes "spare" without any size increase.
MMEdgeLabel    72 bytes (was 80)

Note that this impacts timedistancematrix (positive impact!) more than other classes (especially with pedestrian and bicycle matrices) since there are no heuristics to guide the search. 

## Tasklist

 - [ ] Add tests

## Requirements / Relations


dnesbitt61:
I considered moving sortcost out of the base EdgeLabel. However, since this is tied to double bucket priority queue (sortcost method), we would have to make this a virtual method which would negate the potential savings.

dnesbitt61:
Should BDEdgeLabel be derived from PathEdgeLabel? MMEdgeLabel is derived from PathEdgeLabel, perhaps BDEdgeLabel should be as well? At least we should be consistent. What is the preference?

valhalla/sif/edgelabel.h
```diff
@@ -478,21 +429,92 @@ class EdgeLabel {
...
   Cost cost_;      // Cost and elapsed time along the path.
   float sortcost_; // Sort cost - includes A* heuristic.
-  float distance_; // Distance to the destination.
+};
 
-  // Was originally used for reverse search path to remove extra time where paths intersected
-  // but its now used everywhere to measure the difference in time along the edge vs at the node
+/**
+ * Derived label class used for recosting paths within the LabelCallback.
+ * transition_cost is added to the label for use when recoting a path.
```
kevinkreiser, line 441: :+1: great observation, it never occurred to me when i added recosting!

src/thor/astar_bss.cc
```diff
@@ -194,7 +194,7 @@ void AStarBSSAlgorithm::ExpandForward(GraphReader& graphreader,
     // less cost the predecessor is updated and the sort cost is decremented
     // by the difference in real cost (A* heuristic doesn't change)
     if (current_es->set() == EdgeSet::kTemporary) {
-      EdgeLabel& lab = edgelabels_[current_es->index()];
+      auto& lab = edgelabels_[current_es->index()];
```
kevinkreiser, line 197: excellent call, makes future refactors less painful :wink:

src/thor/multimodal.cc
```diff
@@ -740,21 +740,19 @@ bool MultiModalPathAlgorithm::ExpandFromNode(baldr::GraphReader& graphreader,
 
     // Check if lower cost path
     if (es->set() == EdgeSet::kTemporary) {
-      EdgeLabel& lab = edgelabels[es->index()];
+      auto& lab = edgelabels[es->index()];
       if (newcost.cost < lab.cost().cost) {
-        float newsortcost = lab.sortcost() - (lab.cost().cost - newcost.cost);
```
kevinkreiser, line 745: `newcost.cost` is now equivalent to this?

kevinkreiser:
looks good, since its a very core datastructure with a long constructor/update function that has a lot of parameters that are all implicitly castable to each other should we do any test requests in addition to what the tests are already covering? would also be cool to see if perf changed and report on it

src/thor/multimodal.cc
dnesbitt61: this is the same as what is changed in timedistancematrix code. In these cases there is no A* heuristsic, so the sort cost is the new cost. This basically triggers the adjacency list method to decrease the sort cost in the queue.

kevinkreiser (dismissed):

dnesbitt61:
I compared times and distances within valhalla_run_matrix for my test case. No changes between master and this branch.

valhalla/sif/edgelabel.h
nilsnolde, line 441: nit: 
```suggestion
 * transition_cost is added to the label for use when recosting a path.
```

src/thor/astar_bss.cc
```diff
@@ -98,7 +98,7 @@ void AStarBSSAlgorithm::Init(const midgard::PointLL& origll, const midgard::Poin
 // edges if from_transition is false.
 void AStarBSSAlgorithm::ExpandForward(GraphReader& graphreader,
                                       const GraphId& node,
-                                      const EdgeLabel& pred,
+                                      const BDEdgeLabel& pred,
```
nilsnolde, line 101: not sure how to name this better.. this is a unidir algo. it's also not only for a*, we use it in bidir (dijkstra) matrix too. nothing to take action, just wondering out loud.

nilsnolde (approved):
nice cleanup @dnesbitt61 thanks!

valhalla/sif/edgelabel.h
kevinkreiser: strange must be disabled somehow for this file?

src/thor/astar_bss.cc
kevinkreiser: we could name it `astar_label` as its the one that is "directed" ie with the heuristic, to differentiate it from the undirected one. though we do have other labels like mmlabel that are even more algorithm specific so mmlabel is better in that case imho. naming is hard :smile:

nilsnolde: > we could name it astar_label as its the one that is "directed" ie with the heuristic

like I said, it's also used in costmatrix. it's actually both a bidir edge label and a* edge label so it covers most algos.

anyways, dave actually left a comment. 
```
 * Derived EdgeLabel class used for A* path algorithms and CostMatrix.
 * NOTE - despite the name, this is also used by unidirectional A* algorithms
 * (not simply bidirectional A*).
 ```

valhalla/sif/edgelabel.h
nilsnolde: doesn't seem so, but :shrug:

PR #4450, opened 2023-12-11
hgv destonly

### LINKED ISSUE #4408: hgv:destination support in directed edge
There were some discussion I can't find anymore, where we agreed to add a dest_only flag for HGV. A lot of roads (in Europe at least) allow trucks to use otherwise inaccessible roads for loading/unloading goods etc, which makes it necessary to support this.

I didn't really look at the code yet. It's a bit of work, but shouldn't be too bad. 

We have 4 bits left in DirectedEdge, IMO using one here is being put to good use. Any contrary opinions?

nilsnolde:
fixes #4408


src/thor/bidirectional_astar.cc
```diff
@@ -336,11 +338,13 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
     }
     edgelabels_reverse_.emplace_back(pred_idx, meta.edge_id, opp_edge_id, meta.edge, newcost,
                                      sortcost, dist, mode_, transition_cost, thru,
-                                     (pred.closure_pruning() || !costing_->IsClosed(meta.edge, tile)),
+                                     (pred.closure_pruning() || !costing_->IsClosed(opp_edge, tile)),
```
nilsnolde, line 341: this was bugging me before: why would we look at the traversed edge for closures when we're actually interested in the opposing for reverse search? the same goes for `not_thru` further above, but that might actually not be a directed property, not sure..

nilsnolde: same in all other algorithms. which makes me think this is not a copy/paste fail but intentional. I just don't really get it I guess..

nilsnolde:
this is actually done. all tests pass BUT the isochrone one.. it's ridiculous.. the bike isochrones are off by 300 m in length which doesn't make any sense.. I disabled all changes to dijkstras and to the labels and they STILL fail. for the life of me I can't see how isochrone results can be affected by any other change in this PR!

lua/graph.lua
```diff
@@ -1658,7 +1658,8 @@ function filter_tags_generic(kv)
      kv["link_type"] = kv["link_type"]
   end
 
-  kv["private"] = private[kv["access"]] or private[kv["motor_vehicle"]] or "false"
+  kv["private"] = private[kv["access"]] or private[kv["motor_vehicle"]] or private[kv["motorcar"]] or "false"
```
nilsnolde, line 1661: OMG.. and as I wrote it down it dawned on me.. I added `private[kv["motorcar"]]` (which makes sense IMO). reverting that lets the isochrones succeed.

nilsnolde: bike is affected by normal destonly in that it adds a 10 mins penalty btw.

nilsnolde: I also verified it in the PBF: there's some `motorcar=destination` just around the place where the geometry was impacted. Instead of reverting that change, I updated the isochrone geometries.

valhalla/mjolnir/osmway.h
```diff
@@ -2663,8 +2679,9 @@ struct OSMWay {
   uint16_t use_sidepath_ : 1;
   uint16_t bike_forward_ : 1;
   uint16_t bike_backward_ : 1;
-  bool lit_ : 1;
```
nilsnolde, line 2666: that was a fail in the `lit` PR earlier this year. luckily only on this temp structure

valhalla/sif/edgelabel.h
```diff
@@ -72,19 +73,21 @@ class EdgeLabel {
...
-        dest_only_(edge->destonly()), origin_(0), destination_(0), toll_(edge->toll()),
-        not_thru_(edge->not_thru()), deadend_(edge->deadend()),
+        origin_(0), destination_(0), toll_(edge->toll()), not_thru_(edge->not_thru()),
+        deadend_(edge->deadend()),
         on_complex_rest_(edge->part_of_complex_restriction() || edge->start_restriction() ||
                          edge->end_restriction()),
         closure_pruning_(closure_pruning), path_id_(path_id), restriction_idx_(restriction_idx),
         internal_turn_(static_cast<uint8_t>(internal_turn)), unpaved_(edge->unpaved()),
         has_measured_speed_(has_measured_speed), cost_(cost), sortcost_(sortcost) {
+    dest_only_ = destonly ? destonly : edge->destonly();
```
nilsnolde, line 90: I took this out of the initializer list so no EdgeLabel usage has to be updated unless it's interested in using HGV destonly

lua/graph.lua
```diff
@@ -1658,7 +1658,8 @@ function filter_tags_generic(kv)
...
+  kv["private_hgv"] = private[kv["hgv"]] or private[kv["private"]] or "false"
```
nilsnolde, line 1662:
```suggestion
  kv["private_hgv"] = private[kv["hgv"]] or kv["private"] or "false"
```

src/mjolnir/ferry_connections.cc
```diff
@@ -177,6 +177,7 @@ std::pair<uint32_t, bool> ShortestPath(const uint32_t start_node_idx,
         }
 
         // if the first edge was destonly and the previous edge was as well, add no penalty
+        // TODO(nils): this would have to take into account HGV destonly to be accurate
```
nilsnolde, line 180: maybe log this situation out

src/sif/truckcost.cc
```diff
@@ -458,7 +464,8 @@ bool TruckCost::AllowedReverse(const baldr::DirectedEdge* edge,
   if (!IsAccessible(opp_edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((opp_edge->restrictions() & (1 << pred.opp_local_idx())) && !ignore_restrictions_) ||
       opp_edge->surface() == Surface::kImpassable || IsUserAvoidEdge(opp_edgeid) ||
-      (!allow_destination_only_ && !pred.destonly() && opp_edge->destonly()) ||
+      (!allow_destination_only_ && !pred.destonly() &&
+       (opp_edge->destonly() || opp_edge->destonly_hgv())) ||
```
nilsnolde, line 468:
```suggestion
      opp_edge->destonly_hgv()) ||
```

src/sif/truckcost.cc
```diff
@@ -535,6 +542,8 @@ Cost TruckCost::TransitionCost(const baldr::DirectedEdge* edge,
   uint32_t idx = pred.opp_local_idx();
   Cost c = base_transition_cost(node, edge, &pred, idx);
   c.secs = OSRMCarTurnDuration(edge, node, idx);
+  c.cost +=
+      destination_only_penalty_ * (edge->destonly_hgv() && !pred.destonly() && !edge->destonly());
```
kevinkreiser, line 546: this is good but wont you'll still get penalized from the base transition cost right if the regular dest only is set

kevinkreiser: you can over in base cost just not apply the regular one if you are hgv

valhalla/sif/edgelabel.h
```diff
@@ -72,19 +73,21 @@ class EdgeLabel {
             const bool closure_pruning,
             const bool has_measured_speed,
             const InternalTurn internal_turn,
-            const uint8_t path_id = 0)
+            const uint8_t path_id = 0,
+            const bool destonly = false)
```
kevinkreiser, line 77: we could instead pass the hgv flag here and then edgelabel would be the only place that has to have logic to decide whther to use the regular dest only flag or the hgv flag instead of this ternary/if leaking out into the algorithms

src/thor/bidirectional_astar.cc
nilsnolde: revert for now and leave todo

src/thor/dijkstras.cc
```diff
@@ -238,23 +238,27 @@ void Dijkstras::ExpandInner(baldr::GraphReader& graphreader,
...
       t2 = tile;
       oppedgeid = graphreader.GetOpposingEdgeId(edgeid, t2);
     }
 
     // Add edge label, add to the adjacency list and set edge status
     uint32_t idx = bdedgelabels_.size();
     *es = {EdgeSet::kTemporary, idx};
+    auto destonly =
+        FORWARD ? directededge->destonly() || (costing_->is_hgv() && directededge->destonly_hgv())
+                : opp_edge->destonly() || (costing_->is_hgv() && opp_edge->destonly_hgv());
```
nilsnolde, line 251: lets think about the directionality of edgelabel properties

nilsnolde:
One more thing that comes to mind: if a segment is tagged both `maxweight=10t` and `hgv=destination` I'd expect it to be reachable for origin/destination and truck, no matter its weight. Will try if it is already with this PR, but I don't think so..

src/thor/dijkstras.cc
nilsnolde: So I looked at the code a bit again: on reverse search we add the edge IDs of the expanded edges (i.e. the opposing ones of the final path), but we do add the _cost_ of their opposing edge which also sets the sortcost for the adjacency queue. While we don't have directionality of destonly _yet_ I still think this is the right way to go. See #4456 where IMO we're not adhering to this principle either.

valhalla/sif/edgelabel.h
nilsnolde: unless we require the caller to differentiate wrt direction. I think dijkstras is the only not doing that yet, it'd be trivial to refactor that. eventually we should bake that info into the EdgeLabel, but for now that'd work and still be nicer than having every algo care about it.

nilsnolde: meeh, forget about the previous comment. of course every direction puts the _expansion_ edge into the label, for reverse search we want the opposite. so unfortunately we have to leak it into the algos at this point.

src/thor/dijkstras.cc
nilsnolde: I refactored this a bit. Now it distinguishes between forward and reverse when adding labels during expansion.

src/mjolnir/ferry_connections.cc
nilsnolde: I added an implementation of this but not very happy with it. in fact, I'm pretty sure it's the same (bad) behavior as not acting here at all. it's not easy to log out this particular situation either, since `destonly_hgv` also contains normal `destonly`. I'd vote for leaving things as they are.

nilsnolde: added code to try and handle it: https://github.com/valhalla/valhalla/pull/4450/commits/5395f139feb03be5589cf050e5368cc5aab7722f

I reverted that commit already.

src/sif/truckcost.cc
nilsnolde: handled it in `base_transition_cost`

kevinkreiser (approved):
too bad about the leakage into the algos but understandable

PR #4455, opened 2023-12-13
Sort nodes within a tile by grid id within the tile

dnesbitt61:
# Issue

The hope is that spatial sorting of nodes within a tile can improve performance by improving memory coherence. Nodes that are close spatially should be more likely to be close to each other in tile memory with this change.

## Tasklist

 - [ ] Add tests

## Requirements / Relations


valhalla/baldr/tilehierarchy.h
```diff
@@ -53,6 +53,15 @@ class TileHierarchy {
...
   static GraphId GetGraphId(const midgard::PointLL& pointll, const uint8_t level);
 
+  /**
+   * Returns the grid Id within the tile. A tile is subdivided into a nxn grid.
+   * The grid Id within the tile is used to sort nodes spatially within the tile.
+   * @param pointll  Lat,lng location within the tile.
+   * @param level    Level of the requested tile.
+   * @return  Returns a grid id used for sorting nodes.
+   */
+  static uint32_t GetGridId(const midgard::PointLL& pointll, const uint8_t level);
```
kevinkreiser, line 63: can this stay an implementation detail inside of graphbuilder.cc rather than come into the public api (ie in the headers)

src/baldr/tilehierarchy.cc
```diff
@@ -60,6 +60,32 @@ GraphId TileHierarchy::GetGraphId(const midgard::PointLL& pointll, const uint8_t
   return id;
 }
 
+// Returns the grid Id within the tile. A tile is subdivided into a 32x32 grid.
+// The grid Id within the tile is used to sort nodes spatially.
+const int32_t kGridDivisions = 32;
```
kevinkreiser, line 65: should make this a mjolnir config so we can try different granularities and see what happens

valhalla/mjolnir/node_expander.h
```diff
@@ -217,6 +217,8 @@ struct Node {
   uint32_t end_of;
   // the graphid of the node
   baldr::GraphId graph_id;
+  // grid Id within the tile (used for spatial node sorting)
+  uint32_t grid_id;
```
kevinkreiser, line 221: if we wanted to keep this completely hidden as an implementation detail we could even use the spare bits in graph_id rather add another field here

kevinkreiser: it would certainly be less readable though..

kevinkreiser:
ive always wondered if this would help, thanks for doing it!

valhalla/baldr/tilehierarchy.h
dnesbitt61: done

dnesbitt61:
I ran a test conversion and long bicycle route in PA. Using either 32x32 or 64x64 grid within the tile leads to about an 8% performance improvement. I am seeing some test failures and some slight differences in edge label counts in the test route. I'm not sure what might be wrong as it seemed that sorting nodes within a tile should not impact behavior.

valhalla/mjolnir/node_expander.h
dnesbitt61: not sure we need to keep this hidden - especially if we add a mjolnir config for the grid sizing (that would make it pretty obvious!).

kevinkreiser: yeah sorry i just mean hidden in the sense that its not part of the c++ api (methods we expose in header files). feel free to ignore this :smile: was just an errant thought!

kevinkreiser:
8% thats great!

dnesbitt61:
I want to do a wider conversion / set of performance tests. Also need to figure out the test failures. But this is a start.

nilsnolde:
I like it! Surprisingly little changes. If the `de_benchmark_routes.txt` benchmarks well, that'd be a great proof. Feel free to use the planet from the fossgis instance.

src/baldr/tilehierarchy.cc
dnesbitt61: added as a mjolnir config

nilsnolde:
Will take a look later. Curious: what were those 4 extra byte (or whatever it was) in the end?

dnesbitt61:
If I change to use a non-zero default grid_divisions_within_tile in graphbuilder.cc, e.g.,

   `uint32_t grid_divisions = pt.get<unsigned int>("mjolnir.data_processing.grid_divisions_within_tile", 8);`

I get a few unit test failures. First I get errors in test/scripts/test_valhalla_build_extract.py. I commented out this test as this can be brittle if any changes to graph building occur.

In test/matrix I get a couple errors where duration is 473 (code expects 472). I am not sure this is significant - could be roundoff/order of adding edges to the adjacency list.

In test/predictive_traffic I get an error on truck speed: it gets 80 but expects 0. The comment says: 
`// Truck speed is not set for the edge initaially. In such case normal speed is used.`
I am not sure how a directed edge (this test uses Utrecht tiles) would have a truck speed of 0. At any rate the speed seems to be different when grid_divisions_within_tiles is not 0.

All other tests pass. Are these failures signifcant?

kevinkreiser:
to me the truck speed thing warrants looking into. i guess its saying that it should see a 0 in the data and that costing/graphtile::getspeed should fall back to another value

dnesbitt61:
Actually, the predictive_traffic test uses a fixed GraphID for the directed edge:
```
  // fixture traffic tile 0/003/196.gph
  GraphId id("0/3196/0");
  auto tile = reader.GetGraphTile(id);
  auto de = tile->directededge(id);
When sorting nodes using grids within the tile, the directed edges become sorted differently as well. So I think this test can be expected to fail since directed edge 0 within tile 3196 will be different.
```

kevinkreiser:
@dnesbitt61 excellent then id say there are no blockers

dnesbitt61:
looking more closely, this directed edge has wayid == 0. This directed edge is on level 0 and is **a shortcut edge**. Shortcut edges get truck_speed assigned using a duration along the shortcut using auto/car speeds (see TODO in shortcutbuilder on line 298). So this makes sense to me.

nilsnolde (dismissed):
great boost on average!

nilsnolde:
oops, still missing the test update..

dnesbitt61:
which test update? Do we want a test where we utilize node sorting? If so, I am open to suggestions for where/how to do this.

nilsnolde:
Oh I thought we said we'd just enable it in the default test config at https://github.com/valhalla/valhalla/blob/7eb478460a98bb76d91d7d3bd1a9b2a2e3fe4d63/test/test.cc#L119-L121. That'd change the vast majority of test to using that build option, though there's still a few other hard-coded configs around, e.g. for python & windows. Or we just do it for a bigger graph, like utrecht (that'd be [here](https://github.com/valhalla/valhalla/blob/7eb478460a98bb76d91d7d3bd1a9b2a2e3fe4d63/test/CMakeLists.txt#L173) I think)? Adapting the now failing tests is fine IMO. What do you think?

kevinkreiser (approved):

PR #4468, opened 2023-12-20
Nn scooter track

nilsnolde:
@gknisely found that motorscooter has ETAs of days for small distances on `track`s. The issue was that scooter speed could become 0, which would lead to a `speedfactor_` of 3600. This should fix it.


src/sif/motorscootercost.cc
```diff
@@ -426,9 +426,11 @@ Cost MotorScooterCost::EdgeCost(const baldr::DirectedEdge* edge,
     return {sec * ferry_factor_, sec};
   }
 
+  // prevent scooter speed to become 0
   uint32_t scooter_speed =
-      (std::min(top_speed_, speed) * kSurfaceSpeedFactors[static_cast<uint32_t>(edge->surface())] *
-       kGradeBasedSpeedFactor[static_cast<uint32_t>(edge->weighted_grade())]);
+      std::max(1.f, (std::min(top_speed_, speed) *
```
nilsnolde, line 431: at least 1. alternatively we could leave this std::max and update the `speedfactor_[0]` to the same value as `speedfactor_[1]` (3.6), which is currently set to 3600.

kevinkreiser (approved):
looks good, should refactor speedfactor at some point i8 guess or figure out why we have such a penalty for "closed"

nilsnolde:
oh boy.. that freaking gtfs test is acting up again! guess my band-aid from last year isn't working anymore..

nilsnolde:
that'll block all PRs again on all circleci builds.. I'll go home now hahah will take a look tmrw if no one else did

nilsnolde:
eieiei.. https://github.com/valhalla/valhalla/pull/4089#discussion_r1179438332 :laughing: @kevinkreiser you saw it and seems I never opened an issue to resolve that.. time to put the band-aid into saner code, or not merge until monday haha

PR #4469, opened 2023-12-21
require admins & tz

### LINKED ISSUE #4467: make admins & timezones required
we talked about it and came to the conclusion that there's no point to have timezones & admins optional. there's tons of issues with no admins added, and also having no timezones results in strange behavior if someone requests with time. @kevinkreiser suggested we could host those somewhere so people can pull them without building. we could use github artifacts for that, triggered by a manual github action to re-create them. those artifacts are available for download via the github api. though they're only persisted for 90 days or so I think, but it's easy to automatically generate them every couple of months, github actions has cron capabilities.

could even auto-download them in the code eventually via a config url defaulting to the latest github artifact, but the whole HTTP/curl compilation logic seems a bit daunting currently..

nilsnolde:
fixes #4467 

this is just a start, unfortunately these changes let all gurka tests fail currently with "no edges near location", very cryptic..

in any case, I wouldn't merge this until just before the next release, it's kinda major for anyone running valhalla.


nilsnolde:
Do I remember right that we came to the conclusion to host the DBs on Github artifacts instead of baking them into a release? I know there's the risk of broken admin data, but that's what it is then IMO. If someone has a strong need for better coverage, they can do the work to come up with a sane process to achieve that.

Also this move makes admin & tz db's necessary for all tests. While we could technically generate the tz db for windows too, it'd be easier to just download it in CI.

nilsnolde:
maybe a better route would be: we add a boolean arg to the build process to control whether we require the dbs, which `valhalla_build_tiles` can set to `true`. then it's hardly any changes. I don't really like that tests now require them as well.

wdyt?

PR #4483, opened 2023-12-29
fix `-j` for executables

nilsnolde:
We messed up on https://github.com/valhalla/valhalla/pull/4220, which removed the ability to change anything in the config for the `argparse_utils.h`. This fixes it.

### DIFF
--- a/src/argparse_utils.h
+++ b/src/argparse_utils.h
@@ -16,7 +16,7 @@
  * @param program The executable's name
  * @param opts    The command line options
  * @param result  The parsed result
- * @param pt      The config which will be populated here
+ * @param config  The config which will be populated here
  * @param log     The logging config node's key
  * @param use_threads Whether this program multi-threads
  *
@@ -26,6 +26,7 @@
 bool parse_common_args(const std::string& program,
                        const cxxopts::Options& opts,
                        const cxxopts::ParseResult& result,
+                       boost::property_tree::ptree& conf,
                        const std::string& log,
                        const bool use_threads = false) {
   if (result.count("help")) {
@@ -39,7 +40,6 @@ bool parse_common_args(const std::string& program,
   }
 
   // Read the config file
-  boost::property_tree::ptree conf;
   if (result.count("inline-config")) {
     conf = valhalla::config(result["inline-config"].as<std::string>());
   } else if (result.count("config") &&
--- a/src/mjolnir/valhalla_add_elevation.cc
+++ b/src/mjolnir/valhalla_add_elevation.cc
@@ -10,7 +10,6 @@
 #include "baldr/graphreader.h"
 #include "baldr/graphtile.h"
 #include "baldr/rapidjson_utils.h"
-#include "config.h"
 #include "mjolnir/elevationbuilder.h"
 
 #include "argparse_utils.h"
@@ -66,6 +65,7 @@ int main(int argc, char** argv) {
   const auto program = filesystem::path(__FILE__).stem().string();
   // args
   std::vector<std::string> tiles;
+  boost::property_tree::ptree config;
 
   try {
     // clang-format off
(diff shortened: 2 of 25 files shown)


src/argparse_utils.h
```diff
@@ -26,6 +26,7 @@
 bool parse_common_args(const std::string& program,
                        const cxxopts::Options& opts,
                        const cxxopts::ParseResult& result,
+                       boost::property_tree::ptree& conf,
```
nilsnolde, line 29: this is the main change. the exe's give an uninitialized one in here and we populate it. defining a local `conf` and changing it here makes no sense

nilsnolde: further down we add `mjolnir.concurrency` derived from the `-j` flag, the `mjolnir.concurrency` flag or whatever the hardware has

src/mjolnir/valhalla_add_elevation.cc
```diff
@@ -66,6 +65,7 @@ int main(int argc, char** argv) {
   const auto program = filesystem::path(__FILE__).stem().string();
   // args
   std::vector<std::string> tiles;
+  boost::property_tree::ptree config;
```
nilsnolde, line 68: all exe's have more or less the same changes. some still had the logging initializing stuff which I also removed.

nilsnolde: in fact this whole PR is mostly reverting #4220 and just restoring the way we handle things in `argparse_utils.h` based on the introduced singleton.

src/valhalla_export_edges.cc
```diff
@@ -158,17 +158,16 @@ int main(int argc, char* argv[]) {
     options.add_options()
       ("h,help", "Print this help message.")
       ("v,version", "Print the version of this software.")
-      ("c,column", "What separator to use between columns [default=\\0].", cxxopts::value<std::string>(column_separator)->default_value("\0"s))
+      ("c,config", "Path to the json configuration file.", cxxopts::value<std::string>())
+      ("i,inline-config", "Inline json config.", cxxopts::value<std::string>())
+      ("x,column", "What separator to use between columns [default=\\0].", cxxopts::value<std::string>(column_separator)->default_value("\0"s))
```
nilsnolde, line 163: not sure anyone exercised this exe in the past years, I felt it's fine to break the args to make it consistent with everything else

ImreSamu: I occasionally use it, and perhaps it would be helpful to add a note about this breaking change in the CHANGELOG.

nilsnolde: Oh good to know! Jep will do that

src/mjolnir/valhalla_add_landmarks.cc
```diff
@@ -40,18 +41,7 @@ int main(int argc, char** argv) {
...
-  auto logging_subtree = pt.get_child_optional("mjolnir.logging");
-  if (logging_subtree) {
-    auto logging_config =
-        valhalla::midgard::ToMap<const boost::property_tree::ptree&,
-                                 std::unordered_map<std::string, std::string>>(logging_subtree.get());
-    valhalla::midgard::logging::Configure(logging_config);
-  }
-
-  if (!valhalla::mjolnir::AddLandmarks(pt.get_child("mjolnir"))) {
+  if (!valhalla::mjolnir::AddLandmarks(config.get_child("mjolnir"))) {
```
kevinkreiser, line 44: works like graphreader?

src/mjolnir/valhalla_add_landmarks.cc
```diff
@@ -40,18 +41,7 @@ int main(int argc, char** argv) {
...
 
-  auto pt = valhalla::config();
-
-  // configure logging
-  auto logging_subtree = pt.get_child_optional("mjolnir.logging");
-  if (logging_subtree) {
-    auto logging_config =
-        valhalla::midgard::ToMap<const boost::property_tree::ptree&,
-                                 std::unordered_map<std::string, std::string>>(logging_subtree.get());
-    valhalla::midgard::logging::Configure(logging_config);
```
kevinkreiser, line 51: is this inside valhalla::config() now i expected it was in the parsing args place

kevinkreiser: ah i see its already in the parsing args spot just was redundant!

kevinkreiser (approved):

PR #4485, opened 2024-01-01
Update osmaccessrestriction.h

### LINKED ISSUE #3956: wrong value or comment
At [pbfgraphparser.cc](https://github.com/valhalla/valhalla/blob/master/src/mjolnir/pbfgraphparser.cc#L705) max value for AccessType - **31** (comment in [graphconstants.h](https://github.com/valhalla/valhalla/blob/master/valhalla/baldr/graphconstants.h#L580)), but max value for OSMAccessRestriction::set_type(AccessType type) in [osmaccessrestriction.h](https://github.com/valhalla/valhalla/blob/master/valhalla/mjolnir/osmaccessrestriction.h#L62) - only 4 bits,  max value = **15**

Beeram12:
changed from 4 bits to 6 bits

fixes #3956 

I am here changing size from 4 bits to 6 bits

### DIFF
--- a/valhalla/mjolnir/osmaccessrestriction.h
+++ b/valhalla/mjolnir/osmaccessrestriction.h
@@ -59,7 +59,7 @@ protected:
   uint64_t value_;
 
   struct Attributes {
-    uint16_t type_ : 4;
+    uint16_t type_ : 6;
     uint16_t modes_ : 12;
   };
   Attributes attributes_;
--- a/valhalla/mjolnir/osmaccessrestriction.h
+++ b/valhalla/mjolnir/osmaccessrestriction.h
@@ -59,7 +59,7 @@ protected:
   uint64_t value_;
 
   struct Attributes {
-    uint16_t type_ : 6;
+    uint16_t type_ : 6; // interpreted as AccessType but stored in AccessRestriction::type_
     uint16_t modes_ : 12;
   };
   Attributes attributes_;


Beeram12:
@kevinkreiser  since I am a beginner May I know how to resolve that 1 check to remove all errors

kevinkreiser:
@Beeram12 its not a required check so its not strictly needed. if you were to make a non-trivial change then we'd want you to add a description of the change to the changelog file. in this case no changelog is needed.

valhalla/mjolnir/osmaccessrestriction.h
```diff
@@ -59,7 +59,7 @@ class OSMAccessRestriction {
   uint64_t value_;
 
   struct Attributes {
-    uint16_t type_ : 4;
+    uint16_t type_ : 6;
```
kevinkreiser, line 62: lets put a comment here that this explains whats going on:

```suggestion
    uint16_t type_ : 6;  // interpreted as AccessType but stored in AccessRestriction::type_
```

valhalla/mjolnir/osmaccessrestriction.h
```diff
@@ -59,7 +59,7 @@ class OSMAccessRestriction {
   uint64_t value_;
 
   struct Attributes {
-    uint16_t type_ : 4;
+    uint16_t type_ : 6; // interpreted as AccessType but stored in AccessRestriction::type_
     uint16_t modes_ : 12;
```
kevinkreiser, line 63: ok so i dont think we should merge this and this is why. please note that we are storing this data as 16 bit value. if you increase `type_` to cover 6 bits now we have 18 bits. the compiler will pad this struct to 32 bits but its not what the code says. a gotcha for people to stumble over in the future.

we have two options:

1. revert `type_` back to 4 bits because really we dont ever use the other 2 bits yet and then add a compile time check that max AccessType is < 2^4
2. upgrade attributes to 32bits: `struct Attributes { uint32_t type_ : 6; uint32_t modes_ : 12; uint32_t spare_ : 14; };`

i favor the first option because doubling the size of this struct is wasteful (though its relatively sparsely populated so not so bad)

Beeram12: Thanks for clarifying me

kevinkreiser: @dnesbitt61 also mentioned that we could just go into `AccessRestriction::type_` set the type to 4 bits and then have 2 bits spare in that field

PR #4487, opened 2024-01-03
Do not reclassify ferry connections when hierarchy=false

### LINKED ISSUE #4486: When hierarchies are not built, we should not reclassify links or ferry connections
Currently we always reclassify edges leading to ferry connections. However, if we are not creating hierarchies there is no reason to reclassify edges. The same holds for reclassifying links (ramps, turn channels). In fact, the reclassification can impact costing for bicycle routes (perhaps others) that use road classification in edge costing methods.

dnesbitt61:
Do not reclassify ferry connections if hierarchies are not being used. Add a log statement to indicate no reclassification is occurring in the case where mjolnir config has hierarchy set to false.

Update graphbuilder to call link reclassification if infer_turn_channels is true. However, if reclassify_links is false then the actual reclassification (changing an edges classification) is bypassed but turn channel generation is performed (if config indicates infer_turn_channels=true).

fixes #4486 

Log statements from a Maryland build where infer_turn_channels=true, reclassify_links=false, and hierarchy=false:

```
2024/01/10 13:16:49.685371 [INFO] Finished with 0 link edges reclassified.  Turn channel count = 4935
2024/01/10 13:16:49.709464 [WARN] Not reclassifying ferry connections since no hierarches are being created
```

# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


nilsnolde (approved):
LGTM

dnesbitt61:
Updated so that ReclassifyLinks is called if either reclassify_links and infer_turn_channels is true in the config. Added a reclassify_links bool to ReclassifyLinks (and pass this through to ReclassifyLinkGraph). If this is false, no edges are reclassified (but can be used for infer_turn_channels).

kevinkreiser (dismissed):
should have some tests for this but it is very straight forward

dnesbitt61:
I copied the TEST(RampsTCs, test_tc_infer) test and changed so that reclassify_links is false and infer_turn_channels is true. The test checks the ramps classification (makes sure they stay as motorway). Edge GK seems to still be marked as internal and a turn channel (as before). However, the narrative checks fail now. I am not sure why!

test/gurka/test_ramps_tc.cc
```diff
@@ -290,8 +290,115 @@ void check_edge_classification(baldr::GraphReader& graph_reader,
...
+      {"KJ", {{"highway", "primary"}, {"oneway", "yes"}}},
+      {"KN", {{"highway", "motorway_link"}, {"oneway", "yes"}}},
+      {"LN", {{"highway", "motorway_link"}, {"oneway", "yes"}}},
+      {"NC", {{"highway", "motorway_link"}, {"oneway", "yes"}}},
+  };
+
+  const gurka::nodes nodes = {{"B", {{"highway", "motorway_junction"}, {"ref", "4"}}}};
+
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize_metres);
+  auto map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_ramps_tc_infer",
```
kevinkreiser, line 355: when copying from other tests it might be a good idea to use a different dir. in this case its not a problem because the tests are run synchronously but if you had copied from another test in another file, then we can get random failures when the tests are run in parallel (since they race over messing with the same directory)

kevinkreiser (approved):

test/gurka/test_ramps_tc.cc
dnesbitt61: fixed

PR #4489, opened 2024-01-03
Nn matrix geometry try2

cvvergara:
Removing warnings from merged branch nn-matrix-geometry
(see #4435 )


kevinkreiser:
can you keep the headers private? id rather not expand the libraries public api

cvvergara:
I get you, I will start all over and leave the header problems to another PR

cvvergara:
hmm, I will not move them to valhalla/tyr first to see if it works

cvvergara:
@kevinkreiser  tyr headers are now in src/tyr directory

src/tyr/route_serializer.cc
```diff
@@ -5,7 +5,7 @@
 
 #include "midgard/encoded.h"
 #include "midgard/util.h"
-#include "route_serializer_osrm.cc"
+#include "route_serializer_osrm.h"
```
nilsnolde, line 8: yeah, so much nicer to look at:)

src/tyr/route_summary_cache.h
```diff
@@ -1,3 +1,5 @@
+#ifndef VALHALLA_TYR_ROUTE_SUMMARY_CACHE_H_
+#define VALHALLA_TYR_ROUTE_SUMMARY_CACHE_H_
 #pragma once
```
nilsnolde, line 3: shouldn't `pragma once` be enough here? I thought it's properly supported by most toolchains by now?

valhalla/tyr/serializer_constants.h
```diff
@@ -1,3 +1,6 @@
+#ifndef VALHALLA_TYR_ROUTE_SERIALIZER_CONSTANTS_
+#define VALHALLA_TYR_ROUTE_SERIALIZER_CONSTANTS_
+#pragma once
```
nilsnolde, line 3: same here

nilsnolde (commented):
I guess you hardly changed anything in `route_serializer` & `route_summary_cache` other than splitting them?

src/tyr/route_summary_cache.h
cvvergara: `pragma once` is not standard

nilsnolde (dismissed):
LGTM

valhalla/tyr/serializer_constants.h
nilsnolde: let's just remove those `pragma once` then

nilsnolde (approved):

PR #4490, opened 2024-01-04
fiddling with tar_index test failures

kevinkreiser:
we have a long standing issue with the tar_index test which fails randomly on osx and arm. i'm going to play around with trying to fix it.. more to come later


nilsnolde:
Got on intel OSX and ran this test 100 times in a loop, didn't fail once.. Is anything manipulating the utrecht tar in some other test, so parallel execution might be a problem? I did run `check` a few times as well, but didn't get any failure either..

nilsnolde:
Coincidentally we havent' seen such a failure in > 1 week and quite a few builds.. :man_shrugging:

kevinkreiser:
we should merge this, it is a bug in our test configuration. i cant explain how it would cause the issue we have, it doesnt seem enough but i havent had time to try to reproduce it and we may as well get this small fix

kevinkreiser:
> so parallel execution might be a problem
yes the problem has to be something due to timing

nilsnolde:
go for it. it's weird that there's times when it's not failing for weeks, and then it's failing randomly here and there for 1-2 weeks, and again pause for weeks. :shrug:

PR #4491, opened 2024-01-04
Add time zone info to `/route` and `sources_to_targets` responses

### LINKED ISSUE #4241: timezone info in response `date_time`

chrstnbwnkl:
# Issue

Fixes #4241 

Adds `time_zone_offset` and `time_zone_name` fields to `/route` and `/sources_to_targets` responses while preserving the output in `date_time` output fields. 

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/thor/route_action.cc
```diff
@@ -381,26 +381,28 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
-      // back propagate time information
-      if (!destination->date_time().empty() &&
-          options.date_time_type() != valhalla::Options::invariant) {
-        auto origin_dt =
-            DateTime::offset_date(destination->date_time(),
-                                  reader->GetTimezoneFromEdge(temp_path.back().edgeid, tile),
-                                  reader->GetTimezoneFromEdge(temp_path.front().edgeid, tile),
-                                  -temp_path.back().elapsed_cost.secs);
-        origin->set_date_time(origin_dt);
+      auto out_tz = reader->GetTimezoneFromEdge(temp_path.back().edgeid, tile);
```
chrstnbwnkl, line 384: Previously, the `date_time` was only set on the first or last location (depending on whether or not `/date_time/type` was `ArriveBy`) and then propagated. Now, we also need to call `offset_date` on these to get the time zone information, since we cannot get time zone information in `worker.cc` where `date_time` is initially set on the first/last location.

test/gurka/test_matrix.cc
```diff
@@ -538,3 +538,123 @@ TEST(StandAlone, CostMatrixShapes) {
...
+gurka::map DateTimeTest::map = {};
+gurka::map DateTimeTest::map_tz = {};
+
+TEST_F(DateTimeTest, DepartAt) {
+  rapidjson::Document res_doc;
+  std::string res;
+  {
+    auto api = gurka::do_action(valhalla::Options::sources_to_targets, map_tz, {"A", "G"}, {"A", "G"},
+                                "auto",
+                                {{"/prioritize_bidirectional", "true"},
```
chrstnbwnkl, line 591: We're only testing `costmatrix` now because of a bug discovered in `timedistancematrix` that causes the timezone info of the last found source to be set on all source locations.

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@
    * ADDED: find connection on backward search for bidir matrix algo [#4329](https://github.com/valhalla/valhalla/pull/4329)
    * FIXED: Fix segfault in OSRM serializer with bannerInstructions when destination is on roundabout [#4480](https://github.com/valhalla/valhalla/pull/4481)
    * CHANGED: Adujustment of walk speed when walking on slight downhill [#4302](https://github.com/valhalla/valhalla/pull/4302)
+   * ADDED: find connection on backward search for bidir matrix algo [#4491](https://github.com/valhalla/valhalla/pull/4491)
```
nilsnolde, line 71: not quite the right description ;)

chrstnbwnkl: Oops lol

test/gurka/test_matrix.cc
nilsnolde: I like the tests. It's a lot but it's very easy to follow and we won't get failures in some opaque functions

test/gurka/test_matrix.cc
```diff
@@ -538,3 +538,123 @@ TEST(StandAlone, CostMatrixShapes) {
...
+      |          |
+      C          D
+      |          |
+      ~          ~
+      |          |
+      |          |
+      E          F
+      |          |
+      G----------H
+    )";
```
nilsnolde, line 563: this is really good too! not sure we had any end-to-end test with timezones derived from a gurka map all the way to checking the date_time in the response

nilsnolde (dismissed):
since we had a few rounds before, this looks good to me:) let's show the others next week.

src/thor/route_action.cc
```diff
@@ -381,26 +381,28 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
+      auto in_tz = reader->GetTimezoneFromEdge(temp_path.front().edgeid, tile);
+
+      // we add the timezone info if destination is the last location
+      // and add waiting_secs again from the final destination's datetime, so we output the departing
+      // time at intermediate locations, not the arrival time
+      if ((destination->correlation().original_index() == (options.locations().size() - 1) &&
```
kevinkreiser, line 390: since its only for the last location, any chance we could pull this special case outside the loop? no worries if its not ergonomic

chrstnbwnkl: I think the issue with that is that we use the `PathInfo`'s `edgeid` to get the time zone information here, is there a way to get that time zone information outside of the loop?

nilsnolde: you could get to it via the TripLeg in PBF, but honestly, that's more hassle than it's worth IMHO.

nilsnolde: wdyt @kevinkreiser ?

src/thor/route_action.cc
```diff
@@ -381,26 +381,28 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
     if (temp_paths.empty())
       return false;
     for (auto& temp_path : temp_paths) {
-      // back propagate time information
-      if (!destination->date_time().empty() &&
```
kevinkreiser, line 385: because this is arrive_by we are certain there will always be a date_time on the destination?

src/thor/route_action.cc
kevinkreiser: yep no worries fine here

src/thor/route_action.cc
kevinkreiser: ah i see you have it for the regular locations just not the first one

nilsnolde (approved):
never mind the OSX failure, there's smth wrong with the mapmatch test here and there (or OSX/clang, which I don't trust anymore after the recent stunt..)

nilsnolde (approved):
we already reviewed this last week together in the PR meeting and chris addressed all comments.

should we merge?

kevinkreiser (approved):

Issue #4475, opened 2023-12-23
Golf Cart / NEV / LSV Routing


ianthetechie:
We at Stadia Maps have developed a golf cart routing profile in a fork which is currently private but we're interested to upstream.

The first question to answer is probably whether or not this is something that belongs in Valhalla generally. Our current implementation is somewhat specific to golf carts, but I could also see it evolving to a more general NEV/LSV profile with specific vehicle subtypes (like we have for bicycles already). I think there's a strong case to be made for this as being important to the future of mobility and I'd be happy to put in some work to make our current work more generic if needed before merging.

Let me know if there's interest!

nilsnolde:
Definitely interesting for us! I had a client semi-recently with a golf cart app for a local community somewhere in Florida, with no resources to invest in smth like this. I def agree, it’s important for the future (even though reality is that many many idiots keep buying _bigger_ cars, offense intended). 

Feel free to PR, we can discuss more there. Thanks!

ianthetechie:
I have a feeling we may have interacted with the same client 😂

>  I def agree, it’s important for the future (even though reality is that many many idiots keep buying bigger cars, offense intended).

💯 this 😂 

Perfect; I'll work up a PR early in the new year. I'm working on integrating the latest from `master` right now and verifying that nothing broke + cleaning some stuff up.

nilsnolde:
Hahah was it the guy who was burned badly by MB, being invoiced 7-8 times the usual due to their newish plan structures? I think I sent him your way. He was very sweet I remember.

Great, looking forward to the PR:)

kevinkreiser:
wtf! some 'idiots' need big cars because they have big families or live in places that have rugged terrain or have businesses that require them to pull heavy loads. there are many legitimate reasons to have 'big cars' even if they may be difficult to imagine from your perspective

nilsnolde:
Believe it or not, I have enough perspective for that. Obviously there’s tons of legitimate reasons for big cars as well as legitimate reasons to have a car at all. The context is _golf carts_, so clearly we’re talking about urban environment and I stand 100% by why I said. If you wanna take it personal, I can’t stop you from that, even I can’t see any connection whatsoever..

kevinkreiser:
[...]

PR #4492, opened 2024-01-04
Golf cart costing

ianthetechie:
# Issue

Adds golf cart costing (see #4475; let's discuss if this closes it or not).

Quick hits:
* This has seen production use for 6+ months (not this *exact* code, as I just merged master this week, but substantially the same)
* This *exact* code (incl. the speculative Meili fix) have been tested to generate on a relevant extract covering GA and FL + produce good routes in the golf communities there
* Currently highly specific to golf carts; I could see us expanding to other forms of LSVs / NEVs in the future, as I mentioned in the original issue, but I'm not sure I have enough of an opinion on what that should be yet (the golf cart rules were developed with both municipal reps and people intimately familiar with the rules + using them for daily transportation)
* Initial PR still has some rough edges before merging (see discussions + checklist). Also my IDE has some opinions and I don't know how widely those are shared 😂 (ex: markdown table formatting)
* Includes a few unrelated cleanup bits that either were flagged by clang-tidy or caused issues on my ARM mac. (Note: I do get one failure when running `make -j8 check` locally, but it appears to be somewhat spurious related to floating point comparison; we'll see what the official CI run says.)

## Tasklist

 - [X] Add tests
 - [X] Update the docs with any new request parameters or changes to behavior described
 - [ ] Update the [taginfo](taginfo.json)

## Requirements / Relations

N/A?


valhalla/mjolnir/osmaccess.h
```diff
@@ -224,8 +240,11 @@ struct OSMAccess {
       uint16_t motorroad_tag : 1;
       uint16_t moped_tag : 1;
       uint16_t motorcycle_tag : 1;
-      uint16_t spare : 6;
+      uint16_t golf_cart_tag : 1;
+      uint16_t spare_ : 5;
     } fields;
+    // TODO: What exactly is 64 bits wide? I'm counting 16 + 32 = 48 bits?
+    uint16_t _spare;
     uint32_t v; // this should be 64bits wide for architectures who require word alignment
```
ianthetechie, line 248: I'm not quite sure what's going on here. I see a comment about something being 64 bits wide for alignment... but in this block of fields I'm only counting up 48. I'm occasionally both dense and bad at math though so maybe I missed something.

valhalla/baldr/graphconstants.h
```diff
@@ -42,11 +42,12 @@ constexpr uint16_t kHOVAccess = 128;
 constexpr uint16_t kWheelchairAccess = 256;
 constexpr uint16_t kMopedAccess = 512;
 constexpr uint16_t kMotorcycleAccess = 1024;
+constexpr uint16_t kGolfCartAccess = 2048;
```
ianthetechie, line 45: Worth noting: this PR maxes out all available bit field constants, per the comment on L33 which states we have 12 bits available.

lua/graph.lua
```diff
@@ -996,6 +1038,10 @@ function filter_tags_generic(kv)
     if kv["motorroad"] == "yes" then
       kv["motorroad_tag"] = "true"
     end
+
+    if kv["golf_cart"] == "designated" then
+      kv["golf_cart_designated"] = "true"
```
ianthetechie, line 1043: Worth noting: this felt a bit ad hoc when I wrote it, but I'm also not sure we want to introduce a generalized differentiation between designated and permissive across all costing modes.

In the case of golf carts, it actually is significant, since there are typically local ordinances to the effect of "use paths designated for golf cart use whenever possible, even though another road may be legally traversable for short distances"

src/thor/triplegbuilder.cc
```diff
@@ -1151,11 +1151,15 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
     trip_edge->set_speed(speed);
   }
 
-  uint8_t kAccess = 0;
+  uint16_t kAccess = 0;
```
ianthetechie, line 1154: Widened because this hasn't accommodated the access mask in some time but we just only used the lower values previously.

src/odin/enhancedtrippath.cc
```diff
@@ -123,6 +123,7 @@ const std::string& TripLeg_TravelMode_Name(int v) {
 const std::string& TripLeg_VehicleType_Name(int v) {
   static const std::unordered_map<int, std::string> values{
       {0, "kCar"}, {1, "kMotorcycle"}, {2, "kAutoBus"}, {3, "kTractorTrailer"}, {4, "kMotorScooter"},
+      {5, "kFourWheelDrive"}, {6, "kGolfCart"},
```
ianthetechie, line 126: I think the 4WD value was always missing, but it was never noticed ;)

src/sif/golfcartcost.cc
```diff
@@ -0,0 +1,565 @@
...
+#include "baldr/graphconstants.h"
+#include "baldr/nodeinfo.h"
+#include "midgard/constants.h"
+#include "midgard/util.h"
+#include "proto_conversions.h"
+#include "sif/costconstants.h"
+#include "sif/osrm_car_duration.h"
+#include <cassert>
+
+// TODO: Inline tests like other profiles have
```
ianthetechie, line 12: This should have a question mark after it. I couldn't tell whether those inline tests were used and/or valuable, so I didn't write them when I first implemented the profile. Should I write some up before this gets merged (NOTE: there *are* Gurka tests)?

valhalla/mjolnir/osmway.h
```diff
@@ -2679,9 +2727,15 @@ struct OSMWay {
...
+  // Flags which didn't fit elsewhere
   uint16_t lit_ : 1;
   uint16_t destination_only_hgv_ : 1;
-  uint16_t spare2_ : 2;
+  // 2 access flags that didn't fit in the previous 16 bits
+  uint16_t golf_cart_forward_ : 1;
+  uint16_t golf_cart_backward_ : 1;
+
+  // Sadly this doesn't fit packed into the previous 16 bits so it's on its own for now
+  bool golf_cart_designated_;
```
ianthetechie, line 2738: Worth noting: I added 3 bits in total, which means overflowing the existing nicely packed set of fields :(

ianthetechie (commented):
Self-flagging all the bits that stood out to me as needing a second look. I also know I need to update the CHANGELOG and taginfo.

src/odin/enhancedtrippath.cc
nilsnolde: in fact I removed it recently:) maybe you're not on latest master?

nilsnolde:
great, thanks @ianthetechie ! might be worth to join a PR review meeting, which are happening mondays & wednesdays at 2:30 pm berlin time. this is quite big and there's a few discussable points, possibly better reviewed synchronously. would you be up for that?

src/sif/golfcartcost.cc
kevinkreiser: the inline tests are the only way you can "unit test" the specific parsing functions because they are hidden in this implementation and not exposed in the headers. if you've tested all the stuff that this costing parses as costing options and feel that its covered in gurka there is no need to add them. if you look at other examples you'll see (iirc) that they check that the range objects that get setup are doing the right thing which is not too easy to test outside of this implementation file.

nilsnolde:
Also need to run good ol' format.sh :)

ianthetechie:
@nilsnolde yeah that sounds like the best use of everyone's time ;) Can you send me an invite for Wednesday?

src/odin/enhancedtrippath.cc
ianthetechie: That could have also gotten lost in merge conflict resolution 🎃 If it's gone, then I guess I can set this constant to 5 as well?

valhalla/mjolnir/osmway.h
kevinkreiser: bool wont become a bit. the compiler will pad that to either 1, 4 or 8 bytes depending on your platform. to avoid confusion we should probably just add another `uint16_t` or another type with a specific width (whatever makes sense depending on what we already have in here to get to a round number (multiple of 64 bits)) and then take 1 bit from it and make the rest spare.

ianthetechie: That's a good point for me to follow up on.

The people actually using this stated that they wanted it to be able to ensure that golf carts used appropriate paths even when another road was legally available, but doing a quick check of the areas with overpass turbo, I'm not actually sure they are :P If that's the case, we could possibly free up that bit + simplify the logic a bit.

src/sif/golfcartcost.cc
ianthetechie: Thanks, Kevin. I'll see about adding some quick tests. That said, I'm not too worried about it, and the Gurka tests actually do test the new parameter for max allowable road speed, and vehicle top speed isn't unique to this costing model.

valhalla/mjolnir/osmaccess.h
kevinkreiser: its a bit complicated but let me explain. first, `AccessAttributes` is a `union` of a 16bit field and a 32bit field, that means the compiler will make the whole thing 32bits wide. so we shouldnt add the spare field, that wont do anything. the comment is really just saying that on 64bit platforms, `AccessAttributes` will get padded to 64bits because the other field that is in the outer `OSMAccess` struct is already 64bits (it used to be 32 and this one is 32 so they would both stop at 64).

ianthetechie: oooooooh! Thanks for the explanation; that makes sense.

src/sif/golfcartcost.cc
kevinkreiser: yeah tbh i wouldnt ask you to add the inline tests!

valhalla/mjolnir/osmway.h
kevinkreiser: i was hoping you might say that :smile:

kevinkreiser:
@ianthetechie i sent you the invite, see you then!

ianthetechie:
Is it possible to run `format.sh` on macOS on Apple Silicon at the moment? A quick attempt seems to indicate that it's highly specific to x86 Linux, but I might be missing something.

kevinkreiser:
@ianthetechie its not possible. it relies on mapbox/mason to get a specific version of clang formt. mason is deprecated and there are no builds for your platform. honestly this is a whole problem for any linter. if you dont pin the linter version you get different peoples platforms disagreeing with each other and every other commit reformats the whole code base. this is one of many reasons ive always hated opinionated linting and opposed its use here :smile: but whatever we agreed to do it so we are stuck-ish now. the only real solution is docker. basically run a docker image of ubuntu with the right version of clang format to lint the code. i had written somewhere that this was our only option, im just not jazzed to work on it really..

kevinkreiser:
oh and i should say the arm release build skips lint because of this fact so we can use that build to signal pass or fail on this pr until we are done making logical tweaks to it. then we can lint as a last step

src/odin/enhancedtrippath.cc
ianthetechie: Should be gone now.

valhalla/mjolnir/osmaccess.h
ianthetechie: Reverted in the latest commit.

lua/graph.lua
ianthetechie: This has been removed.

valhalla/mjolnir/osmway.h
ianthetechie: Followed up with the relevant parties and they are on board. I've removed all the designated logic.

nilsnolde:
@ianthetechie can you make your fork editable by maintainers? then it'll be easy to eventually run the format.sh script:)

ianthetechie:
@nilsnolde hmm..... It looks like GitHub [doesn't allow this for org-owned forks](https://github.com/orgs/community/discussions/5634) -_-

I think the options are 1) proceed till the end with this PR, 2) make another fork personally and PR from there, or 3) add you + Kevin as collaborators (and remind myself to not contribute from org forks until GitHub fixes this insanity) ;)

valhalla/baldr/graphtile.h
```diff
@@ -882,16 +882,16 @@ class GraphTile {
   std::size_t lane_connectivity_size_{};
 
   // Predicted speeds
-  PredictedSpeeds predictedspeeds_;
+  PredictedSpeeds predictedspeeds_{};
 
   // Map of stop one stops in this tile.
-  std::unordered_map<std::string, GraphId> stop_one_stops;
+  std::unordered_map<std::string, GraphId> stop_one_stops{};
```
nilsnolde, line 888: did this give a warning or? `unordered_map` shouldn't have problems with not being initialized

ianthetechie: Yeah, clang-tidy or something was giving me linter warnings 🤷

kevinkreiser:
ian can do the linting himself from inside docker or we can update CI so that it spits out the diff when lint fails


i have another general question, is it possible to not make this so specific to golf carts? could we make it for all low speed vehicles and include golf cart tags in the mix?

lua/graph.lua
```diff
@@ -1999,18 +2092,24 @@ function nodes_proc (kv, nokeys)
...
   --do not shut off bike access if there is a highway crossing.
   if bike_tag == 0 and kv["highway"] == "crossing" then
     bike_tag = 4
   end
 
+  --do not shut off golf cart access if there is a highway crossing.
+  if golf_cart_tag == 0 and kv["highway"] == "crossing" then
+    golf_cart_tag = 2048
+  end
+
```
ianthetechie, line 2112: This is something to look into.

Make a gurka test to check if this is a problem for ones not marked as crossing. It might be deleteable.

src/sif/golfcartcost.cc
```diff
@@ -0,0 +1,547 @@
...
+   * mode used by the costing method. It's also used to filter
+   * edges not usable / inaccessible by golf cart.
+   */
+  bool Allowed(const baldr::DirectedEdge* edge,
+               const graph_tile_ptr& tile,
+               uint16_t disallow_mask = kDisallowNone) const override {
+    bool allow_closures = (!filter_closures_ && !(disallow_mask & kDisallowClosure)) ||
+                          !(flow_mask_ & kCurrentFlowMask);
+    return DynamicCost::Allowed(edge, tile, disallow_mask) && !edge->bss_connection() &&
+           (edge->speed_type() == SpeedType::kClassified || tile->edgeinfo(edge).speed_limit() <= max_allowed_speed_limit_) &&
```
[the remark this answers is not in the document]
ianthetechie: Follow-up:

1. Check if tagged (maybe makes it easier to read)
2. Use `edge->speed()` instead of `edgeinfo`

src/sif/golfcartcost.cc
```diff
@@ -0,0 +1,547 @@
...
+           (allow_closures || !tile->IsClosed(edge));
+  }
+
+  // Hidden in source file so we don't need it to be protected
+  // We expose it within the source file for testing purposes
+
+  std::vector<float> speedfactor_;
```
ianthetechie, line 267: This should probably be a fixed sized array (open an issue as affects almost all profiles, not just this one).

ianthetechie (commented):
Capturing some follow-ups from the first review session

src/sif/golfcartcost.cc
```diff
@@ -0,0 +1,547 @@
...
+  // Check access, U-turn, and simple turn restriction.
+  // Allow U-turns at dead-end nodes.
+  if (!IsAccessible(edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
+      ((pred.restrictions() & (1 << edge->localedgeidx())) && !ignore_restrictions_) ||
+      (edge->surface() > kMinimumGolfCartSurface) || IsUserAvoidEdge(edgeid) ||
+      // NOTE: Parking aisles are baked as destination-only, but for golf carts we actually need
+      // to ignore this (per the GIS department of Peachtree City, GA).
+      (!allow_destination_only_ && !pred.destonly() && edge->destonly() && edge->use() != Use::kParkingAisle) ||
+      (pred.closure_pruning() && IsClosed(edge, tile)) ||
+      (edge->speed_type() == SpeedType::kTagged && tile->edgeinfo(edge).speed_limit() > max_allowed_speed_limit_)) {
```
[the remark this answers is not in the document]
ianthetechie: Also fixed.

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -97,43 +98,42 @@ Another special case is `disable_hierarchy_pruning` costing option. As the name
...
-| `service_factor` | A factor that modifies (multiplies) the cost when generic service roads are encountered. The default `service_factor` is 1. |
-| `country_crossing_cost` | A cost applied when encountering an international border. This cost is added to the estimated and elapsed times. The default cost is 600 seconds. |
-| `country_crossing_penalty` | A penalty applied for a country crossing. This penalty can be used to create paths that avoid spanning country boundaries. The default penalty is 0. |
-| `shortest` | Changes the metric to quasi-shortest, i.e. purely distance-based costing. Note, this will disable all other costings & penalties. Also note, `shortest` will not disable hierarchy pruning, leading to potentially sub-optimal routes for some costing models. The default is `false`. |
-| `disable_hierarchy_pruning` | Disable hierarchies to calculate the actual optimal route. The default is `false`. **Note:** This could be quite a performance drainer so there is a upper limit of distance. If the upper limit is exceeded, this option will always be `false`. |
```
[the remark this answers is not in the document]
ianthetechie: Fixed.

lua/graph.lua
ianthetechie: OK, I did a bit of research and testing here...

1. Gurka test (crossing a road where golf carts are not allowed; will be committed shortly) seems to pass just fine if I remove this logic.
2. I think I said in our last meeting that I assumed this was a direct translation of a business requirement. Looking over it again though, I realized I just copied it directly above from the same logic for bicycles ;) I will _not_ touch the bicycle behavior in this PR, but maybe this is room for simplification in the future I guess.
3. Confirmed this only applies at the node level in OSM (just a few hundred invalid way tags) + checked the area quickly for odd looking tags with Overpass Turbo. Shouldn't cause any issues.
4. I did some smoke testing with random routes in the areas we're serving in GA and FL and it doesn't look like it broke anything.

So, after all that, I went ahead and deleted this.

test/gurka/test_golf_cart_preference.cc
```diff
@@ -0,0 +1,80 @@
...
+// Check that golf carts prefer lower classes of roads
+class GolfCartPreference : public ::testing::Test {
+protected:
+  static gurka::map map;
+
+  static void SetUpTestSuite() {
+    constexpr double gridsize = 100;
+
+    // A network with a mixture of road classes, all of which are legally traversable
+    const std::string ascii_map = R"(
```
ianthetechie, line 19: Combine this into the access tests to reduce the number of executables that need to be built.

valhalla/mjolnir/osmway.h
```diff
@@ -2679,9 +2711,12 @@ struct OSMWay {
   uint16_t use_sidepath_ : 1;
   uint16_t bike_forward_ : 1;
   uint16_t bike_backward_ : 1;
+  // Flags which didn't fit elsewhere
   uint16_t lit_ : 1;
   uint16_t destination_only_hgv_ : 1;
-  uint16_t spare2_ : 2;
+  // 2 access flags that didn't fit in the previous 16 bits
+  uint16_t golf_cart_forward_ : 1;
+  uint16_t golf_cart_backward_ : 1;
```
ianthetechie, line 2719: Move this up to the top of the block; doesn't break anything.

src/sif/golfcartcost.cc
```diff
@@ -0,0 +1,547 @@
...
+  bool is_parking_aisle = edge->use() == Use::kParkingAisle;
+
+  // Add penalties based on road class.
+  // The dest only penalty (for private roads) is normally just a one-time fixed penalty on transitions.
+  // However, in the case of golf cart routing, this is usually not sufficient to avoid large areas such as private golf courses which are not routable except as a destination.
+  avoidance_factor += kRoadClassPenaltyFactor[static_cast<uint32_t>(edge->classification())] + (edge->destonly() * !is_parking_aisle * 10);
+
+  // Multiply by speed so that roads are more severely punished for having traffic faster than the golf cart's speed.
+  // Use the speed assigned to the directed edge. Even if we had traffic information we shouldn't use it here.
+  avoidance_factor *= speedpenalty_[edge->speed()];
```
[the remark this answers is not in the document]
ianthetechie: Ian to double check on this

ianthetechie:
Quick update on the status of this... I'm working to ensure the first version doesn't require _too_ much of an overhaul later ;) To that end, I've been having conversations with others involved in the LSV space over the last month, and learned quite a few things about which assumptions are golf-cart specific and/or should be tunables vs being generally applicable, including things like "primary roads are actually routable in downtown LA because speed limits are so low."

I'm continuing work on this in the background and will likely have a more polished generic LSV profile ready in a few weeks (I have completed most of the refactor to generalize terminology locally).

kevinkreiser:
@ianthetechie no worries i very much appreciate the diligence in using this last bit in the access mask and making it as worth it as possible :smile:

ianthetechie:
Yeah, here's a quick info dump on that just so it's recorded in one piece (there will also be comments in the code to note areas for further improvement, specialization, etc.).

First, there is no tagging scheme, either de facto or in use, which covers LSVs as a more general class, nor do I think there is likely to be one anytime soon. LSVs/quadricycles (EU equivalent) mostly behave like cars with somewhat different preferences and a federal/national or state law restricting roads they can legally use based on the posted speed limit.

Thus, **the access bit will have to go to golf carts**, which _do_ have a widely used tagging scheme and are often subject to specific restrictions that are only modeled this way.

Fortunately, there are no scenarios I'm currently aware of preventing LSVs/NEVs which are not easily captured in some other attribute like max allowed (tagged) speed. So we'll get a general LSV profile that's tunable for both vehicle type and max allowed speed, and we can evolve it as we uncover more use cases. Most of my work recently has been around making sure that the first version allows for such expansion.

nilsnolde:
>  So we'll get a general LSV profile that's tunable for both vehicle type and max allowed speed

this is sadly not fully true.. "max" allowed speed is not something we capture today for edges. we have a "tagged" speed, but that's not necessarily the OSM `maxspeed` value: if the same edge has a `maxspeed:practical` (or `default` or whatever that was), then we prefer that over the plain `maxspeed` when we set an edge's `speed` value. the reason is that maxspeed is rarely useful for costing. problem is we can't distinguish those, it'll always just be "tagged". it's not trivial to support that, but certainly possibly with some bit juggling.

we do have `speed_limit` on EdgeInfos, but again, we don't want to use that for costing.

ianthetechie:
> we have a "tagged" speed, but that's not necessarily the OSM maxspeed value: if the same edge has a maxspeed:practical (or default or whatever that was), then we prefer that over the plain maxspeed when we set an edge's speed value. 

Ahh, interesting. That's actually good to know... I guess we'll just need to document that's an edge case for now unless another solution presents itself.

Of course there are options if you're willing to run a custom graph / pre-processed OSM data file specific for LSV routing, but that is indeed inconvenient that my assertion isn't fully true out of the box.

> we do have speed_limit on EdgeInfos, but again, we don't want to use that for costing.

I understand that this is bad (slow), but out of curiosity just how "bad" is it? We aren't running hundreds of req/sec yet, but at moderate load we actually had (maybe still have? I forget 😂) deployed an older version of this PR that accessed the `EdgeInfo::speed_limit` and it was acceptable for our use cases.

Would you be opposed to accessing this when a max legal speed limit is specified and clearly documenting the tradeoffs that you're opting into by doing this? It seems useful to me to unlock that use case unless I'm missing something (quite probably). I'm guessing adding `speed_limit` to `DirectedEdge` isn't an option? Looks like that's already packed...

minter:
@ianthetechie Just curious if there's been any advancement here? This could help me with a project I'm working on.

ianthetechie:
Yes we've been running this in production for a few years now. Basically patches on the main Valhalla codebase. It's a bit behind at the moment but I'd still like to see this merged if there's interest :) My current branch including this and a few other patches lives under the stadiamaps fork (prod branch).

ianthetechie:
> you can have a look at this: https://github.com/valhalla/valhalla/pull/2910#discussion_r585817880. I didn't really in a few years, can't remember well what the suggestion was exactly, but I remember thinking it would work before that particular project I was working on went away.

This was probably one of the final sticking points. This data was critical for our use case so we ended up fetching it anyways even though it may not technically be the most performant going out to info from the directed edge.

minter:
@ianthetechie Awesome! The branch I'm looking at has the last commit at about two years ago - just confirming that's the correct one?

ianthetechie:
Take a look at https://github.com/stadiamaps/valhalla/tree/prod. Last commit was yesterday (that's not indicitive of a rapid dev pace / divergence; just happened that I made a patch yesterday 😂)

minter:
@ianthetechie Oh, got it, so I should use y'all's mainline 👍

PR #4493, opened 2024-01-05
clamp truck speed to 90 kph

### LINKED ISSUE #4472: truck `speed` shouldn't always be derived from trafffic
All traffic I know is car-specific and on higher class roads both historical & live traffic can be MUCH higher than any HGV legally allowed speed, in Germany it can > 140 kmh. IMO we shouldn't take any speed for HGV that's higher than any `maxspeed:hgv` (which we translate to `truck_speed`). One would have to look at the OSM wiki for country-specific HGV speeds, but probably we could by default cap it at 90 kmh for HGV (`bus` has a similar problem..).

I'd like to do translate the country-specific maxspeed OSM wiki to our admin database soon anyways, which should eventually set `truck_speed` for lots of countries. Or very future, integrate https://github.com/westnordost/osm-legal-default-speeds (there's a discussion around this already somewhere).

chrstnbwnkl:
# Issue

Fixes #4472 (short term fix, that is) by limiting truck speed to 90 kph, including when traffic speeds are higher. 

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/sif/truckcost.cc
```diff
@@ -484,7 +484,9 @@ Cost TruckCost::EdgeCost(const baldr::DirectedEdge* edge,
                                          &flow_sources, time_info.seconds_from_now)
                         : fixed_speed_;
 
-  auto final_speed = std::min(edge_speed, top_speed_);
+  auto final_speed = std::min(std::min(edge_speed, edge->truck_speed() ? edge->truck_speed()
```
chrstnbwnkl, line 487: Right now we're getting the `DirectedEdge`'s  `truck_speed()` inside `GetSpeed`. I'll try to pull that out of there so that we only call it here in `EdgeCost`.

valhalla/baldr/graphtile.h
```diff
@@ -620,7 +620,7 @@ class GraphTile {
   /**
    * Convenience method for use with costing to get the speed for an edge given the directed
    * edge and a time (seconds since start of the week). If the current speed of the edge
-   * is 0 then the current speed is ignore and other speed sources are used to prevent
+   * is 0 then the current speed is ignored and other speed sources are used to prevent
```
chrstnbwnkl, line 623: sorry couldn't help it :smile:

nilsnolde (changes requested):
looks great, thanks! one thing we forgot about: shortcuts. it's easy though, you'll need to do the same treatment to shortcut building as you did here in EdgeCost. currently the shortcut is built using car speeds as well. 

it starts here and has 2 more occurrences in the builder:
https://github.com/valhalla/valhalla/blob/7eb478460a98bb76d91d7d3bd1a9b2a2e3fe4d63/src/mjolnir/shortcutbuilder.cc#L387-L390

I don't know why it doesn't at least respect the edges' `truck_speed()` but that'll be a good fix as well.

chrstnbwnkl:
In the shortcut builder, it does take into account the edge's `truck_speed` in `GetSpeed` though, doesn't it? So shouldn't it be enough to just clamp it? And why is `top_speed` not taking into account in the shortcuts?

nilsnolde:
> it does take into account the edge's truck_speed in GetSpeed though, doesn't it

that's true, since there's no traffic at this point. we'll refactor away the stuff in GetSpeed eventually, it's fine to just clamp it  for now.

> And why is top_speed not taking into account in the shortcuts?

Well that's a request option but you're in the data build ;)

src/mjolnir/shortcutbuilder.cc
```diff
@@ -297,7 +297,9 @@ uint32_t ConnectEdges(GraphReader& reader,
   // Add edge and turn duration for truck
   // Currently use the same as for cars. TODO: implement for trucks
```
nilsnolde, line 298: maybe remove the (now) stale comment

src/mjolnir/shortcutbuilder.cc
```diff
@@ -295,9 +295,11 @@ uint32_t ConnectEdges(GraphReader& reader,
   total_duration += edge_duration;
 
   // Add edge and turn duration for truck
-  // Currently use the same as for cars. TODO: implement for trucks
   total_truck_duration += turn_duration;
-  auto const truck_speed = tile->GetSpeed(directededge, kNoFlowMask, kInvalidSecondsOfWeek, true);
+  auto const truck_speed =
+      std::min(tile->GetSpeed(directededge, kNoFlowMask, kInvalidSecondsOfWeek, true),
+               directededge->truck_speed() ? directededge->truck_speed() : kMaxAssumedTruckSpeed);
```
chrstnbwnkl, line 301: Truck speed is set on a directed edge only if there's a `maxspeed:hgv` tag on the way. If that is set, then don't clamp it.

src/mjolnir/shortcutbuilder.cc
```diff
@@ -385,7 +387,8 @@ uint32_t AddShortcutEdges(GraphReader& reader,
       assert(speed != 0);
       float total_duration = length / (speed * kKPHtoMetersPerSec);
       uint32_t const truck_speed =
-          tile->GetSpeed(directededge, kNoFlowMask, kInvalidSecondsOfWeek, true);
+          std::min(tile->GetSpeed(directededge, kNoFlowMask, kInvalidSecondsOfWeek, true),
```
chrstnbwnkl, line 390: Same here: only clamp if `truck_speed` is not set on the edge

src/sif/truckcost.cc
```diff
@@ -97,6 +97,7 @@ constexpr ranged_default_t<float> kTruckLengthRange{0, kDefaultTruckLength, 50.0
 constexpr ranged_default_t<float> kUseTollsRange{0, kDefaultUseTolls, 1.0f};
 constexpr ranged_default_t<uint8_t> kAxleCountRange{2, kDefaultAxleCount, 20};
 constexpr ranged_default_t<float> kUseHighwaysRange{0, kDefaultUseHighways, 1.0f};
+constexpr ranged_default_t<float> kTopSpeedRange{10, kMaxAssumedTruckSpeed, kMaxSpeedKph};
```
chrstnbwnkl, line 100: Add a top speed ranged default that defaults to the clamped speed value of 90 KPH.

src/sif/truckcost.cc
```diff
@@ -484,7 +485,9 @@ Cost TruckCost::EdgeCost(const baldr::DirectedEdge* edge,
                                          &flow_sources, time_info.seconds_from_now)
                         : fixed_speed_;
 
-  auto final_speed = std::min(edge_speed, top_speed_);
+  auto final_speed = std::min(std::min(edge_speed, edge->truck_speed() ? edge->truck_speed()
+                                                                       : kMaxAssumedTruckSpeed),
+                              top_speed_);
```
chrstnbwnkl, line 490: Basically the same as in `shortcutbuilder`, but here we need to take into account the `top_speed` as well: we want to let the user specifiy top speeds higher than our new `kMaxAssumedTruckSpeed`, at least up to `kMaxSpeedKph`.

test/gurka/test_ferry_connections.cc
```diff
@@ -119,12 +119,15 @@ TEST(Standalone, ShortFerry) {
 }
 
 TEST(Standalone, TruckFerryDuration) {
-  // corresponds to 33.3 m/sec (120 km/h) in the below map
-  uint32_t ferry_secs = 9;
+  // corresponds to 21.4 m/sec (~77 km/h) in the below map
```
chrstnbwnkl, line 122: Changes to this test are necessary, since ferry route speeds are clamped by this new `kMaxAssumedTruckSpeed` value as well.

If anyone's wondering about how fast vehicle carrying ferries can go, I just went down that rabbit hole: The HSC Francisco running between Buenos Aires and Montevideo holds the world record for the fastest ferry at ~107 kph.[^1] However, that's the top speed and at an average route duration of 2h30mins for ~220 km [^2], even that would not be clamped.

[^1]: https://www.guinnessworldrecords.com/world-records/505447-fastest-ferry
[^2]: https://es.wikipedia.org/wiki/Francisco_(barco)

test/gurka/test_truck_restrictions.cc
```diff
@@ -63,3 +68,100 @@ INSTANTIATE_TEST_SUITE_P(TruckRestrictions,
                                            std::pair<std::string, std::string>{"hazmat", "true"},
                                            std::pair<std::string, std::string>{"axle_load", "11"},
                                            std::pair<std::string, std::string>{"axle_count", "10"}));
+
+TEST(TruckSpeed, MaxTruckSpeed) {
```
chrstnbwnkl, line 72: Add some tests to make sure traffic speeds are clamped

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -117,7 +117,7 @@ These options are available for `auto`, `bus`, and `truck` costing methods.
 | `country_crossing_penalty` | A penalty applied for a country crossing. This penalty can be used to create paths that avoid spanning country boundaries. The default penalty is 0. |
 | `shortest` | Changes the metric to quasi-shortest, i.e. purely distance-based costing. Note, this will disable all other costings & penalties. Also note, `shortest` will not disable hierarchy pruning, leading to potentially sub-optimal routes for some costing models. The default is `false`. |
 | `disable_hierarchy_pruning` | Disable hierarchies to calculate the actual optimal route. The default is `false`. **Note:** This could be quite a performance drainer so there is a upper limit of distance. If the upper limit is exceeded, this option will always be `false`. |
-| `top_speed` | Top speed the vehicle can go. Also used to avoid roads with higher speeds than this value. `top_speed` must be between 10 and 252 KPH. The default value is 140 KPH. |
+| `top_speed` | Top speed the vehicle can go. Also used to avoid roads with higher speeds than this value. `top_speed` must be between 10 and 252 KPH. The default value is 90 KPH for `truck` and 140 KPH for `auto` and `bus`. |
```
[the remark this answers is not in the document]
nilsnolde: Right, thanks

nilsnolde (approved):
:ship:

gknisely (approved):

PR #4494, opened 2024-01-05
fix wrong time zones in time distance matrix

chrstnbwnkl:
# Issue

In time distance matrix, the time zone of the edge id used to settle the last location was used for getting the date time for all locations (there's no issue for this, found out while writing tests for #4491).

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/thor/timedistancematrix.cc
+++ b/src/thor/timedistancematrix.cc
@@ -230,6 +230,9 @@ void TimeDistanceMatrix::ComputeMatrix(Api& request,
     SetOrigin<expansion_direction>(graphreader, origin, time_info);
     SetDestinationEdges();
 
+    // Collect edge_ids used for settling a location
+    std::unordered_map<u_int32_t, baldr::GraphId> dest_edge_ids;
+
     // Find shortest path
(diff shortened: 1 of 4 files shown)


src/thor/timedistancematrix.cc
```diff
@@ -230,6 +230,9 @@ void TimeDistanceMatrix::ComputeMatrix(Api& request,
     SetOrigin<expansion_direction>(graphreader, origin, time_info);
     SetDestinationEdges();
 
+    // Collect edge_ids used for settling a location
```
nilsnolde, line 233: I'd mention that we only need this to determine the right timezone for each destination

nilsnolde (approved):
Thanks, good catch!

I guess we'll first merge #4491, then you can activate the disabled test(s) here for tdmatrix right?

nilsnolde:
there were typos with `u_int32_t` tripping windows. seems other c++ libs have that aliased. fixed it quickly

kevinkreiser (approved):

kevinkreiser (approved):

PR #4253, opened 2023-08-13
our cmake find modules are not robust to pkgconfig installations

nilsnolde:
I tried for way too long to get a clean build of Valhalla on a fresh (more or less) CentOS 8, with dependencies only installed by `vcpkg`. After an [initial struggle](https://github.com/microsoft/vcpkg/issues/33082) with the build environment, it started to build, but couldn't link properly, it was all linking errors for libgeos and libspatialite. I was going down the rabbit hole of brushing up my CMake knowledge and eventually came to the conclusion that our CMake Find modules for those 2 libraries for some reason don't link the dependencies to spatialite (quite a lot) and geos (mostly its own c++ -> c). I don't know why, since it found the libraries when CMake was configuring. I tried a few more things, but it in the end it still couldn't properly link libz.a to spatialite and I just gave up. Though I'm quite sure I almost had it.. I really do wonder how the hell it's working so smoothly on Windows.

The below solution is to use pkg-config instead of relying on the CMake Find modules. That should work for all platforms/environments with pkg-config installed. Also Windows should be fine if `vcpkg` is used as a package manager. I don't have a strong opinion about merging this, I can keep it as a patch downstream.

I'd PR a bit of README soon, now that I have a vcpkg-only Valhalla build without any system package manager or manual builds (no ENABLE_SERVICES yet).

Issue #4048, opened 2023-03-30
Running tests serially seems to break


ianthetechie:
There appear to be some sort of timing issues or implicit dependencies in the unit tests that magically disappear with `-j8` or similar. I noticed that running the tests serially doesn't work.

The only notable CMake flag used is `-DENABLE_SANITIZERS=ON`. The build sequence should essentially mirror CircleCI. I have been able to reproduce on both an M1 Mac and an x86_64 Linux (Debian) box.

```
$ make clean && make -j8 && make utrecht_tiles && make -j8 tests
$ export ASAN_OPTIONS=detect_leaks=0 && make -j1 check

<some output>

[ 48%] Built target worker_nullptr_tiles
[ 48%] Generating worker_nullptr_tiles.log
[FAIL] worker_nullptr_tiles
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from WorkerNullptrTiles
[...]

PR #4500, opened 2024-01-08
migrate CI to M1

nilsnolde:
hoping this will not end in a shit show :crossed_fingers:


nilsnolde:
oh nice, M1 cuts down the build time by half!

conanfile.txt
```diff
@@ -1,22 +1,19 @@
 [requires]
-boost/1.71.0
+boost/1.83.0
```
nilsnolde, line 2: boost 1.71 can't be used anymore on newer osx versions, it uses functions which were removed in c++17

conanfile.txt
```diff
@@ -1,22 +1,19 @@
...
 
 [generators]
 cmake_find_package
 json
 
 [options]
 boost:header_only=True
-boost:without_atomic=True
```
nilsnolde, line 10: it wouldn't install boost without having these libraries as well. it's not compiling anything though, I guess it's just optional dependencies it now expects for some reason

nilsnolde:
gotta be fucking kidding me.. still the same bullshit error as on intel osx.. see here: https://github.com/valhalla/valhalla/pull/4253#discussion_r1444899142. at least those hours debugging it weren't for naught..

src/mjolnir/adminbuilder.cc
```diff
@@ -142,6 +142,7 @@ void buffer_polygon(const polygon_t& polygon, multipolygon_t& multipolygon) {
   auto* outer_ring = geos_helper_t::from_striped_container(polygon.outer());
   std::vector<GEOSGeometry*> inner_rings;
   inner_rings.reserve(polygon.inners().size());
+  std::cout << polygon.inners().size() << std::endl;
```
nilsnolde, line 145: this is crap of course! open for ideas what to do here..

nilsnolde: fwiw, it doesn't help at all to check for its size for some reason, it'll still enter that condition.. kinda reminds me of python generators, which you can call only once. the second time the same instance is called, they're exhausted.

nilsnolde: could also make it a LOG_INFO and pretend it's useful info :D

nilsnolde:
Well finally the `tar_index` bug re-surfaced:)

test/util_midgard.cc
```diff
@@ -416,7 +416,7 @@ TEST(UtilMidgard, TestTrimPolylineWithFloatGeoPoint) {
   // Worst case is they may quantized at 1.69m intervals (for an epsilon change).
   //  https://stackoverflow.com/a/28420164
   // The length comparisons below do better than that, but not a lot.
-  constexpr double MAX_FLOAT_PRECISION = 0.05; // Should be good for 5cm at this lon/lat
+  constexpr double MAX_FLOAT_PRECISION = 0.06; // Should be good for 5cm at this lon/lat
```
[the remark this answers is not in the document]
ianthetechie: Can confirm; I'm using this on my local working copy or else tests fail on Apple Silicon.

src/mjolnir/adminbuilder.cc
nilsnolde: @ianthetechie can you do me a favor and try to run the „admin_uk_overrides“ (smth like that) with this branch, but without this log line? My M1 is still set up with Rosetta and I lack the patience to switch, my intel osx compiles fine, but CI keeps crapping out. On intel I was on clang 14. This seems like an outrageous bug to me😄if we can merge this, we can finally get rid of conan and have full support for vcpkg which would be great.

ianthetechie: Sure thing @nilsnolde 😂 

By "set up with Rosetta" I assume you mean the thing in the docs about configuring your terminal and stuff with Rosetta and running builds there, right? If so then yeah, I ignored that :P

On to the test... it works for me in debug configuration with this log line removed. When I run in release mode, I get an error about `mode_change_count` in `thor/astar_bss.cc` being set but never referenced cuz I guess the assert macro is a no-op in release mode :P Removing that variable and references, it compiles and the `gurka_admin_uk_override` test passes.

ianthetechie: 15 it would seem. Here's my `clang --version` output:

```
Apple clang version 15.0.0 (clang-1500.1.0.2.5)
Target: arm64-apple-darwin23.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

ianthetechie: And yeah, we're running the same Xcode build (for another day or 2 at least; IIRC Xcode 15.2 just dropped and I'll upgrade soon but currently running 15.1).

ianthetechie: I'm also running Boost 1.83 FWIW. Here's my CMake output.

image

nilsnolde: So your env is as close as it can be to CI, yet it passes for you :exploding_head: I'd like to test out GHA, but we only get crap runners there.

ianthetechie: Hmm, yeah that's really strange. Mine succeeds on that particular test.

I can't actually complete the full test suite though. It craps out in the Python section, probably because I haven't configured my environment perfectly for it. (Worth noting btw that newer versions of python/pip will refuse to sudo install packages globally, so we'll probably need to change how that part of the process to use virtual environments etc.)

The only (highly unsatisfying) explanation that this suggests is something peculiar to running tests in parallel. This doesn't make a lot of sense to me, but I've experienced similar issues in the past (ex: https://github.com/valhalla/valhalla/issues/4048; I think this one may have resolved itself now).

nilsnolde: > It craps out in the Python section

Yeah I can see in your CMake output that it can only find the Interpreter, for the bindings you'll need a dev installation (or point CMake to the right libs/dirs).

> Worth noting btw that newer versions of python/pip will refuse to sudo install packages globally

We already took care of that:)

I'm very convinced this is a compiler bug, somehow only manifesting in circleci OSX machines :D In the admins we don't do any parallel stuff, we also don't depend on tiles/graphreader or so around that code. But :shrug: , I'm done with this crap for now really.

test/util_midgard.cc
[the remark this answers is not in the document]
nilsnolde: I also found where I touched it last, in the other PR: https://github.com/valhalla/valhalla/pull/4253/files#r1442377392, where an even bigger precision tolerance is needed

nilsnolde:
This is as mergeable as it'll get with circleci

kevinkreiser (approved):

Issue #3747, opened 2022-09-13
Using Valhalla as a Library May Miss Proper Includes


kevinkreiser:
over in https://github.com/valhalla/valhalla/pull/2192 we tried to address this issue but landed on a different tactic. the essential problem is that we are vendoring header only libraries at build time but when someone downstream links us and builds against our headers they dont get the vendored headers as well. we need to not only vendor the headers when we install our packages we should also change the way we  include them to be the full path `<valhalla/third_party/...>` so that we make sure downstream users are using the same header they built the library with.

related also to #3734

nilsnolde:
I also wonder if it wouldn’t make sense to not vendor rapidjson anymore and instead rely on package managers? In the event of releasing a Valhalla package for a distribution, I guess it’d be a problem if people also have it installed via their package manager (as rapidjson is public API). I didn’t review yet existing versions for apt-get/dnf/.. and the supported distros. 

Smth we could of course also do once we get there. Just popped into my head when reading this.

kevinkreiser:
Vendeoring is a better solution in my opinion as it pins the version. We just have to actually do it correctly :)

nilsnolde:
It’s def nicer for us :) 

I think from a package manager’s perspective it’s not so great when the package is also available as system package (as rapidjson is by now for almost all distros), in my head public deps would create conflicts if they’re both vendored and system-wide (eg as dependency of another system app/lib). 

But yeah, I’ll get back to it once I aspire to package Valhalla for any distro (mid-term I’m considering pacman) :) or there’ll be a tiny patch file.. never mind me😅

kevinkreiser:
> in my head public deps would create conflicts if they’re both vendored and system-wide (eg as dependency of another system app/lib)

yep thats my point, if we vendored them correctly they couldnt conflict, specifically by including them from the thirdparty directory and not elsewhere. eg. `#include <valhalla/third_party/rapidjson/include/rapidjson.h>` then you cannot have a conflict unless some system include or other system library exposes them in their headers and we get them inadvertently. this does bring up another long term thing we should be doing better with which is not leaking thirdparty includes in our headers which goes along the lines of reducing the scope of our headers to only expose the bits of the library that makes sense for people who want to use it as such and not just throwing every function into the header because its convenient. honestly we could have used a lot more discipline with this over the years so the effort to clean it up now is herculean haha

PR #4477, opened 2023-12-24
test files without warning use -Werror flag

cvvergara:
# Issue

Very few test files have warnings.
To avoid having in the future more files with warnings when contributors change them
This PR adds a couple of lists: tests_with_warnings
And all the tests not on that list will have the `-Werror` flag

## Tasklist

- [x] add files with warning to the lists on test and test/gurka
- [x] modify cmake to add the `-Werror` when compiling files **not** on those lists

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #4504, opened 2024-01-11
include vendored libraries properly

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/3747

this is a huge diff unfortunately. the main aim was to make libvalhalla more usable in a super project. for the first time we wanted to implement libvalhalla and installing it doesn't install the vendored headers used in the public API. to make that happen I changed the following:
- changed the path of all submodules from `./third_party` to `./valhalla/third_party`, so we can include things like `#include <valhalla/third_party/rapidjson/include/rapidjson/writer.h>`
- changed all the `#include`s in the public header files to the pattern above
- install headers for `rapidjson`, `date` and for win also `dirent`

I might've included a few cleanup changes as well, I'll check now in the diff again. 

Also, there's still a few things to do:
- it seems a lot of the gurka tests which were flagged warning free in https://github.com/valhalla/valhalla/pull/4477 are showing now warnings with `gtest`, e.g. when comparing `int` with `uint32_t`; those warnings are within `gtest`, I don't know how to shut up compiler warnings originating in gtest. or maybe those warnings should be triggered? I don't know which of my changes would have triggered that..
- when installing the project now, the vendored libraries are installed to e.g. `/usr/local/include/valhalla/third_party/rapidjson`, but right now those directories include the folder tree of the entire git repository, not only the `include` folders. we filter for `*.h/*.hpp` files but the directories are still all there, plus the headers of any sub-submodules, e.g. for rapidjson:
```
/usr/local/include/valhalla/third_party/rapidjson
├── bin
├── CMakeModules
├── contrib
├── doc
├── docker
├── example
├── include
├── test
└── thirdparty
```
but also 
```
/usr/local/include/valhalla/third_party/rapidjson/thirdparty/gtest/googlemock/include/gmock
├── gmock-actions.h
├── gmock-cardinalities.h
├── gmock-generated-actions.h
├── gmock-generated-function-mockers.h
├── gmock-generated-matchers.h
├── gmock-generated-nice-strict.h
├── gmock.h
├── gmock-matchers.h
├── gmock-more-actions.h
├── gmock-more-matchers.h
├── gmock-spec-builders.h
└── internal
``` 

FWIW, it's kinda surprising that we're only having to install 3 vendored libraries (only 2 on UNIX).


CMakeLists.txt
```diff
@@ -116,9 +116,9 @@ if(ENABLE_COVERAGE)
     message(FATAL_ERROR "no genhtml installed")
   endif()
 
-  set(FASTCOV_PATH ${VALHALLA_SOURCE_DIR}/third_party/fastcov/fastcov.py)
+  set(FASTCOV_PATH ${VALHALLA_SOURCE_DIR}/valhalla/third_party/fastcov/fastcov.py)
```
nilsnolde, line 119: this is the majority of changes

CMakeLists.txt
```diff
@@ -293,7 +293,7 @@ if(ENABLE_TOOLS)
     set_target_properties(${program} PROPERTIES FOLDER "Tools")
     create_source_groups("Source Files" ${path})
     target_link_libraries(${program} valhalla)
-    target_include_directories(${program} PUBLIC ${VALHALLA_SOURCE_DIR}/third_party/cxxopts/include)
+    target_include_directories(${program} PRIVATE ${VALHALLA_SOURCE_DIR}/valhalla/third_party/cxxopts/include)
```
nilsnolde, line 296: cxxopts is actually private

bench/meili/mapmatch.cc
```diff
@@ -1,8 +1,8 @@
 #include <iostream>
 #include <sstream>
 
-#include <benchmark/benchmark.h>
 #include <boost/property_tree/ptree.hpp>
+#include <valhalla/third_party/benchmark/include/benchmark/benchmark.h>
```
nilsnolde, line 5: actually this is the majority of changes:)

scripts/format.sh
```diff
@@ -23,7 +23,7 @@ readonly CLANG_FORMAT=$(pwd)/mason_packages/.link/bin/clang-format
 
 echo "Using clang-format $CLANG_FORMAT_VERSION from ${CLANG_FORMAT}"
 
-find src valhalla test bench -type f -name '*.h' -o -name '*.cc' \
+find src valhalla test bench -path "valhalla/third_party" -prune -type f -name '*.h' -o -name '*.cc' \
```
nilsnolde, line 26: skip any files in `valhalla/third_party`

src/CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ endfunction()
 
 function(valhalla_module)
   set(oneValueArgs NAME)
-  set(multiValueArgs SOURCES SOURCES_WITH_WARNINGS HEADERS DEPENDS INCLUDE_DIRECTORIES SYSTEM_INCLUDE_DIRECTORIES)
+  set(multiValueArgs SOURCES SOURCES_WITH_WARNINGS HEADERS HEADERS_VENDORED DEPENDS INCLUDE_DIRECTORIES SYSTEM_INCLUDE_DIRECTORIES)
```
nilsnolde, line 73: each module defines a `HEADERS_VENDORED` pointing to the submodules

src/CMakeLists.txt
```diff
@@ -133,6 +133,13 @@ function(valhalla_module)
   install(FILES ${MODULE_HEADERS}
     DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/valhalla/${MODULE_NAME}"
     COMPONENT development)
+  
+  install(DIRECTORY ${MODULE_HEADERS_VENDORED}
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/valhalla/third_party/"
+    FILES_MATCHING
+    PATTERN "*.h"
+    PATTERN "*.hpp"
+    PATTERN "include")
```
nilsnolde, line 142: this installs properly, but leaves a lot of empty directories in the install tree and sub-submoduled headers, see PR description

src/CMakeLists.txt
```diff
@@ -233,11 +240,11 @@ target_include_directories(valhalla
...
     ${VALHALLA_BINARY_DIR}/valhalla  # TODO: this path must be removed and changed to #include <valhalla/...> in headers
-    ${VALHALLA_SOURCE_DIR}/third_party/rapidjson/include
-    ${VALHALLA_SOURCE_DIR}/third_party/date/include
-    ${VALHALLA_SOURCE_DIR}/third_party/cpp-statsd-client/include
-    $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/third_party/dirent/include>
+    ${VALHALLA_SOURCE_DIR}/valhalla/third_party/rapidjson/include
+    ${VALHALLA_SOURCE_DIR}/valhalla/third_party/date/include
+    $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/valhalla/third_party/dirent/include>
   PRIVATE
+    ${VALHALLA_SOURCE_DIR}/valhalla/third_party/cpp-statsd-client/include
```
nilsnolde, line 247: this is also not public

src/baldr/CMakeLists.txt
```diff
@@ -106,22 +106,26 @@ list(APPEND sources_with_warnings
...
+    $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/valhalla/third_party/dirent>
+    ${VALHALLA_SOURCE_DIR}/valhalla/third_party/date
   HEADERS
     ${headers}
   INCLUDE_DIRECTORIES
     PUBLIC
       ${includes}
     PRIVATE
       ${CMAKE_CURRENT_BINARY_DIR}
+      ${VALHALLA_SOURCE_DIR}/valhalla/third_party/rapidjson/include
```
nilsnolde, line 128: needs rapidjson

src/baldr/tz_alt.cpp
```diff
@@ -96,7 +96,10 @@
 #  include <windows.h>
 #endif  // _WIN32
 
-#include <date/tz_private.h>
+// date emits a warning otherwise for C++17, see
+// https://github.com/valhalla/valhalla/pull/3878#issuecomment-1365487437
+#define HAS_UNCAUGHT_EXCEPTIONS 1
+#include <valhalla/third_party/date/include/date/tz_private.h>
```
nilsnolde, line 102: not sure why this came up now, but we have the same ifdef in `datetime.h`

src/sif/CMakeLists.txt
```diff
@@ -14,21 +14,23 @@ set(sources
...
+    ${VALHALLA_SOURCE_DIR}/valhalla/third_party/rapidjson
   INCLUDE_DIRECTORIES
     PUBLIC
       ${VALHALLA_SOURCE_DIR}
       ${VALHALLA_SOURCE_DIR}/valhalla
-      $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/third_party/dirent/include>
-    PRIVATE
-      ${VALHALLA_SOURCE_DIR}/third_party/rapidjson/include
+      $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/valhalla/third_party/dirent/include>
+      ${VALHALLA_SOURCE_DIR}/valhalla/third_party/rapidjson/include
```
nilsnolde, line 33: rapidjson is actually public for sif

src/tyr/CMakeLists.txt
```diff
@@ -28,13 +28,16 @@ valhalla_module(NAME tyr
...
+  HEADERS_VENDORED
+    $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/valhalla/third_party/dirent>
+    ${VALHALLA_SOURCE_DIR}/valhalla/third_party/rapidjson
   INCLUDE_DIRECTORIES
     PUBLIC
       ${VALHALLA_SOURCE_DIR}
       ${VALHALLA_SOURCE_DIR}/valhalla
-      $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/third_party/dirent/include>
+      $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/valhalla/third_party/dirent/include>
+      ${VALHALLA_SOURCE_DIR}/valhalla/third_party/rapidjson/include
```
nilsnolde, line 39: rapidjson is public for tyr

test/CMakeLists.txt
```diff
@@ -36,7 +35,7 @@ endif()
 ## Lists tests
 set(tests aabb2 access_restriction actor admin attributes_controller configuration datetime directededge
   distanceapproximator double_bucket_queue edgecollapser edgestatus ellipse encode
-  enhancedtrippath factory graphid graphtile graphtileheader gridded_data grid_range_query grid_traversal instructions
+  enhancedtrippath factory geometry graphid graphtile graphtileheader gridded_data grid_range_query grid_traversal instructions
```
nilsnolde, line 38: randomly found out that this test was missing in the list. should increase coverage:)

test/CMakeLists.txt
```diff
@@ -107,21 +106,21 @@ set(tests_with_warnings
...
   create_source_groups("Source Files" ${test}.cc)
   target_link_libraries(${test} valhalla_test)
+
+  if (UNIX AND ENABLE_SINGLE_FILES_WERROR)
+    if (${test} IN_LIST tests_with_warnings)
+      target_compile_options(${test} PRIVATE -Wall)
+    else()
+      target_compile_options(${test} PRIVATE -Wall -Werror)
+    endif()
+  endif()
```
nilsnolde, line 122: I had some problems and thought this might help, but it didn't in the end. anyways, I think this is much safer for consuming projects, those compile options will not leak now.

test/gurka/CMakeLists.txt
```diff
@@ -60,6 +53,15 @@ if(ENABLE_DATA_TOOLS)
...
     create_source_groups("Source Files" ${FILENAME})
     target_link_libraries(${TESTNAME} valhalla_test)
+
+    if (UNIX AND ENABLE_SINGLE_FILES_WERROR)
+      if (${FILENAME} IN_LIST TESTS_WITH_WARNINGS)
+        target_compile_options(${TESTNAME} PRIVATE -Wall)
+      else()
+        target_compile_options(${TESTNAME} PRIVATE -Wall -Werror)
+      endif()
+    endif()
```
nilsnolde, line 63: same here, avoid leaking the compile options

test/gurka/test_avoids.cc
```diff
@@ -137,7 +137,7 @@ TEST_F(AvoidTest, TestMaxPolygonPerimeter) {
   // make sure the right exception is thrown
   try {
     gurka::do_action(Options::route, avoid_map, req);
-  } catch (const valhalla_exception_t& err) { EXPECT_EQ(err.code, 167); } catch (...) {
+  } catch (const valhalla_exception_t& err) { EXPECT_EQ(err.code, 167UL); } catch (...) {
```
nilsnolde, line 140: this is part of the other TODO: for some reason we're triggering now warnings in `gtest` with `--warning-sign-compare`. I started to change it but did only this file. do you have an idea @cvvergara ? I don't see how my changes can make the previously warning-free gurka files now trigger warnings..

src/CMakeLists.txt
nilsnolde: it actually needs to be for testing apparently..

test/CMakeLists.txt
nilsnolde: But also goals the arm build :(

kevinkreiser:
this feels like way more than what is needed. i very much prefer vendored dependencies at the top level in the repo. only the 3 that are exposed in our headers need special treatment (and i feel like we could/should get that down to just rapidjson).

i think you can accomplish everything you want with just:

1. add the bit to the install to actually install the 3 necessary ones into the valhalla target header install directory
2. when using with another project you need to add the valhalla target header install directory as an include path, eg. you need both /usr/local/include and /usr/local/include/valhalla

i think thats it

kevinkreiser:
> FWIW, it's kinda surprising that we're only having to install 3 vendored libraries 

that is on purpose! i try very hard to keep external things out of our headers

nilsnolde:
> i very much prefer vendored dependencies at the top level in the repo.

me too, I guess I was too much focusing on your suggestion in https://github.com/valhalla/valhalla/issues/3747#issuecomment-1250907379.

just going with
- installing the 3 libraries' headers to e.g. `<prefix>/valhalla/third_party/<repo>/<headers>`
- pushing them also in the `CMAKE_BINARY_DIR` I guess, so the `#include`s can stay the way they are
- adding `<prefix>/valhalla/third_party` as include path in other projects

`date` is probably a bit of work to get rid of. `dirent` will be gone once our `filesystem.h` is gone.

kevinkreiser:
ah yeah sorry I wasn't clearer in my suggestion i was referring to where we install vendored headers not thier location in the repo (I hadn't even considered the latter changing)

kevinkreiser:
and yeah regarding pur own headers I'd like to be even more disciplined. using pimples to hide the last 3 deps, moving implementation out of headers for the same reason. regarding date yeah that one is tough but if we go to c++20 we can remove it similar to filesystem with us moving to c++17

PR #4506, opened 2024-01-12
OSRM Serializer: Adds voice instructions

### LINKED ISSUE #4482: Tyr OSRM serializer needs voiceInstructions in step
There is a proprietary extension of the OSRM routing format used in the Mabox Navi SDK and MapLibre Navi SDK, both in iOS and Android. These SDKs work better when the voiceInstructions attribute is present in the steps. Valhalla should have support for these voiceInstructions.

eikes:
# Issue

There is a proprietary extension of the OSRM routing format used in the Mabox Navi SDK and MapLibre Navi SDK, both in iOS and Android. These SDKs work better when the voiceInstructions attribute is present in the steps. 

This PR adds these voice instructions by using the existing `verbal_pre_transition_instruction` and `verbal_post_transition_instruction` in the maneuver.

Fixes #4482

## Tasklist

 - [x] Add tests
 - [ ] ~~If you made changes to the lua files, update the [taginfo](taginfo.json) too.~~


src/tyr/route_serializer_osrm.cc
```diff
@@ -1454,6 +1456,99 @@ void maneuver_geometry(json::MapPtr& step,
...
+    // not really accumulating seconds ourselves, but it happens elsewhere:
+    accumulated_seconds =
+        end_node_elapsed_seconds - etp->node(node_index).cost().elapsed_cost().seconds();
+    accumulated_distance_km += etp->GetCurrEdge(node_index)->length_km();
+  }
+  // The node_index now indicates the node AFTER which the target_seconds will be reached
+  // we now have to subtract the surplus seconds * speed of this edge from the accumulated_distance_km
+  float speed = etp->GetCurrEdge(node_index)->speed(); // speed is in km/h
+  float hours_to_be_removed = (accumulated_seconds - target_seconds) / 3600;
+  accumulated_distance_km -= speed * hours_to_be_removed;
```
kevinkreiser, line 1494: this worries me a little bit because there are a lot of factors that comprise this final value:

https://github.com/valhalla/valhalla/blob/5540cd7d0fcf67b5a1b8ee16fc5d6cbd2360b4c9/src/thor/triplegbuilder.cc#L1135-L1152

i could imagine or i am a tiny bit worried that we'll see a voice instruction at some crazy spot along the maneuver and find a gotcha here somewhere. cant we just do a linear combination of the edge we stop on?

```suggestion
  float accumulated_distance_km = 0;
  float previous_accumulated_distance_km = 0;
  double accumulated_seconds = 0;
  double previous_accumulated_seconds = 0;
  // Find the node after which the instructions should be heard:
  while (accumulated_seconds < target_seconds && node_index >= prev_maneuver->begin_path_index()) {
    node_index -= 1;
    // not really accumulating seconds ourselves, but it happens elsewhere:
    previous_accumulated_seconds = accumulated_seconds;
    accumulated_seconds =
        end_node_elapsed_seconds - etp->node(node_index).cost().elapsed_cost().seconds();
    previous_accumulated_distance_km = accumulated_distance_km;
    accumulated_distance_km += etp->GetCurrEdge(node_index)->length_km();
  }
  // The node_index now indicates the node AFTER which the target_seconds will be reached
  // we now have to subtract the surplus distance (based on seconds) of this edge from the accumulated_distance_km
  auto surplus_percentage = (accumulated_seconds - target_seconds) / (accumulated_seconds - previous_accumulated_seconds);
  accumulated_distance_km -= (accumulated_distance_km - previous_accumulated_distance_km) * surplus_percentage;
```

src/tyr/route_serializer_osrm.cc
```diff
@@ -1454,6 +1456,99 @@ void maneuver_geometry(json::MapPtr& step,
...
+  // distanceAlongGeometry is the distance along the current step from where on this
+  // voice instruction should be heard.
+  // The voice_instruction_beginning starts shortly after the beginning of the step.
+  // The voice_instruction_end starts shortly before the end of the step.
+  float distance_before_maneuver_end = -1;
+  if (prev_maneuver) {
+    distance_before_maneuver_end =
+        distance_along_geometry(prev_maneuver, etp, distance,
+                                SECONDS_BEFORE_MANEUVER_END_FOR_VOICE_INSTRUCTIONS);
+    // This is the the case for the depart maneuver
```
kevinkreiser, line 1517:
```suggestion
    // This is the case for the depart maneuver
```

src/tyr/route_serializer_osrm.cc
```diff
@@ -1454,6 +1456,99 @@ void maneuver_geometry(json::MapPtr& step,
...
+    if (maneuver_index == 1 && !prev_maneuver->verbal_pre_transition_instruction().empty()) {
+      json::MapPtr voice_instruction_start = json::map({});
+      voice_instruction_start->emplace("distanceAlongGeometry", json::fixed_t{distance, 1});
+      voice_instruction_start->emplace("announcement",
+                                       prev_maneuver->verbal_pre_transition_instruction());
+      voice_instructions_array->emplace_back(std::move(voice_instruction_start));
+    } else if (distance > distance_before_maneuver_end + 110 &&
```
kevinkreiser, line 1524: `110`?

src/tyr/route_serializer_osrm.cc
```diff
@@ -1454,6 +1456,99 @@ void maneuver_geometry(json::MapPtr& step,
...
+  // The voice_instruction_end starts shortly before the end of the step.
+  float distance_before_maneuver_end = -1;
+  if (prev_maneuver) {
+    distance_before_maneuver_end =
+        distance_along_geometry(prev_maneuver, etp, distance,
+                                SECONDS_BEFORE_MANEUVER_END_FOR_VOICE_INSTRUCTIONS);
+    // This is the the case for the depart maneuver
+    if (maneuver_index == 1 && !prev_maneuver->verbal_pre_transition_instruction().empty()) {
+      json::MapPtr voice_instruction_start = json::map({});
+      voice_instruction_start->emplace("distanceAlongGeometry", json::fixed_t{distance, 1});
```
kevinkreiser, line 1520: what if `distance` is longer than the whole maneuver?

kevinkreiser:
looking excellent so far. i have to review it a bit more but i have broader questions:

1. verbal multi-cue for short successive maneuvers, we should have unit tests for that because i can imagine that could be tricky here
2. if i remember right, there are two verbal call outs before a maneuver if the maneuver is long enough, i couldnt figure out how you accomplished that. the first should say like "in Xm turn right onto Y street" and then the second sh ould say "turn onto Y street". i guess for now they both say "turn onto Y street" because we don thave the language support for the prepositional phrase "in Xm" right?

src/tyr/route_serializer_osrm.cc
eis-ioki: My (granted limited) understanding of the elapsed time is that it is the quotient of the edge length divided by the edge speed. So getting the remaining distance along the edge would simply be the product of the speed times the remaining time.

What you are linking to suggests, that it is the other way around, the speed is calculated by dividing the length by the elapsed time (if at all). And your worry is that, the speed is not the best value to rely on. Instead you would suggest, that remaining distance along the edge should be calculated based on the length and elapsed time like this:

```
edge_duration = elapsed_time_at_the_end - elapsed_time_at_the_beginning
distance_along_edge = edge_length * remaining_time / edge_duration
```

Correct?

eikes: My (granted limited) understanding of the elapsed time is that it is the quotient of the edge length divided by the edge speed. So getting the remaining distance along the edge would simply be the product of the speed times the remaining time.

What you are linking to suggests, that it is the other way around, the speed is calculated by dividing the length by the elapsed time (if at all). And your worry is that, the speed is not the best value to rely on. Instead you would suggest, that remaining distance along the edge should be calculated based on the length and elapsed time like this:

```
edge_duration = elapsed_time_at_the_end - elapsed_time_at_the_beginning
distance_along_edge = edge_length * remaining_time / edge_duration
```

Correct?

eis-ioki: I applied your suggested changes and made sure the tests still pass and applied clang-format to it.

src/tyr/route_serializer_osrm.cc
eis-ioki: I left a long explanation for this in the comment

kevinkreiser: can you take the comment and the number and make it an ALL CAPS constant at the top of the file?

kevinkreiser (approved):
im sure after trying it out we'll want to tweak it but just having this even untweaked is a great place to start, thank you!

kevinkreiser:
@nilsnolde we can ask contributors to merge master into thier branches  and then merge when we see they have done that. that way we don't pollute the squash merge with other people's commits

src/tyr/route_serializer_osrm.cc
eikes: Done!

eikes:
> looking excellent so far. i have to review it a bit more but i have broader questions:

Thank you!

> 1. verbal multi-cue for short successive maneuvers, we should have unit tests for that because i can imagine that could be tricky here

I am not entirely sure I understand what you mean. Can you give me an example for verbal multi-cue for short successive maneuvers?

> 2. if i remember right, there are two verbal call outs before a maneuver if the maneuver is long enough, i couldnt figure out how you accomplished that. the first should say like "in Xm turn right onto Y street" and then the second sh ould say "turn onto Y street". i guess for now they both say "turn onto Y street" because we don thave the language support for the prepositional phrase "in Xm" right?

I would love to see the prepositional phrase "in Xm". It would tie in nicely with the `distance_along_geometry` function and it would make the verbal instruction a lot more helpful. Alas, I don't know if I can somehow add it. 

The way I approached this problem was to simply take the verbal instructions which are available in the default valhalla serializer and expose them in the OSRM serializer as well. 

These are the verbal instructions available in the maneuver:

1. text_instruction
2. verbal_transition_alert_instruction
3. verbal_succinct_transition_instruction
4. verbal_pre_transition_instruction
5. verbal_post_transition_instruction

Examples:

```json
  "instruction": "Turn left onto Main Street/Route 123. Continue on Route 123.",
  "verbal_transition_alert_instruction": "Turn left onto Main Street.",
  "verbal_succinct_transition_instruction": "Turn left.",
  "verbal_pre_transition_instruction": "Turn left onto Main Street, Route 123.",
  "verbal_post_transition_instruction": "Continue on Route 123 for 2 kilometers.",
```

The `instruction` isn't explicitly marked as a verbal instruction, but it would work. I don't know how it differs from other instructions. It is already exposed in the `maneuver->instruction` key within each OSRM format `step`. 

From the other instructions I really wasn't sure which instructions would be the best to choose. My impression was that the `verbal_pre_transition_instruction` was the most reliable in terms of it actually being available. The `verbal_post_transition_instruction` made the most sense to add, since it compliments the other information nicely. 

Maybe what you are saying is, that I should add the `verbal_transition_alert_instruction` only a couple of seconds before the end of the maneuver? If you don't mind, maybe in another PR?

eikes:
Having just written that I went back to the docs and found this:

https://valhalla.github.io/valhalla/api/turn-by-turn/api-reference/#trip-legs-and-maneuvers

So the actual order should be:

1. `verbal_transition_alert_instruction` a certain distance before the end of the maneuver (using `distance_along_geometry` with 15 seconds)
2. `verbal_pre_transition_instruction` right before the end the maneuver (using `distance_along_geometry` with 3 (?) seconds)
3. Drive the actual maneuver
4. `verbal_post_transition_instruction` if the maneuver is long enough.

eikes:
I changed the code to use both `verbal_transition_alert_instruction` and `verbal_pre_transition_instruction` in the correct order when they are available. 

What do you think @kevinkreiser ? I would love to see this get merged. Is there anything else that I can do to move this forward?

Getting the distance into the `verbal_transition_alert_instruction` is an issue of its own, as it would affect the regular Valhalla serializer... That should happen in a separate PR

kevinkreiser:
let me take a quick look tomorrow and then yeah we can merge

kevinkreiser:
@eikes why has so much of the test been commented out? it kind of looks like maybe so you could focus on one test maybe and you forgot to uncomment the others? there is a way to do that when running with the tests. if this is what you are trying to do let me know

eikes:
My bad! I was only running these tests. I should have checked again. Sorry! I‘ll fix it ASAP

eikes:
@kevinkreiser The tests are back again, sorry. Yes I was trying to only run my specific tests. I am a bit ashamed to admit that I used the commenting method, because I was too lazy to figure out how to run individual tests. Now that it caused this lapse, I will not use that method anymore. I looked it up and I could and should have used this:

```
test/gurka/gurka_osrm_serializer --gtest_filter=*VoiceInstructions*
```

kevinkreiser:
haha yep thats what i was going to tell you about thanks!

kevinkreiser (approved):

eikes:
🥳

PR #4507, opened 2024-01-13
fix libvalhalla.pc generation

nilsnolde:
~~not sure what caused it, either a cmake or a protobuf update (I guess it's the latter exporting wrong things), but with cmake 3.28.1, protobuf 3.25.1 & gcc 13.2.1,~~ (it was our own fault) the `libvalhalla.pc` would link protobuf & abseil like `-lprotobuf-lite.so.25.1.0` including the version. 

this PR extends the regex to write the linker info to also remove anything after `.so`.

also fixes https://github.com/valhalla/valhalla/issues/3747 by installing rapidjson, date & dirent to `<install_dir>/valhalla/third_party`; updated `libvalhalla.pc.in` so it outputs automatically for pkg-config installations.

I tried this on a libvalhalla implementing project with pkg-config, works nicely.


nilsnolde:
I also wonder if we do the right thing here putting those libraries into `Libs: ` in the pc file. shouldn't most, if not all, libraries go into `Requires: `? we're not packaging those libraries but using them. Possibly some of them could even be marked `Requires.private: `..

if I understand correctly, it also seems we're "leaking" some stuff we probably shouldn't explicitly set as required, like `-lm  -lstdc++  -lm  -lgcc_s  -lgcc  -lc`?

nilsnolde:
I changed it around to be a bit more legible (IMO). Now, with all default options set, I get the following `libvalhalla.pc`:
```
prefix=/usr/local
exec_prefix=/usr/local
libdir=${prefix}/lib
includedir=${prefix}/include

Name: libvalhalla
Description: valhalla c++ library
Version: 3.4.0
Libs: -L${libdir} -lvalhalla
Libs.private: -lm
Requires: protobuf-lite zlib spatialite sqlite3 luajit libprime_server
Requires.private: libcurl
Cflags: -I${includedir}
```

AFAIU `Libs*` and `Requires*` are basically the same, only the latter one uses pkg-config to figure out what those packages need themselves. 

`PKG_CONFIG_PATH="/usr/local/lib/pkgconfig" pkg-config --libs libvalhalla` returns

```
-L/usr/local/lib -lvalhalla -lprotobuf-lite -labsl_log_internal_check_op -labsl_leak_check -labsl_die_if_null -labsl_log_internal_conditions -labsl_log_internal_message -labsl_examine_stack -labsl_log_internal_format -labsl_log_internal_proto -labsl_log_internal_nullguard -labsl_log_internal_log_sink_set -labsl_log_sink -labsl_log_entry -labsl_flags -labsl_flags_internal -labsl_flags_marshalling -labsl_flags_reflection -labsl_flags_private_handle_accessor -labsl_flags_commandlineflag -labsl_flags_commandlineflag_internal -labsl_flags_config -labsl_flags_program_name -labsl_log_initialize -labsl_log_globals -labsl_log_internal_globals -labsl_raw_hash_set -labsl_hash -labsl_city -labsl_low_level_hash -labsl_hashtablez_sampler -labsl_statusor -labsl_status -labsl_cord -labsl_cordz_info -labsl_cord_internal -labsl_cordz_functions -labsl_exponential_biased -labsl_cordz_handle -labsl_crc_cord_state -labsl_crc32c -labsl_crc_internal -labsl_crc_cpu_detect -labsl_bad_optional_access -labsl_str_format_internal -labsl_strerror -labsl_synchronization -labsl_graphcycles_internal -labsl_kernel_timeout_internal -labsl_stacktrace -labsl_symbolize -labsl_debugging_internal -labsl_demangle_internal -labsl_malloc_internal -labsl_time -labsl_civil_time -labsl_time_zone -labsl_bad_variant_access -lutf8_validity -lutf8_range -labsl_strings -labsl_string_view -labsl_strings_internal -labsl_base -lrt -labsl_spinlock_wait -labsl_int128 -labsl_throw_delegate -labsl_raw_logging_internal -labsl_log_severity -lspatialite -lminizip -lrttopo -lfreexl -lproj -lz -lgeos_c -lxml2 -lm -lsqlite3 -lluajit-5.1 -lprime_server -lzmq
```

cmake/PkgConfig.cmake
```diff
@@ -1,3 +1,46 @@
...
+function(set_variable_from_rel_or_absolute_path var root rel_or_abs_path)
+  if(IS_ABSOLUTE "${rel_or_abs_path}")
+    set(${var} "${rel_or_abs_path}" PARENT_SCOPE)
+  else()
+    set(${var} "${root}/${rel_or_abs_path}" PARENT_SCOPE)
+  endif()
+endfunction()
+
+# configure a pkg-config file libvalhalla.pc
+function(configure_proj_pc)
```
nilsnolde, line 10: this function is called from the root CMakeLists.txt and sets up the libvalhalla.pc properly depending on the build options.

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(ENABLE_UNDEFINED_SANITIZER "Use UB sanitizer for Debug build" OFF)
 option(ENABLE_TESTS "Enable Valhalla tests" ON)
 option(ENABLE_WERROR "Convert compiler warnings to errors. Requires ENABLE_COMPILER_WARNINGS=ON to take effect" OFF)
 option(ENABLE_BENCHMARKS "Enable microbenchmarking" ON)
+option(INSTALL_VENDORED_HEADERS "Install vendored public headers to <install_dir>/valhalla/third_party" ON)
```
nilsnolde, line 35: I'm open to remove this, but I thought it might be nice to give the option to opt out.

nilsnolde, line 35: I'm open to remove this, but I thought it might be nice to give the option to opt out.

cmake/PkgConfig.cmake
```diff
@@ -1,3 +1,55 @@
...
+  set_variable_from_rel_or_absolute_path("libdir" "$\{prefix\}" "${CMAKE_INSTALL_LIBDIR}")
+  set_variable_from_rel_or_absolute_path("includedir" "$\{prefix\}" "${CMAKE_INSTALL_INCLUDEDIR}")
+  # Build strings of dependencies
+  set(LIBS "")
+  set(REQUIRES "zlib")
+  set(LIBS_PRIVATE "${CMAKE_THREAD_LIBS_INIT}")
+  set(REQUIRES_PRIVATE "")
+  if (INSTALL_VENDORED_LIBS)
+    set(CFLAGS "-I$\{includedir\}/valhalla/third_party")
+  endif()
```
nilsnolde, line 22: add to CFLAGS if they are installed

nilsnolde:
this is super-seded by #4516

PR #4510, opened 2024-01-16
added gha workflow

### LINKED ISSUE #4508: GHA to detect new IANA version
I randomly checked https://github.com/eggert/tz/tags and there's been a new release again a few weeks ago. let's create a small GHA to check if our submodule commit is outdated. it can run on a schedule once a week. then we create a badge to track the status of that workflow in the main README and it'll be right in our faces when we're outdated.

fancy pants extra: auto-create a PR upgrading `tz`, which should just work in the majority of cases. at least it'd work with the current new release.

chrstnbwnkl:
# Issue

Fixes #4508 

This GHA workflow accomplishes the bare minimum: it compares the commit checksums of the commit pointed to by the latest tag and the commit currently checked out in the submodule and fails if they're not equal.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/.github/workflows/check_tz_releases.yml
+++ b/.github/workflows/check_tz_releases.yml
@@ -1,0 +1,31 @@
+name: Check if tz git submodule is up to date
+
+on:
+  schedule:
+    - cron: "0 0 * * 0"
+
+jobs:
+  check_tz:
+    name: Check if latest tag commit differs from the commit the submodule currently points to
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: "recursive"
+          fetch-depth: 0
+
+      - name: Run script
+        shell: bash
+        run: |
+          set -e 
+
+          latest_tag=$(git -C third_party/tz ls-remote --tags --refs --quiet origin | tail --lines 1 | awk '{print $2}' | cut --delimiter='/' --fields=3 )
+          latest_tag_commit=$(git -C third_party/tz rev-list -n 1 $latest_tag)
+          current_commit=$(git -C ./third_party/tz rev-parse HEAD)
+
+          if [[ $latest_tag_commit != $current_commit ]]; then 
+              echo "New tz release available: ${latest_tag}"
+              exit 1
+          fi
+
+          echo "tz up to date"


nilsnolde (approved):

.github/workflows/check_tz_releases.yml
```diff
@@ -0,0 +1,31 @@
+name: Check if tz git submodule is up to date
+
+on:
+  schedule:
+    - cron: "0 0 * * 0"
```
kevinkreiser, line 5: if it doesnt run as part of a pr how are we notified of failure?

nilsnolde: Runs once a week on a schedule. As proposed in #4508, we can add the badge to the readme, so people can see it too.

nilsnolde: Next PR can do that, also bumping the release which shouldn’t fail anything anyways. You ok with that?

nilsnolde: Ah sorry Chris already explained that:)

kevinkreiser: > also bumping the release which shouldn’t fail anything anyways. You ok with that?

you are asking about doing a new release? its fine with me

nilsnolde: No sorry wasn’t clear: I’m asking if you’re ok with having the status badge if this workflow in the readme.

kevinkreiser: i dont mind if its in the readme but i do think its not all that noticable there. do you actually check the readme, i certainly dont hahaha.

it would be cool if did an http post to the github api to add a comment to a never-closed github timezone update issue whenever it fails. then we'd all get a notification. the github api is very trivial to work with the only question i have is how to get perms into the gha runner to do it. surely github has info on this

kevinkreiser: ah ok yeah its trivial too: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

i would just forget the badge and modify the script to curl to the github api a post payload to a timezone update issue when it fails and do nothing when it passes. that would be best imho

kevinkreiser: sure!

PR #4512, opened 2024-01-16
added aggregation logic to filter stage of tile building

gknisely:
# Issue

Added logic during the graph filtering stage to “aggregate” or combine edges that cross a removed pedestrian edge. This logic removes the node and updates directed edges (and edgeinfo) records to combine the 2 edges.  Tested with a full conversion of the planet.  

Logic:

Filtering logic stayed the same except for using the `mode_change` flag to temporarily indicate that we should consider aggregating the edges at this node.  `mode_change` is only set to true when edge attributes match, we don't end at same node (no loops), the edges do not have different names, end node of edges are not in a different tile, no traffic signal, no signs exist at the node(named_intersection), nodeinfo->intersection() not equal to `IntersectionType::kFork`, and we are not one of the following node types:
`NodeType::kGate`
`NodeType::kTollBooth`
`NodeType::kTollGantry`
`NodeType::kBollard`
`NodeType::kSumpBuster`
`NodeType::kBorderControl`  

Next, we loop over the tiles twice to determine if we can aggregate the edges at his node by running a validation check.  If we can't we set `mode_change` back to false at this node.   We can't aggregate edges when we cross into another country as the access can be different in each country.  Moreover, we will not aggregate edges that starts, ends, or is part of a complex restriction, part of a simple restriction, has an access restriction, or has a traffic signal on the edge.  

Also, we toss out edges loop back to the node.  We currently have these in the data now and we should consider how to toss them in the future as we could create islands.  As of today, only saw ~180 of these examples:   [example 1](https://www.openstreetmap.org/way/993706522)  [example 2](https://www.openstreetmap.org/way/975845893)

We continue another pass over the tiles and aggregate the edges.  We have to take into consideration the edge direction and build and set the new shape, length, and curvature for the edge.  The process for the validation and aggregation step was used from the restriction builder logic.   

Finally, we call `UpdateEndNodes` again to set the new end nodes.

Performance results (when excluding pedestrian edges):

Routing benchmark results:

- 17.5799% decrease in the edgelabels_forward size
- 16.8923% decrease in the edgelabels_reverse size

Matrix benchmark results:

- 14.0226% decrease in edgelabels size

Tile size change (when excluding pedestrian edges):

- 10.1876% decrease

Total run time for filter stage is now 1 hour (when excluding pedestrian edges).

## Tasklist

 - [x] Add tests


gknisely:
@nilsnolde I added the test that you requested

src/mjolnir/graphfilter.cc
```diff
@@ -27,10 +29,157 @@ uint32_t n_original_nodes = 0;
...
 uint32_t can_aggregate = 0;
+uint32_t aggregated = 0;
 
 // Group wheelchair and pedestrian access together
 constexpr uint32_t kAllPedestrianAccess = (kPedestrianAccess | kWheelchairAccess);
 
+struct EdgeId {
+  uint64_t way_id;
+  GraphId graph_id;
+};
```
nilsnolde, line 40: seems this isn't used anywhere?

src/mjolnir/graphfilter.cc
```diff
@@ -27,10 +29,157 @@ uint32_t n_original_nodes = 0;
...
+
+bool CanAggregate(const DirectedEdge* de) {
+  if (de->start_restriction() || de->part_of_complex_restriction() || de->end_restriction() ||
+      de->restrictions() || de->traffic_signal() || de->access_restriction()) {
+    return false;
+  }
+  return true;
+}
+
+bool ExpandFromNode(GraphReader& reader,
```
nilsnolde, line 50: these pain me a bit. we have so many small-ish expansion algos distributed across the code base, most probably doing mostly the same thing. nothing to take action right now, but it'd  be really nice if we could consolidate them a bit more "in the future"©

src/mjolnir/graphfilter.cc
```diff
@@ -27,10 +29,157 @@ uint32_t n_original_nodes = 0;
...
+                         uint64_t& way_id,
+                         const graph_tile_ptr& prev_tile,
+                         GraphId prev_node,
+                         GraphId current_node,
+                         const NodeInfo* node_info) {
+
+  for (size_t j = 0; j < node_info->edge_count(); ++j) {
+    GraphId edge_id(prev_tile->id().tileid(), prev_tile->id().level(), node_info->edge_index() + j);
+    const DirectedEdge* de = prev_tile->directededge(edge_id);
+    auto edge_info = prev_tile->edgeinfo(de);
```
nilsnolde, line 80: can we live with a reference here instead of a copy or is that hidden in the `auto`? if it's implicit, let's add the `&` so it's clearer

src/mjolnir/graphfilter.cc
```diff
@@ -27,10 +29,157 @@ uint32_t n_original_nodes = 0;
...
+          return true;
+        }
+        aggregated++;
+
+        bool found;
+        if (visited_nodes.find(de->endnode()) == visited_nodes.end()) {
+          visited_nodes.insert(de->endnode());
+
+          // expand with the same way_id
+          found = ExpandFromNode(reader, shape, en, from_node, isos, forward, last_node,
```
nilsnolde, line 130: this breaks my head a bit, `ExpandFromNode` is calling `ExpandFromNodeInner` which is calling `ExpandFromNode` which is calling `ExpandFromNodeInner` which is calling :laughing:

src/mjolnir/graphfilter.cc
```diff
@@ -27,10 +29,157 @@ uint32_t n_original_nodes = 0;
...
+                    const graph_tile_ptr& prev_tile,
+                    GraphId prev_node,
+                    GraphId current_node) {
+
+  auto tile = prev_tile;
+  if (tile->id() != current_node.Tile_Base()) {
+    tile = reader.GetGraphTile(current_node);
+  }
+
+  auto node_info = tile->node(current_node);
```
nilsnolde, line 162: can you add the `*` here too? helps somewhat when reading the code

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+                       const GraphId& from_node,
+                       const graph_tile_ptr& tile,
+                       const DirectedEdge* directededge) {
+
+  // Get the tile at the end node.  Skip if node is in another tile.
+  // mode_change is not set for end nodes that are in diff tiles
+  if (directededge->endnode().tile_value() == tile->header()->graphid().tile_value()) {
+
+    // original edge data.
+    auto edgeinfo = tile->edgeinfo(directededge);
```
nilsnolde, line 411: similar here, reference, pointer, copy?

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+}
+
+// If we cross into another country we can't aggregate the edges
+// as the access can be different in each country.  Also, bollards
+// and or gates could exist at the node blocking access.
+//
+// Also, we need to handle islands that we created by tossing the
+// pedestrian edges.  For example://
+// https://www.openstreetmap.org/way/993706522
+// https://www.openstreetmap.org/way/975845893
```
nilsnolde, line 442: nice crazy examples :D I guess with default reachability constraints loki wouldn't ever snap inside these islands would it?

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+        processed_nodes.insert(directededge->endnode());
+        no_agg_ways.insert(edgeinfo.wayid());
+        return;
+      }
+
+      std::unordered_set<std::string> isos;
+      bool isForward = directededge->forward();
+      auto id = directededge->endnode();
+
+      isos.insert(tile->admin(sn_info->admin_index())->country_iso()); // start node
```
nilsnolde, line 476: is country_iso enough or we'd even need to consider state boundaries? aren't there different legislations across US states as well?

also would keeping a set of `admin_index()` not be enough instead of the full iso code?

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+        processed_nodes.insert(directededge->endnode());
+        return;
+      }
+
+      // walk in the correct direction.
+      uint64_t wayid = edgeinfo.wayid();
+      if (!Aggregate(id, reader, shape, en, from_node, wayid, isos, isForward)) {
+        LOG_WARN("ValidateData - failed to validate node.  Will not aggregate.");
+        std::cout << "End node: " << directededge->endnode() << " WayId: " << edgeinfo.wayid()
+                  << std::endl;
```
nilsnolde, line 489: debug stuff?

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+
+      if (isos.size() > 1) { // already in diff country
+        processed_nodes.insert(directededge->endnode());
+        return;
+      }
+
+      // walk in the correct direction.
+      uint64_t wayid = edgeinfo.wayid();
+      if (!Aggregate(id, reader, shape, en, from_node, wayid, isos, isForward)) {
+        LOG_WARN("ValidateData - failed to validate node.  Will not aggregate.");
```
nilsnolde, line 487: how noisy is this?

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+        const DirectedEdge* directededge = tile->directededge(idx);
+        if (no_agg_ways.find(tile->edgeinfo(directededge).wayid()) != no_agg_ways.end()) {
+          processed_nodes.insert(directededge->endnode());
+        }
+      }
+    }
+
+    if (reader.OverCommitted()) {
+      reader.Trim();
+    }
```
nilsnolde, line 552: once is enough inside this outer loop no (s. line 533)? maybe we can move it to the end of the loop?

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+          }
+          tilebuilder.AddLaneConnectivity(laneconnectivity);
+        }
+
+        // Names can be different in the forward and backward direction
+        diff_names = tilebuilder.OpposingEdgeInfoDiffers(tile, directededge);
+
+        bool added;
+        auto edgeinfo = tile->edgeinfo(directededge);
+        std::string encoded_shape = edgeinfo.encoded_shape();
```
nilsnolde, line 673: we can probably avoid this copy if we're just overwriting it with the new encoded shape just below?

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+            lc.set_to(tilebuilder.directededges().size());
+          }
+          tilebuilder.AddLaneConnectivity(laneconnectivity);
+        }
+
+        // Names can be different in the forward and backward direction
+        diff_names = tilebuilder.OpposingEdgeInfoDiffers(tile, directededge);
+
+        bool added;
+        auto edgeinfo = tile->edgeinfo(directededge);
```
nilsnolde, line 672: again auto hiding smth or no need for a copy

src/mjolnir/graphfilter.cc
nilsnolde: oh you actually do it at the very end, so I guess both of these can go

src/mjolnir/graphfilter.cc
```diff
@@ -258,6 +784,8 @@ void UpdateEndNodes(GraphReader& reader, std::unordered_map<GraphId, GraphId>& o
       auto iter = old_to_new.find(edge->endnode());
       if (iter == old_to_new.end()) {
         LOG_ERROR("UpdateEndNodes - failed to find associated node");
+        std::cout << std::to_string(edge->endnode().value) << " "
+                  << std::to_string(tile->edgeinfo(edge).wayid()) << std::endl;
```
nilsnolde, line 788: debug?

test/gurka/test_graphfilter.cc
```diff
@@ -0,0 +1,601 @@
...
+
+             M              P         R              W    Y
+             |              |         |              |    |
+             |              |         |              |    |
+        A----B----C----D----E----F----G----H----I----J----K----L
+                  |    |         |         |    |         |
+                  |    |         Q----S----T    |         |
+                  N    O              |         V         X
+                                      |
+                                      U
```
nilsnolde, line 26: to be fully honest, it feels like the maps in here are way more detailed than they need to be, which would add a lot of cognitive overhead if we ever have to manipulate them again.

also, there wasn't ever a graph filter test? good that it's added now:)

test/gurka/test_graphfilter.cc
```diff
@@ -0,0 +1,601 @@
...
+  };
+
+  auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize_metres, {5.11668, 52.07826});
+  auto map =
+      gurka::buildtiles(layout, ways, nodes, {}, work_dir,
+                        {{"mjolnir.admin", {VALHALLA_SOURCE_DIR "test/data/language_admin.sqlite"}},
+                         {"mjolnir.include_pedestrian", "false"}});
+  // CIL should be deleted
+  {
+    GraphReader graph_reader = GraphReader(map.config.get_child("mjolnir"));
```
nilsnolde, line 157: nit: would suffice to have one graph reader for the whole test

nilsnolde (commented):
mostly nits from my side:)

src/mjolnir/graphfilter.cc
```diff
@@ -219,7 +393,359 @@ void FilterTiles(GraphReader& reader,
...
+    std::copy(orig_edges, orig_edges + n, std::back_inserter(directededges));
+
+    nodeid = GraphId(tile_id.tileid(), tile_id.level(), 0);
+    for (uint32_t i = 0; i < tile->header()->nodecount(); ++i, ++nodeid) {
+      NodeInfo nodeinfo = tilebuilder.node(i);
+      bool found = (processed_nodes.find(nodeid) != processed_nodes.end());
+
+      // We can not aggregate at this node.  Turn off the mode change(aggregation) bit
+      if (found) {
+        nodeinfo.set_mode_change(false);
```
kevinkreiser, line 572: can we leave a comment on this bit in nodeinfo that we are repurosing it temporarily for this filtering use-case

src/mjolnir/graphfilter.cc
```diff
@@ -27,10 +29,157 @@ uint32_t n_original_nodes = 0;
...
+                    std::unordered_set<std::string>& isos,
+                    bool forward,
+                    GraphId& last_node,
+                    std::unordered_set<GraphId>& visited_nodes,
+                    uint64_t& way_id,
+                    const graph_tile_ptr& prev_tile,
+                    GraphId prev_node,
+                    GraphId current_node);
+
+bool ExpandFromNodeInner(GraphReader& reader,
```
kevinkreiser, line 63: add javadoc to this to tell how this is used

src/mjolnir/graphfilter.cc
gknisely: Ah leftover from porting code from restriction builder.  ty

src/mjolnir/graphfilter.cc
gknisely: ha agree

src/mjolnir/graphfilter.cc
gknisely: We need the starting shape to build the final shape as we aggregate the edge.  So yes I need to have a copy of this as if we aggregate we update it.

test/gurka/test_graphfilter.cc
gknisely: Yeah I always find real world examples and port them over to gurka.  I have been burned too many times by creating "fake" tests.  This is why they seem so complex.  You can use LL to see the location from where I created the test from.

src/mjolnir/graphfilter.cc
gknisely: yes.  will remove

test/gurka/test_graphfilter.cc
gknisely: This is not the case in all places as we are reconverting the data multiple times.  So I can reuse when we can do so.

src/mjolnir/graphfilter.cc
gknisely: yeah these islands exist in the data now.  Not too many of them, but would love to toss just in case someone started there

src/mjolnir/graphfilter.cc
gknisely: I think that we are fine here.  I followed the same logic used in the graph validator and shortcut builder.  They only look at the country ISO for the end nodes.

test/gurka/test_graphfilter.cc
nilsnolde: jep all good, doesn't really hurt with the mini maps

src/mjolnir/graphfilter.cc
gknisely: that would be my guess too

src/mjolnir/graphfilter.cc
gknisely: I am commenting out.  Validator will catch us if we have an issue.

src/mjolnir/graphfilter.cc
gknisely: ty

gknisely:
@nilsnolde @kevinkreiser @dnesbitt61 this is ready for review again

nilsnolde (approved):
works well for me, thanks:)

gknisely:
@nilsnolde i fixed some formatting.  can you stamp

nilsnolde (approved):

Issue #4514, opened 2024-01-20
Overhaul CMakeLists.txt


nilsnolde:
I recently read a great tutorial on CMake for packaging: https://alexreinking.com/blog/building-a-dual-shared-and-static-library-with-cmake.html. Lots of good pointers on how to modernize a CMake setup. I'd like to take a stab at that over the next weeks. And clean some of the mess we patched over the years.

kevinkreiser:
yeah this is the primary reason I stopped doing the ppa. it was non trivial to do static and dynamic in the same build whereas with autotools its was the default

nilsnolde:
another thing I'd like to add to this, since it will become a topic soon:

so far we have been fairly relaxed about min version of cmake. I've been reading up about it a bit (e.g. [this article](https://alexreinking.com/blog/how-to-use-cmake-without-the-agonizing-pain-part-1.html)). and I think we should require whatever version is best for us to maintain this project. cmake makes it _crazy_ easy to update a version, even tarballs from their homepage are installable without root access. I don't think there's a good argument to make our own life harder just so some devs don't have to make sure their software is up to date. for debian/ubuntu there's kitware PPA's, other platforms are almost always on newest version anyways. and centos/rhel users.. well, they can build from source if need be.

once things are packaged and available over package managers, there's hardly any non-techies having to deal with this anyways.

PR #4516, opened 2024-01-20
clean up cmake

nilsnolde:
preparation for #4514 

cleans/fixes some cmake stuff, e.g.
- make it more modular by outsourcing some functions into `*.cmake` files, e.g. grouping source files for IDEs, finding Valhalla version etc
- make all vendored include directories `SYSTEM` so compiler warnings shut up automatically. I think it's been a problem here and there, e.g. with `date` and can be pretty annoying, especially if there's no more releases as with some of our vendored libs
- quite a bit more minor things, will comment inline

this also includes #4507 which I'd like to merge first. also had to include this log line for the weird osx clang bug.

lastly, it removes conan, which is more trouble than it's worth. all important OSs have versions > 1.71 by now. I felt it's fine to do it here, it's very mechanical and easy to see what's happening.


CMakeLists.txt
```diff
@@ -52,13 +51,10 @@ include(CheckCXXCompilerFlag)
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcolor-diagnostics")
 elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
-  set(COLOR_FLAG "-fdiagnostics-color=auto")
   check_cxx_compiler_flag("-fdiagnostics-color=auto" HAS_COLOR_FLAG)
-  if(NOT HAS_COLOR_FLAG)
-    set(COLOR_FLAG "")
+  if(HAS_COLOR_FLAG)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COLOR_FLAG}")
```
nilsnolde, line 56: pure refactor, logic was more awkward than it had to be

CMakeLists.txt
```diff
@@ -85,18 +81,6 @@ else()
   message(FATAL_ERROR "Unrecognized build type. Use one of Debug, Release, RelWithDebInfo, MinRelSize, None")
 endif()
 
-function(create_source_groups prefix)
```
nilsnolde, line 88: this was moved into its own file and is included below

CMakeLists.txt
```diff
@@ -107,103 +91,47 @@ if(ENABLE_CCACHE AND (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILE
...
-  # bail if there's no conan installed
-  message(STATUS "No compatible boost version detected, using conan...")
-  find_program(conan_FOUND conan)
-  if (NOT conan_FOUND)
-    message(FATAL_ERROR "conan needs to be installed for boost, see https://docs.conan.io/en/latest/installation.html.")
+# cURL
+set(curl_targets "")
+if (ENABLE_HTTP)
+  find_package(CURL)
+  if (CURL_FOUND)
```
nilsnolde, line 105: curl had strange logic before which I honestly didn't quite understand.. I think it was a bit more complicated than it needed to be

CMakeLists.txt
```diff
@@ -212,29 +140,26 @@ else()
   message(FATAL_ERROR "Required target protobuf::libprotobuf-lite or protobuf::libprotobuf is not defined")
 endif()
 
-add_library(libprime_server INTERFACE IMPORTED)
+# libprime_server
+# see https://gitlab.kitware.com/cmake/cmake/-/issues/19467
+set(libprime_server_targets "")
```
nilsnolde, line 145: I think it's better to do this `set(... )` here than `add_library()`

CMakeLists.txt
```diff
@@ -212,29 +140,26 @@ else()
...
-  pkg_check_modules(libprime_server REQUIRED libprime_server>=0.6.3)
-  # workaround for https://gitlab.kitware.com/cmake/cmake/issues/15804
-  find_library(libprime_server_LIBRARY
-    NAME ${libprime_server_LIBRARIES}
-    HINTS ${libprime_server_LIBRARY_DIRS})
-  set_target_properties(libprime_server PROPERTIES
-    INTERFACE_LINK_LIBRARIES "${libprime_server_LIBRARY}"
-    INTERFACE_INCLUDE_DIRECTORIES "${libprime_server_INCLUDE_DIRS}"
-    INTERFACE_COMPILE_DEFINITIONS HAVE_HTTP)
+  pkg_check_modules(libprime_server IMPORTED_TARGET libprime_server>=0.6.3)
```
nilsnolde, line 147: `IMPORTED_TARGET` lets us use `PkgConfig::libprime_server` i.e. looking like a normal cmake target

CMakeLists.txt
```diff
@@ -212,29 +140,26 @@ else()
...
 if(ENABLE_DATA_TOOLS)
   add_compile_definitions(DATA_TOOLS)
   find_package(SQLite3 REQUIRED)
   find_package(SpatiaLite REQUIRED)
-  find_package(LuaJIT)
-  add_library(Lua::Lua INTERFACE IMPORTED)
-  set_target_properties(Lua::Lua PROPERTIES
-    INTERFACE_LINK_LIBRARIES "${LUA_LIBRARIES}"
-    INTERFACE_INCLUDE_DIRECTORIES "${LUA_INCLUDE_DIR}")
+  pkg_check_modules(LuaJIT REQUIRED IMPORTED_TARGET luajit)
```
nilsnolde, line 162: this was also a bit hacky..

CMakeLists.txt
```diff
@@ -255,22 +180,11 @@ if(ENABLE_PYTHON_BINDINGS)
       COMPONENT python)
   else()
     set(ENABLE_PYTHON_BINDINGS OFF)
-    message(WARNING "Skipping Python bindings... Install a development Python version including headers to compile the bindings.")
+    message(WARNING "Python development version not found, skipping Python bindings...")
   endif()
 endif()
 
 ## Executable targets
-function(get_source_path PATH NAME)
```
nilsnolde, line 263: moved to its own cmake module

CMakeLists.txt
```diff
@@ -292,11 +206,11 @@ if(ENABLE_TOOLS)
...
     create_source_groups("Source Files" ${path})
-    target_link_libraries(${program} valhalla)
-    target_include_directories(${program} PUBLIC ${VALHALLA_SOURCE_DIR}/third_party/cxxopts/include)
-    if(MSVC)
-      target_link_libraries(${program} ${CURL_LIBRARIES})
-    endif()
+    target_link_libraries(${program}
+      valhalla
+      # TODO(nils): make this a target-specific dependency? I'm sure it's not many needing this..
+      $<$<BOOL:${CURL_FOUND}>:CURL::libcurl>)
```
nilsnolde, line 212: looking at the sources/headers I'm quite sure libcurl is needed for some executables. I didn't really make the effort to find out which..

CMakeLists.txt
```diff
@@ -369,6 +293,26 @@ if(ENABLE_BENCHMARKS AND ENABLE_DATA_TOOLS AND NOT WIN32)
   add_subdirectory(bench)
 endif()
 
+## Coverage report targets
```
nilsnolde, line 296: moved from somewhere up top for legibility

src/CMakeLists.txt
```diff
@@ -1,47 +1,14 @@
 include(GNUInstallDirs)
+include(ValhallaVersion)
+include(ValhallaWarnings)
 
-list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
-
-## Set relative to the current source and binary paths
+## Set relative binary path
 set(VALHALLA_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
-if(NOT VALHALLA_SOURCE_DIR)
```
nilsnolde, line 7: this was already defined in the root one

src/CMakeLists.txt
```diff
@@ -1,47 +1,14 @@
...
-  set(VALHALLA_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/..)
-endif()
 
-## Get Valhalla version
-file(STRINGS "${VALHALLA_SOURCE_DIR}/valhalla/valhalla.h" version_lines REGEX "VALHALLA_VERSION_(MAJOR|MINOR|PATCH)")
```
nilsnolde, line 12: moved to its own cmake module

src/CMakeLists.txt
```diff
@@ -161,10 +116,6 @@ function(valhalla_module)
   endif()
 endfunction()
 
-add_subdirectory(${VALHALLA_SOURCE_DIR}/third_party/robin-hood-hashing ${CMAKE_BINARY_DIR}/third_party/robin-hood-hashing)
-SET(CPP_STATSD_STANDALONE OFF CACHE BOOL "ignore test targets" FORCE)
-add_subdirectory(${VALHALLA_SOURCE_DIR}/third_party/cpp-statsd-client ${CMAKE_BINARY_DIR}/third_party/cpp-statsd-client)
```
nilsnolde, line 166: these don't need to be added as cmake projects, we only need their headers

src/CMakeLists.txt
```diff
@@ -186,13 +137,13 @@ set(valhalla_hdrs
     ${VALHALLA_SOURCE_DIR}/valhalla/worker.h
     ${VALHALLA_SOURCE_DIR}/valhalla/filesystem.h
     ${VALHALLA_SOURCE_DIR}/valhalla/proto_conversions.h
+    ${VALHALLA_SOURCE_DIR}/valhalla/config.h
```
nilsnolde, line 140: this is now a header we need to install

src/CMakeLists.txt
```diff
@@ -245,13 +196,23 @@ target_link_libraries(valhalla
     ${libvalhalla_link_libraries}
   PRIVATE
     $<$<BOOL:${ENABLE_COVERAGE}>:gcov>
-    Threads::Threads
-    cpp-statsd-client)
```
nilsnolde, line 249: no need for linking

src/baldr/curler.cc
```diff
@@ -6,7 +6,7 @@
 #include <stdexcept>
 #include <string>
 
-#ifdef CURL_STATICLIB
+#ifdef ENABLE_HTTP
```
nilsnolde, line 9: `CURL_STATICLIB` actually means to link a static library, I don't think we should use it to conditionally include this source code

src/loki/status_action.cc
```diff
@@ -33,7 +33,7 @@ time_t get_tileset_last_modified(const std::shared_ptr<GraphReader>& reader) {
 namespace valhalla {
 namespace loki {
 void loki_worker_t::status(Api& request) const {
-#ifdef HAVE_HTTP
+#ifdef ENABLE_HTTP
```
nilsnolde, line 36: might as well use the cmake option name

nilsnolde:
hmpf, OSX has a linker failure on the python lib, which should be fine. worse is that we have a similar segfault as in https://github.com/valhalla/valhalla/pull/4253#issuecomment-1877655919. only this time we're not touching spatialite.. I'll debug that another time

cmake/ValhallaBin2Header.cmake
```diff
@@ -127,7 +127,7 @@ function(BIN2H)
     endif()
 endfunction()
 
-if(CMAKE_SCRIPT_MODE_FILE AND "${CMAKE_SCRIPT_MODE_FILE}" MATCHES "Binary2Header.cmake$")
+if(CMAKE_SCRIPT_MODE_FILE AND "${CMAKE_SCRIPT_MODE_FILE}" MATCHES "ValhallaBin2Header.cmake$")
```
kevinkreiser, line 130: whats the significance of the `Valhalla` prefix on all of these?

CMakeLists.txt
kevinkreiser: i thought normal targets looked more like: `boost::boost`

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(ENABLE_UNDEFINED_SANITIZER "Use UB sanitizer for Debug build" OFF)
 option(ENABLE_TESTS "Enable Valhalla tests" ON)
 option(ENABLE_WERROR "Convert compiler warnings to errors. Requires ENABLE_COMPILER_WARNINGS=ON to take effect" OFF)
 option(ENABLE_BENCHMARKS "Enable microbenchmarking" ON)
+option(INSTALL_VENDORED_HEADERS "Install vendored public headers to <install_dir>/valhalla/third_party" ON)
```
kevinkreiser, line 35: do we really need this option? i get that a person could install their own but then they have 2 new problems:

1. what version to install
2. which ones to install

id rather we dont even give the option and just install what we know will work if they want to do something special they can worry about it. or am i missing an important use-case here?

CMakeLists.txt
```diff
@@ -107,103 +91,51 @@ if(ENABLE_CCACHE AND (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILE
...
+  if (CURL_FOUND)
+    set(curl_targets CURL::libcurl)
+    target_compile_definitions(CURL::libcurl INTERFACE ENABLE_HTTP)
+    if (MSVC)
+      # TODO(nils): this seems too blankety.. in CI python_valhalla fails
+      link_libraries(CURL::libcurl)
+    endif()
+  else()
+    set(ENABLE_HTTP OFF)
+    message(WARNING "cURL not found, skipping ENABLE_HTTP...")
```
kevinkreiser, line 114: why not fail? i know its optional but if its on and it doesnt work better to tell the person hey install curl then continue with the build

CMakeLists.txt
```diff
@@ -107,103 +91,51 @@ if(ENABLE_CCACHE AND (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILE
...
-  # and turn it off again
-  set(CMAKE_FIND_PACKAGE_PREFER_CONFIG OFF)
-endif()
-
-message(STATUS "Using pbf headers from ${PROTOBUF_INCLUDE_DIR}")
-message(STATUS "Using pbf libs from ${PROTOBUF_LIBRARY}")
-message(STATUS "Using pbf release libs from ${PROTOBUF_LIBRARY_RELEASE}")
-message(STATUS "Using pbf debug libs from ${PROTOBUF_LIBRARY_DEBUG}")
+# Protobuf
+# prefer CONFIG mode over MODULE mode, there were a few CMake/protobuf fuckups
```
kevinkreiser, line 125: a bit to spicy for me
```suggestion
# prefer CONFIG mode over MODULE mode, there were a few CMake/protobuf messups
```

CMakeLists.txt
kevinkreiser: is that because you can see the thing that is set versus that being implicit when calling `add_library`?

CMakeLists.txt
```diff
@@ -212,29 +144,26 @@ else()
...
-    INTERFACE_LINK_LIBRARIES "${libprime_server_LIBRARY}"
-    INTERFACE_INCLUDE_DIRECTORIES "${libprime_server_INCLUDE_DIRS}"
-    INTERFACE_COMPILE_DEFINITIONS HAVE_HTTP)
+  pkg_check_modules(libprime_server IMPORTED_TARGET libprime_server>=0.6.3)
+  if (libprime_server_FOUND)
+    target_compile_definitions(PkgConfig::libprime_server INTERFACE ENABLE_SERVICES)
+    set(libprime_server_targets PkgConfig::libprime_server)
+  else()
+    set(ENABLE_SERVICES OFF)
+    message(WARNING "prime_server not found, skipping ENABLE_SERVICES...")
```
kevinkreiser, line 157: same point, i dont think its a good idea to just warn and move on, this is major functionality within the library being disabled. the immediate result of this will be 20 github issues about "i built valhalla and the docs say to run valhalla_service but i cant find it"

CMakeLists.txt
kevinkreiser: it would be all of them if ENABLE_HTTP is turned on and none of them otherwise. this is because curl is used from within the valhalla library which every single one of these executables links. i would just remove the TODO

CMakeLists.txt
```diff
@@ -360,6 +276,18 @@ install(FILES COPYING CHANGELOG.md
...
+if (INSTALL_VENDORED_HEADERS)
+  install(DIRECTORY ${CMAKE_SOURCE_DIR}/third_party/rapidjson/include/
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/valhalla/third_party")
+  install(DIRECTORY ${CMAKE_SOURCE_DIR}/third_party/date/include/
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/valhalla/third_party")
+  if (WIN32)
+    install(DIRECTORY ${CMAKE_SOURCE_DIR}/third_party/dirent/include/
+      DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/valhalla/third_party")
+  endif()
+endif()
```
kevinkreiser, line 289: good catch, but also good motivation to reorganize our public interface to move these to implementation and out of our headers

test/CMakeLists.txt
```diff
@@ -385,7 +386,7 @@ if(ENABLE_SERVICES)
   add_dependencies(run-skadi_service test_directories)
 endif()
 
-if(ENABLE_PYTHON_BINDINGS AND ENABLE_DATA_TOOLS)
+if(ENABLE_PYTHON_BINDINGS)
```
nilsnolde, line 389: revert, this is testing

CMakeLists.txt
nilsnolde: we talked about it in review: I'll remove this option. in the current state of the repo it's pointless. eventually we'll likely get rid of any vendored libs to make peace with package managers and their requirements, but that's pending actual confirmation that vendored header-only libraries aren't ok for compile-time-only dependencies. if it's too hard, we could even incorporate those headers into our repo with proper attribution. anyways, digressing, but this was on some level part of my thought here :D

cmake/ValhallaBin2Header.cmake
nilsnolde: just to make it obvious that its our own modules, not vendored from somewhere like the rest

nilsnolde:
https://github.com/valhalla/valhalla/pull/4516/commits/522b303f68387bdedeaa8b2ce24f95bb11d9fe12: 
- fail cmake config if e.g. ENABLE_SERVICES=ON but libprime_server wasn't found
- switch as many `find_package()` to pkg-config's `pkg_check_modules()`. a few notable exceptions:
  - protobuf: it's non-trivial to get the right path for `protoc` from pkg-config in a platform-agnostic manner. fwiw, using pkg-config to find protobuf takes many seconds because of their lunatic deep-endency trees; also the recent inclusion of `absl` did mess up the .pc file as well apparently, so it's mildly unstable no matter which way we go. the only downside of using cmake here is that providing a custom protobuf installation is a bit more complex, see the building.md where I added a bit of info for that
  - boost: doesn't install a .pc file
  - python: didn't try but the .pc file doesn't look very usable, also 
  - sqlite3: OSX ships with a system sqlite3 which doesn't alllow extensions such spatialite and I'd rather avoid hard-coding some `PKG_CONFIG_PATH`, it'd also lead to many issues "segfaulting in make_spatialite_cache" (observed that while trying it in CI)

https://github.com/valhalla/valhalla/pull/4516/commits/c2f8cd02565e22645c1ac9de1637a4f5146c26f5: removes `INSTALL_VENDORED_HEADERS`

I also added a bit more info in `docs/building.md` about how to use libvalhalla as a dependency and bring your own (protobuf, curl etc) dependencies.

finally, I'll still need to make win work before this can be merged.

scripts/install-linux-deps.sh
```diff
@@ -19,6 +19,7 @@ env DEBIAN_FRONTEND=noninteractive sudo apt install --yes --quiet \
     git \
     jq \
     lcov \
+    libboost-all-dev \
```
nilsnolde, line 22: I thought I saw that debian also packages each header-only boost as dev, but seems I was mistaken. very very annoying..

nilsnolde:
I wonder how windows even gets that far in CI that easily. locally it failed on the `tz` stuff where `make leapseconds` is executed in `baldr/CMakeLists.txt`. I had to install GNU's make & awk and even then it failed bcs the stupid MS shell is too dumb to do a proper `mv`. simply removing the `.out` of `third_party/tz/leapseconds.out` worked and I can finally reproduce the linking failure. I updated the `building.md` with win troubleshooting.

.azure-pipelines.yml
```diff
@@ -54,8 +64,8 @@ jobs:
       - task: Cache@2
         displayName: "Cache vcpkg's packages"
         inputs:
-          key: .\vcpkg.json | "$(VCPKG_REF)" | ".v2"
-          path: "$(VCPKG_DIR)"
+          key: .\vcpkg.json | "$(VCPKG_REF)"
+          path: '%BUILD_DIR%/vcpkg_installed'
```
nilsnolde, line 68: since we're using `vcpkg.json` now it's considered "manifest" mode, which effectively means cmake/vcpkg is figuring out what to install at configuration time, as opposed to "classic" mode where we'd install things manually.

.azure-pipelines.yml
```diff
@@ -70,20 +80,13 @@ jobs:
...
+          %VCPKG_INSTALLATION_ROOT%\vcpkg.exe version
         displayName: 'Install vcpkg'
         condition: ne(variables.CACHE_RESTORED, 'true')
 
-      - script: |
-          %VCPKG_DIR%\vcpkg.exe install
-          %VCPKG_DIR%\vcpkg.exe list
-        displayName: 'Install vcpkg packages'
-        condition: ne(variables.CACHE_RESTORED, 'true')
-
```
nilsnolde, line 86: as far as I understood we don't need to manually install things anymore

.azure-pipelines.yml
nilsnolde: and in manifest mode, vcpkg is installing to build/vcpkg_installed. which explains why we didn't have any cache hits anymore..

CMakeLists.txt
```diff
@@ -296,10 +211,7 @@ if(ENABLE_TOOLS)
     set_target_properties(${program} PROPERTIES FOLDER "Tools")
     create_source_groups("Source Files" ${path})
     target_link_libraries(${program} valhalla)
-    target_include_directories(${program} PUBLIC ${VALHALLA_SOURCE_DIR}/third_party/cxxopts/include)
-    if(MSVC)
-      target_link_libraries(${program} ${CURL_LIBRARIES})
-    endif()
```
nilsnolde, line 302: this wasn't necessary. via baldr we're already linking publicly to curl which propagates to the executables automatically on all platforms

src/mjolnir/ingest_transit.cc
```diff
@@ -18,7 +18,6 @@
 #include <boost/format.hpp>
 #include <boost/property_tree/ptree.hpp>
 #include <boost/tokenizer.hpp>
-#include <curl/curl.h>
```
nilsnolde, line 21: who knows why that was here..

src/CMakeLists.txt
```diff
@@ -245,15 +196,26 @@ target_include_directories(valhalla
 target_link_libraries(valhalla
   PUBLIC
     ${libvalhalla_link_libraries}
+    $<$<BOOL:$(WIN32)>:ws2_32>
```
nilsnolde, line 199: actually this is only needed in baldr & mjolnir, but I couldn't find a way to make it a proper "target" so it could be used in baldr/CMakeLists.txt. meaning it's probably broken when building static libraries for each module, but that's not smth one would usually do on windows anyways I guess

kevinkreiser (approved):

Issue #3587, opened 2022-04-01
New exclude_tolls, exclude_briges and exclude_tunnels costing options


azarz:
Hello!

I'm currently working on isochrones/isodistances using Valhalla, and I have the specific need to exclude tolls, bridges and/or tunnels. I saw that a [PR doing exactly that but with unpaved roads ](https://github.com/valhalla/valhalla/pull/3240) got accepted. If I submit a similar PR, but fulfilling my specific needs, is it likely to be accepted?

Cheers,

azarz:
Yes, that's a specific demand from the end users who specifically want to exclude anything that's above ground. The isocurves will be contained in very restricted areas.

azarz:
I think I'll do the necessary things on a fork anyway because Valhalla is such a great tool for isochrones! I'll submit a pull request when the options are fully developped and tested.

PR #4524, opened 2024-01-24
Hard avoids highway ferry

ADMUKHE:
Issue [#3587](https://github.com/valhalla/valhalla/issues/3587)

This PR is an extension of PR [#3733 ](https://github.com/valhalla/valhalla/pull/3733). In addition to the exclude_bridges, exclude_tunnels ,exclude_tolls options, this PR introduces exclude_ferries, exclude_highways. This new options works similar to exclude_tolls/tunnels/bridges. 

We noticed that in PR [#3733 ](https://github.com/valhalla/valhalla/pull/3733), the PR was not merged since some perf tests are required. We want to help with the perf tests so any hint on where to start with the perf tests would be great! 

The default behaviour of valhalla remains same as the PR [#3733 ](https://github.com/valhalla/valhalla/pull/3733). It returns 400 and by default does not allow the use of hard excludes unless otherwise set using the service_limits.status.allow_hard_exclusions parameter.

The reason for the seperate PR is we are not sure if IGNF is using their branch for other purposes. Nevertheless we added their changes as well since we would anyway need to add the service_limits.status.allow_hard_exclusions parameter. 

Also we did not add the changes in the [#4318](https://github.com/valhalla/valhalla/pull/4318) since that deals with the soft avoid options which is also different from the issue this PR wants to resolve. 

## Tasklist

 - [x] Add tests

## Requirements / Relations

https://github.com/valhalla/valhalla/pull/3733


src/sif/autocost.cc
```diff
@@ -425,7 +425,18 @@ bool AutoCost::Allowed(const baldr::DirectedEdge* edge,
       edge->surface() == Surface::kImpassable || IsUserAvoidEdge(edgeid) ||
       (!allow_destination_only_ && !pred.destonly() && edge->destonly()) ||
       (pred.closure_pruning() && IsClosed(edge, tile)) ||
-      (exclude_unpaved_ && !pred.unpaved() && edge->unpaved()) || !IsHOVAllowed(edge)) {
+      (exclude_unpaved_ && !pred.unpaved() && edge->unpaved()) || !IsHOVAllowed(edge) ||
```
nilsnolde, line 428: can we please:
- initialize a `has_excludes` or so at the beginning of each cost model which is true if _any_ of them is set? this is just way too many checks right now for _every_ edge which are basically never triggered on average
- make this a function in DynamicCost, it's used in every cost model, avoids a lot of duplication

src/loki/worker.cc
```diff
@@ -71,6 +71,16 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
     throw valhalla_exception_t{124};
   }
 
+  auto costings = options.costings();
+  for (const auto& pair : costings) {
+    if (!allow_hard_exclusions &&
+        (pair.second.options().exclude_bridges() || pair.second.options().exclude_tolls() ||
+         pair.second.options().exclude_tunnels() || pair.second.options().exclude_highways() ||
+         pair.second.options().exclude_ferries())) {
+      throw valhalla_exception_t{145};
```
nilsnolde, line 80: make this a warning please, not an exception. there's nothing a user can do about it anyways, so he's forced to send the same request again, might as well let it pass right away instead of throwing.

proto/options.proto
```diff
@@ -312,6 +312,21 @@ message Costing {
     uint32 axle_count = 81;
     float use_lit = 82;
     bool disable_hierarchy_pruning = 83;
+    oneof has_exclude_bridges {
```
nilsnolde, line 315: why `oneof`?`isn't this defaulting to `false`?

test/gurka/test_hard_exclusions.cc
```diff
@@ -0,0 +1,120 @@
...
+#include <gtest/gtest.h>
+
+using namespace valhalla;
+
+namespace {
+const std::vector<std::string> kSupportedCostingModels = {"auto",          "taxi",       "bus",
+                                                          "truck",         "pedestrian", "bicycle",
+                                                          "motor_scooter", "motorcycle"};
+
+const std::vector<std::string> kExclusionParameters = {"exclude_bridges", "exclude_tunnels",
```
nilsnolde, line 12: what about ferry and highway/trunk?

src/sif/autocost.cc
nilsnolde: what about shortcuts btw?

CHANGELOG.md
```diff
@@ -134,6 +134,12 @@
    * ADDED: "has_transit_tiles" & "osm_changeset" to verbose status response [#4062](https://github.com/valhalla/valhalla/pull/4062)
    * ADDED: time awareness to CostMatrix for e.g. traffic support [#4071](https://github.com/valhalla/valhalla/pull/4071)
    * UPDATED: transifex translations [#4102](https://github.com/valhalla/valhalla/pull/4102)
+   * ADDED: Added `exclude_tolls` request parameter [#3733](https://github.com/valhalla/valhalla/pull/3733)
```
nilsnolde, line 137: one entry is enough

nilsnolde (changes requested):
not sure how I feel about all these hard exclusion properties tbh. most will seriously screw up   tons of routes. there's also a question how shortcuts are handled and if it's even doing the right thing in the current state. tests to prove would be good.

nilsnolde:
> We noticed that in PR https://github.com/valhalla/valhalla/pull/3733, the PR was not merged since some perf tests are required. We want to help with the perf tests so any hint on where to start with the perf tests would be great!

1. run a valhalla service with at least europe
2. run this with `time`: https://github.com/valhalla/valhalla/blob/master/run_route_scripts/run_with_server.py with e.g. this file: https://github.com/valhalla/valhalla/blob/master/test_requests/de_benchmark_routes.txt which has a lot of routes of varying lengths
3. do 2. multiple times while 1. is running for both master and your PR; the first run will be meaningless as the service isn't initiated

best do it on a bare minimum linux server with as little background stuff running as possible, to not pollute the pseudo benchmarks.

src/sif/autocost.cc
ADMUKHE: @nilsnolde for shortcuts we do check if edge is shortcut and exclude_bridge/tunnel is true then we do not use it. As per your discussion in [#3733](https://github.com/valhalla/valhalla/pull/3733) you decided on this change! Can you please specify exactly what could be the issue with shortcuts if we exclude edges part of shortcut?

ADMUKHE: https://github.com/valhalla/valhalla/pull/3733#issuecomment-1235883038 here its mentioned that we can disable shortcut use

nilsnolde: the initial PR is 1.5 years old, I have a hard time remembering what I did last week;) that's the problem with opening new PRs based on old ones, I won't go back and read across several PRs. that's why it's important for the person who raises a PR to be as informative as needed for reviewers to have an easy-ish time. 

now that you pointed it out I squinted again at the code and yeah I see the last line in each cost model update handles shortcuts in a way that should be fine (I think).

that's kinda a case in point of above though: those changes are so repetitive it's very easy to miss these small important updates in a review and it'd be very nice to have it commented in the PR so those misunderstandings don't happen. keep in mind we/I do this on a voluntary basis and the easier you make the reviewer's lives the likelier they'll actually put in the unpaid time to do the review. just as a recent example, e.g. in https://github.com/valhalla/valhalla/pull/4516 one can see how reviewers will have an easier time when the PR person is flagging non-trivial changes explicitly and explaining a little bit right where the changes happen (not in the PR description).

chrstnbwnkl:
What's the state of this PR @ADMUKHE? I can help bringing it over the finish line.

ADMUKHE:
@chrstnbwnkl i checked out the avoid ferries PR, can we somehow merge the 2 PRs into one for the reviewers?

proto/options.proto
ADMUKHE: I am wondering then why is it used for has_exclude_cash_only_tolls this also defaults to false!

src/loki/worker.cc
ADMUKHE: In this [comment](https://github.com/valhalla/valhalla/pull/3733#issuecomment-1235883038 ) if it is detected that one of these options is set and its not allowed, return a 400

src/sif/autocost.cc
ADMUKHE: > can we please:
> 
> * initialize a `has_excludes` or so at the beginning of each cost model which is true if _any_ of them is set? this is just way too many checks right now for _every_ edge which are basically never triggered on average

This is Done

> * make this a function in DynamicCost, it's used in every cost model, avoids a lot of duplication

This is Done

Regarding the shortcuts i guess its disabled correctly as mentioned in [Comment](https://github.com/valhalla/valhalla/pull/3733#issuecomment-1242590288)

As you have asked, i commented on the files in the PR for easier review

test/gurka/test_hard_exclusions.cc
ADMUKHE: New Tests are added after refactoring

ADMUKHE:
Refactoring, tests modified and server config is added as requested for hard avoids.

If this PR is merged then it includes PR  [#3733 ](https://github.com/valhalla/valhalla/pull/3733).
It also covers the PR [4812](https://github.com/valhalla/valhalla/pull/4812/files# ) except the exclude ferry for search filter 

@kevinkreiser  [4818](https://github.com/valhalla/valhalla/issues/4818) this is for the search filter part which as you already mentioned is just for the edges we use at the input locations. it doesnt apply to the rest of the route.

Please review further incase any more changes is required.

scripts/valhalla_build_config
ADMUKHE, line 1: Server side config, since this PR is meant for people who wants to experiment so hard exclusions is set to false

src/loki/worker.cc
ADMUKHE, line 1: Throws error if hard exclusion is set to false but hard excludes are requested

src/sif/autocost.cc
ADMUKHE, line 1: When each cost is initialized then has excludes flag are set if any of the excludes is present in the request. Checking for the exclusions is done in the dynamiccost in order to avoid code duplication.

proto/options.proto
ADMUKHE, line 1: oneof is removed

test/gurka/test_hard_exclusions_ferry.cc
ADMUKHE, line 1: Tests for ferry exclusion, no route should be found if hard avoids are impended

valhalla/sif/dynamiccost.h
ADMUKHE, line 1: CheckExclusion function is called if has_exclude flag is initialized to true and only then we check if in the Allowed or AllowedReverse function if the edge is excluded. The shortcut issue is handled here. If edge is a shortcut and one of exclude bridge or exclude tunnel is set then we exclude the edge.

valhalla/sif/edgelabel.h
ADMUKHE, line 1: bridge and tunnel are added to the edge label for easier checks

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -258,6 +258,26 @@ These options are available for transit costing when the multimodal costing mode
...
 | `filters` | A way to filter for one or more ~~`stops`~~ (TODO: need to re-enable), `routes`, or `operators`. Filters must contain a list of so-called Onestop IDs, which is (supposed to be) a unique identifier for GTFS data, and an `action`. The OneStop ID is simply the feeds's directory name and the object's GTFS ID separated by an underscore, i.e. a route with `route_id: AUR` in `routes.txt` from the feed `NYC` would have the OneStop ID `NYC_AUR`, similar with operators/agencies. <ul><li>`ids`: any number of Onestop IDs</li><li>`action`: either `exclude` to exclude all of the `ids` listed in the filter or `include` to include only the `ids` listed in the filter</li></ul>
 
+##### Hard exclusions -> **EXPERIMENTAL**
+
+The following options are available for all costing methods. Those options are not available by default, the server config must have `service_limits.allow_hard_exclusions` set to true in order to allow them. If not allowed and set to true, the server will return a 400 error code. If allowed and set to true, it is highly plausible that no route will be found.
```
kevinkreiser, line 263: better to do as @chrstnbwnkl was doing in another pr and simply add a warning and ignore the hard exclude, imho. previously we didnt have warnings so it wasnt possible but now that we do i think this is a better outcome

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -258,6 +258,26 @@ These options are available for transit costing when the multimodal costing mode
...
+
+| Vehicle Options | Description |
+| :-------------------------- | :----------- |
+| `exclude_bridges` | This value indicates whether or not the path may include bridges. If `exclude_bridges` is set to 1 it is allowed to start and end with bridges, but is not allowed to have them in the middle of the route path, otherwise they are allowed. If set to true, it is highly plausible that no path will be found. Default false. |
```
kevinkreiser, line 267: nit: dont mix 1/0 with true/false. if its boolean just always use true and false

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -258,6 +258,26 @@ These options are available for transit costing when the multimodal costing mode
...
+
+- Turn-by-Turn API: Query a transit route query and parse the returned JSON maneuver  for `transit_info` to find `operator_onestop_id` and the route `onestop_id`. A `transit_stop` contains the `onestop_id` for the stop.
+- [Mobility Explorer](https://github.com/transitland/mobility-explorer): Click a single route, stop, or operator on the map, or use the drop-down menu to find the Onestop ID for routes and operators. The Onestop ID, among other details, is listed in the sidebar.
+- [Transitland](https://transit.land/): Use the Transitland Datastore API to query directly for stops, routes, and operators using a number of options. For example, you can filter for only [subway routes](http://transit.land/api/v1/routes?vehicle_type=metro) or [bus routes](http://transit.land/api/v1/routes?vehicle_type=bus). See the [Transitland Datastore API documentation](https://transit.land/documentation/datastore/api-endpoints.html) for details.
```
kevinkreiser, line 280: did this get removed previously and accidentally added back?

kevinkreiser: we do still support some form of this filtering but we no longer interface with transit.land as it was too cumbersome in practice despite all of the cool things its provides

src/loki/worker.cc
kevinkreiser: i agree with nils, a warning and just setting the option back to false is far better than a 400. the user still gets a usable response but they are told that we couldnt adhere to what they wanted so we gave them the next best thing

src/sif/autocost.cc
kevinkreiser: ok yeah i mentioned that one could disable it on costing but that is really really going to be a performance hit. i wonder how bad it would be all the shortcut builder to care about tunnel and bridge attribution. im tempted to think it might not be so bad compared to the alternative here. an experiment would be worth it at some point and if it proves not too bad we could replace the costing skipping shortcuts with that

kevinkreiser:
@chrstnbwnkl and i can review this on wednesday. to me it looks like this is pretty much done other than the warning and letting the request through. one more thing we should do is build fresh data for it and do RAD testing. i can download the planet tomorrow and do that in advance

docs/docs/api/turn-by-turn/api-reference.md
ADMUKHE: i removed this, this was accidentally added may be

chrstnbwnkl:
Yeah it's fine for me to drop the other one, I'll just leave some comments here. My PR also included the ferry search filter, but that doesn't have to be in scope in this PR of course.

src/loki/worker.cc
```diff
@@ -71,6 +71,16 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
     throw valhalla_exception_t{124};
   }
 
+  auto costings = options.costings();
+  for (const auto& pair : costings) {
+    if (!allow_hard_exclusions &&
+        (pair.second.options().exclude_bridges() || pair.second.options().exclude_tolls() ||
+         pair.second.options().exclude_tunnels() || pair.second.options().exclude_highways() ||
+         pair.second.options().exclude_ferries())) {
```
chrstnbwnkl, line 79: I wonder if using the singular nouns here would be more in line with the matching `use_*` parameters, although the plural sounds a little better

docs/docs/api/turn-by-turn/api-reference.md
ADMUKHE: This is Done

test/gurka/test_hard_exclusions_ferry.cc
```diff
@@ -0,0 +1,86 @@
+#include "gurka.h"
```
chrstnbwnkl, line 1: this can probably go into `test_ferry_connections.cc`

valhalla/sif/edgelabel.h
```diff
@@ -444,12 +459,17 @@ class EdgeLabel {
...
   uint32_t unpaved_ : 1;
   uint32_t has_measured_speed_ : 1;
+
   // Flag if this edge had HGV access
   uint32_t hgv_access_ : 1;
   uint32_t spare : 12;
 
   Cost cost_;      // Cost and elapsed time along the path.
   float sortcost_; // Sort cost - includes A* heuristic.
+                   // Flag indicating edge is a bridge.
```
chrstnbwnkl, line 469: probably better for legibility to move this just above the spare

docs/docs/api/turn-by-turn/api-reference.md
ADMUKHE: This is done

chrstnbwnkl (dismissed):

src/loki/worker.cc
kevinkreiser: i vote for plurals

ADMUKHE: this is just to keep it seperate from exclude options in the search filter!

valhalla/sif/edgelabel.h
ADMUKHE: do you mean // Flag indicating edge is a bridge.
  uint32_t bridge_ : 1;
  // Flag indicating edge is a tunnel.
  uint32_t tunnel_ : 1; after hgv_access and before spare?

ADMUKHE:
> @chrstnbwnkl
> 
> > My PR also included the ferry search filter,
> 
> lets still get that one in though!

can we add tolls and highways search filters as well alon with the ferry? then we can have all hard excludes in the middle of the routes as well as for origin/destinations

chrstnbwnkl:
Correct me if I'm wrong but can't the `max_road_class` search filter used here for highway exclusion?

ADMUKHE:
> Correct me if I'm wrong but can't the `max_road_class` search filter used here for highway exclusion?

max_road_class (string, defaults to "motorway"): highest road class allowed. Acc to the documentation! Then we can squeeze in the toll and ferry in this PR?

kevinkreiser:
yes to both questions. max_road_class handles highway/motorway and yes we can add the other 2 filters

valhalla/sif/edgelabel.h
ADMUKHE: Done

src/loki/worker.cc
ADMUKHE: Done

test/gurka/test_hard_exclusions_ferry.cc
ADMUKHE: I added it as a seperate file since others are also seperate for hard exclusions like unpaved etc. What do you think?

kevinkreiser: we have the problem of having too many tests in separate files. it makes the build times very long because each test file is a separate executable that must be built and run. for that reason, where we can, we like to put tests into existing test files, whichever ones that most closely match the area of functionality that we are testing.

ADMUKHE: Moved!

ADMUKHE:
> yes to both questions. max_road_class handles highway/motorway and yes we can add the other 2 filters

Done

ADMUKHE:
@chrstnbwnkl @kevinkreiser 

I did some more test and there is one particular scenario where the hard avoids (where it is not allowed to have them in the middle of the route path) is difficult to be applied.
Let's say we have a map like this and a test case as the following:

``` const std::string ascii_map = R"(
  E----F----G----H----I----A----J----K----L----O----P----Q----R
                                                  |       |              |        |
                                                  |       |              |        |
                                                  |       |              |        |
                                                  |       |              |        |
                                                 M----N            T----U
  )";

const gurka::ways ways = {
    {"EF", {{"highway", "motorway"}}},
    {"FG", {{"highway", "residential"}}},
    {"GH", {{"highway", "residential"}}},
    {"HI", {{"highway", "residential"}}},
    {"IA", {{"highway", "residential"}}},
    {"IJ", {{"highway", "residential"}}},
    {"JK", {{"highway", "motorway"}}},
    {"KL", {{"highway", "residential"}}},
    {"JM", {{"highway", "residential"}}},
    {"MN", {{"highway", "residential"}}},
    {"NK", {{"highway", "residential"}}},
    {"LO", {{"highway", "motorway"}}},
    {"OP", {{"highway", "motorway"}}},
    {"PQ", {{"highway", "residential"}}},
    {"QR", {{"highway", "residential"}}},
    {"OT", {{"highway", "residential"}}},
    {"TU", {{"highway", "residential"}}},
    {"UP", {{"highway", "residential"}}},
};

TEST_P(ExclusionTestExcludeHighways, InTheBeginningWithPossibleExit) {
  
  check_result("1", {"L", "U"}, {"LO", "OT", "TU"}, map, GetParam()); // this test fail and {"LO", "OP", "UP"} is the route returned
} ```

I want to go from L to U, but LO and OP are motorway. Since it starts on a motorway as per the logic it should go on that way eventhough we have a hard avoid.
But it can make a possible exit to OT since OP is a motorway. 
However the route continues on OP which is the middle of the path and should be avoided based on what we want to acheive here.

It could be due to the logic where we check the predecessor edge is not highway/bridge/toll/tunnel etc in check exclusions in dynamiccost.h.

But is there a way we can make the algorithm take the possible exit if there is any!

**Comment by ADMUKHE** [comment]
> Yeah the predecessor check is what's causing this, and I don't think it's trivial to check for exits when deciding whether or not the edge should be allowed. On the other hand, not having the predecessor check would lead to issues where Valhalla fails when a route starts on the motorway because it can't expand anywhere.
> 
> Not sure if use_highway=0 could help mitigate this at least a little?

Thanks for the suggestions! How far is this PR from being merged?

**Comment by kevinkreiser** [comment]
> How far is this PR from being merged?
this pr is the top priority one for review and merging in my opinion as its the most active. id like to have merged it today but i ended up having a conflict in the morning and so havent reviewed the latest. i also need to look at your test example and see if i can suggest soemthing, havent read it properly yet

**Comment by ADMUKHE** [comment]
> Yeah the predecessor check is what's causing this, and I don't think it's trivial to check for exits when deciding whether or not the edge should be allowed. On the other hand, not having the predecessor check would lead to issues where Valhalla fails when a route starts on the motorway because it can't expand anywhere.
> 
> Not sure if use_highway=0 could help mitigate this at least a little?

Yes, when we put use_#option_name# = 0 along with the hard exclude, we take the nearest possible exit due to the weights assigned to the edges.  Test for L->Q works as expected

but if we start the route on the toll road and there is a possible exit right in that location then setting the use_toll = 0 does not help. Check test for O->Q in the script.

Also when we start with say a "toll", put search filter for exclude_toll(which is supposed to work for locations?) along with hard avoid exclude_tolls then it does not work as I expect. Check test SearchFilterExclusion from x->y. The valhalla query is 

```
{"locations":[{"lat":0.000000,"lon":0.002695,"search_filter":{**"exclude_toll":true**}},{"lat":0.000000,"lon":0.015271}],"costing":"auto","costing_options":{"auto":{"speed_types":["freeflow","constrained","predicted"],**"exclude_tolls":"1"**}},"verbose":true,"shape_match":"map_snap"}
```

Also the user can expect "No possible route found" if there is no edge like OT but i am not sure why its taking OP when clearly OP is a toll road.


I have a small test file which use different combination of hard/soft/search filters, which demonstrates this better. I have marked the test that fails. Could be i am unaware of certain behaviour in valhalla. 

``` 
#include "gurka.h"

#include <gtest/gtest.h>

using namespace valhalla;

namespace {
const std::vector<std::string> kCostingModelsExcludeHighways = {"auto"};

const std::vector<std::string> kExclusionParameters = {"exclude_tolls"};

constexpr double grid_size_meters = 100.;

  L--x-O----P----Q-y---R
       |    |         
       |    |         
       |    |         
       |    |         
       T----U         
  )";

const gurka::ways ways = {
    {"LO", {{"highway", "motorway"},{"toll", "yes"}}},
    {"OP", {{"highway", "motorway"}, {"toll", "yes"}}},    {"PQ", {{"highway", "residential"}}},
    {"QR", {{"highway", "residential"}}}, {"OT", {{"highway", "residential"}}},
    {"TU", {{"highway", "residential"}}}, {"UP", {{"highway", "residential"}}},
};

const auto layout = gurka::detail::map_to_coordinates(ascii_map, grid_size_meters);


void check_result_with_searchfilter_addition(const std::vector<std::string>& waypoints,
                  const std::vector<std::string>& expected_names,
                  const gurka::map& map) {

  const std::string& request_filtered =
      (boost::format(
           R"({"locations":[{"lat":%s,"lon":%s,"search_filter":{"exclude_toll":true}},{"lat":%s,"lon":%s}],"costing":"auto","costing_options":{"auto":{"speed_types":["freeflow","constrained","predicted"],"exclude_tolls":"1"}},"verbose":true,"shape_match":"map_snap"})") %
       std::to_string(map.nodes.at(waypoints[0]).lat()) % std::to_string(map.nodes.at(waypoints[0]).lng()) %
       std::to_string(map.nodes.at(waypoints[1]).lat()) % std::to_string(map.nodes.at(waypoints[1]).lng()))
          .str();
  auto result_filtered = gurka::do_action(valhalla::Options::route, map, request_filtered);

  gurka::assert::raw::expect_path(result_filtered, expected_names);
}

void check_result_with_use_optionname(const std::string& exclude_parameter_value,
                  const std::vector<std::string>& waypoints,
                  const std::vector<std::string>& expected_names,
                  const gurka::map& map,
                  const std::vector<std::string>& props = {}) {

  const std::string& costing = props[0];
  const std::string& exclude_parameter = props[1];
  const std::string& use_parameter = "tolls";
  const std::string& use_parameter_value = "0";
  const auto result = gurka::do_action(valhalla::Options::route, map, waypoints, costing,
                                       {
                                 {"/costing_options/" + costing + "/use_" + use_parameter, use_parameter_value},
                                 {"/costing_options/" + costing + "/" + exclude_parameter, exclude_parameter_value}
                               });
  gurka::assert::raw::expect_path(result, expected_names);
}


};


class HardExclusionTestBug : public ::testing::TestWithParam<std::vector<std::string>> {
protected:
  static gurka::map map;
  static gurka::map mapNotAllowed;

  static void SetUpTestSuite() {
    map = gurka::buildtiles(layout, ways, {}, {}, "test/data/hard_exclude_highways_test",
                            {{"service_limits.allow_hard_exclusions", "true"}});

    mapNotAllowed = gurka::buildtiles(layout, ways, {}, {}, "test/data/hard_exclude_highways_test");
  }
};

gurka::map HardExclusionTestBug::map = {};
gurka::map HardExclusionTestBug::mapNotAllowed = {};

// This one fails
TEST_P(HardExclusionTestBug, SearchFilterExclusion) {
  check_result_with_searchfilter_addition({"x", "y"}, {"OT", "TU", "UP","PQ","QR"}, map);
}

//This one fails
TEST_P(HardExclusionTestBug, UseOptionSetZero) {
  check_result_with_use_optionname("1", {"O", "Q"}, {"OT", "TU", "UP","PQ"}, map, GetParam());
}

TEST_P(HardExclusionTestBug, WithPossibleExitUseOptionSetZero) {
  check_result_with_use_optionname("1", {"L", "Q"}, {"LO", "OT", "TU","UP", "PQ"}, map, GetParam());
}



INSTANTIATE_TEST_SUITE_P(ExclusionTestBugSoftHardSuite,
                         HardExclusionTestBug,
                         ::testing::ValuesIn([]() {
                           std::vector<std::vector<std::string>> values;
                           for (const auto& costing : kCostingModelsExcludeHighways) {
                             for (const auto& param : kExclusionParameters) {
                               values.push_back({costing, param});
                             }
                           }
                           return values;
                         }()));



```

**Comment by chrstnbwnkl** [comment]
The way most `use_*` parameters work is that these values between 0 and 1 get converted to a factor that the cost is multiplied with. The interpolation range differs between parameters. For `use_highway`, the value 0 corresponds to a factor of 8, while for `use_tolls` it's 4. That's why

**Comment by ADMUKHE** [comment]
> The way most `use_*` parameters work is that these values between 0 and 1 get converted to a factor that the cost is multiplied with. The interpolation range differs between parameters. For `use_highway`, the value 0 corresponds to a factor of 8, while for `use_tolls` it's 4. That's why I was saying that these parameters may be used to mitigate your issue at least a bit, but they don't solve the problem entirely. The only way I see would be to change the internal range that `use_toll` translates to, but I'm not sure that's a good idea w.r.t. consistency of the API.

Do you have any insights on why the search filter exclusion would not work? for example in the test 

```
// This one fails
TEST_P(HardExclusionTestBug, SearchFilterExclusion) {
  check_result_with_searchfilter_addition({"x", "y"}, {"OT", "TU", "UP","PQ","QR"}, map);
}
```

**Comment by chrstnbwnkl** [comment]
Have you checked the resulting path? Seems like what would happen is that the edge OT would be a valid candidate for the start location, so the resulting route would start there

**Comment by ADMUKHE** [comment]
> Have you checked the resulting path? Seems like what would happen is that the edge OT would be a valid candidate for the start location, so the resulting route would start there

actual_names
    Which is: { "OP", "PQ" }

**Comment by kevinkreiser** [comment]
@chrstnbwnkl want to debug some of these test cases tomorrow so we can come back with answers here and suggests on how to move forward?

**Comment by ADMUKHE** [comment]
> @chrstnbwnkl want to debug some of these test cases tomorrow so we can come back with answers here and suggests on how to move forward?

Sounds good and thanks for the update

**Comment by chrstnbwnkl** [comment]
@kevinkreiser and I had a look and it seems like this is a bug that occurs when a location is node snapped. The search filtering logic doesn't seem to get called there, so the toll road ends up being a candidate edge for location x, even though it shouldn't be. This should be an easy fix, just apply the search filter in the part of the candidate search where node snapped locations are handled.

**Comment by johannes-no** [comment]
A short update, @ADMUKHE will be temporarily unavailable so I will do my best to replace her here

**Comment by kevinkreiser** [comment]
did we need to enable or change some tests now that the bug with loki search is taken care of?

**Comment by janusz-anue** [comment]
Heyhey, I wanted to take over this task. Can you point me to what is to be done to get this branch in a state to complete?

**Comment by chrstnbwnkl** [comment]
As far as I can tell, this is almost done. There seem to be two pending things:

1. resolve any conflicts with the base branch
2. possibly add some node snapping tests discussed further up, at least I can't find any in the diff so far

Let's give @ADMUKHE  some time to respond to this, there's no need to rush this from my side :)

**Comment by janusz-anue** [comment]
I will look into the node snapping tasks, @ADMUKHE  will not be able to assist for a long period of time thats why I want to look into this and help out :)

**Comment by chrstnbwnkl** [comment]
Ah sorry, then please go ahead :smile: should be as easy as incorporating the test mentioned [here](https://github.com/valhalla/valhalla/pull/4524#issuecomment-2261682790), it should pass if the is up to date with master.

**Comment by janusz-anue** [comment]
So I updated the branch, added said test and ran it. Now the first and the third test pass as expected, the second one "UseOptionSetZero" still fails:
```
actual_names
    Which is: { "OP", "PQ" }
  expected_names
    Which is: { "OT", "TU", "UP", "PQ" }
```

Its weird because the third test "WithPossibleExitUseOptionSetZero" works as expected. The difference is that it started at 'L' and not on 'O'.

**Comment by chrstnbwnkl** [comment]
Oh yep, that one's not using any search filter, so it will snap to OP as the start location. If this is the only test that checks the behavior when `use_highway` is set to 0, then I'd say fix it so that the start location can start on a non-highway edge and find its way around highways.

**Comment by janusz-anue** [comment]
But if it starts at 'O', why doesn't the router route you through O-T-U-P given that we set exlude_tolls to 1 and use_tolls to 0

**Comment by janusz-anue** [comment]
To my understanding it is because of the predecessor check, that you have also mentioned before. But this would ultimately provide an inconsistency: 
  - if the start is on a hard-avoided edge, the exclude option is ignored until we have left the edges (use_options could help mitigate this)
  - but if the destination is on an hard-avoided edge, the router will give no route because it cant find a path to the edge

**Comment by kevinkreiser** [comment]
ive not looked at this in a bit but next time i meet up with @chrstnbwnkl we'll discuss what needs to be done here about the remaining issue you two are discussing. bear with me!

**Comment by janusz-anue** [comment]
Alright, thanks for the update!

**Comment by chrstnbwnkl** [comment]
So the failing test is not really testing the functionality of neither the `exclude_`s nor the search filters, instead it's testing a soft exclusion, which works under some circumstances, but not all of the time. If you modify the test map like this, it will work the way you want it to: 


```

  L--x-O--c-P----Q-y---R
       |    |         
       T----U         
  )";

const gurka::ways ways = {
    {"LO", {{"highway", "motorway"}, {"toll", "yes"}}},
    {"cP", {{"highway", "motorway"}, {"toll", "yes"}}},
    {"Oc", {{"highway", "motorway"}, {"toll", "yes"}}},
    {"PQ", {{"highway", "residential"}}},
    {"QR", {{"highway", "residential"}}},
    {"OT", {{"highway", "residential"}}},
    {"TU", {{"highway", "residential"}}},
    {"UP", {{"highway", "residential"}}},
};

```

We'd prefer you remove this test though, as it's not directly related to the features you're introducing. If you want hard exclusions after you've started/ended on an excluded edge, there's a way to do that (look into `no_thru_pruning` or  `closure_pruning`). We rather not extend the algorithms, instead we could make an issue about moving much of that pruning logic into costing so we could be more generic about it (and hence accommodate more cases like this).

File: valhalla/sif/edgelabel.h [hunk 12]
```diff
@@ -443,8 +458,14 @@ class EdgeLabel {
...
   uint32_t has_measured_speed_ : 1;
+
   // Flag if this edge had HGV access
   uint32_t hgv_access_ : 1;
+  // Flag indicating edge is a bridge.
+  uint32_t bridge_ : 1;
+  // Flag indicating edge is a tunnel.
+  uint32_t tunnel_ : 1;
+
   uint32_t spare : 12;
```
janusz-anue, line 469: We added two bits for bridge and tunnel label, shouldn't the spare then be subtracted by 2?
So it would be `uint32_t spare : 10`

janusz-anue:
Thanks for the hint with these parameters, I didn't know about them, will test them out. Otherwise the test you mentioned isn't included in this PR, it was just for own testing.

valhalla/sif/edgelabel.h
chrstnbwnkl: Oh yes, good catch!

janusz-anue: Updated it

CHANGELOG.md
```diff
@@ -212,6 +212,14 @@
...
    * ADDED: time awareness to CostMatrix for e.g. traffic support [#4071](https://github.com/valhalla/valhalla/pull/4071)
    * UPDATED: transifex translations [#4102](https://github.com/valhalla/valhalla/pull/4102)
+   * ADDED: Added `exclude_tolls` request parameter [#4524](https://github.com/valhalla/valhalla/pull/4524)
+   * ADDED: Added `exclude_bridges` request parameter [#4524](https://github.com/valhalla/valhalla/pull/4524)
+   * ADDED: Added `exclude_tunnels` request parameter [#4524](https://github.com/valhalla/valhalla/pull/4524)
+   * ADDED: Added `service_limits.allow_hard_exclusions` server config parameter [#4524](https://github.com/valhalla/valhalla/pull/4524)
+   * ADDED: Added `exclude_highways` request parameter [#4524](https://github.com/valhalla/valhalla/pull/4524)
+   * ADDED: Added `exclude_ferries` request parameter [#4524](https://github.com/valhalla/valhalla/pull/4524)
+   * ADDED: Added `exclude_ferry` location search filter [#4524](https://github.com/valhalla/valhalla/pull/4524)
+   * ADDED: Added `exclude_toll` location search filter [#4524](https://github.com/valhalla/valhalla/pull/4524)
```
kevinkreiser, line 222: nit: can we make this a single changelog entry just listing all the new things added. its ok to have multiple sentences describing whats going on

janusz-anue: No problem, its done

janusz-anue:
Is there anything else we can do? :)

chrstnbwnkl:
Sorry for the delay here, we've accumulated a bit of a backlog :sweat_smile: from my side this is good to go, but let's see what @kevinkreiser says, maybe we'll do a final review round.

src/loki/worker.cc
```diff
@@ -69,6 +69,43 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
...
+      if (hard_exclude_options.exclude_ferries()) {
+        hard_exclude_options.set_exclude_ferries(false);
+        exclusion_detected = true;
+      }
+
+      if (exclusion_detected) {
+        add_warning(api, 208);
+      }
+    }
+  }
```
kevinkreiser, line 107: lets make this more succinct:

```suggestion
  bool exclusion_detected = false;
  if (!allow_hard_exclusions) {
    for (auto& pair : options.costings()) {
      auto& opts = pair.second.options();
      exclusion_detected = exclusion_detected || opts.exclude_bridges() || opts.exclude_tolls() || opts.exclude_tunnels() || opts.exclude_highways() || opts.exclude_ferries();
      opts.set_exclude_bridges(false);
      opts.set_exclude_tolls(false);
      opts.set_exclude_tunnels(false);
      opts.set_exclude_highways(false);
      opts.set_exclude_ferries(false);
    }
  }
  if (exclusion_detected)
    add_warning(api, 208);
```

valhalla/sif/dynamiccost.h
```diff
@@ -289,6 +289,52 @@ class DynamicCost {
...
+                       bool exclude_highways,
+                       bool exclude_ferries) const {
+    return (exclude_bridges && !pred.bridge() && edge->bridge()) ||
+           (exclude_tunnels && !pred.tunnel() && edge->tunnel()) ||
+           (exclude_tolls && !pred.toll() && edge->toll()) ||
+           IsExcludedRoadClass(exclude_highways, pred.classification(), edge->classification()) ||
+           (exclude_ferries &&
+            !(pred.use() == baldr::Use::kFerry || pred.use() == baldr::Use::kRailFerry) &&
+            (edge->use() == baldr::Use::kFerry || edge->use() == baldr::Use::kRailFerry)) ||
+           (edge->is_shortcut() && (exclude_bridges || exclude_tunnels));
```
kevinkreiser, line 335: yeah so as i mentioned in my previous comment, this line will be an absolute perf killer. i know we default to false in the config so you cant even use it without opting in however i still wonder if we just remove bridges and tunnels from shortcut building how bad perf in general would drop. my guess is tunnels dont matter but bridges probably really get eaten up by shortcuts. this brings me to another general question why do we want hard avoids of bridges and tunnels? what is the use-case? to me it seems like a poor mans heavy vehicle option, where you arent sure if you'd make it in the height limit of a tunnel or the weight limit of a bridge.

all of that to say i dont want you to make any changes to this. the fact that its opt-in is good enough. i just wanted to ask the question about the use-case.

src/sif/autocost.cc
```diff
@@ -454,7 +456,9 @@ bool AutoCost::AllowedReverse(const baldr::DirectedEdge* edge,
       opp_edge->surface() == Surface::kImpassable || IsUserAvoidEdge(opp_edgeid) ||
       (!allow_destination_only_ && !pred.destonly() && opp_edge->destonly()) ||
       (pred.closure_pruning() && IsClosed(opp_edge, tile)) ||
-      (exclude_unpaved_ && !pred.unpaved() && opp_edge->unpaved()) || !IsHOVAllowed(opp_edge)) {
+      (exclude_unpaved_ && !pred.unpaved() && opp_edge->unpaved()) || !IsHOVAllowed(opp_edge) ||
+      (has_excludes_ && CheckExclusions(opp_edge, pred, exclude_bridges_, exclude_tunnels_,
```
kevinkreiser, line 460: since `has_excludes_` is accessible from inside `CheckExclusions` can we just put it in there? also why do we need to pass these flags to it, its a member function, it has access to all of them, the only operand should be `opp_edge` right?

kevinkreiser:
looks mostly good to me. just had 2 little changes to simplify it a bit

src/loki/worker.cc
janusz-anue: Yep I have applied the suggestion

src/sif/autocost.cc
janusz-anue: I dont know why these flags are implemented as parameters to the function, I think it was an oversight, I took them out and only left edge and pred in it.
I now put `has_excludes_` in the `CheckExclusions`-Function and I also went ahead and removed the `IsExcludedRoadClass`-Function and inlined it in `CheckExclusions` so it is all implemented in there.

kevinkreiser:
this is ready for merge if you wouln't mind just inlining that one function so we dont drop any perf on the costing calls in the tight loop of the main algorithms i think that would be great. thanks again for sticking with this so long i know its been a huge effort and taken a lot of time but i really appreciate your efforts

janusz-anue:
No problem, I just resolved some merge conflicts and marked the function as inline

kevinkreiser (approved):

chrstnbwnkl (approved):

kevinkreiser:
i know this isnt up to date with master but i tried it locally merging master and everything passed

PR #4530, opened 2024-01-25
Fix crash using costmatrix (and potentially timedistancebssmatrix)

### LINKED ISSUE #4526: Unknown: CHECK failed: (index) < (current_size_):
can someone hint me on what might be the problem here? I get this error out of no where (it was working normally)

dnesbitt61:
Add empty strings for date_time, time_zone_offset, and time_zone_name. The matrix serializer expects these arrays to have same size as the time and distance arrays.

#fixes #4526 (note - to test the default limits for pedestrian matrix need to be increased to handle the request listed in the issue).

Note: I left a TODO comment in timedistancebssmatrix as a reminder to set date_time and timezone info (as is done in timedistancematrix).

# Issue

## Tasklist

 - [ ] Add tests

## Requirements / Relations


nilsnolde (approved):
thanks a lot @dnesbitt61! basically it's segfaulting every time a connection isn't found?! that should be happening all the time then.. weird it wasn't caught before.. 

my memory failed me.. before the change, we guaranteed that there's a date_time object always, but not after the recent date_time refactor, so my bad..

amrelhewy09:
hey I was wondering if this PR was merged into the docker image with the latest tag. Thanks in advance!

nilsnolde:
sure, every push to master creates a new docker image.

PR #4533, opened 2024-01-26
GHA for Windows

### LINKED ISSUE #4531: ditch azure for github actions
no idea why I waited so long for this.. see here for a starting point: https://github.com/microsoft/vcpkg/issues/26346#issuecomment-1320604180. GHA has 4 core/16 GB while azure seems to have less than half of that.. a no-brainer.

nilsnolde:
fixes #4531


.github/workflows/win.yml
```diff
@@ -0,0 +1,134 @@
...
+          ${{ github.workspace }}\build\$BUILD_TYPE\valhalla_run_route.exe --config .\test\win\valhalla.json -j "{\"locations\": [{\"lat\": 52.10205, \"lon\": 5.114651}, {\"lat\": 52.093113, \"lon\": 5.100918}], \"costing\": \"auto\"}"
+          ${{ github.workspace }}\build\$BUILD_TYPE\valhalla_run_isochrone.exe --config .\test\win\valhalla.json -j "{\"locations\": [{\"lat\": 52.10205, \"lon\": 5.114651}], \"costing\": \"auto\", \"contours\":[{\"time\":15,\"color\":\"ff0000\"}]}"
+      
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        # only run this if manually invoked or a previous job failed
+        if: ${{ (github.event_name == 'workflow_dispatch' && inputs.debug_enabled) || failure() }}
+        with:
+          detached: true
+          timeout-minutes: 15
```
nilsnolde, line 133: super useful action, opens a SSH session on the github server if:
- we triggered it manually and checked the boolean flag when doing so, in case we'd like to try some stuff on windows
- there was a failure in any step in which case it'll wait for 15 minutes before terminating the SSH session

.github/workflows/win.yml
```diff
@@ -0,0 +1,134 @@
...
+defaults:
+  run:
+    shell: bash
+
+env:
+  BUILD_TYPE: Release
+  MSVC_VERSION: '2022'
+  VCPKG_VERSION: '2024.01.12'
+  VCPKG_INSTALL_OPTIONS: --x-abi-tools-use-exact-versions
+  VCPKG_DISABLE_COMPILER_TRACKING: ON
```
nilsnolde, line 42: these 2 are supposedly making sure that we have less cache misses if the vcpkg toolchain was updated, e.g newer cmake

kevinkreiser (approved):

Issue #4474, opened 2023-12-23
second pass for matrix's unconnected pairs


nilsnolde:
Discovered in TomTom data, couldn't find a proper example in OSM for this. The same request works with OSM which might just be because in OSM there's a few more ways in that not_thru region to carry it over the 256 node limit beyond which we assume `not_thru = false`. But that's just a guess so far. Need to investigate further.

nilsnolde:
Nope, it's not that.. In fact it's a very bad case, where both trees hit a different `no_thru` edge. /route only works on a second pass as well, after disabling `no_thru`. 2 things we should do anyways, both of which would make this matrix call  work:
1. https://github.com/valhalla/valhalla/pull/3812: mark all edges inside a no_thru region
2. Re-run the matrix only for unconnected pairs in a second pass, just like /route. 

Both should be done eventually. I'd prefer to do 2. first, it'd have a bigger impact for matrix, also destonly, hierarchy limit relaxing etc could be dealt with.

PR #4535, opened 2024-01-28
Refactor matrix_action

nilsnolde:
as a preparation to https://github.com/valhalla/valhalla/issues/4474

similar to routing algos, I created a base class for matrix algos and unified the main public `SourceToTarget` function. IMO the logic is a bit nicer now, but also it'll be much much easier to handle second passes this way.

I should add this is really a pure refactor, no functional change is intended.


src/thor/costmatrix.cc
```diff
@@ -49,9 +49,9 @@ class CostMatrix::ReachedMap : public robin_hood::unordered_map<uint64_t, std::v
 
 // Constructor with cost threshold.
 CostMatrix::CostMatrix(const boost::property_tree::ptree& config)
-    : max_reserved_labels_count_(config.get<uint32_t>("max_reserved_labels_count_bidir_dijkstras",
+    : MatrixAlgorithm(config),
+      max_reserved_labels_count_(config.get<uint32_t>("max_reserved_labels_count_bidir_dijkstras",
                                                       kInitialEdgeLabelCountBidirDijkstra)),
-      clear_reserved_memory_(config.get<bool>("clear_reserved_memory", false)),
```
nilsnolde, line 54: this is now in the base class

src/thor/costmatrix.cc
```diff
@@ -132,19 +132,17 @@ void CostMatrix::SourceToTarget(Api& request,
                                 baldr::GraphReader& graphreader,
                                 const sif::mode_costing_t& mode_costing,
                                 const sif::travel_mode_t mode,
-                                const float max_matrix_distance,
-                                const bool has_time,
-                                const bool invariant,
-                                const ShapeFormat& shape_format) {
```
nilsnolde, line 138: I pulled all these out of the signature and set them inside the function instead. was pointless to have them in the signature, it'll all in the `request`, except for `has_time` which is being set via the base class's `set_has_time()`

src/thor/matrix_action.cc
```diff
@@ -21,43 +21,9 @@ constexpr uint32_t kCostMatrixThreshold = 5;
...
-  };
-  auto timedistancematrix = [&]() {
-    if (options.shape_format() != no_shape)
-      add_warning(request, 207);
-    return time_distance_matrix_.SourceToTarget(request, *reader, mode_costing, mode,
-                                                max_matrix_distance.find(costing)->second,
-                                                options.matrix_locations(),
-                                                options.date_time_type() == Options::invariant);
-  };
+MatrixAlgorithm* thor_worker_t::get_matrix_algorithm(Api& request, const bool has_time) {
```
nilsnolde, line 24: the diff in here is impossible to read. will show tmrw in code editor what's going on there. the basic gist is:
- add a `get_matrix_algorithm` function to thor_worker_t (like for routing), to get the properties of that class
- remove the lambdas for the matrix algos and decide here which one to use

src/thor/matrix_action.cc
```diff
@@ -81,32 +47,60 @@ std::string thor_worker_t::matrix(Api& request) {
...
+      check_matrix_time(request, options.prioritize_bidirectional() ? Matrix::CostMatrix
+                                                                    : Matrix::TimeDistanceMatrix);
+
+  // allow all algos to be cancelled
+  for (auto* alg : std::vector<MatrixAlgorithm*>{
+           &costmatrix_,
+           &time_distance_matrix_,
+           &time_distance_bss_matrix_,
+       }) {
+    alg->set_interrupt(interrupt);
```
nilsnolde, line 94: I added the `interrupt` function now for all matrix algos finally

src/thor/timedistancematrix.cc
```diff
@@ -8,33 +8,14 @@
 using namespace valhalla::baldr;
 using namespace valhalla::sif;
 
-namespace {
-static bool IsTrivial(const uint64_t& edgeid,
-                      const valhalla::Location& origin,
```
nilsnolde, line 13: this was duplicated with the exact same function in `PathAlgorithm`

src/tyr/matrix_serializer.cc
```diff
@@ -2,7 +2,7 @@
 
 #include "baldr/json.h"
 #include "proto_conversions.h"
-#include "thor/matrix_common.h"
```
nilsnolde, line 5: I also remove this header and pulled out what was there into the new header

test/matrix.cc
```diff
@@ -220,7 +220,7 @@ TEST(Matrix, test_matrix) {
   CostMatrix cost_matrix;
   cost_matrix.SourceToTarget(request, reader, mode_costing, sif::TravelMode::kDrive, 400000.0);
   auto matrix = request.matrix();
-  for (uint32_t i = 0; i < matrix.times().size(); ++i) {
+  for (int i = 0; i < matrix.times().size(); ++i) {
```
nilsnolde, line 223: made this file warning-free

valhalla/thor/pathalgorithm.h
```diff
@@ -155,32 +155,32 @@ class PathAlgorithm {
...
+/**
+ * Check for path completion along the same edge. Edge ID in question
+ * is along both an origin and destination and origin shows up at the
+ * beginning of the edge while the destination shows up at the end of
+ * the edge.
+ * @param  edgeid       Edge id.
+ * @param  origin       Origin path location information.
+ * @param  destination  Destination path location information.
+ */
+inline bool IsTrivial(const baldr::GraphId& edgeid,
```
nilsnolde, line 169: this is changed from a member function to an inline function

nilsnolde:
Also, in a further step, we could think about somehow merging `PathAlgorithm` with `MatrixAlgorithm` and also add a bit more common functionality around expansion details if possible, so we don't have to maintain so many places doing the same thing. That needs a lot more thought though, but is still kinda my end goal.

valhalla/thor/matrixalgorithm.h
```diff
@@ -0,0 +1,192 @@
...
+
+  // for tracking the expansion of the algorithm visually
+  expansion_callback_t expansion_callback_;
+
+  uint32_t max_reserved_labels_count_;
+
+  // if `true` clean reserved memory for edge labels
+  bool clear_reserved_memory_;
+
+  // resizes all PBF sequences except for date_times
```
chrstnbwnkl, line 116: stale comment

kevinkreiser (approved):

nilsnolde:
fixed the build @kevinkreiser , if you have a sec to re-approve

kevinkreiser (approved):

PR #4536, opened 2024-01-29
second pass for CostMatrix

### LINKED ISSUE #4474: second pass for matrix's unconnected pairs

nilsnolde:
fixes #4474 

based on #4535

EDIT: on second thought, I'll leave TDMatrix as TODO for now. Less to review and less relevant overall. Not even sure it makes sense at all. Both TDMatrix algos don't seem to care about nothru pruning or even destonly (yet), hierarchies are not relevant either. Only thing they care about are conditional destonly but those are super rare.

TODO:
- [x] costmatrix
- ~~[ ] timedistancematrix~~
- ~~[ ] timedistancebssmatrix~~
- [x] tests
- [x] add a config parameter to control this?
- [x] Add a warning how many connections needed a 2nd pass

thoughts:
- need to go through costmatrix again and make sure when a connection A -> B is not found, but B -> A is, that **both** A & B are doing another half-dijkstra, otherwise we don't get the bidirectional benefit; but of course need to additionally make sure they're not looking for anything else.. **EDIT**: checking the code again and running manually through the tests in debugging mode (impossible to prove with a test), I can confirm this is what's happening.


src/thor/bidirectional_astar.cc
```diff
@@ -90,7 +90,7 @@ void BidirectionalAStar::Clear() {
   // Set the ferry flag to false
   has_ferry_ = false;
   // Set not thru pruning to true
-  not_thru_pruning_ = true;
+  set_not_thru_pruning(true);
```
nilsnolde, line 93: there's a function for this, I find it nicer for comprehension to use it

src/thor/costmatrix.cc
```diff
@@ -262,21 +257,28 @@ void CostMatrix::SourceToTarget(Api& request,
...
-  for (auto& connection : best_connection_) {
-    uint32_t target_idx = count % target_location_list.size();
-    uint32_t source_idx = count / target_location_list.size();
+  bool connection_failed = false;
+  for (uint32_t connection_idx = 0; connection_idx < best_connection_.size(); connection_idx++) {
+    auto best_connection = best_connection_[connection_idx];
+    // if this is the second pass we don't have to process previously found ones again
+    if (costing_->pass() > 0 && !(matrix.second_pass(connection_idx))) {
+      continue;
+    }
```
nilsnolde, line 272: if we're on the second pass and we didn't set the flag on the first pass, then don't bother processing anymore

src/thor/costmatrix.cc
```diff
@@ -286,40 +288,31 @@ void CostMatrix::SourceToTarget(Api& request,
...
-      // Add empty strings to make sure pbf arrays are populated (serializer
-      // requires this)
-      auto* pbf_date_time = matrix.mutable_date_times()->Add();
-      *pbf_date_time = "";
-      auto* pbf_time_zone_offset = matrix.mutable_time_zone_offsets()->Add();
-      *pbf_time_zone_offset = "";
-      auto* pbf_time_zone_name = matrix.mutable_time_zone_names()->Add();
-      *pbf_time_zone_name = "";
+      // let's try a second pass for this connection
+      matrix.mutable_second_pass()->Set(connection_idx, true);
```
nilsnolde, line 296: `second_pass` is defaulting to false, `true` is signaling the second pass that it wasn't found initially

src/thor/costmatrix.cc
```diff
@@ -357,10 +350,18 @@ void CostMatrix::Initialize(
...
     for (uint32_t j = 0; j < locs_count_[MATRIX_REV]; j++) {
+      const auto connection_idx = i * static_cast<uint32_t>(target_locations.size()) + j;
       if (equals(source_locations.Get(i).ll(), target_locations.Get(j).ll())) {
         best_connection_.emplace_back(empty, empty, trivial_cost, 0.0f);
         best_connection_.back().found = true;
+      } else if (costing_->pass() > 0 && !matrix.second_pass(connection_idx)) {
+        // we've found this connection in a previous pass, we only need the time & distance
+        best_connection_.emplace_back(empty, empty, Cost{0.0f, matrix.times(connection_idx)},
+                                      matrix.distances(connection_idx));
+        best_connection_.back().found = true;
```
nilsnolde, line 361: if we have previously found connections, we transfer the important attributes which make up the response

src/thor/costmatrix.cc
```diff
@@ -370,15 +371,21 @@ void CostMatrix::Initialize(
 
   // Set the remaining number of sources and targets
   locs_remaining_[MATRIX_FORW] = 0;
-  for (const auto& s : locs_status_[MATRIX_FORW]) {
+  for (auto& s : locs_status_[MATRIX_FORW]) {
     if (!s.unfound_connections.empty()) {
       locs_remaining_[MATRIX_FORW]++;
+    } else {
+      // don't look at sources which don't have unfound connections, important for second pass
+      s.threshold = 0;
```
nilsnolde, line 379: setting the threshold to 0 for sources which have all destination found on the first pass, makes sure we don't even expand those sources anymore

src/thor/costmatrix.cc
```diff
@@ -538,7 +550,7 @@ bool CostMatrix::ExpandInner(baldr::GraphReader& graphreader,
                         pred_dist, newcost.cost);
   }
 
-  return true;
+  return !(pred.not_thru_pruning() && meta.edge->not_thru());
```
nilsnolde, line 553: allow u-turns if we hit a not thru edge

src/thor/route_action.cc
```diff
@@ -312,6 +312,7 @@ std::vector<std::vector<thor::PathInfo>> thor_worker_t::get_path(PathAlgorithm*
 
   // Check if we should run a second pass pedestrian route with different A*
   // (to look for better routes where a ferry is taken)
+  // TODO(nils): how would a second pass find a better route, if it changes nothing ferry-related?
```
nilsnolde, line 315: this seemed strange to me

src/thor/matrix_action.cc
```diff
@@ -81,32 +51,85 @@ std::string thor_worker_t::matrix(Api& request) {
...
+
+  auto* algo = get_matrix_algorithm(request, has_time, costing);
+  LOG_INFO("matrix::" + std::string(algo->name()));
+
+  // TODO(nils): TDMatrix doesn't care about either destonly or no_thru
+  if (algo->name() != "costmatrix") {
+    algo->SourceToTarget(request, *reader, mode_costing, mode,
+                         max_matrix_distance.find(costing)->second);
+    return tyr::serializeMatrix(request);
+  }
```
nilsnolde, line 110: we don't do second passes for tdmatrices (bss or normal), both don't have a concept of neither destonly nor not_thru

nilsnolde: actually unidir a* also allows for destonly, not sure why?

src/thor/costmatrix.cc
```diff
@@ -286,40 +288,31 @@ void CostMatrix::SourceToTarget(Api& request,
...
-      *pbf_date_time = dt_info.date_time;
-
-      auto* pbf_time_zone_offset = matrix.mutable_time_zone_offsets()->Add();
-      *pbf_time_zone_offset = dt_info.time_zone_offset;
-
-      auto* pbf_time_zone_name = matrix.mutable_time_zone_names()->Add();
-      *pbf_time_zone_name = dt_info.time_zone_name;
+      *matrix.mutable_date_times(connection_idx) = dt_info.date_time;
+      *matrix.mutable_time_zone_offsets(connection_idx) = dt_info.time_zone_offset;
+      *matrix.mutable_time_zone_names(connection_idx) = dt_info.time_zone_name;
```
nilsnolde, line 293: I also changed this whole logic, it became unbearable now that we're "randomly" processing connections, i.e. not in sequence anymore. I realized that `RepeatedPtrFields` don't have a `Set()` method but also their `operator()` methods aren't const, so we can just assign new values. probably that's quite a bit worse for performance, but I doubt it's very measurable wrt to the whole expansion.

nilsnolde: basically now we're initializing all `repeated string` fields with `""` in `reserve_pbf_arrays()`, just like the other fields which have `Set()`, in a way reverting some of what was done in #4530

valhalla/thor/costmatrix.h
```diff
@@ -99,55 +97,29 @@ class CostMatrix {
...
-   * Set a callback that will throw when the path computation should be aborted
-   * @param interrupt_callback  the function to periodically call to see if
-   *                            we should abort
+   * Get the algorithm's name
+   * @return the name of the algorithm
    */
-  void set_interrupt(const std::function<void()>* interrupt_callback) {
-    interrupt_ = interrupt_callback;
+  inline std::string_view name() override {
+    return "costmatrix";
```
nilsnolde, line 118: this should probably be a `constexpr`

nilsnolde: Actually we can use the proto_conversion stuff here too, so we only have one place where we hard-code those names

valhalla/thor/matrixalgorithm.h
```diff
@@ -113,16 +149,26 @@ class MatrixAlgorithm {
...
+  // on first pass, resizes all PBF sequences and defaults to 0 or ""
+  inline static void reserve_pbf_arrays(valhalla::Matrix& matrix, size_t size, uint32_t pass = 0) {
+    if (pass == 0) {
+      matrix.mutable_from_indices()->Resize(size, 0U);
+      matrix.mutable_to_indices()->Resize(size, 0U);
+      matrix.mutable_distances()->Resize(size, 0U);
+      matrix.mutable_times()->Resize(size, 0U);
+      matrix.mutable_second_pass()->Resize(size, false);
+      // repeated strings don't support Resize(), likely this is not ideal in terms of performance
+      for (size_t i = 0; i < size; i++) {
```
nilsnolde, line 161: reserve here before adding

test/test.cc
```diff
@@ -380,7 +380,8 @@ boost::property_tree::ptree make_config(const std::string& path_prefix,
         "service": {
           "proxy": "ipc://%%/thor"
         },
-        "source_to_target_algorithm": "select_optimal"
+        "source_to_target_algorithm": "select_optimal",
+        "costmatrix_allow_second_pass": true
```
nilsnolde, line 384: remove this and only turn it on in the test

valhalla/thor/costmatrix.h
```diff
@@ -110,6 +111,15 @@ class CostMatrix : public MatrixAlgorithm {
    */
   void Clear() override;
 
+  /**
+   * Get the algorithm's name
+   * @return the name of the algorithm
+   */
+  inline const std::string& name() override {
+    const static std::string& name = MatrixAlgoToString(Matrix::CostMatrix);
```
nilsnolde, line 119: return the function result

kevinkreiser:
this looks broadly good to me other than the tiny nitpicks we just discussed. let me know when those are patched and ill put a ship. would also be nice if @dnesbitt61 had a sec to look

nilsnolde:
@kevinkreiser I addressed the comments

kevinkreiser (approved):

PR #4541, opened 2024-02-01
bump cxxopts to v3 (major version upgrade)

nilsnolde:
was just trying a vcpkg build with dependencies from vcpkg instead of our own vendored ones and realized we could also bump our vendored cxxopts version to v3.x. it's a major version upgrade, but not much changed, mainly the namespace (and name) of their exception.


src/argparse_utils.h
```diff
@@ -1,3 +1,5 @@
+#include <iostream>
```
nilsnolde, line 1: must've been removed from cxxopts.h so now it bit us

kevinkreiser (approved):

Issue #4540, opened 2024-02-01
Multimodal and wheelchair: GTFS wheelchair_boarding not considered for in-station transfers


jthiard:
Hi,
I recently jumped into the multimodal routing algorithm, thanks to the new GTFS loader. Which works really great :+1: 

I'm mainly interested in wheelchair routing - currently working on an highly specialized custom costing, feeded with proprietary data). When playing around with multimodal routes, I realized that:

- `wheelchair_boarding` in GTFS stops is taken into account when transitioning from pedestrian to transit or the opposite: if the pedestrian costing `type` is `wheelchair`, the user won't be able to start or end a route from a stop without `wheelchair_boarding`
- This is not true for in-station transfers (changing trip at the same stop): the multimodal algorithm will happily ask the user to change bus in a stop without `wheelchair_boarding`. But in real life this would mean changing vehicle, so unboarding from the first one and boarding into the second one. Which sounds wrong if the stop has not `wheelchair_boarding` capable.

I realize multimodal + wheelchair is kinda dragon zone. So I first wanted to share my findings and the fix I did on my side to see if it sounds good to you. If so I'll be happy to try to write a test case and contribute a fix.

**What I found:**

GTFS `wheelchair_boarding` is set on transit stop nodes during transit ingestion:

 https://github.com/valhalla/valhalla/blob/c9c241873bab2155c814d5d04ab740f492427b06/src/mjolnir/ingest_transit.cc#L278-L279

This info will be translated into access mask on the hraph node during transit conversion:
https://github.com/valhalla/valhalla/blob/c9c241873bab2155c814d5d04ab740f492427b06/src/mjolnir/convert_transit.cc#L1061-L1063
and then
https://github.com/valhalla/valhalla/blob/c9c241873bab2155c814d5d04ab740f492427b06/src/mjolnir/convert_transit.cc#L593-L596

When connecting transit stops to the graph, new directed edges connected to this stop will inherit the access mask from the stop node:
https://github.com/valhalla/valhalla/blob/c9c241873bab2155c814d5d04ab740f492427b06/src/mjolnir/transitbuilder.cc#L233-L258

And finally, for some reason, access mask on the stop node in then reset (probably because this stop can be traversed on transit):
https://github.com/valhalla/valhalla/blob/c9c241873bab2155c814d5d04ab740f492427b06/src/mjolnir/transitbuilder.cc#L299-L300

[...]

PR #4542, opened 2024-02-01
Consider transit stop access for in station transfers

jthiard:
:warning: No test-case written so far, and there might be some side effects.

Try to fix in-station transfers in stops with GTFS `wheelchair_boarding` to false, when pedestrian costing use the `wheelchair` transport type.

# Issue

#4540

## Tasklist

 - [ ] Add tests

### DIFF
--- a/src/mjolnir/transitbuilder.cc
+++ b/src/mjolnir/transitbuilder.cc
@@ -296,8 +296,6 @@ void ConnectToGraph(GraphTileBuilder& tilebuilder_local,
     // Add the node and directed edges
     nb.set_edge_index(edge_index);
 
-    // reset the access to defaults.
-    nb.set_access((kPedestrianAccess | kWheelchairAccess | kBicycleAccess));
     nb.set_edge_count(tilebuilder_transit.directededges().size() - edge_index);
     tilebuilder_transit.nodes().emplace_back(std::move(nb));
   }
--- a/src/thor/multimodal.cc
+++ b/src/thor/multimodal.cc
@@ -390,9 +390,15 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
         } else {
           if (pred.tripid() > 0) {
             // tripId > 0 means the prior edge was a transit edge and this
-            // is an "in-station" transfer. Add a small transfer time and
-            // call GetNextDeparture again if we cannot make the current
-            // departure.
+            // is an "in-station" transfer.
+            // This mean changing vehicle (unboard, wait, board again).
+            // So first check if the pedestrian costing allow traversing
+            // this node.
+            if (!pc->Allowed(nodeinfo)) {
+              continue;
+            }
+            // Then add a small transfer time and call GetNextDeparture again
+            // if we cannot make the current departure.
             // TODO - let's get the transfers.txt implemented!
             if (offset_time.day_seconds() + 30 > departure->departure_time()) {
               departure =


src/mjolnir/transitbuilder.cc
```diff
@@ -296,8 +296,6 @@ void ConnectToGraph(GraphTileBuilder& tilebuilder_local,
     // Add the node and directed edges
     nb.set_edge_index(edge_index);
 
-    // reset the access to defaults.
-    nb.set_access((kPedestrianAccess | kWheelchairAccess | kBicycleAccess));
```
jthiard, line 300: Keep the access mask inherited from the GTFS.

This should not break the multimodal costing which doesn't evaluate the node access, nor normal pedestrian routes since those nodes are connections between the transit graph and the pedestrian graph.

src/thor/multimodal.cc
```diff
@@ -390,9 +390,15 @@ bool MultiModalPathAlgorithm::ExpandForward(GraphReader& graphreader,
...
-            // is an "in-station" transfer. Add a small transfer time and
-            // call GetNextDeparture again if we cannot make the current
-            // departure.
+            // is an "in-station" transfer.
+            // This mean changing vehicle (unboard, wait, board again).
+            // So first check if the pedestrian costing allow traversing
+            // this node.
+            if (!pc->Allowed(nodeinfo)) {
+              continue;
+            }
```
jthiard, line 399: If `wheelchair_boarding` was set to `false` and access mask keeped on the stop node, a normal pedestrian costing won't allow wheelchair users to "walk" (unboard and board)  through this node.

jthiard:
What I understand from GTFS wheelchair handling in Valhalla:

- GTFS trips `wheelchair_accessible` will be used to qualify transit edge, evaluated by transit costing. This works well to my knowledge
- GTFS stops `wheelchair_boarding` will be used to qualify "transit connection" edges, ie edges created to connect the transit graph with the road network. Those edges will be evaluated by the pedestrian costing. This work well when starting or ending a route, or for transfers when changing stop through the road network).

During a in-station transfer (changing vehicle at the same stop), I propose to call the pedestrian costing on the transit stop node to evaluate accessibility and the ability to change vehicle.

nilsnolde:
Sounds good on first glance, thanks! Adding a test for this will be a bit of a challenge, transit costing is not really a first-class citizen in our testing framework, but it's of course possible. Let us know if you need some pointers.

jthiard:
> I'm curious: how did you even find out our wheelchair support?

Oh that's simple, diving into the code :smile: 
We work on fine pedestrian routing for "standard" pedestrians and for various kinds of disability profiles, at small scales (basically a city or an urban area). And Valhalla is an amazing code base for this.

There is no globally available data at this time so we have our own (highly unstable) data model and our own (highly volatiles) costings. Hopefully there will be standardized data one day, at least in Europe with ongoing work on NeTEx profiles for accessibility.

So TBH we don't rely on the standard pedestrian costing, this is why in my fix the multimodal algorithm don't check the node access mask, and instead rely on whatever costing has been given as the pedestrian costing to check access.

>  Adding a test for this will be a bit of a challenge, transit costing is not really a first-class citizen in our testing framework, but it's of course possible.

I will give a try and let you know, thanks.

nilsnolde:
I have little idea (so far) what is really needed for proper wheelchair routing. I've been working in Heidelberg when they did their PoC at GIScience; I've not been involved but I remember they did a crazy effort to update OSM with sidewalks and curb info. AFAIK the wider OSM community seems to shift towards modeling sidewalks as separate entities (at least in Europe), which would help a lot in making wheelchair routing more proper. But there's also lots of people who very much don't like that, bcs it makes OSM maintenance so much harder, and I can kinda get the argument. If it's not _somehow_ in OSM there's very little chance we'd ever be able to really support good wheelchair routing. Unless there's an ODbL-compliant derivative of OSM dedicated to that which is hard to imagine ever existing..

jthiard:
I'm working again on this because I realized multimodal isochrones have the same issue with `wheelchair_boarding`.

Isochrones use a `thor:Isochrone` algo inheriting straight from Djiskstras, so my changes to `thor:MultimodalPathAlgorithm` doesn't help here.

I really need to work on a way to test this with a minimal multimodal tileset before moving forward -> still draft for now :sweat_smile:

PR #4543, opened 2024-02-02
Enable pbf output for isochrone endpoint

chrstnbwnkl:
# Issue

Adds pbf output option for the isochrone endpoint. 

The way the geometries are represented in the proto message right now is pretty simple, but analogous to how the outputs are serialized to GeoJSON: each contour is treated as a ring and its left to the end user to figure out  the exact relationships between them. In most cases, this is easy because there is only one exterior ring and possibly holes. But I can identify at least two scenarios where the response contains a feature (or `contour` message in pbf) with more than one exterior ring:

  1. multi-locations, if their expansions do not intersect
  2. when the expansion includes a ferry edge
  3. possibly when using transit (haven't tested this yet but assume it's the same as with ferry, because you can only enter or leave at a specific stop, not along the way)

In these case, multiple exterior rings are produced, as well as inner rings that would have to be associated to the right outer one.

In any case, maybe there is a better way of representing the output geometries that includes taking some of the heavy lifting off the users by associating inner and outer rings correctly and construct a multipolygon feature for each contour. (EDIT: see proposal in #4575 )
  

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/thor/isochrone_action.cc
```diff
@@ -46,11 +46,11 @@ std::string thor_worker_t::isochrones(Api& request) {
   // we have parallel vectors of contour properties and the actual geojson features
   // this method sorts the contour specifications by metric (time or distance) and then by value
   // with the largest values coming first. eg (60min, 30min, 10min, 40km, 10km)
-  auto isolines =
-      grid->GenerateContours(contours, options.polygons(), options.denoise(), options.generalize());
+  auto contours =
```
chrstnbwnkl, line 49: renaming for consistency

src/tyr/isochrone_serializer.cc
```diff
@@ -12,16 +12,16 @@ using namespace valhalla::baldr::json;
...
+using namespace valhalla;
+using namespace tyr;
 
-std::string serializeIsochrones(const Api& request,
-                                std::vector<midgard::GriddedData<2>::contour_interval_t>& intervals,
-                                midgard::GriddedData<2>::contours_t& contours,
-                                bool polygons,
-                                bool show_locations) {
+std::string
+serializeIsochroneJson(Api& request,
```
chrstnbwnkl, line 20: separate json and pbf construction

src/tyr/isochrone_serializer.cc
```diff
@@ -142,5 +142,68 @@ std::string serializeIsochrones(const Api& request,
...
+    con.set_metric_value(std::get<1>(interval));
+
+    // for each feature
+    for (const auto& feature : contour) {
+      // for each ring/contour
+      for (const auto& ring : feature) {
+        // construct a geometry
+        Isochrone_Geometry geom;
+        for (const auto& pair : ring) {
+          geom.add_coords(round(pair.lng() * 1e6));
```
chrstnbwnkl, line 171: coords are just scaled for now, not delta encoded

src/thor/isochrone_action.cc
```diff
@@ -17,13 +17,13 @@ std::string thor_worker_t::isochrones(Api& request) {
   auto costing = parse_costing(request);
 
   // name of the metric (time/distance, value, color)
-  std::vector<GriddedData<2>::contour_interval_t> contours;
+  std::vector<GriddedData<2>::contour_interval_t> intervals;
```
nilsnolde, line 20: yeah thanks for that!

src/tyr/isochrone_serializer.cc
```diff
@@ -142,5 +142,66 @@ std::string serializeIsochrones(const Api& request,
...
+  Isochrone& isochrone = *request.mutable_isochrone();
+
+  // construct contours
+  for (size_t isoline_index = 0; isoline_index < contours.size(); ++isoline_index) {
+    const auto& contour = contours[isoline_index];
+    const auto& interval = intervals[isoline_index];
+
+    valhalla::Isochrone::Contour con;
+    con.set_metric(std::get<2>(interval) == "time" ? Isochrone_metric_type_time
+                                                   : Isochrone_metric_type_distance);
```
nilsnolde, line 158: nit: short form is somewhat nicer IMO
```suggestion
    con.set_metric(std::get<2>(interval) == "time" ? Isochrone::time
                                                   : Isochrone::distance);
```

src/tyr/isochrone_serializer.cc
```diff
@@ -142,5 +142,66 @@ std::string serializeIsochrones(const Api& request,
...
+    for (const auto& feature : contour) {
+      // for each ring/contour
+      for (const auto& ring : feature) {
+        // construct a geometry
+        Isochrone_Geometry geom;
+        for (const auto& pair : ring) {
+          geom.add_coords(round(pair.lng() * 1e6));
+          geom.add_coords(round(pair.lat() * 1e6));
+        }
+        con.add_geometries()->CopyFrom(geom);
```
nilsnolde, line 172: not sure if `CopyFrom` is best here.. sounds like it's copying;) there's also a `SwapSmth` function(s) or change the order of things by creating a contour at the top with `auto* con = isochrone.mutable_contours->Add()`  and then a `auto* geom = con.mutable_geometries().Add()` where you can do `geom->add_coords()`.

src/valhalla_run_isochrone.cc
```diff
@@ -177,7 +177,7 @@ int main(int argc, char* argv[]) {
 
   // Compute the isotile
   auto t1 = std::chrono::high_resolution_clock::now();
-  Isochrone isochrone;
+  valhalla::thor::Isochrone isochrone;
```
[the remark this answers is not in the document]
nilsnolde: hm, yeah, it'd be much nicer if we had a `valhalla::pbf` namespace, this is always a problem

nilsnolde (changes requested):
Nice, thanks for this! Just a few nits from my side and one minor problem in the pbf serializer.

src/tyr/isochrone_serializer.cc
```diff
@@ -153,27 +153,23 @@ std::string serializeIsochronePbf(Api& request,
...
     for (const auto& feature : contour) {
       // for each ring/contour
       for (const auto& ring : feature) {
         // construct a geometry
-        Isochrone_Geometry geom;
+        auto* geom = con->mutable_geometries()->Add();
         for (const auto& pair : ring) {
-          geom.add_coords(round(pair.lng() * 1e6));
-          geom.add_coords(round(pair.lat() * 1e6));
+          geom->add_coords(round(pair.lng() * 1e6));
```
nilsnolde, line 168: Oh right, I didn’t see that initially. I guess it’s fine to mix lat/lon in the same array. Another concept would be to do lats and lons in separate arrays, but I don’t have a strong opinion on it. The latter is possibly a tiny bit more trivial to parse out for a client.

nilsnolde (dismissed):

src/tyr/isochrone_serializer.cc
chrstnbwnkl: I mean can it get any more trivial :smile: we can provide examples somewhere in the docs if necessary
```cpp
for (int i = 0; i < coords.size(); i += 2)
```

nilsnolde (dismissed):
:ship:

src/tyr/isochrone_serializer.cc
```diff
@@ -142,5 +142,62 @@ std::string serializeIsochrones(const Api& request,
...
+namespace tyr {
+
+std::string serializeIsochrones(Api& request,
+                                std::vector<midgard::GriddedData<2>::contour_interval_t>& intervals,
+                                midgard::GriddedData<2>::contours_t& contours,
+                                bool polygons,
+                                bool show_locations) {
+
+  // if the user requested pbf output
+  if (request.options().format() == Options_Format_pbf) {
```
chrstnbwnkl, line 189: not sure how I missed this one... if #4575 gets discarded in favor of this one, this needs to be removed

nilsnolde: oops, didn't see that either.. wouldn't break anything, but yeah it's redundant

nilsnolde: I'd say we should merge this first then merge into #4575 to reduce the changes

src/tyr/isochrone_serializer.cc
```diff
@@ -142,5 +142,62 @@ std::string serializeIsochrones(const Api& request,
...
+    return serializeIsochronePbf(request, intervals, contours);
+  }
+
+  switch (request.options().format()) {
+    case Options_Format_pbf:
+      return serializeIsochronePbf(request, intervals, contours);
+    case Options_Format_json:
+      return serializeIsochroneJson(request, intervals, contours, polygons, show_locations);
+    default:
+      throw;
```
nilsnolde, line 199: what is the bare `throw` doing here actually? it seems off, there's nothing catching any exception before this

proto/isochrone.proto
```diff
@@ -0,0 +1,24 @@
...
+
+message Isochrone {
+
+  enum metric_type {
+    time = 0;
+    distance = 1;
+  }
+
+  message Geometry {
+    repeated sint32 coords = 1;
```
chrstnbwnkl, line 14: put packed = true

chrstnbwnkl:
Closing in favor of #4575

PR #4544, opened 2024-02-02
Make vendored libraries optional

### LINKED ISSUE #4529: remove third_party libs as much as possible
while it's very convenient for us, it's not compatible with general package managing rules, to have public interface depending on vendored libraries. I guess all libs only used in tests are fine, no one wants to package tests. this could also be (partly, see below) solved by removing those headers from our public interface which would be a great thing to do anyways.

I'll still need to confirm, but I think even compile-time header-only libs are at least frowned upon. because of security implications and possibly licensing problems.

nilsnolde:
based on #4541 

fixes #4529 

Let users decide whether to use our vendored libraries or their own for all major ones which we need for compile-/ or run-time (e.g. not `just_gtfs`, not found in any package manager). 

I tried a full/default vcpkg build turning all vendored libraries off, worked smoothly. To be fair, quite a few libraries wouldn't be found with other package managers, e.g. `date` or `dirent.h`. But it doesn't really matter, bcs we fallback to our submodules.

What this PR does:
- introduces a new `PREFER_SYSTEM_DEPS` which, if turned `ON` (default `OFF`), will use `find_package()` to resolve `date`, `rapidjson`, `cxxopts`, `robin-hood-hashing`, `pybind11`, `dirent.h` (Win only). If turned `ON` and CMake doesn't find the library, we fall back to our vendored version of those libs and emit a `WARNING`. This way nothing changes for anyone who doesn't want to use this feature
- remove `lz4` submodule and depend on external package via `pkg-config`, it's available everywhere and stable
- add `PREFER_SYSTEM_DEPS=ON` for the Windows CI build which already uses `vcpkg`
- add a section to `building.md` for `vcpkg`


.github/workflows/win.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ${{ env.VCPKG_ROOT }}
-          key: vcpkg=${{ env.VCPKG_VERSION }}-msvc=${{ env.MSVC_VERSION }}-json=${{ hashFiles('vcpkg.json') }}-v2
+          key: 'vcpkg=${{ env.VCPKG_VERSION }}-msvc=${{ env.MSVC_VERSION }}-json=${{ hashFiles('vcpkg.json') }}-v2'
```
nilsnolde, line 67: we're still not having any cache hits..

CMakeLists.txt
```diff
@@ -93,6 +94,10 @@ endif()
 include(ValhallaSanitizerOptions)
 include(ValhallaSourceGroups)
 
+# prefer CONFIG mode over MODULE mode, which versions configuration on the package, not CMake
+# NOTE: this is only supported for cmake >= 3.15, but shouldn't be a problem in real life
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
```
nilsnolde, line 99: let's turn this on for everything, it should _always_ be preferred if available as those config modules are shipped with the libraries, instead of cmake

CMakeLists.txt
```diff
@@ -118,16 +123,57 @@ add_definitions(-DBOOST_NO_CXX11_SCOPED_ENUMS)
 add_definitions(-DBOOST_ALLOW_DEPRECATED_HEADERS)
 add_definitions(-DBOOST_BIND_GLOBAL_PLACEHOLDERS)
 
+# resolve vendored libraries
+set(date_include_dir ${VALHALLA_SOURCE_DIR}/third_party/date/include)
```
nilsnolde, line 127: this is the main changes, defaulting to vendored paths but trying to change them to system paths if opted in with `PREFER_SYSTEM_DEPS`

docs/docs/building.md
```diff
@@ -52,34 +72,9 @@ sudo make -C build install
 
 ### Building from Source - macOS
 
-#### Configuring Rosetta for ARM64 MacBook
-
-Check your architecture typing `arch` in the terminal. In case the result is `arm64` set up Rosetta terminal to emulate x86_64 behavior. Otherwise, skip this step.
```
nilsnolde, line 57: rosetta should be a thing of the past I think

src/baldr/CMakeLists.txt
```diff
@@ -38,14 +38,10 @@ set(includes
...
      )
 
-# treat date library as system
 set(system_includes 
-  ${VALHALLA_SOURCE_DIR}/third_party/date/include
-  $<$<BOOL:${WIN32}>:${VALHALLA_SOURCE_DIR}/third_party/dirent/include>
-  ${VALHALLA_SOURCE_DIR}/third_party/rapidjson/include)
-if(APPLE)
-  list(APPEND system_includes ${VALHALLA_SOURCE_DIR}/third_party/date/include/date)
-endif()
```
nilsnolde, line 48: I removed this in all CMakeLists.txt, `${VALHALLA_SOURCE_DIR}/third_party/date/include/date` is not even a valid path

.github/workflows/win.yml
```diff
@@ -37,7 +37,7 @@ defaults:
 env:
   BUILD_TYPE: Release
   MSVC_VERSION: '2022'
-  VCPKG_VERSION: '2024.01.12'
+  VCPKG_VERSION: '8040303'
```
nilsnolde, line 40: newest commit on vcpkg, the "old" release failed to resolve libspatialite with pkg-config for some reason (worked until this PR and I didn't change anything regarding pkg-config or spatialite..)

kevinkreiser (approved):

PR #4547, opened 2024-02-03
attempt to fix infinite loop in isochrones generation

### LINKED ISSUE #4546: In isochrone: Infinite loop in bresenham_line, caused by Nan, caused by divide by zero, due to zero length edge
# Summary

While generating isochrones for 160,000 locations, we found 2 locations that would result in no response from the isochrone API. Repeat attempts would gradually lock each handler thread until the server became entirely unresponsive. 

Debugging showed the handler threads were getting stuck in an infinite loop in `bresenham_line()`.

Sorry, may be a duplicate of #3823 but wanted to post a bit of detail so opening a new bug. Can move this there if preferred.

# Steps to reproduce

So far I've reproduced this in these versions
* `master` at `0cc7d6ec7` compiled on macOS M1
* 3.4.0 compiled for macOS M1
* 3.4.0 using the project Docker images

To make the steps easier to read, I'm going refer to the following requests from the steps below

isochrone-ok
```
curl --request GET --url 'http://localhost:8002/isochrone?json=%7B%22locations%22%3A%5B%7B%22lat%22%3A%2244.89%22%2C%22lon%22%3A%22-0.568%22%7D%5D%2C%22costing%22%3A%22auto%22%2C%22contours%22%3A%5B%7B%22time%22%3A10.0%7D%5D%7D'
```
isochrone-bad
```
curl --request GET --url 'http://localhost:8002/isochrone?json=%7B%22locations%22%3A%5B%7B%22lat%22%3A%2244.836048%22%2C%22lon%22%3A%22-0.565914%22%7D%5D%2C%22costing%22%3A%22auto%22%2C%22contours%22%3A%5B%7B%22time%22%3A10.0%7D%5D%7D'
```

1. Generate tile data (I think we might be using a slightly customised version of the car profile Lua script, I will need to check and confirm)
2. Start the service with 1 thread: `valhalla_service <path to config>.json 1`
3. Execute "isochrone-ok" request and confirm there is a response
4. Execute "isochrone-bad" request, there is no response, the client times out, the server process is using 100% of 1 CPU core.
5. Execute "isochrone-ok" request, there is no response, the client times out.

# Findings from debugging

Pausing the debugger while it's locked up showed up the problem loop. It is stuck in an endless loop in `bresenham_line()` in tiles.cc, because `x0` and `y0` are both `NaN`. 

Screenshot 2024-01-23 at 18 05 28

Tracing this back it's because of a divide by zero in `OriginEdgeShape()`, here:

Screenshot 2024-01-26 at 21 12 40

So it seems there is an edge in the tile data that has a zero length. On line 22 the Distance call returns 0.0 On line 25 that results in a divide by zero. It evaluates to NaN. This then carries forward polluting code that follows, which continues executing until it gets stuck in the loop in `bresenham_line()`.

Here's a diagram of the callstack with my notes:

image

# Open questions

This is as far as I've got. The two questions I have at this point are:

1. Why is there a zero length edge? 
  * Could it be there is something wrong with our tile data? 
  * Could this have been caused by a version mismatch between the server and the data processing scripts?
2. Is there a good place in the code that we could detect the 0 or the NaN?
  * Should `bresenham_line()` be responsible for checking for `NaN`? or `OriginEdgeShape()` for 0? Or could it be detected at boot or data load time?

kevinkreiser:
fixes #4546 

we may still want to protect bresenhams separately but perhaps this is a start


kevinkreiser:
@munckymagik did you have time to give this code a try?

src/thor/isochrone.cc
```diff
@@ -17,10 +17,22 @@ constexpr float METRIC_PADDING = 10.f;
 template <typename PrecisionT>
 std::vector<GeoPoint<PrecisionT>> OriginEdgeShape(const std::vector<GeoPoint<PrecisionT>>& pts,
                                                   double distance_along) {
+  // just the endpoint really
+  if (distance_along == 0)
```
kevinkreiser, line 21: if you are snapped to the end of the edge we just return a segment of zero length at the end

src/thor/isochrone.cc
```diff
@@ -17,10 +17,22 @@ constexpr float METRIC_PADDING = 10.f;
...
+  // consume shape until we reach the desired distance
   double suffix_len = 0;
   for (auto from = std::next(pts.rbegin()), to = pts.rbegin(); from != pts.rend(); ++from, ++to) {
+    // add whatever this segment of shape contributes to the overall distance
     PrecisionT len = from->Distance(*to);
     suffix_len += len;
+
+    // skip 0 length segments
+    if (len == 0)
+      continue;
```
kevinkreiser, line 33: this is not unneeded. if `len` is 0 then it cant change `suffix_len` which means it cant change the outcome of that `if` below and get in there and do a divide by 0.

.github/workflows/lint.yml
```diff
@@ -1,4 +1,4 @@
-name: syntax
+name: lint
```
kevinkreiser, line 1: i prefer to call it like it is

nilsnolde (approved):
LGTM, thanks for the additional comments!

PR #4554, opened 2024-02-04
move osx ci to github actions

nilsnolde:
fixes #4550 (move mac-arm64 to GHA)
fixes #4558 (circleci had a broken path for ccache caching)
fixes #4557 (remove running benchmarks in ci)

PR #4560, opened 2024-02-05
remove bench

### LINKED ISSUE #4557: remove "benchmarks" from CI?
they're pretty pointless IMO. we can keep building them if some day we'd like to actually do smth with them, but I guess CI machines are not the right environment for that in any case, way too spotty in terms of actual resources.. 

it'd shave off a good 6 mins from the circleci builds if we stopped exercising them.

nilsnolde:
fixes #4557 

based on #4554 

removes the `./bench` and everything related (CMake flags, CI jobs etc).


kevinkreiser (approved):
feels great to shed this

kevinkreiser (approved):

nilsnolde:
No idea why the win ci is flaky..

Issue #4562, opened 2024-02-05
configurable map area and tile sizes


akesmarki:
Feature request:
 
It would be great if there would be a possibility in Valhalla to specify a custom tile layout with custom whole map area and tile sizes. 
For example that would enable us to create maps with smaller/bigger tiles, or create only map slices.

nilsnolde:
I have reservations regarding "map area", see https://github.com/valhalla/valhalla/pull/4563#issuecomment-1927444883. best to open the issue and get feedback on feature requests _before_  working on smth:)

akesmarki:
If I assume the whole map area is always 360x180 degrees, then the number of columns and number of rows in the tile grid are also redundant. 
If we go in that way, only the tilesizes array is enough in the config file.
What do you think ?

nilsnolde:
Jep sorry, should've mentioned that. Exactly, only the tilesizes are good:)

PR #4563, opened 2024-02-05
Configurable tiling layout

akesmarki:
# Issue

[#4562]

## Tasklist

 - [ ] Add tests

## Requirements / Relations

We build on the recently introduced config singleton. 
As described in the linked issue, this PR is about the configurable tiling scheme/layout.
The new keys in valhalla.json are optional, and the fallback when they are not specified is the Valhalla/OSM default layout. 

Attila Kesmarki - attila.kesmarki@nng.com on behalf of NNG LLC.


nilsnolde:
Thanks, we always wanted to have configurable tile sizes. However, why the coordinates? Personally I wouldn't like to have that in the code, bcs it's redundant: either you clip the PBF before, or you can use `valhalla_build_extract` to produce extracts from a pre-built tileset. I can't imagine any other requirement that would warrant adding this, but maybe I'm missing smth?

akesmarki:
When I cut the coded numbers from the source, I didn't want to just go halfway. But you are right, the same result can be achieved with other tools.
I will remove the minpt key.

dnesbitt61 (commented):
I have seen issues with some tile sizes if I recall. I seem to remember tile sizes have to be evenly divisible into 1.
0.125, 0.25, 0.5 work, but
0.3, 0.6, 0.75 I think cause issues.
Perhaps this should be tested with some of these tile size variations to make sure they work. If there are limitations they should be spelled out somewhere.

dnesbitt61:
Also, if tile bounds are provided what happens in the OSM pbf extends outside the bounds? Is this handled properly or does this cause a crash in valhalla_build_tiles?

dnesbitt61:
It seems better to always use a full earth tiling system (rather than allowing min, max lat,lon for the tiles) to create consistent tile Ids.

danthe42:
I agree, it's a good idea, and the safest option.

Maybe extending the GraphTileHeader with a new field: total number of columns in the tile grid, on that given level would be enough. In that case the tilesize can be calculated easily as 360/total_columns so we don't need to use float type.

src/baldr/tilehierarchy.cc
```diff
@@ -1,44 +1,100 @@
...
+  };
+
+  try {
+    // Many tests don't specify any configuration, so catch the "Config singleton was not initialized
+    // before usage" exception here and fallback to default OSM tile grid.
+    const boost::optional<const boost::property_tree::ptree&> tilesizesptree =
+        config().get_child_optional("baldr.tilesizes");
+
+    if (tilesizesptree) {
+      BOOST_FOREACH (auto& v, *tilesizesptree) { tilesizesvector.push_back(v.second.get<float>("")); }
```
kevinkreiser, line 31: modern c++ has foreach for a long time, i know ptree is a pain but i think it should work with:

```suggestion
    for(auto& v : *tilesizesptree) { 
      tilesizesvector.push_back(v.second.get<float>(""));
    }
```

src/baldr/tilehierarchy.cc
```diff
@@ -1,44 +1,100 @@
...
+                                                                                 tilesizesvector[3])),
+                                                 static_cast<int32_t>(std::round(wholemapbb.Height() /
+                                                                                 tilesizesvector[3])),
+                                                 static_cast<unsigned short>(kBinsDim)}},
+  };
+}
+
+const std::vector<TileLevel>& TileHierarchy::levels() {
+  if (levels_.size() == 0) {
+    getLevelsfromConfig();
```
kevinkreiser, line 87: getting these values from config doesnt seem like the right idea to me. as @dnesbitt61 it invites the possibility of having a config that doesnt match the actual tiles.

if we want to make this configurable, we have to:

1. validate the config doesnt break the rules (evenly divisble adjacent levels)
2. the config is only used to build tiles (ie it is a mjolnir config)
3. the service and library discover the tilesize from the tiles themselves via graphtileheader

kevinkreiser: that last one is a deal breaker in that it would require a major version change. we would have to add something to the graphtileheader and add code to graphreader et al to read that. we can do that as a new feature in valhalla v3 but if we make tiles that arent using the default sizes and old valhalla v3 code wants to read those, it will break because it wont know anything about the configurable sizes and having to get that info from graphtileheader. sadly to actually do this we need to do it in a new major version

akesmarki: Yes, old valhalla v3 code would not be able to use the new tiles. Can't we handle this feature as a new, experimental one and warn the users about this ? 

@kevinkreiser , could you advice how to progress with this ?
-  Should this PR be cancelled now, and later the design/implementation of the final graphtile-embedded tilesizes can be discussed, when preparing to switch to v4.
 - Or, would it worth to include the configurable tilesize feature in some form (based on this PR) in the current, v3 series as a temporary solution, so, at least we could start experimenting it ?

nilsnolde: Hm, but this modification doesn't solve the initial problem of breaking a consuming application if the tiles were built non-standard, right? Then what's the point if there can't be a graph reader being able to read them?

> Thanks, we always wanted to have configurable tile sizes.

I was very confused here haha I thought about hierarchy limits, not tile dimensions.

akesmarki: That's true, it does not solve the compatibility issue, but, at least it enables this experimental feature in v3. I don't know about the v4 release plans, but if it's in the far future, would it not be better to give some possibility to experiment with this feature ?

akesmarki: Well, there would be 2 use cases:

- I believe all future releases from master would be able to convert OSM maps with a custom layout if this is merged.  Am I wrong ?
- Also, we could produce graph tiles from other sources with custom layout.

The feature is on our forked branch, we can use it, so I won't feel bad if you refuse to accept this contribution.

nilsnolde: Jep then you could _build_ tiles that way, but you won’t be able to _read_ the same tiles right? The graph reader will still expect the default resolution. Or am I missing smth?

akesmarki: Nope, I could produce graph tiles using a custom tile grid, and use the valhalla-app viewer/tester to calculate routes on it when I specified the same custom layout in the valhalla.json.
So it seems that the graph reader and the whole valhalla_service used this custom layout correctly.

nilsnolde: Huh interesting.. then I’d be clearly missing smth😄

akesmarki:
Any insights about this Pull Request by the maintainers ? Should I change it somehow, cancel it, or should it be put "on hold"  ?

kevinkreiser:
i mean i am truly shocked that we've been so diligent that all you needed to change was the tile hierarchy to get it to work! in terms of merging it into master i guess i wouldnt be against it if the default is the current configuration. this way no one gets screwed over unless they decide to do it to themselves haha

to prove that this actually works for literally everything id like to see a commit where we change the default (in all the test configs) to be some other valid configuration. if we can do that any everything works then we can revert just that part and review and merge it. specifically i mean changing mainly this to some other valid configuration: https://github.com/valhalla/valhalla/blob/master/test/test.cc#L123

like @nilsnolde im just shocked that its possible we havent made any other assumptions around this particular design choice and that its all only accessed from this one place. a rare moment where we didn't make a mistake and kept our selves disciplined haha

src/baldr/tilehierarchy.cc
```diff
@@ -1,44 +1,104 @@
...
 namespace valhalla {
 namespace baldr {
 
-const std::vector<TileLevel>& TileHierarchy::levels() {
-  // Static tile levels
-  static const std::vector<TileLevel> levels_ = {
+static std::vector<TileLevel> levels_;
+static std::vector<TileLevel> transit_levels_;
+
+void getLevelsfromConfig() {
```
nilsnolde, line 14: what I'd like to change is the coding style and would ask to make it more valhalla like:
- either camel case with the first letter capitalized or snake case
- no `bSizesInConfigOk`, `tilerowsvector` etc indicating the type as part of the variable name
- variable names always snake case (we'd like that for functions too actually, but somehow java-ish google style crept in at some point)

PR #4567, opened 2024-02-06
fix config generator with thor.costmatrix_allow_second_pass

### LINKED ISSUE #4565: costmatrix second pass configuration parameter name inconsistency in valhalla_build_config
Hello, we are testing new second pass feature in costmatrix [(4536)](https://github.com/valhalla/valhalla/pull/4536).

In src is taken "thor.costmatrix_allow_second_pass":
[worker_cc_line](https://github.com/valhalla/valhalla/blob/e406c10bbcc01c57a991ec8f3b9beace0113fbe4/src/thor/worker.cc#L104)

In python valhalla_build_config is parameter enabling this named "costmatrix_second_pass": [valhalla_build_config_line](https://github.com/valhalla/valhalla/blob/e406c10bbcc01c57a991ec8f3b9beace0113fbe4/scripts/valhalla_build_config#L122)

So IMHO valhalla_build_script is wrong and should be change current:
```json
costmatrix_second_pass': False,
```
should be change to:
```json
costmatrix_allow_second_pass': False,
```

nilsnolde:
fixes #4565 

in the test I actually put the right config name, that's why it worked :smile:


gknisely (approved):

nilsnolde:
thx, cowboying it

PR #4573, opened 2024-02-07
fix win vcpkg caching

nilsnolde:
trying around until smth works. wish there's be an option to turn off certain CI jobs per push..

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -67,16 +67,15 @@ jobs:
 
       - name: Build Tests
         run: make -C build -j$(sysctl -n hw.logicalcpu) tests
-          
-      - name: Run Tests
-        run: make -C build -j$(sysctl -n hw.logicalcpu) check
 
       - name: Save ccache
-        id: cache-ccache-save
         uses: actions/cache/save@v4
         with:
           path: ~/Library/Caches/ccache
           key: ${{ steps.cache-ccache-restore.outputs.cache-primary-key }}
+          
+      - name: Run Tests
+        run: make -C build -j$(sysctl -n hw.logicalcpu) check
       
       - name: Setup tmate session
         uses: mxschmitt/action-tmate@v3
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -29,7 +29,7 @@ on:
         default: false
 
 jobs:
-  build_osx_tests:
+  build_osx:
     runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -29,7 +29,7 @@ on:
         default: false
 
 jobs:
-  build:
+  build_osx_tests:
     runs-on: macos-14
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -38,7 +38,7 @@ env:
   VCPKG_DISABLE_COMPILER_TRACKING: ON
 
 jobs:
-  build:
+  build_win:
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -109,7 +109,7 @@ jobs:
             -DENABLE_CCACHE=OFF \
             -DENABLE_SERVICES=OFF \
             -DENABLE_BENCHMARKS=OFF \
-            -DPREFER_EXTERNAL=ON
+            -DPREFER_EXTERNAL_DEPS=ON
 
       - if: ${{ steps.vcpkg-restore.outputs.cache-hit != 'true' }}
         name: Save vcpkg packages (if cache miss)
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -88,7 +88,6 @@ jobs:
 
       - name: Setup Developer Command Prompt for VS
         uses: ilammy/msvc-dev-cmd@v1
-        shell: bash
 
       - name: Configure CMake
         shell: bash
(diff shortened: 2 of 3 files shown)


.github/workflows/osx.yml
```diff
@@ -67,16 +67,15 @@ jobs:
...
 
       - name: Save ccache
-        id: cache-ccache-save
         uses: actions/cache/save@v4
         with:
           path: ~/Library/Caches/ccache
           key: ${{ steps.cache-ccache-restore.outputs.cache-primary-key }}
+          
+      - name: Run Tests
+        run: make -C build -j$(sysctl -n hw.logicalcpu) check
```
nilsnolde, line 78: save the ccache cache before we run the tests

.github/workflows/win.yml
```diff
@@ -83,16 +71,26 @@ jobs:
...
+        id: vcpkg-restore
+        uses: actions/cache/restore@v3
+        with:
+          key: vcpkg=${{ env.VCPKG_VERSION }}-msvc=${{ env.MSVC_VERSION }}-json=${{ hashFiles('vcpkg.json') }}
+          path: |
+            vcpkg/*
+            !vcpkg/downloads
+            !vcpkg/docs
+            !vcpkg/buildtrees
+            vcpkg/downloads/tools
```
nilsnolde, line 87: pff, this was super super annoying to get right.. they do have some serious growing pains, think I got involved too early and things change around while they can't break stuff that previously worked.. or they have the "valhalla syndrome" of trying to cover every use case :wink: but worse docs... 

for the interested reader (and my own future reference), assuming `vcpkg` was installed to the root of the repo:

there's 3 (!!) way how one can run `vcpkg`:
1. classic mode: 
  - **no** `vcpkg.json` in the root of the repo
  - `vcpkg install <pkg1> <pkg2>...` to install packages
  - will install stuff to `./vcpkg/installed`
2. manifest mode (command line):
  - `vcpkg.json` in the root of the repo
  - `vcpkg install` (note, no packages given)
  - will install stuff to `./vcpkg_installed`
3. manifest mode (CMake integrated)
  - `vcpkg.json` in the root of the repo
  - simply `cmake -B build .` will install the packages
  - will install stuff to `build/vcpkg_installed`

what I've been doing was combining 2. & 3.: manually run `vcpkg install` to `./vcpkg_installed` which looks to work just fine and then it installs _another_ time with `cmake -B build .` to `build/vcpkg_installed` which is super quick and seems like it was "cached" from the `vcpkg install`. where it really cached from is `./vcpkg/installed` which was already initiated by `vcpkg install`. but it's not doing anything if you cache the `./vcpkg_installed` directory, it will just be ignored by cmake.

TLDR: the modern cmake way of dealing with vcpkg is to not install anything manually and let cmake configuring take care of it. in terms of caching paths, we want the paths of the vcpkg installation. that finally worked. just we want to ignore the temp build stuff (`vcpkg/downloads` etc), the only interesting thing is located in `vcpkg/installed` (and some cached tool stuff like cmake, pkgconf, 7z.. in `vcpkg/downloads/tools`).

jeez, that is so confusing.. I can hear kevin yell "told ya" hahaha I'm still standing by vcpkg, even there's some questionable design involved..

.github/workflows/win.yml
```diff
@@ -111,14 +109,26 @@ jobs:
...
+        name: Save vcpkg packages (if cache miss)
+        uses: actions/cache/save@v3
+        with:
+          key: ${{ steps.vcpkg-restore.outputs.cache-primary-key }}
+          path: |
+            vcpkg/*
+            !vcpkg/downloads
+            !vcpkg/docs
+            !vcpkg/buildtrees
+            vcpkg/downloads/tools
```
nilsnolde, line 124: this saves 500 smth MB (150 or so compressed). keeping all subdirectories of `./vcpkg` easily triples the size.

.github/workflows/win.yml
```diff
@@ -111,14 +109,26 @@ jobs:
...
-            -DPREFER_EXTERNAL=ON \
+            -DPREFER_EXTERNAL=ON
+
+      - if: ${{ steps.vcpkg-restore.outputs.cache-hit != 'true' }}
+        name: Save vcpkg packages (if cache miss)
+        uses: actions/cache/save@v3
+        with:
+          key: ${{ steps.vcpkg-restore.outputs.cache-primary-key }}
+          path: |
+            vcpkg/*
```
nilsnolde, line 120: also `vcpkg/*` is a notation that's needed when you want to exclude subdirectories, it's basically a bug in `action/cache`

nilsnolde:
ok, vcpkg packages are caching fine again. but what a mess.. took a good day to get it right.

fwiw, I looked into ccache for windows/msvc. while it seems to be working, there's lots of reports of not doing thaaat much in terms of speed (yet?). seems it's very slow compared to osx/linux. 

@mloskot do you use some compiler caching for msvc and can share some experience?

mloskot:
@nilsnolde No, not really, or nothing beyond incremental builds supported by VS/VC++. I have also refactored my large projects to utiliise MSBuild parallelisation of `.proj` and `.vcxproj` building and parallel compilation, so there is some consideration put into PCH etc. I used to use IncrediBuild but MSBuild can nowadays do wonders really.

There is https://stashed.io, but I have not used it myself.

nilsnolde:
might also merge this while you're at it. so we're down to 20ish mins again across all workflows

.github/workflows/osx.yml
```diff
@@ -29,7 +29,7 @@ on:
         default: false
 
 jobs:
-  build:
+  build_osx_tests:
```
kevinkreiser, line 32: thanks for this but also why n ot match what you did below and just do `build_osx`

kevinkreiser (approved):

kevinkreiser (approved):

kevinkreiser:
> might also merge this while you're at it.

i assumed you were refering to me so i did the deed apologies if i misunderstood

nilsnolde:
Jep sure😄

PR #4574, opened 2024-02-07
still respect strict dimensional restrictions if `ignore_restrictions=true`

### LINKED ISSUE #4571: Ignore_restrictions for truck routing disregards physical real-world restrictions like maxweight and maxheight. Is it ok?
Hello, when the 'ignore_restrictions' option is set for truck routing, it disregards factors such as 'maxheight' and 'maxweight.' These are critical real-world physical restrictions that cannot be overlooked as they could lead to dangerous situations (hitting bridge e.g.). If someone wants a truck route where 'maxweight' or 'maxheight' is ignored, they can simply make a request where heigh and weight for vehicle is set to 0. Therefore, my question is: shouldn't these 'maxweight' and 'maxheight' restrictions still be adhered to even when 'ignore_restriction' is enabled for truck routing?

nilsnolde:
fixes #4571


test/gurka/test_route.cc
```diff
@@ -33,7 +33,7 @@ class IgnoreAccessTest : public ::testing::Test {
     const std::string ascii_map = R"(
           B----------C
          /            \
-        A<-------------D
+        A<-------------D----G----H---I
```
nilsnolde, line 36: add a turn restriction & a `maxheight` to trigger both car & truck

valhalla/sif/dynamiccost.h
```diff
@@ -601,26 +605,25 @@ class DynamicCost {
...
             // We are in range at the time we are allowed at this edge
-            if (access_type == baldr::AccessType::kTimedAllowed)
-              return true;
-            else if (access_type == baldr::AccessType::kDestinationAllowed)
-              return allow_conditional_destination_ || is_dest;
-            else
-              return false;
+            switch (access_type) {
+              case baldr::AccessType::kTimedAllowed:
+                restricted = false;
```
nilsnolde, line 610: this was a bug: we shouldn't `return true` within this loop. later restrictions might be `false` and that's what we want to get to. however, `return false` within the loop is fine of course

nilsnolde:
Hm, some relevant test failing..

valhalla/baldr/graphconstants.h
```diff
@@ -739,6 +739,8 @@ enum class AccessType : uint8_t {
   kTimedAllowed = 6,
   kTimedDenied = 7,
   kDestinationAllowed = 8,
+  // TODO(nils): this is a problem. when evaluating restrictions in dynamiccost,
+  // we rely on the fact that non-time restrictions come first
   kMaxAxles = 9
```
nilsnolde, line 744: didn't think about it at the time: if a `maxaxle` restriction is combined with a `kTimeAllowed` restriction, then we won't respect the `maxaxle` one. this is because they are sorted in the graph tiles and maxaxles would be evaluated after timed restrictions which return true/false if active.

nilsnolde: but also there was no other way to do this without breaking the tile structure I guess. maybe we should just do 2 loops when evaluating restrictions, one for dimensions/axles etc, one for timed ones

valhalla/sif/dynamiccost.h
```diff
@@ -569,58 +569,58 @@ class DynamicCost {
...
     const std::vector<baldr::AccessRestriction>& restrictions =
         tile->GetAccessRestrictions(edgeid.id(), access_mode);
 
-    bool time_allowed = false;
+    // do 2 loops to not disregard "maxaxles" (restrictions are sorted, see graphconstants.h)
+    for (const auto& restriction : restrictions) {
+      if (!ModeSpecificAllowed(restriction)) {
+        return false;
+      }
+    }
```
nilsnolde, line 583: this is unfortunately necessary, as `maxaxles` was added after the time restrictions and would be disregarded if a timed restriction triggers and returns early

valhalla/sif/dynamiccost.h
```diff
@@ -569,58 +569,58 @@ class DynamicCost {
...
+          // remember restrictions are sorted! see baldr/graphconstants.h
+          switch (access_type) {
+            case baldr::AccessType::kTimedAllowed:
               return true;
-            else if (access_type == baldr::AccessType::kDestinationAllowed)
+            case baldr::AccessType::kDestinationAllowed:
               return allow_conditional_destination_ || is_dest;
-            else
+            default:
+              // baldr::AccessType::kTimedDenied
```
nilsnolde, line 613: not much to see here, just made the code a bit terser and used a switch/case

valhalla/sif/dynamiccost.h
```diff
@@ -569,58 +569,58 @@ class DynamicCost {
...
+    for (const auto& restriction : restrictions) {
+      if (!ModeSpecificAllowed(restriction)) {
+        return false;
+      }
+    }
 
-    for (size_t i = 0; i < restrictions.size(); ++i) {
-      const auto& restriction = restrictions[i];
-      // Compare the time to the time-based restrictions
+    bool time_allowed = true;
```
nilsnolde, line 585: calling the variable `time_allowed` and defaulting to `false` is weird.

valhalla/sif/dynamiccost.h
```diff
@@ -569,58 +573,62 @@ class DynamicCost {
...
+    if (!(edge->access_restriction() & access_mode))
       return true;
 
     const std::vector<baldr::AccessRestriction>& restrictions =
         tile->GetAccessRestrictions(edgeid.id(), access_mode);
 
-    bool time_allowed = false;
+    // do 2 loops to not disregard "maxaxles" (restrictions are sorted, see graphconstants.h)
 
+    TimeRestriction time_type = TimeRestriction::NO_TIME;
```
nilsnolde, line 584: doing an enum here makes it easier to handle things in the loop below

nilsnolde:
just adding @kevinkreiser 's input: 
- `ignore_restrictions` shouldn't change in scope, documented or not. was mostly for map-matching initially to really do ignore _any_ restriction. so we can add a `ignore_safe_restrictions` or so (it's a bit more than just turn restrictions, e.g. timed ones & maxaxles etc)
- the bug fix should be done in a way to change the sorting in the tiles so we don't have to touch the code in dynamiccost.h, which currently results in a bit of a performance sink when many timed restrictions are available per edge (e.g. proprietary data)

I'll close here and open another one at some point.

PR #4463, opened 2023-12-18
Fix ferry cost times

chrstnbwnkl:
Seems like `base_transition_cost.secs` was getting overridden by `OSRMCarTurnDuration` for all motorized vehicle profiles, instead of being added to it.

# Issue

What issue is this PR targeting? If there is no issue that addresses the problem, please open a corresponding issue and link it here.
fixes #4219 

## Tasklist

 - [x] Add tests
 - [x] Add #fixes with the issue number that this PR addresses
 - [ ] Update the docs with any new request parameters or changes to behavior described
 - [x] Update the [changelog](CHANGELOG.md)
 - [ ] If you made changes to the lua files, update the [taginfo](taginfo.json) too.

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

PR #4575, opened 2024-02-08
Isochrone PBF output and multipolygon support

chrstnbwnkl:
# Issue

This PR builds on #4543 but takes it step further by grouping contour rings into multipolygons by means of pont in polygon checks.

## Tasklist

 - [x] Add tests


src/tyr/isochrone_serializer.cc
```diff
@@ -12,40 +12,183 @@ using namespace valhalla::baldr::json;
...
+  for (const auto inner : inner_ptrs) {
+    // construct bbox
+    AABB2<PointLL> inner_bbox(std::vector(inner->rbegin(), inner->rend()));
+    bool found_exterior;
+    // go over exterior rings from smallest to largest
+    for (size_t i = results.size(); i > 0; --i) {
+      AABB2<PointLL> outer_bbox(std::vector(results[i - 1][0]->rbegin(), results[i - 1][0]->rend()));
+
+      // contain check
+      if (outer_bbox.Contains(inner_bbox)) {
```
chrstnbwnkl, line 67: this is the easy way, but would in some cases lead to "misgrouped" inner rings. the better solution would be to do actual polygon-within-polygon checks

src/tyr/isochrone_serializer.cc
```diff
@@ -142,5 +284,68 @@ std::string serializeIsochrones(const Api& request,
...
+                                std::vector<midgard::GriddedData<2>::contour_interval_t>& intervals,
+                                midgard::GriddedData<2>::contours_t& contours,
+                                bool polygons,
+                                bool show_locations) {
+
+  switch (request.options().format()) {
+    case Options_Format_pbf:
+      return serializeIsochronePbf(request, intervals, contours);
+    case Options_Format_json:
+      if (polygons)
```
chrstnbwnkl, line 343: This still needs to change, probably should keep the current serialization as the default and only use the new method when explicitly requested by the user.

test/isochrone.cc
```diff
@@ -85,30 +104,32 @@ void try_isochrone(loki_worker_t& loki_worker,
...
+        test_shape_equality(actual_geom, expected_geom);
+
+      } else {
+        uint32_t actual_poly_size = rp(coord_selector).Get(response)->GetArray().Size();
+        uint32_t expected_poly_size = rp(coord_selector).Get(response)->GetArray().Size();
+
+        ASSERT_EQ(actual_poly_size, expected_poly_size);
+        for (uint32_t j = 0; j < actual_poly_size; ++j) {
+
+          // we only compare exterior rings
```
chrstnbwnkl, line 125: I noticed we never compare full geometry, as we leave out the holes. maybe check those as well?

src/tyr/isochrone_serializer.cc
```diff
@@ -12,40 +12,218 @@ using namespace valhalla::baldr::json;
...
+    std::for_each(inner_ptrs.begin(), inner_ptrs.end(),
+                  [&results](const contour_t* c) { results[0].push_back(c); });
+    return results;
+  }
+
+  // now we're dealing with multiple exterior rings
+  // iterate over outer rings and for each inner ring check if the inner ring is within the outer ring
+  for (const auto inner : inner_ptrs) {
+    // construct bbox
+    AABB2<PointLL> inner_bbox(std::vector(inner->rbegin(), inner->rend()));
```
chrstnbwnkl, line 60: `AABB2` constructor expects vector, not list. there's probably a better way to solve this :smile:

nilsnolde:
One bigger question to be discussed: would this warrant a request param to enable proper multipolygons?

IMO our current GeoJSON polygon support is broken and this is actually fixing that bug. Apart from the fact that it's contours and not polygons and it's a crutch to begin with. But if we offer this we should do it as correctly as we can.

Most libs seem to be forgiving enough to render them half ok-ish even now (with multiple outer rings per polygon feature..). Also I'd guess 99% of consumers use some library's GeoJSON implementation, feeding in just the full thing and getting lucky rendering it properly, so nothing would change for them. As @chrstnbwnkl found out, our own web app at https://github.com/gis-ops/valhalla-app needs a bit of work to handle this, but that's just because we weren't doing using leaflet properly for GeoJSON in the first place (using `L.polygon()` instead of `L.GeoJSON()`).

src/tyr/isochrone_serializer.cc
chrstnbwnkl: use point in polygon

src/tyr/isochrone_serializer.cc
```diff
@@ -12,40 +12,218 @@ using namespace valhalla::baldr::json;
...
+                });
+
+  // exactly one exterior ring, so all inners go in one group
+  if (results.size() == 1) {
+    std::for_each(inner_ptrs.begin(), inner_ptrs.end(),
+                  [&results](const contour_t* c) { results[0].push_back(c); });
+    return results;
+  }
+
+  // iterate over outer rings and for each inner ring check if the inner ring is within the exterior
```
kevinkreiser, line 56: the check can simply take a single point from the inner and doing point in polygon on a given outer. i would consider using the simplified winding number algorithm as proposed in the graphics gems book under "An Incremental Angle Point in Polygon Test". here's a sample implementation of that: https://jeremytammik.github.io/tbc/a/0491_point_in_poly.htm

test/isochrone.cc
chrstnbwnkl: Let's do that

kevinkreiser (dismissed):
looks excellent pbf support and multipolygon with corrected inner outer relationships is awesome

chrstnbwnkl:
So there's a floating point issue with the ARM build again. Shape tolerance in the isochrone test is exceeded by a bit, but I'm hesitant to increase it again, since I just did that not too long ago in #4463.

nilsnolde:
I don’t think we can do too much here and have to deal with it as it comes

docs/docs/api/isochrone/api-reference.md
```diff
@@ -63,18 +63,16 @@ Most JavaScript-based GeoJSON renderers, including [Leaflet](http://leafletjs.co
...
 The Isochrone service is in active development. To report software issues or suggest enhancements, open an issue in the [Valhalla GitHub repository](https://github.com/valhalla/valhalla/issues).
 
 Several other options are being considered as future service enhancements. These include:
 
 * ~~Using distance rather than time for each unit.~~
-* Generating outer contours or contours with interior holes for regions that cannot be accessed within the specified time.
+* ~~Generating outer contours or contours with interior holes for regions that cannot be accessed within the specified time.~~
 * ~~Options to control the minimum size of interior holes.~~
-* Removing self intersections from polygonal contours.
+* ~~Removing self intersections from polygonal contours.~~
```
nilsnolde, line 75: :partying_face:

src/tyr/isochrone_serializer.cc
```diff
@@ -12,117 +12,193 @@ using namespace valhalla::baldr::json;
...
+using grouped_contours_t = std::vector<contour_group_t>;
+using contour_interval_t = std::tuple<size_t, float, std::string, std::string>;
+
+grouped_contours_t GroupContours(const bool polygons, const feature_t& contours) {
+  grouped_contours_t results;
+
+  // if the user requested linestrings, we'll give them linestrings
+  if (!polygons || contours.size() < 2) {
+    std::for_each(contours.begin(), contours.end(),
+                  [&results](const contour_t& c) { results.push_back({&c}); });
```
nilsnolde, line 32: probably we can reserve here with `results.reserve(contours.size()`?

src/tyr/isochrone_serializer.cc
```diff
@@ -12,117 +12,193 @@ using namespace valhalla::baldr::json;
...
+  // exactly one exterior ring, so all inners go in one group
+  if (results.size() == 1) {
+    std::for_each(inner_ptrs.begin(), inner_ptrs.end(),
+                  [&results](const contour_t* c) { results[0].push_back(c); });
+    return results;
+  }
+
+  // iterate over outer rings and for each inner ring check if the inner ring is within the exterior
+  // ring
+  for (const auto inner : inner_ptrs) {
```
nilsnolde, line 58:
```suggestion
  for (const auto* inner : inner_ptrs) {
```

it's easy to get confused if we hide stuff in `auto`

src/tyr/isochrone_serializer.cc
```diff
@@ -12,117 +12,193 @@ using namespace valhalla::baldr::json;
...
+
+    // get the first point of the ring (could be any though)
+    PointLL inner_pt = inner->front();
+    bool found_exterior = false;
+    // go over exterior rings from smallest to largest
+    for (size_t i = results.size(); i > 0; --i) {
+      contour_t ext = *results[i - 1][0];
```
nilsnolde, line 65: if you don't want the pointer, rather take a ref to the underlying value or you do a copy
```suggestion
      const contour_t& ext = *results[i - 1][0];
```

src/midgard/util.cc
```diff
@@ -545,6 +609,33 @@ template bool intersect<PointLL>(const PointLL& u,
...
+    adjust_delta(delta, vertex, next_vertex, pt);
+    angle = angle + delta;
+    quad = next_quad;
+  }
+  return (angle == 4) || (angle == -4);
+};
+
+template bool point_in_poly<valhalla::midgard::PointLL, std::list<valhalla::midgard::PointLL>>(
+    const valhalla::midgard::PointLL&,
+    std::list<valhalla::midgard::PointLL>&);
```
nilsnolde, line 637: the polygon is not `const`?

src/tyr/isochrone_serializer.cc
```diff
@@ -12,117 +12,193 @@ using namespace valhalla::baldr::json;
...
+        results[i - 1].push_back(inner);
+        found_exterior = true;
+        break;
+      }
+    }
+
+    // TODO: reverse winding and make it an exterior ring
+    if (!found_exterior) {
+      LOG_WARN("No exterior ring contour found for inner contour.");
+    }
```
nilsnolde, line 79: did you observe this? probably could happen in the contouring algo?

nilsnolde (commented):
good stuff:) just a few nits here and there from my side

src/tyr/isochrone_serializer.cc
chrstnbwnkl: They're nested vectors, of which we don't know the size at runtime, so not sure this makes any difference

nilsnolde: You’re right, didn’t see the nesting

src/midgard/util.cc
chrstnbwnkl: whoops, updated that

src/tyr/isochrone_serializer.cc
chrstnbwnkl: Haven't  observed this yet, even after excessive playing around (with different graph extracts even)

nilsnolde (approved):

kevinkreiser (approved):

Issue #4576, opened 2024-02-08
Consider using `-Wimplicit-fallthrough`


r-barnes:
`-Wimplicit-fallthrough` requires all non-contiguous switch statement cases to terminate with `break` or `[[fallthrough]]`.

Enabling this has caught many errors in codebases I've worked with. It looks like Valhalla is very nearly able to compile with the flag already, so it locks in the current level of safety.

PR #4577, opened 2024-02-08
Add missing `break` to graphconstants.h

r-barnes:
Adding this break allows the code to compile with `-Wimplicit-fallthrough` enabled as `-Werror`.

This has prevented many bugs in our codebase, so it's useful to have upstream code able to compile with the same flags enabled.

# Issue

#4576 

## Tasklist

 - [ X ] Add tests
 - [ X ] Add #fixes with the issue number that this PR addresses
 - [ X ] Update the docs with any new request parameters or changes to behavior described
 - [ X ] Update the [changelog](CHANGELOG.md) - likely unnecessary
 - [ X ] If you made changes to the lua files, update the [taginfo](taginfo.json) too.

## Requirements / Relations

N/A

### DIFF
--- a/valhalla/baldr/graphconstants.h
+++ b/valhalla/baldr/graphconstants.h
@@ -764,6 +764,7 @@ inline float GetOffsetForHeading(RoadClass road_class, Use use) {
     case Use::kPedestrian:
     case Use::kBridleway: {
       offset *= 0.5f;
+      break;
     }
     default:
       break;


nilsnolde:
Great, thanks! Indeed that happens all the time.. Not sure how to read your comment: you want to enable this single flag on the whole codebase? I’d be happy with that tbh. 

We have this mechanism to make files warning-free (well “-Werror”) but I think it’d be also great to enable single warnings-as-error on the whole codebase to lock it in and not allow any more of those, no matter which file.

kevinkreiser:
honestly i find it pedantic but adding the extra break is also not that egregious either

kevinkreiser (approved):

nilsnolde:
That particular “break” is pedantic, but that warning is one of the more useful ones. Happens to me all the time when I write a switch/case and it’s easily becoming a bug.

PR #4585, opened 2024-02-14
Fix roundoff issue in Tiles

### LINKED ISSUE #4584: Roundoff issue within Tiles
I found an issue with tiling that seems to be due to use of float in the Row and Col methods in tiles.h.
 ```
 PointLL ll(179.999978, -16.805363);
  uint8_t level = TileHierarchy::levels().back().level;
  auto tiling = TileHierarchy::get_tiling(level);
  auto tile_id = tiling.TileId(ll);
  LOG_INFO("TileId = " + std::to_string(tile_id));
  auto base_ll = tiling.Base(tile_id);
  LOG_INFO("Base LL = " + std::to_string(base_ll.lat()) + "," + std::to_string(base_ll.lng()));
```
this produces:
```
2024/02/14 13:28:14.930670 [INFO] TileId = 421920
2024/02/14 13:28:14.930714 [INFO] Base LL = -16.750000,-180.000000
```
which is incorrect. If I change Row, Col methods to double then I get the correct tile.
```
2024/02/14 13:30:00.544148 [INFO] TileId = 421919
2024/02/14 13:30:00.544171 [INFO] Base LL = -17.000000,179.750000
```
One question is whether fixing this would lead to any compatibility issues with prior data?

### PR DESCRIPTION
**Pull request by dnesbitt61**
# Issue

#fixes #4584
Using float in the Tiles Row and Col methods can lead to an incorrect tile id being computed (along with incorrect base lat, lng). Using double as the argument to these methods fixes the issue.
For discussion - will this change cause any compatibility issues?
## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/valhalla/midgard/tiles.h
+++ b/valhalla/midgard/tiles.h
@@ -123,7 +123,7 @@ public:
    * @param   y   y coordinate
    * @return  Returns the tile row. Returns -1 if outside the tile system bounds.
    */
-  int32_t Row(const float y) const {
+  int32_t Row(const double y) const {
     // Return -1 if outside the tile system bounds
     if (y < tilebounds_.miny() || y > tilebounds_.maxy()) {
       return -1;
@@ -139,7 +139,7 @@ public:
    * @param   x   x coordinate
    * @return  Returns the tile column. Returns -1 if outside the tile system bounds.
    */
-  int32_t Col(const float x) const {
+  int32_t Col(const double x) const {
     // Return -1 if outside the tile system bounds
     if (x < tilebounds_.minx() || x > tilebounds_.maxx()) {
       return -1;
--- a/valhalla/midgard/tiles.h
+++ b/valhalla/midgard/tiles.h
@@ -123,7 +123,7 @@ public:
    * @param   y   y coordinate
    * @return  Returns the tile row. Returns -1 if outside the tile system bounds.
    */
-  int32_t Row(const double y) const {
+  int32_t Row(const typename coord_t::value_type y) const {
     // Return -1 if outside the tile system bounds
     if (y < tilebounds_.miny() || y > tilebounds_.maxy()) {
       return -1;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by gknisely** [comment]
@dnesbitt61 need to update the changelog

File: valhalla/midgard/tiles.h [hunk 1]
```diff
@@ -123,7 +123,7 @@ template <class coord_t> class Tiles {
    * @param   y   y coordinate
    * @return  Returns the tile row. Returns -1 if outside the tile system bounds.
    */
-  int32_t Row(const float y) const {
+  int32_t Row(const double y) const {
```
kevinkreiser, line 126: to correctly adhere to the template parameter we should probably use:

```suggestion
  int32_t Row(const typename coord_t::value_type y) const {
```

kevinkreiser:
im trying to reason about your comment on data compatibilty. old code would expect that a node would land in a different tile than new code. which means old data built with old code would have certain nodes land in the wrong tiles. this would be a problem if we used this method in a scenario where:

1. we had old code
2. we had new data
3. loki used it to find the right tile to look for candidate edges
4. but the n ew data wou ld have the edges in a different tile

the good news is this doesnt really matter because loki never stops looking in one place it looks at closest first and next closes next and so on. the only risk is that in this scenario it finds something that isnt the actual closest thing, which does suck but its not catastrophic. to me i think we can ship this without worrying about compatibility. the data wont crash anything it will at worst give sub par results for these specific edge cases

valhalla/midgard/tiles.h
dnesbitt61: agree, fixed.

kevinkreiser (approved):

dnesbitt61:
how can codecov fail! I added a unit test. Oh well

nilsnolde:
🎲 🎲

nilsnolde (approved):
good catch again thanks!

PR #4589, opened 2024-02-19
refactor how costing options are parsed

nilsnolde:
I was working more on #3833 in the past days and realized it's currently not possible to add warnings for clamped costing options. the reason was that we _always_ parse _all_ costing options, regardless what the request actually wants. that way we'd get costing specific warnings for clamped values which aren't relevant for the requested costing, e.g. `top_speed` would trigger a warning by default since motor_scooter defaults to 45 kph while car defaults to 120 kph. 

I did find it strange that we even do parse everything always, so I dug a bit. from what I can tell, there's 2 main reasons why that was done:
1. recostings can be any kind of costing and comes with JSON and PBF which makes parsing a bit awkward
2. multi-modal requests (transit, BSS) can have multiple costings

those are totally understandable and it worked out so far. anyways, I took a shot at refactoring how things are done, so we can 
- support warnings for clamped costing options
- avoid doing a tiny little bit more work than we have to; in 99.99% of requests people are only interested in a single costing model (no recostings, no multimodal)


src/sif/dynamiccost.cc
```diff
@@ -489,16 +489,16 @@ void ParseBaseCostOptions(const rapidjson::Value& json,
 void ParseCosting(const rapidjson::Document& doc,
                   const std::string& costing_options_key,
                   Options& options) {
-  // if specified, get the costing options in there
-  for (auto i = Costing::Type_MIN; i <= Costing::Type_MAX; i = Costing::Type(i + 1)) {
+  // get the needed costing options in there
+  for (const auto& costing_type : kCostingTypeMapping.at(options.costing_type())) {
```
nilsnolde, line 493: instead of going over all possible costings, I did a map with the relevant ones per request costing. the value has to be a vector to support multimodal, but most costings map to a single one

nilsnolde: this function is only being used for valid `options` and we can expect a `option.costing_type` to be defined.

src/valhalla_path_comparison.cc
```diff
@@ -267,18 +267,19 @@ int main(int argc, char* argv[]) {
...
   if (valhalla::Costing_Enum_Parse(routetype, &costing)) {
     request.mutable_options()->set_costing_type(costing);
   } else {
     throw std::runtime_error("No costing method found");
   }
+
+  if (!map_match) {
+    rapidjson::Document doc;
+    sif::ParseCosting(doc, "/costing_options", *request.mutable_options());
+  }
```
nilsnolde, line 281: lots of changes deal with this: `ParseCosting` has to be after setting the `options.costing_type`

src/worker.cc
```diff
@@ -586,6 +586,38 @@ void parse_contours(const rapidjson::Document& doc,
   }
 }
 
+// parse all costings needed to fulfill the request, including recostings
+void parse_recostings(const rapidjson::Document& doc,
```
nilsnolde, line 590: before the code was avoiding to extract the costings from `recostings` due  to JSON/PBF (my guess) and preferred to parse all costing options instead. now it's parsing only the ones which are specified.

test/parse_request.cc
```diff
@@ -254,8 +254,8 @@ std::string get_request_str(const std::string& grandparent_key,
                             const std::string& parent_key,
                             const std::string& key,
                             const bool specified_value) {
-  return R"({")" + grandparent_key + R"(":{")" + parent_key + R"(":{")" + key + R"(":)" +
-         std::string(specified_value ? "true" : "false") + R"(}}})";
+  return R"({"costing":")" + parent_key + R"(",")" + grandparent_key + R"(":{")" + parent_key +
```
nilsnolde, line 257: jeez, this whole test file is an abomination hahaha

valhalla/sif/costfactory.h
```diff
@@ -43,6 +43,7 @@ class CostFactory {
     Register(Costing::pedestrian, CreatePedestrianCost);
     Register(Costing::truck, CreateTruckCost);
     Register(Costing::transit, CreateTransitCost);
+    Register(Costing::multimodal, CreateTransitCost); // dummy so it behaves like the rest
```
nilsnolde, line 46: this was necessary for the changes in `CreateModeCosting`

valhalla/sif/costfactory.h
```diff
@@ -100,21 +101,18 @@ class CostFactory {
 
   mode_costing_t CreateModeCosting(const Options& options, TravelMode& mode) {
     mode_costing_t mode_costing;
-    // Set travel mode and construct costing
+    // Set travel mode and construct costing(s) for this type
+    for (const auto& costing : kCostingTypeMapping.at(options.costing_type())) {
```
nilsnolde, line 105: use the same costing map here

test/parse_request.cc
kevinkreiser: its jsons fault really. c++ needs a way like python where you can do format strings as block literals:

```python
f'''
{"i am": "a literal with variable {1+2} substitution"}
'''
```

src/worker.cc
```diff
@@ -842,6 +874,12 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
     }
   }
 
+  // TODO(nils): if we parse the costing options before the ignore_closures logic,
```
kevinkreiser, line 877: then why not just move it back down where it was before?

test/parse_request.cc
nilsnolde: urgh yeah.. but it's also those 100s of helper functions with the generally annoying habit of not failing tests in an obvious way and you have to start the debugger to know which line really failed. personally I take a 10k LOC test file repeating itself over and over instead of a 1k LOC test file which abstracts too much

src/worker.cc
nilsnolde: I actually wanted to have it right where the costing is created many lines up. 

I was just quickly testing if that's really the problem and left it that way to correlate it better to ignore_closures stuff since that's what makes it break and I don't immediately see why (and want to investigate another day). so it's more for future convenience that I left it here. I'll move it to the original place of course, if it's important to you

kevinkreiser (approved):

PR #4590, opened 2024-02-19
Update CHANGELOG.md

nilsnolde:
was just about to comment on this, but then figured I'll try my luck right now:) 

removes the "crystal bowl" from the changelog and just puts "Unreleased" which IMO is more to the point.


kevinkreiser (approved):
kristallkugel, also: crystal ball

it was only there to give a hint what to change to. should have the release process fail if it's not the right format maybe one day

PR #4594, opened 2024-02-20
GeoTIFF serialization for isochrones

chrstnbwnkl:
# Issue

Adds GeoTIFF as an output format to the isochrones endpoint, which returns the 2-D grid as a raster instead of contours. 

image

The rough idea is this: when a user requests a raster output  by specifing `"format": "geotiff"` in the request JSON, once the 2-D expansion grid is calculated, the grid is serialized as a GeoTIFF dataset and returned as raw bytes, instead of running the contouring algorithm. 

## Implementation Summary

  - uses [GDAL](https://gdal.org/) for writing the raster dataset; when building Valhalla, one needs to actively opt in to this feature by setting the cmake flag `ENABLE_GDAL` (`OFF` by default)
  - raster data type: using `uint16_t` gives plenty range for distance and time metrics, so I decided to scale the values to seconds (which maxes out  at ~18 hours) and 10 meter steps (~650 kms) 
  - each metric gets its own band
  - the data grid has an immense padding, which leaves more than 50% of cells never reached by the expansion (in most cases probably more than 80% though), so the final raster is truncated to the minimum extent that actually contains values not equal to the set max value, plus just a little padding that reassures users we didn't accidentally cut off part of the grid :sweat_smile:  

For more info, see the initial discussion: #4581. 

## TODO

There are still one or two things left to do:
  - figure out a way to return the file buffer as a string without copying it
  - ~~find out why raster values in the added tests are unexpectedly high (probably related to type casting)~~
 

## Future steps

This is a basic first step, but it doesn't have to be the end of it. One could

  - add color information for renderers by making use of color tables (I have fiddled around with it but haven't gotten far :sweat_smile:)
  -  add some more metadata that can be useful for the ouput interpretation, e.g. about the snapped locations (analogous to `show_locations`) 

## Tasklist

 - [x] Add tests


src/loki/worker.cc
```diff
@@ -357,6 +357,12 @@ loki_worker_t::work(const std::list<zmq::message_t>& job,
         result.messages.emplace_back(request.SerializeAsString());
         break;
       case Options::isochrone:
+// return early if geotiff was requested but GDAL isn't enabled
+#ifndef ENABLE_GDAL
```
chrstnbwnkl, line 361: be prepared to see plenty of these... I'm open to any suggestions that reduce these wherever possible

src/thor/worker.cc
```diff
@@ -15,6 +15,10 @@
 
 #include <boost/property_tree/ptree.hpp>
 
+#ifdef ENABLE_GDAL
+#include <gdal_priv.h>
+#endif
+
```
chrstnbwnkl, line 21: `tyr` has no way to maintain state, so `thor_worker_t` will hold the geotiff driver and pass an opaque pointer to it to the serializer function (note that [GDAL is not thread-safe](https://gdal.org/user/multithreading.html#gdal-api-re-entrant-but-not-thread-safe))

src/tyr/isochrone_serializer.cc
```diff
@@ -265,17 +351,39 @@ std::string serializeIsochronePbf(Api& request,
...
 
+#ifdef ENABLE_GDAL
 std::string serializeIsochrones(Api& request,
                                 std::vector<midgard::GriddedData<2>::contour_interval_t>& intervals,
-                                midgard::GriddedData<2>::contours_t& contours,
-                                bool polygons,
-                                bool show_locations) {
+                                const std::shared_ptr<const midgard::GriddedData<2>>& isogrid,
+                                valhalla::thor::geotiff_driver_t geotiff_driver) {
+#else
```
nilsnolde, line 359: this looks like it'd be better served with an overload instead of `#ifdef`ing in the arguments

valhalla/thor/worker.h
```diff
@@ -6,6 +6,9 @@
 #include <vector>
 
 #include <boost/property_tree/ptree.hpp>
+#ifdef ENABLE_GDAL
+#include <gdal_priv.h>
+#endif
```
nilsnolde, line 11: it'd be great if we wouldn't leak gdal into our headers. it's easy to push this particular one into its .cc file, not sure about the serializer though. the reason is: if it's in our public headers, then a consuming application also needs the gdal headers, while it'd only need a gdal library if it's in our .cc files (i.e. only in our static/shared library)

src/valhalla_run_isochrone.cc
```diff
@@ -79,6 +83,12 @@ int main(int argc, char* argv[]) {
   // Process json request
   Api request;
   ParseApi(json_str, valhalla::Options::isochrone, request);
+
+#ifdef ENABLE_GDAL
+  if (request.options().format() == Options_Format_geotiff) {
+    GDALRegister_GTiff();
+  }
+#endif
```
nilsnolde, line 91: haha that's noble but honestly, these executables are hardly used by anyone (AFAIK). dave uses the matrix one now and then for perf testing I believe, but other than that they're not too useful IMO. more for future reference:)

src/worker.cc
```diff
@@ -138,6 +138,8 @@ const std::unordered_map<unsigned, valhalla::valhalla_exception_t> error_codes{
     {445, {445, "Shape match algorithm specification in api request is incorrect. Please see documentation for valid shape_match input.", 400, HTTP_400, OSRM_INVALID_URL, "wrong_match_type"}},
     {499, {499, "Unknown", 400, HTTP_400, OSRM_INVALID_URL, "unknown"}},
     {503, {503, "Leg count mismatch", 400, HTTP_400, OSRM_INVALID_URL, "wrong_number_of_legs"}},
+    {504, {504, "This service does not support GeoTIFF serialization.", 400, HTTP_400, OSRM_INVALID_VALUE, "wrong_number_of_legs"}},
```
nilsnolde, line 141:
```suggestion
    {504, {504, "This service does not support GeoTIFF serialization.", 400, HTTP_400, OSRM_INVALID_VALUE, "unknown"}},
```
? not sure how/if that maps to any osrm error description

CMakeLists.txt
```diff
@@ -203,6 +204,15 @@ if(ENABLE_SERVICES)
   endif()
 endif()
 
+# gdal
+set(gdal_target "")
+if (ENABLE_GDAL)
+  find_package(GDAL REQUIRED)
+  add_compile_definitions(ENABLE_GDAL)
+  set(gdal_target GDAL::GDAL)
+endif()
```
nilsnolde, line 213: apparently a FindGDAL module is only available since cmake 3.14. it's not at all a problem to bump the cmake version, 3.12 is ancient. another option would be to use pkg-config (see e.g. `zlib`, `spatialite` etc), which is my preferred option as then we're not relying on cmake versioned Find* modules which is really crap generally IMO, or on cmake-based package manager installations for some package-provided GdalConfig.cmake file (often not the case on debian based)

cmake/ValhallaPkgConfig.cmake
```diff
@@ -33,6 +33,9 @@ function(configure_valhalla_pc)
   if(ENABLE_SERVICES)
     list(APPEND REQUIRES libprime_server)
   endif()
+  if(ENABLE_GDAL)
+    list(APPEND REQUIRES GDAL::GDAL)
```
nilsnolde, line 37: this need to be 
```suggestion
    list(APPEND REQUIRES gdal)
```

it's the name of the .pc file in <lib_dir>/pkgconfig/.

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -620,4 +620,5 @@ The codes correspond to code returned from a particular [Valhalla project](https
 |**5xx** | **Tyr project codes** |
 |500 | Failed to parse intermediate request format |
 |501 | Failed to parse TripDirections |
+|504 | GeoTIFF serialization not supported by service |
```
nilsnolde, line 623: you're _very_ diligent haha not sure this section has seen any updates since I joined the project

nilsnolde (commented):
great stuff, thanks!

not sure I'll join tmrw for the meeting so I left some comments here. maybe there's some way to hide the gdal headers in the implementation instead of exposing them in our headers. anyways, kevin is the master pimpl'er here :D

nilsnolde:
maybe @dnesbitt61 you want to look at this too, kevin mentioned you had this idea before too:)

src/worker.cc
chrstnbwnkl: Yeah I didn't get what these errors have to do with OSRM anyway, probably compatibility with clients for mapbox?

nilsnolde: Jep that’s right. Same as the whole “osrm” format business for the response.

src/tyr/isochrone_serializer.cc
chrstnbwnkl: then that type would need to made available though even though `ENABLE_GDAL` is off?

src/proto_conversions.cc
```diff
@@ -10,7 +10,7 @@ const std::string& MatrixAlgoToString(const valhalla::Matrix::Algorithm algo) {
   static const std::unordered_map<valhalla::Matrix::Algorithm, const std::string> algos{
       {valhalla::Matrix::CostMatrix, "costmatrix"},
       {valhalla::Matrix::TimeDistanceMatrix, "timedistancematrix"},
-      {valhalla::Matrix::TimeDistanceBSSMatrix, "timedistancbssematrix"},
+      {valhalla::Matrix::TimeDistanceMatrix, "timedistancbssematrix"},
```
chrstnbwnkl, line 13: geez, didn't mean to change this... I think this was tidy's fault

src/tyr/isochrone_serializer.cc
```diff
@@ -134,6 +140,86 @@ void addLocations(Api& request, valhalla::baldr::json::ArrayPtr& features) {
...
+    // seconds or 10 meter steps
+    float scale_factor = metric_idx == 0 ? 60 : 100;
+    for (int32_t i = 0; i < ext_y; ++i) {
+      for (int32_t j = 0; j < ext_x; ++j) {
+        auto tileid = isogrid->TileId(j + box[0], i + box[1]);
+        data[i * ext_x + j] =
+            static_cast<uint16_t>(isogrid->DataAt(tileid, metric_idx) * scale_factor);
+      }
+    }
+    auto band = geotiff_dataset->GetRasterBand(nbands == 2 ? (metric_idx + 1) : 1);
```
chrstnbwnkl, line 203: set the nodata value

src/tyr/isochrone_serializer.cc
```diff
@@ -134,6 +140,86 @@ void addLocations(Api& request, valhalla::baldr::json::ArrayPtr& features) {
...
+  std::stringstream ss;
+  ss << "/vsimem/" << std::this_thread::get_id() << "_"
+     << std::chrono::high_resolution_clock::now().time_since_epoch().count();
+  return ss.str();
+}
+
+std::string serializeGeoTIFF(Api& request,
+                             std::vector<contour_interval_t> intervals,
+                             std::shared_ptr<const GriddedData<2>> isogrid,
+                             valhalla::thor::geotiff_driver_t geotiff_driver) {
```
chrstnbwnkl, line 155: move this to the serializer

.github/workflows/osx.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       
       - name: Install dependencies
         run: |
-          HOMEBREW_NO_AUTO_UPDATE=1 brew install autoconf automake protobuf cmake ccache libtool sqlite3 libspatialite luajit curl wget czmq lz4 spatialite-tools unzip boost
+          HOMEBREW_NO_AUTO_UPDATE=1 brew install autoconf automake protobuf cmake ccache libtool sqlite3 libspatialite luajit curl wget czmq lz4 spatialite-tools unzip boost gdal
```
chrstnbwnkl, line 43: run this on all targets

test/isochrone.cc
```diff
@@ -329,6 +333,144 @@ TEST(Isochrones, test_max_reserved_labels_count) {
...
+  bool no_intermediate_values = true;
+  for (size_t i = 0; i < array_size; ++i) {
+    if (dataArray[i] > 0 && dataArray[i] < min_max[1])
+      no_intermediate_values = false;
+  }
+  ASSERT_EQ(no_intermediate_values, false);
+  VSIFCloseL(handle);
+}
+
+TEST(Isochrones, test_geotiff_output_time) {
```
chrstnbwnkl, line 380: we don't care about the exact dimensions, just test the corners

nilsnolde:
so it seems debian & fedora based distros install gdal headers to `/usr/include/gdal/` while pacman/arch/pkgbuild, brew & vcpkg install it to `/usr/include/` directly. while I'd more agree with the former pattern (gdal has tons of installed headers), the latter one is coming from gdal's own cmake. I wouldn't really like to hack the cmake cxx/c flags for gdal. but maybe I'm not quite at full cognitive height anymore.. I asked upstream: https://github.com/OSGeo/gdal/issues/9276

src/tyr/isochrone_serializer.cc
```diff
@@ -2,12 +2,17 @@
...
 #include "midgard/pointll.h"
+#include "thor/worker.h"
 #include "tyr/serializers.h"
 
 #include <cmath>
 #include <sstream>
 #include <utility>
 
+#ifdef ENABLE_GDAL
+#include <gdal/gdal_priv.h>
```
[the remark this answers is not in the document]
kevinkreiser: i thought the same thing but the examples show it for the c++ types: https://gdal.org/tutorials/raster_api_tut.html

it might be a similar situation to geos though wehre the c api is stable and the c++ one less so. i havent checked on gdal use in c++ since long long ago when i first used it in valhalla for elevation datasets

nilsnolde:
yeah, ubuntu packaging messes up the .pc file: https://github.com/OSGeo/gdal/issues/9276#issuecomment-1958181391

nilsnolde:
uh oh, windows is failing with some msvc specific stuff :fearful:

CMakeLists.txt
```diff
@@ -113,12 +112,19 @@ if (ENABLE_HTTP OR ENABLE_DATA_TOOLS)
   endif()
 endif()
 
+# prefer CONFIG mode over MODULE mode, which versions configuration on the package, not CMake
+# NOTE: this is only supported for cmake >= 3.15, but shouldn't be a problem in real life
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
```
nilsnolde, line 117: I put this now way up top so all `find_package` calls will prefer CONFIG mode if available (i.e. the appropriate cmake module was part of the dependency's installation)

src/tyr/isochrone_serializer.cc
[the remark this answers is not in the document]
nilsnolde: uff I don't hope it's the same as geos haha that was a huge PITA. but I doubt it, geos was flagging that in capital letters in their readme, gdal doesn't seem to do that.

chrstnbwnkl:
Is tidy hallucinating or am I?

CMakeLists.txt
nilsnolde: Can you also bump the cmake minimum version to 3.14 @chrstnbwnkl ? Just to be consistent.

src/worker.cc
nilsnolde: looking at all comments this seems to be the only one left where we need to do smth. without going into osrm error codes, IMO "unknown" should never be totally wrong :) at least `wrong_number_of_legs` is wrong

src/tyr/isochrone_serializer.cc
nilsnolde: is this still a relevant TODO?

nilsnolde (approved):

kevinkreiser (approved):

Issue #4599, opened 2024-02-22
Don't contract nodes connecting edges with different names? Why?


nilsnolde:
I was debugging our TomTom converter and did a deeper analysis of the expansion where I found that also in OSM we break shortcuts fairly often when the names of two edges aren't exactly the same. For example here with `secondary` roads:
https://www.openstreetmap.org/way/294613493: has `ref=227` and `name=Žatecká`
https://www.openstreetmap.org/way/523627234: has `ref=227`

I can't see how names are important for expansion/costing, which is the only place where shortcuts matter right? Or is there some stuff in map-matching I'm not aware of? If so, would it be enough to have _some_ name match of the 2 edges (would resolve the above case)?

nilsnolde:
Or they're used for narration too? Somehow I always figured we'd need to recover the base edges for that..

nilsnolde:
Also why do the `speeds()` have to match for edges in the same direction? The shortcut gets a new speed anyways, based on the total length and duration, and it's quite common for lower class roads like tertiary to have alternating speeds.

kevinkreiser:
yep you are right its a product of hysterical raisins! in the early days we didnt recover shortcuts before narrating the path i added that once we started supporting live traffic. anyway because we shipped paths with shortcuts to narrative/maneuver generation we couldnt contract on properties that are important to that process. now that we bit the bullet and implemented recovering the shortcuts we could drop this as a requirement for contracting a node.

basically its as you said, when making shortcuts back in the day we had to worry about both costing and narrative/maneuvers but now that we recover the shortcuts before the latter happens we only need to worry about costing during shortcut creation. good catch!

nilsnolde:
Awesome! Then I’ll remove those restrictions, that’ll speed up a bit.

nilsnolde:
this landed a while ago

PR #4602, opened 2024-02-23
TODOs to create longer shortcuts

nilsnolde:
EDIT: I'll leave this open as draft, it contains all potential todos I could think of. I'll open another PR with the changes to remove names and speed.

relates to #4599 but goes much wider in scope.

for now, I just left a few TODOs in places which I think could be re-investigated. I might miss a few important concepts downstream of creating shortcuts, e.g. I find the `shortcuts` mask and `superseded` still a bit hard to make full sense of.

I'd like to discuss those TODOs before I jump in to work on it. but I think (and hope) we can create quite a few less and quite a bit longer shortcuts in the future.

I should expand on my thoughts a bit though:

it seems we only contract a node when it has **2 outbound edges** on the same hierarchy, but also we even only allow those 2 edges to have the same FRC (for costing purposes). that's the key to most TODOs. 

consider these 2 shortcuts we produce today:
```
A--------------B-------------C
```
where AB and BC differ in access modes and/or access restrictions and/or tolls and/or most other attributes. 

note that for an algorithm it doesn't matter what B is connected to otherwise. if it's connected to another edge of the same hierarchy, B wouldn't ever be contracted. so an algorithm would need to traverse BOTH AB and BC in any case, it has no way of turning anywhere from B. hence it doesn't matter if break up the shortcut and we might as well not and assign the most restrictive attributes to the full thing ABC and contract B away.

does that make any sense?

### DIFF
--- a/src/mjolnir/shortcutbuilder.cc
+++ b/src/mjolnir/shortcutbuilder.cc
@@ -51,6 +51,7 @@ bool EdgesMatch(const graph_tile_ptr& tile, const DirectedEdge* edge1, const Dir
 
   // Make sure access matches. Need to consider opposite direction for one of
   // the edges since both edges are outbound from the node.
+  // TODO(nils): would make sense to take the "minimum" access, i.e. AND access masks?
   if (edge1->forwardaccess() != edge2->reverseaccess() ||
       edge1->reverseaccess() != edge2->forwardaccess()) {
     return false;
@@ -58,11 +59,15 @@ bool EdgesMatch(const graph_tile_ptr& tile, const DirectedEdge* edge1, const Dir
 
   // Neither directed edge can have exit signs or be a roundabout.
   // TODO - other sign types?
+  // TODO(nils): not sure these make sense either anymore.. it's fine to have shortcuts _within_
+  //   roundabouts no? we won't ever contract nodes which have more than 2 outbound edges
   if (edge1->sign() || edge2->sign() || edge1->roundabout() || edge2->roundabout()) {
     return false;
   }
 
   // Neither edge can be part of a complex turn restriction or differ on access restrictions
+  // TODO(nils): complex restrictions make sense, but access restrictions: same as access masks
+  //   IMO, we should take the most restrictive value for the entire shortcut
   if (edge1->start_restriction() || edge1->end_restriction() || edge2->start_restriction() ||
       edge2->end_restriction() || edge1->access_restriction() != edge2->access_restriction()) {
     return false;
@@ -72,6 +77,11 @@ bool EdgesMatch(const graph_tile_ptr& tile, const DirectedEdge* edge1, const Dir
   // NOTE: might want "better" bridge attribution. Seems most overpasses
   // get marked as a bridge and lead to less shortcuts - so we don't consider
   // bridge and tunnel here
+  // TODO(nils):
+  //   - links restriction could be removed? we don't care about it costing I think
+  //   - most other attributes (other than classification & surface) could also take the "most
+  //   restrictive"
+  //      for the entire shortcut
   if (edge1->classification() != edge2->classification() || edge1->link() != edge2->link() ||
       edge1->use() != edge2->use() || edge1->toll() != edge2->toll() ||
       edge1->destonly() != edge2->destonly() || edge1->destonly_hgv() != edge2->destonly_hgv() ||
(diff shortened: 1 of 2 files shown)


src/mjolnir/shortcutbuilder.cc
```diff
@@ -69,6 +74,11 @@ bool EdgesMatch(const graph_tile_ptr& tile, const DirectedEdge* edge1, const Dir
   // NOTE: might want "better" bridge attribution. Seems most overpasses
   // get marked as a bridge and lead to less shortcuts - so we don't consider
   // bridge and tunnel here
+  // TODO(nils):
+  //   - links restriction could be removed? we don't care about it costing I think
+  //   - most other attributes (other than classification & surface) could also take the "most
+  //   restrictive"
+  //      for the entire shortcut
```
nilsnolde, line 81: actually: if the attribute is used to cost by **a factor** (i.e. it scales with length of the edge), then we do need to break up. but if it's like `destination_only` lump penalty we don't need to

dnesbitt61: there can't be many (any?) destination_only motorways/trunks. I assume you are looking only at attributes that impact driving costing classes (pretty sure bicycle and pedestrian skip shortcuts).

src/mjolnir/shortcutbuilder.cc
```diff
@@ -214,12 +226,16 @@ bool CanContract(GraphReader& reader,
...
   // If either opposing directed edge has exit signs return false
+  // TODO(nils): can this be removed? if there's a third outbound edge involved to exit to
+  //   we won't look at this node anyways..
   if (oppdiredge1->sign() || oppdiredge2->sign()) {
     return false;
   }
 
   // Do not allow a shortcut on a ramp crossing at a traffic signal or where
   // more than 3 edges meet.
+  // TODO(nils): why? this doesn't seem important for costing/expansion, also we only allow exactly 2
```
dnesbitt61, line 237: this seems like the sort of thing that is added to fix some issue that was found without it. There are tons of special cases (as I remember) with ramps as they move from highway to "regular" roads.

nilsnolde: yeah I'm just wondering how that can have any effect on shortcut creation, since if there's a ramp, that node would have at least 3 outgoing edges and we're early-abort its  contraction

nilsnolde: should I work on this, I guess there's a ton of little scenarios I'll need to test

src/mjolnir/shortcutbuilder.cc
nilsnolde: > there can't be many (any?) destination_only motorways/trunks

that's true. even tertiary would be rare I imagine (the least FRC we even contract)

dnesbitt61: Was there a change to create shortcuts on secondary and tertiary? Or is my memory bad? I thought we only created shortcuts on level 0 (motorways, trunks, primary).

nilsnolde: Also level 1:)

nilsnolde: That’s why I think it’d make quite an impact

nilsnolde:
initial results are encouraging: I used an extract with ~ 1 Mio ways and just by removing names & speeds restriction from shortcuts we went down from 10166 shortcuts in level 0 (avg 5 edges/shorcut) to 6978 shortcuts (avg 8 edges/shortcut). even better results for level 1!

test/astar.cc
```diff
@@ -1570,9 +1570,8 @@ TEST(BiDiAstar, test_recost_path) {
            3  4        5
   )";
   const gurka::ways ways = {
-      // make ABC to be a shortcut
+      // make ABCDE to be a shortcut
```
nilsnolde, line 1573: somehow this test is still failing.. it now takes the supposedly longer route, not ABCDE and I don't understand why yet..

src/mjolnir/shortcutbuilder.cc
```diff
@@ -736,8 +750,13 @@ void ShortcutBuilder::Build(const boost::property_tree::ptree& pt) {
...
     LOG_INFO("Creating shortcuts on level " + std::to_string(tile_level->level));
-    [[maybe_unused]] uint32_t count = FormShortcuts(reader, *tile_level);
-    LOG_INFO("Finished with " + std::to_string(count) + " shortcuts");
+    [[maybe_unused]] shortcut_stats stats = FormShortcuts(reader, *tile_level);
+    [[maybe_unused]] uint32_t avg = 0;
+    if (stats.first)
+      avg = stats.second / stats.first;
+    LOG_INFO("Finished with " + std::to_string(stats.first) + " shortcuts superseding " +
+             std::to_string(stats.second) + " edges, average ~" + std::to_string(avg) +
+             " edges per shortcut");
```
nilsnolde, line 759: print more info to the build logs. only the number of shortcuts is meaningless without the total number of superseded edges

src/mjolnir/shortcutbuilder.cc
```diff
@@ -77,17 +90,10 @@ bool EdgesMatch(const graph_tile_ptr& tile, const DirectedEdge* edge1, const Dir
...
-  std::vector<std::string> edge2names = tile->GetNames(edge2);
-  std::sort(edge1names.begin(), edge1names.end());
-  std::sort(edge2names.begin(), edge2names.end());
-  if (edge1names != edge2names)
-    return false;
-
   // if they have access restrictions those must match (for modes that use shortcuts)
+  // TODO(nils): same as above: same as access masks
+  //   IMO, we should take the most restrictive value for the entire shortcut
+  //   will be a bit harder to determine for timed restrictions though
```
nilsnolde, line 96: one more comment here: access restrictions are mode specific. but that doesn't matter for most as those modes are "static" (always the same, e.g. height/weight etc). so it still makes sense to only copy the most restrictive one for each type.

only for conditional restrictions modes are dynamic. to not make our life too hard here, I'd propose to simply copy all of the conditional ones onto the shortcut. there's few ones anyways, even in commercial datasets, and if we'd break up shortcuts because of them we'd look at them anyways.

nilsnolde:
getting back to this next week

gknisely:
@nilsnolde definitely should test multiple countries/areas with this change.  All of the test files in `request_files.txt.TEMPLATE` look good to use.  I would add demo_routes.txt, es_routes.txt, na_motorroad.txt, uk_routes.txt, and probably a few city to city examples.

nilsnolde:
`request_files.txt.TEMPLATE` where's that? wrong fork? :smile:

PR #4603, opened 2024-02-23
Fix out-of-range linestrings in expansion results

### LINKED ISSUE #3869: Expansion returning out-of-range edges
I never really used the endpoint so far, just been testing a bit since I want to include it in a new QGIS plugin.

Seems it's including tons of completely out-of-range edges for isochrones, sprinkled all over the place:
image

That's even far outside the 10 mins/km buffer we have for the isotile. Will look at it when I have more time.

chrstnbwnkl:
# Issue

Fixes #3869. The early return caused the callback not be set to null, so it was called even though the current edge was outside of the expansion tracking thresholds.

Hard to produce a test case with gurka, the ascii map would need to be quite big to account for the metric padding :smile: If deemed necessary, I can add a test on the Utrecht tiles and make sure all the returned edges are connected to their pred edge. 
 
## Tasklist

 - [ ] Add tests

### DIFF
--- a/src/thor/isochrone.cc
+++ b/src/thor/isochrone.cc
@@ -323,7 +323,7 @@ ExpansionRecommendation Isochrone::ShouldExpand(baldr::GraphReader& /*graphreade
   // updating or pushing the label back
   // prune the edge if its start is above max contour
 
-  ExpansionRecommendation recommendation = time > max_seconds_ && dist > max_meters_
+  ExpansionRecommendation recommendation = (time > max_seconds_ && dist > max_meters_)
                                                ? ExpansionRecommendation::prune_expansion
                                                : ExpansionRecommendation::continue_expansion;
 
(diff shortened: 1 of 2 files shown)


src/thor/isochrone.cc
```diff
@@ -322,8 +322,10 @@ ExpansionRecommendation Isochrone::ShouldExpand(baldr::GraphReader& /*graphreade
   // max but need to consider others so we just continue here. Tells MMExpand function to skip
   // updating or pushing the label back
   // prune the edge if its start is above max contour
-  if (time > max_seconds_ && dist > max_meters_)
-    return ExpansionRecommendation::prune_expansion;
+
+  ExpansionRecommendation recommendation = time > max_seconds_ && dist > max_meters_
```
nilsnolde, line 326: nit: can you put parantheses around the check? it's not breaking anything, it's mostly consistent style I guess:)

nilsnolde (approved):
Ah that makes a lot of sense! Na, it's fine without a test, we trust you did some test runs and the problem disappeared. Just a tiny nit and I'd merge.

nilsnolde (approved):

nilsnolde:
formatting is through so it's good to merge:)

nilsnolde:
good that your next PR is fixing the OSX workflow, it's a bit annoying. also no idea what's wrong with the windows workflow, it's also failing randomly with a 127 process error, which, typically win, has no description whatsoever..

PR #4605, opened 2024-02-23
Osmnode size reduction

gknisely:
# Issue

We don't want to provision a larger hard drive in AWS due to the increased size of the OSMNode structure that was added for the multi-lingual PR.  During the latest planet data conversion the process failed due to not having enough hard drive space.  This was due to the increased size of the OSMNode structure which in turn increased the OSMWayNode structure.  

Move the linguistic indexes that were added during the multi-lingual PR to a separate structure and save the data to a new sequence file that is not sorted.  Add an index to the OSMNode struct for lookup of the linguistic record.  Refactored graphbuilder and tests as needed.

fixes https://github.com/valhalla/valhalla/issues/4548

Old way_nodes.bin file size: 277248921600
New way_nodes.bin file size: 138012593488

New linguistics_node.bin file size: 4892616

## Tasklist

 - [ ] Add tests
 - [x] Add #fixes with the issue number that this PR addresses
 - [ ] Update the docs with any new request parameters or changes to behavior described
 - [x] Update the [changelog](CHANGELOG.md)
 - [ ] If you made changes to the lua files, update the [taginfo](taginfo.json) too.

## Requirements / Relations

 Link any requirements here. Other pull requests this PR is based on?

Issue #3760, opened 2022-09-25
collecting missing documentation of request options


nilsnolde:
This is a collection issue for missing documentation on request parameters. Anyone can add to this when they encounter smth that's not in the API docs.

- [ ] `prioritize_bidirectional`
- [ ] `ignore_restrictions` & friends

PR #4606, opened 2024-02-24
"soft" ignore restrictions

### LINKED ISSUE #4571: Ignore_restrictions for truck routing disregards physical real-world restrictions like maxweight and maxheight. Is it ok?

chrstnbwnkl:
# Issue

Follow-up to #4574, fixes #4571. ~~ Open to better ideas for a name. I'll list more as well if I come up with any. @nilsnolde's first take was `ignore_safe_restrictions`. ~~

Edit: sticking with `ignore_non_vehicular_restrictions` now. 

## Tasklist

 - [x] Add tests


src/sif/truckcost.cc
```diff
@@ -439,7 +439,8 @@ inline bool TruckCost::Allowed(const baldr::DirectedEdge* edge,
                                uint8_t& restriction_idx) const {
   // Check access, U-turn, and simple turn restriction.
   if (!IsAccessible(edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
-      ((pred.restrictions() & (1 << edge->localedgeidx())) && !ignore_restrictions_) ||
+      ((pred.restrictions() & (1 << edge->localedgeidx())) &&
+       (!ignore_restrictions_ && !ignore_common_restrictions_)) ||
```
nilsnolde, line 443: - also needs to be added to the reverse equivalent
- also needs to be added to all other costings (pedestrian probably not)
- a bit nitty: but we can define a `ignore_turn_restrictions` somewhere in dynamiccost which will be set to `ignore_turn_restrictions = ignore_restrictions_ || ignore_common_restrictions_` so we can avoid the `&&` right here for every single call of `Allowed`

valhalla/sif/dynamiccost.h
```diff
@@ -605,7 +605,7 @@ class DynamicCost {
         if (access_type == baldr::AccessType::kTimedAllowed)
           time_allowed = true;
 
-        if (current_time == 0) {
+        if (current_time == 0 || ignore_common_restrictions_) {
```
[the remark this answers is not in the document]
nilsnolde: actually (and this would be fixing a bug): we should add smth similar to the `Restricted()` function further up which looks at complex TRs. it could use the new internal `ignore_turn_restrictions_` and early abort

nilsnolde (changes requested):
Not sure about another name either (other than `ignore_safe_restrictions` which is also not very good), possibly only `ignore_turn_conditional_restrictions`, bit long but describes best what it does. 

There's a few more things to do, ideally they'd be tested with a parameterized test somewhere.

Also I know these kind of options weren't documented before, but we really should start documenting them. Would you be up for that? If you feel like including the other ones as well, that'd be much appreciated. There's a collection of undocumented options here: https://github.com/valhalla/valhalla/issues/3760. One after the other:)

chrstnbwnkl:
I think `ignore_turn_conditional_restrictions` does not convey that we are ignoring e.g. `maxaxles`, so in that regard `ignore_safe_restrictions` actually describes it better IMO :smile: 

Anyway just ignore everything around naming in this PR for now, it's a bit all over the place. I'll add some more tests that test some more mode specific cases where `ignore_<whatever>_restrictions` shouldn't lead to a successful route.

test/gurka/test_truck_restrictions.cc
```diff
@@ -49,6 +49,7 @@ TEST_P(TruckRestrictionTest, NotAllowed) {
   try {
     gurka::do_action(Options::route, map, {"A", "D"}, "truck",
                      {{"/costing_options/truck/" + option, v}});
+    FAIL() << "Expected no path to be found";
```
chrstnbwnkl, line 52: this is just a small fix, hazmat was actually not set properly, so there was no exception thrown

test/gurka/test_ignore_restrictions.cc
```diff
@@ -0,0 +1,178 @@
...
+                         {{"/date_time/type", "1"}, {"/date_time/value", "2020-10-10T13:00"}});
+    FAIL() << "Expected route to fail.";
+  } catch (const valhalla_exception_t& err) { EXPECT_EQ(err.code, 442); } catch (...) {
+    FAIL() << "Expected different error code.";
+  }
+
+  //...but succeed with ignore_common_restrictions
+  valhalla::Api route_succeed =
+      gurka::do_action(valhalla::Options::route, map, {"A", "F"}, costing,
+                       {{"/costing_options/" + costing + "/ignore_common_restrictions", "1"},
```
nilsnolde, line 95: the docs say smth else currently

nilsnolde (changes requested):
apart from naming: there's still costings missing.. ped not necessary, but why not the rest?

nilsnolde (commented):
looks good to me now! 

then we can focus on the naming of it. I don't have a very strong opinion, though the only option I don't really like tbh is the current `ignore_common_restrictions`.

valhalla/sif/dynamiccost.h
```diff
@@ -997,6 +1000,8 @@ class DynamicCost {
   bool shortest_;
 
   bool ignore_restrictions_{false};
+  bool ignore_non_vehicular_restrictions_{false};
+  bool ignore_turn_restrictions_{false};
```
nilsnolde, line 1004: we should add a quick note that it's not a request option, just a combination of the 2 above.

valhalla/sif/dynamiccost.h
```diff
@@ -605,7 +608,7 @@ class DynamicCost {
         if (access_type == baldr::AccessType::kTimedAllowed)
           time_allowed = true;
 
-        if (current_time == 0) {
+        if (current_time == 0 || ignore_non_vehicular_restrictions_) {
```
nilsnolde, line 611: I'd put this into the `if` on line 602, so `time_allowed` will stay false if this option is given and we don't need it in the `return` below.

proto/options.proto
```diff
@@ -312,6 +312,9 @@ message Costing {
     uint32 axle_count = 81;
     float use_lit = 82;
     bool disable_hierarchy_pruning = 83;
+    oneof has_ignore_non_vehicular_restrictions {
+      bool ignore_non_vehicular_restrictions = 84;
+    }
```
nilsnolde, line 317: no oneof

nilsnolde:
I edited the things we talked about @kevinkreiser , should be good now

nilsnolde (dismissed):

nilsnolde:
oops that broke smth

nilsnolde:
so the normal release build fails, but the arm & x86 lint-debug build work. that's weird.. they're failing with
```
2024/03/04 16:17:01.373591 [INFO] Parsing nodes...
terminate called after throwing an instance of 'std::runtime_error'
  what():  test_linguistic_node.bin(open): No such file or directory
make[3]: *** [test/CMakeFiles/run-graphparser.dir/build.make:73: test/graphparser.log] Error 134
```

does that ring a bell @gknisely? seems related to https://github.com/valhalla/valhalla/pull/4605?

gknisely:
> so the normal release build works, but the arm & x86 lint-debug build fail. that's weird.. they're failing with
> 
> ```
> 2024/03/04 16:17:01.373591 [INFO] Parsing nodes...
> terminate called after throwing an instance of 'std::runtime_error'
>   what():  test_linguistic_node.bin(open): No such file or directory
> make[3]: *** [test/CMakeFiles/run-graphparser.dir/build.make:73: test/graphparser.log] Error 134
> ```
> 
> does that ring a bell @gknisely? seems related to #4605?
> 
> the failing test also seems to be pretty unrelated to this PRs changes, so it seems flaky.

Should just be creating a temporary file.  Odd that this is failing.  Are you sure unrelated?

nilsnolde:
> Are you sure unrelated?

@gknisely at least this PR doesn't touch any of the graphbuilder logic or the test that fails. Just costings and some of those tests. So I'd say yeah, I'm very sure. Also the release build works. It's "just" the debug and arm build that fail that way. Locally it also doesn't fail.

nilsnolde:
hm yeah, now it's the exact opposite! the release build is failing with the same problem, debug & arm builds are fine.. had a quick look at the test, but can't see anything wrong. the only way I'd see this failing would be if gtest would execute tests in parallel which I don't think it does.

nilsnolde:
since the flaky test also fails on master's arm workflow currently, this PR should be good to merge.

nilsnolde (dismissed):

nilsnolde:
once it passes the master merge, I'd merge this too. it's been reviewed and amended.

nilsnolde (approved):

PR #4607, opened 2024-02-24
make shortcut logging more useful

nilsnolde:
log out
- total shortcuts
- total superseded edges
- avg amount of edges per shortcut for convenience

### DIFF
--- a/src/mjolnir/shortcutbuilder.cc
+++ b/src/mjolnir/shortcutbuilder.cc
@@ -29,6 +29,9 @@ using namespace valhalla::mjolnir;
 
 namespace {
 
+// total_shortcut_count and avg_edge_per_shortcut
+using shortcut_stats = std::pair<uint32_t, uint32_t>;
+
 // Simple structure to hold the 2 pair of directed edges at a node.
 // First edge in the pair is incoming and second is outgoing
 struct EdgePairs {
@@ -337,7 +340,7 @@ bool IsEnteringEdgeOfContractedNode(GraphReader& reader, const GraphId& nodeid, 
 }
 
 // Add shortcut edges (if they should exist) from the specified node
-uint32_t AddShortcutEdges(GraphReader& reader,
+shortcut_stats AddShortcutEdges(GraphReader& reader,
                           const graph_tile_ptr& tile,
                           GraphTileBuilder& tilebuilder,
                           const GraphId& start_node,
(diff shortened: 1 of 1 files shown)


src/mjolnir/shortcutbuilder.cc
```diff
@@ -337,18 +340,18 @@ bool IsEnteringEdgeOfContractedNode(GraphReader& reader, const GraphId& nodeid,
...
+                                const GraphId& start_node,
+                                const uint32_t edge_index,
+                                const uint32_t edge_count,
+                                std::unordered_map<uint32_t, uint32_t>& shortcuts) {
   // Shortcut edges have to start at a node that is not contracted - return if
   // this node can be contracted.
   EdgePairs edgepairs;
   if (CanContract(reader, tile, start_node, edgepairs)) {
-    return 0;
+    return {0UL, 0UL};
```
kevinkreiser, line 354: nit UL is probably 64bits wide, ie on 64 bit architectures. you could use just U I'd you wanted but if we wanted to be technically correct for all architectures you should use the exact type

src/mjolnir/shortcutbuilder.cc
```diff
@@ -359,6 +362,7 @@ uint32_t AddShortcutEdges(GraphReader& reader,
   // Iterate through directed edges of the base node
   uint32_t shortcut = 0;
   uint32_t shortcut_count = 0;
+  int32_t total_edge_count = 0;
```
kevinkreiser, line 365:
```suggestion
  uint32_t total_edge_count = 0;
```

src/mjolnir/shortcutbuilder.cc
```diff
@@ -29,6 +29,9 @@ using namespace valhalla::mjolnir;
 
 namespace {
 
+// total_shortcut_count and total_superseded_edges
+using shortcut_stats = std::pair<uint32_t, uint32_t>;
```
kevinkreiser, line 33: nit this typedef is only used for the return type. in the functions that return it and in the function that accumulates it its not used instead separate variables are used. we could just drop the typedef and return pair explicitly

kevinkreiser (commented):
looks good just a couple nits and that typo using the signed int

src/mjolnir/shortcutbuilder.cc
nilsnolde: I'll just do what the vast majority of valhalla is doing, passing ints:) what's ULL then on 64 bit archs?

nilsnolde: changed it to `return {uint32_t(0), uint32_t(0)};` instead

nilsnolde:
no need for re-approval actually, just cognitive overhead:)

PR #4608, opened 2024-02-24
contract over nodes connecting edges with differing names/speed

### LINKED ISSUE #4599: Don't contract nodes connecting edges with different names? Why?

nilsnolde:
fixes #4599 

most of the changes are concerning tests to play nicely with the new logic. the real changes are just removing a few lines.

I'll share some more insights once I got the last test to build, `astar` has an obnoxious test which is hard to adapt to the new logic..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ CMakeLists.txt.user
 .idea
 /.tidytmp
 vcpkg*/
+*.log
 
 # python
 .venv
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@
    * FIXED: fix config generator with thor.costmatrix_allow_second_pass [#4567](https://github.com/valhalla/valhalla/pull/4567)
    * FIXED: infinite loop or other random corruption in isochrones when retrieving partial shape of an edge [#4547](https://github.com/valhalla/valhalla/pull/4547)
    * FIXED: Aggregation updates: update opposing local idx after aggregating the edges, added classification check for aggregation, and shortcut length changes [#4570](https://github.com/valhalla/valhalla/pull/4570)
+   * FIXED: Fix segfault in OSRM serializer with bannerInstructions when destination is on roundabout [#4480](https://github.com/valhalla/valhalla/pull/4481)
+   * FIXED: Fix segfault in costmatrix (date_time and time zone always added). [#4530](https://github.com/valhalla/valhalla/pull/4530)
+   * FIXED: Fixed roundoff issue in Tiles Row and Col methods [#4585](https://github.com/valhalla/valhalla/pull/4585)
 * **Enhancement**
    * UPDATED: French translations, thanks to @xlqian [#4159](https://github.com/valhalla/valhalla/pull/4159)
    * CHANGED: -j flag for multithreaded executables to override mjolnir.concurrency [#4168](https://github.com/valhalla/valhalla/pull/4168)
@@ -75,7 +78,6 @@
    * UPDATED: updated country access overrides [#4460](https://github.com/valhalla/valhalla/pull/4460)
    * CHANGED: date_time refactor as a preparation to return DST/timezone related offset in the response [#4365](https://github.com/valhalla/valhalla/pull/4365)
    * ADDED: find connection on backward search for bidir matrix algo [#4329](https://github.com/valhalla/valhalla/pull/4329)
-   * FIXED: Fix segfault in OSRM serializer with bannerInstructions when destination is on roundabout [#4480](https://github.com/valhalla/valhalla/pull/4481)
    * CHANGED: Adujustment of walk speed when walking on slight downhill [#4302](https://github.com/valhalla/valhalla/pull/4302)
    * CHANGED: Do not reclassify ferry connections when no hierarchies are to be generated [#4487](https://github.com/valhalla/valhalla/pull/4487)
    * ADDED: Added a config option to sort nodes spatially during graph building [#4455](https://github.com/valhalla/valhalla/pull/4455)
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -61,6 +61,7 @@ CMakeLists.txt.user
 .idea
 /.tidytmp
 vcpkg*/
+*.log
```
nilsnolde, line 64: doing some graph build experiments and logging to .log. happens frequently, might as well put it in .gitignore

src/mjolnir/shortcutbuilder.cc
```diff
@@ -403,13 +388,6 @@ std::pair<uint32_t, uint32_t> AddShortcutEdges(GraphReader& reader,
         std::reverse(shape.begin(), shape.end());
       }
 
-      // Get names - they apply over all edges of the shortcut
-      auto names = edgeinfo.GetNames();
-      auto tagged_values = edgeinfo.GetTaggedValues();
-      auto linguistics = edgeinfo.GetLinguisticTaggedValues();
-
-      auto types = edgeinfo.GetTypes();
```
nilsnolde, line 411: those are not needed anymore either

test/gurka/test_shortcut.cc
```diff
@@ -389,13 +359,11 @@ TEST(Shortcuts, ShortcutRestrictions) {
   baldr::GraphReader reader(map.config.get_child("mjolnir"));
 
   // test we got the right shortcuts edges, implicitly means they were broken properly
-  for (const auto& end_node : {"C", "J", "L"}) {
-    const auto shortcut =
-        gurka::findEdge(reader, layout, "highway", end_node, baldr::GraphId{}, 0, true);
+  for (const auto& pair :
+       node_pairs{{"A", "C"}, {"C", "A"}, {"H", "J"}, {"J", "H"}, {"J", "L"}, {"L", "J"}}) {
+    const auto shortcut = gurka::findEdgeByNodes(reader, layout, pair.first, pair.second);
```
nilsnolde, line 364: the `findEdges` is tripping now that we don't keep the name for the shortcut anymore. in the end actually _could_ keep it, it has no adverse effects other than a few more lines of redundant code (which is why I removed it)

test/gurka/test_shortcut.cc
```diff
@@ -67,38 +67,6 @@ TEST(Shortcuts, LoopWithoutShortcut) {
   EXPECT_FALSE(shortcut.Is_Valid()) << "Shortcuts found. Check the map.";
 }
 
-// Here no shortcuts are created. There could be one from A to C with speed 80 but in the opposite
```
nilsnolde, line 70: this is entirely redundant now

test/astar.cc
```diff
@@ -1571,7 +1571,7 @@ TEST(BiDiAstar, test_recost_path) {
   )";
   const gurka::ways ways = {
       // make ABC to be a shortcut
-      {"ABC", {{"highway", "primary"}, {"maxspeed", "80"}}},
+      {"ABC", {{"highway", "motorway"}, {"maxspeed", "80"}}},
```
nilsnolde, line 1574: this is needed so we still create separate shortcuts

CHANGELOG.md
```diff
@@ -36,6 +36,9 @@
    * FIXED: fix config generator with thor.costmatrix_allow_second_pass [#4567](https://github.com/valhalla/valhalla/pull/4567)
    * FIXED: infinite loop or other random corruption in isochrones when retrieving partial shape of an edge [#4547](https://github.com/valhalla/valhalla/pull/4547)
    * FIXED: Aggregation updates: update opposing local idx after aggregating the edges, added classification check for aggregation, and shortcut length changes [#4570](https://github.com/valhalla/valhalla/pull/4570)
+   * FIXED: Fix segfault in OSRM serializer with bannerInstructions when destination is on roundabout [#4480](https://github.com/valhalla/valhalla/pull/4481)
```
nilsnolde, line 39: those were misplaced in the "Enhancement" section. I didn't make the effort to align them properly time-wise, still better than before

src/mjolnir/restrictionbuilder.cc
```diff
@@ -270,8 +270,8 @@ ComplexRestrictionBuilder CreateComplexRestriction(const OSMRestriction& restric
 };
 
 struct Result {
-  uint32_t forward_restrictions_count;
-  uint32_t reverse_restrictions_count;
+  uint32_t forward_restrictions_count = 0;
```
nilsnolde, line 273: this finally fixes an issue where we report ridiculous numbers for restrictions, like 2 billion restrictions on level 2

test/gurka/test_avoids.cc
```diff
@@ -1,4 +1,5 @@
 #include "gurka.h"
+#include <boost/format.hpp>
```
nilsnolde, line 2: fixes osx ci

nilsnolde:
this is ready for review now

nilsnolde:
next week more..

nilsnolde:
there's some minor improvement for `de_benchmark_routes.txt` with 115 vs 108 secs for the bunch on a persistent server, so like 6-7% improvement. spot checking 15-20 route diffs there was no difference in narration or time/distance.

performance improvement will have an exponential relationship with overall distance, but since bidir a* is pretty directed, it doesn't show as much for regional routes. I'd expect the p90 routes to have a clearer improvement, e.g. moscow -> madrid or so. diffs for (dijkstra) costmatrix are a lot more pronounced, 20 x 20 random locations in germany over 50 requests show 200 secs (branch) vs 260 secs (master).

here the shortcut status for germany-latest:
| features (accumulative) |     level 0     |                  |                       |     level 1     |                  |                       |
|-------------------------|:---------------:|:----------------:|:---------------------:|:---------------:|:----------------:|:---------------------:|
|                         | total shortcuts | total superseded | avg edges per shorcut | total shortcuts | total superseded | avg edges per shorcut |
| master                  | 161872          | 849550           | 6                     | 623364          | 3863670          | 5                     |
| branch                  | 107146          | 889932           | 8                     | 359422          | 359422           | 11                    |

resulting in a minor size reduction of 15 MB for level tiles (411 MB total) and 4 MB for level 0 tiles (113 MB total).

I think this particular one doesn't really need a full planet RAD test yet. It's pretty straight forward and it's fairly easy to see that it won't have any impact on expansion (other than performance) or narration. however, next ones are more sensitive and there I'd run a full planet RAD.

what do you think @gknisely @kevinkreiser @dnesbitt61 ?

test/astar.cc
```diff
@@ -1679,6 +1679,81 @@ TEST(BiDiAstar, test_recost_path) {
   }
 }
 
+// TODO(nils): this test fails currently, because bidir A* has a problem with 2 shortcuts between the
+// same nodes: https://github.com/valhalla/valhalla/issues/4609
```
nilsnolde, line 1683: this disabled test is the reason why the reported coverage is so low BTW.

also, good to mention: if I run this on master (with the appropriate edits to make sure it's 2 shortcuts with different superseded edges between A and E), it fails too.

Issue #3212, opened 2021-07-14
Inexplicable artefacts in isochrones


mesozoic-drones:
Hello! I've noticed that in multiple cases valhalla generates rather strange isochrones with artefacts. Is this a bug or do I need to tune API parameters in some way?

Here are some example screenshots. 

The most eye-catching case. Artefacts on water:
<img width="500" alt="Screenshot 2021-07-12 at 15 47 58" src="https://user-images.githubusercontent.com/54934129/125652219-955be151-6f0a-4082-bb29-07503dfbdd0b.png">

Another example of artefacts on water:
<img width="500" alt="Screenshot 2021-07-12 at 15 48 24" src="https://user-images.githubusercontent.com/54934129/125652553-69c99f04-b6e5-4051-8dea-5328472abc32.png">

Artefacts on park:
<img width="500" alt="Screenshot 2021-07-12 at 19 43 10" src="https://user-images.githubusercontent.com/54934129/125652657-bc616ee7-66ef-4fb2-a8a0-88667a78e952.png">

Artefact near building:
<img width="500" alt="Screenshot 2021-07-12 at 15 47 13" src="https://user-images.githubusercontent.com/54934129/125652906-42ff228c-8964-4e4c-a841-a821989281d3.png">

Additional information:
- Valhalla uses tiles with the same OSM planet version as the vector tiles on map. 
- I double-checked that there are no OSM ways (like piers, paths etc) that are not shown on map but present in valhalla tiles.
- "generalize" is set to 2.5 meters. **If I use the default, artefacts become more simplified but are still present.**
- All isochrones are built in pedestrian mode.

Thank you in advance!

kevinkreiser:
@mesozoic-drones the way it works underneath is that we make what is essentially a raster image and then we generate vector contours from that rasterization. that step of taking something discretized and adding precision to it often creates issues like the one in your last image where there is a cusp near where some geometry is, basically the center of the pixel that got marked by that little piece of road is not right on the road and so when it gets turned into vector it offsets the vector contour.

the other artifacts are basically that same process but occur where you have holes or where the gradient between pixels values is really high. when the gradient is high we need to interpolate vector lines in a tight space and the way numerical precision works out we get those values that we get :smile: again though the reason the vectors that come out arent smooth is because the pixels that are set in the raster have a jagged edge. basically the underlying data looks like this:

[...]

PR #4610, opened 2024-02-25
Expose isochrone grid size via API

akadouri:
# Issue

As disused in https://github.com/valhalla/valhalla/issues/3212 pedestrian isochrones could improve their boundaries with a smaller "grid_size."

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/worker.cc
```diff
@@ -1130,6 +1130,14 @@ void from_json(rapidjson::Document& doc, Options::Action action, Api& api) {
   // if specified, get the show_locations boolean in there
   options.set_show_locations(rapidjson::get<bool>(doc, "/show_locations", options.show_locations()));
 
+  // if specified, get the isochrone_grid_size in there
+  auto isochrone_grid_size = rapidjson::get_optional<float>(doc, "/isochrone_grid_size");
```
akadouri, line 1134: Should I `throw valhalla_exception_t{#}` if the value is invalid?

docs/docs/api/isochrone/api-reference.md
```diff
@@ -47,7 +47,8 @@ The isochrone service uses the `auto`, `bicycle`, `pedestrian`, and `multimodal`
...
 | `generalize` | A floating point value in meters used as the tolerance for [Douglas-Peucker](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm) generalization. Note: Generalization of contours can lead to self-intersections, as well as intersections of adjacent contours. |
-| `show_locations` | A boolean indicating whether the input locations should be returned as MultiPoint features: one feature for the exact input coordinates and one feature for the coordinates of the network node it snapped to. Default false. 
+| `show_locations` | A boolean indicating whether the input locations should be returned as MultiPoint features: one feature for the exact input coordinates and one feature for the coordinates of the network node it snapped to. Default false. |
+| `isochrone_grid_size` | A floating point value used to determine the grid size over which the isochrone is calculated.
```
akadouri, line 51: Not sure how to describe the unit. "Length in degrees at location"?

src/thor/isochrone.cc
```diff
@@ -125,17 +125,20 @@ void Isochrone::ConstructIsoTile(const bool multimodal,
...
-  float grid_size = multimodal ? dlat / 500.0f : dlat / 300.0f;
-  if (grid_size < 0.001f) {
-    grid_size = 0.001f;
-  } else if (grid_size > 0.005f) {
-    grid_size = 0.005f;
-  } else {
-    // Round to nearest 0.001
-    int r = std::round(grid_size * 1000.0f);
-    grid_size = static_cast<float>(r) * 0.001f;
+  float grid_size = api.options().isochrone_grid_size();
```
nilsnolde, line 128: Hm I’ll need to read up a bit on what my past self thought in that issue:)

Intuitively I’m not sure I like an (unbound) request option for such a pretty sensitive parameter. I’d rather like us pick a (much!) better value automatically. It’s not easy to document this parameter bcs it’s hard to convey what it means. When that’s the case I’m usually more in favor of hiding those kind of details. But of course (adjusting) the resolution of the grid is smth a more keen user might be interested in. 

Hm, at the very least I’d make it bound with a lower limit (relative to the costing speed!), so to not screw up hosted servers. Maybe even including a way to opt out of it in the config generator? Maybe we should restrict total grid size instead? To make sure a user can’t ask for 200 km car isochrone with the resolution of low range ped..

Other opinions?

nilsnolde: Read up on it a bit, and I still agree with what I said there. It should be fairly easy to hide this kind of thing and provide a better estimate of the resolution automatically. A request option will not make sense intuitively, even to me, it’s just hard to envision how to scale this by hand. Then it’ll be a lot of trying around to micro-optimize that setting.

What do you think @akadouri? Would you be up for investigating an improved implicit approach?

proto/options.proto
```diff
@@ -489,4 +489,7 @@ message Options {
   bool banner_instructions = 55;                                   // Whether to return bannerInstructions in the OSRM serializer response
   float elevation_interval = 56;                                   // Interval for sampling elevation along the route path.  [default = 0.0];
   bool voice_instructions = 57;                                    // Whether to return voiceInstructions in the OSRM serializer response
+  oneof has_isochrone_grid_size {
+    float isochrone_grid_size = 58;                                // Grid size to calculate isochrone
```
kevinkreiser, line 493: no need for `oneof` the default value of zero as a sentinel would be fine to note that the person didnt request it

src/thor/isochrone.cc
kevinkreiser: i am also uneasey about exposing this. with a really high resolution it could significantly increase the time required to mark all the pixels and then to serialize the result. 

if the primary use case is further smoothing of the results we have a couple options:

1. use the new geotiff raster output. double triple or whatever the resolution of the geotif and then apply raster based smoothing on it (eg gaussian). you'd then have to use something like gdal to contour the image to get the vectors out of it thought
2. implement other types of "simplification/generalization". right now we have the RDP algorithm implemented but there are a ton of simplification algrithms that smooth the resulting geometry, grass has a nice collection of them: https://grass.osgeo.org/grass83/manuals/v.generalize.html

PR #4613, opened 2024-02-27
contract edges with differing non-conditional access restrictions

nilsnolde:
based on #4608 

if 2 edges have differing conditional access restrictions, this PR will **not** contract the node. this is done because kTimeAllowed would be a bit more work to consolidate onto shortcuts, we'd have to only include _overlapping_ time periods of the base edges. certainly possible, I left a TODO, just wanted to keep this PR lean. also conditionals are (at least in OSM) sufficiently rare to hardly have an impact.

if 2 edges have other differing access restrictions (maxheight/weight/width etc), we'll contract that node and keep the most restrictive attribute of all base edges for the entire shortcut.

I think this should work fine, but I do owe some tests. also some RAD tests. personally I'd still be happy in this case to only RAD test `de_benchmark_routes.txt`, which together with gurka tests, should prove that it works the way I'm thinking it does.


nilsnolde:
Updating the table of https://github.com/valhalla/valhalla/pull/4608#issuecomment-1964801663:

| features (accumulative) |     level 0     |                  |                       |     level 1     |                  |                       |
|-------------------------|:---------------:|:----------------:|:---------------------:|:---------------:|:----------------:|:---------------------:|
|                         | total shortcuts | total superseded | avg edges per shorcut | total shortcuts | total superseded | avg edges per shorcut |
| master                  | 161872          | 849550           | 6                     | 623364          | 3863670          | 5                     |
| names/speed      | 107146          | 889932           | 8                     | 359422          | 3976950           | 11                    |
| restrictions          | 107792          | 906868           | 8                     | 349522          | 3985508           | 11                    |

nilsnolde:
I think there's a problem with opposing edges of shortcuts. If I read the current code correctly, then we use the last base edge's `opp_local_idx` for the new shortcut, but if the node of that last base edge will have a shortcut at some point later, then the `opp_local_idx` should be all wrong no? There's other questions I have regarding this, especially in the light of this PR. We can discuss on Monday.

src/mjolnir/shortcutbuilder.cc
```diff
@@ -29,6 +29,50 @@ using namespace valhalla::mjolnir;
...
+  //   kTimeDenied is fine to just append all restrictions of the base edges, but kTimeAllowed
+  //   will be harder, there we'll have to merge overlapping time periods
+  void update_nonconditional(const std::vector<AccessRestriction>&& other_restrictions) {
+    for (const auto& new_ar : other_restrictions) {
+      // update the modes for the edge attribute regardless
+      if (new_ar.type() == AccessType::kTimedAllowed || new_ar.type() == AccessType::kTimedDenied ||
+          new_ar.type() == AccessType::kDestinationAllowed) {
+        continue;
+      }
+      modes |= new_ar.modes();
```
nilsnolde, line 55: did this small change to not record modes of restrictions we don't care about

nilsnolde:
no rush here, I wanted to understand better how shorcut's opposites are generated

nilsnolde:
got a minute to talk with @kevinkreiser about this. and yeah, there's some problems with how we build shortcuts today (and same in this PR), see #4656 .

that's for another PR though, as we're not breaking anything here, so all good for final review.

src/mjolnir/restrictionbuilder.cc
```diff
@@ -270,8 +270,8 @@ ComplexRestrictionBuilder CreateComplexRestriction(const OSMRestriction& restric
 };
 
 struct Result {
-  uint32_t forward_restrictions_count;
-  uint32_t reverse_restrictions_count;
+  uint32_t forward_restrictions_count = 0;
+  uint32_t reverse_restrictions_count = 0;
```
nilsnolde, line 274: need to initialize it, further down we only add to it. resulted in bogus restriction counts for the first level

PR #4620, opened 2024-03-05
Clear state after time domain push

molind:
# Issue

For string with two or more time intervals e.g. `Mo-Sa 09:00-21:00; Su 10:00-20:00` current version returns second interval for all days of week.

After push we need to reset timedomain.

What do you think?

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/mjolnir/timeparsing.cc
+++ b/src/mjolnir/timeparsing.cc
@@ -320,6 +320,7 @@ std::vector<uint64_t> get_time_range(const std::string& str) {
             timedomain.set_end_mins(min);
 
             time_domains.push_back(timedomain.td_value());
+            timedomain = TimeDomain();
           }
         }
         return time_domains;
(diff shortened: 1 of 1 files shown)


kevinkreiser:
@gknisely this should be testable with some gurka right?

molind:
Also, I've noticed that `24/7` isn't parsed at all. Maybe that makes sense, because in Valhalla, this function is used to parse data when the road is closed.

gknisely:
> DOW for `09:00-21:00` is empty. Is that expected behaviour?

Yes.  Means the restriction is in place everyday for the time noted above

gknisely:
> Also, I've noticed that `24/7` isn't parsed at all. Maybe that makes sense, because in Valhalla, this function is used to parse data when the road is closed.

Yes.  People usually just add `access=no`

gknisely:
The old tests are here https://github.com/valhalla/valhalla/blob/master/test/timeparsing.cc but should be able to add a gurka test too

PR #4625, opened 2024-03-11
Conflict with signinfo's temporary linguistic node sequence file caused test failures. 

gknisely:
# Issue

graphparser test is randomly failing.   Ref: https://github.com/valhalla/valhalla/pull/4606#issuecomment-1977634974

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/test/graphparser.cc
+++ b/test/graphparser.cc
@@ -65,30 +65,11 @@ void DoConfig() {
     file << "{ \
       \"mjolnir\": { \
       \"id_table_size\": 1000, \
-      \"tile_dir\": \"test/data/parser_tiles\" \
+      \"tile_dir\": \"test/data/graph_parser_tiles\" \
       } \
     }";
   } catch (...) {}
   file.close();
-
-  boost::property_tree::ptree conf;
-  rapidjson::read_json(config_file, conf);
-
-  // Get the tile directory (make sure it ends with the preferred separator
-  std::string tile_dir = conf.get<std::string>("mjolnir.tile_dir");
-  if (tile_dir.back() != filesystem::path::preferred_separator) {
-    tile_dir.push_back(filesystem::path::preferred_separator);
-  }
-
-  ways_file = tile_dir + ways_file;
-  way_nodes_file = tile_dir + way_nodes_file;
-  nodes_file = tile_dir + nodes_file;
-  edges_file = tile_dir + edges_file;
-  access_file = tile_dir + access_file;
-  from_restriction_file = tile_dir + from_restriction_file;
-  to_restriction_file = tile_dir + to_restriction_file;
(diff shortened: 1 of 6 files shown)


test/graphparser.cc
```diff
@@ -677,7 +678,7 @@ void DoConfig() {
     file << "{ \
       \"mjolnir\": { \
       \"id_table_size\": 1000, \
-      \"tile_dir\": \"test/data/parser_tiles\" \
```
kevinkreiser, line 680: utrecht was using this? at this point its happened to us so many times that we had directory collisions we might try having the code generate a random suffix for the directories we create :smile:

kevinkreiser (approved):

test/graphparser.cc
gknisely: yes.  https://github.com/valhalla/valhalla/blob/master/test/utrecht.cc#L55

gknisely: multiple locations

gknisely:
Still having issues.  Looking at it

test/graphparser.cc
nilsnolde: > we might try having the code generate a random suffix for the directories we create

I’d be in favor of that. As long as we print it out it’d be a good time investment. Would fit well in your other test cleanup PR:)

gknisely: Determined that this was not the cause of the failures

test/signinfo.cc
```diff
@@ -30,7 +30,7 @@ TEST(Signinfo, ExitToTest) {
   std::vector<SignInfo> signs;
   std::vector<std::string> linguistics;
   std::vector<std::pair<std::string, bool>> default_languages;
-  const std::string linguistic_node_file = "test_linguistic_node.bin";
+  const std::string linguistic_node_file = "test_sign_linguistic_node.bin";
```
gknisely, line 33: This file was the cause of the problem.  In the multi-threaded environment the `test_linguistic_node.bin` was getting deleted for the `graphparser.cc` test.

dnesbitt61 (dismissed):

dnesbitt61 (approved):

test/signinfo.cc
nilsnolde: so this test also builds to the same directory as `graphparser.cc`? if so, shouldn't we change that?

gknisely: Most are all running in the same directory.  This test is testing specifically signs and is not building data.  This is a temporary file as it is required for `CreateSignInfoList`  We are not testing linguistic node signage here as it is done in the gurka tests.  The real solution is to migrate the old tests to gurka.

Issue #4538, opened 2024-01-30
costmatrix algorithm do not respect oneway on trivial routes


radekvermirovsky:
If we use this request to route endpoint:

```json
{
	"locations": [
        {
		"lat": 50.22878,
		"lon": 15.83895
	},
        {
		"lat": 50.22888,
		"lon": 15.83918
	}
	],
       "verbose":false,
	"directions_type": "none",
	"costing": "auto",
	"prioritize_bidirectional": true
}
```
oneway is respected properly. 
Foggis
[https://valhalla.openstreetmap.de/directions?profile=car&wps=15.83895%2C50.22878%2C15.83918%2C50.22888](url)
Polyline
[https://valhalla.github.io/demos/polyline/?unescape=true&polyline6=true#qyux~Aguve%5DpUhx%40x%60%40%60uAjTiStQmQvQoQim%40smBu%40iC%7DSgu%40ox%40ju%40lDzL](url)
Screenshot:
<img width="832" alt="image" src="https://github.com/valhalla/valhalla/assets/149080356/e81c7e6f-ed06-439a-8334-e460b1e78130">

If we try find same route for matrix, using costmatrix algorithm:
```json
{
	"sources": [
		{
			"lat": 50.22878,
			"lon": 15.83895
		}
	],
	"targets": [
		{
			"lat": 50.22888,
			"lon": 15.83918
		}
	],
	"verbose": false,
        "shape_format":"polyline6",
	"costing": "auto",
	"prioritize_bidirectional": true
}
```
response is:

```json
{
    "algorithm": "costmatrix",
    "units": "kilometers",
    "sources_to_targets": {
        "shapes": [
            [
                "cyux~Aquve]wDyM"
            ]
        ],
        "durations": [
            [
                2
            ]
        ],
        "distances": [
            [
                0.020
            ]
        ]
    }
}
```
Polyline:
[https://valhalla.github.io/demos/polyline/?unescape=true&polyline6=true#cyux~Aquve%5DwDyM](url)
Screenshot:
<img width="930" alt="image" src="https://github.com/valhalla/valhalla/assets/149080356/dad0c1ef-9c8f-4e6e-99a4-1cadbf78053a">

nilsnolde:
Actually I'll keep this open and close the other one, this has a test case

radekvermirovsky:
[...]

PR #4626, opened 2024-03-12
fix CostMatrix for trivial routes with oneways

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4538


src/thor/costmatrix.cc
```diff
@@ -252,9 +253,10 @@ bool CostMatrix::SourceToTarget(Api& request,
       throw valhalla_exception_t{430};
     }
     // Allow this process to be aborted
-    if (interrupt_ && (n++ % kInterruptIterationsInterval) == 0) {
+    if (interrupt_ && (interrupt_n++ % kInterruptIterationsInterval) == 0) {
```
nilsnolde, line 256: needed to come up with `interrupt_n`: `n` is used to abort the expansion (among other things), but it was never incremented unless there was an interrupt. i.e. matrix is vulnerable to home-cooked DDOS right now.. this should fix it now.

src/thor/costmatrix.cc
```diff
@@ -726,26 +728,26 @@ bool CostMatrix::Expand(const uint32_t index,
 // Check if the edge on the forward search connects to a reached edge
 // on the reverse search trees.
 void CostMatrix::CheckForwardConnections(const uint32_t source,
-                                         const BDEdgeLabel& pred,
+                                         const BDEdgeLabel& fwd_pred,
```
nilsnolde, line 731: I did tons of renaming to make it more undestandable

src/thor/costmatrix.cc
```diff
@@ -766,42 +768,48 @@ void CostMatrix::CheckForwardConnections(const uint32_t source,
...
 
     // Special case - common edge for source and target are both initial edges
-    if (pred.predecessor() == kInvalidLabel && opp_predidx == kInvalidLabel) {
-      // TODO: shouldnt this use seconds? why is this using cost!?
-      float s = std::abs(pred.cost().secs + opp_el.cost().secs - opp_el.transition_cost().cost);
+    if (fwd_pred.predecessor() == kInvalidLabel && rev_predidx == kInvalidLabel) {
+      // bail if either edge wasn't allowed (see notes in SetSources/Targets)
+      if (!fwd_pred.path_id() || !rev_label.path_id()) {
+        return;
+      }
```
nilsnolde, line 782: this is the actual change: `path_id()` is not used in CostMatrix of course, so I use it in `SetSources/Targets` to record whether the edge we want to connect on is actually traversable in both directions.

src/thor/costmatrix.cc
```diff
@@ -766,42 +768,48 @@ void CostMatrix::CheckForwardConnections(const uint32_t source,
...
+
+      // remember: transition_cost is abused in SetSources/Targets: cost is secs, secs is length
+      float s =
+          std::abs(fwd_pred.cost().secs + rev_label.cost().secs - rev_label.transition_cost().cost);
```
nilsnolde, line 786: also there was a TODO here before asking why this is using cost and seconds. I removed that and instead added the note above.

src/thor/costmatrix.cc
```diff
@@ -1064,11 +1081,12 @@ void CostMatrix::SetTargets(baldr::GraphReader& graphreader,
       const DirectedEdge* directededge = tile->directededge(edgeid);
 
       // Get the opposing directed edge, continue if we cannot get it
-      GraphId opp_edge_id = graphreader.GetOpposingEdgeId(edgeid);
+      graph_tile_ptr opp_tile = tile;
+      GraphId opp_edge_id = graphreader.GetOpposingEdgeId(edgeid, opp_tile);
```
nilsnolde, line 1085: in `SetTargets`, we need the opposing tile for the Allowed() method below

test/test.h
```diff
@@ -23,6 +23,7 @@
 #include <gtest/gtest.h>
 
 #include <boost/algorithm/string/replace.hpp>
+#include <boost/format.hpp>
```
nilsnolde, line 26: OSX was again failing with some gurka tests not including boost/format.hpp.. let's just do it here and forget about it..

src/thor/costmatrix.cc
kevinkreiser: good catch! just a note: its still vulnerable (and all other endpoints) in the default configuration as well since we dont set a default timeout for the requests in the default config (i still think this is fine though and that we should not prescribe a timeout value)

kevinkreiser (approved):
totally cool to do renames/refactors but it would be much more convenient for review to separate the too though a pain in the butt i totally get it. anyway i looked at the two spots where you set/check the cached `allowed` value via the path_id overload. i guess re-evaluating `allowed` wouldn't be a good option since it requires getting the edges again which we dont have at that spot in the code. we could just keep another vector of this info and look it up in that. we do a moderate amount of overloading of things in the databuilding and its always confusing but i get why you did it, its certainly the most convenient place.

other than that i looked at one tiny bit of refactoring to use the cached graphtileptr in the opposing functions (because this is a change and because its often a place where ive made mistakes). anyway that looked good but i didnt look at any of the other refactoring because it was too much :smile:

PR #4628, opened 2024-03-13
use a singleton pattern to register the gdal driver once in one place

### LINKED ISSUE #4623: GeoTIFF segfaulting on Mac ARM python bindings
Thought I'd try out this new GeoTIFF feature on Mac ARM.

I'm running it through python bindings so I can't get a lot of debugging info.

Has anyone successfully ran this on an ARM build of any kind?

I'm doing regular isochrone request (which works), with "format":"geotiff" added

kevinkreiser:
fixes #4623 (hopefully)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ option(ENABLE_SINGLE_FILES_WERROR "Convert compiler warnings to errors for singl
 option(PREFER_EXTERNAL_DEPS "Whether to use internally vendored headers or find the equivalent external package" OFF)
 # useful to workaround issues likes this https://stackoverflow.com/questions/24078873/cmake-generated-xcode-project-wont-compile
 option(ENABLE_STATIC_LIBRARY_MODULES "If ON builds Valhalla modules as STATIC library targets" OFF)
-option(ENABLE_GDAL "Whether to include GDAL; currently only used for raster serialization of isotile grid" OFF)
+option(ENABLE_GDAL "Whether to include GDAL; currently only used for raster serialization of isotile grid" ON)
 
 set(LOGGING_LEVEL "" CACHE STRING "Logging level, default is INFO")
 set_property(CACHE LOGGING_LEVEL PROPERTY STRINGS "NONE;ALL;ERROR;WARN;INFO;DEBUG;TRACE")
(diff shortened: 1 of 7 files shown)


src/tyr/isochrone_serializer.cc
```diff
@@ -169,7 +186,7 @@ std::string serializeGeoTIFF(Api& request, const std::shared_ptr<const GriddedDa
   char** geotiff_options = NULL;
   geotiff_options = CSLSetNameValue(geotiff_options, "COMPRESS", "PACKBITS");
 
-  // TODO: check if there's an easy way to only instantiate the driver once
+  gdal_singleton_t::get();
```
kevinkreiser, line 189: this first time this is called (per process) the constructor above is called and the driver registration happens. subsequent calls are no-ops

CMakeLists.txt
```diff
@@ -191,12 +191,15 @@ else()
   message(FATAL_ERROR "Required target protobuf::libprotobuf-lite or protobuf::libprotobuf is not defined")
 endif()
 
-# gdal
-set(gdal_target "")
+# unless you said you didnt want gdal we try to turn it on, if we cant we tell you
```
kevinkreiser, line 194: i thought it would be better to default turning it ON if we can and if we couldnt we simply report that was the case. if you tell it you definitely dont want it by setting it to OFF then we dont even try

nilsnolde (commented):
Great, thanks. Just one question about GDAL=OFF @kevinkreiser . Could even do one CI build without, but no need to either IMO

CMakeLists.txt
kevinkreiser: yep will do that real quick

nilsnolde (approved):

src/tyr/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ set(system_includes
   ${date_include_dir}
   $<$<BOOL:${WIN32}>:${dirent_include_dir}>)
 
-
+set(tyr_dependencies )
```
nilsnolde, line 23: I guess this is accidental @kevinkreiser ?

nilsnolde (changes requested):
I think I approved from my phone, just seeing one nitty thing. also, can you remove the gtiff driver loading thing in the isochrone.cc? once that's gone, we know this PR works for the actor.

kevinkreiser:
> can you remove the gtiff driver loading thing in the isochrone.cc test

ha! yeah i grepped the `include` and `src` directories but forgot to grep `test`!

nilsnolde:
Oh jeez I’m really sorry, lousy review: I forgot about the docs in docs/docs/building.md, they still say default is OFF. 

Anyways no worries if you don’t feel like it

test/isochrone.cc
```diff
@@ -355,7 +355,6 @@ void check_raster_edges(size_t x, size_t y, uint16_t* data) {
 }
 
 TEST(Isochrones, test_geotiff_output_distance) {
-  GDALRegister_GTiff();
```
kevinkreiser, line 358: i dont really like removing these because the tests themselves use gdal directly to load the response from the library. now in this case technically we can remove them because hidden in the source of the serializer we are doing this initialization but the tests (and just in general any calling code) shouldnt know/assume that its preconditions are met by something it calls into. the only way to get around this would be to expose gdal initialization as a public api in the library and have the tests call that but i hate that idea even worse.

nilsnolde: Ah I didn’t check that.. I kinda assumed those driver inits were there to make the test work (it wouldn’t have without). That’s true.. anyways, the alternative to test this bit is indeed worse.

nilsnolde (approved):

test/isochrone.cc
kevinkreiser: the other idea is to leave them there though what i would do is in the suite initialization call it once

nilsnolde: Both is fine with me tbh, I get your point, but also think it’ll be fine probably forever ®

PR #4629, opened 2024-03-13
add libgdal to final docker image

nilsnolde:
currently the docker build is failing in master when compiling the python bindings. that's pointing to a much bigger problem, where `ENABLE_GDAL=ON` would currently not even be able to run `valhalla_service`, because `libgdal32` wasn't installed in the final docker image, just in the building intermediate docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,7 +73,7 @@ COPY --from=builder /usr/lib/python3/dist-packages/valhalla/* /usr/lib/python3/d
 # install all the posix locales that we support
 RUN export DEBIAN_FRONTEND=noninteractive && apt update && \
     apt install -y \
-      libcurl4 libczmq4 libluajit-5.1-2 \
+      libcurl4 libczmq4 libluajit-5.1-2 libgdal32 \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -73,7 +73,7 @@ COPY --from=builder /usr/lib/python3/dist-packages/valhalla/* /usr/lib/python3/d
 # install all the posix locales that we support
 RUN export DEBIAN_FRONTEND=noninteractive && apt update && \
     apt install -y \
-      libcurl4 libczmq4 libluajit-5.1-2 \
+      libcurl4 libczmq4 libluajit-5.1-2 libgdal32 \
```
kevinkreiser, line 76: I would have thought it'd be:
```suggestion
      libcurl4 libczmq4 libluajit-5.1-2 libgdal-dev \
```

kevinkreiser: or oh I guess it's the runner so only need the lib but why 32? is it a version suffix?

kevinkreiser: ok yeah must be the version. unfortunate naming!

nilsnolde: Yeah it’s like “libcurl4”, some (to me) strange sort of package versioning

Issue #4609, opened 2024-02-25
bidir A* has a problem with shortcuts with same start & end node


nilsnolde:
I've discovered this during the possibly only test that properly tests shortcuts end to end in `test/astar.cc`. with this map where `ABCDE` is faster than `AXYE`, we still take `AXYE`:

```
           X-----------Y
          /             \
    1----A               E---2
          \             /
           B--C--------D
```

I created a failing test which is currently disabled: https://github.com/valhalla/valhalla/blob/34a705465226aa3135d9cb9d62ef48598bdec7bc/test/astar.cc#L1684.

it seems very similar to the issue with trivial routes and our bidir algorithms.

nilsnolde:
This should really be a sufficiently rare problem. I’d rather do the low effort thing and log out if that happens. If we see many of those in a planet build, we can prioritize maybe.

nilsnolde:
see https://github.com/valhalla/valhalla/pull/4632#issuecomment-2017701669.

it's not rare at all!

nilsnolde:
there's no impact on performance at all @radekvermirovsky .

costmatrix might be affected, I didn't look at that.

PR #4632, opened 2024-03-14
log shortcuts which have the same start/end node

nilsnolde:
relates to https://github.com/valhalla/valhalla/issues/4609, but doesn't fix it. I'd first like to see if this is even happening anywhere (though I'm sure it does) and how often.


gknisely (approved):

nilsnolde:
ok, so I ran a planet with this and have only very conservative numbers of min 50k shortcuts having the same start/end nodes, i.e. problem for bidir a* (costmatrix?). it logged so much, the still available log rotation only captured a few mins of processing. anyways, it's like a quite a few percent of shortcuts where that happens.

so we should really fix the issue. I'll close here.

PR #4633, opened 2024-03-15
fixed CostMatrix shapes for routes against trivial oneways

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4538#issuecomment-1999524920

### DIFF
--- a/src/thor/costmatrix.cc
+++ b/src/thor/costmatrix.cc
@@ -1243,9 +1243,10 @@ std::string CostMatrix::RecostFormPath(GraphReader& graphreader,
   auto source_vertex = PointLL{source_edge.ll().lng(), source_edge.ll().lat()};
   auto target_vertex = PointLL{target_edge.ll().lng(), target_edge.ll().lat()};
   std::vector<PointLL> points;
-  for (const auto& path_edge : path_edges) {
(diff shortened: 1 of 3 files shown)


src/thor/costmatrix.cc
```diff
@@ -1243,9 +1243,10 @@ std::string CostMatrix::RecostFormPath(GraphReader& graphreader,
   auto source_vertex = PointLL{source_edge.ll().lng(), source_edge.ll().lat()};
   auto target_vertex = PointLL{target_edge.ll().lng(), target_edge.ll().lat()};
   std::vector<PointLL> points;
-  for (const auto& path_edge : path_edges) {
-    auto is_first_edge = path_edge == path_edges.front();
-    auto is_last_edge = path_edge == path_edges.back();
```
nilsnolde, line 1248: for the case of seemingly trivial routes which go against the oneway, both of these are true and then further down it'd trim the shape according to a trivial route which is not doing the right thing to do. we have to do this index based to really tell if it's the first/last edge.

kevinkreiser: yeah these needed to be pointer comparison not equality operator

kevinkreiser (approved):

Issue #4433, opened 2023-12-04
CostMatrix fails on trivial routes


nilsnolde:
On trivial routes, the CostMatrix fails to connect on the correct edge. It always chooses the wrong direction it seems. Not sure yet if that's consistent, but seems so.

Found when working on #4432

nilsnolde:
closing in favor of #4538

nilsnolde:
this is referenced in the code and I realized I closed this prematurely. it has nothing to do with oneways, it's just confused about which directed edge is the right one. however, that's also affecting the results, it'll report the cost/time of the wrong edge, see https://github.com/valhalla/valhalla/pull/4634

chrstnbwnkl:
I had a quick look, but haven't gotten very far. I just noticed that it doesn't _always_ connect on the wrong edge. In the example test in #4634 if you switch source and target, you'll see that it actually connects via the backward edge.

chrstnbwnkl:
Fixed in #4634 and #5001

PR #4634, opened 2024-03-15
fix trivial route for CostMatrix

nilsnolde:
showing https://github.com/valhalla/valhalla/issues/4433. it's not fixed yet, just a test written to showcase the problem.

I recently did a band-aid for this, just so we get the right shape out of trivial routes with CostMatrix (i.e. right direction of the shape). however, the problem is worse: it'll also assume the wrong directed edge when calculating time/cost. this PR is just the failing test which will need to pass once it's fixed.

### DIFF
--- a/valhalla/thor/costmatrix.h
+++ b/valhalla/thor/costmatrix.h
@@ -194,6 +194,8 @@ protected:
    * @param  pred    Edge label of the predecessor.
    * @param  n       Iteration counter.
    * @param  graphreader the graph reader instance
+   * @param  options     the request options to check for the position along origin and destination
+   *                     edges
    */
   void CheckForwardConnections(const uint32_t source,
                                const sif::BDEdgeLabel& pred,
@@ -230,6 +232,8 @@ protected:
    * @param  pred        Edge label of the predecessor.
    * @param  n           Iteration counter.
    * @param  graphreader the graph reader instance
+   * @param  options     the request options to check for the position along origin and destination
+   *                     edges
    */
   void CheckReverseConnections(const uint32_t target,
                                const sif::BDEdgeLabel& pred,


michaelhrabanek:
We've found another real life example showcasing this problem, leaving it here for future reference:
Using /route on oneway:
```
{
"costing": "auto",
"locations": [

		{
			"lon": 17.3993797302246,
			"lat": 49.2897300720215

		},
		{
			"lon": 17.3997802734375,
			"lat": 49.2892189025879

		}
],
	
    "shape_format": "polyline6",
    "narrative": false
}
```
returns correct shape and distance:
    "length": 0.063,

image

But for matrix:
```
{
	"costing": "auto",
	"sources": [
		{
			"lon": 17.3993797302246,
			"lat": 49.2897300720215

		}
	],
	"targets": [
		{
			"lon": 17.3997802734375,
			"lat": 49.2892189025879

		}
	],
	
	"shape_format": "polyline6",
    "prioritize_bidirectional": true
}
```
  "distance": 0.630,
image

chrstnbwnkl:
FWIW, I just added another small test that reproduces the issue reported by @michaelhrabanek, not sure yet though if they're related

chrstnbwnkl:
I spend some more time looking at this and I think I've got the culprit. 

In `SetSources` and `SetTargets`, the sort cost  for the correlated edges is set to the partial cost (i.e. `1 - pct_along * EdgeCost`) from the correlated location along the edge to its end. This is generally not bad I guess, but if I understand correctly it will always mess up the priority queue in case of trivial paths where source and target are along the same edge, because the edge leading away from the target (in case of forward expansion) will always have the lower sort cost . And since both edges (meaning the forward and backward edge along which the trivial route runs) are already placed in the reached map from the beginning, the first call to `CheckForwardConnections` will already mark the connection as found and thus no further connections are considered, even though in the second call to `CheckForwardConnections` the "right" edge is checked.

src/thor/costmatrix.cc
```diff
@@ -55,9 +55,9 @@ CostMatrix::CostMatrix(const boost::property_tree::ptree& config)
       max_reserved_locations_count_(
           config.get<uint32_t>("max_reserved_locations_costmatrix", kMaxLocationReservation)),
       check_reverse_connections_(config.get<bool>("costmatrix_check_reverse_connection", false)),
-      access_mode_(kAutoAccess),
-      mode_(travel_mode_t::kDrive), locs_count_{0, 0}, locs_remaining_{0, 0},
-      current_pathdist_threshold_(0), targets_{new ReachedMap}, sources_{new ReachedMap} {
+      access_mode_(kAutoAccess), mode_(travel_mode_t::kDrive), locs_count_{0, 0},
```
chrstnbwnkl, line 58: only reformatting...

valhalla/baldr/double_bucket_queue.h
```diff
@@ -98,7 +98,7 @@ template <typename label_t> class DoubleBucketQueue final {
   }
 
   /**
-   * Clear all labels from the low-level buckets and the overflow buckets and deallocates buckets
+   * Clear all labels from the low-level buckets and the overflow bucket and deallocate the buckets'
```
chrstnbwnkl, line 101: typo

valhalla/sif/hierarchylimits.h
```diff
@@ -46,8 +46,7 @@ struct HierarchyLimits {
                                 // always allowed. Used for A*.
 
   /**
-   * Set hierarchy limits for the specified level using a property tree.
-   * @param  pt     Property tree
+   * Set hierarchy limits for the specified level.
```
chrstnbwnkl, line 49: updated stale docstring

src/thor/costmatrix.cc
```diff
@@ -798,8 +801,15 @@ void CostMatrix::CheckForwardConnections(const uint32_t source,
 
     // Special case - common edge for source and target are both initial edges
     if (fwd_pred.predecessor() == kInvalidLabel && rev_predidx == kInvalidLabel) {
-      // bail if either edge wasn't allowed (see notes in SetSources/Targets)
-      if (!fwd_pred.path_id() || !rev_label.path_id()) {
+      // bail if forward edge wasn't allowed (see notes in SetSources/Targets)
+      if (!fwd_pred.path_id()) {
```
chrstnbwnkl, line 805: We don't really care if the reverse edge wasn't allowed, just the forward one

kevinkreiser (approved):
excellent debugging on this one

Issue #4568, opened 2024-02-06
ignore `hgv=no` on second pass via costing option


nilsnolde:
it's fairly common that `hgv=no` is tagged on a road, e.g. in inner cities, but certain companies have a legal permit to enter anyways. it's similar to `hgv=permit` but currently we handle that the same as `hgv=private` which is fine, that's a hardly used tag anyways and would make such an option data-driven when there's actually no such data. instead I'd propose to make it a costing option, e.g. `ignore_truck_access` which would switch a truck's access mask to a car's access mask (not sure if possible in code, but smth like that). ~~we already have a `ignore_restrictions` flag, but that's for all other restrictions (turn, dimensions etc), not base access.~~ (see below, we also have `ignore_access`)

also smth we could just turn on in a second pass.

I think that would help logistics companies a lot. any opinions on this?

nilsnolde:
Ha, we do have `ignore_access` as costing option (really gotta get the docs in order..). It's doing too much though for the use case described above: it just cares if _any_ mode has access and if so, it grant the current mode access. so we'd go down ped/bike paths, ignore bollards etc.

radekvermirovsky:
That would be greatly appreciated for our use case. We encounter many scenarios where our goal is to adhere to truck restrictions as much as possible. However, in cases where no alternative exists, we accept breaking restriction on edge to find a route at any cost. The 'hgv_access_penalty' appears to be precisely what would solve this issue. It would be a significant improvement over falling back to car mode routing, as it attempts to respect restrictions wherever possible, and only break it on edges where is no other choice. So, +1 from me

PR #4638, opened 2024-03-18
refactor path pruning for not_thru/destonly/closures

nilsnolde:
as preparation for https://github.com/valhalla/valhalla/issues/4568

in that issue I promised to look into the current path pruning conditions which we apply to origin/destination correlated edges to allow closed/destonly/not_thru, but otherwise prohibit them (and enable some again on a 2nd pass). this is what I did:
- I realized `not_thru` is handled differently than the others and the special handling didn't make sense to me, so I refactored it to be consistent with the others
- checking all the logic there led me down a serious rabbit hole of refactoring `*EdgeLabel` a bit and making sure all algos & costings use those things properly and consistently

however, eventually, after I refactored everything and made sure it's all sound, I ran a few existing tests and one particular one kinda makes my first refactor moot (treating `not_thru` the same as the others). the particular test that's failing is a quite a rare case but was added particularly to avoid the scenario where it's breaking on in this PR. however, we might be able to solve this in a different way rather than reverting my refactor (which does help a lot understanding things IMO). more comments on that test.

### DIFF
--- a/src/sif/autocost.cc
+++ b/src/sif/autocost.cc
@@ -423,7 +423,8 @@ bool AutoCost::Allowed(const baldr::DirectedEdge* edge,
   if (!IsAccessible(edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((pred.restrictions() & (1 << edge->localedgeidx())) && !ignore_turn_restrictions_) ||
       edge->surface() == Surface::kImpassable || IsUserAvoidEdge(edgeid) ||
-      (!allow_destination_only_ && !pred.destonly() && edge->destonly()) ||
+      (destination_only_pruning_ && pred.destonly_pruning() && edge->destonly()) ||
+      (not_thru_pruning_ && pred.not_thru_pruning() && edge->not_thru()) ||
       (pred.closure_pruning() && IsClosed(edge, tile)) ||
       (exclude_unpaved_ && !pred.unpaved() && edge->unpaved()) || !IsHOVAllowed(edge)) {
     return false;
@@ -448,7 +449,8 @@ bool AutoCost::AllowedReverse(const baldr::DirectedEdge* edge,
   if (!IsAccessible(opp_edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((opp_edge->restrictions() & (1 << pred.opp_local_idx())) && !ignore_turn_restrictions_) ||
       opp_edge->surface() == Surface::kImpassable || IsUserAvoidEdge(opp_edgeid) ||
-      (!allow_destination_only_ && !pred.destonly() && opp_edge->destonly()) ||
+      (destination_only_pruning_ && pred.destonly_pruning() && opp_edge->destonly()) ||
+      (not_thru_pruning_ && pred.not_thru_pruning() && edge->not_thru()) ||
       (pred.closure_pruning() && IsClosed(opp_edge, tile)) ||
       (exclude_unpaved_ && !pred.unpaved() && opp_edge->unpaved()) || !IsHOVAllowed(opp_edge)) {
     return false;
@@ -785,7 +787,8 @@ bool BusCost::Allowed(const baldr::DirectedEdge* edge,
   if (!IsAccessible(edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((pred.restrictions() & (1 << edge->localedgeidx())) && !ignore_restrictions_) ||
       edge->surface() == Surface::kImpassable || IsUserAvoidEdge(edgeid) ||
-      (!allow_destination_only_ && !pred.destonly() && edge->destonly()) ||
+      (destination_only_pruning_ && pred.destonly_pruning() && edge->destonly()) ||
+      (not_thru_pruning_ && pred.not_thru_pruning() && edge->not_thru()) ||
       (pred.closure_pruning() && IsClosed(edge, tile)) ||
       (exclude_unpaved_ && !pred.unpaved() && edge->unpaved())) {
     return false;
@@ -810,7 +813,8 @@ bool BusCost::AllowedReverse(const baldr::DirectedEdge* edge,
   if (!IsAccessible(opp_edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((opp_edge->restrictions() & (1 << pred.opp_local_idx())) && !ignore_turn_restrictions_) ||
       opp_edge->surface() == Surface::kImpassable || IsUserAvoidEdge(opp_edgeid) ||
-      (!allow_destination_only_ && !pred.destonly() && opp_edge->destonly()) ||
+      (destination_only_pruning_ && pred.destonly_pruning() && opp_edge->destonly()) ||
+      (not_thru_pruning_ && pred.not_thru_pruning() && edge->not_thru()) ||
       (pred.closure_pruning() && IsClosed(opp_edge, tile)) ||
       (exclude_unpaved_ && !pred.unpaved() && opp_edge->unpaved())) {
     return false;
@@ -964,7 +968,8 @@ bool TaxiCost::Allowed(const baldr::DirectedEdge* edge,
   if (!IsAccessible(edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((pred.restrictions() & (1 << edge->localedgeidx())) && !ignore_restrictions_) ||
       edge->surface() == Surface::kImpassable || IsUserAvoidEdge(edgeid) ||
-      (!allow_destination_only_ && !pred.destonly() && edge->destonly()) ||
+      (destination_only_pruning_ && pred.destonly_pruning() && edge->destonly()) ||
+      (not_thru_pruning_ && pred.not_thru_pruning() && edge->not_thru()) ||
       (pred.closure_pruning() && IsClosed(edge, tile)) ||
       (exclude_unpaved_ && !pred.unpaved() && edge->unpaved())) {
     return false;
@@ -989,7 +994,8 @@ bool TaxiCost::AllowedReverse(const baldr::DirectedEdge* edge,
   if (!IsAccessible(opp_edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((opp_edge->restrictions() & (1 << pred.opp_local_idx())) && !ignore_turn_restrictions_) ||
       opp_edge->surface() == Surface::kImpassable || IsUserAvoidEdge(opp_edgeid) ||
-      (!allow_destination_only_ && !pred.destonly() && opp_edge->destonly()) ||
+      (destination_only_pruning_ && pred.destonly_pruning() && opp_edge->destonly()) ||
+      (not_thru_pruning_ && pred.not_thru_pruning() && edge->not_thru()) ||
       (pred.closure_pruning() && IsClosed(opp_edge, tile)) ||
       (exclude_unpaved_ && !pred.unpaved() && opp_edge->unpaved())) {
     return false;
--- a/src/sif/dynamiccost.cc
+++ b/src/sif/dynamiccost.cc
@@ -146,9 +146,9 @@ DynamicCost::DynamicCost(const Costing& costing,
                          const TravelMode mode,
                          uint32_t access_mask,
                          bool penalize_uturns)
-    : pass_(0), allow_transit_connections_(false), allow_destination_only_(true),
-      allow_conditional_destination_(false), travel_mode_(mode), access_mask_(access_mask),
-      closure_factor_(kDefaultClosureFactor), flow_mask_(kDefaultFlowMask),
+    : pass_(0), allow_transit_connections_(false), destination_only_pruning_(true),
+      conditional_destination_only_pruning_(true), not_thru_pruning_(true), travel_mode_(mode),
+      access_mask_(access_mask), closure_factor_(kDefaultClosureFactor), flow_mask_(kDefaultFlowMask),
       shortest_(costing.options().shortest()),
       ignore_restrictions_(costing.options().ignore_restrictions()),
       ignore_non_vehicular_restrictions_(costing.options().ignore_non_vehicular_restrictions()),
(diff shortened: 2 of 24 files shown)


src/sif/autocost.cc
```diff
@@ -423,7 +423,8 @@ bool AutoCost::Allowed(const baldr::DirectedEdge* edge,
   if (!IsAccessible(edge) || (!pred.deadend() && pred.opp_local_idx() == edge->localedgeidx()) ||
       ((pred.restrictions() & (1 << edge->localedgeidx())) && !ignore_turn_restrictions_) ||
       edge->surface() == Surface::kImpassable || IsUserAvoidEdge(edgeid) ||
-      (!allow_destination_only_ && !pred.destonly() && edge->destonly()) ||
+      (destination_only_pruning_ && pred.destonly_pruning() && edge->destonly()) ||
+      (not_thru_pruning_ && pred.not_thru_pruning() && edge->not_thru()) ||
```
nilsnolde, line 427: I moved the `not_thru`  check into `Allowed(Reverse)`

src/sif/dynamiccost.cc
```diff
@@ -146,9 +146,9 @@ DynamicCost::DynamicCost(const Costing& costing,
                          const TravelMode mode,
                          uint32_t access_mask,
                          bool penalize_uturns)
-    : pass_(0), allow_transit_connections_(false), allow_destination_only_(true),
-      allow_conditional_destination_(false), travel_mode_(mode), access_mask_(access_mask),
-      closure_factor_(kDefaultClosureFactor), flow_mask_(kDefaultFlowMask),
+    : pass_(0), allow_transit_connections_(false), destination_only_pruning_(true),
```
nilsnolde, line 149: renamed the members to be consistent (now it has the opposite meaning of before, so I had to change all logic where this is used or set)

src/thor/astar_bss.cc
```diff
@@ -219,9 +219,11 @@ void AStarBSSAlgorithm::ExpandForward(GraphReader& graphreader,
     }
 
     // Add to the adjacency list and edge labels.
+    // TODO(nils): no pruning is enabled for this algo for now, needs some investigation if a second
+    // pass is possible for this algo (bike doesn't allow that)
     uint32_t idx = edgelabels_.size();
     edgelabels_.emplace_back(pred_idx, edgeid, GraphId(), directededge, newcost, sortcost, dist, mode,
-                             transition_cost, false, true, false, InternalTurn::kNoTurn,
+                             transition_cost, false, false, false, false, InternalTurn::kNoTurn,
```
nilsnolde, line 226: no pruning at all, see comment

src/thor/bidirectional_astar.cc
```diff
@@ -361,9 +361,8 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
                         pred.path_distance() + meta.edge->length(), newcost.cost);
   }
 
-  // we've just added this edge to the queue, but we won't expand from it if it's a not-thru edge that
-  // will be pruned. In that case we want to allow uturns.
-  return !(pred.not_thru_pruning() && meta.edge->not_thru());
```
nilsnolde, line 366: now we're now even adding not_thru edges to the queue, so this would be moot

src/thor/bidirectional_astar.cc
```diff
@@ -422,14 +421,14 @@ void BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
     // If so, it means we are attempting a u-turn. In that case, lets wait with evaluating
     // this edge until last. If any other edges were emplaced, it means we should not
     // even try to evaluate a u-turn since u-turns should only happen for deadends
-    uturn_meta = pred.opp_local_idx() == meta.edge->localedgeidx() ? meta : uturn_meta;
+    bool is_uturn = pred.opp_local_idx() == meta.edge->localedgeidx();
```
nilsnolde, line 424: just added this bool, we need it twice

src/thor/bidirectional_astar.cc
```diff
@@ -709,8 +708,7 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
 
       // Prune path if predecessor is not a through edge or if the maximum
       // number of upward transitions has been exceeded on this hierarchy level.
-      if ((fwd_pred.not_thru() && fwd_pred.not_thru_pruning()) ||
```
nilsnolde, line 712: a settled edge can't be `not_thru` anymore

src/thor/bidirectional_astar.cc
```diff
@@ -1024,13 +1022,18 @@ void BidirectionalAStar::SetOrigin(GraphReader& graphreader,
...
     // flags on small loops. Set this to false here to override this for now.
-    edgelabels_forward_.back().set_not_thru(false);
+    edgelabels_forward_.back().set_not_thru_pruning(false);
 
-    pruning_disabled_at_origin_ = pruning_disabled_at_origin_ ||
-                                  !edgelabels_forward_.back().closure_pruning() ||
-                                  !edgelabels_forward_.back().not_thru_pruning();
+    pruning_disabled_at_origin_ =
+        pruning_disabled_at_origin_ || !edgelabels_forward_.back().closure_pruning() ||
+        directededge->not_thru() || !edgelabels_forward_.back().destonly_pruning();
```
nilsnolde, line 1036: `destonly_pruning` is also a valid condition to extend the search in the other direction

src/thor/bidirectional_astar.cc
```diff
@@ -337,14 +336,15 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
       dist = astarheuristic_forward_.GetDistance(t2->get_node_ll(meta.edge->endnode()));
     }
     edgelabels_reverse_.emplace_back(pred_idx, meta.edge_id, opp_edge_id, meta.edge, newcost,
-                                     sortcost, dist, mode_, transition_cost, thru,
-                                     (pred.closure_pruning() || !costing_->IsClosed(meta.edge, tile)),
+                                     sortcost, dist, mode_, transition_cost, not_thru_pruning,
+                                     pred.closure_pruning() || !(costing_->IsClosed(opp_edge, t2)),
```
nilsnolde, line 340: it's easy to miss: I changed to checking if `opp_edge` is closed when expanding reverse because that's the one we want to go on in the final path

nilsnolde: there's quite a few of these fixes sprinkled in this PR

src/thor/bidirectional_astar.cc
```diff
@@ -1106,12 +1109,11 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader,
       dist = astarheuristic_forward_.GetDistance(tile->get_node_ll(opp_dir_edge->endnode()));
     }
     edgelabels_reverse_.emplace_back(kInvalidLabel, opp_edge_id, edgeid, opp_dir_edge, cost, sortcost,
-                                     dist, mode_, c, !opp_dir_edge->not_thru(),
-                                     !(costing_->IsClosed(directededge, tile)),
+                                     dist, mode_, c, false, !(costing_->IsClosed(directededge, tile)),
+                                     !(directededge->destonly() ||
```
nilsnolde, line 1113: when setting the destination however, we want to look at the passed directed edge (not the opposing), because we put the opposing in the queue

valhalla/sif/edgelabel.h
```diff
@@ -70,24 +70,24 @@ class EdgeLabel {
             const TravelMode mode,
             const uint32_t path_distance,
             const uint8_t restriction_idx,
+            const bool not_thru_pruning,
```
nilsnolde, line 73: I added both `not_thru_pruning` and `destonly_pruning` as first-class citizens to EdgeLabel, also timedistancematrix wants to look at them I think

test/gurka/test_route.cc
```diff
@@ -526,6 +526,7 @@ TEST(Standalone, BridgingEdgeIsRestricted) {
 TEST(Standalone, AvoidExtraDetours) {
   // Check that we don't take extra detours to get the target point. Here is
   // a special usecase that breaks previous logic about connection edges in bidirectional astar.
+  // TODO(nils): this currently fails because both trees hit a not_thru area on DE (fwd) and DC(rev)
```
nilsnolde, line 529: so in the map below, the route wants to from 1 -> 2 with bidir a*. the test sets D -> A edges and D -> F edges as `not_thru`, which means neither can the fwd search see the rev tree, nor can the rev search see any of the fwd tree and they both expand into the HIJK region, taking a massive detour.

I _think_ it fails now because we don't even record `not_thru` edges which means a connection on those can't happen anymore.

2 things:
1. this test describes a real failure mapbox noticed on their master a while back, but jeez this is kinda rare.. and not sure it really justifies the extra complex logic that was there before
2. but ok, it shows a rare bug. however, the real problem here is that we _do_ find a connection, it's just a huge detour. if we wouldn't find a connection in the first pass, that's fine, we would find it on a second pass. anyways, obviously it takes `CD` & `DE` edges in either case. in master it'd make that transition directly to have the path `{ "AB", "BC", "CD", "DE", "EF" }`. in this PR, it can't and instead `D` loops back on itself, i.e. with path `{ "AB", "BC", "CD", "DH", "HI", "IJ", "JK", "KH", "DH", "DE", "EF" }` which simply adds the sub-path `"DH", "HI", "IJ", "JK", "KH", "DH"`. now what I wonder: couldn't we detect that while recosting or in triplegbuilder and just cut that completely superfluous part out of the final path?

nilsnolde:
isochrones fail as well, the polygons changed their shape again with this PR, I'm not sure why yet, actually not much changed for dijkstra. but the IMO more important tests dedicated to `not_thru` & `destination_only` all pass at least..

test/gurka/test_route.cc
nilsnolde: actually I'm pretty sure this could fairly easily be done. the only thing to be wary about that I can think of are turn restrictions. if in the case above `CD` -> `DE` was restricted, we'd allow the detour, otherwise we'd cut the superfluous part and recost the remainder of the path accordingly. it'll need to be recorded while `recost_forward` is running, I'll try an implementation later. 

it's only interesting for bidirectional algorithms. and while it'll be a bit more of a burden since we'd expand that detour where we wouldn't have to, this should be a very rare case and would maybe help other cases of unnecessary detours as well?!

src/thor/unidirectional_astar.cc
```diff
@@ -267,27 +267,28 @@ inline bool UnidirectionalAStar<expansion_direction, FORWARD>::ExpandInner(
       best_path.second = cost.cost;
     }
 
+    // not_thru is the same for both trees
+    bool not_thru_pruning = pred.not_thru_pruning() || !meta.edge->not_thru();
```
nilsnolde, line 271: in fact, this algorithm never used not_thru pruning.. it recorded it in the label, but didn't make use of it.

test/gurka/test_conditional_restrictions.cc
```diff
@@ -243,7 +243,10 @@ TEST_F(ConditionalRestrictions, DestinationRestrictionOnMidEdgeIsNotValid_OnlyAl
   }
 }
 
-TEST_F(ConditionalRestrictions, DestinationRestrictionOnMidEdgeIsValid_ManyAlternatives) {
+// TODO: this triggers a not_thru pruning failing the expected route, which wasn't looked at before in
+// unidir a* and now we enter a second pass in which the _both_ the not_thru pruning and conditional
+// restriction are dismissed
+TEST_F(ConditionalRestrictions, DISABLED_DestinationRestrictionOnMidEdgeIsValid_ManyAlternatives) {
```
nilsnolde, line 249: this is unfortunate but also exposed a bug in the previous behavior of unidir a* where it was expanding into not_thru regions (or maybe it was wanted that way because it's unidirectional?). since it now does not expand into those regions, it fails on a first pass and on the second pass it finds the fastest route, right over the conditional restriction. 

that behavior is really unfortunate.. and brings up more considerations.. not_thru is only a performance technique, but restrictions are legal constructs. it sucks a bit that we conflate both into a second pass.

I can think of 2 things to improve:
- actually do https://github.com/valhalla/valhalla/pull/3812 where we set all edges in a not_thru region to `not_thru`. the hope is that we wouldn't need a second pass at all, because we could accurately determine if a location is inside such a region and if so, we allow the other (or only) tree to expand into them, which would be surely better than wasting a full expansion and go into a second pass. however, I didn't fully investigate all potential side effect, so TBD..
- the fact that there's a not_thru edge somewhere in that test map is far from obvious.. and very surely will trip another future test map (and possibly does so already without noticing bcs of second passes?). I'd propose to turn all edge into `not_thru = false` by default and let tests who want it set the property explicitly.

nilsnolde:
ok, I'm down to 2 failing tests. both can be explained:
1. `gurka/conditional_restrictions` fails because unidir a* previously would expand into not_thru regions. this PR changes that and the test map has a hidden not_thru where it's expected to find a path. which triggers it to go into a second pass, then ignoring both restrictions & not_thru, and then finding the "wrong" path, see more in https://github.com/valhalla/valhalla/pull/4638#discussion_r1531413753
2. `gurka/route` actually fails legitimately and this PR introduces a bug which was fixed before. however, I could imagine that bug could be fixed in another way too, see https://github.com/valhalla/valhalla/pull/4638#discussion_r1529380686.

all in all: while this PR fixes a lot of subtle bugs, I kinda dug my own hole here trying to refactor the pruning things to make them more understandable. it's a LOT of changes for such a small-ish impact. even I'd try and abstract the concept a bit more in another PR, I'm not sure this is all worth it. also looking at the 2nd test failure, this PR is actually introducing a previously fixed bug. I'm proposing another solution to that particular bug which might solve other cases too, but again, might not be worth it..

hope we'll get some time to discuss this.

nilsnolde:
after talking to @dnesbitt61 about this, I'll put this on hold. the changes to not_thru pruning are not good (generating an already fixed bug) and my proposal to edit the final path is another potential hole with tons of problem cases I don't think of right now. I'll take out the small bug fixes from this PR and call it a day. refactoring this properly would take a lot of work it seems (more even than already put in here).

he had another idea which seems worthwhile and could be investigated at some point: only turn off not_thru pruning once we get closer to the destination/origin on the fwd/rev tree. which seems intuitively correct and could save us from needing a second pass as well, especially coupled with being able to determine accurately if a dest/origin is inside such a region with #3812.

PR #4642, opened 2024-03-20
some bug fixes ported over from #4638

nilsnolde:
more inline

### DIFF
--- a/src/sif/pedestriancost.cc
+++ b/src/sif/pedestriancost.cc
@@ -550,7 +550,6 @@ public:
          (edge->use() == baldr::Use::kRailFerry && pred->use() != baldr::Use::kRailFerry);
 
     // Additional penalties without any time cost
-    c.cost += destination_only_penalty_ * (edge->destonly() && !pred->destonly());
     c.cost +=
         alley_penalty_ * (edge->use() == baldr::Use::kAlley && pred->use() != baldr::Use::kAlley);
     c.cost +=
--- a/src/sif/pedestriancost.cc
+++ b/src/sif/pedestriancost.cc
@@ -550,6 +550,7 @@ public:
          (edge->use() == baldr::Use::kRailFerry && pred->use() != baldr::Use::kRailFerry);
 
     // Additional penalties without any time cost
+    c.cost += destination_only_penalty_ * (edge->destonly() && !pred->destonly());
     c.cost +=
         alley_penalty_ * (edge->use() == baldr::Use::kAlley && pred->use() != baldr::Use::kAlley);
     c.cost +=
@@ -602,7 +603,7 @@ PedestrianCost::PedestrianCost(const Costing& costing)
   speed_ = costing_options.walking_speed();
   step_penalty_ = costing_options.step_penalty();
   elevator_penalty_ = costing_options.elevator_penalty();
-  max_grade_ = costing_options.max_grade();
+  max_grade_ = costing_options.max_grade();private
 
   if (type_ == PedestrianType::kFoot) {
     max_hiking_difficulty_ = static_cast<SacScale>(costing_options.max_hiking_difficulty());
--- a/src/sif/pedestriancost.cc
+++ b/src/sif/pedestriancost.cc
@@ -603,7 +603,7 @@ PedestrianCost::PedestrianCost(const Costing& costing)
   speed_ = costing_options.walking_speed();
   step_penalty_ = costing_options.step_penalty();
   elevator_penalty_ = costing_options.elevator_penalty();
-  max_grade_ = costing_options.max_grade();private
+  max_grade_ = costing_options.max_grade();
 
   if (type_ == PedestrianType::kFoot) {
     max_hiking_difficulty_ = static_cast<SacScale>(costing_options.max_hiking_difficulty());
--- a/src/sif/recost.cc
+++ b/src/sif/recost.cc
@@ -91,7 +91,7 @@ void recost_forward(baldr::GraphReader& reader,
     // TODO: if this edge begins a restriction, we need to start popping off edges into queue
     // so that we can find if we reach the end of the restriction. then we need to replay the
     // queued edges as normal
-    uint8_t time_restrictions_TODO = -1;
+    uint8_t time_restrictions_TODO = baldr::kInvalidRestriction;
     // if its not time dependent set to 0 for Allowed method below
     const uint64_t localtime = offset_time.valid ? offset_time.local_time : 0;
     // we should call 'Allowed' method even if 'ignore_access' flag is true in order to
--- a/src/thor/astar_bss.cc
+++ b/src/thor/astar_bss.cc
@@ -221,7 +221,7 @@ void AStarBSSAlgorithm::ExpandForward(GraphReader& graphreader,
     // Add to the adjacency list and edge labels.
     uint32_t idx = edgelabels_.size();
     edgelabels_.emplace_back(pred_idx, edgeid, GraphId(), directededge, newcost, sortcost, dist, mode,
-                             transition_cost, false, true, false, InternalTurn::kNoTurn,
+                             transition_cost, false, false, false, InternalTurn::kNoTurn,
                              baldr::kInvalidRestriction);
     *current_es = {EdgeSet::kTemporary, idx};
     adjacencylist_.add(idx);
@@ -446,7 +446,7 @@ void AStarBSSAlgorithm::SetOrigin(GraphReader& graphreader,
     // of the path.
     uint32_t d = static_cast<uint32_t>(directededge->length() * (1.0f - edge.percent_along()));
     BDEdgeLabel edge_label(kInvalidLabel, edgeid, directededge, cost, sortcost, dist,
-                           travel_mode_t::kPedestrian, baldr::kInvalidRestriction, true, false,
+                           travel_mode_t::kPedestrian, baldr::kInvalidRestriction, false, false,
                            sif::InternalTurn::kNoTurn);
     // Set the origin flag and path distance
     edge_label.set_origin();
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -365,7 +365,7 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
 
   // we've just added this edge to the queue, but we won't expand from it if it's a not-thru edge that
   // will be pruned. In that case we want to allow uturns.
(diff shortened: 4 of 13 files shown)


src/sif/pedestriancost.cc
```diff
@@ -550,7 +550,6 @@ class PedestrianCost : public DynamicCost {
          (edge->use() == baldr::Use::kRailFerry && pred->use() != baldr::Use::kRailFerry);
 
     // Additional penalties without any time cost
-    c.cost += destination_only_penalty_ * (edge->destonly() && !pred->destonly());
```
nilsnolde, line 553: ped with destonly penalty sounds weird:)

src/sif/recost.cc
```diff
@@ -91,7 +91,7 @@ void recost_forward(baldr::GraphReader& reader,
     // TODO: if this edge begins a restriction, we need to start popping off edges into queue
     // so that we can find if we reach the end of the restriction. then we need to replay the
     // queued edges as normal
-    uint8_t time_restrictions_TODO = -1;
+    uint8_t time_restrictions_TODO = baldr::kInvalidRestriction;
```
nilsnolde, line 94: this you'll see a lot, why not use it if it already exists.

src/thor/astar_bss.cc
```diff
@@ -221,7 +221,7 @@ void AStarBSSAlgorithm::ExpandForward(GraphReader& graphreader,
     // Add to the adjacency list and edge labels.
     uint32_t idx = edgelabels_.size();
     edgelabels_.emplace_back(pred_idx, edgeid, GraphId(), directededge, newcost, sortcost, dist, mode,
-                             transition_cost, false, true, false, InternalTurn::kNoTurn,
+                             transition_cost, false, false, false, InternalTurn::kNoTurn,
```
nilsnolde, line 224: also this you'll see a lot: this flag is setting `closure_pruning` which is not smth we want for lots of algos (not that it matters at all, but it's again more explicit IMO to set `false` here)

src/thor/bidirectional_astar.cc
```diff
@@ -309,7 +309,9 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
                           : astarheuristic_reverse_.Get(t2->get_node_ll(meta.edge->endnode()), dist));
 
   // not_thru_pruning_ is only set to false on the 2nd pass in route_action.
-  bool thru = not_thru_pruning_ ? (pred.not_thru_pruning() || !meta.edge->not_thru()) : false;
+  // We allow settling not_thru edges so we can connect both trees on them.
```
nilsnolde, line 312: this comment is what eventually wasted a good 2 days of work on #4638 when I discovered that at the very end. very subtle..

src/thor/bidirectional_astar.cc
```diff
@@ -337,8 +339,8 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
       dist = astarheuristic_forward_.GetDistance(t2->get_node_ll(meta.edge->endnode()));
     }
     edgelabels_reverse_.emplace_back(pred_idx, meta.edge_id, opp_edge_id, meta.edge, newcost,
-                                     sortcost, dist, mode_, transition_cost, thru,
-                                     (pred.closure_pruning() || !costing_->IsClosed(meta.edge, tile)),
+                                     sortcost, dist, mode_, transition_cost, not_thru_pruning,
+                                     (pred.closure_pruning() || !costing_->IsClosed(opp_edge, t2)),
```
nilsnolde, line 343: I'm 100% sure now that we need to look at the opposing edge here because we're expanding on the reverse tree. case in point: AllowedReverse will prune on the opposing edge as well, not the one that gets expanded. it's really just common sense, though easy to miss..

you'll see that in a lot of `ExpandInner` functions throughout the PR

src/thor/bidirectional_astar.cc
```diff
@@ -363,7 +365,7 @@ inline bool BidirectionalAStar::ExpandInner(baldr::GraphReader& graphreader,
 
   // we've just added this edge to the queue, but we won't expand from it if it's a not-thru edge that
   // will be pruned. In that case we want to allow uturns.
-  return !(pred.not_thru_pruning() && meta.edge->not_thru());
+  return not_thru_pruning_ && !(pred.not_thru_pruning() && meta.edge->not_thru());
```
nilsnolde, line 368: on a second pass we don't want to allow a u-turn if this was actually not_thru

src/thor/bidirectional_astar.cc
```diff
@@ -422,13 +424,13 @@ void BidirectionalAStar::Expand(baldr::GraphReader& graphreader,
     // If so, it means we are attempting a u-turn. In that case, lets wait with evaluating
     // this edge until last. If any other edges were emplaced, it means we should not
     // even try to evaluate a u-turn since u-turns should only happen for deadends
-    uturn_meta = pred.opp_local_idx() == meta.edge->localedgeidx() ? meta : uturn_meta;
+    bool is_uturn = pred.opp_local_idx() == meta.edge->localedgeidx();
```
nilsnolde, line 427: just putting this in a variable to make it clearer

src/thor/bidirectional_astar.cc
```diff
@@ -1028,9 +1030,9 @@ void BidirectionalAStar::SetOrigin(GraphReader& graphreader,
     // flags on small loops. Set this to false here to override this for now.
     edgelabels_forward_.back().set_not_thru(false);
 
-    pruning_disabled_at_origin_ = pruning_disabled_at_origin_ ||
-                                  !edgelabels_forward_.back().closure_pruning() ||
-                                  !edgelabels_forward_.back().not_thru_pruning();
+    pruning_disabled_at_origin_ =
+        pruning_disabled_at_origin_ || !edgelabels_forward_.back().closure_pruning() ||
+        !edgelabels_forward_.back().not_thru_pruning() || edgelabels_forward_.back().destonly();
```
nilsnolde, line 1035: if the destination/origin projects on destonly edge, it's also a reason to extend expansion in the opposite direction.

src/thor/bidirectional_astar.cc
```diff
@@ -1109,9 +1111,9 @@ void BidirectionalAStar::SetDestination(GraphReader& graphreader,
                                      dist, mode_, c, !opp_dir_edge->not_thru(),
                                      !(costing_->IsClosed(directededge, tile)),
                                      static_cast<bool>(flow_sources & kDefaultFlowMask),
-                                     sif::InternalTurn::kNoTurn, -1,
-                                     opp_dir_edge->destonly() ||
-                                         (costing_->is_hgv() && opp_dir_edge->destonly_hgv()));
+                                     sif::InternalTurn::kNoTurn, kInvalidRestriction,
+                                     directededge->destonly() ||
+                                         (costing_->is_hgv() && directededge->destonly_hgv()));
```
nilsnolde, line 1116: however, in `SetDestination` functions we enqueue the _opposing_ edge, so really we need to look at the correlated edge here

src/thor/costmatrix.cc
```diff
@@ -512,16 +512,16 @@ bool CostMatrix::ExpandInner(baldr::GraphReader& graphreader,
   }
 
   // not_thru_pruning_ is only set to false on the 2nd pass in matrix_action.
-  // TODO(nils): one of these cases where I think reverse tree should look at the opposing edge,
-  //   not the expanding one, same for quite some attributes below (and same in bidir a*)
-  bool thru = not_thru_pruning_ ? (pred.not_thru_pruning() || !meta.edge->not_thru()) : false;
+  // We allow settling not_thru edges so we can connect both trees on them.
```
nilsnolde, line 515: changes in CostMatrix are basically the exact same as in bidir A*

src/thor/timedistancebssmatrix.cc
```diff
@@ -315,11 +315,11 @@ void TimeDistanceBSSMatrix::SetOrigin(GraphReader& graphreader, const valhalla::
       dist = static_cast<uint32_t>(directededge->length() * percent_along);
 
     } else {
-      opp_edge_id = graphreader.GetOpposingEdgeId(edgeid);
+      opp_edge_id = graphreader.GetOpposingEdgeId(edgeid, endtile);
```
nilsnolde, line 318: just some caching stuff

src/thor/unidirectional_astar.cc
```diff
@@ -466,7 +466,7 @@ std::vector<std::vector<PathInfo>> UnidirectionalAStar<expansion_direction, FORW
   midgard::PointLL destination_new(destination.correlation().edges(0).ll().lng(),
                                    destination.correlation().edges(0).ll().lat());
   Init(origin_new, destination_new);
-  float mindist = astarheuristic_.GetDistance(origin_new);
+  float mindist = astarheuristic_.GetDistance(FORWARD ? origin_new : destination_new);
```
nilsnolde, line 469: this seemed like another bug. on reverse we want to get the `mindist` from the origin to the destination

valhalla/sif/costfactory.h
```diff
@@ -101,6 +101,7 @@ class CostFactory {
 
   mode_costing_t CreateModeCosting(const Options& options, TravelMode& mode) {
     mode_costing_t mode_costing;
+    mode = TravelMode::kMaxTravelMode;
```
nilsnolde, line 104: some of the tests were all of a sudden throwing a "mode might be uninitialized" warning. it's technically possible I guess, in case we'd happen to add another costing type which isn't represented in `kCostingTypeMapping`. makes me think of https://github.com/valhalla/valhalla/pull/4492, in case you're not yet aware of it @ianthetechie :)

ianthetechie: heh ;) Thanks for the heads up @nilsnolde. This is vaguely familiar but making a note to review when I get back to finalizing that PR.

src/sif/pedestriancost.cc
nilsnolde: hm, maybe it has a point since we also turn `private` roads into `destination_only` and even pedestrian shouldn't take those..

nilsnolde: it's a bit unfortunate though, "residents only" is pointless for ped but we can't distinguish

nilsnolde:
Did the RAD test for `de_benchmark_routes.txt` and it's the same in perf & results for the 20ish routes I compared.

I guess the only thing which could have a "negative" impact is adding a location's `destonly` as a criterion to extend the search in the other direction, but really that's an improvement, it might save us from a second pass.

PR #4650, opened 2024-03-22
truck penalty for hgv=no edges

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/4568

based on https://github.com/valhalla/valhalla/pull/4642

- added `hgv_no_access_penalty` costing option in the range [0.f, `kMaxPenalty`[
- if `kMaxPenalty` we treat `hgv=no` edges the same as today for truck costing, i.e. we restrict access
- if < `kMaxPenalty` we:
  - don't apply the penalty to correlated edge with `hgv=no`
  - only start applying the penalty after it expands on its first `hgv=yes`

so in the end it works similar to destonly pruning, also with a penalty, only there’s no pruning of turned on.


docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -148,6 +148,7 @@ The following options are available for `truck` costing.
 | `axle_load` | The axle load of the truck (in metric tons). Default 9.07. |
 | `axle_count` | The axle count of the truck. Default 5. |
 | `hazmat` | A value indicating if the truck is carrying hazardous materials. Default false. |
+| `hgv_no_access_penalty` | A penalty applied to roads with no HGV/truck access. If set to a value less than 43200 seconds, HGV will be allowed on these roads and the penalty applies. Default 43200, i.e. trucks are not allowed. |
```
nilsnolde, line 151: open for a different name:)

lua/graph.lua
```diff
@@ -91,7 +91,7 @@ access = {
 ["forestry"] = "false",
 ["destination"] = "true",
 ["customers"] = "true",
-["official"] = "false",
+["official"] = "true",
```
nilsnolde, line 94: AFAIU `official` and `permissive`  are signifying allowed access

proto/options.proto
```diff
@@ -313,6 +313,9 @@ message Costing {
     float use_lit = 82;
     bool disable_hierarchy_pruning = 83;
     bool ignore_non_vehicular_restrictions = 84;
+    oneof has_hgv_no_access_penalty {
```
nilsnolde, line 316: has a default of `kMaxPenalty`

src/mjolnir/pbfgraphparser.cc
```diff
@@ -4990,7 +4990,8 @@ void PBFGraphParser::ParseRelations(const boost::property_tree::ptree& pt,
                                                         OSMPBF::Interest::CHANGESETS),
                           callback);
   }
-  LOG_INFO("Finished with " + std::to_string(osmdata.restrictions.size()) + " simple restrictions");
+  LOG_INFO("Finished with " + std::to_string(osmdata.restrictions.size()) +
+           " simple turn restrictions");
```
nilsnolde, line 4994: "restrictions" is a bit too generic IMO

src/mjolnir/restrictionbuilder.cc
```diff
@@ -752,7 +752,7 @@ void RestrictionBuilder::Build(const boost::property_tree::ptree& pt,
     std::vector<std::promise<Result>> promises(threads.size());
 
     // Start the threads
-    LOG_INFO("Adding Restrictions at level " + std::to_string(tl->level));
+    LOG_INFO("Adding complex turn restrictions at level " + std::to_string(tl->level));
```
nilsnolde, line 755: same here

src/sif/truckcost.cc
```diff
@@ -126,6 +127,14 @@ class TruckCost : public DynamicCost {
 
   virtual ~TruckCost();
 
+  /**
+   * If we want to penalize HGV access, pretend we're driving a car so hgv=no edges are allowed.
+   * @return  Returns car access if we want to penalize hgv=no or truck access by default.
+   */
+  virtual uint32_t access_mode() const override {
+    return truck_access_mask_;
```
nilsnolde, line 135: `truck_access_mask_` will be `kTruckAccess` by default, but `kAutoAccess` if `hgv_no_access_penalty`  is set in the request

src/sif/truckcost.cc
```diff
@@ -447,7 +465,7 @@ inline bool TruckCost::Allowed(const baldr::DirectedEdge* edge,
     return false;
   }
 
-  return DynamicCost::EvaluateRestrictions(access_mask_, edge, is_dest, tile, edgeid, current_time,
+  return DynamicCost::EvaluateRestrictions(kTruckAccess, edge, is_dest, tile, edgeid, current_time,
```
nilsnolde, line 468: then we have to pass `kTruckAccess` here explicitly as we still want to evaluate truck restrictions regardless of that costing option

test/gurka/test_matrix.cc
```diff
@@ -820,3 +820,72 @@ TEST(StandAlone, CostMatrixTrivialRoutes) {
...
+  gurka::map map = gurka::buildtiles(layout, ways, {}, {}, "test/data/hgv_no_access_penalty",
+                                     {{"service_limits.max_timedep_distance_matrix", "50000"}});
+
+  std::unordered_map<std::string, std::string> cost_matrix =
+      {{"/costing_options/truck/hgv_no_access_penalty", "2000"},
+       {"/sources/0/date_time", "2024-03-20T09:00"},
+       {"/prioritize_bidirectional", "1"}};
+  std::unordered_map<std::string, std::string> td_matrix =
+      {{"/costing_options/truck/hgv_no_access_penalty", "2000"},
+       {"/sources/0/date_time", "2024-03-20T09:00"}};
```
nilsnolde, line 858: making a time request makes sure we can call both costmatrix & tdmatrix in a simple loop below

test/gurka/test_route.cc
```diff
@@ -1153,3 +1153,85 @@ TEST_F(DateTimeTest, Invariant) {
...
+  // if hgv_no_penalty is on we should still respect the maxweight restriction on CD
+  // so we should take the next-best hgv=no edge with JK
+  const std::string ascii_map = R"(
+    A-1--B----C----D----E--2-F----G----H--3-I
+              |    |
+              J----K
+              |    |             
+              |    |
+              L----M
+           )";
```
nilsnolde, line 1167: basically the same map as for the matrix test

valhalla/sif/dynamiccost.h
```diff
@@ -497,10 +497,10 @@ class DynamicCost {
 
     // If forward, check if the edge marks the end of a restriction, else check
     // if the edge marks the start of a complex restriction.
-    if ((forward && (edge->end_restriction() & access_mode())) ||
-        (!forward && (edge->start_restriction() & access_mode()))) {
+    if ((forward && (edge->end_restriction() & access_mask_)) ||
```
nilsnolde, line 500: this is in `Restricted` for complex restrictions. instead of doing this I'd rather do what `EvaluateRestrictions` is doing and pass the access mask as argument. before this PR that was actually fully redundant but now it does make sense to pass it as argument.

src/sif/truckcost.cc
nilsnolde: let's make this protected

src/sif/truckcost.cc
```diff
@@ -368,6 +381,11 @@ TruckCost::TruckCost(const Costing& costing)
   for (uint32_t d = 0; d < 16; d++) {
     density_factor_[d] = 0.85f + (d * 0.025f);
   }
+
+  // determine what to do with hgv=no edges
+  bool no_hgv_access_penalty_active = !(costing_options.hgv_no_access_penalty() == kMaxPenalty);
+  no_hgv_access_penalty_ = no_hgv_access_penalty_active * costing_options.hgv_no_access_penalty();
+  truck_access_mask_ = no_hgv_access_penalty_active ? kAutoAccess : kTruckAccess;
```
nilsnolde, line 388: do bitwise or here

valhalla/sif/dynamiccost.h
```diff
@@ -497,10 +497,10 @@ class DynamicCost {
...
+        (!forward && (edge->start_restriction() & access_mask_))) {
```
nilsnolde, line 501: everything else uses `access_mask_` instead of calling the public api and makes sense as long as we're inside the class

nilsnolde:
ok, did the changes we talked about @kevinkreiser . I also saw that triplegbuilder uses `costing->access_mode()` to get the restrictions: 
https://github.com/valhalla/valhalla/blob/962719a352d06198dfbfae2c653c00f6aa99e060/src/thor/triplegbuilder.cc#L1248-L1253

which is fine though, we'd want them in the final trip edges for car mode as well if the new penalty was turned on.

docs/docs/api/turn-by-turn/api-reference.md
kevinkreiser: i dont know of another way to do it but i dislike having to document implementation details like the max penalty. it removes the ability for us to change it in the future which i also know would be a jerk move but still like to have the option for it to not be set in stone. anyway not asking you to do something different, more thinking outloud in case something comes to someone else mind about this.

lua/graph.lua
kevinkreiser: yeah i think it would be easy for us to have messed this up by reading it as "official use only" which is a special U.S.-centric thing that means basically cops and emergency vehicles can use it but no one else. good catch

proto/options.proto
kevinkreiser: would have be cool to have the default be 0, not use `oneof` and then just had 0 mean not allowed but i guess there is a tiny argument for you to be able to say its both allowed and the penalty is actually 0. anyway would have allowed us to not document the internal max penalty or been able to pick a smaller soft limit or something i dont i know haha

src/sif/truckcost.cc
```diff
@@ -368,6 +372,12 @@ TruckCost::TruckCost(const Costing& costing)
   for (uint32_t d = 0; d < 16; d++) {
     density_factor_[d] = 0.85f + (d * 0.025f);
   }
+
+  // determine what to do with hgv=no edges
+  bool no_hgv_access_penalty_active = !(costing_options.hgv_no_access_penalty() == kMaxPenalty);
+  no_hgv_access_penalty_ = no_hgv_access_penalty_active * costing_options.hgv_no_access_penalty();
```
kevinkreiser, line 378: nit: is this necessary, cant you just set it regardless? it will only be applied when the access doesnt have truck right? which wont happen anyway if its not "active" or am i missing something? just a nit not important

src/sif/truckcost.cc
```diff
@@ -551,6 +561,10 @@ Cost TruckCost::TransitionCost(const baldr::DirectedEdge* edge,
     c.cost += low_class_penalty_;
   }
 
+  // Penalty if the request wants to avoid hgv=no edges instead of disallowing
+  c.cost +=
+      no_hgv_access_penalty_ * (pred.has_hgv_access() && !(edge->forwardaccess() & kTruckAccess));
```
kevinkreiser, line 566: cool so we only penalize on entering it :+1:

valhalla/sif/dynamiccost.h
```diff
@@ -599,6 +599,18 @@ class DynamicCost {
...
+   * search to get the tile's access restrictions.
+   *
+   * @param access_mode        The access mode to get restrictions for
+   * @param edge               The edge to check for restrictions
+   * @param is_dest            Is there a destination on the edge?
+   * @param tile               The edge's tile
+   * @param current_time       Needed for time dependent restrictions
+   * @param tz_index           The current timezone index
+   * @param restriction_idx    Records the restriction in the tile for later retrieval
+   */
```
kevinkreiser, line 613: :pray:

valhalla/sif/edgelabel.h
```diff
@@ -433,7 +444,9 @@ class EdgeLabel {
   // Flag indicating edge is an unpaved road.
   uint32_t unpaved_ : 1;
   uint32_t has_measured_speed_ : 1;
-  uint32_t spare : 13;
+  // Flag if this edge had HGV access
+  uint32_t hgv_access_ : 1;
+  uint32_t spare : 12;
```
kevinkreiser, line 449: one day we'll just have the whole mask here and be done with it haha

kevinkreiser (approved):

src/sif/truckcost.cc
nilsnolde: if we don't do it here and want to use it in transitioncost, then we'd have to check the same thing, i.e. if it's ` == kMaxPenalty`. instead of having that ternary being done every time we call TransitionCost, I'd rather do it once here.

nilsnolde:
fyi, since the last master merge we only updated vicky's test files, it's safe to merge without running CI again.

PR #4651, opened 2024-03-22
no need to search for GDAL when building data

nilsnolde:
just a small cleanup: don't even look for gdal if the installation will only be used for building data. even I just installed gdal on the building server when I saw the warning :smile: then I thought "why?".


chrstnbwnkl:
ah yeah I guess that makes sense now that `ENABLE_GDAL` is on by default

chrstnbwnkl (approved):

PR #4661, opened 2024-03-26
Marking unused code

cvvergara:
# Issue

## Tasklist

 - [x] fixing test warnings
 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/test/gurka/test_instructions_roundabout.cc
+++ b/test/gurka/test_instructions_roundabout.cc
@@ -69,11 +69,13 @@ TEST_F(InstructionsRoundabout, RoundaboutEnterOnly) {
 
   // TODO: known issue - future update to end on a roundabout
   //  Verify the enter_roundabout instructions
-  //      gurka::assert::raw::expect_instructions_at_maneuver_index(result, maneuver_index,
-  //                                                                "Enter the roundabout.", "Enter
-  //                                                                the roundabout.", "Enter the
-  //                                                                roundabout.", "Enter the
-  //                                                                roundabout.", "");
+#if 0
+  gurka::assert::raw::expect_instructions_at_maneuver_index(result, maneuver_index,
+                                                            "Enter the roundabout.",
+                                                            "Enter the roundabout.",
+                                                            "Enter the roundabout.",
+                                                            "Enter the roundabout.", "");
+#endif
 }
 
 // enter_roundabout_verbal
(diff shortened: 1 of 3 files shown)


test/gurka/test_instructions_roundabout.cc
```diff
@@ -69,11 +69,13 @@ TEST_F(InstructionsRoundabout, RoundaboutEnterOnly) {
 
   // TODO: known issue - future update to end on a roundabout
   //  Verify the enter_roundabout instructions
-  //      gurka::assert::raw::expect_instructions_at_maneuver_index(result, maneuver_index,
-  //                                                                "Enter the roundabout.", "Enter
-  //                                                                the roundabout.", "Enter the
-  //                                                                roundabout.", "Enter the
-  //                                                                roundabout.", "");
+#if 0
```
nilsnolde, line 72: was a "multiline comment" warning? I always wondered why that would be a _compiler_ warning.. to me, it's kinda a case in point that we shouldn't use `-WAll/Extra` at all and instead choose the few compiler warnings we actually want to care about. there's many very good ones, but this isn't one of them IMHO. 

that would also solve the annoying thing where we semi-regularly break the `-WAll` CI builds because some compiler update introduced some esoteric new warning which we then have to comply with.

it's just unfortunately a serious PITA to write out all interesting warnings.. on top of that clang & gcc here and there have different names for the same warning! thanks for that.. 

so yeah, in summary, I just wanna express how much I loathe these nonsense warnings :smile: I do understand why you do `if 0` (easy to grep, won't be compiled) and it makes sense here instead of removing it seems, reading the comment. I still hope that we don't have to do those changes too much

nilsnolde (approved):

test/gurka/test_instructions_roundabout.cc
kevinkreiser: my two cents: i hate this convention :smile:

there is nothing wrong with commenting something out. if the compiler is warning about this then as nils said we have it configured wrong. this is just way way too pedantic to be reasonable. can we avoid doing more of this?

if we desperately want to use the preprocessor to disable this code we should define a macro for tests to use like `NON_WORKING_TEST_CODE` or something. the idea with these things are kind of like the author leaving a sketch of something that is TODOs but isnt ready yet. having `ifdef 0` looks more like stuff you leave around while developing but want to actually delete before you PR the change

PR #4663, opened 2024-03-26
fixed connection check for costmatrix when the connection edge is on correlated edge

nilsnolde:
in costmatrix we already had a special check for trivial routes, but not for one where the connecting edge is the correlated edge on one tree, but not the other. statistically that should happen more often than real trivial routes (i.e. one edge routes). in that case we were just returning kinda bogus.

### DIFF
--- a/src/thor/bidirectional_astar.cc
+++ b/src/thor/bidirectional_astar.cc
@@ -805,15 +805,15 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
 // search tree. Check if this is the best connection so far and set the
 // search threshold.
 bool BidirectionalAStar::SetForwardConnection(GraphReader& graphreader, const BDEdgeLabel& pred) {
-  // Find pred on opposite side
-  GraphId oppedge = pred.opp_edgeid();
-  EdgeStatusInfo oppedgestatus = edgestatus_reverse_.Get(oppedge);
-  auto opp_pred = edgelabels_reverse_[oppedgestatus.index()];
-
   // Disallow connections that are part of an uturn on an internal edge
   if (pred.internal_turn() != InternalTurn::kNoTurn) {
     return false;
   }
+
+  // Find pred on opposite side
+  GraphId oppedge = pred.opp_edgeid();
+  EdgeStatusInfo oppedgestatus = edgestatus_reverse_.Get(oppedge);
+  auto opp_pred = edgelabels_reverse_[oppedgestatus.index()];
 
   // Disallow connections that are part of a complex restriction
   if (pred.on_complex_rest()) {
@@ -837,8 +837,11 @@ bool BidirectionalAStar::SetForwardConnection(GraphReader& graphreader, const BD
   } else {
     // If no predecessor on the forward path get the predecessor on
     // the reverse path to form the cost.
-    uint32_t predidx = opp_pred.predecessor();
-    float oppcost = (predidx == kInvalidLabel) ? 0 : edgelabels_reverse_[predidx].cost().cost;
+    uint32_t opp_predidx = opp_pred.predecessor();
+    // TODO(nils): this can't be happen, it'd mean that this is a trivial route
+    // , which is not supposed to be here! same in SetReverseConnection
+    float oppcost =
+        (opp_predidx == kInvalidLabel) ? 0.f : edgelabels_reverse_[opp_predidx].cost().cost;
     c = pred.cost().cost + oppcost + opp_pred.transition_cost().cost;
   }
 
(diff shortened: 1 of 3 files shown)


src/thor/bidirectional_astar.cc
```diff
@@ -805,16 +805,16 @@ BidirectionalAStar::GetBestPath(valhalla::Location& origin,
...
   // Disallow connections that are part of an uturn on an internal edge
   if (pred.internal_turn() != InternalTurn::kNoTurn) {
     return false;
   }
 
+  // Find pred on opposite side
+  GraphId oppedge = pred.opp_edgeid();
+  EdgeStatusInfo oppedgestatus = edgestatus_reverse_.Get(oppedge);
+  auto opp_pred = edgelabels_reverse_[oppedgestatus.index()];
+
```
nilsnolde, line 817: the diff is strange.. I moved the block above that, not this one.. we should early-return on 
```
  if (pred.internal_turn() != InternalTurn::kNoTurn) {
    return false;
  }
```

src/thor/bidirectional_astar.cc
```diff
@@ -837,8 +837,11 @@ bool BidirectionalAStar::SetForwardConnection(GraphReader& graphreader, const BD
   } else {
     // If no predecessor on the forward path get the predecessor on
     // the reverse path to form the cost.
-    uint32_t predidx = opp_pred.predecessor();
-    float oppcost = (predidx == kInvalidLabel) ? 0 : edgelabels_reverse_[predidx].cost().cost;
+    uint32_t opp_predidx = opp_pred.predecessor();
+    // TODO(nils): this can't be happen, it'd mean that this is a trivial route
+    // , which is not supposed to be here! same in SetReverseConnection
+    float oppcost =
+        (opp_predidx == kInvalidLabel) ? 0.f : edgelabels_reverse_[opp_predidx].cost().cost;
```
nilsnolde, line 844: this check shouldn't be necessary, it'd mean we have a trivial route in bidir a*

src/thor/costmatrix.cc
```diff
@@ -896,29 +904,36 @@ void CostMatrix::CheckReverseConnections(const uint32_t target,
         uint32_t d = std::abs(static_cast<int>(rev_pred.path_distance()) +
                               static_cast<int>(fwd_label.path_distance()) -
                               static_cast<int>(fwd_label.transition_cost().secs));
-        best_connection_[source_idx].Update(fwd_edgeid, rev_pred.edgeid(), Cost(s, s), d);
-        best_connection_[source_idx].found = true;
+        best_connection_[idx].Update(fwd_edgeid, rev_pred.edgeid(), Cost(s, s), d);
```
nilsnolde, line 907: this is mostly changing `source_idx` name to `idx`; it's the connection's index, not the source index

src/thor/costmatrix.cc
```diff
@@ -896,29 +904,36 @@ void CostMatrix::CheckReverseConnections(const uint32_t target,
...
+        best_connection_[idx].found = true;
+
+        // Update status and update threshold if this is the last location
+        // to find for this source or target
+        UpdateStatus(source, target);
+      } else if (fwd_predidx == kInvalidLabel) {
```
nilsnolde, line 913: this is the actual bug fix. before we'd go directly into the `else` below in case it wasn't a trivial route. but that block of code looks at the opposing's `transition_cost`, which is not what we'd expect if the opposing is a correlated edge: in `SetTargets` we use `transition_cost` to store the full edge's length and the full duration for it. 

so this needs to be handled in a different way to get the connection's length & duration

src/thor/costmatrix.cc
```diff
@@ -896,29 +904,36 @@ void CostMatrix::CheckReverseConnections(const uint32_t target,
...
+        best_connection_[idx].found = true;
 
         // Update status and update threshold if this is the last location
         // to find for this source or target
         UpdateStatus(source, target);
       } else {
-        float oppcost = (fwd_predidx == kInvalidLabel) ? 0 : fwd_edgelabels[fwd_predidx].cost().cost;
-        float c = rev_pred.cost().cost + oppcost + fwd_label.transition_cost().cost;
+        auto opp_pred_label = fwd_edgelabels[fwd_predidx];
+        auto cost = rev_pred.cost() + opp_pred_label.cost() + fwd_label.transition_cost();
```
nilsnolde, line 928: the `else` I just refactored to be terser

nilsnolde:
I was preparing for the bidir A* for costmatrix and reviewing the costmatrix code for the zillionth time. each time I find smth else! I doubt this was the last time :sweat_smile: 

I guess I should add a test for this..

nilsnolde:
I can't reproduce this behavior in gurka. I can see that it finds a connection which uses that `transition_cost` of an initial edge but it also finds other connections for the same edge combination which have different predecessors, so it always ends up being fine.

another PR to put on hold. in theory it's a problem but until it's not proven I won't spend time on it.

PR #4669, opened 2024-04-01
Significantly improve performance of graphbuilder

lseelenbinder:
# Issue

Since we do frequent planet builds, we have a solid pulse of the performance of tile building on similar hardware. In our experience, the performance of tile building has always been pretty consistent until a few weeks ago something changed, and it dropped off a cliff (roughly 2x increase in build times).

I went on a fact finding mission and turned up a few significant improvements based on profiling. This iterate is focused on the `BuildTileSet` portion of building, but I believe similar techniques could be used to improve other portions. Based on the latest profile, `enhance` is now the most costly (in terms of CPU time) part of tile build.

Things I did:

- Switched to using a queue for the graph tile builder. With high enough concurrency, worker threads ran out of work using the iterator approach (not all tiles are the same complexity), so pretty frequently a planet build would be stuck with only 4 of 12 worker threads doing anything (or worse) for hours. A queue keeps all the workers busy for much longer.
- Introduced a `R*tree` for holding language polys and then based on that improvement realized that `boost::geometry::covers` algorithm strategy is by default quite bad, so I switched to a better algorithm.
- After fixing that up, a mutex in sqlite became the bottleneck, and I removed global memory stat tracking for sqlite to remove that issue.
- I also removed some duplication of code and work by preprocessing the language tags per admin poly instead of per node that used a given admin poly.
- Lots of minor clang tidy stuff.

## Results

For a build of a recent Switzerland extract, graph tile build times of my machine dropped from ~6m to 40s:

```
# BEFORE
2024/04/01 10:19:05.957162 [INFO] Building 118 tiles with 12 threads...
2024/04/01 10:25:24.672623 [INFO] Finished

# AFTER
2024/04/01 10:13:31.291478 [INFO] Building 118 tiles with 12 threads...
2024/04/01 10:14:11.345470 [INFO] Finished
```

Flamegraph before:
image

Flamegraph after:
image

## Tasklist

 - [ ] Add tests

## Requirements / Relations


src/mjolnir/graphbuilder.cc
```diff
@@ -1449,7 +1458,12 @@ void GraphBuilder::Build(const boost::property_tree::ptree& pt,
   unsigned int threads =
       std::max(static_cast<unsigned int>(1),
                pt.get<unsigned int>("mjolnir.concurrency", std::thread::hardware_concurrency()));
-  std::string tile_dir = pt.get<std::string>("mjolnir.tile_dir");
+
+  auto tile_dir = pt.get<std::string>("mjolnir.tile_dir");
+  // Disable sqlite3 internal memory tracking (results in a high-contention mutex, and we don't care
+  // about marginal sqlite memory usage).
+  sqlite3_config(SQLITE_CONFIG_MEMSTATUS, false);
```
kevinkreiser, line 1465: i never noticed this option before, how much of an impact would you say it had?

kevinkreiser:
would you mind throwing in a changelog entry? this looks good to me

valhalla/mjolnir/admin.h
```diff
@@ -18,13 +18,18 @@
 using namespace valhalla::baldr;
 using namespace valhalla::midgard;
 
-namespace valhalla {
-namespace mjolnir {
+namespace valhalla::mjolnir {
+
+namespace bg = boost::geometry;
```
nilsnolde, line 23: I don't think we should change this to `valhalla::mjolnir`. I get that it's terser, but I'd rather continue with the same convention as in the rest of the project.

nilsnolde: not sure it makes a real difference but usually we keep those namespace aliases outside of our own namespaces

src/mjolnir/admin.cc
```diff
@@ -8,8 +8,7 @@
 
 #include <spatialite.h>
 
-namespace valhalla {
-namespace mjolnir {
+namespace valhalla::mjolnir {
```
nilsnolde, line 11: same as below: let's keep the previous convention

src/mjolnir/graphbuilder.cc
lseelenbinder: At higher levels of concurrency (> 8), sys time dominates once the `covered_by` performance is fixed.

High-level root cause:
- The `ST_AsText(geom)` from the admin queries generates a *lot* of text
- Building the return values results in a lot of memory allocations while fetching/building the polygons' WKT.
- [SQLite3 tracks memory stats](https://www.sqlite.org/malloc.html#_memory_status) at a global level (behind a single `static` mutex).
- The mutex spinlock becomes the bottleneck around returning the admin queries (i.e., a lot of system time, and more than 50% of the profile samples are in that spinlock).

We could also fix this long-term by adjusting the approach in any of the following ways:
- Directly using SQLite for the PiP lookup. I think this might be faster because we don't have to fetch and parse for every admin poly for every tile, and then run the PiP for every node.
- Switching from `boost::geometry` to `geos` to allow us to return WKB for less (de-)serialization cost.

lseelenbinder:
> would you mind throwing in a changelog entry? this looks good to me

Done 👍

nilsnolde (changes requested):
looks good, thanks. I'd just ask to revert the namespace changes (`valhalla::mjolnir`), rather sticking with our convention, even it's not the tersest.

> until a few weeks ago something changed, and it dropped off a cliff (roughly 2x increase in build times).

hm, that'd be not so good.. can you approx say between which commits/dates that happened? I remember you're quite up-to-date usually right? sadly I made some experiments on the fossgis server, I can't currently see how long our planet build took. also I don't monitor performance regularly..

did you already have a chance to try on it on the planet as well?

src/mjolnir/admin.cc
```diff
@@ -37,7 +36,7 @@ uint32_t GetMultiPolyId(const std::multimap<uint32_t, multi_polygon_type>& polys
   uint32_t index = 0;
   point_type p(ll.lng(), ll.lat());
   for (const auto& poly : polys) {
-    if (boost::geometry::covered_by(p, poly.second)) {
+    if (bg::covered_by(p, poly.second, bg::strategy::within::crossings_multiply<point_type>())) {
```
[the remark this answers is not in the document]
lseelenbinder: The simple answer is it uses ray-casting from the point to see how many times it crosses the polygon. It's a variant of this: https://en.wikipedia.org/wiki/Point_in_polygon#Ray_casting_algorithm.

Default uses the winding algorithm.

lseelenbinder:
> looks good, thanks. I'd just ask to revert the namespace changes (`valhalla::mjolnir`), rather sticking with our convention, even it's not the tersest.

Done 😄

> > until a few weeks ago something changed, and it dropped off a cliff (roughly 2x increase in build times).
> 
> hm, that'd be not so good.. can you approx say between which commits/dates that happened? I remember you're quite up-to-date usually right? sadly I made some experiments on the fossgis server, I can't currently see how long our planet build took. also I don't monitor performance regularly..

We're 99% certain it's related to worker starvation (fixed by using the queue). We spent approximately 60-70% of the build time with only 50% of the workers doing anything. The other improvements were related to low-hanging fruit I found after running a profile and realizing the cause was mostly workers having nothing to do.

> did you already have a chance to try on it on the planet as well?

Not yet…we're in a weird state because of the LSV changes (#4475) we're working on merging into upstream.

nilsnolde:
> We're 99% certain it's related to worker starvation (fixed by using the queue)

ah ok, so it wasn't changing all of a sudden, rather a slowly degrading process then? 

I had the same thought a while back, but never acted on it. back then I wanted to sort the tiles for tile size before slicing them into the current threading approach, to give a better work distribution to the threads. but using a locked queue instead is as good IMO.

I'll check this out on the fossgis instance and try it for a round. no need to wait with merging, it's good either way.

lseelenbinder:
> > We're 99% certain it's related to worker starvation (fixed by using the queue)
> 
> ah ok, so it wasn't changing all of a sudden, rather a slowly degrading process then?

It did change rather suddenly, but we assume that's due to a slight change in the data causing the workload to get more badly allocated. We've seen significant changes in (wall clock) runtime based on adding or removing a worker, likely due to how well the work happened to be distributed.

nilsnolde (approved):

nilsnolde:
it'll take a while for vcpkg to sort it out I think. just fyi, not a show stopper IMO.

nilsnolde (approved):

gknisely (approved):

nilsnolde:
Yeah damn! I was just looking at the fossgis instance today and realized we built a planet in WAY less than 24 hours, like 16 or so! I just saw it but didn’t have a closer look into the logs or visiting the PR again. Anyways, that’s really great:) as far as I remember it used to be 30ish+ hours on that very same machine!

nilsnolde:
I remember thinking how Switzerland was an unfair comparison because of the amount of tiles and how many of those would be empty, elevating the effect of this PR. Im amazed at the consequences on the planet!

lseelenbinder:
Yes, the biggest impact was definitely the ability to keep all the worker threads busy during the multi-threaded build. In our build, the single-threaded parts of the process (e.g., parsing) dominated the total time, so we could increase concurrency to eek out a bit more performance, but I think we're at around ~16h best case for the planet based on how long the single-threaded portions take.

nilsnolde:
Yeah exactly, I can’t remember the actual graph build (not parsing, not enhancing) taking that long to make up for this difference. Once I’m back in civilization I’ll need to take a closer look again:)

Issue #3605, opened 2022-04-29
Incorrect value for cost_threshold in costmatrix 


monteirodom:
In costmatrix, the searches (forward and backward) are stopped when a cost.duration is higher than a computed threshold.
That threshold is computed as (auto mode): `cost_threshold = max_matrix_distance / kCostThresholdAutoDivisor;`
As the check is based on a duration, that divisor is equivalent to a speed (56 km/h in auto mode).
> In accordance with the much more explicit computation in timedistancematrix

But that speed is in km/h and the cost.duration is in seconds. But the computation of the threshold lacks the conversion of the units (_again that conversion is explicit in timedistancematrix_).
In consequence, the many to many searches stop far too early, unless we set max_matrix_distance with a huge value.

The issue was already raised in #1554, but the fix in #1555 didn't address the problem correctly.
> It used a `2.0` coefficient, but the right conversion from km/h to m/s should be `3.6`

That issue arises much particularly for routes with ferries, as usually the ferry edge is very long and has a very low speed.

By the way, why convert a distance into a time for the check? Yo **do need** a hipothesis of  an average speed. The check can be done with a distance, by replacing :
`if (pred.cost().secs > current_cost_threshold_)`
with 
`if (pred.path_distance() > max_matrix_distance)`
Which is much more meaningful with the parameter in the setup.

monteirodom:
I'm also concerned by that one ... wich mixes duration with cost
```
-        float s = std::abs(pred.cost().secs + opp_el.cost().secs - opp_el.transition_cost());
+        // TODO: shouldnt this use seconds? why is this using cost!?
+        float s = std::abs(pred.cost().secs + opp_el.cost().secs - opp_el.transition_cost().cost);
```
The TODO was added by @kevinkreiser at #2425, but left as is since then.

kevinkreiser:
yeah i noticed that while doing another refactor. instead of changing the behavior i decided to adapt to just the refactor and let the TODO for later. does look like we should take another bug scrub over the code

monteirodom:
Is there anyone working on this issue?
I can come with a PR with the fix described on the issue. But I'm not very easy about the consequences on this part of code that uses the threshold for bucket allocation:
```
    // Allocate the adjacency list and hierarchy limits for target location.
    // Use the cost threshold to size the adjacency list.
    target_adjacency_.emplace_back(DoubleBucketQueue<BDEdgeLabel>(0, current_cost_threshold_,
[...]

PR #4671, opened 2024-04-03
astar costmatrix

### LINKED ISSUE #4445: Try A* for bidir matrix
We talked about it offline. While we're not exactly agreeing on the optimality of the approach (also no one tried to write a proof), I'd like to see the expansion result of this. Basically: for each source, when we expand an edge, we take the closest target as the A* heuristic for the sortcost. We start off directing the search to the closest target to each source, once that's found, we direct it to the next closest target etc. While that does change the significance of the queue order immediately after a target was found, it does IMO not invalidate any costed edge thus far and should only result in some "wasted" expansion where the edges close to the previously found target are still being popped off. Vice versa for expanding from the targets.

I'm not sure I could even attempt to "write a proof". I guess I'd be happy if the results of a test file would be identical to today. Performance-wise this should be quite a lot better, nothing can be as bad as Dijkstra:)

nilsnolde:
fixes #4445 

it's not that many changes actually. most tests passed other than 1 small test in `test/matrix.cc` which makes sense that it failed now. it's pretty straight-forward, the most important points are:

- initialize a vector of astar heuristics, one for each source & target
- for each source we initialize the heuristic with the cost to the _closest_ target, and vice versa
- when we want to use it for cost sorting of an edge for a source expansion, we use the _minimum_ heuristic of that edge to each target, and vice versa

that's really it. the heuristics are at least admissible. note, we never change the sortcost of existing labels once the closest target/source was found. I can imagine scenarios where it _might_ find less optimal routes, e.g. when the closest target is very close but requires a LOT of expansion (e.g. a waterbody needing a ferry) and another target's expansion "randomly" meets that source's "misguided" expansion.

FWIW, I already experimented a lot with it, my observations so far:
- EDIT: forgot the most important part, the performance improvements! since many PRs I have a kind of benchmark request file for matrix, 20 random locations (symmetrical) within Germany per request and 50 requests. with that file, request time decreased ~60%. that's better than I was hoping for really. I didn't measure it, but the memory requirements should scale down very similarly, less expansion means less labels
- for long distances I couldn't make out any differences in route geometry
- for urban environments we do get quite a lot of differences. however, those differences currently also exist with bidir a* routing and this PR makes those routes closer to routing. for the few ones I investigated a bit more, I could boil it down to the fact that we add some transitioncost via `stopimpact` during expansion. however, now we definitely use less shortcuts for short-ish distances compared to dijkstra, since we do a lot less expansion. and shortcuts don't have that `stopimpact` added (though it does have the `OSRMTurnDuration` stuff). the reason is that we only add `stopimpact` for cost and `OSRMTurnDuration` only for secs. when I commented out that bit of code, the routes were the same. that sucks somewhat and deserves its [own issue](https://github.com/valhalla/valhalla/issues/4672) IMO.
- what I can see consistently is that the route distance is almost always off by a few meters. @dnesbitt61 just recently got rid of some shortcut inaccuracies, not sure where that comes from now. /route doesn't have that problem


src/thor/costmatrix.cc
```diff
@@ -190,7 +193,7 @@ bool CostMatrix::SourceToTarget(Api& request,
               if (targets.empty() && locs_status_[MATRIX_FORW][source].threshold > 0) {
                 // TODO(nils): shouldn't we extend the search here similar to bidir A*
                 //   i.e. if pruning was disabled we extend the search in the other direction
-                locs_status_[MATRIX_FORW][i].threshold = -1;
+                locs_status_[MATRIX_FORW][source].threshold = -1;
```
nilsnolde, line 196: this was a bug, I wonder how it never segfaulted!

src/thor/costmatrix.cc
```diff
@@ -318,6 +321,8 @@ void CostMatrix::Initialize(
 
   locs_count_[MATRIX_FORW] = source_locations.size();
   locs_count_[MATRIX_REV] = target_locations.size();
+  astar_heuristics_[MATRIX_FORW].resize(target_locations.size());
```
nilsnolde, line 324: for each **source** we create one heuristic for the **reverse tree** and vice versa

src/thor/costmatrix.cc
```diff
@@ -326,22 +331,40 @@ void CostMatrix::Initialize(
...
-  // Add initial sources status
-  for (const auto exp_dir : {MATRIX_FORW, MATRIX_REV}) {
-    const auto count = locs_count_[exp_dir];
-    locs_status_[exp_dir].reserve(count);
-    hierarchy_limits_[exp_dir].resize(count);
-    adjacency_[exp_dir].resize(count);
-    edgestatus_[exp_dir].resize(count);
-    edgelabel_[exp_dir].resize(count);
+  const uint32_t bucketsize = costing_->UnitSize();
+  const float range = kBucketCount * bucketsize;
```
nilsnolde, line 335: changed this to be the bucket size as in bidirectional_astar.cc. this might change again, see https://github.com/valhalla/valhalla/issues/3605#issuecomment-2028938018

src/thor/costmatrix.cc
```diff
@@ -326,22 +331,40 @@ void CostMatrix::Initialize(
...
-      adjacency_[exp_dir][i].reuse(0, current_cost_threshold_, costing_->UnitSize(),
-                                   &edgelabel_[exp_dir][i]);
-      locs_status_[exp_dir].emplace_back(kMaxThreshold);
-      hierarchy_limits_[exp_dir][i] = hlimits;
+      edgelabel_[is_fwd][i].reserve(max_reserved_labels_count_);
+      locs_status_[is_fwd].emplace_back(kMaxThreshold);
+      hierarchy_limits_[is_fwd][i] = hlimits;
+      // for each source/target init the other direction's astar heuristic
+      auto& ll = locations[i].ll();
+      astar_heuristics_[!is_fwd][i].Init({ll.lng(), ll.lat()}, costing_->AStarCostFactor());
```
nilsnolde, line 358: so a source create a heuristic for the reverse tree, so that each edge on the reverse search can find the clostest/lowest-cost source for the sortcost

src/thor/costmatrix.cc
```diff
@@ -326,22 +331,40 @@ void CostMatrix::Initialize(
...
+
+      // get the min heuristic to all targets/sources for this source's/target's adjacency list
+      float min_heuristic = std::numeric_limits<float>::max();
+      for (uint32_t j = 0; j < other_count; j++) {
+        auto& other_ll = other_locations[j].ll();
+        auto heuristic = astar_heuristics_[!is_fwd][i].Get({other_ll.lng(), other_ll.lat()});
+        min_heuristic = std::min(min_heuristic, heuristic);
+      }
```
nilsnolde, line 366: and to initialize the queue we get the mincost to the closest target for each source. and vice versa

src/thor/costmatrix.cc
```diff
@@ -429,7 +452,7 @@ bool CostMatrix::ExpandInner(baldr::GraphReader& graphreader,
     // edges while still expanding on the next level since we can still transition down to
     // that level. If using a shortcut, set the shortcuts mask. Skip if this is a regular
     // edge superseded by a shortcut.
-    if (hierarchy_limits_[FORWARD][index][meta.edge_id.level() + 1].StopExpanding()) {
+    if (hierarchy_limits_[FORWARD][index][meta.edge_id.level() + 1].StopExpanding(pred.distance())) {
```
nilsnolde, line 455: now that we're using astar we can also disable hierarchy limits within a few kilometers of source/target to try and get higher quality routes.

src/thor/costmatrix.cc
```diff
@@ -442,6 +465,8 @@ bool CostMatrix::ExpandInner(baldr::GraphReader& graphreader,
     return true;
   }
 
+  // TODO(nils): refactor this whole thing to only have a single if (FORWARD) {}
```
nilsnolde, line 468: there's a lot of ternaries in this algorithm, most of which can be refactored away. probably the same in most other algorithms.

src/thor/costmatrix.cc
```diff
@@ -516,12 +547,23 @@ bool CostMatrix::ExpandInner(baldr::GraphReader& graphreader,
...
   *meta.edge_status = {EdgeSet::kTemporary, idx};
   if (FORWARD) {
-    edgelabels.emplace_back(pred_idx, meta.edge_id, opp_edge_id, meta.edge, newcost, mode_, tc,
-                            pred_dist, not_thru_pruning,
+    if (hierarchy_limits_[MATRIX_FORW][index][meta.edge_id.level()].max_up_transitions !=
+        kUnlimitedTransitions) {
+      // Override distance to the destination with a distance from the origin.
+      // It will be used by hierarchy limits
+      dist = astar_heuristics_[MATRIX_REV][index].GetDistance(t2->get_node_ll(meta.edge->endnode()));
+    }
```
nilsnolde, line 564: same logic as in bidirectional_astar.cc: don't need the heuristic distance if we're not having any hierarchy limits. though I wonder if that couldn't be made a constant for each request; AFAIK we never change to `kUnlimitedTransitions` dynamically..

test/matrix.cc
```diff
@@ -256,7 +256,7 @@ TEST(Matrix, test_matrix) {
       ++found;
     }
   }
-  EXPECT_EQ(found, 10) << " not the number of results as expected";
+  EXPECT_EQ(found, 11) << " not the number of results as expected";
```
nilsnolde, line 259: this tested that we find less connections once we decrease the max matrix distance in the config. seems there was a close case and a* actually finds it because it's better directed.

src/thor/costmatrix.cc
nilsnolde: hmpf, this is a serious performance drainer though.. I just ran some tests again and all of a sudden that one test file took 100 secs instead of 60 secs.. that's a massive slowdown, also considering we didn't do that before either. I'll run a few more long-ish urban tests to see the change in route quality (taking bidir a* /route as reference).

nilsnolde:
just to have it more visible:

including a distance to the source/target when determining hierarchy limits (as is done for /route bidir a*) results in a massive slowdown (at least up to 60% slower, e.g. 100 secs instead of 60 secs).

kevinkreiser:
ive always been interested to try this but the guy on my shoulder always talked me out of it because of its "technical" incorrectness. its great to hear that in practice is pretty much equivalent to running individual bidir a* requests. would be interesting to see those results actually!

src/thor/costmatrix.cc
nilsnolde: talked quickly with @kevinkreiser: will remove and leave a comment for future us

nilsnolde:
I can try to paste some pics of it. But maybe it’s even better if I update the fossgis instance with this PR, then everyone can play around a bit. If you use our qgis plugin, I recently added the matrix geometries there as well for debugging, which proved invaluable to make sense of changed routes.

kevinkreiser:
was thinking about this last night and again I can't rationalize how this works. all of the expansion you do with one goal heuristic which is part of the cost, is instantly invalidated once you switch goals. that means new labels won't sort properly. labels near the previous goal will be expanded and purged basically before you start using the new ones or I guess they will at least compete for a time with the old labels... I guess worst case it's like running back to back a* for each target but best case some targets are near each other and you can use some of the expansion that you did. using that expansion though with the wrong sort cost would have to mean suboptimal paths I think... or maybe just suboptimal expansion but still optimal paths if we do enough expansions which is still less expansion than dijkstras?

is there a way we could opt into this behavior just to avoid pitfalls later or is there a way to understand it as harmless that im not grasping.

kevinkreiser:
yeah maybe that's the key suboptimal sort cost leads to suboptimal expansion but the actual cost is still correct so not suboptimal paths. is that the way to graph this properly (barring any discussion about other pruning we do that affects correctness)

nilsnolde:
> all of the expansion you do with one goal heuristic which is part of the cost, is instantly invalidated once you switch goals. 

not quite. because we do not use a _single_ heuristic for an entire expansion, but for each edge we use the heuristic of the closest target/source (fwd/rev), so no edge is ever costed "wrongly" and nothing should ever be invalidated in terms of sortcost.

> yeah maybe that's the key suboptimal sort cost leads to suboptimal expansion but the actual cost is still correct so not suboptimal paths. is that the way to grok this properly (barring any discussion about other pruning we do that affects correctness)

right, that's how I'd express it. I can still see how this could potentially lead to suboptimal paths in edge-case scenarios, e.g. what I describe in the PR description:
> I can imagine scenarios where it might find less optimal routes, e.g. when the closest target is very close but requires a LOT of expansion (e.g. a waterbody needing a ferry) and another target's expansion "randomly" meets that source's "misguided" expansion.

nilsnolde:
once a target is found, the edges around it, which were put in the queue before, still have a very low sortcost, which is technically wrong at that moment, because the next target is much further away than their sortcost suggests. however, once they're settled, the next expanded edges on those branches will have a high sortcost, which basically leads to a few redundant edges being in the queue, but doesn't hurt too much IMO.

nilsnolde:
>  I guess worst case it's like running back to back a* for each target but best case some targets are near each other and you can use some of the expansion that you did.

and fwiw this is how I think about it too. but in a many to many scenario the worst case very rarely really happens.

nilsnolde:
Btw, if you install our QGIS plugin and request against the fossgis server, you can play around a bit for yourself and try to break it;) the QGIS plugin has the matrix geometry stuff implemented so you’ll actually see the paths it finds. 

Of course you can’t compare it on the same instance to Dijkstra CostMatrix, but at least to routing results. In an earlier version, and mostly for dev, I had a request boolean “use_costmatrix_astar”, which would toggle astar vs dijkstra. It’s fugly to do that in a single algorithm so for convenience I duplicated the entire CostMatrix to CostMatrixAstar. My hope was that we don’t have to do this and can live with the (very) rare occasions where A* CostMatrix would return sub-par results compared to full Dijkstra.

src/thor/costmatrix.cc
kevinkreiser: the second part of the condition must be very uncommon? code coverage says none of our tests make it happen either

src/thor/costmatrix.cc
kevinkreiser: i wonder if this matters at all to performance or if we are all memory bound anyway. would be interesting if we used a fast approximate sqrt in the heuristic instead of actual sqrt like we do today

kevinkreiser: (i guess i dont mean this specific instance of intitializing but when its calling `GetAstarHeuristic` in the tight loop)

kevinkreiser (approved):

kevinkreiser:
what was the bit that we talked about thats missing from this that exists in the regular bidir algorithm, hierarchy limits? dont forget to write a note about it in the relevant place in the code :pray:

src/thor/costmatrix.cc
nilsnolde: fwiw now, left a comment here: https://github.com/valhalla/valhalla/pull/4671/files#diff-a2ca5ea82be964856f22db7fe58defcbe2f92e6ec7646566e5040dfef9c8229eR546

nilsnolde: fwiw now, left a comment here: https://github.com/valhalla/valhalla/pull/4671/files#diff-a2ca5ea82be964856f22db7fe58defcbe2f92e6ec7646566e5040dfef9c8229eR546

src/thor/costmatrix.cc
nilsnolde: > would be interesting if we used a fast approximate sqrt in the heuristic instead of actual sqrt like we do today

good point. even just working in integer space might already speed up? if there's accuracy to the meter, that'd be good enough right?

PR #4674, opened 2024-04-03
change costmatrix max_distance threshold to a distance threshold instead of duration

nilsnolde:
fixes https://github.com/valhalla/valhalla/issues/3605


src/thor/costmatrix.cc
```diff
@@ -583,7 +583,7 @@ bool CostMatrix::Expand(const uint32_t index,
 
   // Get edge label and check cost threshold
   auto pred = edgelabels[pred_idx];
-  if (pred.cost().secs > current_cost_threshold_) {
+  if (pred.path_distance() > current_pathdist_threshold_) {
```
nilsnolde, line 586: as discussed in https://github.com/valhalla/valhalla/issues/3605, it's much more intuitive to use the path distance

src/thor/costmatrix.cc
```diff
@@ -338,8 +338,8 @@ void CostMatrix::Initialize(
       // Allocate the adjacency list and hierarchy limits for this source.
       // Use the cost threshold to size the adjacency list.
       edgelabel_[exp_dir][i].reserve(max_reserved_labels_count_);
-      adjacency_[exp_dir][i].reuse(0, current_cost_threshold_, costing_->UnitSize(),
-                                   &edgelabel_[exp_dir][i]);
+      adjacency_[exp_dir][i].reuse(0, GetBucketRange(current_pathdist_threshold_),
```
nilsnolde, line 341: I'd still take the previous logic to estimate the cost range for the queue. or do what bidirectional_astar does and just make it a constant (20k currently)?

src/thor/costmatrix.cc
nilsnolde: Since this is per expansion, we should probably take half the value?

nilsnolde: We have some tests for this which are failing right now. Probably can’t make them pass entirely, but I’d try to get close to that

src/thor/costmatrix.cc
```diff
@@ -147,7 +127,7 @@ bool CostMatrix::SourceToTarget(Api& request,
   auto& source_location_list = *request.mutable_options()->mutable_sources();
   auto& target_location_list = *request.mutable_options()->mutable_targets();
 
-  current_cost_threshold_ = GetCostThreshold(max_matrix_distance);
+  current_pathdist_threshold_ = max_matrix_distance / 2;
```
nilsnolde, line 130: half the max distance as threshold since we're bidirectional here

test/matrix.cc
```diff
@@ -231,7 +231,7 @@ TEST(Matrix, test_matrix) {
 
   CostMatrix cost_matrix_abort_source;
   cost_matrix_abort_source.SourceToTarget(request, reader, mode_costing, sif::TravelMode::kDrive,
-                                          90000.0);
+                                          7000.0);
```
nilsnolde, line 234: this here was testing that we don't find certain connections based on the max matrix distance. since we changed the concept to distance now, it seems we need to lower that significantly to have a similar effect.

nilsnolde:
this is ready for review too.

note, the timedistancematrix still has the same concept. I didn't change that (yet) because it's using the threshold differently than costmatrix. possibly it'll need to be refactored a bit before we could do the same to it.

valhalla/thor/costmatrix.h
```diff
@@ -164,11 +164,11 @@ class CostMatrix : public MatrixAlgorithm {
...
 
   /**
-   * Get the cost threshold based on the current mode and the max arc-length distance
-   * for that mode.
+   * Get the cost range for the doublebucket queue based on the current mode and the max arc-length
+   * distance for that mode.
    * @param  max_matrix_distance   Maximum arc-length distance for current mode.
    */
-  float GetCostThreshold(const float max_matrix_distance);
+  float GetBucketRange(const float max_matrix_distance);
```
kevinkreiser, line 171: you can just remove this. seems like you intended to replace it at first and then inline you just did the simple math instead, no function call needed

nilsnolde:
Oops, only removed it from the cc file it seems.. I left a todo, I think it’d make sense to have a „smarter“ bucketing of the queue, the CostMatrix at least.

nilsnolde:
removed the function from all files @kevinkreiser

kevinkreiser (approved):

PR #4670, opened 2024-04-02
update vpckg due to xz having a compromised release in previous version

nilsnolde:
ref https://github.com/microsoft/vcpkg/issues/37893

reading further it seems github actually disabled the whole repo! might not even work with this newer version.. in worst case win builds will be broken for a few days/weeks..

PR #4679, opened 2024-04-09
 Add vcpkg overlay port for xz package to fix windows CI workflow

michaelhrabanek:
Related to (https://github.com/valhalla/valhalla/pull/4670)

This PR introduces a vcpkg overlay port targeting a mirror repository of the xz compression library. Given recent concerns regarding the security of newer versions, this PR specifically utilizes version 5.4 of xz, which is currently not known to have any security vulnerabilities.


nilsnolde:
Thanks!

Don’t we have to tell vcpkg that we want to use the overlay?

vcpkg-configuration.json
```diff
@@ -0,0 +1,5 @@
+{
```
nilsnolde, line 1: Can this go into the existing vcpkg.json?

overlay-ports/liblzma/add_support_ios.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
nilsnolde, line 1: I’d call the new directory smth like “vcpkg-overlay-ports” to make it clearer.

vcpkg-configuration.json
michaelhrabanek: AFAIK it can't and you have to add vcpkg-configuration.json to use overlays (https://learn.microsoft.com/en-us/vcpkg/concepts/manifest-mode#configuration-file)

overlay-ports/liblzma/add_support_ios.patch
michaelhrabanek: That was my original plan, but since "vcpkg*/" is gitignored, I renamed it. I guess I can force it... or should we come with another name?

michaelhrabanek:
> Thanks!
> 
> Don’t we have to tell vcpkg that we want to use the overlay?

vcpkg looks for vcpkg-configuration.json in same directory as vcpkg.json and if it exists, it's included. While resolving packages, overlay ports take priority.

overlay-ports/liblzma/add_support_ios.patch
nilsnolde: > but since "vcpkg*/" is gitignored

hmpf right! then we can reverse it maybe, like `overlay-ports-vcpkg`, that should go through .gitignore

nilsnolde (commented):
it's all good actually, except for the directory name change. I think Win is failing due to some libexpat stuff which is unrelated to this PR. 

Can you try to update the vcpkg commit in [`.github/workflows/win.yml`](https://github.com/valhalla/valhalla/blob/1b5bea03a61597a60314056325da8eeaf87c5f8f/.github/workflows/win.yml#L34) to the most recent master commit in vcpkg? that should pull a newer version of expat.

michaelhrabanek:
> it's all good actually, except for the directory name change. I think Win is failing due to some libexpat stuff which is unrelated to this PR.
> 
Dir renamed. 
> Can you try to update the vcpkg commit in [`.github/workflows/win.yml`](https://github.com/valhalla/valhalla/blob/1b5bea03a61597a60314056325da8eeaf87c5f8f/.github/workflows/win.yml#L34) to the most recent master commit in vcpkg? that should pull a newer version of expat.

Actually, it was just some hiccup with libexpat repo, all working now. FYI that bminor xz mirror went down too - I changed the overlay to older, archived mirror, which never included that xz backdoor, so it should stay up hopefully.

nilsnolde (approved):

PR #4686, opened 2024-04-12
Consider more forward & backward tags  

### LINKED ISSUE #4434: More forward/backward tags
It's not very abundant, but for HGV again, it's important to be as good as we can respecting restrictions. `maxweight`/`maxheight` etc all have forward/backward tags. Obviously on commercial datasets much better populated than in OSM, where my main interest would be. It's not that hard to support, might as well do it.

chrstnbwnkl:
# Issue

Fixes #4434 

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/sif/autocost.cc
```diff
@@ -454,8 +454,8 @@ bool AutoCost::AllowedReverse(const baldr::DirectedEdge* edge,
     return false;
   }
 
-  return DynamicCost::EvaluateRestrictions(access_mask_, edge, false, tile, opp_edgeid, current_time,
-                                           tz_index, restriction_idx);
+  return DynamicCost::EvaluateRestrictions(access_mask_, opp_edge, false, tile, opp_edgeid,
```
chrstnbwnkl, line 457: There's a couple of these. Up until now, access restrictions were always symmetrical (meaning an edge's access restrictions were always equal to the ones of it's opposing edge), so it wasn't relevant that in `AllowedReverse`, the current edge's access restrictions were checked instead of the opposing one's.

nilsnolde: good catch :smile:

valhalla/baldr/graphconstants.h
```diff
@@ -742,6 +742,9 @@ enum class AccessType : uint8_t {
   kMaxAxles = 9
 };
 
+// Used for access restrictions. Conveys the direction in which the access restriction applies
+enum class AccessRestrictionDirection : uint8_t { kBoth = 0, kForward = 1, kBackward = 2 };
+
```
nilsnolde, line 747: can this go into valhalla/mjolnir/osmaccessrestriction.h?

nilsnolde (dismissed):
looks great, thanks! just a small change request on where to put an enum.

for others, note that this has little impact on OSM data where those attributes are hardly populated. however, in commercial data they do appear more frequently, especially the maxspeed:hgv directional ones.

nilsnolde (approved):

valhalla/mjolnir/osmaccessrestriction.h
```diff
@@ -63,7 +76,8 @@ class OSMAccessRestriction {
     uint16_t modes_ : 12;
   };
   Attributes attributes_;
-  uint16_t spare_[3];
+  AccessRestrictionDirection direction_;
+  uint16_t spare_[2];
```
chrstnbwnkl, line 80: add another `uint8_t

nilsnolde (approved):

PR #4687, opened 2024-04-13
Add missing docs

Trietes:
# Issue

Adding some of the missing docs here. Most of #3760 should be done now. Following are still left:
- `speed_types`: Don't know about these, so I won't document them.
- `restriction_probability`: Doesn't make sense to me in combination with OSM data. I would come to the same conclusion as mentioned in the issue.
- `mode_factor`: Don't know about this one, so I won't document it.

## Tasklist

 - [X] Update the docs with any new request parameters or changes to behavior described


docs/docs/api/isochrone/api-reference.md
```diff
@@ -65,20 +67,6 @@ When making a map, drawing the isochrone contours as lines is more straightforwa
 
 (TODO: write something about rendering the GeoTIFF output.)
 
-## Future work on the isochrone service
-
-The Isochrone service is in active development. To report software issues or suggest enhancements, open an issue in the [Valhalla GitHub repository](https://github.com/valhalla/valhalla/issues).
-
-Several other options are being considered as future service enhancements. These include:
-
-* ~~Using distance rather than time for each unit.~~
```
nilsnolde, line 74: Good catch! Nice to see those 7-8 year old TODOs were solved naturally over time:)

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -116,6 +123,7 @@ These options are available for `auto`, `bus`, and `truck` costing methods.
 | `country_crossing_cost` | A cost applied when encountering an international border. This cost is added to the estimated and elapsed times. The default cost is 600 seconds. |
 | `country_crossing_penalty` | A penalty applied for a country crossing. This penalty can be used to create paths that avoid spanning country boundaries. The default penalty is 0. |
 | `shortest` | Changes the metric to quasi-shortest, i.e. purely distance-based costing. Note, this will disable all other costings & penalties. Also note, `shortest` will not disable hierarchy pruning, leading to potentially sub-optimal routes for some costing models. The default is `false`. |
+| `use_distance` | A factor that allows controlling the contribution of distance and time to the route costs. The value is in range between 0 and 1, where 0 only takes time into account (default) and 1 only distance. A factor of 0.5 will weight them roughly equally. **Note:** this costing is currently not available for truck costing. |
```
nilsnolde, line 126: Actually it’s only implemented for car, also not motorcycle etc

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -276,6 +288,7 @@ Directions options should be specified at the top level of the JSON object.
 | `language` | The language of the narration instructions based on the [IETF BCP 47](https://tools.ietf.org/html/bcp47) language tag string. If no language is specified or the specified language is unsupported, United States-based English (en-US) is used. [Currently supported language list](#supported-language-tags) |
 | `directions_type` |  An enum with 3 values. <ul><li>`none` indicating no maneuvers or instructions should be returned.</li><li>`maneuvers` indicating that only maneuvers be returned.</li><li>`instructions` indicating that maneuvers with instructions should be returned (this is the default if not specified).</li></ul> |
 | `format` | Four options are available: <ul><li>`json` is default valhalla routing directions JSON format</li><li>`gpx` returns the route as a GPX (GPS exchange format) XML track</li><li>`osrm` creates a OSRM compatible route directions JSON</li><li>`pbf` formats the result using protocol buffers</li></ul> |
+| `shape_format` | If the format is `osrm`: Specifies the optional format for the path shape of each connection. One of `polyline6` (default), `polyline5`, `geojson` or `no_shape`. |
```
nilsnolde, line 291: shape_format can’t have the value „osrm“ IIRC. Also this is only implemented for some endpoints, I think trace_attributes and matrix only, so it’d rather belong there. Or you checked and it’s even implemented for /route?

valhalla/sif/dynamiccost.h
```diff
@@ -1017,7 +1017,7 @@ class DynamicCost {
   // Penalties that all costing methods support
   float maneuver_penalty_;         // Penalty (seconds) when inconsistent names
   float alley_penalty_;            // Penalty (seconds) to use a alley
-  float destination_only_penalty_; // Penalty (seconds) using private road, driveway, or parking aisle
+  float destination_only_penalty_; // Penalty (seconds) using destination only roads
```
nilsnolde, line 1020: In fact the previous comment was more correct, however missing that it also includes actual „destination“ tags.

nilsnolde (changes requested):
Thanks a lot for the effort! I think there’s just a few minor changes. I’ll do the ones you left soon as well!

docs/docs/api/turn-by-turn/api-reference.md
Trietes: I've just checked it again. Its little messy :D
- Matrix has it (except TDMatrix?!) and its already documented.. 
- Turn by turn also has it if you specify `"format":"osrm"`. I've just gave it a test and code can be seen here: https://github.com/valhalla/valhalla/blob/afd01613cd430bdc9550cfd7352dea7552004be5/src/tyr/route_serializer_osrm.cc#L280C1-L280C12
- Elevation has it and its documented.
- For trace_attributes I couldn't find anything. But I also didn't try it out.

valhalla/sif/dynamiccost.h
Trietes: True. I was completely focusing on destination only signs, but forgott about valhalla treating private roads (for good reasons) similar to destination only.

Trietes:
Thanks for the fast review! I've just added another commit to address your comments

docs/docs/api/turn-by-turn/api-reference.md
nilsnolde: Ah right elevation I meant, not trace_attributes.. Ok cool, didn’t know about the option for osrm, but I guess it makes sense as its default is polyline5. Then never mind!

nilsnolde:
Seems like the small change in the header file causes the formatting to fail. Can you run format.sh?

PR #4644, opened 2024-03-21
Removed voice & banner instructions from last step and added ssmlAnnouncements

Trietes:
# Issue

This PR is targetting to solve two of the issues mentioned in #4640. I've added simple voice ssmlAnnouncements to make the routing answers compatible with the navigation SDKs. I've also removed the voice and banner instructions from the last step since they are not required by the SDKs and just lead to confusing instructions on the screen.

## Tasklist

 - [X] Add tests
 - [X] Add #fixes with the issue number that this PR addresses
 - [x] Update the [changelog](CHANGELOG.md)

Issue #4640, opened 2024-03-19
Improved OSRM voice and banner instructions


Trietes:
I just had the time to test out the new banner and voice instructions of Valhalla in the Mapbox Navigation SDK. I've just did some testing and probably I'll try out Maplibre soon. First huge thanks for bringing it to Valhalla @eikes ! Was a quiet good experience to get it working.
I just want to collect few things I've noticed while using it, so it can be improved. Maybe somebody else also has some feedback, so we can collect it here.

- [x] The ssmlAnnouncement for the voice instructions seems to be required for the SDK. As a simple workaround I just wrapped the normal announcement into a opening and closing 'speak' tag and added it as ssmlAnnouncement. For example `"Your destination is on the right."` will be added as `"<speak>Your destination is on the right.</speak>"`. Might be interesting to add this into Valhalla to make it compatible and later on also have the possibility to add more detailed ssml instructions.
- [x] One thing which is little annoying is that there are too many voice instructions. Especially right before doing a turn the navigation system is talking constantly. Valhalla is returning e.g. this here: 
```
"voiceInstructions": [
    {
         "announcement": "Rechts abbiegen.",
         "distanceAlongGeometry": 76.3
     },
     {
         "announcement": "Rechts abbiegen. Dann Das Ziel wird erreicht.",
         "distanceAlongGeometry": 26.3
     }
 ]
```
This step is 100 meters long and has those two instructions during that distance. It could replaced with one voice instruction. Comparing it with other APIs, there is just the second instruction which is speaked out right in the middle between those two.
- [x] There shouldn't be any voice and banner instructions in the last step of the OSRM answer. This leads to having two banner instructions with "destination reached" in the banner view: 
![grafik](https://github.com/valhalla/valhalla/assets/33934221/fd213c10-0fbd-4783-afdb-5d2f12538f2a)

Trietes:
Just wanted to give a small update on this. PR https://github.com/valhalla/valhalla/pull/4644 is solving point 1&3 from the list.
Fine tuning the voice instructions itself is little more complicated. As already mentioned in the original PR which added voice instructions, it would e.g. need adding distance information for the alert voice instructions. I'm going to play around and experiment some things on a local build to improve the instructions and timings. I've to see if I'll have another PR on this. At least I'll share my experience with it here.

eikes:
Thank you for your appreciation.

[...]

PR #4696, opened 2024-04-19
Improve voice instructions

Trietes:
# Issue

Trying to do more work on #4640.

## Tasklist

 - [ ] Add tests
 - [ ] Do some more real-world testing

## Requirements / Relations

Also contains the changes made in #4644. So we can either skip #4644 and wait for this one or quickly merge as soon as we know that it works with Maplibre and later on merge this one.


src/tyr/route_serializer_osrm.cc
```diff
@@ -12,6 +12,8 @@
 #include "midgard/util.h"
 #include "odin/enhancedtrippath.h"
 #include "odin/util.h"
+#include "odin/narrative_builder_factory.h"
+#include "odin/narrativebuilder.h"
```
Trietes, line 16: Having these here is ofc debatable. We have following problem here: Currently we only build instructions during building the maneuver. But we decide the distance when a maneuver is played in the osrm_serializer and the instructions do depend on the distance (e.g. turn right in 200 meters). So the two options in my mind are:
- Calculate the distances here and add the 'IN 200 meters' information here. This needs the dependency and that's how I did it.
- Inject the distance information during maneuver building. So we later on just call here e.g. `verbal_transition_alert_instruction()` and we also get the distance when to play it.

src/tyr/route_serializer_osrm.cc
```diff
@@ -96,9 +98,11 @@ const constexpr PointLL::first_type DOUGLAS_PEUCKER_THRESHOLDS[19] = {
     2.6,      // z18
 };
 
-const constexpr double SECONDS_BEFORE_VERBAL_TRANSITION_ALERT_INSTRUCTION = 15.0;
-const constexpr double SECONDS_BEFORE_VERBAL_PRE_TRANSITION_INSTRUCTION = 5.0;
+const constexpr double SECONDS_BEFORE_VERBAL_TRANSITION_ALERT_INSTRUCTION = 35.0;
+const constexpr double SECONDS_BEFORE_VERBAL_PRE_TRANSITION_INSTRUCTION = 10.0;
```
Trietes, line 102: Just increased those values. 5 and 15 seconds sound nice in theory but on the road I experienced that this is way too late. Would be cool to get some feedback of somebody else on this

src/tyr/route_serializer_osrm.cc
```diff
@@ -96,9 +98,11 @@ const constexpr PointLL::first_type DOUGLAS_PEUCKER_THRESHOLDS[19] = {
...
+const constexpr double MIN_DISTANCE_VERBAL_PRE_TRANSITION_INSTRUCTION = 40;
```
Trietes, line 103: Adding a minimum distance for the pre transition alert here. I experienced that on low speed roads we will just announce right before the turn due to the low speed. That also often feels way too late. On such low speed roads its even fine to announce 30 seconds before the transition if you can already can see it.
One thing to take into account here is that a routing graph only reflects (ideally) the center of the road and doesn't represent the width of a road.
I'm not sure if I 100% like this concept, but it helped in my case

src/tyr/route_serializer_osrm.cc
```diff
@@ -1482,7 +1486,12 @@ json::ArrayPtr voice_instructions(const valhalla::DirectionsLeg::Maneuver* prev_
                                   const valhalla::DirectionsLeg::Maneuver& maneuver,
                                   const double distance,
                                   const uint32_t maneuver_index,
-                                  valhalla::odin::EnhancedTripLeg* etp) {
+                                  valhalla::odin::EnhancedTripLeg* etp,
+                                  const valhalla::Options& options) {
+  // narrative builder for custom pre alert instructions
+  MarkupFormatter nullFormatter;
+  std::unique_ptr<NarrativeBuilder> narrative_builder = NarrativeBuilderFactory::Create(options, etp, nullFormatter);
```
Trietes, line 1493: Also not 100% happy with this. The functions which we don't need the markup formatter, so I'm just passing a null formatter.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1494,65 +1503,94 @@ json::ArrayPtr voice_instructions(const valhalla::DirectionsLeg::Maneuver* prev_
   // shortly (10 meters at the given speed) after the maneuver has started.
   // The voice_instruction_beginning starts shortly after the beginning of the step.
   // The voice_instruction_end starts shortly before the end of the step.
-  float distance_before_verbal_transition_alert_instruction = -1;
-  float distance_before_verbal_pre_transition_instruction = -1;
-  if (prev_maneuver) {
```
Trietes, line 1499: Just moved this `if(prev_maneuver)` out of the function. We should always have it.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1494,65 +1503,94 @@ json::ArrayPtr voice_instructions(const valhalla::DirectionsLeg::Maneuver* prev_
...
+    voice_instruction_beginning->emplace("ssmlAnnouncement",
+                                      "<speak>" + prev_maneuver->verbal_post_transition_instruction() + "</speak>"); 
+    voice_instructions_array->emplace_back(std::move(voice_instruction_beginning));
+  }
+
+  // If there is an alert instruction and we have enough time to play it, we will play it
+  // TODO: don't add alert instruction if there was already a pre/post transition aleart which is close by
+  if (!maneuver.verbal_transition_alert_instruction().empty() 
+    && distance_before_verbal_transition_alert_instruction > -1.0
+    && prev_maneuver->time() > SECONDS_BEFORE_VERBAL_TRANSITION_ALERT_INSTRUCTION + APPROXIMATE_VERBAL_PRERANSITION_SECONDS) {
```
Trietes, line 1566: This is one of the most important changes. We will not have the alert instructions for short maneuvers. Doesn't make sense in my mind to alert something on a 50 meter road. Just play the transition instruction in that case is fine enough. Two instructions on that small maneuver are way too much. 
Actually we should even go further and collapse multiple maneuver into a single one if the maneuver itself is too short

src/tyr/route_serializer_osrm.cc
```diff
@@ -1494,65 +1503,94 @@ json::ArrayPtr voice_instructions(const valhalla::DirectionsLeg::Maneuver* prev_
...
       // as it is capped to the maneuver length
-      voice_instruction_end
-          ->emplace("distanceAlongGeometry",
-                    json::fixed_t{distance_before_verbal_transition_alert_instruction, 1});
+      distance_along_geometry = distance_before_verbal_transition_alert_instruction;
     }
-    voice_instruction_end->emplace("announcement", maneuver.verbal_transition_alert_instruction());
+    //building voice instructions for the alert
+    float distance_km = (float) distance_along_geometry / 1000.0f; 
+    std::string instruction = narrative_builder->FormVerbalAlertApproachInstruction(distance_km, maneuver.verbal_transition_alert_instruction());
```
Trietes, line 1580: Here is the part where I'm enhancing the alert instruction by the distance information. This needs the narrative_builder.

src/tyr/route_serializer_osrm.cc
```diff
@@ -1832,29 +1871,28 @@ json::ArrayPtr serialize_legs(const google::protobuf::RepeatedPtrField<valhalla:
 
       // Add banner instructions if the user requested them
       if (options.banner_instructions()) {
-        if (prev_step) {
+        if (prev_step && prev_maneuver) {
```
Trietes, line 1874: Just ensuring here that we have the prev_maneuver. Actually it should always be present

Trietes (commented):
It just a draft right now. Even if it works fine in my case it should be tested by somebody else to verify if it works in reality on the road.

ianthetechie (commented):
We'll need a review from one of the core maintainers to comment on the parts that you flagged as needing a design review, but throwing in my 2 cents, this 1) does appear to work (running it on some servers now after some local testing), and 2) is a significant improvement.

I'd suggest marking this ready for review @Trietes unless you're still waiting on something I missed.

Trietes:
Thanks for checking it out and giving it a test. I've been waiting for somebody to test it out and saying it is at least better than before :D 
I'm only concerned of including the narrative builder here. Works fine, but I don't know if it is good practice. If the maintainers are fine with it, I can add some tests and make the PR ready.

src/tyr/route_serializer_osrm.cc
kevinkreiser: yeah i think the long term way we should do it is the second way primarily because it is odins main function, deciding where maneuvers are. deciding when to warn about them is very closely related in my opinion.

having said that, you already did all the work and there has been slow progress on this effort in the past so i think it could stay this way until someone has time for a larger refactor. i would wager that others would agree especially because there are other odin deps mixed in here as well (which was already not great and was for the same reason).

as an aside.. maybe we should have had the actor stuff that tyr provides be part of the valhalla namespace exposed in the main valhalla header and then just have odin subsume all the serialization stuff (essentially dividing up tyrs parts and then removing tyr completely). i think it would be reasonably appropriate for odin to generate maneuvers narrative and do serialization. its essentially serializing the path into maneuvers and then serializing maneuvers into narrative. may as well do the final serialization into json or pbf bytes. this of course ignores that there are some endpoints that have no directions or narrative and maybe those would seem strange calling into odin.. meh just some thoughts :smile:

Trietes: Thanks for having a look! I completely agree. The second way feels more correct, but would need some more time. Would also be better so the normal valhalla json answer could be enhanced by the same distance information which might be helpful.

But then let me clean up this one and write some test. I'll also open a cleanup issue for doing it the second way. Maybe I'll have some time in the future to clean it up.

Trietes:
Seems like GitHub doesn't like my rebase. I'm just done with adapting the tests and cleaning the stuff up. I'll open a clean new pull request now

PR #4701, opened 2024-04-22
Address some "unused code" errors

michaelkirk:
# Issue

> What issue is this PR targeting? If there is no issue that addresses the problem, please open a corresponding issue and link it here.

I'm assuming a tracking issue is overly ceremonious in this case, but let me know.

## Tasklist

None of these seemed applicable, but let me know if I'm off base:

 - [ ] Add tests

### DIFF
--- a/src/thor/astar_bss.cc
+++ b/src/thor/astar_bss.cc
@@ -513,7 +513,7 @@ std::vector<PathInfo> AStarBSSAlgorithm::FormPath(baldr::GraphReader& graphreade
   // Work backwards from the destination
   std::vector<PathInfo> path;
   travel_mode_t old = travel_mode_t::kPedestrian;
-  int mode_change_count = 0;
+  [[maybe_unused]] int mode_change_count = 0;
 
   for (auto edgelabel_index = dest; edgelabel_index != kInvalidLabel;
        edgelabel_index = edgelabels_[edgelabel_index].predecessor()) {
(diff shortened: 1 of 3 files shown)


src/thor/astar_bss.cc
```diff
@@ -513,7 +513,9 @@ std::vector<PathInfo> AStarBSSAlgorithm::FormPath(baldr::GraphReader& graphreade
   // Work backwards from the destination
   std::vector<PathInfo> path;
   travel_mode_t old = travel_mode_t::kPedestrian;
+  #ifndef NDEBUG
   int mode_change_count = 0;
```
kevinkreiser, line 517: you can use `[[maybe_unused]]` instead

michaelkirk: done.

test/logging.cc
```diff
@@ -57,12 +57,12 @@ TEST(Logging, FileLoggerTest) {
     results.emplace_back(std::async(std::launch::async, work));
   }
 
-  // dont really care about the results but we can pretend
-  int exit_code = 0;
   for (auto& result : results) {
     try {
       result.get();
```
kevinkreiser, line 62: instead of doing try catch at all could you just do this:

```suggestion
      ASSERT_NO_THROW(result.get());
```

michaelkirk: done.

kevinkreiser:
its a shame ci cant run properly with your prs

PR #4705, opened 2024-04-23
rm unused import fixes building tests without prime_server

michaelkirk:
# Issue

The tests fail to build without prime_server, even with `-DENABLE_SERVICES=OFF -DENABLE_HTTP=OFF`.

I was going to do something fancier with cmake including different tests depending on the config, but it actually looks like this import was unused anyway.

## Context

I'm trying to build valhalla for use as a library (really I'm only looking at using `odin` at the moment), so I don't need the service integration. Further, prime_server is failing to build for me, but that's for another issue.

## Tasklist

I don't think any of these are relevant to this PR, but LMK:

 - [ ] Add tests

### DIFF
--- a/test/gurka/test_elevation.cc
+++ b/test/gurka/test_elevation.cc
@@ -8,12 +8,9 @@
 #include <filesystem>
 
 #include <gtest/gtest.h>
-#include <prime_server/http_protocol.hpp>
-#include <prime_server/prime_server.hpp>
 
 using namespace valhalla;
 using namespace valhalla::gurka;
-using namespace prime_server;
 
 const std::string workdir = "test/data/gurka_elevation";
 


nilsnolde (approved):
cool thanks. I wonder why circleci doesn't work for your PRs specifically, the other one too.. other external contributors don't have that problem I think (not many who are not part of the org somehow). anyways, OSX builds and runs the tests, that's good enough for this PR.

michaelkirk:
> I wonder why circleci doesn't work for your PRs specifically,

I might be wrong, but I think I ran into something similar in another repository:
https://github.com/flatgeobuf/flatgeobuf/issues/280

tldr: I don't have a circle CI account. Last time I checked, using circle CI seems to require me to give circle ci read/write access to every one of my repositories, which is a pretty tough sell for me.

Issue #4465, opened 2023-12-19
"has_*" not set when narrative = false


chrstnbwnkl:
When calling `/route` and passing the option `"narrative": false`, no `has_*` attributes seem to be set to  `true`, neither in the leg nor in the overall trip summary.

Request to reproduce:

```
curl --request POST \
  --url http://valhalla1.openstreetmap.de/route \
  --header 'Content-Type: application/json' \
  --data '{
	"locations": [

		{
			"lon": 11.503716,
			"lat": 54.774158
		},
		{
			"lon": 11.14769,
			"lat": 54.45068
		}
	],
	"costing": "auto",
"narrative": false
}'
```

This returns a route that makes use of a ferry as well as a toll road and a highway, but `has_ferry`, `has_toll` and `has_highway` set to false. When omitting the `narrative` option, these are correctly set to true.

chrstnbwnkl:
So turns out the `narrative` option has been deprecated a while ago in #1602 in favor of `directions_type`. However, with `directions_type` set to `none` I still observe the same behavior.

nilsnolde:
Whaaat? I never knew haha interesting.. So the problem is that the `has_*` stuff is set somewhere in odin which we bypass completely with `none`?

michaelhrabanek:
I'm currently working on this issue. I plan to submit a pull request soon. If anyone has already started working on this or has any insights to share, please let me know

chrstnbwnkl:
I remember quickly looking into the cause. IIRC, in `DirectionsBuilder::Build()`, the has_* attributes that we're talking about here are propagated upwards from the maneuvers. Since the maneuvers are not built with `directions_type` set to `none`, the logic inside `PopulateDirectionsLeg` that would propagate these attributes to the leg is never executed. Hope this helps a little!

PR #4706, opened 2024-04-23
Correctly set has_highway, has_ferry and has_toll attributes when directions_type is set to none

michaelhrabanek:
Fixes https://github.com/valhalla/valhalla/issues/4465

### DIFF
--- a/proto/trip.proto
+++ b/proto/trip.proto
@@ -272,6 +272,11 @@ message TripLeg {
       uint32 end_shape_index = 2;
     }
   };
+  message Summary {
+    bool has_toll = 1;
+    bool has_highway = 2;
+    bool has_ferry = 3;
+  }
 
   uint64 osm_changeset = 1;
   uint64 trip_id = 2;
@@ -286,6 +291,7 @@ message TripLeg {
   repeated Incident incidents = 11;
   repeated string algorithms = 12;
   repeated Closure closures = 13;
+  Summary summary = 14;
 }
 
 message TripRoute {
--- a/test/gurka/test_route_summary.cc
+++ b/test/gurka/test_route_summary.cc
@@ -205,3 +205,48 @@ TEST(TestRouteSummary, DupSummaryFix) {
 
   std::filesystem::remove_all(workdir);
 }
+
+TEST(Standalone, TripLegSummary) {
+  const std::string ascii_map = R"(
+      A---B---C---D
+    )";
+
+  const gurka::ways ways = {{"AB", {{"highway", "motorway"}}},
+                            {"BC", {{"highway", "motorway"}, {"toll", "yes"}}},
+                            {"CD", {{"route", "ferry"}}}};
+  const double gridsize = 100;
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize);
+  const std::string workdir = "test/data/gurka_test_route_summary";
+
+  valhalla::gurka::map map = gurka::buildtiles(layout, ways, {}, {}, workdir);
+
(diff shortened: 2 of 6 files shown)


proto/trip.proto
```diff
@@ -272,6 +272,11 @@ message TripLeg {
       uint32 end_shape_index = 2;
     }
   };
+  message Summary {
```
nilsnolde, line 275: I don't know what else to call it, though `Summary` doesn't make much sense without `time`, `length` etc from the equivalent one in directions.proto. or maybe we should put the one from directions.proto into `common.proto` and import it in both places. but then for trip legs we only use this sub-set of attributes to check whether a leg encounters toll etc?

nilsnolde: I think I'd prefer that. WDYT @kevinkreiser @dnesbitt61 ?

nilsnolde (dismissed):
thanks @michaelhrabanek LGTM mostly! except an open question on the proto message. all feedback appreciated:)

nilsnolde:
oh yeah, and you'll have to run ./scripts/format.sh once. careful, it'll install python packages (black, flake8) into the current environment, even if it's system python.

proto/trip.proto
kevinkreiser: using the same name with a different meaning is really not great. why can't the be top level in the leg? also isn't this info technically already available in the edges of the leg?

kevinkreiser: oh isee directions summary does have this info

nilsnolde: Yeah and that gets it from the maneuvers, which is why we have that bug. We needed a way to bring the info from thor to odin so I suggested this way.

should we just pull out the „Summary“ message to common.proto and let both TripLeg and DirectionsLeg share it?

kevinkreiser: yep sounds good lets share it from common.proto

test/gurka/test_route_summary.cc
```diff
@@ -205,3 +205,53 @@ TEST(TestRouteSummary, DupSummaryFix) {
...
+                            {"CD", {{"route", "ferry"}}}};
+  const double gridsize = 100;
+  const auto layout = gurka::detail::map_to_coordinates(ascii_map, gridsize);
+  const std::string workdir = "test/data/gurka_test_route_summary";
+
+  valhalla::gurka::map map = gurka::buildtiles(layout, ways, {}, {}, workdir);
+
+  valhalla::Api result0 = gurka::do_action(valhalla::Options::route, map, {"A", "B"}, "auto",
+                                           {{"/directions_type", "none"}});
+  EXPECT_TRUE(result0.trip().routes(0).legs(0).summary().has_highway());
```
nilsnolde, line 225: Can you also add a few EXPECT to see that the response JSON has those attributes as well? Currently only PBF response is tested. You can look at the signature of do_action, one overload accepts somewhere a string ref where it’ll put the JSON response in. Lots of other gurka tests are doing that too.

michaelhrabanek: Sure, something like this is ok? https://github.com/valhalla/valhalla/pull/4706/commits/7965887a0f3eda82a8c77b6057d68c269df4c974

proto/trip.proto
michaelhrabanek: Ok, thanks for suggestions, done

test/gurka/test_route_summary.cc
nilsnolde: yeah perfect:) thanks!

PR #4716, opened 2024-05-02
auto-update the fossgis instances

nilsnolde:
adds a github action workflow to update the fossgis instances automatically whenever smth interesting was pushed to master. things can still go wrong, but really shouldn't very often. I didn't touch the servers in quite a while, other than installing gdal a while back.

just a general note how I set it up back then:
- one server has two sets of valhalla servers "installed", controlled by supervisor, where only one service is online at all times. once a new graph is pushed we start/stop the supervisor instances so we don't have any outages
- builder server has a supervisord process running on `while true` (until eternity) building the planet graph, updating the planet PBF right after incrementally and pushing the new tarred graph to the services server via ssh and starting/stopping its supervisor processes remotely

### DIFF
--- a/.github/workflows/update_public_servers.yml
+++ b/.github/workflows/update_public_servers.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - nn-update-fossgis-gha
     paths:
+      - .github/workflows
       - 'src'
       - 'valhalla'
       - 'proto'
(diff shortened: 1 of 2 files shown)


nilsnolde:
hmpf, seems like I'll need to merge first. shouldn't it work for some reason we can revert

PR #4717, opened 2024-05-02
auto-update the fossgis instances

nilsnolde:
follow-up to https://github.com/valhalla/valhalla/pull/4716. will now try to run the action itself.

### DIFF
--- a/scripts/update_public_server.sh
+++ b/scripts/update_public_server.sh
@@ -1,4 +1,5 @@
 # updates Valhalla on the FOSSGIS servers
+# accepts one argument, either "builder" or "service"
 
 set -e
 
--- a/scripts/update_public_server.sh
+++ b/scripts/update_public_server.sh
@@ -9,6 +9,10 @@ git config --global --add safe.directory /src/valhalla
 
 git -C "${src_dir}" checkout master
 git -C "${src_dir}" pull
+git submodule update --init --recursive
+
+# remove the build folder first
+rm -r "${src_dir}"/build
 
 if [[ $server == "builder" ]]; then
     cmake -S "${src_dir}" -B "${src_dir}/build" \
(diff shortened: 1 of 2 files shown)


kevinkreiser (approved):
feel free to merge this if it helps testing!

nilsnolde:
I could manually trigger the workflow from this branch, but thanks. seems to work now. what it does on pushes to master (and only on changes to interesting directories):
- ssh's into each machine with a repo-wide github secret for the private key
- pulls, builds and installs the project on each machine
- on the service machine it re-generates the valhalla config, on the builder machine the forever-looping script is doing that itself

do note, this won't have an immediate effect on either the service or the builder. we don't restart the processes, we wait for the next planet graph build to finish which will automatically restart all processes on both machines so the changes will be picked up then.

from my side, this is good for merging @kevinkreiser, but feel free to look over the 2 whole files (which are also this PR's changes).

nilsnolde:
the script runs as `sudo` because of the installation. I know I could make `/usr/local` allowed for the user, but doesn't really matter IMO, the workflow can't run from a fork in a PR, only maintainers have access to the secrets. but I'm 99% sure that master merges will run regardless if from a fork or not.

scripts/update_public_server.sh
```diff
@@ -5,8 +5,14 @@ set -e
...
 src_dir="/src/valhalla"
 
+git config --global --add safe.directory /src/valhalla
+
 git -C "${src_dir}" checkout master
 git -C "${src_dir}" pull
+git submodule update --init --recursive
+
+# remove the build folder first
+rm -r "${src_dir}"/build
```
nilsnolde, line 15: ccache is installed

kevinkreiser (approved):

PR #4719, opened 2024-05-03
always save vcpkg cache

nilsnolde:
let's try to always overwrite the vcpkg binary cache.. vcpkg is very annoying to configure for caching it seems, it (very) regularly invalidates a cache when smth in the build toolchain updates and then it decides to rebuild all packages which takes 45 mins.. I quickly looked for a solution but all I could find were very very long Github issues.. I _think_ most of the invalidations are caused by vcpkg's own copy of cmake which updates quite often..


nilsnolde:
I ran the job again and it successfully cached vcpkg packages again. so this approves itself:)

Issue #4601, opened 2024-02-23
some (backwards-compatible) improvements to `/expansion`


nilsnolde:
2 things I'd like to add to make this more useful for our debugging:
- `is_shortcut` to know whether a traversed edge was a shortcut
- choose which stage of the expansion is being logged: sometimes, if I want to compare 2 expansions in terms of total edges explored I'd like to know the total amount of `reached` or `settled` or `rejected`  (s. below) edges. today I have to comment the relevant portions in the code. would cut down the total response size tremendously as well and usually one is only interested in one of the 2-3 stages
- (optional since perf affecting): add one more stage of logging in `ExpandInner` where we log edges which didn't make the cut because of access restrictions or whatever, i.e. not `reached` but `rejected` or so

also again, smth is messed up with the "expansion history", i.e. some of the algos log the wrong predecessor or current edge, it often makes no sense.

nilsnolde:
after that, I'd personally be happy to remove the BETA label finally

kevinkreiser:
sorry for being quiet so long. the whole point of the expansion action **is** debugging. i know that over the years others have found other interesting use cases for it but we need to maintain this key functionality. just to be clear what i mean by that ill link the original work here: 

https://github.com/valhalla/valhalla/pull/1849 
https://github.com/valhalla/demos/pull/233

at any rate i think we need to maintain the ability to do this kind of analysis with this endpoint. im not opposed to enhancing it so that it can serve more use-cases that arent debugging but we cant lose the original purpose.

PR #4728, opened 2024-05-12
Improvements to /expansion

michaelhrabanek:
This PR addresses the issue described in https://github.com/valhalla/valhalla/issues/4601, implementing the solution proposed by @nilsnolde. Currently, the expansion process is overly verbose because it logs edges multiple times—for each stage of expansion and when an edge is reached from another branch. This PR keeps track of every affected edge during expansion only once, with all properties associated with it (necessary for the final response assembly). Properties such as cost and duration for already tracked edges are updated only at stages that are higher or same (to keep track of latest ones). This also means we keeping only latest prev_edge_it so we lose information about whether an edge was reached from other branches. I'm not sure how significant this issue might be. Thoughts?

This is still a draft, and gurka tests are missing, as I would appreciate some feedback before proceeding further.

## Tasklist

 - [x] Add tests


kevinkreiser:
i left a comment on the issue: https://github.com/valhalla/valhalla/issues/4601#issuecomment-2106398348

kevinkreiser:
just to be super clear i dont have any problem with us updating this. but id rather it be something you opt into with a flag. maybe `dedupe=true` or something. and when that flag is on the callback will check if the edge is already in the output and if so only update if the info should override the edge properties (lower cost, final state etc). otherwise (dedupe=false, the default) the code should always add a new feature as it does today so that we can still make sense of the timeseries information.

michaelhrabanek:
Ok, I've added dedupe option, which defaults to false and keeps the original behaviour.

docs/docs/api/expansion/api-reference.md
```diff
@@ -12,10 +12,11 @@ The expansion service wraps the `route`, `isochrone` and `sources_to_targets` se
 
 Since this service wraps other services, the request format mostly follows the ones of the [route](../turn-by-turn/api-reference.md#inputs-of-a-route), [isochrone](../isochrone/api-reference.md#inputs-of-the-isochrone-service) and [matrix](../matrix/api-reference.md#inputs-of-the-matrix-service). Additionally, it accepts the following parameters:
 
-| Parameter                         | Description                           |
+| Parameter                         | Description                                                                                                                                                                                                                                                                                                                          |
```
kevinkreiser, line 15: nit: probably dont need the whitespace

docs/docs/api/expansion/api-reference.md
```diff
@@ -12,10 +12,11 @@ The expansion service wraps the `route`, `isochrone` and `sources_to_targets` se
...
+| `action` (required)               | The service whose expansion should be tracked. Currently one of `route`, `isochrone` or `sources_to_targets`.                                                                                                                                                                                                                        | 
+| `skip_opposites` (optional)       | If set to `true` the output won't contain an edge's opposing edge. Opposing edges can be thought of as both directions of one road segment. Of the two, we discard the directional edge with higher cost and keep the one with less cost. Default false.                                                                             | 
+| `dedupe` (optional)               | If set to `true`, the output will contain each edge only once, significantly reducing the response size. The expansion will keep track of already visited edges and override their properties, ensuring that only the newer one or the one with higher edge states is returned. Default false.                                       |
```
kevinkreiser, line 19: thank you very much for documenting!

proto/expansion.proto
```diff
@@ -18,9 +18,10 @@ message Expansion {
   repeated uint32 costs = 1 [packed=true];
   repeated uint32 durations = 2 [packed=true];
   repeated uint32 distances = 3 [packed=true];
-  repeated EdgeStatus edge_status = 4;
+  repeated EdgeStatus edge_status = 4 [packed=true];
```
kevinkreiser, line 21: this is technically a breaking change but its a beta service so oh well i guess

proto/expansion.proto
```diff
@@ -18,9 +18,10 @@ message Expansion {
...
   repeated uint32 edge_id = 5 [packed=true];
   repeated uint32 pred_edge_id = 6 [packed=true];
+  repeated uint32 edge_status_flags = 7 [packed=true];
 
-  repeated Geometry geometries = 7;
+  repeated Geometry geometries = 8;
```
kevinkreiser, line 26: moving things around in proto is very much a breaking change again it doesnt matter because we call this service beta but we should try not to do this when we dont need to

docs/docs/api/expansion/api-reference.md
michaelhrabanek: I've changed it in skip_opposites too, for consistency

proto/expansion.proto
michaelhrabanek: You're right of course, but since we're breaking it anyway, I've moved it around ... well just for aesthetic reasons tbh.

docs/docs/api/expansion/api-reference.md
michaelhrabanek: Oh, sorry, I've overlooked this, it's fixed now

src/thor/expansion_action.cc
```diff
@@ -78,43 +118,56 @@ std::string thor_worker_t::expansion(Api& request) {
...
+        auto tile = reader.GetGraphTile(edgeid);
+        if (tile == nullptr) {
+          LOG_ERROR("thor_worker_t::expansion error, tile no longer available" +
+                    std::to_string(edgeid.Tile_Base()));
+          return;
+        }
+        const auto* edge = tile->directededge(edgeid);
+        // unfortunately we have to call this before checking if we can skip
+        // else the tile could change underneath us when we get the opposing
+        auto shape = tile->edgeinfo(edge).shape();
```
kevinkreiser, line 138: you're saying that because of the call to `GetOpposingEdgeId` below? if so the simple solution would be to, inside of the `skip_opps` `if` block to add `auto opp_tile = tile;` and use `opp_tile` in the call to `GetOpposingEdgeId`. if i misunderstood this comment let me know!

michaelhrabanek: This is original comment from @nilsnolde actually, I haven't touch it, it just got marked by diff as I've clean up some unused code around those lines

kevinkreiser: i mean would be good to move the grabbing of the shape below the `if` though in case that method bails before using it

michaelhrabanek: Good point, I've moved it and removed the original comment

src/thor/expansion_action.cc
```diff
@@ -46,12 +46,49 @@ void writeExpansionProgress(Expansion* expansion,
...
+    uint8_t first_significant = current | candidate;
+    // connected(LSb) is the highest status, so check from right
+    for (uint8_t i = 0; i < Expansion_EdgeStatus_EdgeStatus_ARRAYSIZE; i++) {
+      uint8_t mask = 1 << i;
+      if (!(first_significant & mask))
+        continue;
+      return candidate & mask;
+    }
+    return false;
+  }
```
kevinkreiser, line 90: its not obvious to me why we need the complication of remembering all the statuses an edge has seen using a mask. it is true that in bi/multidirectional search an edge can be reached and settled multiple times (once for each search direction) but we dont need to know the prior statuses we only need to know the most important (lowest value) one its ever seen. tell me what im missing here. cant this just be:

```suggestion
  // check if status is higher or same – as we will keep track of the latest one
  static bool is_latest_status(uint8_t current, uint8_t candidate) {
    return candidate < current;
  }
```

src/thor/expansion_action.cc
```diff
@@ -78,43 +118,56 @@ std::string thor_worker_t::expansion(Api& request) {
...
+            auto edge_stages = edge_state.at(edgeid).stages_mask;
+            // Keep only properties of last/highest status of edge, but update what stages the edge
+            // has seen
+            if (!expansion_properties_t::is_latest_status(edge_stages, stage)) {
+              edge_state.at(edgeid).stages_mask |= stage;
+              return;
+            }
+            stage |= edge_state.at(edgeid).stages_mask;
+          }
+          edge_state[edgeid] =
```
kevinkreiser, line 164: i very much appreciate that we dont fill up this large map when not deduping, thank you for that!

src/thor/expansion_action.cc
```diff
@@ -78,43 +118,56 @@ std::string thor_worker_t::expansion(Api& request) {
...
+                const Expansion_EdgeStatus status = Expansion_EdgeStatus_reached,
+                const float duration = 0.f, const uint32_t distance = 0, const float cost = 0.f) {
+        algo = algorithm;
+
+        auto tile = reader.GetGraphTile(edgeid);
+        if (tile == nullptr) {
+          LOG_ERROR("thor_worker_t::expansion error, tile no longer available" +
+                    std::to_string(edgeid.Tile_Base()));
+          return;
+        }
```
kevinkreiser, line 134: i know this was here before but it seems crazy pedantic. the graph reader can fail to get a tile for sure but not one its already seen (ie we know is in the tileset). the worst case is that its a cache miss and it has to pull it from disk or from network but if the algorithm is passing you this graph id for an edge its seen it will have to be able to get the edge. imho we can delete this check, if this happens someone is deleting tiles from the data storage while the application is running which is not worth protecting against

src/thor/expansion_action.cc
```diff
@@ -78,43 +118,56 @@ std::string thor_worker_t::expansion(Api& request) {
...
+        Polyline2<PointLL>::Generalize(shape, gen_factor, {}, false);
+        if (dedupe) {
+          uint8_t stage = 1 << status;
+          if (edge_state.contains(edgeid)) {
+            auto edge_stages = edge_state.at(edgeid).stages_mask;
+            // Keep only properties of last/highest status of edge, but update what stages the edge
+            // has seen
+            if (!expansion_properties_t::is_latest_status(edge_stages, stage)) {
+              edge_state.at(edgeid).stages_mask |= stage;
+              return;
```
kevinkreiser, line 160: can we refactor this section just a little bit to add an early exit before the shape stuff? something like this (pseudo code, i suggest renaming the `stage` member to `status` if the thing i said about the mask makes sense):

```c++
// if we've seen it before we can update if better and bail either way
auto maybe_state = edge_state.find(edgeid); // works for dedupe == false, since its empty
if (maybe_state != edge_state.cend()) {
  if (maybe_state->status > status)
    edge_state[edgeid].status = status;
  return;
}

// get the shape ready because we're going to use it
const auto* edge = tile->directededge(edgeid);
auto shape = tile->edgeinfo(edge).shape();
if (!edge->forward())
Polyline2<PointLL>::Generalize(shape, gen_factor, {}, false);

// either we track it for serialization later or
if (dedupe) {
  edge_state[edgeid] = expansion_properties_t(prev_edgeid, status, duration, distance, shape, cost);
} // we simply serialize it now
else {
  writeExpansionProgress(expansion, edgeid, prev_edgeid, shape, exp_props, status, status, duration, distance, cost, false);
}
```

src/tyr/expansion_serializer.cc
```diff
@@ -62,8 +62,17 @@ std::string serializeExpansion(Api& request, const std::string& algo) {
...
     }
-    if (exp_props.count(Options_ExpansionProperties_edge_status))
-      writer("edge_status", Expansion_EdgeStatus_Enum_Name(expansion.edge_status(i)));
+    if (exp_props.count(Options_ExpansionProperties_edge_status)) {
+      if (request.options().dedupe()) {
+        static const std::vector<std::string> status_captions{{"status_connected"},
+                                                              {"status_settled"},
+                                                              {"status_reached"}};
+        for (uint8_t j = 0; j < status_captions.size(); j++)
+          writer(status_captions[j], (bool)(expansion.edge_status_flags(i) & 1 << j));
```
kevinkreiser, line 71: yeah i think this is unneccesary the best status that is reached implies any others below it were reached

kevinkreiser (dismissed):
i had a few suggested changes but for the most part it looks good!

src/thor/expansion_action.cc
michaelhrabanek: Hm, yes, that would simplify a lot of things. I implemented it this way because it was the proposed solution in the original issue (https://github.com/valhalla/valhalla/issues/4601). To be honest, I assumed it had some use case I wasn't aware of but would be useful anyway. @nilsnolde, could you please shed some insight on this?

proto/expansion.proto
nilsnolde: maybe it's time after this PR to take it out of BETA state?

src/thor/expansion_action.cc
nilsnolde: yeah, @kevinkreiser is of course right. some things only come to you when it's in code I guess :) sorry for the extra hassle I caused!

src/thor/expansion_action.cc
nilsnolde: fair

michaelhrabanek:
Ok, so I got rid of all statuses the edge has seen. Now, it only contains the latest highest one. It has simplified of lot things actually.

src/thor/expansion_action.cc
```diff
@@ -52,6 +51,31 @@ void writeExpansionProgress(Expansion* expansion,
...
+  std::vector<midgard::PointLL> shape;
+  uint32_t distance;
+  float cost;
+
+  expansion_properties_t() = default;
+  expansion_properties_t(baldr::GraphId prev_edgeid,
+                         Expansion_EdgeStatus status,
+                         float duration,
+                         uint32_t distance,
+                         std::vector<midgard::PointLL> shape,
```
kevinkreiser, line 69:
```suggestion
                         std::vector<midgard::PointLL>&& shape,
```

src/thor/expansion_action.cc
```diff
@@ -78,43 +105,51 @@ std::string thor_worker_t::expansion(Api& request) {
...
-    if (!edge->forward())
-      std::reverse(shape.begin(), shape.end());
-    Polyline2<PointLL>::Generalize(shape, gen_factor, {}, false);
-
-    writeExpansionProgress(expansion, edgeid, prev_edgeid, shape, exp_props, status, duration,
-                           distance, cost);
-  };
+  auto track_expansion =
+      [&expansion, &opp_edges, &gen_factor, &skip_opps, &exp_props, &algo, &edge_state,
+       &dedupe](baldr::GraphReader& reader, baldr::GraphId edgeid, baldr::GraphId prev_edgeid,
```
kevinkreiser, line 110: at this point why not just capture them all instead of listing them out?
```suggestion
      [&](baldr::GraphReader& reader, baldr::GraphId edgeid, baldr::GraphId prev_edgeid,
```

src/thor/expansion_action.cc
```diff
@@ -24,7 +22,8 @@ void writeExpansionProgress(Expansion* expansion,
                             const Expansion_EdgeStatus& status,
                             const float& duration,
                             const uint32_t& distance,
-                            const float& cost) {
+                            const float& cost,
+                            const bool dedupe) {
```
kevinkreiser, line 26: tidy
```suggestion
                            const float& cost) {
```

src/thor/expansion_action.cc
```diff
@@ -146,6 +181,15 @@ std::string thor_worker_t::expansion(Api& request) {
     // anyway
   }
 
+  // assemble the properties from latest/highest stages it went through
+  if (dedupe) {
+    for (const auto& e : edge_state) {
+      writeExpansionProgress(expansion, e.first, e.second.prev_edgeid, e.second.shape, exp_props,
+                             e.second.status, e.second.duration, e.second.distance, e.second.cost,
+                             true);
```
kevinkreiser, line 189: tidy
```suggestion
                             e.second.status, e.second.duration, e.second.distance, e.second.cost);
```

kevinkreiser (commented):
ive noted just a little bit of cleanup but this is basically done. thanks for slogging through

michaelhrabanek:
> ive noted just a little bit of cleanup but this is basically done. thanks for slogging through

Thank you Kevin, I've incorporated your changes

chrstnbwnkl (approved):

Issue #1069, opened 2018-01-17
Add country specific speeds


dnesbitt61:
Add logic to set speeds based on country. See: https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Maxspeed#Default_speed_limits

This also refers to places= tags - so issue #236 is related.

https://wiki.openstreetmap.org/wiki/Places

westnordost:
I have been researching into this and was about to open a ticket here, asking how Valhalla infers implicit speed limits (= when no sign is posted).

In StreetComplete, I am [working on sourcing the information](https://github.com/westnordost/StreetComplete/issues/813) that is missing to infer these speed l limits now also for the United States, as the jurisdiction differs for each state. The tags set will then probably look something like i.e.
`maxspeed:type=US-AK:residential` (= road uses the implicit speed limit for residential districts in Alaska)

The linked wiki page is far from complete. It would be nice, if there was an implementation for that data. This doesn't need to be part of Valhalla but could also be another library that is then simply used by Valhalla. 
90% of this library would be researching, determining and maintenance of these implicit speed limits - that is, rule sets written in per-country configuration files.
If anyone is poised to do this - casting the legislations that determine implicit speed limits per state into machine-readable configuration files, I might give some input there as I both already researched into the topic and have some ideas how these files could be structured.

1ec5:
Note that, although @westnordost’s [distillation of default limits](https://wiki.openstreetmap.org/wiki/Default_speed_limits) is much simplified from the source legislation, there are a lot of distinctions that require context beyond just the containing national and regional administrative boundaries. For example, a common feature among U.S. state rules is that “residential” areas are treated differently than “business” districts – a zoning rule that might even be out of scope for OSM. (Other rules are a little more tractable, taking into consideration the highway class, route network, or number of lanes.)

So while I think the basic ability to vary the default speed limit by national or regional boundary is useful, we should keep in mind the limits of this approach for automatically, authoritatively backfilling for missing speed limits or missing speed limit signs.

ianthetechie:
Agreed. At the moment, I see this not as a way to backfill data into OSM directly, but to provide more sane defaults for the purposes of estimating route duration.

westnordost:
For cases like residence and business zones, indeed, it is an open question how to record these. Same with the urban /rural distinction in most places outside the US as well.

We can distinguish two cases:

[...]

PR #4739, opened 2024-05-22
Simple legal default speeds

### LINKED ISSUE #3334: Infer default speed limits
This issue ticket's purpose is to
 
1. get feedback from you guys if inferring default speed limits is desired at all
2. get feedback from you guys if the default speed limit data as provided is in a format you are able to make use of etc.
3. finally, serve as a starting-area / feature request to implement this

## Assumptions

The maximum allowed speed on any given road is an important factor to rate the costing of edges in the graph and travel time.

`maxspeed` coverage in OSM is sketchy and likely will always be: Currently, **only about 12%** of all roads have a speed limit tagged.

Additionally, `maxspeed` is just for cars. Other vehicles such as trucks, buses, motorhomes, cars with trailers or motorcycles will, depending on the road type, have different maximum speed limits that are often below the signed one.

## Implicit speed limits

When speed limits are not signed, the default speed limits apply. _(Maybe more so outside the United States. I heard that in the US, things are usually always explicitly signed.)_ Regardless of the reason why `maxspeed` is missing - nothing is signed or information is still missing - the actual speed limits can be inferred by looking at the legislation. 

**For example**, in Hungary, the default speed limits are as follows
- 20 km/h on a living street
- 30 km/h on a bicycle road
- 50 km/h on any other urban road
- 90 km/h on a non-urban road, except for vehicles pulling a trailer, with a [GVWR](https://en.wikipedia.org/wiki/Vehicle_weight#Gross_vehicle_weight_rating)of 3.5t, buses or trucks for which it is 70 km/h
- 110 km/h on motorroads, except for the vehicle types mentioned above for which it is still 70 km/h
- 130 km/h on a motorway. For buses however, it's 100 km/h. For vehicles, pulling a trailer, with a GVWR of 3.5t or more and trucks, it is 80 km/h

So, if a road in Hungary is tagged with e.g. `highway=living_street` or `motorroad=yes`, we can easily infer the implicit speed limit from the above information. Hungary however is a rather easy example.

### Sourcing implicit speed limits

The legislations as the one above have been collected over the years on the OSM wiki page [Default speed limits](https://wiki.openstreetmap.org/wiki/Default_speed_limits) in a well-researched (with links to the actual legislation) and machine-readable form. @ianthetechie and me have been working on [a Python parser for this table](https://github.com/westnordost/osm-default-speeds) that outputs the information in it as a JSON-file so that it can easily be consumed by (routing) software.

This is how it currently looks like: [default_speeds.json](https://github.com/westnordost/osm-default-speeds/blob/master/default_speeds.json)

<details>
<summary>See how it looks for Hungary</summary>

```json
"HU": [
    {
      "filter": "highway=living_street or living_street=yes",
      "name": "living street",
      "tags": {
        "maxspeed": "20"
      }
    },
    {
      "filter": "bicycle_road=yes or cyclestreet=yes",
      "name": "bicycle road",
      "tags": {
        "maxspeed": "30"
      }
    },
    {
      "filter": "{is_urban}",
      "name": "urban",
      "tags": {
        "maxspeed": "50"
      }
    },
    {
      "tags": {
        "maxspeed": "90",
        "maxspeed:bus": "70",
        "maxspeed:hgv": "70",
        "maxspeed:conditional": "70 @ (weightrating>3.5); 70 @ (trailer)"
      }
    },
    {
      "filter": "motorroad=yes",
      "name": "motorroad",
      "tags": {
        "maxspeed": "110",
        "maxspeed:bus": "70",
        "maxspeed:hgv": "70",
        "maxspeed:conditional": "70 @ (weightrating>3.5); 70 @ (trailer)"
      }
    },
    {
      "filter": "highway~^motorway(_link)?$",
      "name": "motorway",
      "tags": {
        "maxspeed": "130",
        "maxspeed:bus": "100",
        "maxspeed:hgv": "80",
        "maxspeed:conditional": "80 @ (weightrating>3.5); 80 @ (trailer)"
      }
    }
  ]
```
</details>

### Inherent complexity

Already in the example for Hungary, the following becomes clear:

#### Filter-based mapping

There can not be a 1:1 mapping of `highway=X` to speed limit rules because 
- there is no (single) tag for "urban road". It may be sourced from other information such as `source:maxspeed=HU:urban` and similar tags, by looking at the location of `traffic_sign=city_limit` signs, whether or not a road is in- or outside of a respective city limit boundary or fuzzily by looking at tags such as `lit=yes/no` and similar (because roads outside of settlements are usually not lit)
- sometimes different tags are synonymous in OSM
- sometimes the legal road type matches multiple different tags on OSM
- sometimes the legal road type depends on such things as whether a road is lit or not (e.g. United Kingdom), whether the oncoming traffic is segregated (e.g. Germany) or whether the road is paved at all (e.g. Georgia, US)

Instead, it must be based on filtering by OSM tags. For example, the default speed limit in Illinois for roads where the oncoming traffic is segregated and there are 2 or more lanes in each direction would be 65 mph. For interstates, it would be 70 mph.
The filters (in pseudo-Overpass-Wizard, `~` meaning it would be a regex-match) for that would be:
- `oneway ~ ^yes|-1$ and lanes >= 2 and dual_carriageway != no` -> 65 mph
- `ref ~ ^I and highway ~ ^motorway|motorway_link$`

To have a parser for such pseudo-Overpass-Wizard syntax adds quite the complexity, I believe.

#### Reverse country/state geocoder 

Not every motorway with a `ref` code starting with `I` is an US interstate highway. This rule may only be applied in the US. Thus, the router must know in which country/state any given road is located. (Should be no problem though if you use a database with geospatial extensions. I can give pointers.)

#### Advanced maxspeed and maxspeed:conditional parsing

The information in the default speed limits table is translated into OSM tags, including `maxspeed:conditional` and `maxspeed:<vehicle type>`. Thus, to interpret the limits correctly, it's necessary for the (router) software to also parse the information in these tags correctly.
Whether this adds complexity or not depends on if Valhalla can already parse these, as these are normal OSM maxspeed tags.

## Your Feedback necessary

So, the [mentioned parser](https://github.com/westnordost/osm-default-speeds) hasn't been worked on for quite some time because we need feedback from you. No point in creating something if it isn't used or doesn't even have the chance of being used. See the points at the very beginning of this ticket.

I may have some time in the near future to improve on this and also maintain the source data, so:

Do you think this data is useful? Can you work with this data? What is necessary to work with it? Do you have suggestions on how else this data could be represented? Do you have other questions?

chrstnbwnkl:
This PR adds support for simple legal default speeds. 

Fixes #3334 and #1069.

Following the discussions in #1069 and #3334 this PR attempts to facilitate assignment of country specific legal speeds based on simple rules. The idea is to allow the user to provide a JSON file that follows the schema proposed in [this repo](https://github.com/westnordost/osm-legal-default-speeds) (see example JSON [here](https://raw.githubusercontent.com/westnordost/osm-legal-default-speeds/master/demo/distribution/legal_default_speeds.json)). If present, valhalla will try to set edge speeds based on the edges' admin info where no tagged speeds were present.

Since in most countries there are rules simply on the notion of whether a road segment is in an urban or rural area, this assignment of default speeds happens in the enhance stage of the graph build, where the density estimate is available. 

The proposed logic further attempts to set speeds according to rules specific to a certain road class or road use. While the aforementioned JSON schema allows for much more complicated rules, this logic only implements a few of these for now. I think it is possible to extend this in the future by adding another legal speed assignment stage in earlier stages of the graph build where more information from the original way tags are available. Though it won't be possible to implement all of these rules, since some are dependent on vehicle information only available in costing.

WRT the customizable speed config added in #3055, that config, if present, will still override any speeds set by this logic. 

## Tasklist

 - [x] Add tests


src/mjolnir/graphenhancer.cc
```diff
@@ -1584,9 +1590,19 @@ void enhance(const boost::property_tree::ptree& pt,
...
+        if (found_speed && directededge.speed_type() == SpeedType::kClassified) {
+          // EdgeInfo edgeinfo = tilebuilder->edgeinfo(&directededge);
+          tilebuilder->set_speed_limit(directededge.edgeinfo_offset(), found_speed);
+        }
+
         // Speed assignment
         speed_assigner.UpdateSpeed(directededge, density, infer_turn_channels, end_node_code,
-                                   end_node_state_code);
+                                   end_node_state_code,
+                                   found_speed == 0 || !update_speed_from_legal_speeds);
```
chrstnbwnkl, line 1605: test for speed limit if speed assigner overrides legal default speeds

src/mjolnir/legal_speed.h
```diff
@@ -0,0 +1,319 @@
...
+
+  // if the value is not numeric, check if it's "walk" else default to 0
+  uint32_t speed = num_ix < 1 ? s == "walk" ? 10 : 0 : stoi(s.substr(0, num_ix));
+
+  if (s.size() > 3 && s.substr(s.size() - 3) == "mph") {
+    speed = static_cast<uint32_t>(std::round(speed * kMPHtoKPH));
+  }
+
+  return speed;
+};
```
chrstnbwnkl, line 71: extra semicolon

chrstnbwnkl:
I ran some route tests on a Switzerland extract and compared routes on a graph with legal speeds set as edge speeds vs a graph with no extra speed assignments, and the results are not very convincing (results from the default speed graph in light blue, default graph in purple):

image

image

Routes via rural areas tend to deviate from arterial roads onto lower road classes. Without looking at this in more detail, it seems very likely that these smaller roads are assigned high speeds because the general rural legal speed for Switzerland is 80 kph (which is higher than what Valhalla assumes for these road classes if no maxspeed tag exists). 

In urban areas, I observed almost no differences, presumably in part due to the high occurence of `maxspeed` tags. 

I think @nilsnolde was right, the purely density based assignment does not make much sense for setting edge speeds, maybe that should only be applied in cases where we have legal speeds that are based on some more specific rule (form of way, road use, etc.).

chrstnbwnkl:
Update: 

To circumvent issues with speeds that deviate heavily from Valhalla's default speeds, I decided to clamp those to the default speeds for now. Using these speeds as actual edge speeds is a highly experimental feature anyway, and the primary use case of the legal speed information really should be setting the speed limit rather than the speed.

chrstnbwnkl:
I'm going to close this. This was one of my first PRs to Valhalla, and I was young and naive :smile: in order to properly infer legal default speeds, we need to have a whole lot more information than

  - whatever we have on the edge at the enhance stage, plus
  - a very vague estimate of whether we're in a urban/rural area 
 
because I think in order for this feature be useful, the speed limit must be more than an educated guess. 

Some anecdotal evidence: I sometimes have the pleasure to drive around in a car with a modern in-car navigation system, with speed limit information that is almost always correct and up-to-date. I automatically start to rely on that information in order to not go past the speed limit (obviously I try to go by the signage but I'll sometimes forget what it said, etc). 

If I imagine that speed limit information to be nothing more than a "good guess", I would be very frustrated, because the worst outcome is that if I rely on this information, I may drive to fast and get a speeding ticket.

So in my opinion, a PR that tackles this properly must not only have access to more information from OSM (to find out whether e.g. an edge runs through a school zone), but also, and more importantly, a good source on whether a certain edge runs through built-up area or not. This information is not in OSM, so you would have to get it elsewhere. In Germany, for example, you can get it from ALKIS data (cadastral data). 

If anyone has interest in tackling this, feel free to contact me via mail.

PR #4743, opened 2024-05-28
Adds documentation for `ignore_oneways` option

eikes:
# Issue

There is a so far undocumented option called `ignore_oneways` which does just that, it allows finding routes which go against one-way restrictions. This PR adds documentation for this option.

## Tasklist


eikes:
Here is where the option is defined:
https://github.com/valhalla/valhalla/blob/c14d3ac8e4745a17e5e99a50dac32c838fc3475b/proto/options.proto#L256-L258

Here it gets parsed:
https://github.com/valhalla/valhalla/blob/c14d3ac8e4745a17e5e99a50dac32c838fc3475b/src/sif/dynamiccost.cc#L392

Here it gets passed on:
https://github.com/valhalla/valhalla/blob/c14d3ac8e4745a17e5e99a50dac32c838fc3475b/src/sif/dynamiccost.cc#L157

Here it is used:
https://github.com/valhalla/valhalla/blob/c14d3ac8e4745a17e5e99a50dac32c838fc3475b/valhalla/sif/dynamiccost.h#L321
https://github.com/valhalla/valhalla/blob/c14d3ac8e4745a17e5e99a50dac32c838fc3475b/valhalla/sif/dynamiccost.h#L343

There are some tests that verify this works correctly:

https://github.com/valhalla/valhalla/blob/c14d3ac8e4745a17e5e99a50dac32c838fc3475b/test/gurka/test_route.cc#L69-L135

PR #4766, opened 2024-06-14
Add missing include in `baldr/admin.h`

chrstnbwnkl:
# Issue

I've been starting to get compilation errors recently due the use of `std::find()` in `baldr/admin.cc`, apparently caused by a missing include:

```
[build] /home/chris/dev/github/valhalla/src/baldr/admin.cc: In member function ‘std::string valhalla::baldr::Admin::state_iso() const’:
[build] /home/chris/dev/github/valhalla/src/baldr/admin.cc:109:57: error: no matching function for call to ‘find(std::array<char, 3>::const_iterator, std::array<char, 3>::const_iterator, char)’
[build]   109 |              : std::string(state_iso_.begin(), std::find(state_iso_.begin(), state_iso_.end(), '\0'));
[build]       |                                                ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

 Not sure what's the root cause, but seems like Valhalla was just relying on it being included elsewhere? 

## Tasklist

### DIFF
--- a/valhalla/baldr/admin.h
+++ b/valhalla/baldr/admin.h
@@ -1,6 +1,7 @@
 #ifndef VALHALLA_BALDR_ADMIN_H_
 #define VALHALLA_BALDR_ADMIN_H_
 
+#include <algorithm>
 #include <array>
 #include <cstdint>
 #include <cstring>
--- a/valhalla/baldr/admin.h
+++ b/valhalla/baldr/admin.h
@@ -1,7 +1,6 @@
 #ifndef VALHALLA_BALDR_ADMIN_H_
 #define VALHALLA_BALDR_ADMIN_H_
 
-#include <algorithm>
 #include <array>
 #include <cstdint>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


kevinkreiser:
youre exactly right. what happens is compilers come with their own implementations of the standard library. what then happens is you use one of the headers, lets say `vector`, and in your compilers version of the standard library they happened to also include `algorithm` because one of the vector functions needed it or it was convenient. you throw in a call to `std::find` and it works because you were inadvertantly relying on that include you didnt know about. time passes and a new compiler release happens or you take your code to another platform with a different compiler and now your code doesnt work because vector on that platform isnt also giving access to algorithm.

what we should do but havent done is strictly follow the "include what you use" idiom. itd be a nice cleanup to run some toolign to detect this stuff but it might be a bit of a long and winding road. anyway easily fixed as you noticed.

valhalla/baldr/admin.h
```diff
@@ -1,6 +1,7 @@
 #ifndef VALHALLA_BALDR_ADMIN_H_
 #define VALHALLA_BALDR_ADMIN_H_
 
+#include <algorithm>
```
kevinkreiser, line 4: if we only need access to this in admin.cc then put the include there. better to keep our headers as slim as possible so that people using our headers dont get more stuff than they need (not that it bloats their binaries but just because we should advertize only what we need to, ie only what is used directly in this header file)

kevinkreiser (approved):

nilsnolde:
wtf haha never seen this. must be a glitch on the circleci machine. not sure how circleci works these days. back when I registered, I did so through github and it was "just working" bcs I think it checked my permissions to this repo (were the same as yours today). when you sign up with circleci, does it ask for your github access? I know they're being google'y-evil about it requesting wayyy too much info, but :shrug:

kevinkreiser:
meh who cares about stupid CI messing with us

PR #4784, opened 2024-06-27
Add `expansion_type` property to `/expansion`

chrstnbwnkl:
# Issue

Adds the expansion property `expansion_type`. 

Some considerations:
  - not sure these casts between enum types are so good, I just thought the more explicit detour via int would be a bit too verbose 
  - in theory also supports multimodal expansion type, but expansion action doesn't support it (shouldn't be too much work though? :smile: ) so I left it undocumented for now
  - I kept the json value as int intentionally, I think strings just bloat the response up way too much. I know `edge_status` does it differently, let me know if you think consistency is more important here
  - didn't add tests that actually check the property values, just made sure the property exists on the output if requested by the user, as with the existing tests


src/thor/expansion_action.cc
```diff
@@ -78,43 +82,44 @@ std::string thor_worker_t::expansion(Api& request) {
...
-    }
-
-    if (!edge->forward())
-      std::reverse(shape.begin(), shape.end());
-    Polyline2<PointLL>::Generalize(shape, gen_factor, {}, false);
-
-    writeExpansionProgress(expansion, edgeid, prev_edgeid, shape, exp_props, status, duration,
-                           distance, cost);
-  };
+  auto track_expansion =
```
chrstnbwnkl, line 85: most of this hunk is just formatting, I just pass the new `expansion_type` arg on to the writer func

proto/expansion.proto
```diff
@@ -15,12 +15,19 @@ message Expansion {
     reached = 2;
   }
 
+  enum ExpansionType {
+    forward = 0;
+    reverse = 1;
+    multimodal = 2;
```
kevinkreiser, line 21: maybe drop this for now even if it does match path algo header stuff?

chrstnbwnkl, line 21: scratch that

src/thor/expansion_action.cc
```diff
@@ -104,50 +110,51 @@ std::string thor_worker_t::expansion(Api& request) {
...
-        return;
-      opp_edges.insert(edgeid);
-    }
+  auto track_expansion =
+      [&](baldr::GraphReader& reader, baldr::GraphId edgeid, baldr::GraphId prev_edgeid,
+          const char* algorithm = nullptr,
+          const Expansion_EdgeStatus status = Expansion_EdgeStatus_reached,
+          const float duration = 0.f, const uint32_t distance = 0, const float cost = 0.f,
+          const Expansion_ExpansionType expansion_type = Expansion_ExpansionType_forward) {
+        algo = algorithm;
```
chrstnbwnkl, line 119: aw man, these diffs again... all that changed here is the extra `expansion_type` argument that's being passed around, and it has some effects on the formatting.

kevinkreiser (approved):

PR #4787, opened 2024-06-29
Add `--inline-config` arg to `valhalla_build_elevation` 

chrstnbwnkl:
# Issue

Just noticed that the elevation build script was missing the option to just pass an inline config.

### DIFF
--- a/scripts/valhalla_build_elevation
+++ b/scripts/valhalla_build_elevation
@@ -89,6 +89,13 @@ parser.add_argument(
     type=Path,
 )
 parser.add_argument(
+    "-i",
+    "--inline-config",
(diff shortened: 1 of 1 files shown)


scripts/valhalla_build_elevation
```diff
@@ -322,14 +329,17 @@ if __name__ == "__main__":
...
         with open(args.config) as f:
             config = json.load(f)
 
+    if args.inline_config:
+        config.update(**json.loads(args.inline_config))
+
     if args.outdir:
         elevation_fp = args.outdir
-    elif config is not None:
+    elif config:
```
chrstnbwnkl, line 342: empty dicts are falsy

kevinkreiser (approved):

PR #4793, opened 2024-07-03
Restructure top speeds for truck

### LINKED ISSUE #4789: Truck costing ignoring top_speed
Was looking to update from version 3.4.0 to a main branch build and noticed our truck routes were all ~20% slower. 

I see https://github.com/valhalla/valhalla/pull/4493 clamped truck speed to 90 KPH / 55 MPH. The intention there makes sense to me (avoiding unbounded truck speed), but 55 mph is a bit slow for some highways and doesn't seem like the new top_speed can be overridden with a costing_option?

> `top_speed` Top speed the vehicle can go. Also used to avoid roads with higher speeds than this value. top_speed must be between 10 and 252 KPH. The default value is 90 KPH for truck and 140 KPH for auto and bus.

For an example request, a small section of I-95 in Virginia, US: 
```
/route?json={"locations":[{"lat":37.858795,"lon":-77.453613,"type":"break"},{"lat":37.872482,"lon":-77.453914,"type":"break"}],"costing":"truck"}
```
 on our 3.4.0 build
```
"summary": {
      ...
      "time": 48.725,
      "length": 1.529,
      "cost": 45.071
},
```
vs on main branch
```
"summary": {
      ...
      "time": 61.177,
      "length": 1.529,
      "cost": 56.589
 },
```
This is okay on its own, but I expected raising `top_speed` would allow going up to the speed limit, which https://www.openstreetmap.org/way/48842029 has a tagged max speed of 70 MPH / 112 KPH. 

adding ~`"costing_options":{"top_speed":120}}`~  `"costing_options":{"truck":{"top_speed":120}}` doesn't take affect and still get the slower time that aligns with a speed of 90 KPH.

chrstnbwnkl:
# Issue
Fixes #4789. 

This properly enabled `top_speed` for trucks. Until now, the option was a ranged default, but even the allowed range was [10,252], the parameter was always clamped to the default value, 90. This PR addresses that with the following changes:

  - top speed now is clamped to the edge speed instead of `kMaxAssumedTruckSpeed`
  - when a `truck_speed` is set, choose the minimum of `truck_speed`, edge speed or `top_speed` 
  - `kMaxAssumedTruckSpeed` (and hence, `top_speed` default for truck) is now raised from 90 to 120 KPH. This is more in line with `kMaxAssumedSpeed` which at 140 kph is set at the higher end of how fast cars will actually go. I have gotten frequent feedback about 90 sometimes being to low, and in this case is really just necessary for build time estimation of truck durations for shortcuts to play well with run time dynamic costing
  - in shortcutbuilder, we previously set `truck_speed` to `kMaxAssumedTruckSpeed` i.e. 90KPH where no truck speed is set, but with `top_speed` now fully controllable, this really needs to be the edge's `speed` as the fallback

Thinking about it, though, I think #4493 and this PR really are just band aids for a bigger problem, which is trying to get good ETA's for truck routes... IMO this PR would improve the situation a bit by extending the flexibility of `top_speed`, but also worsen the default situation again. Anyway, leaving it here to be tried out and contested :smile:  

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/src/sif/autocost.cc
+++ b/src/sif/autocost.cc
@@ -79,6 +79,8 @@ constexpr ranged_default_t<float> kUseDistanceRange{0, kDefaultUseDistance, 1.0f
 constexpr ranged_default_t<float> kAutoHeightRange{0, kDefaultAutoHeight, 10.0f};
 constexpr ranged_default_t<float> kAutoWidthRange{0, kDefaultAutoWidth, 10.0f};
 constexpr ranged_default_t<uint32_t> kProbabilityRange{0, kDefaultRestrictionProbability, 100};
+constexpr ranged_default_t<uint32_t> kVehicleSpeedRange{10, baldr::kMaxAssumedSpeed,
+                                                        baldr::kMaxSpeedKph};
 
 constexpr float kHighwayFactor[] = {
     1.0f, // Motorway
@@ -694,6 +696,7 @@ void ParseAutoCostOptions(const rapidjson::Document& doc,
   JSON_PBF_DEFAULT(co, false, json, "/include_hot", include_hot);
   JSON_PBF_DEFAULT(co, false, json, "/include_hov2", include_hov2);
   JSON_PBF_DEFAULT(co, false, json, "/include_hov3", include_hov3);
+  JSON_PBF_RANGED_DEFAULT(co, kVehicleSpeedRange, json, "/top_speed", top_speed);
 }
 
 cost_ptr_t CreateAutoCost(const Costing& costing_options) {
--- a/src/sif/dynamiccost.cc
+++ b/src/sif/dynamiccost.cc
@@ -108,10 +108,10 @@ constexpr float kDefaultClosureFactor = 9.0f;
 // non-closure end
 constexpr ranged_default_t<float> kClosureFactorRange{1.0f, kDefaultClosureFactor, 10.0f};
 
-constexpr ranged_default_t<uint32_t> kVehicleSpeedRange{10, baldr::kMaxAssumedSpeed,
-                                                        baldr::kMaxSpeedKph};
 constexpr ranged_default_t<uint32_t> kFixedSpeedRange{0, baldr::kDisableFixedSpeed,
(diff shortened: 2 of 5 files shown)


src/sif/autocost.cc
```diff
@@ -694,6 +696,7 @@ void ParseAutoCostOptions(const rapidjson::Document& doc,
   JSON_PBF_DEFAULT(co, false, json, "/include_hot", include_hot);
   JSON_PBF_DEFAULT(co, false, json, "/include_hov2", include_hov2);
   JSON_PBF_DEFAULT(co, false, json, "/include_hov3", include_hov3);
+  JSON_PBF_RANGED_DEFAULT(co, kVehicleSpeedRange, json, "/top_speed", top_speed);
```
chrstnbwnkl, line 699: there was a previously unnoticed bug that prevented costing profiles to overwrite ranged defaults set in `dynamiccost`' base cost. Most likely affected `motorscootercost` as well

src/sif/dynamiccost.cc
```diff
@@ -402,9 +400,6 @@ void ParseBaseCostOptions(const rapidjson::Value& json,
   co->set_disable_hierarchy_pruning(
       rapidjson::get<bool>(json, "/disable_hierarchy_pruning", co->disable_hierarchy_pruning()));
 
-  // top speed
-  JSON_PBF_RANGED_DEFAULT(co, kVehicleSpeedRange, json, "/top_speed", top_speed);
-
   // destination only penalty
```
chrstnbwnkl, line 403: add it to bike as well

chrstnbwnkl: Actually, I just looked at it again and currently, top speed is not used at all in  bike and pedestrian (in `DynamicCost` it's used in `SpeedPenalty()`, but that's not being called in neither `PedestrianCost` nor `BicycleCost`, so deleting it from `DynamicCost` should be fine.

kevinkreiser (approved):

PR #4799, opened 2024-07-12
Handle list arguments with nargs instead of list type when overriding…

### LINKED ISSUE #4798: valhalla_config_build does not handle list arguments correctly
This issue was previously noted a few years ago, but it appears not to be fixed. See https://github.com/valhalla/valhalla/pull/3225

Essentially, when you build a config file with valhalla_build_config.py, and you pass in arguments to a path with a Python type of `list`, ArgumentParser will apply split() on the argument. That means the argument is streated as a string and splits it into characters, which is not the desired behavior.

Today, the only `list` type argument is `statsd-tags`. 
Example behavior:
```
% ./scripts/valhalla_build_config --statsd-tags list_item_ list_item_2
...
"statsd": {
    "port": 8125,
    "prefix": "valhalla",
    "tags": [
      "[",
      "l",
      "i",
      "s",
      "t",
      "_",
      "i",
      "t",
      "e",
      "m",
      "_",
      ",",
      "l",
      "i",
      "s",
      "t",
      "_",
      "i",
      "t",
      "e",
      "m",
      "_",
      "2",
      "]"
    ]
  },
...
```

The desired output is this:
```
% ./scripts/valhalla_build_config --statsd-tags ["list_item_1", "list_item_2"]
...
"statsd": {
    "port": 8125,
    "prefix": "valhalla",
    "tags": [
      "list_item_1",
      "list_item_2"
    ]
  },
...
```

asmyth01:
# Issue

fixes #4798 

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
    * REMOVED: needs_ci_run script [#4423](https://github.com/valhalla/valhalla/pull/4423)
    * REMOVED: unused vehicle types in AutoCost and segway; renamed kTruck to "truck" instead of "tractor_trailer" [#4430](https://github.com/valhalla/valhalla/pull/4430)
    * REMOVED: ./bench and related files/code [#4560](https://github.com/valhalla/valhalla/pull/4560)
+   * FIXED: Handle list type arguments correctly when overriding config with valhalla_build_config [#4799](https://github.com/valhalla/valhalla/pull/4799)
 * **Bug Fix**
    * FIXED: gcc13 was missing some std header includes [#4154](https://github.com/valhalla/valhalla/pull/4154)
    * FIXED: when reclassifying ferry edges, remove destonly from ways only if the connecting way was destonly [#4118](https://github.com/valhalla/valhalla/pull/4118)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
    * REMOVED: needs_ci_run script [#4423](https://github.com/valhalla/valhalla/pull/4423)
    * REMOVED: unused vehicle types in AutoCost and segway; renamed kTruck to "truck" instead of "tractor_trailer" [#4430](https://github.com/valhalla/valhalla/pull/4430)
    * REMOVED: ./bench and related files/code [#4560](https://github.com/valhalla/valhalla/pull/4560)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
    * REMOVED: needs_ci_run script [#4423](https://github.com/valhalla/valhalla/pull/4423)
    * REMOVED: unused vehicle types in AutoCost and segway; renamed kTruck to "truck" instead of "tractor_trailer" [#4430](https://github.com/valhalla/valhalla/pull/4430)
    * REMOVED: ./bench and related files/code [#4560](https://github.com/valhalla/valhalla/pull/4560)
+   * FIXED: Handle list type arguments correctly when overriding config with valhalla_build_config [#4799](https://github.com/valhalla/valhalla/pull/4799)
```
kevinkreiser, line 6: please move this below the bug fix heading

scripts/valhalla_build_config
```diff
@@ -596,7 +596,10 @@ def add_leaf_args(
             t = type(tree)
 
         arg = '--' + path.replace('_', '-').replace('\0', '-')
-        parser_.add_argument(arg, type=t, help=h, default=tree)
+        if t == list:
+            parser_.add_argument(arg, nargs="+", help=h, default=tree)
+        else:
+            parser_.add_argument(arg, type=t, help=h, default=tree)
```
kevinkreiser, line 602: nit: a ternary would be nicer imho:
```suggestion
        parser_.add_argument(arg, nargs="+" if t == list else t, help=h, default=tree)
```

kevinkreiser:
thank you for this cleanup!

scripts/valhalla_build_config
asmyth01: @kevinkreiser That would be more concise, but I'm afraid this isn't going to work in this case. Your suggestion would replace the value of `nargs` with the result of the ternary operator. What I really want to do here is use the `nargs` parameter only if `t==list`, else set the paramter `type=t`.

kevinkreiser: oh sorry i didnt look closely :smile: ignore the suggestion haha

kevinkreiser (approved):

PR #4805, opened 2024-07-16
docker uses newer ubuntu 24.04

### LINKED ISSUE #4796: Update Ubuntu base image version in Docker image
Ubuntu 23.04 has reached EOL and 24.04 is the latest LTS release.

kevinkreiser:
fixes #4796 (hopefully)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@
 #  binaries that can target the target architecture. from there bob is your uncle maybe..
 
 ####################################################################
-FROM ubuntu:23.04 as builder
+FROM ubuntu:24.04 as builder
 MAINTAINER Kevin Kreiser <kevinkreiser@gmail.com>
 
 ARG CONCURRENCY
@@ -54,7 +54,7 @@ RUN rm -rf valhalla
 
 ####################################################################
 # copy the important stuff from the build stage to the runner image
-FROM ubuntu:23.04 as runner
+FROM ubuntu:24.04 as runner
 MAINTAINER Kevin Kreiser <kevinkreiser@gmail.com>
 
 # basic paths


kevinkreiser:
this doesnt work yet, deps script fails on python deps. i'll patch it shortly

chrstnbwnkl:
seems like shapely 2.x is making trouble, since apt installs 1.x. Why exactly is shapely installed additionally via pip? :smile:

kevinkreiser:
ok this is building to completion now however i feel like we should test all the python scripts that we distribute with the installation to make sure they all work

Dockerfile
```diff
@@ -67,15 +67,15 @@ LABEL org.opencontainers.image.source = "https://github.com/valhalla/valhalla"
...
 
 # we need to add back some runtime dependencies for binaries and scripts
 # install all the posix locales that we support
 RUN export DEBIAN_FRONTEND=noninteractive && apt update && \
     apt install -y \
-      libcurl4 libczmq4 libluajit-5.1-2 libgdal32 \
+      libcurl4 libczmq4 libluajit-5.1-2 libgdal34 \
       libprotobuf-lite32 libsqlite3-0 libsqlite3-mod-spatialite libzmq5 zlib1g \
-      curl gdb locales parallel python3-minimal python3-distutils python-is-python3 \
+      curl gdb locales parallel python3-minimal python3-setuptools python-is-python3 \
```
chrstnbwnkl, line 78: Do we need `setuptools`? I know `distutils` was recently deprecated and some of that functionality moved to setuptools, but I think it was only used for that `strtobool` utility function used in `valhalla_build_config`. But that got replaced in #4017

kevinkreiser: probably not then! that was why I was suggesting we test the Python bits but now that you mention it I think thats the only thing I was wondering about

chrstnbwnkl:
Would it be a good idea to add a build argument to the Dockerfile to optionally run tests at build time? That would help make these dependency updates easier, at least wrt all the python scripts

kevinkreiser:
it doesnt hurt for sure

kevinkreiser:
merged this will fix if its problematic

PR #4811, opened 2024-07-19
Revive `max_distance` for pedestrian costing

### LINKED ISSUE #4804: `max_distance` broken in Pedestrian routing
Seems like an implicit type conversion here:

https://github.com/valhalla/valhalla/blob/e73131c6e20a7c06538f52c239f62f3542876fcf/src/sif/pedestriancost.cc#L651

That would explain why no path is only found when `max_distance > 1.0`

chrstnbwnkl:
# Issue

Fixes #4804 . What I though was a small bug caused by an unwanted type conversion turned into a bit of a scavenger hunt :smile: I'm not sure how `max_distance` was so neglected, because it turns out that it just didn't work with bidirectional Astar, since that  algorithm doesn't properly track path distances. Did pedestrian pathfinding use a unidirectional algorithm when `max_distance` was first introduced? 

Anyway,  this is how I'm bringing it back to life:

  - allow edges on both search branches to be added to the queue unless their path distance exceeds the total max distance. Cutting that distance in half wouldn't make much sense, because search paths don't necessarily meet half way (maybe it could be set lower so the algorithm can bail earlier, but then the costing needs to be aware of which algorithm is calling it, and we need to agree on an appropriate percentage of max distance)
  - only allow a connection to be marked as found when the accumulated path distance does not exceed the allowed distance 
 

## Tasklist

 - [x] Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
    * FIXED: Missing algorithm include in `baldr/admin.h` [#4766](https://github.com/valhalla/valhalla/pull/4766)
    * FIXED: Handle list type arguments correctly when overriding config with valhalla_build_config [#4799](https://github.com/valhalla/valhalla/pull/4799)
    * FIXED: `top_speed` range not fully allowed for trucks [#4793](https://github.com/valhalla/valhalla/pull/4793)
+   * FIXED: fix `max_distance` for pedestrian costing [#4811](https://github.com/valhalla/valhalla/pull/4811)
 
 * **Enhancement**
    * UPDATED: French translations, thanks to @xlqian [#4159](https://github.com/valhalla/valhalla/pull/4159)
--- a/src/sif/dynamiccost.cc
+++ b/src/sif/dynamiccost.cc
@@ -120,8 +120,8 @@ BaseCostingOptionsConfig::BaseCostingOptionsConfig()
     : dest_only_penalty_{0.f, kDefaultDestinationOnlyPenalty, kMaxPenalty},
       maneuver_penalty_{0.f, kDefaultManeuverPenalty, kMaxPenalty},
       alley_penalty_{0.f, kDefaultAlleyPenalty, kMaxPenalty},
-      gate_cost_{0.f, kDefaultGateCost, kMaxPenalty}, gate_penalty_{0.f, kDefaultGatePenalty,
-                                                                    kMaxPenalty},
+      gate_cost_{0.f, kDefaultGateCost, kMaxPenalty},
+      gate_penalty_{0.f, kDefaultGatePenalty, kMaxPenalty},
       private_access_penalty_{0.f, kDefaultPrivateAccessPenalty, kMaxPenalty},
       country_crossing_cost_{0.f, kDefaultCountryCrossingCost, kMaxPenalty},
       country_crossing_penalty_{0.f, kDefaultCountryCrossingPenalty, kMaxPenalty},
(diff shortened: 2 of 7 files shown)


kevinkreiser:
so for regular routing bidir a* for pedestrian im not sure we ever cared about max distance in costing, instead we enforce this as a request limit and look at the crow flies distance.

max_distance instead was primarily used, iirc for limiting the walking one would have to do in a multimodal route.

i suppose we could adhere to it for regular routing but its pretty well redundant with the initial distance check. like if the requester is telling you they dont want to walk further than 10km but the service limit for pedestrian is 1000km then but they send us 2 locations that are 15km apart why would we compute the route until the costing tells us to bail. why not just look at the crows flies and either let the route through or dont. so i guess if we want to allow potentially wasteful computation we should at least add a check in worker.cc that checks if you want a max distance limit and immediately fail the request if the locations are too far apart.

i personally dont think its a very compelling use case but there are edge cases where straight line distance is a poor estimate of the final route length (low network density areas or areas with big water features etc). fairly rare tbh but it happens.

chrstnbwnkl:
Ahh yeah that makes much more sense, then the original meaning just kinda got lost :smile: The initial check is a no-brainer, you're right. I have a specific case where a location snaps to a ferry edge in a pedestrian route request, and since the road between the ferry  and the rest of the street network is marked as `foot=no`, Valhalla does not find a route, but only after searching for what seems like a long time.

The request: https://valhalla.openstreetmap.de/directions?profile=pedestrian&wps=2.161216735839844%2C41.38466556442377%2C2.1835756301879887%2C41.36402281373148


This is the way that is essentially blocking this: https://www.openstreetmap.org/way/132559639#map=16/41.3642/2.

Granted, this is the very definition of an edge case, but a bit of an annoying one :smile: even with hard avoids on ferries (#4524) this problem would persist (unless snapping to ferry edges would be disallowed), and lower service limits for pedestrian and crow flies checks wouldn't do anything in that case.

kevinkreiser:
> Valhalla does not find a route, but only after searching for what seems like a long time.

well for that case we usually rely on the number of iterations in the algorithm determining when to give up. i thought we made that configurable but im doubting my memory. i wonder if there is a misconfiguration there somewhere

anyway, i think this is getting to the heart of the issue, which is that `max_distance` is a user facing preference for shaping the kinds of routes you get back whereas the iterations and crows flies service limits checks are about operational diligence (protecting the service from abuse or wasted cycles).

chrstnbwnkl:
I get your point about `max_distance` being user facing, so having something configurable on the server side would be the better approach. 

In bidir astar, seems like there is the `iterations_threshold_` member. Can't find any configuration option, it's just set to numeric_limits of `float` in the c-tor and then lowered appropriately as connections are found. 

Anyway, I'm fine with limiting `max_distance` to a simple crow flies check, for the multimodal stuff there now are `transit_start_end_max_distance` and `transit_transfer_max_distance` anyway.

nilsnolde:
We’re snapping pedestrian to ferry edges? If that’s right, it sounds pretty wrong 😄 And seems like a severe mapping error when the only way off a ferry prohibits ped. 

I think it’d be better to document max_distance properly (as part of multimodal) and leave it pretty much as it is. Using it for the same endpoint with a different meaning would be pretty confusing IMO. Also I don’t really see the benefit: if a user doesn’t want to walk more than 10 km by crow distance, it can be done client side.

chrstnbwnkl:
> and leave it pretty much as it is.

Right, except it's broken as documented in #4804 :smile: So I would at least fix that and update the documentation.

chrstnbwnkl:
Hold on, if the only use case for `max_distance` is mm, then `max_distance` has no business being in `Allowed` anyway, right? 

https://github.com/valhalla/valhalla/blob/e73131c6e20a7c06538f52c239f62f3542876fcf/src/sif/pedestriancost.cc#L650-L651

Since everything related to maximum walking distance is done in the multimodal algorithms? I don't see it being called anywhere else though. Was it replaced by `transit_start_end_max_distance` and `transit_transfer_max_distance`?

PR #4812, opened 2024-07-22
Hard avoid ferries

chrstnbwnkl:
# What? 

Adds two request parameters: 

  - `exclude_ferry` costing option. When set to `true`, ferries will be completely disallowed on a route; defaults to `false` 
  - `exclude_ferry` location search filter. When passed in a location object, that location will not snap to ferry edges

# Why?

The `use_ferry` and `ferry_cost` parameters can be used to heavily penalize ferries, but they will never be fully excluded. I think adding the possibility to fail in those cases can be a feature. There are good reasons for users to want to absolutely avoid ferries, e.g. due to missing schedules, or simply an unwillingness to use that mode of transport (anyone who's ever been seasick will agree :smile:).

# Notes 

I used a previous PR (#4524) as a basis but I ended up removing pretty much all of it. Excluding ferries does not seem to be as much of an issue wrt performance as other excludes that previous PRs attempted to implement, so I think there is no need to hide it behind a server side config option.


valhalla/sif/edgelabel.h
```diff
@@ -446,6 +446,7 @@ class EdgeLabel {
   uint32_t has_measured_speed_ : 1;
   // Flag if this edge had HGV access
   uint32_t hgv_access_ : 1;
+  // Flag indicating edge is a bridge.
```
chrstnbwnkl, line 449: this must  go

chrstnbwnkl:
I ended up moving the `exclude_ferry` check in `Allowed` and `AllowedReverse` just before `IsClosed` in the motorized costing profiles, and left the order untouched in the other ones because I think ferry edge occurence is statistically rare anyway, plus I'm not sure it will be the most widely used costing option anyway, so other checks are likelier to evaluate to true anyway.

dnesbitt61:
Is there anything to prevent a long running request where a ferry must be taken to complete the route? A request would fail only when the forward or reverse search trees are exhausted in cases where a ferry is the only connection. England to anywhere in Europe is one case, but there may be worse ones?

chrstnbwnkl:
Good point @dnesbitt61... Is this something that server side config parameters that limit the maximum number of iterations for each algorithm could mitigate?

kevinkreiser:
yeah that is what the iterations limit is meant to do. at the moment its just a general limit but one could imagine having the number based on request parameters (hard avoids) or distance between locations. would take some tweaking to get it right for sure but it might be a better experience operationally and from the users perspective

kevinkreiser:
as a quicker way to get this merged we could maybe consider having a config option to disable the ability to request such a hard avoid to mitigate peoples ability to DOS a service that didnt want to allow for such traffic

PR #4817, opened 2024-07-26
costmatrix cleanup was not resetting properly

chrstnbwnkl:
# Issue

In `CostMatrix`, when second pass is used, `not_thru_pruning` is set to false to increase the chances of finding a route, but the reset was missing afterwards.

### DIFF
--- a/src/thor/costmatrix.cc
+++ b/src/thor/costmatrix.cc
@@ -71,7 +71,7 @@ void CostMatrix::Clear() {
   if (check_reverse_connections_)
     sources_->clear();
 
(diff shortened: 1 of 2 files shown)


src/thor/costmatrix.cc
```diff
@@ -105,6 +105,7 @@ void CostMatrix::Clear() {
     astar_heuristics_[is_fwd].clear();
   }
   best_connection_.clear();
+  set_not_thru_pruning(true);
```
kevinkreiser, line 108: oops!

kevinkreiser (approved):

PR #4828, opened 2024-08-02
Fix oneway ferry connections classification

kinkard:
# Issue

Oneway ferry connections were not properly classified which caused really interesting behavior - short route can still utilize that "low class edge" but after some margin route starts to fail because of the hierarchy limits.

### Route example

Ok route - https://valhalla.openstreetmap.de/directions?profile=car&wps=18.297982%2C57.6379361%2C18.10152053833008%2C59.26772287446247
image

Failed route - https://valhalla.openstreetmap.de/directions?profile=car&wps=18.297982%2C57.6379361%2C18.0710935%2C59.3251172
image

Failed route expansion graph
image

### Real world example

Example of such separate oneway connections - https://www.openstreetmap.org/#map=18/58.90352/17.95579
image

### Before this PR

```
$ ./valhalla_build_tiles -c valhalla.json sweden-latest.osm.pbf
...
[INFO] Finished ReclassifyFerryEdges: ferry_endpoint_count = 145, 3981 edges reclassified. Failed both directions for 54 connections.
...
```

### After this PR

```sh
$ ./valhalla_build_tiles -c valhalla.json sweden-latest.osm.pbf
...
[INFO] Finished ReclassifyFerryEdges: ferry_endpoint_count = 145, 3987 edges reclassified. Failed both directions for 54 connections.
..
```

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/mjolnir/ferry_connections.cc
```diff
@@ -380,35 +380,46 @@ void ReclassifyFerryConnections(const std::string& ways_file,
...
+          }
         } else {
           // Check if oneway inbound to the ferry
           bool inbound =
               (edge.first.sourcenode_ == node_itr.position()) ? edge_rev_access : edge_fwd_access;
           auto ret = ShortestPath(node_itr.position(), end_node_idx, ways, way_nodes, edges, nodes,
                                   inbound, remove_destonly);
           total_count += ret.first;
-          if (ret.second > 0) {
+          if (ret.second) {
```
kinkard, line 402: I can't pass by comparing bool with zero

src/mjolnir/ferry_connections.cc
```diff
@@ -380,35 +380,46 @@ void ReclassifyFerryConnections(const std::string& ways_file,
...
-        }
-        ferry_endpoint_count++;
 
-        // Reclassify the first/start edge if a connection to higher class roads
-        // is found. Do this AFTER finding shortest path so we do not immediately
-        // determine we hit the specified classification
-        sequence<Edge>::iterator element = edges[edge.second];
-        auto update_edge = *element;
-        if (inbound_path_found && outbound_path_found &&
-            update_edge.attributes.importance > kFerryUpClass) {
```
kinkard, line 406: The problem was in this condition - reclassification happens only if both inbound and outbound edges are met. So the first one will not be reclassified

src/mjolnir/ferry_connections.cc
```diff
@@ -380,35 +380,46 @@ void ReclassifyFerryConnections(const std::string& ways_file,
...
-          total_count++;
+          // Reclassify the first/start edge if a connection to higher class roads
+          // is found. Do this AFTER finding shortest path so we do not immediately
+          // determine we hit the specified classification
+          sequence<Edge>::iterator element = edges[edge.second];
+          auto update_edge = *element;
+          if (ret.second && update_edge.attributes.importance > kFerryUpClass) {
+            update_edge.attributes.importance = kFerryUpClass;
+            update_edge.attributes.reclass_ferry = remove_destonly;
+            element = update_edge;
```
kevinkreiser, line 418: element is really just a temporary right? cant we just remove this line?
```suggestion
```

kevinkreiser: oh no we are writing the updates back to the sequence haha sorry

src/mjolnir/ferry_connections.cc
```diff
@@ -380,35 +380,46 @@ void ReclassifyFerryConnections(const std::string& ways_file,
...
-          update_edge.attributes.importance = kFerryUpClass;
-          update_edge.attributes.reclass_ferry = remove_destonly;
-          element = update_edge;
-          total_count++;
+          // Reclassify the first/start edge if a connection to higher class roads
+          // is found. Do this AFTER finding shortest path so we do not immediately
+          // determine we hit the specified classification
+          sequence<Edge>::iterator element = edges[edge.second];
+          auto update_edge = *element;
+          if (ret.second && update_edge.attributes.importance > kFerryUpClass) {
```
kevinkreiser, line 415: nit: the `ret.second` check that you fixed above is already happening in this scope, you could just move this inside there?

src/mjolnir/ferry_connections.cc
```diff
@@ -380,35 +380,46 @@ void ReclassifyFerryConnections(const std::string& ways_file,
...
+            update_edge.attributes.importance = kFerryUpClass;
+            update_edge.attributes.reclass_ferry = remove_destonly;
+            element = update_edge;
+            total_count++;
+          }
         }
+        ferry_endpoint_count++;
       }
 
       // Log cases where reclassification fails
```
kevinkreiser, line 425: do we need to update the logging to not report false positives for failures when there are ferry connections that can only happen in or outbound? i dont know of any such mapped places but maybe there are ferries that only let you get off and drive away from the ferry and spots where you can only board the ferry (like separate entrance and exit). i suppose at the moment the code would need another else if (and test case) to cover the outbound only case?

kevinkreiser: maybe we should just always run whichever ones have access and reclass whichever ones succeed. like if its forward access check outbound if it has reverse access check inbound and whichever ones work we'll reclass them. i havent looked at this in a long while so im probably oversimplifying it :smile:

kinkard: > i suppose at the moment the code would need another else if (and test case) to cover the outbound only case?

Actually this PR fixes such case as well, I'll cover it by test.

> maybe we should just always run whichever ones have access and reclass whichever ones succeed. like if its forward access check outbound if it has reverse access check inbound and whichever ones work we'll reclass them.

Well, currently it should work as the following:
- two way edge - reclassify it only if we found both inbound and outbound shortest paths via it
- one way edge - reclassify it if we found the corresponding (inbound for inbound edge and outbound for outbou nd edge) shortest paths

kinkard: > Actually this PR fixes such case as well, I'll cover it by test.

Please take a look - 078ac205b274beb3d4ac671a22e85d6c92f9301d

kevinkreiser (approved):

PR #4829, opened 2024-08-02
Remove unused headers

kinkard:
# Issue

This PR removes unused C++ headers. There is no direct value from that aside small build time speed up, which is always nice

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 ## Unreleased
 * **Removed**
-   * REMOVED: unused headers [#4829](https://github.com/valhalla/valhalla/pull/4829)
    * REMOVED: needs_ci_run script [#4423](https://github.com/valhalla/valhalla/pull/4423)
    * REMOVED: unused vehicle types in AutoCost and segway; renamed kTruck to "truck" instead of "tractor_trailer" [#4430](https://github.com/valhalla/valhalla/pull/4430)
    * REMOVED: ./bench and related files/code [#4560](https://github.com/valhalla/valhalla/pull/4560)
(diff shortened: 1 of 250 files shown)


kevinkreiser:
very nice many thanks for this! we always wanted to do the include what you use cleanup, i think its mentioned somewhere in some issue. would be great to have a linter rule for this.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 * **Removed**
+   * REMOVED: unused headers [#4829](https://github.com/valhalla/valhalla/pull/4829)
```
kevinkreiser, line 3: we do the changelog entries such that new ones are at the bottom of the current unreleased section

src/baldr/streetname_us.cc
```diff
@@ -1,7 +1,5 @@
-#include <iostream>
-
-#include "baldr/streetname.h"
 #include "baldr/streetname_us.h"
+#include "baldr/streetname.h"
```
kevinkreiser, line 2: ill be shocked if the linter doesnt complain about reordering these (i think its alphabetically sorts them)

kevinkreiser: strange, i guess it doesnt care about sorting then!

kinkard: I believe `clang-format` has a specific rule for the header that has the same name as the current source file.

kinkard:
I used really opiniated linter for this cleanup - https://gist.github.com/kinkard/f0d41b4e5f1f29347b8bdb4cb723defa

PR #4835, opened 2024-08-06
fix location search_filter for node snaps

kevinkreiser:
this morning @chrstnbwnkl and i debugged an issue with another pr that is currently open. we found that the search filter was not always filtering out candidate edges that should be.

after a few minutes of looking we quickly noticed that indeed, we use the search filter when deciding the initial candidates to use but if a candidate becomes a node snap, the other edges connected at the node (which were likely already filtered) are not checked a second time, instead they are simply included.

this pr fixes that by again checking if an edge should be filtered before including it.

more info here: https://github.com/valhalla/valhalla/pull/4524#issuecomment-2269095092

also it seems that we found this issue in testing before but werent able to debug to find out what was going wrong. so it seems another test breaks and we'll need to patch that up before merging this.

### DIFF
--- a/src/loki/search.cc
+++ b/src/loki/search.cc
@@ -22,23 +22,22 @@ template <typename T> inline T square(T v) {
   return v * v;
 }
 
-bool is_search_filter_triggered(const DirectedEdge* edge,
-                                const DynamicCost& costing,
-                                const graph_tile_ptr& tile,
-                                const Location::SearchFilter& search_filter) {
+bool search_filter(const DirectedEdge* edge,
+                   const DynamicCost& costing,
+                   const graph_tile_ptr& tile,
+                   const Location::SearchFilter& filter) {
   // check if this edge matches any of the exclusion filters
   uint32_t road_class = static_cast<uint32_t>(edge->classification());
-  uint32_t min_road_class = static_cast<uint32_t>(search_filter.min_road_class_);
-  uint32_t max_road_class = static_cast<uint32_t>(search_filter.max_road_class_);
+  uint32_t min_road_class = static_cast<uint32_t>(filter.min_road_class_);
+  uint32_t max_road_class = static_cast<uint32_t>(filter.max_road_class_);
 
   // Note that min_ and max_road_class are integers where, by default, max_road_class
   // is 0 and min_road_class is 7. This filter rejects roads where the functional
   // road class is outside of the min to max range.
   return (road_class > min_road_class || road_class < max_road_class) ||
-         (search_filter.exclude_tunnel_ && edge->tunnel()) ||
-         (search_filter.exclude_bridge_ && edge->bridge()) ||
-         (search_filter.exclude_ramp_ && (edge->use() == Use::kRamp)) ||
-         (search_filter.exclude_closures_ && (costing.flow_mask() & kCurrentFlowMask) &&
+         (filter.exclude_tunnel_ && edge->tunnel()) || (filter.exclude_bridge_ && edge->bridge()) ||
+         (filter.exclude_ramp_ && (edge->use() == Use::kRamp)) ||
+         (filter.exclude_closures_ && (costing.flow_mask() & kCurrentFlowMask) &&
           tile->IsClosed(edge));
 }
 
@@ -323,7 +322,7 @@ struct bin_handler_t {
         // do we want this edge, note we have to re-evaluate the filter check because we may be
         // seeing these edges a second time (filtered out before)
         if (costing->Allowed(edge, tile, kDisallowShortcut) &&
-            !is_search_filter_triggered(edge, *costing, tile, location.search_filter_)) {
+            !search_filter(edge, *costing, tile, location.search_filter_)) {
           auto reach = get_reach(id, edge);
           PathLocation::PathEdge
               path_edge{id,   0, node_ll, distance, PathLocation::NONE, reach.outbound, reach.inbound,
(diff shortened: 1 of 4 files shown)


test/gurka/test_search_filter.cc
```diff
@@ -61,6 +61,23 @@ TEST_F(SearchFilter, Unfiltered) {
   gurka::assert::osrm::expect_steps(result, {"AB", "BC"});
   gurka::assert::raw::expect_path(result, {"AB", "BC"});
 }
+TEST_F(SearchFilter, NodeSnapped) {
+  auto from = "B";
+  auto to = "C";
```
kevinkreiser, line 66: b to c is a tunnel way, so if we node snap to b, the edge bc should not be allowed as a candidate. before the fix it was and this test failed. with the update the test passes and the route goes around the tunnel.

test/gurka/test_traffic.cc
```diff
@@ -110,7 +110,6 @@ TEST(Traffic, BasicUpdates) {
                "it's noticed the changes in the live traffic file"
             << std::endl;
   {
-
```
kevinkreiser, line 113: the test below this one was failing at first because i blocked an edge and its opp edge if edge was hard filtered. the problem with that is that traffic closures can close only one direction of an edge which means i needed to check both directions separately. which you'll see in search.cc

src/loki/search.cc
```diff
@@ -532,6 +535,13 @@ struct bin_handler_t {
       bool all_prefiltered = true;
       for (p_itr = begin; p_itr != end; ++p_itr, ++c_itr) {
         c_itr->sq_distance = std::numeric_limits<double>::max();
+        // TODO: for closings its possible that we could use the opposing edge if edge is filtered
```
kevinkreiser, line 538: imho we should solve this now so we dont have to debug this later :smile: its hard to write a test for this becasue it depends on order of edges in the data but i think we can just fix the logic and allow the rest of the tests to pass as a good enough signal its not broken anything.

kevinkreiser: ill take care of it tomorrow or so

kevinkreiser:
ok last commit should take care of full robustness to this problem. was a bit more of a change than i liked but sadly its a bit complicated

src/loki/search.cc
```diff
@@ -22,23 +22,22 @@ template <typename T> inline T square(T v) {
   return v * v;
 }
 
-bool is_search_filter_triggered(const DirectedEdge* edge,
-                                const DynamicCost& costing,
-                                const graph_tile_ptr& tile,
-                                const Location::SearchFilter& search_filter) {
```
kevinkreiser, line 28: no change here other than shortening names. since we call this a bunch more now an inside a bunch of `if`s i really wanted a shorter method name and to make it more closely match the other similar filter ones

src/loki/search.cc
```diff
@@ -320,8 +319,10 @@ struct bin_handler_t {
             tangent_angle(index, candidate.point, info.shape(),
                           GetOffsetForHeading(edge->classification(), edge->use()), edge->forward());
         auto layer = info.layer();
-        // do we want this edge
-        if (costing->Allowed(edge, tile, kDisallowShortcut)) {
+        // do we want this edge, note we have to re-evaluate the filter check because we may be
+        // seeing these edges a second time (filtered out before)
+        if (costing->Allowed(edge, tile, kDisallowShortcut) &&
+            !search_filter(edge, *costing, tile, location.search_filter_)) {
```
kevinkreiser, line 325: this one and the one right below here are in the `correlate_node` method. this was the primary bug we found in real test cases. node snapping is common in gurka because of.. conventions :smile:

src/loki/search.cc
```diff
@@ -418,19 +420,22 @@ struct bin_handler_t {
       PathLocation::PathEdge path_edge{candidate.edge_id, length_ratio, candidate.point,
                                        distance,          side,         reach.outbound,
                                        reach.inbound,     angle};
-      // correlate the edge we found
-      if (side_filter(path_edge, location, reader) || heading_filter(location, angle) ||
-          layer_filter(location, layer)) {
+      // correlate the edge we found if its not filtered out
```
kevinkreiser, line 423: ok so here inside `correlate_edge` we still need to re-evaluate the search filter because its possible that one of the edges is only filtered because its a traffic closure, thus leaving only one of the directions of the edge pair open to this location. it shouldnt be possible to get here in a case where both directions are filtered or otherwise not allowed so i didnt had a check for it. if it happens, and it shouldnt (would be a logic error), the result will be no candidates for a given location (and hopefully a bug report)

src/loki/search.cc
```diff
@@ -514,14 +519,22 @@ struct bin_handler_t {
         continue;
       }
 
+      // lots of places below where we might like to know about the opp edge
+      const DirectedEdge* opp_edge = nullptr;
+      graph_tile_ptr opp_tile = tile;
+      GraphId opp_edgeid;
```
kevinkreiser, line 525: there are now 3 different places where we fallback to testing the opposing edge, one for `allowed`, one for `search_filter` and one for `reach` check. it was messy but i pulled the declaration of these up here so i could use the same ones for everything.

kevinkreiser: would be nice if there were an easy way to memoize the getting of the opposing triplet of variables

chrstnbwnkl (approved):

PR #4842, opened 2024-08-07
Unfinished: Multithreaded Mjolnir parser

SwedishBarbarossa:
# Issue

**Consider this an unfinished a work in progress.**

Makes the mjolnir `OSMPBF::Parser::parse` method multithreaded. I implemented a threadpool class for it to do that, which is in `mjolnir/util.cc`.

The threadpool means a Parser class object must now be instantiated to work.

## Things i suggest checking before a merge

 - [ ]  Make sure it actually works and doesn't corrupt the data or crash repeatedly etc. I could not get the unit tests to run - I'm not used to C++ development.
 - [ ]  Make sure it doesn't leak memory.
 - [ ]  Fetch the configured thread count where the `Parse` function is called. I hard-coded them as 4 in this commit. I'm not 100% sure I got all of them - but the compiler should have complained if I missed one.
 - [ ]  I think I forgot to remove unused imports :cold_sweat:.
 - [ ]  I have genuinely no idea what this free method did, and it seems I removed it by mistake:
 ```cpp
void Parser::free() {
  google::protobuf::ShutdownProtobufLibrary();
}
```
 - [ ]  Get it to adhere to the repo style.

## Background

I started on this loosely based on a stated intent to implement multithreading [here](https://github.com/valhalla/valhalla/issues/4801). Made the PR based on [this](https://github.com/valhalla/valhalla/discussions/4841) exchange.

Even if its just a jumping off place to see what actually compiled and worked before you implement a better solution, I thought it was a waste not to share before I move on.

### DIFF
--- a/src/mjolnir/landmarks.cc
+++ b/src/mjolnir/landmarks.cc
@@ -323,8 +323,9 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
   }
 
   LOG_INFO("Parsing nodes and storing landmarks...");
+  OSMPBF::Parser parser(4);
   for (auto& file_handle : file_handles) {
-    OSMPBF::Parser::parse(file_handle, static_cast<OSMPBF::Interest>(OSMPBF::Interest::NODES),
+    parser.parse(file_handle, static_cast<OSMPBF::Interest>(OSMPBF::Interest::NODES),
                           callback);
   }
 
--- a/src/mjolnir/osmpbfparser.cc
+++ b/src/mjolnir/osmpbfparser.cc
@@ -285,42 +285,70 @@ Member::Member(Member&& other)
     : member_type(other.member_type), member_id(other.member_id), role(std::move(other.role)) {
 }
 
+// parser constructor
+Parser::Parser(int numThreads) : thread_pool(numThreads) {}
+
 void Parser::parse(std::ifstream& file, const Interest interest, Callback& callback) {
+  // Read file in the main thread and queue tasks for worker threads
+  read_file(file, interest, callback);
+
+  // Wait for all jobs to complete
+  thread_pool.WaitForAllJobs();
+}
+
+void Parser::read_file(std::ifstream& file, const Interest interest, Callback& callback) {
   char* buffer = new char[MAX_UNCOMPRESSED_BLOB_SIZE];
   char* unpack_buffer = new char[MAX_UNCOMPRESSED_BLOB_SIZE];
 
-  // start from the top
+  // Start from the top
   file.clear();
   file.seekg(0, std::ios::beg);
 
-  // while there is more to read
+  // While there is more to read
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: src/mjolnir/landmarks.cc [hunk 1]
```diff
@@ -323,8 +323,9 @@ bool BuildLandmarkFromPBF(const boost::property_tree::ptree& pt,
   }
 
   LOG_INFO("Parsing nodes and storing landmarks...");
+  OSMPBF::Parser parser(4);
```
kinkard, line 326: Better to use something like `"mjolnir.concurrency"` or `std::thread::hardware_concurrency()` instead of hardcoded `4`, e.g. like here: https://github.com/valhalla/valhalla/blob/58a289de51638c55f479101bf5352160902e0182/src/mjolnir/graphbuilder.cc#L1473-L1475

src/mjolnir/osmpbfparser.cc
```diff
@@ -285,42 +285,70 @@ Member::Member(Member&& other)
...
-        parse_header_block(unpack_buffer, sz);
-      } else {
-        LOG_WARN("Unknown blob type: " + header.type());
-      }
+
+      // Create a task and queue it for processing
+      Task task = { buffer, unpack_buffer, header, sz };
+      thread_pool.QueueJob([this, task, interest, &callback]() {
+        this->process_task(task, interest, callback);
+      });
```
kinkard, line 321: Unfortunately the `callback` here get mutated in this function, creating a race condition - multiple threads will mutate the same `callback` simultaneously.

kevinkreiser:
yeah as @kinkard mentioned, calling the `osmgraphparser` callback are not threadsafe. when you call them they are all trying to modify the same internal state (which is mostly attached to `osmdata_`).

this was exactly my worry when i had described how to parallelize this before. i'll quickly summarize that again here for posterity.

to do this we need to add the parallelism in pbfgraphparser itself not in osmpbfparser. what should happen is that we make threads in pbfgraphparser and each of those writes to their own sequence file (via the callbacks) from their own osmpbfparser. at the end of each parse pass we'll have `n` sequence files which we'll want to merge sort.

i have to refresh my memory on what gotchas might be in there regarding some global state that all threads might need access to. basically anything that accesses `osmdata_` in the callbacks. a lot of it is stats and can be combined later but some of it is not so simple like the index for the names, combining them would mean incrementing them by the size of the previous threads name indices. anyway it wont be completely trivial.

its not that one cannot parallelize the processing in the way that is suggested here but one would have to put synchronization inside the callback function whenever `osmdata_` is being modified. without reviewing that again its hard to say whether or not this would be prohibitive or not. its probably worth looking at as its certainly a simplier implementation than having to write the thread result merging code however its also easier to screw up in the future (ie by forgetting to think about the implications of not synchronizing something in the callback).

PR #4846, opened 2024-08-12
perf: Move encoded elevation to avoid redundant allocation

kinkard:
# Issue

Basically this PR fixes `performance-move-const-arg` in this function and thus avoids redundant allocation during the setting elevation data into EdgeInfo.
```cpp
// Set the encoded elevation vector.
void EdgeInfoBuilder::set_encoded_elevation(const std::vector<int8_t>& encoded_elevation) {
  if (!encoded_elevation.empty()) {
    encoded_elevation_ = std::move(encoded_elevation);
  }
}
```

## Tasklist

 - [ ] Add tests

## Requirements / Relations


kinkard:
I would probably do a bigger scope with `performance-move-const-arg` fixes in another PR

Issue #4553, opened 2024-02-04
maxspeed:conditional is not considered


SamuelBrucksch:
I just checked `maxspeed` property of an edge for a road with conditional speed limit, and it seems to ignore the `maxspeed:conditional` property.

Then i checked the lua file and it indeed only uses the `maxspeed` tag.

It's however not as easy as to replace it in lua, as conditional speed limits usually mean parsing the time and date, when they should be applied.

For me it would also be fine, to only get it as metadata, like the string in a property like `speedLimit_raw`, which contains the tag value, which is not processed, as we only need it to display the current speed limit, it shouldn't affect the actual planning too much anyways.

Would it be easily possible, to just add an additional property to the edge data, or would that rather require a bigger change? Is it maybe possible, to already attach additional data somewhow? Would it be possible, for example to replace an existent value, which is not so important, to contain the data we need, so the effort implementation wise is not that high?

What do you think would b the easiest way to do it?

Apart from that, it could actually also be handled by Valhalla, when a time is passed, valhalla could check if the time of arrival at that edge is within the conditional time frame or not and use the conditional speed limit instead then.

Here is a sample way, which has such a conditional speed limit:
https://www.openstreetmap.org/way/25575714

nilsnolde:
it could be handled the same as the other conditional `AccessRestrictions` (though it's not one but that's what the structure is called). IRC there's plenty of "space" and those tags are so sparse, no problem adding them I guess  (it's doing binary search to retrieve access restrictions). that would at least make them part of the graph and could be retrieved by `trace_attributes`/`locate` or so. I might not think of some gotcha though.

it could also be used in costing, don't see a big problem there off the top of my head. 

in summary, all the bits & pieces to fairly easily add them are already there (specifically parsing the time component), you'll just need to follow the code paths around access restrictions.

PR #4851, opened 2024-08-15
Provide conditional speed limits from "maxspeed:conditional" in `/locate` and proto `/route` responses

kinkard:
# Issue

fixes https://github.com/valhalla/valhalla/issues/4553

This PR adds "maxspeed:conditional" osm key parsing, that then stored into `EdgeInfo`.

- Added new 'conditional_speed_limits' field to the trip proto `/route` response
- Added new "conditional_speed_limits" field into the "edge_info" `/locate` response, for example:
```json
{
  "edge_info": {
    "bike_network": {
      "local": false,
      "mountain": false,
      "national": false,
      "regional": false
    },
    "conditional_speed_limits": {
      "19:00-06:00": 130
    },
    "mean_elevation": null,
    "names": [
      "A2",
      "E 35"
    ],
    "shape": "clasbB_hznHih@jZmg@fXm~B|fAc~BheAqyE|wByjGrqCcy@n[u{@nZ_uGdkB_bFnuAe{Czz@",
    "speed_limit": 100,
    "way_id": 127580804
  }
}
```

image

## The motivation behind this change

Basically, this PR targets two use cases:
- To have an ability to draw a correct speed limit in the nav app - as the `trip` proto now contains such info, an app can detect that conditional speed limit should be displayed instead of the default one.
- To have an ability to debug such cases - `/locate` response will show such conditional speed limits in a human readable format if any

## Technical implementation motivation

- Why not using `AccessRestriction` - restrictions are heavily used in the graph expansion and I doubt that one more type of restriction with one more enum entry will not impact core Valhalla algorithms, which I would like to avoid.
- Why using `EdgeInfo` - `EdgeInfo` already stores `speed_limit`, so related info looks suitable here. Also, we pay only for ways/edges that have such conditional speed limits, which are quite few.
- Why store them in tagged values - tagged values already support multiple values, so we pay one extra bytes per one conditional speed limit, resulting in total 5 bytes (speed is stored in the `TimeDomain` sparse bits) per conditional speed limit which are extremely sparse.
- Conditional speed limits are stored unsorted as I don't know how sorting them will benefit us.
- Return a string (like "19:00 - 6:00") from `/locate` and structured parsed value from `/route` - json format is extremely unfriendly to big structures, so for debugging purpose one single string is much more readable that 12 separate fields. At the same time, if downstream users want to use this info (like drawing a speed limit icon in the corner of the screen) they will benefit from already parsed format that is much much easier to use.
For example:
```python
def condition_is_active(condition):
    dt = datetime.datetime.now()
    if condition.dow_mask == 0 or not (condition.dow_mask & (1 << ((dt.weekday() + 1) % 7))):
      return False

    # todo: handle other stuff
    return True
```

## Tasklist

 - [x] Add tests

## Requirements / Relations


proto/trip.proto
```diff
@@ -119,6 +119,25 @@ message TripLeg {
...
+    // uint64_t begin_mins_ : 6;    // begin minutes, 0 if not set
+    // uint64_t begin_month_ : 4;   // begin month, from 1 to 12, 0 if not set
+    // uint64_t begin_day_dow_ : 5; // begin day of month or nth dow, i.e. 1st Sunday
+    // uint64_t begin_week_ : 3;    // which week does this start, i.e. 1st week in Oct
+    // uint64_t end_hrs_ : 5;       // end hours, 0 if not set
+    // uint64_t end_mins_ : 6;      // end minutes, 0 if not set
+    // uint64_t end_month_ : 4;     // end month, from 1 to 12, 0 if not set
+    // uint64_t end_day_dow_ : 5;   // end day of month or nth dow, i.e. last Sunday
+    // uint64_t end_week_ : 3;      // which week does this end, i.e. last week in Oct
+    uint64 condition = 2;
```
kinkard, line 138: The string format is really strange and really hard to parse, so downstream users will benefit from the bitfield format that then can be used on devices to determine which speed limit to show.
But I'm open for ideas how to do it better.

src/mjolnir/timeparsing.cc
```diff
@@ -163,84 +237,44 @@ std::vector<uint64_t> get_time_range(const std::string& str) {
...
+
+              // Feb 2-14 => Feb#2-Feb#14
+              if (RegexFound(condition, kRangeWithinMonth.first)) {
+                condition = FormatCondition(condition, kRangeWithinMonth);
+              } else {
+
+                // Nov - Mar => Nov-Mar
+                if (RegexFound(condition, kMonthRange.first)) {
+                  condition = FormatCondition(condition, kMonthRange);
+                }
```
kinkard, line 277: This condition is heavily used in Lithuania to annotate 110 kph speed limit during winter (130 otherwise) -https://en.wikipedia.org/wiki/Speed_limits_in_Lithuania
For specific examples see any A1 way in this relation - https://www.openstreetmap.org/relation/2895620

valhalla/baldr/conditional_speed_limit.h
```diff
@@ -0,0 +1,64 @@
...
+  uint64_t begin_day_dow_ : 5; // begin day of month or nth dow, i.e. 1st Sunday
+  uint64_t begin_week_ : 3;    // which week does this start, i.e. 1st week in Oct
+  uint64_t end_hrs_ : 5;       // end hours
+  uint64_t end_mins_ : 6;      // end minutes
+  uint64_t end_month_ : 4;     // end month
+  uint64_t end_day_dow_ : 5;   // end day of month or nth dow, i.e. last Sunday
+  uint64_t end_week_ : 3;      // which week does this end, i.e. last week in Oct
+  uint64_t spare_ : 2;
+
+  uint64_t speed_limit_ : 8; // speed limit in KPH
```
kinkard, line 58: Basically `TimeDomain::spare_` are used here to store the speed limit. Alternatively the bit mask can be used to write speed limit on top of the `TimeDomain` value with cost of the implicit dependency if one day someone decides to use `TimeDomain::spare_` for something, so I decided to go with copy&paste here.

src/baldr/edgeinfo.cc
```diff
@@ -82,6 +90,13 @@ EdgeInfo::EdgeInfo(char* ptr, const char* names_list, const size_t names_list_le
...
+  if (ei_.had_conditional_speed_limits) {
+    conditional_speed_limits_count_ = static_cast<uint8_t>(*ptr);
+    ptr += sizeof(uint8_t);
+
+    conditional_speed_limits_ = reinterpret_cast<ConditionalSpeedLimit*>(ptr);
+    ptr += conditional_speed_limits_count_ * sizeof(ConditionalSpeedLimit);
+  }
 
   // Set encoded elevation pointer
   encoded_elevation_ = reinterpret_cast<int8_t*>(ptr);
```
kinkard, line 102: Unfortunately there was not enough space to put count into the header, just a single bit, so the size goes here.
I put `conditional_speed_limits_` **before** elevation as its size is unknown in the `EdgeInfo` constructor and it was quite hard to pass it everywhere.

valhalla/baldr/edgeinfo.h
```diff
@@ -290,12 +297,12 @@ class EdgeInfo {
...
-    uint32_t extended_wayid1_ : 8;     // Next next byte of the way id
-    uint32_t extended_wayid_size_ : 2; // How many more bytes the way id is stored in
-    uint32_t has_elevation_ : 1;       // Does the edgeinfo have elevation?
-    uint32_t spare0_ : 1;              // not used
+    uint32_t name_count_ : 4;                  // How many name infos we expect
+    uint32_t encoded_shape_size_ : 16;         // How many bytes long the encoded shape is
+    uint32_t extended_wayid1_ : 8;             // Next next byte of the way id
+    uint32_t extended_wayid_size_ : 2;         // How many more bytes the way id is stored in
+    uint32_t has_elevation_ : 1;               // Does the edgeinfo have elevation?
+    uint32_t had_conditional_speed_limits : 1; // Does the edgeifo have conditional speed limits?
```
kinkard, line 305: I'm really not sure about consuming the last bit of the `EdgeInfo` header for `had_conditional_speed_limits` because of the backward-forward compatibility reasons.
Alternatively, I use this bit as a marker for the "second header" and store required info there.

kinkard:
Moved `(Nov - Mar)` condition parsing fix into a separate PR - https://github.com/valhalla/valhalla/pull/4857

proto/trip.proto
kinkard: Okay, surprisingly, it is extremely hard to use it in a languages that do not support bitfields (like Python).

kevinkreiser: yeah i had originally suggested a bitfield for this (or one of the other restriction formats) to @gknisely but he was like "this is a massive pain im not doing it" hahaha. and yeah for python you'd have to do the whole `ctype` struct to get a bitfield to work. it is quite verbose without the bitfield (ie a json field for each integer) but then again the locate output only comes with all this information when you specify `verbose=true` so maybe its ok?

kevinkreiser: oh and if you are talking about the bitfield stored in a single integer in proto and then consuming the proto format as output from the service then i think sadly its better to force the requester to figure out the bitfield than to bloat the proto with all the various fields

kinkard: Well, I've played a bit with bits unpacking like this
```python
import ctypes
class TimeDomain(ctypes.Structure):
    _fields_ = [
        ("day_dow_type", ctypes.c_uint64, 1),
        ("dow", ctypes.c_uint64, 7),
        ("begin_hrs", ctypes.c_uint64, 5),
        ("begin_mins", ctypes.c_uint64, 6),
        ("begin_month", ctypes.c_uint64, 4),
        ("begin_day_dow", ctypes.c_uint64, 5),
        ("begin_week", ctypes.c_uint64, 3),
        ("end_hrs", ctypes.c_uint64, 5),
        ("end_mins", ctypes.c_uint64, 6),
        ("end_month", ctypes.c_uint64, 4),
        ("end_day_dow", ctypes.c_uint64, 5),
        ("end_week", ctypes.c_uint64, 3),
        ("spare", ctypes.c_uint64, 10),
    ]
td = TimeDomain.from_buffer_copy(ctypes.c_uint64(12884906752))
```

and the biggest problem with it is that you need additional tests to cover how value is unpacked. 
In the de685b9601afdffe0b9b9b61507a74f98dea3c0e I've added a `message TimeDomain` and it doesn't look that bad, while the downstream usage seems much more convenient.

valhalla/baldr/timedomain.h
```diff
@@ -310,19 +312,20 @@ union TimeDomain {
...
+    uint64_t begin_hrs : 5;     // begin hours, 0 if not set
+    uint64_t begin_mins : 6;    // begin minutes, 0 if not set
+    uint64_t begin_month : 4;   // begin month, from 1 (January) to 12 (December), 0 if not set
+    uint64_t begin_day_dow : 5; // begin day of month or nth dow, i.e. 1st Sunday
+    uint64_t begin_week : 3;    // which week does this start, i.e. 1st week in Oct
+    uint64_t end_hrs : 5;       // end hours, 0 if not set
+    uint64_t end_mins : 6;      // end minutes, 0 if not set
+    uint64_t end_month : 4;     // end month, from 1 (January) to 12 (December), 0 if not set
+    uint64_t end_day_dow : 5;   // end day of month or nth dow, i.e. last Sunday
+    uint64_t end_week : 3;      // which week does this end, i.e. last week in Oct
```
kinkard, line 328: just a doc change, as I anyway came to better (imho) wording for some fields

proto/trip.proto
gknisely: > yeah i had originally suggested a bitfield for this (or one of the other restriction formats) to @gknisely but he was like "this is a massive pain im not doing it" hahaha. and yeah for python you'd have to do the whole `ctype` struct to get a bitfield to work. it is quite verbose without the bitfield (ie a json field for each integer) but then again the locate output only comes with all this information when you specify `verbose=true` so maybe its ok?

Not a pain...just no time

kevinkreiser: sorry yeah thats all i meant

valhalla/baldr/conditional_speed_limit.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
+
+#include <cstdint>
+#include <string>
+
+namespace valhalla {
+namespace baldr {
+
+struct ConditionalSpeedLimit {
```
kevinkreiser, line 9: any chance we can simplify this. could you union all the bits that make up the condtion with some member maybe call it `condition_` and then remove the private section entirely. then you should be able to initialize it without a constructor using braces i think like `ConditionalSpeedLimit{speed, condition}`. where we can i just like to have a simple as possible structures. (note i didnt tryy this so there be some gotcha that makes this not work, apologies if so)

valhalla/baldr/conditional_speed_limit.h
```diff
@@ -0,0 +1,64 @@
...
+  uint64_t begin_day_dow_ : 5; // begin day of month or nth dow, i.e. 1st Sunday
+  uint64_t begin_week_ : 3;    // which week does this start, i.e. 1st week in Oct
+  uint64_t end_hrs_ : 5;       // end hours, 0 if not set
+  uint64_t end_mins_ : 6;      // end minutes, 0 if not set
+  uint64_t end_month_ : 4;     // end month, from 1 (January) to 12 (December), 0 if not set
+  uint64_t end_day_dow_ : 5;   // end day of month or nth dow, i.e. last Sunday
+  uint64_t end_week_ : 3;      // which week does this end, i.e. last week in Oct
+  uint64_t spare_ : 2;
+
+  uint64_t speed_limit_ : 8; // speed limit in KPH
```
kevinkreiser, line 58: i guess this is 8 bytes wide just because the compiler will pad to that anyway so may as well do it explicitly?

valhalla/baldr/conditional_speed_limit.h
```diff
@@ -0,0 +1,64 @@
...
+  uint64_t begin_mins_ : 6;    // begin minutes, 0 if not set
+  uint64_t begin_month_ : 4;   // begin month, from 1 (January) to 12 (December), 0 if not set
+  uint64_t begin_day_dow_ : 5; // begin day of month or nth dow, i.e. 1st Sunday
+  uint64_t begin_week_ : 3;    // which week does this start, i.e. 1st week in Oct
+  uint64_t end_hrs_ : 5;       // end hours, 0 if not set
+  uint64_t end_mins_ : 6;      // end minutes, 0 if not set
+  uint64_t end_month_ : 4;     // end month, from 1 (January) to 12 (December), 0 if not set
+  uint64_t end_day_dow_ : 5;   // end day of month or nth dow, i.e. last Sunday
+  uint64_t end_week_ : 3;      // which week does this end, i.e. last week in Oct
+  uint64_t spare_ : 2;
```
kevinkreiser, line 56: i konw that it technically wouldnt be a POD, but can we just use `TimeDomain` here instead of copying? i think the memory layout will still be the same and that the `sequence` header wont complain but maybe im remembering that wrong

valhalla/baldr/conditional_speed_limit.h
kevinkreiser: oh derp i didnt see the `:8`

kevinkreiser:
we're working our way through the review here, bear with us

valhalla/baldr/conditional_speed_limit.h
kinkard: Initially I've started from the following fragile bit magic, that used `TimeDomain` and utilised the `TimeDomain::spare` bits to store the speed limit. But this looks too fragile

```cpp
struct ConditionalSpeedLimit {
  ConditionalSpeedLimit(uint8_t speed, uint64_t time_domain) {
    encoded_ = time_domain | (static_cast<uint64_t>(speed) << (64 - 8));
  }

  uint32_t speed_limit() const {
    return encoded_ >> (64 - 8);
  }
  uint64_t condition() const {
    return encoded_ & ((-1) >> 8);
  }

  // mix of bits from TimeDomain and speed limit in the higher 8 bits
  uint64_t encoded_;
};
```

valhalla/baldr/conditional_speed_limit.h
kinkard: Well, I can try this
```cpp
struct ConditionalSpeedLimit {
  uint32_t speed_limit() const {
    return speed_limit_;
  }

  uint64_t condition() const {
    return condition_;
  }


  uint64_t condition_: 56;   // TimeDomain meaningful bits
  uint64_t speed_limit_: 8;  // speed limit in KPH
};
```

the all thing will become much smaller, but I'm not sure about this implicit dependency if one day someone decides to use `TimeDomain::spare` bits for something.

kinkard: Actually, I like how it looks like

src/baldr/conditional_speed_limit.cc
```diff
@@ -0,0 +1,180 @@
...
+    default:
+      return "";
+  }
+}
+} // namespace
+
+namespace valhalla {
+namespace baldr {
+
+std::string ConditionalSpeedLimit::condition_str() const {
```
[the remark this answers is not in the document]
kinkard: Addressed in ffcf854278e9a902d02711cc2fc5a7c5f907822f

valhalla/baldr/conditional_speed_limit.h
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Provides the `TimeDomain` value of the condition, that defines when this speed limit is applied.
+   * @return `TimeDomain` value of the condition.
+   */
+  uint64_t condition() const {
+    return condition_;
+  }
+
+  uint64_t condition_ : 56;  // TimeDomain meaningful bits
```
kevinkreiser, line 26: can we rename this to `time_domain_` so its clear that that is what this is? also why it is 56 bits, when the `DateRange` object is 54 bits. i see you only use 8 for the speed so you dont need all 10 spare but still dont you want the condition thing to match exactly the timedomain union? i wonder if it would be easier if we use a union here to get at a speed limit that overrides the spare:

```
union conditional_speed_t {
   TimeDomain td_;
   struct speed_limit {
     uint64_t padding_ : 54; // this would pad over the timedomain part
     uint64_t speed_ : 8;
     uint64_t spare_ : 2;
   };
};
```

then we could maybe just throw the to_string function for the condition into the timedomain header/impl and keep this thing tiny?

valhalla/baldr/datetime.h
```diff
@@ -212,6 +212,38 @@ bool is_conditional_active(const bool type,
...
+ * @param   begin_month   begin month
+ * @param   begin_day_dow if kNthDow, then which dow to start the restriction.
+ *                        if kYMD then it is the day of the month
+ * @param   end_week      only set for kNthDow.  which week in the month
+ * @param   end_month     end month
+ * @param   end_day_dow   if kNthDow, then which dow to end the restriction.
+ *                        if kYMD then it is the day of the month
+ * @return String with time range that represents the condition
+ */
+std::string conditional_to_str(const bool type,
```
kevinkreiser, line 234: can we not just pass the TimeDomain object into this and let it do its thing internally, why do we need to hav all the fields separate, is there use case where we dont have a timedomain already parsed? i see you are using it for a test, maybe you can just do an equality check between the proto fields and the real TimeDomain fields in the test rather than to_stringing them

valhalla/baldr/edgeinfo.h
kinkard:
```cpp
mutable std::multimap<TaggedValue, std::string> tag_cache_;
```
Well, the only thing I don't like is putting u64 values into a string. Well, technically it is not a problem to put bytes and then do `std::reinterpret_cast<uint64_t*>(tag.second.data())`, it just looks... well, it looks strange.

valhalla/baldr/datetime.h
kinkard: > can we not just pass the `TimeDomain` object into this and let it do its thing internally

I just wanted to keep it in par with `is_conditional_active()`, that takes the condition as separate fields.

> maybe you can just do an equality check between the proto fields and the real TimeDomain fields

Actually, this is what I do in tests, `conditional_to_str()` only used for pretty print if test failed.  And I used it in tests only because it already exists for `/locate` json output.

valhalla/baldr/conditional_speed_limit.h
kinkard: Addressed in 23799605a30ab0ac75d7e4f49f05e03f4b1b2860

valhalla/baldr/datetime.h
kinkard: Addressed in 23799605a30ab0ac75d7e4f49f05e03f4b1b2860

valhalla/baldr/edgeinfo.h
kinkard: Alternatively, I can remove all the stuff around parsing, encoding and so on, and just keep conditional speed limits as an original string inside the tagged values (similar to how Valhalla handles linguistics, ), that Valhalla will simply return. @kevinkreiser wdyt about this approach?

kevinkreiser: true, that would be the simplest option! i might still suggest you normalize the values by sending it through the parsing and to_string methods you have. on the flip side of that it means that we cant store every single strange permutation people put in the data (but that might be a good thing anyway?)

i guess id also note that its likely a good deal cheaper (maybe half?) to store the fixed size `DateRange` struct than the strings, which does very much appeal to me. in the end though i dont feel too strongly about either approach. id say i have a slight preference for the above, keep the parse and to_string stuff, round trip the raw values to normalize them and store the string in the key value. whats your preference?

chrstnbwnkl: I think an upside of the normalization would be that it tells the user what sense valhalla actually made of a condition. E.g. if there's a "PH" value in that string (which means the speed only applies on public holidays, for which there's no support at the moment), and that string just goes straight into the `/locate` output I see people could be more likely to complain why it's not being applied on christmas eve :smile:

kinkard: > whats your preference?

I would like to leave all the of parsing the suprisingly complex [opening hours](https://wiki.openstreetmap.org/wiki/Key:opening_hours) format at the Valhalla side and return somehow structured output, so downstream users can write logic like `condition.dow_mask & (1 << today.get_weekday())`.

Another argument for keeping fixed size parsed value is that it makes possible to use it in the costing (or where else speeds are used) one day.

kevinkreiser: > possible to use it in the costing 

this was the argument i was going to make but decided to not complicate the discussion, im glad you were thinking it too :smile:

kinkard: Talking about the costing - I can't find any `EdgeInfo` usage in the costing, so maybe some another place is more suitable for storing conditional speed limits?

kevinkreiser: this is why i didnt want to complicate the discussion :smile: ill try to do so succinctly.

we purposely dont give edgeinfo to the costing because its expensive to pull out of the tile relative to the other items. its meant for things that are sparsely populated, conditional, or just vanity metrics not something we cost on in the tight loop of the algorithm.

that said you can get it because you get the tile and the edge. if you have the edge and you have the tile then you have a way to get the edgeinfo.

at the moment speed limit isnt that interesting to costing because its usually baked into the `directededge->speed` but in this case, if the algorithm was trackign time and the edge had conditional speed limits and the speed calculation needed to fall back to the speed limit (because no real time, no historical, etc), then it would make sense to pull out the conditions and check if one overlaps this time. as you can see it would be quite a complication (a rare one) but indeed, as you stated it would be better to have it in the parsed struct if we plan on doing that.

so how about we keep the fixed size struct, we throw it in with the tagged values and we save using it in GetSpeed for a rainy day :wink:

kinkard: Thanks for the great explanation! I like the approach you suggested, I just need to convince myself that it is a good idea to do such hack:
```cpp
// somewhere at writing
ConditionalSpeedLimit limit;
std::string value(reinterpret_cast<char*>(&limit), sizeof(limit));

// and then at reading
const std::string limit_str = edge_info.GetTags()[kConditionalSpeedLimit];
const uint8_t count = limit_str.front();
for (size_t pos = 1; pos < 1 + count * sizeof(ConditionalSpeedLimit); pos += sizeof(ConditionalSpeedLimit)) {
  const ConditionalSpeedLimit limit = *reinterpret_cast<ConditionalSpeedLimit*>(&limit_str.data() + pos);
  // use/check limit...
}
```

kinkard: All names (tagged values) are stored using null terminator, which should be hacked around as zero bytes are possible in the `ConditionalSpeedLimit`.

src/baldr/edgeinfo.cc
```diff
@@ -9,10 +8,7 @@ namespace {
 
 // should return true if not equal to TaggedValue::kLinguistic
 bool IsNonLiguisticTagValue(char ch) {
-  static const std::unordered_set<TaggedValue> kNameTags =
-      {TaggedValue::kBridge, TaggedValue::kTunnel,   TaggedValue::kBssInfo, TaggedValue::kLayer,
-       TaggedValue::kLevel,  TaggedValue::kLevelRef, TaggedValue::kLandmark};
-  return kNameTags.count(static_cast<TaggedValue>(static_cast<uint8_t>(ch))) > 0;
+  return static_cast<TaggedValue>(ch) != TaggedValue::kLinguistic;
 }
```
kinkard, line 12: @gknisely Maybe I'm messing up with some lost sacred knowledge here, so please correct me if I'm doing a mistake.

valhalla/baldr/edgeinfo.h
kinkard: As far as I got, it is not a problem to have strings with `\0` somewhere inside, for example we don't dump all tagged values into a json. So let's try - e79f54d39a65720e2432c155e8f1a9f4ce5e9f5c

src/mjolnir/osmdata.cc
```diff
@@ -308,6 +320,31 @@ bool write_linguistic(const std::string& filename, const LinguisticMultiMap& lin
...
+  if (!file.is_open()) {
+    LOG_ERROR("write_linguistic failed to open output file: " + filename);
+    return false;
+  }
+
+  // Convert the multi map into a flat representation
+  std::vector<TempConditionalSpeedLimit> flat;
+  flat.reserve(speed_map.size());
+  for (auto it = speed_map.cbegin(); it != speed_map.cend(); ++it) {
+    flat.emplace_back(it->first, it->second);
```
chrstnbwnkl, line 336: Can you delete `TmpConditionalSpeedLimit` and just write to file in the loop?

src/mjolnir/osmdata.cc
```diff
@@ -525,6 +562,29 @@ bool read_linguistic(const std::string& filename, LinguisticMultiMap& ling_map)
...
+  std::ifstream file(filename, std::ios::in | std::ios::binary);
+  if (!file.is_open()) {
+    LOG_ERROR("read_conditional_speed_limits failed to open input file: " + filename);
+    return false;
+  }
+
+  // Read the count and then the temporary access restriction list
+  uint32_t count = 0;
+  file.read(reinterpret_cast<char*>(&count), sizeof(uint32_t));
+  std::vector<TempConditionalSpeedLimit> flat(count);
```
chrstnbwnkl, line 577: Why not skip making the `flat` vector and emplace into the map right away with a loop?

test/gurka/test_elevation.cc
```diff
@@ -230,4 +237,28 @@ TEST(Standalone, ElevationCompareToSkadi) {
...
+    EXPECT_EQ(condition.begin_mins(), 0);
+    EXPECT_EQ(condition.begin_month(), 0);
+    EXPECT_EQ(condition.begin_day_dow(), 0);
+    EXPECT_EQ(condition.begin_week(), 0);
+    EXPECT_EQ(condition.end_hrs(), 6);
+    EXPECT_EQ(condition.end_mins(), 0);
+    EXPECT_EQ(condition.end_month(), 0);
+    EXPECT_EQ(condition.end_day_dow(), 0);
+    EXPECT_EQ(condition.end_week(), 0);
+  }
```
chrstnbwnkl, line 263: Why add this in the elevation test?

chrstnbwnkl (changes requested):
Looks good generally, just a couple of questions!

src/baldr/edgeinfo.cc
```diff
@@ -445,6 +456,17 @@ json::MapPtr EdgeInfo::json() const {
     edge_info->emplace("speed_limit", static_cast<uint64_t>(speed_limit()));
   }
 
+  const auto cond_limits_range = GetTags().equal_range(TaggedValue::kConditionalSpeedLimits);
+  if (cond_limits_range.first != cond_limits_range.second) {
```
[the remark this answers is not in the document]
kinkard: Well, I doubt that I know a proper formatting for all tagged values, especially `kLandmark` and `BssInfo`. Probably worth to add in a separate PR

src/baldr/timedomain.cc
```diff
@@ -0,0 +1,197 @@
...
+  // handle range ending after loop
+  if (prev && preprev) {
+    // we passed the range end - print the previous day
+    ss << '-' << days[6];
+  }
+}
+
+std::string month_name(valhalla::baldr::MONTH month) {
+  using valhalla::baldr::MONTH;
+  switch (month) {
```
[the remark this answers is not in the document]
kinkard: There is no point in the unordered map as we do not resolve strings into enum, only enum into a string. As an option, I can do something like this
```cpp
std::string month_name(valhalla::baldr::MONTH month) {
  const std::string months[] = {"Jan", "Feb", /*...*/};
  return months[static_cast<size_t>(month);
}
```

src/mjolnir/osmdata.cc
kinkard: Well, all items are stored via temp flat structure and via flat vector. I definitely don't want to refactor whole `osmdata.cc` in this PR.

src/mjolnir/osmdata.cc
kinkard: Well, all osmdata items are stored via temp flat structure and via flat vector. I definitely don't want to refactor whole osmdata.cc in this PR. Probably there were a reasons for such approach that I don't know

test/gurka/test_elevation.cc
kinkard: This test helped me to find and fix a bug in the initial implementation. Removed in 58bf2016c51e512002253d39f0ac55a104296a23

src/mjolnir/osmdata.cc
```diff
@@ -308,6 +320,31 @@ bool write_linguistic(const std::string& filename, const LinguisticMultiMap& lin
...
+  for (auto it = speed_map.cbegin(); it != speed_map.cend(); ++it) {
+    flat.emplace_back(it->first, it->second);
+  }
+
+  // Write the count and then the via ids
+  uint32_t sz = flat.size();
+  file.write(reinterpret_cast<const char*>(&sz), sizeof(uint32_t));
+  file.write(reinterpret_cast<const char*>(flat.data()),
+             flat.size() * sizeof(TempConditionalSpeedLimit));
+  file.close();
```
kevinkreiser, line 344: sure i see others have followed this pattern but we can do a little better. you can avoid having to copy all the data before writing it and can even avoid having the temporary structure both by simply writing it directly to the file:

```suggestion
  // Convert the multi map into a flat representation
  uint32_t sz = speed_map.size();
  file.write(reinterpret_cast<const char*>(&sz), sizeof(uint32_t));
  for(const auto& sp : speed_map) {
    file.write(reinterpret_cast<const char*>(&sp.first), sizeof(sp.first));
    file.write(reinterpret_cast<const char*>(&sp.second), sizeof(sp.second));
  }
  file.close();
```

src/mjolnir/osmdata.cc
```diff
@@ -525,6 +562,29 @@ bool read_linguistic(const std::string& filename, LinguisticMultiMap& ling_map)
...
+  file.read(reinterpret_cast<char*>(&count), sizeof(uint32_t));
+  std::vector<TempConditionalSpeedLimit> flat(count);
+  file.read(reinterpret_cast<char*>(flat.data()), count * sizeof(TempConditionalSpeedLimit));
+  file.close();
+
+  // Iterate through the temporary access restriction list and add to the restriction multi-map
+  for (const auto& l : flat) {
+    speed_map.emplace(l.way_id, l.conditional_speed_limit);
+  }
+  return true;
```
kevinkreiser, line 585:
```suggestion
  // Read the count and fill out our multimap
  uint32_t count = 0;
  file.read(reinterpret_cast<char*>(&count), sizeof(uint32_t));
  while(count--) {
    std::pair<uint64_t, ConditionalSpeedLimit> sp;
    file.read(reinterpret_cast<char*>(&sp), sizeof(sp));
    speed_map.emplace(std::move(sp));
  }  
  file.close();
  return true;
```

kevinkreiser: its not super important but its just a bit less copying and less code to go directly from file to the multimap. you certainly dont have to refactor all the other methods that dont do this. its just a suggestion to avoid some copying and avoid the temporary structure. i guess one option would be to just use the pair isntead of the temporary structure and keep the copies to the vector. or dont do either hahaha

src/baldr/timedomain.cc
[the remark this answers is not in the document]
kevinkreiser: a vector is also a fine solution, should be `static` though

src/mjolnir/osmdata.cc
kinkard: Added in b08327ce2b9c8cbbbf0ac0bd8826a39f096baa17

src/mjolnir/osmdata.cc
kinkard: Added in b08327ce2b9c8cbbbf0ac0bd8826a39f096baa17

src/mjolnir/osmdata.cc
kinkard: Addressed in b08327ce2b9c8cbbbf0ac0bd8826a39f096baa17

src/mjolnir/osmdata.cc
kinkard: Addressed in b08327ce2b9c8cbbbf0ac0bd8826a39f096baa17

src/baldr/timedomain.cc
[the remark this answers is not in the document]
kinkard: Addressed in 811a140eaefee5a2ec26b99ff63b803fed3908b9

src/baldr/edgeinfo.cc
kinkard: All tests are green. I just got confused by the inverse logic

src/baldr/edgeinfo.cc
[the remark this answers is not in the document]
kevinkreiser: we really just wanted a switch case with only this one filled out so that we have a template to fill out the others. you dont have to fill the others out just would be nice to have a switch here so we see there is work to do

test/gurka/test_conditional_speedlimit.cc
```diff
@@ -0,0 +1,313 @@
...
+TEST_F(ConditionalSpeedlimit, RouteApiProto) {
+  const auto result = gurka::do_action(valhalla::Options::route, map, {"D", "Q"}, "auto",
+                                       {{"/date_time/value", "2024-08-15T21:00"}});
+  ASSERT_EQ(result.trip().routes_size(), 1);
+  ASSERT_EQ(result.trip().routes(0).legs_size(), 1);
+  const auto leg = result.trip().routes(0).legs(0);
+  ASSERT_EQ(leg.node_size(), 8);
+
+  // Regular speed limit are not affected, even if date was passed
+  EXPECT_EQ(leg.node(0).edge().speed_limit(), 100); // DC
```
chrstnbwnkl, line 136: Can you add a call to `gurka::raw::assert_path` (or something like that :smile:) to make sure the path is actually as expected?

valhalla/baldr/timedomain.h
```diff
@@ -309,20 +311,27 @@ union TimeDomain {
...
```
kevinkreiser, line 334: thank you for spelling more of this out for our future selves when we need to remember

kevinkreiser (dismissed):
this looks great, there were two little nitpicks but we can merge this when you say you are ready, let us know!

src/baldr/edgeinfo.cc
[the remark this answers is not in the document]
kinkard: Added in dbe27c30b6cb2716a8bcd7223e1d539b3d3deb9f

test/gurka/test_conditional_speedlimit.cc
kinkard: Added in 12118f1fe45fb9913a3fa2197bcd4c829dcb6b66

kinkard:
> this looks great, there were two little nitpicks but we can merge this when you say you are ready, let us know!

Let's merge it if the recent switch-case addition around `EdgeInfo::Tags()` in https://github.com/valhalla/valhalla/pull/4851/commits/dbe27c30b6cb2716a8bcd7223e1d539b3d3deb9f looks ok for you.
I'm thinking that it would be nice to have a structured output from `/locate` output as well, but there is no `pbf` format output for `/locate` (structured in `pbf` and human-readable in `json`), so probably in another PR ;)

src/baldr/edgeinfo.cc
```diff
@@ -445,6 +456,39 @@ json::MapPtr EdgeInfo::json() const {
...
+      }
+      case TaggedValue::kTunnel:
+        break;
+      case TaggedValue::kBridge:
+        break;
+    }
+  }
+  if (conditional_speed_limits) {
+    edge_info->emplace("conditional_speed_limits", std::move(conditional_speed_limits));
+  }
```
kinkard, line 490: Well, probably it would be more aligned with other parts to always send `"conditional_speed_limits"` even if empty, but the idea of adding quite a big string to the response where it would be rarely used hurts me.

kevinkreiser: i prefer the way you did it!

kevinkreiser (approved):

PR #4857, opened 2024-08-18
Fix `(Nov - Mar)` condition parsing

kinkard:
# Issue

This PR itself is a substract from the https://github.com/valhalla/valhalla/pull/4851 into a separate PR that addresses a separate issue.

The `(Nov - Mar)` condition (with spaces) is extremely common in Lithuania. It is used to annotate 110 kph speed limit during winter (130 otherwise) - https://en.wikipedia.org/wiki/Speed_limits_in_Lithuania. Almost all A1 ways (e.g. https://www.openstreetmap.org/way/966218614) have it.

## Tasklist

 - [x] Add tests

## Requirements / Relations

### DIFF
--- a/test/timeparsing.cc
+++ b/test/timeparsing.cc
@@ -460,6 +460,55 @@ TEST(TimeParsing, TestConditionalRestrictions) {
   }
 }
 
+TEST(TimeParsing, TestConditionalMaxspeed) {
+  TryConditionalRestrictions("(19:00-06:00)", 0, 0, 0, {0, 0, 0, 19, 0}, {0, 0, 0, 6, 0});
+  TryConditionalRestrictions("(06:00-18:00)", 0, 0, 0, {0, 0, 0, 6, 0}, {0, 0, 0, 18, 0});
+
+  std::string condition = "(07:00-09:00,13:00-16:00; SH off)";
+  ASSERT_EQ(get_time_range(condition).size(), 2);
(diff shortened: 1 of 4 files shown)


kevinkreiser:
i very much appreciate you breaking this out into a smaller bit that is much easier to review and test

test/timeparsing.cc
```diff
@@ -460,6 +460,78 @@ TEST(TimeParsing, TestConditionalRestrictions) {
...
+    TryConditionalRestrictions(conditions[3], 0, 0, 62, {8, 15, 0, 0, 0}, {12, 31, 0, 0, 0});
+
+    ASSERT_EQ(get_time_range(conditions[4]).size(), 0);
+    ASSERT_EQ(get_time_range(conditions[5]).size(), 0);
+  }
+
+  TryConditionalRestrictions("(Jun 1-Aug 31 00:00-24:00)", 0, 0, 0, {6, 1, 0, 0, 0},
+                             {8, 31, 0, 0, 0});
+
+  // non-standart seasons that are not supported
```
kevinkreiser, line 530: linter wants you to fix this typo
```suggestion
  // non-standard seasons that are not supported
```

kinkard:
The note about perf commit in this PR:

The function `get_time_range` takes 2.4% of the way_callback time and noticeable part of it was spent in regex construction and assigning.

Before:
image

After:
image

Which is nice, taking in account that this PR adds one more regex to check. So more regexes in less time, which is always good

kevinkreiser (approved):

Issue #4878, opened 2024-09-06
Control level in location search 


chrstnbwnkl:
As discussed in #3650, there is currently no way of controlling the desired level for locations, which is necessary for proper indoor navigation. I see two possible ways this could be implemented: 

- analogous to the `preferred_layer` parameter introduced in #3270 
- as a search filter 

The difference lies in how Valhalla handles cases when no locations were found: the search filter is a "hard" filter, while the other option would be "soft" (similar to heading/side of street), meaning if no candidates are found, loki will fall back to previously filtered edges. I'm not sure yet if there is a clearly superior option, but I'm leaning towards the search filter (i.e. hard filter), because falling back to locations on other levels seems of little use to me.

PR #4879, opened 2024-09-06
Support multiple levels and level ranges

chrstnbwnkl:
# Issue

Precursor to #4878. 

Adds proper parsing of integer values stored in the `level` tag, including ranges. This is in accordance with the [osm wiki](https://wiki.openstreetmap.org/wiki/Key:level) and the [Simple Indoor Tagging](https://wiki.openstreetmap.org/wiki/Indoor_Mapping) scheme. 

Also adds suppoprt for multiple, semicolon separated `level_ref` values. 

## Tasklist

 - [x] Add tests

### DIFF
--- a/lua/graph.lua
+++ b/lua/graph.lua
@@ -1870,11 +1870,6 @@ function filter_tags_generic(kv)
   kv["bike_local_ref"] = lref
   kv["bike_network_mask"] = bike_mask
 
-  -- turn semicolon into colon due to challenges to store ";" in string
-  if kv["level"] ~= nil then
-    kv["level"] = kv["level"]:gsub(";", ":")
-  end
-
   -- Explicitly turn off access for construction type. It's done for backward compatibility
   -- of valhalla tiles and valhalla routing. In case we allow non-zero access then older
   -- versions of router will work with new tiles incorrectly. They would start to route
--- a/src/baldr/edgeinfo.cc
+++ b/src/baldr/edgeinfo.cc
@@ -404,11 +404,13 @@ int8_t EdgeInfo::layer() const {
   return static_cast<int8_t>(value.front());
 }
 
-std::vector<std::string> EdgeInfo::level() const {
+std::vector<int16_t> EdgeInfo::level() const {
   const auto& tags = GetTags();
-  std::vector<std::string> values;
+  std::vector<int16_t> values;
   for (auto [itr, range_end] = tags.equal_range(TaggedValue::kLevel); itr != range_end; ++itr) {
-    values.emplace_back(itr->second);
+    try {
+      values.emplace_back(static_cast<uint16_t>(std::stoi(itr->second)));
+    } catch (...) { LOG_ERROR("Unable to parse numerical level value " + itr->second); }
   }
   return values;
 }
--- a/src/baldr/edgeinfo.cc
+++ b/src/baldr/edgeinfo.cc
@@ -404,24 +404,22 @@ int8_t EdgeInfo::layer() const {
   return static_cast<int8_t>(value.front());
 }
 
-std::string EdgeInfo::level() const {
+std::vector<std::string> EdgeInfo::level() const {
   const auto& tags = GetTags();
-  auto itr = tags.find(TaggedValue::kLevel);
-  if (itr == tags.end()) {
-    return "";
+  std::vector<std::string> values;
+  for (auto [itr, range_end] = tags.equal_range(TaggedValue::kLevel); itr != range_end; ++itr) {
+    values.emplace_back(itr->second);
   }
-  const std::string& value = itr->second;
-  return value;
+  return values;
 }
 
-std::string EdgeInfo::level_ref() const {
+std::vector<std::string> EdgeInfo::level_ref() const {
   const auto& tags = GetTags();
-  auto itr = tags.find(TaggedValue::kLevelRef);
-  if (itr == tags.end()) {
-    return "";
+  std::vector<std::string> values;
+  for (auto [itr, range_end] = tags.equal_range(TaggedValue::kLevelRef); itr != range_end; ++itr) {
+    values.emplace_back(itr->second);
   }
-  const std::string& value = itr->second;
-  return value;
+  return values;
 }
 
 json::MapPtr EdgeInfo::json() const {
--- a/src/mjolnir/osmway.cc
+++ b/src/mjolnir/osmway.cc
@@ -1069,7 +1069,33 @@ void OSMWay::GetTaggedValues(const UniqueNames& name_offset_map,
     // level
     auto tokens = GetTagTokens(name_offset_map.name(level_index_));
     for (const auto& t : tokens) {
-      names.emplace_back(encode_tag(TaggedValue::kLevel) + t);
+
+      // get ranged values
+      const char dash = '-';
+      auto dash_pos = t.find(dash);
+
+      // if there is a dash and its not at the first position,
+      // we have a range
+      if (dash_pos != std::string::npos && dash_pos != 0) {
+        std::vector<std::string> range_start_end;
+        boost::algorithm::split(
+            range_start_end, t, [dash](const char c) { return c == dash; },
+            boost::algorithm::token_compress_on);
+        if (range_start_end.size() == 2) {
+          try {
+            auto start = std::stoi(range_start_end[0]);
+            auto end = std::stoi(range_start_end[1]);
+
+            for (; start <= end; ++start) {
+              names.emplace_back(encode_tag(TaggedValue::kLevel) + std::to_string(start));
+            }
+          } catch (...) { LOG_ERROR("Unable to process level range: " + t); }
+        } else {
+          LOG_ERROR("Unable to process level range: " + t);
+        }
+      } else { // no range
+        names.emplace_back(encode_tag(TaggedValue::kLevel) + t);
+      }
     }
   }
   if (level_ref_index_ != 0) {
--- a/src/odin/enhancedtrippath.cc
+++ b/src/odin/enhancedtrippath.cc
@@ -525,17 +525,19 @@ std::vector<std::pair<std::string, bool>> EnhancedTripLeg_Edge::GetNameList() co
 
 std::vector<std::string> EnhancedTripLeg_Edge::GetLevelRef() const {
   std::vector<std::string> level_refs;
+  std::vector<std::string> levels;
+
+  // try to get level_refs, else create some from levels as fallback
   if (!tagged_value().empty()) {
     for (int t = 0; t < tagged_value().size(); ++t) {
       if (tagged_value().Get(t).type() == TaggedValue_Type_kLevelRef) {
         level_refs.emplace_back(tagged_value().Get(t).value());
-        break;
       } else if (tagged_value().Get(t).type() == TaggedValue_Type_kLevel) {
-        level_refs.emplace_back("Level " + tagged_value().Get(t).value());
+        levels.emplace_back("Level " + tagged_value().Get(t).value());
       }
     }
   }
-  return level_refs;
+  return level_refs.empty() ? levels : level_refs;
 }
 
 float EnhancedTripLeg_Edge::GetLength(const Options::Units& units) {
--- a/src/odin/enhancedtrippath.cc
+++ b/src/odin/enhancedtrippath.cc
@@ -523,19 +523,19 @@ std::vector<std::pair<std::string, bool>> EnhancedTripLeg_Edge::GetNameList() co
   return name_list;
 }
 
-std::string EnhancedTripLeg_Edge::GetLevelRef() const {
-  std::string level_ref;
+std::vector<std::string> EnhancedTripLeg_Edge::GetLevelRef() const {
+  std::vector<std::string> level_refs;
   if (!tagged_value().empty()) {
     for (int t = 0; t < tagged_value().size(); ++t) {
       if (tagged_value().Get(t).type() == TaggedValue_Type_kLevelRef) {
-        level_ref = tagged_value().Get(t).value();
+        level_refs.emplace_back(tagged_value().Get(t).value());
         break;
       } else if (tagged_value().Get(t).type() == TaggedValue_Type_kLevel) {
-        level_ref = "Level " + tagged_value().Get(t).value();
+        level_refs.emplace_back("Level " + tagged_value().Get(t).value());
       }
     }
   }
-  return level_ref;
+  return level_refs;
 }
 
 float EnhancedTripLeg_Edge::GetLength(const Options::Units& units) {
--- a/src/odin/maneuversbuilder.cc
+++ b/src/odin/maneuversbuilder.cc
@@ -1157,7 +1157,11 @@ void ManeuversBuilder::InitializeManeuver(Maneuver& maneuver, int node_index) {
 
   // Set the end level ref
   if (curr_edge && !curr_edge->GetLevelRef().empty()) {
-    maneuver.set_end_level_ref(curr_edge->GetLevelRef());
+    if (curr_edge->GetLevelRef().size() > 1) {
+      maneuver.set_end_level_ref("");
+    } else {
+      maneuver.set_end_level_ref(curr_edge->GetLevelRef()[0]);
+    }
   }
 
   // Elevator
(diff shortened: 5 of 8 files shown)


lua/graph.lua
```diff
@@ -1870,11 +1870,6 @@ function filter_tags_generic(kv)
   kv["bike_local_ref"] = lref
   kv["bike_network_mask"] = bike_mask
 
-  -- turn semicolon into colon due to challenges to store ";" in string
```
chrstnbwnkl, line 1873: I suppose this was added to avoid splitting the value string?

src/mjolnir/osmway.cc
```diff
@@ -1069,7 +1069,33 @@ void OSMWay::GetTaggedValues(const UniqueNames& name_offset_map,
     // level
     auto tokens = GetTagTokens(name_offset_map.name(level_index_));
     for (const auto& t : tokens) {
-      names.emplace_back(encode_tag(TaggedValue::kLevel) + t);
+
```
chrstnbwnkl, line 1072: This is the actual parsing of the initial string: 
- splits it by semicolon
- for each token, check if the token denotes a range
- if it is a range, add a tagged value for each value in the range
- if it is not a range, simply store the value

Another way would be to push this out to edgeinfo and just store the string as a single tagged value, but I suppose this is sparse data (and most edges only connect 2 floors anyways).

src/baldr/edgeinfo.cc
```diff
@@ -404,24 +404,24 @@ int8_t EdgeInfo::layer() const {
...
+    try {
+      values.emplace_back(static_cast<uint16_t>(std::stoi(itr->second)));
+    } catch (...) { LOG_ERROR("Unable to parse numerical level value " + itr->second); }
   }
-  const std::string& value = itr->second;
-  return value;
+  return values;
 }
 
-std::string EdgeInfo::level_ref() const {
```
chrstnbwnkl, line 417: A comment about backwards compatibility:  this will lead to `level_ref()` returning single values (the first one it finds in the map) when there are multiple, but IMO the way this was previously handled also didn't make much sense. Maneuvers that end on a flight of stairs would have the instruction "Take the stairs to level 1;2".

src/odin/enhancedtrippath.cc
```diff
@@ -523,19 +523,21 @@ std::vector<std::pair<std::string, bool>> EnhancedTripLeg_Edge::GetNameList() co
   return name_list;
 }
 
-std::string EnhancedTripLeg_Edge::GetLevelRef() const {
-  std::string level_ref;
+std::vector<std::string> EnhancedTripLeg_Edge::GetLevelRef() const {
```
chrstnbwnkl, line 526: When there are no `level_ref` values, fall back to the `level` ones.

src/odin/maneuversbuilder.cc
```diff
@@ -1157,7 +1157,11 @@ void ManeuversBuilder::InitializeManeuver(Maneuver& maneuver, int node_index) {
 
   // Set the end level ref
   if (curr_edge && !curr_edge->GetLevelRef().empty()) {
-    maneuver.set_end_level_ref(curr_edge->GetLevelRef());
+    if (curr_edge->GetLevelRef().size() > 1) {
```
chrstnbwnkl, line 1160: If there is more than one `level_ref` value for the current edge, set `end_level_ref` to empty, so the instruction would be something like "Take the stairs" if the maneuver ends on stairs.

src/odin/maneuversbuilder.cc
```diff
@@ -4006,10 +4014,12 @@ void ManeuversBuilder::AddLandmarksFromTripLegToManeuvers(std::list<Maneuver>& m
 
     for (auto node = man->begin_node_index(); node < man->end_node_index(); ++node) {
       auto curr_edge = trip_path_->GetCurrEdge(node);
-      // multipoint routes with `through` or `via` types can have consecutive copies of the same edge
+      // multipoint routes with `through` or `via` types can have consecutive copies of the same
```
chrstnbwnkl, line 4017: formatting

chrstnbwnkl (commented):
comments

src/mjolnir/osmway.cc
chrstnbwnkl: Maybe I'm mistaken but shouldn't the largest two byte varint be `(2**13)-1=8191`? 6 bits on the first (one for the sign), 7 on the second byte

kevinkreiser: yep you have it right. when i was just talking off the cuff earlier today and doing the math in my head i made two mistakes in quick succession:

1. i didnt think about the sign, which made me think it was 7 + 7 = 14 bits
2. i misremembered that 2^16 isnt 16k but rather 64k, so i took 16k in half 2 times rather than 64k in half 2 times

chrstnbwnkl:
One more thing that this PR doesn't address yet is fractional levels. There is limited documentation on this in the OSM wiki (see https://wiki.openstreetmap.org/wiki/Key:level#Level_designations), but according to [taginfo](https://taginfo.openstreetmap.org/keys/level#values) these values do occur in the wild. 

 Using one point decimal precision should cover >99% of all cases, but there are values where higher precision seems to be necessary (https://taginfo.openstreetmap.org/tags/?key=level&value=0.25#overview). In order to save space, it would be wise to use something like the varint encoding that @kevinkreiser suggested here, with the ability to store whether extra precision is necessary for a given level value (i.e. default to 1 point decimal precision, and use a sentinel value to indicate the next value requires two point decimal precision).

proto/common.proto
```diff
@@ -344,6 +344,7 @@ message TaggedValue {
     kLevel = 4;
     kLevelRef = 5;
     kLandmark = 6;
+    kLevels = 7;
```
chrstnbwnkl, line 347: Introduce a new enum value to not break compatibility

src/baldr/edgeinfo.cc
```diff
@@ -415,24 +460,44 @@ int8_t EdgeInfo::layer() const {
   return static_cast<int8_t>(value.front());
 }
 
-std::string EdgeInfo::level() const {
+bool EdgeInfo::includes_level(float lvl) {
```
chrstnbwnkl, line 463: I have removed the `level()` method and replaced it with `includes_level(lvl)`, because that's all we really do with the levels information at this point, and just returning the levels array that includes the separator values is of little use.

src/baldr/edgeinfo.cc
```diff
@@ -415,24 +460,44 @@ int8_t EdgeInfo::layer() const {
...
 }
 
-std::string EdgeInfo::level_ref() const {
+std::vector<std::string> EdgeInfo::level_ref() const {
   const auto& tags = GetTags();
-  auto itr = tags.find(TaggedValue::kLevelRef);
-  if (itr == tags.end()) {
-    return "";
+  std::vector<std::string> values;
+  for (auto [itr, range_end] = tags.equal_range(TaggedValue::kLevelRef); itr != range_end; ++itr) {
```
chrstnbwnkl, line 497: `level:ref` tags weren't split before, now they are split by semicolons.

src/odin/enhancedtrippath.cc
chrstnbwnkl: Edit: only if there is a single level value though; when forming the text instructions and a maneuver ends on a flight of stairs, we don't want the instruction to be "Take the stairs to level 1;2".

docs/docs/api/locate/api-reference.md
```diff
@@ -334,8 +334,16 @@ Here are some sample results with `verbose` set to `true`:
 ```
 
 ### Attribute Descriptions for Responses
+#### Edge
```
chrstnbwnkl, line 337: There are no real docs describing the output of `/locate`, and because the proposed `edge_info.levels` merits some explanation, I figured I'd start.

chrstnbwnkl:
Not sure yet what happens during varint deserialization that leads to negative values to always be -64... will have to take another look at this

valhalla/mjolnir/osmway.h
```diff
@@ -14,6 +14,34 @@
...
+  level_range_t(float s) : start(s), end(s){};
+  level_range_t() {
+    start = valhalla::baldr::kMinLevel;
+    end = valhalla::baldr::kMaxLevel;
+  };
+
+  /**
+   * get the max value this range covers
+   */
+  float max() {
```
chrstnbwnkl, line 33: remove this

src/mjolnir/osmway.cc
```diff
@@ -1067,10 +1098,100 @@ void OSMWay::GetTaggedValues(const UniqueNames& name_offset_map,
...
+          }
+          start = std::stof(token);
+        } catch (...) {
+          LOG_ERROR("Invalid level: " + token + "; way_id " + std::to_string(osmwayid_));
+          continue;
+        }
+        range.start = start;
+        range.end = start;
+      }
+      if ((range.start > values.back().max() || values.empty()) && range.valid()) {
```
chrstnbwnkl, line 1157: use std::sort instead, and get rid of the range struct

src/mjolnir/osmway.cc
```diff
@@ -1067,10 +1098,100 @@ void OSMWay::GetTaggedValues(const UniqueNames& name_offset_map,
...
+
+    // sentinel value that marks discontinuity
+    std::string sep = encode_level(kLevelRangeSeparator);
+
+    auto last_it = --values.end();
+    for (auto it = values.begin(); it != values.end(); ++it) {
+      auto& range = *it;
+      levels_encoded.append(encode_level(range.start * pow(10, precision)));
+      if (range.end != range.start)
+        levels_encoded.append(encode_level(range.end * pow(10, precision)));
```
chrstnbwnkl, line 1183: encode in one go instead of per value in the list

src/mjolnir/osmway.cc
```diff
@@ -1067,10 +1098,100 @@ void OSMWay::GetTaggedValues(const UniqueNames& name_offset_map,
...
+
+      if (it != last_it)
+        levels_encoded.append(sep);
+    }
+
+    std::string precision_enc = encode_level(static_cast<float>(precision));
+    std::string encoded =
+        encode_tag(TaggedValue::kLevels) +
+        encode_level(static_cast<float>(levels_encoded.size() + precision_enc.size())) +
```
chrstnbwnkl, line 1192: encode the number of ints, not the string size

src/mjolnir/osmway.cc
chrstnbwnkl: never mind

src/baldr/edgeinfo.cc
```diff
@@ -28,6 +28,27 @@ json::ArrayPtr names_json(const std::vector<std::string>& names) {
...
+  while (byte & 0x80 || shift == 0) {
+    byte = int32_t(*encoded);
+    result |= (byte & 0x7f) << shift;
+    shift += 7;
+    ++encoded;
+    ++i;
+  }
+
+  return (result & 1 ? ~result : result) >> 1;
+}
```
kevinkreiser, line 50:
```suggestion
int32_t parse_varint(const char*& encoded) {
  int32_t byte = 0, shift = 0, result = 0;

  while (byte & 0x80 || shift == 0) {
    byte = int32_t(*encoded);
    result |= (byte & 0x7f) << shift;
    shift += 7;
    ++encoded; 
  }

  return (result & 1 ? ~result : result) >> 1;
}
```

src/baldr/edgeinfo.cc
```diff
@@ -43,6 +64,13 @@ std::vector<std::string> parse_tagged_value(const char* ptr) {
       size_t landmark_size = landmark_name.size() + 10;
       return {std::string(ptr, landmark_size)};
     }
+    case TaggedValue::kLevels: {
+      // parse out the size to construct the string
+      uint32_t i = 0; // a place to store the size of the size varint
+      auto start = ptr + 1;
+      int size = static_cast<int>(parse_varint(start, i));
+      return {std::string(ptr, size + i + 1)};
+    }
```
kevinkreiser, line 73:
```suggestion
      // parse out the size to construct the string
      auto start = ptr + 1;
      int size = static_cast<int>(parse_varint(start));
      return {std::string(start, size)};
    }
```

kevinkreiser, line 73:
```suggestion
    case TaggedValue::kLevels: {
      // parse out the size to construct the string
      auto start = ptr + 1;
      int size = static_cast<int>(parse_varint(start));
      return {std::string(ptr, (start + size) - ptr)};
    }
```

src/baldr/edgeinfo.cc
```diff
@@ -415,24 +467,43 @@ int8_t EdgeInfo::layer() const {
...
+  auto itr = tags.find(TaggedValue::kLevels);
   if (itr == tags.end()) {
-    return "";
+    return false;
   }
-  const std::string& value = itr->second;
-  return value;
+  try {
+    auto decoded = std::get<0>(decode_levels(itr->second));
+    // TODO: use binary search
```
chrstnbwnkl, line 478: if we construct a vector of pairs instead, we could use `std::lower_bound` or `upper_bound`, which would be using binary search.

src/mjolnir/osmway.cc
chrstnbwnkl: call `pow` once

test/gurka/test_indoor.cc
```diff
@@ -1,13 +1,25 @@
...
 
 #if !defined(VALHALLA_SOURCE_DIR)
 #define VALHALLA_SOURCE_DIR
 #endif
 
 using namespace valhalla;
+using namespace mjolnir;
 const std::unordered_map<std::string, std::string> build_config{{}};
 
+struct range_t {
```
chrstnbwnkl, line 15: this is just to make the test a bit more concise

test/gurka/test_indoor.cc
```diff
@@ -280,4 +272,140 @@ TEST_F(Indoor, CombineStepsManeuvers) {
...
+    layout = gurka::detail::map_to_coordinates(ascii_map, gridsize_metres);
+    map = gurka::buildtiles(layout, ways, {}, {}, "test/data/gurka_levels");
+  }
+};
+
+gurka::map Levels::map = {};
+std::string Levels::ascii_map = {};
+gurka::nodelayout Levels::layout = {};
+
+TEST_F(Levels, EdgeInfoIncludes) {
```
chrstnbwnkl, line 313: This tests `includes_level`

test/scripts/test_valhalla_build_extract.py
```diff
@@ -151,15 +151,15 @@ def test_create_extracts(self):
         tile_count = len(tile_resolver.matched_paths)
 
         # test that the index has the right offsets/sizes
-        exp_tuples = ((2560, 25568, 296768), (301056, 410441, 665624), (968704, 6549282, 6137080))
+        exp_tuples = ((2560, 25568, 296768), (301056, 410441, 665624), (968704, 6549282, 6137088))
```
chrstnbwnkl, line 154: size of tiles changed

chrstnbwnkl:
I can't wrap my head around the test fails in CI...  I have tried to reproduce this using two machines, and inside a Docker setup, to no avail...

src/mjolnir/osmway.cc
```diff
@@ -13,6 +14,50 @@ constexpr uint32_t kMaxNodesPerWay = 65535;
...
+ * @return a string containing the encoded value
+ */
+std::string encode_level(const float lvl, const int precision = 0) {
+  std::string encoded;
+  // most of the time precision will be zero and level values will be lower
+  // than 4191 (even in the case of higher precision)
+  encoded.reserve(2);
+
+  int val = static_cast<int>(lvl * pow(10, precision));
+  const bool negative = val < 0;
```
chrstnbwnkl, line 34: The previous CI fails arose from a `static_cast` from int to unsigned int back to int. According to C++17 (§7.8/3 (Integral Conversions):

> If the destination type is signed, the value is unchanged if it can be represented in the destination type;
otherwise, the value is implementation-defined 

So instead, we perform a more manual conversion (taken from [another polyline implementation](https://github.com/vahancho/polylineencoder/blob/07db1476f369ff64c9bc2778021b064710847764/src/polylineencoder.h#L194-L211)

kevinkreiser (approved):
looks great to me, thanks for the deep dive on my mistake with the static_cast!

src/mjolnir/osmway.cc
chrstnbwnkl: Edit: after review with @kevinkreiser we decided to just use `reinterpret_cast` instead

Issue #2205, opened 2020-02-03
Map matching "failure" on loop route


dnesbitt61:
I am seeing an abbreviated map matching result (trace_route) on a path that begins and ends along the same edge. The original trace was long (95km) and only matched the initial maneuver. The return was not a failure, but the vast majority of the route was not matched. To me this is worse than a failure since the return seemed to indicate a success, but 99% of the trace was unmatched. If I removed the last 2 shape points (so the route ended before turning back onto the same edge as the start) the full path was matched. 

I was able to generate a simple test case of a route that just loops around the block:

{"shape":[{"lat": 39.958439, "lon": -75.609932},{"lat": 39.958340, "lon": -75.610153},{"lat": 39.958012, "lon": -75.610909},{"lat": 39.957520, "lon": -75.610672},{"lat": 39.957169, "lon": -75.610413},{"lat": 39.957081, "lon": -75.610336},{"lat": 39.957470, "lon": -75.609451},{"lat": 39.957630, "lon": -75.609062},{"lat": 39.958099, "lon": -75.609306},{"lat": 39.958389, "lon": -75.609520},{"lat": 39.958561, "lon": -75.609657},{"lat": 39.958450, "lon": -75.609909}],"shape_match":"map_snap", "costing":"bicycle", "format": "osrm", "shape_format": "polyline5"}

This is only matching the trace until the point of the first turn. The following images shows the route that was used to generate this test case (uses shape points from the initial route). The route starts by going west along West Gay St, that is the only maneuver returned.

![image](https://user-images.githubusercontent.com/1838768/73691959-649e5b00-46a1-11ea-95c4-8f779afa1cd5.png)

dnesbitt61:
@kevinkreiser - not sure if you have seen something like this...let me know if this duplicates any other issues.

dnesbitt61:
There are 8 path edges in the MatchResult, so I think the issue is within thor:map_match method. There are no disconnected_edges in the result either.
 
If the last shape input is removed the full route (ending at intersection of Potter Alley and W Gay St) is returned.

dnesbitt61:
Here is a similar route in Utrecht that perhaps can be used as a test case:

{"shape":[
{"lat": 52.073929, "lon": 5.087180},
{"lat": 52.074268, "lon": 5.086840},
{"lat": 52.074730, "lon": 5.086450},
{"lat": 52.074940, "lon": 5.087080},
{"lat": 52.075081, "lon": 5.087490},
[...]

PR #4895, opened 2024-09-19
Loop match walk or snap fix

ianthetechie:
# Issue

More fun with matching loops (#2205 and others)!

I ran into this when debugging another loop issue when map matching GTFS feeds, which are a great source of loops. I found a problem. I found _a_ fix. But I don't know if it's the correct one ;)

There is an optimization(?) for identifying trivial routes on a single edge in the walking route matcher. I assume that "trivial" in this case means "only ever traverses (some of) a single edge." This appears to be causing problems. In such cases, if you're in `walk_or_snap` mode, you'll get the following server error:

```json
{
  "error_code": 299,
  "error": "Unknown: Could not build directions for TripLeg",
  "status_code": 400,
  "status": "Bad Request"
}
```

The route successfully matches in the more expensive `map_snap` mode, but it fails in `edge_walk` mode. The curious thing is that, in `walk_or_snap` mode, the error handling around https://github.com/valhalla/valhalla/blob/3a385045919967a14d5c9cc57610b2111936ac64/src/thor/trace_route_action.cc#L111 is such that it will fall back to `map_snap` when `edge_walk` fails. Indeed, the expected warning log is printed. However, it never succeeds.

Why it does not successfully find a match, and fails with this error about building trip legs, I have not had time to uncover, but since almost all of this code appears to be dealing with mutable references, it is quite difficult to reason about.

My fix comes from observing that the "trivial" case detection is way too eager. I thought of a few ways to fix this, but also observed that it was being hit for my routes as well as a constructed example (in the unit tests now) after breaking from the preceeding while loop due to examining a distance greater than the edge length. That looks to me like a situation where the route is almost certainly *not* trivial, so I have cut out the check using the break condition.

## Tasklist

 - [x] Add tests

### DIFF
--- a/test/mapmatch.cc
+++ b/test/mapmatch.cc
@@ -1232,6 +1232,33 @@ TEST(Mapmatch, test_loop_matching) {
   }
 }
 
+TEST(Mapmatch, test_loop_matching_walk_or_snap) {
+  // The walk_or_snap case sometimes incorrectly identifies trivial cases where it can optimize away.
+  // This tests loops which are not trivial.
+  std::vector<std::string> test_cases = {
+      // Simple loop around a block; incorrectly identified as trivial in such a way that
+      // it caused an exception in TripLegBuilder (no maneuvers)
+      R"({"shape_match":"walk_or_snap", "shape":[
+          {"lat": 52.11870746741026, "lon": 5.117783546447755, "type": "break", "node_snap_tolerance": 0},
+          {"lat": 52.11905661947785, "lon": 5.118234157562257, "type": "via", "node_snap_tolerance": 0},
+          {"lat": 52.11934977334695, "lon": 5.117751359939576, "type": "via", "node_snap_tolerance": 0},
+          {"lat": 52.118852398789194, "lon": 5.116935968399049, "type": "via", "node_snap_tolerance": 0},
+          {"lat": 52.11852959643748, "lon": 5.117445588111878, "type": "via", "node_snap_tolerance": 0},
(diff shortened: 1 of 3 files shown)


test/mapmatch.cc
```diff
@@ -1232,6 +1232,33 @@ TEST(Mapmatch, test_loop_matching) {
...
+          {"lat": 52.118852398789194, "lon": 5.116935968399049, "type": "via", "node_snap_tolerance": 0},
+          {"lat": 52.11852959643748, "lon": 5.117445588111878, "type": "via", "node_snap_tolerance": 0},
+          {"lat": 52.11870746741026, "lon": 5.117783546447755, "type": "break", "node_snap_tolerance": 0}],
+          "costing":"auto"})",
+  };
+
+  api_tester tester;
+  for (size_t i = 0; i < test_cases.size(); ++i) {
+    // Verify that the map_snap output matches the walk_or_snap output.
+    // This approach is inspired by the above.
```
ianthetechie, line 1254: Quick note in case it sparks any ideas... I attempted the same logic as above with verifying that a route over the same points was identical, but the /route response always ended up with a spurious keep right + continue after the 2nd maneuver. Figuring that out is out of scope of this PR, so I went with something a bit more simple. We expect the `map_snap` and `walk_or_snap` results to match, where they did not before.

kevinkreiser: we'd need to look at the trip path. there has to be a difference in how the edges look like two copies of the same edge where a via point lands for example. its unfortunate that that would happen but i totally could see that it could happen.

kevinkreiser (approved):

Issue #3699, opened 2022-08-05
trace_attributes: duplicate waypoints trigger edge_index bug


nilsnolde:
ref. https://github.com/valhalla/valhalla/issues/3371#issuecomment-1205090389

apparently in some situations duplicate waypoints trigger an already solved bug: #3371 . the following request shows it:

```
curl --request POST \
  --url https://valhalla1.openstreetmap.de/trace_attributes \
  --header 'Content-Type: application/json' \
  --data '{
	"shape": [
		{
			"lat": 57.6512141667,
			"lon": 12.0509125
		},
		{
			"lat": 57.6512141667,
			"lon": 12.0509125
		},
		{
			"lat": 57.6512141667,
			"lon": 12.0509125
		},
		{
			"lat": 57.65126,
			"lon": 12.0507619444
		},
		{
			"lat": 57.6512294444,
			"lon": 12.0505425
		},
		{
			"lat": 57.6512294444,
			"lon": 12.0505425
		}
	],
	"search_radius": 150,
	"shape_match": "map_snap",
	"costing": "auto",
	"format": "osrm"
}'
```

when the duplicates are removed it works fine. although it's not only duplicates.. if you just remove the middle two it'll also work, although there are only duplicates left. it'll also work if the first _or_ last two are removed, but not if only the first _or_ last.. very strange behavior!

ianthetechie:
Similar to other bugs of this class like #3626 and #3371, I ended up hitting an example of this and at first thought it was due to degenerate user input, but it turns out that it can happen even when tracing attributes on a route from Valhalla.

Here's a fairly small polyline that will reproduce the issue in `map_snap` mode: `guszaAfirtnDoA~bBGnIOzOmAz`DC`EY~e@g@fdBY~wAYxo@}BdjFAx@?x@s@b}BC|ODpMR~H`. It seems to appear near intersections, though in this case it does not occur along a road change.

In the trace serializer, what's happening is that the `match_result.edeid.Is_Valid()` check succeeds because, well, it DOES have a valid edge ID ;)

https://github.com/valhalla/valhalla/blob/0c848128c7fe213f3af96e48501e61e39a451923/src/tyr/trace_serializer.cc#L401

However, `match_result.edge_index` is still `meili::kInvalidEdgeIndex`, so that's what gets serialized ;)

[...]

PR #4911, opened 2024-09-30
Don't return edge invalid edge_index values

ianthetechie:
# Issue

This is a _mitigation_ related to #3699. While this addresses one of the symptoms, it does not address the underlying problem. That said, I think that simply omitting the edge index rather than returning an invalid one, is better than the status quo. As noted in the linked issue, if anyone has any pointers on how to address the root cause, I'm happy to work on that.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

N/A


ianthetechie:
Ha! Just realized this was still open when looking at our prod branch divergence 😂 Anyone have an objection to this PR / better ideas for how to solve it if we don't want to merge a mitigation?

kevinkreiser:
I saw this but forgot about it. Not merging it wasn't deliberate. vann ou update it?

ianthetechie:
No worries :) Updated.

kevinkreiser:
looks like its non trivial to merge. there are lots of diffs in the main map matching test. if we can i think i'll suggest to @chrstnbwnkl that instead of review sometimes we'll use of the 2 weekly meetings we have to do code changes to accomodate other peoples work. there are a bunch of small things ive started but havent finished and stuff like this is a place where its small enough i should, with a little bit of work, be able to suggest a way forward to unblock the basic issue here

ianthetechie:
Interesting; I had no issues when building this as a patch on our prod branch (which is maybe a month or 2 behind master). The PR I opened yesterday, which doesn't touch map matching and is based off master as of yesterday, also shows some map matching diffs in a failed CI run, which is quite unexpected: https://github.com/valhalla/valhalla/actions/runs/11779814286/job/32834504650.

kevinkreiser:
yeah it totally builds just the map matching test fails a bunch of test cases. could be as easy as the tests expecting a different value now. ive not really thought about the semantic difference tbh

HarelM:
Any updates on this? I just came across this issue when trying to add a test for rallista's valhalla wrapper to allow route trace and route attributes.

kinkard:
@HarelM do you have a request that reproduces this issue?

HarelM:
I might be able to generate one from the following PR I opened to valhalla-mobile:
https://github.com/Rallista/valhalla-mobile/pull/95
The following is the fix I added in order to overcome the failure in the test:
https://github.com/Rallista/valhalla-mobile/pull/95/changes/697380ff5d64b85c416ee7d3feaa6433eb9dd4b6
It adds a filter to the test to preventing if from failing due to this issue.
Does it count?

kinkard:
> I might be able to generate one from the following PR I opened to valhalla-mobile:
https://github.com/Rallista/valhalla-mobile/pull/95
The following is the fix I added in order to overcome the failure in the test:
https://github.com/Rallista/valhalla-mobile/commit/697380ff5d64b85c416ee7d3feaa6433eb9dd4b6
It adds a filter to the test to preventing if from failing due to this issue.
Does it count?

I can't say that these steps are simple enough to reproduce and fix the root cause issue in Valhalla, at least for me.

HarelM:
Is there any way I can help move this forward?
It's worth mentioning (which you probably know) that this is a PR that someone else had opened to fix this issue, so I'm not the first person to come across this problem.

kinkard:
> I'm not the first person to come across this problem

That's fine. I was wondering if you have a reliable way to reproduce this issue so we can understand the underlying root cause that leads to such invalid `edge_index` values instead of filtering such invalid values in serializer.

HarelM:
I can try and port the test from valhalla-mobile to this repo's test infra if that can help... (More like ask Claude to do it, but I'll be happy to help out). Not sure I'll be able to add it to this PR, but I can open a different PR. WDYT?

nilsnolde:
this still produces the same thing: https://github.com/valhalla/valhalla/issues/3699.

kinkard:
@HarelM can you please check if https://github.com/valhalla/valhalla/pull/6278 fixes your issue?

kinkard:
Superseded by https://github.com/valhalla/valhalla/pull/6278

PR #4925, opened 2024-10-08
dont use static_cast for negative varint encoding

kevinkreiser:
over in #4879 @chrstnbwnkl noticed that when varint encoding negative values certain values fail the tests. this turned out to be a result of how we handle the movement of the sign bit to do the encoding for negative values. the main issue with my initial implementation was that to get the integer into something that we could reliably bitshift i `static_cast`ed it to unsigned. `static_cast` is meant really to do a conversion in the human sense of understanding numbers and for negatives it wraps around which isnt what we want. what we want is to treat it like a blob of bytes that we can do bitwise operations on, hence `reinterpret_cast` is the right tool for the job

### DIFF
--- a/valhalla/midgard/encoded.h
+++ b/valhalla/midgard/encoded.h
@@ -177,7 +177,7 @@ std::string encode(const container_t& points, const int precision = ENCODE_PRECI
   // handy lambda to turn an integer into an encoded string
   auto serialize = [&output](int number) {
     // move the bits left 1 position and flip all the bits if it was a negative number
-    number = number < 0 ? ~(static_cast<unsigned int>(number) << 1) : (number << 1);
+    number = number < 0 ? ~(*reinterpret_cast<unsigned int*>(&number) << 1) : (number << 1);
     // write 5 bit chunks of the number
     while (number >= 0x20) {
       int nextValue = (0x20 | (number & 0x1f)) + 63;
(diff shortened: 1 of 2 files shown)


valhalla/midgard/encoded.h
```diff
@@ -177,7 +177,7 @@ std::string encode(const container_t& points, const int precision = ENCODE_PRECI
   // handy lambda to turn an integer into an encoded string
   auto serialize = [&output](int number) {
     // move the bits left 1 position and flip all the bits if it was a negative number
-    number = number < 0 ? ~(static_cast<unsigned int>(number) << 1) : (number << 1);
+    number = number < 0 ? ~(*reinterpret_cast<unsigned int*>(&number) << 1) : (number << 1);
```
kevinkreiser, line 180: yeah so we want to tell the compiler hey just think of this signed integer as some bytes that we want to do bitwise manipulation to, that is what `reinterpret_cast` does for us. `static_cast` is meant to do an actual conversion which in this case we dont actually want and i just threw in there without thinking about it all those years ago

chrstnbwnkl (approved):
It's highly unlikely that this bug never occured :smile: I saw wrongly encoded values occur with small negative values like -2 or -1 (that led to being encoded as -64 instead), so it is possible that no one noticed because that would have led to sub-meter differences using GeoJSON or polyline6, but probably to several meter differences using polyline5? Anyway, thanks for the explainer :smile:

valhalla/midgard/encoded.h
sebrockm: @kevinkreiser I'm fairly sure that the original implementation was correct and `reinterpret_cast` is incorrect (if I understood correctly that the goal is to not change the bit pattern at all, just the type). See https://stackoverflow.com/questions/1751346/interpret-signed-as-unsigned
Quote from the accepted answer:
> Note: In a two’s complement representation, this conversion is conceptual and there is no change in the bit pattern (if there is no truncation
> ...
> The mapping performed by reinterpret_cast is implementation-defined. [Note: it might, or might not, produce a representation different from the original value. ]

PR #4926, opened 2024-10-08
Level location filtering

### LINKED ISSUE #4878: Control level in location search

chrstnbwnkl:
# Issue

resolves #4878.  
 
## Tasklist

 - [x] Add tests


valhalla/baldr/pathlocation.h
```diff
@@ -228,6 +229,14 @@ struct PathLocation : public Location {
       l.search_filter_.exclude_bridge_ = loc.search_filter().exclude_bridge();
       l.search_filter_.exclude_ramp_ = loc.search_filter().exclude_ramp();
       l.search_filter_.exclude_closures_ = loc.search_filter().exclude_closures();
+      if (loc.search_filter().has_level()) {
+        l.search_filter_.level_ = loc.search_filter().level();
+        if (loc.has_search_cutoff_case()) {
+          // set smaller default and add warning
+          l.search_cutoff_ = 300;
+          add_warning(request, 401, std::to_string(300) + " meters");
```
chrstnbwnkl, line 237: Not sure yet about this, this would add one warning per location. When serializing out to json, we could dedupe these.

test/gurka/test_multi_layer_loki.cc
```diff
@@ -21,7 +21,7 @@ struct Waypoint {
```
chrstnbwnkl, line 21: renamed to avoid confusion

src/worker.cc
```diff
@@ -408,6 +410,10 @@ void parse_location(valhalla::Location* location,
     // search_filter.exclude_ramp
     location->mutable_search_filter()->set_exclude_ramp(
         rapidjson::get<bool>(*search_filter, "/exclude_ramp", false));
+    auto level =
+        rapidjson::get<float>(*search_filter, "/level", static_cast<float>(baldr::kMaxLevel));
+    if (level != baldr::kMaxLevel)
+      location->mutable_search_filter()->set_level(level);
```
kevinkreiser, line 416: need to worry about when person uses a level that is higher than max and also the raw proto input variant of this has to get validated

valhalla/baldr/pathlocation.h
```diff
@@ -228,6 +230,9 @@ struct PathLocation : public Location {
       l.search_filter_.exclude_bridge_ = loc.search_filter().exclude_bridge();
       l.search_filter_.exclude_ramp_ = loc.search_filter().exclude_ramp();
       l.search_filter_.exclude_closures_ = loc.search_filter().exclude_closures();
+      if (loc.search_filter().has_level_case()) {
+        l.search_filter_.level_ = loc.search_filter().level();
```
chrstnbwnkl, line 234: set this to kMaxLevel when we get it off the json

src/worker.cc
chrstnbwnkl: no real need though to clamp since max level is already the numerical limit of `float`, right?

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -46,7 +46,7 @@ To build a route, you need to specify two `break` locations. In addition, you ca
...
...oads marked as bridges</li><li>`exclude_ramp` (boolean, defaults to `false`): whether to exclude link roads marked as ramps, note that some turn channels are also marked as ramps</li><li>`exclude_closures` (boolean, defaults to `true`): whether to exclude roads considered closed due to live traffic closure. **Note:** This option cannot be set if `costing_options.<costing>.ignore_closures` is also specified. An error is returned if both options are specified. **Note 2:** Ignoring closures at destination and source locations does NOT work for date_time type `0/1` & `2` respectively</li><li>`min_road_class` (string, defaults to `"service_other"`): lowest road class allowed</li><li>`max_road_class` (string, defaults to `"motorway"`): highest road class allowed</li><li>**BETA** `level` (float): if specified, will only consider edges that are on or traverse the passed floor level. It will set `search_cutoff` to a default value of 300 meters if no cutoff value is passed. Additionally, if a `search_cutoff` is passed, it will be clamped to 1000 meters.</li></ul>Road classes from highest to lowest are: motorway, trunk, primary, secondary, tertiary, unclassified, residential, service_other. |
```
chrstnbwnkl, line 49: added the beta label

src/loki/worker.cc
```diff
@@ -48,15 +51,35 @@ void loki_worker_t::parse_locations(google::protobuf::RepeatedPtrField<valhalla:
...
+        has_302 = true;
+      }
+      // if there is a level search filter and
       if (!location.has_street_side_tolerance_case())
         location.set_street_side_tolerance(default_street_side_tolerance);
 
       if (!location.has_street_side_max_distance_case())
         location.set_street_side_max_distance(default_street_side_max_distance);
     }
+    if (has_302)
```
chrstnbwnkl, line 78: max two warnings per request instead of per location

src/worker.cc
```diff
@@ -436,6 +442,8 @@ void parse_location(valhalla::Location* location,
   if (!location->search_filter().has_max_road_class_case()) {
     location->mutable_search_filter()->set_max_road_class(valhalla::kMotorway);
   }
+  if (!location->search_filter().has_level_case())
```
chrstnbwnkl, line 445: always set default sentinel if not set

chrstnbwnkl:
Just the occasional win fail, seems unrelated

src/loki/search.cc
```diff
@@ -34,11 +34,13 @@ bool search_filter(const DirectedEdge* edge,
   // Note that min_ and max_road_class are integers where, by default, max_road_class
   // is 0 and min_road_class is 7. This filter rejects roads where the functional
   // road class is outside of the min to max range.
+
```
kevinkreiser, line 37: nit: remove

kevinkreiser (approved):

PR #4941, opened 2024-10-21
Optionally add admin crossings to route response

chrstnbwnkl:
# Issue

Adds an `admin_crossings` request parameter for `/route`. The motivation is that currently, an additional call to `/trace_attributes` is necessary to obtain any information about admins for the route in JSON format. With this PR, a user can optionally request admin info, which adds two keys to each leg on the JSON response: 

- `admins`: an array of admin objects that the route crosses, that contains:  
  -  `country_iso`
  - `state_iso` 
  - `country_text`
  - `state_text` 
- `admin_crossings`, an array of objects denoting the start and and end shape indices along which there is an admin crossing (including the keys `from_admin_index` and `to_admin_index`, which are indices into the `admin` array)

Now that I spell it out, maybe it makes more sense to name the request parameter to something more generic `admins`, to be able to reuse this across end points in the future.    

## Tasklist

 - [x] Add tests


chrstnbwnkl:
All OSX Test builds are failing, looks like it's possibly Python 3.13 related

test/gurka/test_admin.cc
```diff
@@ -9,7 +10,7 @@ using namespace valhalla;
 const std::unordered_map<std::string, std::string> build_config{
     {"mjolnir.data_processing.use_admin_db", "false"}};
 
-class Admin : public ::testing::Test {
+class AdminTest : public ::testing::Test {
```
chrstnbwnkl, line 13: `Admin` was shadowing `baldr::Admin`

proto/options.proto
```diff
@@ -498,4 +498,7 @@ message Options {
   bool voice_instructions = 57;                                    // Whether to return voiceInstructions in the OSRM serializer response
   bool dedupe = 58;                                                // Keep track of edges and override their properties during expansion,
                                                                    // ensuring that each edge appears in the output only once. [default = false]
+  oneof has_admin_crossings {
+    bool admin_crossings = 59;                                     // Include administrative boundary crossings
+  }
```
kevinkreiser, line 503:
```suggestion
    bool admin_crossings = 59;                                     // Include administrative boundary crossings
```

src/tyr/route_serializer_valhalla.cc
```diff
@@ -530,6 +530,43 @@ void legs(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper_t
       writer.end_array(); // elevation
     }
 
+    // Does the user want admin info?
+    if (api.options().admin_crossings() && trip_leg_itr->admin_size() > 0) {
```
kevinkreiser, line 534: so maybe we only want this when its more than 1 meaning an actual crossing occured, you left one and went to at least one other?

src/tyr/route_serializer_valhalla.cc
```diff
@@ -530,6 +530,43 @@ void legs(const valhalla::Api& api, int route_index, rapidjson::writer_wrapper_t
...
+            writer.end_object();
+          }
+          ++node_itr;
+          ++next_node_itr;
+        }
+        writer.end_array();
+      }
+    }
+
     writer.start_object("summary");
```
kevinkreiser, line 570: this feels a bit like summary info maybe it should go inside the summary? i dont feel strongly about it though so feel free to ignore

kevinkreiser (dismissed):
this looks good maybe just one more test with a border that is the same geom as the edge you are traveling on might be fun, happens a lot for admin level 3 (or whatever states are)

src/tyr/route_serializer_valhalla.cc
chrstnbwnkl: Moved it into the summary

src/tyr/route_serializer_valhalla.cc
chrstnbwnkl: We should always get the `admins` array even if it's only one (it's never going to be less than that), and only get the crossings when there are at least two admins along the route. Updated the code to reflect that.

kevinkreiser (approved):

kevinkreiser (approved):

Issue #3650, opened 2022-06-07
Indoor Navigation


dmenneck:
I am currently facing the use case of navigating in buildings. Unfortunately I couldn’t find any examples so far and wanted to ask if some of you have already made experience with it and/or can estimate if indoor navigation with Valhalla is possible.

The underlying OSM data will be tagged with schemas like IndoorOSM or Simple Indoor Tagging.

kevinkreiser:
@dmenneck it is possible. in #3519 we added some basic support for this. i'm neither familiar with "IndoorOSM" nor "Simple Indoor Tagging" but if it uses the osm data model and the tags are ones that our processing cares about it should somewhat work. you can always give it a try and see how far it goes and what shortcomings you find. it would be great to cataloge a list of that here so we can address it as more of a larger project.

dmenneck:
Thank you @kevinkreiser !

Is there any documentation somewhere where I can find the tags that Valhalla expects for Indoor Navigation? Which OSM tags should I use for mapping a building and making Valhalla able to navigate over the graph?

nilsnolde:
You can look at the graph lua file in this PR which added the indoor stuff to the graph: https://github.com/valhalla/valhalla/pull/3509/files, which is fairly straight forward to derive from.

dmenneck:
You‘re right @nilsnolde, it’s actually straight forward, thanks!

Since Valhalla can handle stairs/elevators/escalators it’s obviously possible to navigate from one floor to another. I’m curious how a request would be structured to navigate eg from the first floor to the fourth.

Do you have an example for that?

kevinkreiser:
@dmenneck the unit tests in the pr linked above does some test requests but at the present time we dont have a way to select what floor you are on if you want to set a destination on a certain floor. we had some discussion about it before but its very similar to the "level" parameter we would use to select a road when multiple roads are on top of one another. if your question is focused on this problem the answer at the moment is we've not yet implemented the ability to select the "level" or "floor" but if the destination or origin that you pick is clearly on top of one walk way vs another it will pick that one regardless of the floor its on. for ambiguous cases where they are on top of each other it will depend on the ordering of the data unfortunately.

kevinkreiser:
both work should work already but the problem is when hallways on two different floors are on top of each other, in that ambiguous case there is no way to tell the router which floor you meant

kevinkreiser:
[...]

PR #4942, opened 2024-10-21
Generate level change info for `/route` response

chrstnbwnkl:
# Issue

As discussed in #3650, Valhalla does not return any information about traversed levels in the route response. Hence, there is no way to properly render route shapes in multi-level contexts. This PR addresses this by adding a `level_changes` key to the route JSON response, which is a sparse array of arrays containing two elements: 

 - the index into the shape at which a new level has been reached 
 - the level as a floating point value 

A current caveat is that since level information is only stored on the edge rather than the node, a level change is only detected once the trip reaches an edge that is situated on only one level.      

## Tasklist

 - [x] Add tests


src/loki/search.cc
```diff
@@ -34,6 +34,7 @@ bool search_filter(const DirectedEdge* edge,
   // Note that min_ and max_road_class are integers where, by default, max_road_class
   // is 0 and min_road_class is 7. This filter rejects roads where the functional
   // road class is outside of the min to max range.
+
```
chrstnbwnkl, line 37: remove

src/thor/triplegbuilder.cc
```diff
@@ -1052,11 +1052,14 @@ void ProcessTunnelBridgeTaggedValue(valhalla::StreetName* trip_edge_name,
  * @param  has_junction_name  True if named junction exists, false otherwise
  * @param  start_tile         The start tile of the start node
  * @param  blind_instructions Whether instructions should be generated for blind users
- *
+ * @param  prev_level         The previous level the path is on. Gets updated if the new edge is
+ *                            on a different level.
+ * @param  trip               The trip path
+ * @param  shape_index        The index of the edges first point into the final leg shape
```
chrstnbwnkl, line 1058: update docstring

src/thor/triplegbuilder.cc
```diff
@@ -1069,16 +1072,15 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
                           const uint32_t start_node_idx,
                           const bool has_junction_name,
                           const graph_tile_ptr& start_tile,
-                          const uint8_t restrictions_idx,
-                          float elapsed_secs,
-                          bool blind_instructions) {
+                          bool blind_instructions,
+                          EdgeInfo& edgeinfo,
+                          const std::pair<std::vector<std::pair<float, float>>, uint32_t>& levels) {
```
chrstnbwnkl, line 1077: pass edgeinfo and levels, and while at it simplify the  signature a bit by simply passing `edge_itr`

src/thor/triplegbuilder.cc
```diff
@@ -1190,7 +1192,7 @@ TripLeg_Edge* AddTripEdge(const AttributesController& controller,
     if (mode == sif::TravelMode::kPublicTransit) {
       // TODO(nils): get the actual speed here by passing in the elapsed seconds (or the whole
       // pathinfo)
-      speed = directededge->length() / elapsed_secs * kMetersPerSectoKPH;
+      speed = directededge->length() / edge_itr->elapsed_cost.secs * kMetersPerSectoKPH;
```
chrstnbwnkl, line 1195: some small changes as a result of that refactor

kevinkreiser (approved):

PR #4945, opened 2024-10-22
Fix macOS CI 

chrstnbwnkl:
# Issue

CMake is picking up the wrong Python, ~~probably because the default version installed by homebrew is now 3.13, which is newer than the system one (3.12.x). See if changing the default policy to first matching instead of newest fixes this.~~

Edit: I think some package recently started depending on Python. I found out that CMake's `FindPython` always prefers non-system Python, and we did not have this issue ever before. I tried  (unsuccessfully) to force CMake to use system Python. The easiest solution seems to be to just install it directly via homebrew and use it.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -59,15 +59,15 @@ jobs:
             ccache-osx-
 
       - name: Configure CMake
-        run: cmake -B build -DENABLE_SINGLE_FILES_WERROR=OFF -DENABLE_GDAL=ON
+        run: |
+          source .venv/bin/activate
+          cmake -B build -DENABLE_SINGLE_FILES_WERROR=OFF -DENABLE_GDAL=ON
 
       - name: Build Valhalla
         run: make -C build -j$(sysctl -n hw.logicalcpu)
 
       - name: Build Tests
-        run: |
-          source .venv/bin/activate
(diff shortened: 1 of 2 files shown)


kevinkreiser (dismissed):

.github/workflows/osx.yml
```diff
@@ -33,14 +33,15 @@ jobs:
...
         with:
-          python-version: '3.12' 
-      
+          python-version: "3.12"
+
       - name: Install dependencies
         run: |
-          HOMEBREW_NO_AUTO_UPDATE=1 brew install autoconf automake protobuf cmake ccache libtool sqlite3 libspatialite luajit curl wget czmq lz4 spatialite-tools unzip boost gdal
+          HOMEBREW_NO_AUTO_UPDATE=1 brew install python autoconf automake protobuf cmake ccache libtool sqlite3 libspatialite luajit curl wget czmq lz4 spatialite-tools unzip boost gdal
+          export PATH="$(brew --prefix python)/libexec/bin:$PATH"
```
chrstnbwnkl, line 44: Only two changes except for whitespace and double quotes formatting:
1) install python via homebrew
2) update $PATH so homebrew python is

kevinkreiser (approved):

PR #4960, opened 2024-11-05
Improve steps maneuvers

chrstnbwnkl:
# Issue

Currently, steps maneuvers don't generally lead to a dedicated maneuver. I suppose this is because steps can be trivial (i.e. it's not necessary in an outdoor routing situation to always indicate steps even if it's only a couple of stairs).  This PR adds dedicated maneuvers for stairs that traverse levels.

## Tasklist

 - [x] Add tests


chrstnbwnkl:
Never mind about the additional elevator maneuvers, I had a wrong understanding about what maneuvers are supposed to represent (a point and some part of the route leading up to it). Still then, the way maneuvers work for node-modeled elevators does not seem to be correct, because they include the portion after the elevator, not the portion leading up to it.

kevinkreiser (approved):

PR #4965, opened 2024-11-11
Support the HTTP Accept-Language header

### LINKED ISSUE #4948: Accept-Language header support
Getting localized directions to users is something that's interesting to me, and I'm wondering if there's a better way than the current approach of specifying a single 2 letter language code. Setting a language code in the query string or POST body is fine if you know, for example, that you need directions in Korean AND you also know that Valhalla supports Korean (it doesn't :P). But maybe the user also understands some other supported language which they'd prefer over English directions?

There's a web standard for this already; both web browsers and (good) mobile HTTP clients send a useful hierarchy of wants in the Accept-Language header, and some APIs (like Pelias) take this into account to automatically localize content. Valhalla could improve the UX by supporting this header too, parsing the priority list, and identifying a language for the response based on what's supported.

Is there interest in supporting this directly in Valhalla? If so, I'd be happy to take a swing at implementation.

ianthetechie:
# Issue

Fixes #4948.

## Tasklist

 - [X] Add tests

## Requirements / Relations

N/A

### DIFF
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -1,3 +1,4 @@
+#include <regex>
 #include <sstream>
 #include <typeinfo>
 #include <unordered_map>
@@ -1339,6 +1340,61 @@ std::string serialize_error(const valhalla_exception_t& exception, Api& request)
   return body.str();
 }
 
+// Regex for matching components of an Accept-Language header.
+const std::regex language_re(R"(([a-zA-Z0-9\-]{2,}|\*)(?:;q=(\d(?:\.\d+)?))?)");
+
+std::string ParseAcceptLanguage(headers_t headers) {
+  auto accept_language_header = headers.find("Accept-Language");
+  if (accept_language_header != headers.end()) {
+    // Parse the list of languages.
+    // The final argument says we want to get groups 1 and 2 from each match.
+    std::sregex_token_iterator it(accept_language_header->second.begin(), accept_language_header->second.end(), language_re, {1, 2});
+    std::sregex_token_iterator end;
+
+    // Iterate through weighted matches
+    bool found_usable_language = false;
+    // Pre-allocate; ternary necessary because with no matches, length will segfault
+    std::vector<std::pair<float, std::string>> languages(it != end ? it->length() : 0);
+    while (it != end) {
+      std::string language = *it++;
+      float q_value = 1.0;
+
+      if (it != end && it->length() > 0) {
+          q_value = std::stof(*it);
+      }
+
+      // Check if it's a usable language first.
+      // In the case of headers, it's often implicit without the developer having to act,
+      // so we gracefully ignore any unsupported options and pick the best one.
+      // We do not currently consider things like falling back to a different variant
+      // of a supported language, but this could be considered in the future.
+      if (odin::get_locales().find(language) != odin::get_locales().end()) {
+        // If the quality is 1 (max) and we don't have any usable matches yet,
+        // we can skip further checks
+        if (q_value == 1.0 && !found_usable_language) {
+          return language;
+        }
+        found_usable_language = true;
+
+        languages.push_back(std::make_pair(q_value, language));
+      }
+
+      ++it;
+    }
+
+    auto max_it = std::max_element(languages.begin(), languages.end(),
+                                   [](const std::pair<float, std::string>& a, const std::pair<float, std::string>& b) {
+                                     return a.first < b.first;
+                                   });
+
+    if (max_it != languages.end()) {
+      return max_it->second;
+    }
+  }
+
+  return "";
+}
+
 void ParseApi(const std::string& request, Options::Action action, valhalla::Api& api) {
   // maybe parse some json
   auto document = from_string(request, valhalla_exception_t{100});
@@ -1412,6 +1468,13 @@ void ParseApi(const http_request_t& request, valhalla::Api& api) {
       array.PushBack({value, allocator}, allocator);
     }
     document.AddMember({kv.first, allocator}, array, allocator);
+  }
+
+  if (!document.HasMember("language")) {
+    // Fall back to the Accept-Language header
+    auto language = ParseAcceptLanguage(request.headers);
+
+    document.AddMember({"language", allocator}, {language, allocator}, allocator);
   }
 
   // parse out the options
--- a/test/parse_request.cc
+++ b/test/parse_request.cc
@@ -373,6 +373,14 @@ Api get_request(const std::string& request_str, const Options::Action action) {
   return request;
 }
 
+Api get_http_request(const std::string& body, const prime_server::headers_t headers = prime_server::headers_t{}) {
+  Api api;
+  prime_server::http_request_t request(prime_server::method_t::POST, "/route", body, prime_server::query_t{}, headers);
+  ParseApi(request, api);
+  return api;
+}
+
+
 ///////////////////////////////////////////////////////////////////////////////
 // test parsing methods
 std::string get_costing_str(Costing::Type costing) {
@@ -1641,6 +1649,21 @@ void test_closure_factor_parsing(const Costing::Type costing_type,
   validate(key, expected_value, options.closure_factor());
 }
 
+void test_language_parsing(const std::string expected_value,
+                           const std::string language_arg,
+                           const std::string language_header) {
+  const std::string key = "language";
+  auto req_body = language_arg.empty() ? "{}" : get_request_str(key, language_arg);
+
+  prime_server::headers_t headers{};
+  if (!language_header.empty()) {
+    headers.insert({"Accept-Language", language_header});
+  }
+
+  Api request = get_http_request(req_body, headers);
+  validate(key, expected_value, request.options().language());
+}
+
 // utility functions for testing disable_hierarchy_pruning
 // Create costing options (reference: /test/astar.cc)
 void create_costing_options(Options& options, Costing::Type type) {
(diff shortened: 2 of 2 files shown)


ianthetechie:
There's a pretty high probability the formatting checks fail (I'm developing on macOS). I've enabled maintainer edits, so hopefully someone else can just push a fix :)

docs/docs/api/turn-by-turn/api-reference.md
```diff
@@ -303,7 +303,7 @@ Directions options should be specified at the top level of the JSON object.
 | Options | Description |
 | :------------------ | :----------- |
 | `units` | Distance units for output. Allowable unit types are miles (or mi) and kilometers (or km). If no unit type is specified, the units default to kilometers. |
-| `language` | The language of the narration instructions based on the [IETF BCP 47](https://tools.ietf.org/html/bcp47) language tag string. If no language is specified or the specified language is unsupported, United States-based English (en-US) is used. [Currently supported language list](#supported-language-tags) |
+| `language` | The language of the narration instructions based on the [IETF BCP 47](https://tools.ietf.org/html/bcp47) language tag string. See the [8urrently supported language list](#supported-language-tags). If no language is specified, the HTTP API will look at the [`Accept-Language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) header and attempt to select a supported language using the q-factor weighting in the header. If no header is present, or the specified language(s) are unsupported or invalid, `en-US` is used as the fallback. |
```
CommanderStorm, line 306:
```suggestion
| `language` | The language of the narration instructions based on the [IETF BCP 47](https://tools.ietf.org/html/bcp47) language tag string. See the [currently supported language list](#supported-language-tags). If no language is specified, the HTTP API will look at the [`Accept-Language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) header and attempt to select a supported language using the q-factor weighting in the header. If no header is present, or the specified language(s) are unsupported or invalid, `en-US` is used as the fallback. |
```

src/worker.cc
```diff
@@ -1414,6 +1470,13 @@ void ParseApi(const http_request_t& request, valhalla::Api& api) {
     document.AddMember({kv.first, allocator}, array, allocator);
   }
 
+  if (!document.HasMember("language")) {
```
kevinkreiser, line 1473: this only checks the json, you need to also check if `options.language` is unset from pbf input

kevinkreiser: actually looks like it short circuits up above so you'll need to maybe regoranize this a bit more to support pbf properly

ianthetechie: d'oh.... I forgot about the PBF logic short circuiting; it looks like JSON and query string are together though so I *think* it handles those two.

ianthetechie: I think the latest commits should handle this. Though I'm not quite sure the best way to write a unit test for it. Can't say I've ever sent an HTTP API request as PBF, though I probably should try that out at some point.

src/worker.cc
```diff
@@ -1341,6 +1342,62 @@ std::string serialize_error(const valhalla_exception_t& exception, Api& request)
   return body.str();
 }
 
+// Regex for matching components of an Accept-Language header.
+const std::regex language_re(R"(([a-zA-Z0-9\-]{2,}|\*)(?:;q=(\d(?:\.\d+)?))?)");
+
+std::string ParseAcceptLanguage(headers_t headers) {
```
kevinkreiser, line 1348:
```suggestion
std::string ParseAcceptLanguage(const headers_t& headers) {
```

test/parse_request.cc
```diff
@@ -1641,6 +1650,21 @@ void test_closure_factor_parsing(const Costing::Type costing_type,
   validate(key, expected_value, options.closure_factor());
 }
 
+void test_language_parsing(const std::string expected_value,
+                           const std::string language_arg,
+                           const std::string language_header) {
```
kevinkreiser, line 1655:
```suggestion
void test_language_parsing(const std::string& expected_value,
                           const std::string& language_arg,
                           const std::string& language_header) {
```

src/worker.cc
kevinkreiser: the approach you are taking is just fine imho. a couple things to try:

1. the service expects the top level object, so you'll want instead of `options` an `api` object with an `options` object nested inside of it. its probably erroring because its using an empty `api` object
2. im not sure if only `costing_type` is needed. i think you might need to fill out the costings map where the one you care about needs to be at the mapped to the enum int that corresponds with bus costing. but lets hope you only need step one

ianthetechie: That's just a quirk of the text format. The message type is specified on the command line for `protoc` already (`valhalla.Api`).

If I try adding a wrapper, `protoc` throws an error: `input:1:5: Message type "valhalla.Api" has no field named "api".` So it sure thinks it's constructing an API message.

But.... crap :P I think I found the bug (the request format was fine). `from_json` is nuking the `Api` fields any time the document parameter is non-empty 😅 So my method of option merging with shared logic is sorta to blame, but...

1. I'm pretty sure that's unnecessary. We already call `api.Clear()`just a few lines before trying to parse the protobuf body.
2. The only other usage of this method is the regular JSON options parsing path.
3. I think it's cleaner to only have this logic once (the `with_accept_language` closure). I can rewrite this if you'd prefer another approach though.

I'll do some more testing now and report the results.

ianthetechie: Base message (just a handy request):

```
options {
  action: trace_route
  costing_type: bus
  shape: [
    { ll: { lat: 42.335447, lng: -71.086526 }},
    { ll: { lat: 42.335694, lng: -71.087059 }},
    { ll: { lat: 42.336708, lng: -71.088014 }},
    { ll: { lat: 42.336968, lng: -71.088246 }},
    { ll: { lat: 42.336737, lng: -71.088636 }},
    { ll: { lat: 42.336389, lng: -71.089048 }}
  ]
  format: pbf
}
```

How to test with `curl`:

```shell
curl -XPOST --location 'http://127.0.0.1:8002/trace_route' --header 'Content-Type: application/x-protobuf' --data-binary "@$HOME/v.bin" | protoc --decode valhalla.Api api.proto
```

Test 1: Response comes back in English as expected.

Test 2: Add a language string within the `options` protobuf object like `language: "et"` and `language: "de"`. The responses come back in the expected language.

Test 3: Set the `Accept-Language` header. As expected, this has no effect if a language is explicitly set in the options body.

Test 4: Set the `Accept-Language` header a few different ways (single value, multiple values, unsupported languages, etc.). Seems to be working as expected!

ianthetechie:
Just merged in master; quick bump on this as it's a pretty small PR and I think it'd make the default experience better for a lot of people :)

src/worker.cc
```diff
@@ -1358,6 +1356,62 @@ std::string serialize_error(const valhalla_exception_t& exception, Api& request)
   return body.str();
 }
 
+// Regex for matching components of an Accept-Language header.
+const std::regex language_re(R"(([a-zA-Z0-9\-]{2,}|\*)(?:;q=(\d(?:\.\d+)?))?)");
+
+std::string ParseAcceptLanguage(const headers_t& headers) {
+  auto accept_language_header = headers.find("Accept-Language");
+  if (accept_language_header != headers.end()) {
```
nilsnolde, line 1364: very nitty: 
```suggestion
  if (accept_language_header == headers.end()) {
    return "";
  }
  // rest of the code
```

just for readability

src/worker.cc
```diff
@@ -1358,6 +1356,62 @@ std::string serialize_error(const valhalla_exception_t& exception, Api& request)
...
+    std::sregex_token_iterator it(accept_language_header->second.begin(),
+                                  accept_language_header->second.end(), language_re, {1, 2});
+    std::sregex_token_iterator end;
+
+    // Iterate through weighted matches
+    bool found_usable_language = false;
+    // Pre-allocate; ternary necessary because with no matches, length will segfault
+    std::vector<std::pair<float, std::string>> languages(it != end ? it->length() : 0);
+    while (it != end) {
+      std::string language = *it++;
```
nilsnolde, line 1376: another very nit: it might actually be worth it to pre-increment on the previous line (to avoid the string copying of post-increment), and add a small comment that we (might) increment this iterator twice per loop for lang & q

src/worker.cc
```diff
@@ -1358,6 +1356,62 @@ std::string serialize_error(const valhalla_exception_t& exception, Api& request)
...
+      // of a supported language, but this could be considered in the future.
+      if (odin::get_locales().find(language) != odin::get_locales().end()) {
+        // If the quality is 1 (max) and we don't have any usable matches yet,
+        // we can skip further checks
+        if (q_value == 1.0 && !found_usable_language) {
+          return language;
+        }
+        found_usable_language = true;
+
+        languages.push_back(std::make_pair(q_value, language));
```
nilsnolde, line 1396: another possible optimization: can't we just do smth like
```suggestion
    // outside the loop
    auto lang = std::make_pair(0.0f, "");
        // in the loop
        if (q_value > lang.first) {
          lang = std::make_pair(q_value, language);
        }
    // after the loop
    if (found_usable_language)
      return lang.second;
```

src/worker.cc
```diff
@@ -1497,6 +1551,14 @@ void ParseApi(const http_request_t& request, valhalla::Api& api) {
     throw valhalla_exception_t{101};
   };
 
+  // Sets the request document language parameter from the Accept-Language header
+  auto with_accept_language = [request](rapidjson::Document& document) {
+    auto language = ParseAcceptLanguage(request.headers);
+    auto& allocator = document.GetAllocator();
+
+    document.AddMember({"language", allocator}, {language, allocator}, allocator);
```
nilsnolde, line 1559: rapidjson has an awful api.. `rapidjson::SetValueByPointer(doc, "/language", language)` is a little nicer to read IMO

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@
    * ADDED: include level change info in `/route` response [#4942](https://github.com/valhalla/valhalla/pull/4942)
    * ADDED: steps maneuver improvements [#4960](https://github.com/valhalla/valhalla/pull/4960)
    * ADDED: instruction improvements for node-based elevators [#4988](https://github.com/valhalla/valhalla/pull/4988)
+   * ADDED: Support the HTTP Accept-Language header as a fallback for instruction generation [#4965](https://github.com/valhalla/valhalla/pull/4965)
```
nilsnolde, line 64: could you move this to the end of the list again?

nilsnolde (approved):
very nice PR, sorry for the long wait on it! feel free to skip my nits:)

nilsnolde:
@ianthetechie should we also set the [`Content-Language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Language) response header appropriately? I'm not sure if that's easy enough at this point, might be that response headers are set somewhere totally different..

nilsnolde:
with https://github.com/valhalla/web-app/pull/286 just merged, I got reminded of this one. are you still interested here @ianthetechie?

ianthetechie:
Yeah lemme see if I can clean this up in a week or two (Jan is kinda crazy with family schedule). Feel free to just push any commits yourself if it's just minor stuff to get this over the line; you're probably faster at C++ than me :)

PR #4973, opened 2024-11-14
Per level elevator penalty for node based elevators

chrstnbwnkl:
# Issue

Elevator penalties introduced in #3509 are applied to transitions via elevator type nodes, without taking into consideration the number of traversed levels. With support for levels on edge info, we can now apply a penalty per level traversed. 

The down side is that levels are stored in edge info, which is usually not something available in costing, since it

   1. has to be read from a location in the tile that is likely not in memory during path finding and 
   2. it has to be parsed first.

 As discussed with @kevinkreiser though, elevator nodes should be rare and unlikely to occur somewhere in the middle of a route, so in the few cases where these are encountered, it may be fine to access that information. 

If I'm not mistaken, in the calling contexts we don't have the predecessor tile, so my solution is to pass a reference to the graph reader and get the tile through the predecessor's edge id, but I'm open to suggestions on how to improve this :smile:  

## Tasklist

 - [x] Add tests


src/sif/pedestriancost.cc
```diff
@@ -744,16 +754,32 @@ Cost PedestrianCost::EdgeCost(const baldr::DirectedEdge* edge,
 // Returns the time (in seconds) to make the transition from the predecessor
 Cost PedestrianCost::TransitionCost(const baldr::DirectedEdge* edge,
                                     const baldr::NodeInfo* node,
-                                    const EdgeLabel& pred) const {
+                                    const EdgeLabel& pred,
+                                    const graph_tile_ptr& tile,
+                                    baldr::GraphReader& reader) const {
```
kevinkreiser, line 759: really unfortunate to give such power to the costing function, one could do a lot of damage having access to the whole graph inside of costing. ill give a bit of thought if there is another way to get the info we need without passing the graphreader

src/thor/astar_bss.cc
```diff
@@ -156,7 +156,9 @@ void AStarBSSAlgorithm::ExpandForward(GraphReader& graphreader,
 
     auto edge_cost = current_costing->EdgeCost(directededge, tile);
     Cost normalized_edge_cost = {edge_cost.cost * current_costing->GetModeFactor(), edge_cost.secs};
-    auto transition_cost = current_costing->TransitionCost(directededge, nodeinfo, pred);
+    auto reader_getter = [&]() { return baldr::LimitedGraphReader(graphreader); };
```
kevinkreiser, line 159: can we limit the scope of capture to just the reader?
```suggestion
    auto reader_getter = [&graphreader]() { return baldr::LimitedGraphReader(graphreader); };
```

kevinkreiser (approved):

kevinkreiser (approved):

PR #4981, opened 2024-11-20
Fix valhalla_benchmark-loki

dnesbitt61:
Add a default value to reach. Let me know if this value is OK (1000) as a default.
Construct PointLL correctly (lon, lat)!


kevinkreiser:
the default in config is probably the one to use: https://github.com/valhalla/valhalla/blob/61949e92c72037ac2cfd59f458cbd6b199ed8ff0/scripts/valhalla_build_config#L93

src/valhalla_benchmark_loki.cc
```diff
@@ -188,7 +188,7 @@ int main(int argc, char** argv) {
         throw std::runtime_error("Latitude must be in the range [-90, 90] degrees");
       }
       float lon = valhalla::midgard::circular_range_clamp<float>(std::stof(parts[1]), -180, 180);
-      valhalla::midgard::PointLL ll(lat, lon);
+      valhalla::midgard::PointLL ll(lon, lat);
```
kevinkreiser, line 191: every time!

kevinkreiser (approved):
pre-emptive ship, probably should use a reach of 50 so it matches what we use for most requests

kevinkreiser (approved):

PR #4994, opened 2024-11-25
Fix macOS CI – homebrew links two versions of pkg-config

chrstnbwnkl:
Seems like the issue is fixed upstream but hasn't been deployed yet.


chrstnbwnkl:
oops wrong branch

PR #4962, opened 2024-11-06
docs: fix the outputs of the matrix service

CommanderStorm:
# Issue

While testing the docs, I noticed this formatting oversight for the matrix API while building a rust client.
(given that this seems like just a docs typo fix, I am going to not spam you with this. Hope this is okay..)

## Tasklist

 - [x] Update the docs with any new request parameters or changes to behavior described

## Requirements / Relations

Related to https://github.com/valhalla/valhalla/pull/4957 in that it also touches the docs

PR #5000, opened 2024-11-28
fix!: a typo in the matrix APIs response: `algorithm: "timedistancbssematrix"`

### LINKED ISSUE #4999: algorithms of `timedistancebssmatrix` are typo-ed
Looks like there is a typo, but it is also present in the actual response sent.

https://github.com/valhalla/valhalla/blob/e5e695f5209440aabea02004fa3999f49ca154de/src/proto_conversions.cc#L13

The actual response likely should be
```diff
- {valhalla::Matrix::TimeDistanceBSSMatrix, "timedistancbssematrix"}, 
+ {valhalla::Matrix::TimeDistanceBSSMatrix, "timedistancebssmatrix"},
```

_Originally posted by @M393 in https://github.com/valhalla/valhalla/pull/4962#discussion_r1856522691_

CommanderStorm:
Resolves #4999

Pulled out since this might be a breaking change => might be worth to tag as such and put on the sidelines

## Tasklist

 - [x] Add tests
        ^- not currently documented, but if the update to the matrix api output is merged beforehand, that needs to change: https://github.com/valhalla/valhalla/pull/4962

## Requirements / Relations

- will need to adjust https://github.com/valhalla/valhalla/pull/4962 (or this PR)


CommanderStorm:
I have not added a Changelog entry as I somewhat expect this to get pushed a few releases until the next MSRV major.
For the changelog:

> BREAKING: Fixed a typo in the matrix-APIs responses `algorithm` field: `timedistancbssematrix` is now `timedistancebssmatrix`.

kevinkreiser:
yeah im not sure if we should do this it is a breaking change. i know that @xlqian is the primary author and user of this so i guess if he is ok then we can make the change. what do you say @xlqian ?

kevinkreiser (approved):

PR #5002, opened 2024-12-02
More CI fixes

chrstnbwnkl:
- remove macOS hack from the other day again
- update circleci workflows to use their own recommended base images

PR #5005, opened 2024-12-02
fix: Double free crash during tiles build inside libxml2 on concurrent `spatialite_cleanup_ex()` calls

kinkard:
# Issue

Fixes https://github.com/valhalla/valhalla/issues/4904

Sadly, [`spatialite_cleanup_ex()` calls](https://www.gaia-gis.it/fossil/libspatialite/file?name=src/spatialite/spatialite.c) ([via `free_internal_cache()`](https://www.gaia-gis.it/fossil/libspatialite/file?name=src/connection_cache/alloc_cache.c)) the `xmlCleanupParser()`, which is [not thread-safe](https://github.com/GNOME/libxml2/blob/3b38f19b406951a5eedf71776d3864d7f2e03265/encoding.c#L700-L720).

This PR introduces a static mutex that guarantees that all `spatialite_cleanup_ex()` are called sequentially.

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
    * FIXED: `incremental_build_tiles` script works again [#4909](https://github.com/valhalla/valhalla/pull/4909)
    * FIXED: Fix ability to use Valhalla via cmake `add_subdirectory` [#4930](https://github.com/valhalla/valhalla/pull/4930)
    * FIXED: Fix valhalla_benchmark_loki benchmark application. [#4981](https://github.com/valhalla/valhalla/pull/4981)
+   * FIXED: Double free crash during tiles build inside libxml2 on concurrent `spatialite_cleanup_ex()` calls [#5005](https://github.com/valhalla/valhalla/pull/5005)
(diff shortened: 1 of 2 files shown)


kevinkreiser (dismissed):

kinkard:
Created a `libspatialite` ticket for this issue - https://www.gaia-gis.it/fossil/libspatialite/tktview/855ef62a68b9ac6e500b54883707b2876c390c01

kinkard:
Probably a better solution is to use thread local connections to sqlite and spatielite to avoid libxml inner state corruption. What do you think?

chrstnbwnkl:
I just tested your changes in conjunction with the circleCI runner updates and all checks pass (#5007), so I'd say this is good to merge. Don't worry about the test fails here, those are addressed in #5002, so IMO we can merge this one first. Only nitpick: maybe add a comment inside `make_spatialite_cache` linking the spatialite ticket or your analysis of the root cause over in #4904.

PR #5010, opened 2024-12-05
Customizable hierarchy limits 

### LINKED ISSUE #4088: Make Hierarchy Limits Configurable
We have holes in our test coverage because we cant make algorithms use parts of the graph based on these thresholds. If we make them configurable via vahalla.json then we can increase our test coverage. Also it makes it much easier for people to experiment with the routing by changing the limit (several times in the past we had to tweak it to get sane routes so this would make that easier)

chrstnbwnkl:
# Issue

Fixes #4088. 

This PR makes hierarchy limits customizable via the config JSON as well as via request parameters. Right now, hierarchy limits are hard coded. This PR proposes two ways to customize these values: 

1. server-side: defaults are set per algorithm inside the "thor" section (currently only `costmatrix`, `unidirectional_astar` and `bidirectional_astar` use hierarchy limits). 
2. client-side via request parameters: hierarchy limits can be set in the `costing_options`. In order to not break backwards compatibility, there is a new section `hierarchy_limits` in the `service_limits` server side config that has a `allow_modification` flag, which is false by default. Also, `max_allowed_up_transitions` and `max_expand_within_distance` can be set to clamp parameters passed via request params. 

~~The general idea is that user specified hierarchy limits are set in `DynamicCost`. If these are not set, each algorithm uses its own defaults passed down from the thor worker. In order to avoid copying back and forth from/to protobuf, I decided to remove the `HierarchyLimits` class and use the proto message format instead.~~
The general idea is to keep hierarchy limits in `DynamicCost` just like before, but have the thor worker do the clamping/overriding since it's got access to the relevant config options.

## Tasklist

 - [x] Add tests


proto/options.proto
```diff
@@ -68,6 +68,25 @@ message AvoidEdge {
...
+ * Hierarchy limits controls expansion and transitions between hierarchy
+ * levels. It allows limiting expansion on hierarchy levels once
+ * some number of "upward" transitions have been made (e.g. from the local
+ * level to the arterial level). Expansion on a level is allowed within some
+ * distance from the destination location. This also allows control of where
+ * upward and downward transitions are allowed based on distance from the
+ * destination.
+ */
+message HierarchyLimits {
+ uint32 up_transition_count = 1;   // Keep track of # of upward transitions
```
chrstnbwnkl, line 81: Using this for internal state is not ideal, I get that, but I don't want to have a wrapper class just for this attribute.

scripts/valhalla_build_config
```diff
@@ -118,11 +118,45 @@ config = {
         'max_reserved_labels_count_bidir_astar': 1000000,
         'max_reserved_labels_count_dijkstras': 4000000,
         'max_reserved_labels_count_bidir_dijkstras': 2000000,
-        'costmatrix_check_reverse_connection': False,
-        'costmatrix_allow_second_pass': False,
-        'max_reserved_locations_costmatrix': 25,
         'clear_reserved_memory': False,
         'extended_search': False,
+        'costmatrix': {
+            'check_reverse_connections': False,
```
chrstnbwnkl, line 124: we should have done this from the beginning... the bad news is that those have defaults in code, we should probably remove those so users are forced to update.

scripts/valhalla_build_config
```diff
@@ -118,11 +118,45 @@ config = {
...
+            'allow_second_pass': False,
+            'max_reserved_locations': 25,
+            'hierarchy_limits': {
+                'max_up_transitions': {
+                    '1': 400,
+                    '2': 100,
+                }, 
+            }
```
chrstnbwnkl, line 132: costmatrix does not use `expand_within_distance` right now

scripts/valhalla_build_config
```diff
@@ -118,11 +118,45 @@ config = {
...
+        },
+        'bidirectional_astar': {
+            'hierarchy_limits': {
+                'max_up_transitions':  {
+                    '1': 80,
+                    '2': 20,
+                } ,
```
chrstnbwnkl, line 139: no need to set  this for level 0, there are no up transitions from there.

scripts/valhalla_build_config
```diff
@@ -118,11 +118,45 @@ config = {
...
+            }
+        },
+        'unidirectional_astar': {
+            'hierarchy_limits': {
+                'max_up_transitions':  {
+                    '1': 80,
+                    '2': 20,
+                } ,
+                'expand_within_distance': {
+                    '0': 1e8,
```
chrstnbwnkl, line 154: = 100 thousand km, basically meaning "no limit" like it did before

src/loki/worker.cc
```diff
@@ -92,6 +93,8 @@ void loki_worker_t::parse_costing(Api& api, bool allow_none) {
     throw valhalla_exception_t{124};
   }
 
+  check_hierarchy_limits(api);
```
chrstnbwnkl, line 96: We parse costing options in loki, so I'd say we do the same here, even though hierarchy limits are only used in thor. 

The high level idea is: 
  - parse it off the json inside `ParseBaseCostingOptions` 
  - validate it inside loki worker's `parse_costing`, since the worker knows about the service limits (are users allowed to modify it, what are the max allowed values)
  - set it in `DynamicCost` only if there is anything on the protobuf options and expose a flag that tells algorithms whether there are customized hierarchy limits
  - in the  algorithms that use hierarchy limits, prefer user provided hierarchy limits, fall back to config

kevinkreiser: talked about this in a video call but basically to  really support all the different algorithms having their own service limits we'll need thor worker to parse these service limits and do the check there where it decides which algorithm will be used

src/sif/hierarchylimits.cc
```diff
@@ -2,6 +2,29 @@
 
 using namespace valhalla::sif;
 
-bool HierarchyLimits::StopExpanding(const float dist) const {
-  return (up_transition_count > max_up_transitions && dist > expansion_within_dist);
+namespace valhalla {
+namespace sif {
+
+bool StopExpanding(const valhalla::HierarchyLimits& hierarchy_limits, const float dist) {
```
chrstnbwnkl, line 8: make these inline

src/thor/astar_bss.cc
```diff
@@ -539,5 +539,7 @@ std::vector<PathInfo> AStarBSSAlgorithm::FormPath(baldr::GraphReader& graphreade
   return path;
 }
 
+void AStarBSSAlgorithm::RelaxHierarchyLimits() {
```
chrstnbwnkl, line 542: leave this in costing and have the worker administer the defaults

src/sif/dynamiccost.cc
```diff
@@ -261,18 +270,28 @@ float DynamicCost::GetModeFactor() {
 }
 
 // Gets the hierarchy limits.
-std::vector<HierarchyLimits>& DynamicCost::GetHierarchyLimits() {
+std::vector<HierarchyLimits> DynamicCost::GetHierarchyLimits() {
   return hierarchy_limits_;
 }
 
+// Gets mutable hierarchy limits.
```
chrstnbwnkl, line 277: This is now necessary, because the thor worker managing the hierarchy limits needs a copy of the originally passed hierarchy limits. Since we have different defaults for the different routing algorithms, and we may use them interchangeably for any given location pair, we always need two versions: one for bidirectional and one for timedep forward/reverse.

src/sif/dynamiccost.cc
```diff
@@ -401,6 +420,32 @@ void ParseBaseCostOptions(const rapidjson::Value& json,
   co->set_disable_hierarchy_pruning(
       rapidjson::get<bool>(json, "/disable_hierarchy_pruning", co->disable_hierarchy_pruning()));
 
+  // hierarchy limits
```
chrstnbwnkl, line 423: parsing it off the json: don't set anything on the proto if we can't find any hierarchy limits in the json

src/thor/bidirectional_astar.cc
```diff
@@ -1329,17 +1337,6 @@ std::vector<std::vector<PathInfo>> BidirectionalAStar::FormPath(GraphReader& gra
   return paths;
 }
 
-void BidirectionalAStar::ModifyHierarchyLimits() {
```
chrstnbwnkl, line 1332: since this is just a static division we just store those modified values as the defaults in the config and remove this

src/thor/costmatrix.cc
```diff
@@ -53,8 +53,8 @@ CostMatrix::CostMatrix(const boost::property_tree::ptree& config)
       max_reserved_labels_count_(config.get<uint32_t>("max_reserved_labels_count_bidir_dijkstras",
                                                       kInitialEdgeLabelCountBidirDijkstra)),
       max_reserved_locations_count_(
-          config.get<uint32_t>("max_reserved_locations_costmatrix", kMaxLocationReservation)),
-      check_reverse_connections_(config.get<bool>("costmatrix_check_reverse_connection", false)),
+          config.get<uint32_t>("costmatrix.max_reserved_locations", kMaxLocationReservation)),
```
chrstnbwnkl, line 56: I moved these from `thor` to the new `thor.costmatrix` section. Right now this would silently ignore values at the old location, which sucks. So the question is: do we warn? I would like to avoid crashing the service the way we do in service limits. I personally hate when that happens to me :smile:

chrstnbwnkl: Another thing we could do is release 3.5.2 and pin a warning at the top of the release notes.

src/thor/unidirectional_astar.cc
```diff
@@ -603,6 +611,9 @@ template <const ExpansionType expansion_direction, const bool FORWARD>
 void UnidirectionalAStar<expansion_direction, FORWARD>::ModifyHierarchyLimits(
     const float dist,
     const uint32_t /*density*/) {
+
+  if (!costing_->DefaultHierarchyLimits())
```
chrstnbwnkl, line 615: If the user passed in any custom hierarchy limits, don't do this distance based modification.

test/astar.cc
```diff
@@ -1261,6 +1272,7 @@ TEST(Astar, test_complex_restriction_short_path_fake) {
   vs::TravelMode mode;
   auto costs = vs::CostFactory().CreateModeCosting(options, mode);
   ASSERT_TRUE(bool(costs[int(mode)]));
+  set_hierarchy_limits(costs[int(mode)], true);
```
chrstnbwnkl, line 1275: some of these tests circumvent using the workers and call the algorithms directly, so set some default hierarchy limits in the costing.

valhalla/thor/costmatrix.h
```diff
@@ -333,20 +333,6 @@ class CostMatrix : public MatrixAlgorithm {
     return infos;
   };
 
-  void ModifyHierarchyLimits() {
-    // Distance threshold optimized for unidirectional search. For bidirectional case
```
chrstnbwnkl, line 337: `expand_within_dist` wasn't used anyway in CostMatrix, it was tested but the performance drain turned out to be too big.

src/worker.cc
```diff
@@ -1345,6 +1347,126 @@ void ParseApi(const std::string& request, Options::Action action, valhalla::Api&
   from_json(document, action, api);
 }
 
+hierarchy_limits_config_t
+parse_hierarchy_limits_from_config(const boost::property_tree::ptree& config,
```
chrstnbwnkl, line 1351: reads both the service limits as well as the default values from the config

src/worker.cc
```diff
@@ -1345,6 +1347,126 @@ void ParseApi(const std::string& request, Options::Action action, valhalla::Api&
...
+            ? default_max_up_transitions->get_value<uint32_t>(kDefaultMaxUpTransitions[it->level])
+            : kDefaultMaxUpTransitions[it->level]);
+
+    // if the algorithm uses distance to decide whether to expand a given level, set that property
+    // as well
+    if (uses_dist) {
+      auto default_expand_within_dist = config.get_child_optional(
+          "thor." + algorithm + ".hierarchy_limits.expand_within_distance." +
+          std::to_string(it->level));
+      found = found && (algorithm == "costmatrix" || default_expand_within_dist);
```
chrstnbwnkl, line 1404: the expand within distance is currently not used in costmatrix

src/worker.cc
```diff
@@ -1345,6 +1347,126 @@ void ParseApi(const std::string& request, Options::Action action, valhalla::Api&
...
+  }
+
+  if (!found)
+    LOG_WARN("Incomplete config for hierarchy limits found for " + algorithm +
+             ". Falling back to defaults");
+
+  return {max_hierarchy_limits, default_hierarchy_limits};
+};
+
+bool check_hierarchy_limits(std::vector<HierarchyLimits>& hierarchy_limits,
```
chrstnbwnkl, line 1420: this is where it all comes together: in thor's worker (matrix and route actions), this function is used to clamp/overwrite any user passed values based on the config.

src/worker.cc
```diff
@@ -1345,6 +1347,126 @@ void ParseApi(const std::string& request, Options::Action action, valhalla::Api&
...
+    }
+
+    if (limits.expand_within_dist() < 0 || // float might be negative
+        limits.expand_within_dist() > config.max_limits[i].expand_within_dist()) {
+      limits.set_expand_within_dist(config.max_limits[i].expand_within_dist());
+      add_warning = true;
+    }
+  }
+
+  cost->SetDefaultHierarchyLimits(default_limits);
```
chrstnbwnkl, line 1465: This flag is only needed because in unidirectional a*, the hierarchy limits are always modified based on the distance between origin and destination. This should only happen though if we are using default hierarchy limits (i.e. no user supplied hierarchy limits).

kevinkreiser (approved):

chrstnbwnkl:
I compared ~10k routes (mostly from `test_requests/random_routes.txt` and `timed_conditional_routes` to test unidirectional algos as well) using `run_with_server.py` between master and this feature branch. I was getting diffs, and it turns out I forgot to add the default values for `expand_within_dist` for bidirectional astar (they were previously modified within the algorithm, but now this happens in the worker, and only if neither user provided nor config defaults are present). 

So now I'm getting no diffs, and timing the commands gives me about 136 seconds for both master and this branch.

Just one question left we didn't address earlier: I moved two options into the new `costmatrix` section under `thor`: `costmatrix_check_forward_connections` and `costmatrix_allow_second_pass` are now nested under `costmatrix` (with the costmatrix suffix removed). This forces users to update their config, otherwise it will now silently fall back to the default in code. Not sure if we should do something more sensible here...

src/thor/route_action.cc
```diff
@@ -359,15 +359,51 @@ void thor_worker_t::path_arrive_by(Api& api, const std::string& costing) {
...
     // Get the algorithm type for this location pair
     thor::PathAlgorithm* path_algorithm =
         this->get_path_algorithm(costing, *origin, *destination, options);
     path_algorithm->Clear();
+
+    // once we know which algorithm will be used, set the hierarchy limits accordingly
+    bool is_bidir = path_algorithm == &bidir_astar;
+    auto& hierarchy_limits = is_bidir ? hierarchy_limits_bidir : hierarchy_limits_unidir;
+
+    // only check hierarchy limits if not already done for the current algorithm
```
chrstnbwnkl, line 392: only call `check_hierarchy_limits` once per algorithm on multi leg routes

kevinkreiser (approved):

src/worker.cc
```diff
@@ -1345,6 +1347,129 @@ void ParseApi(const std::string& request, Options::Action action, valhalla::Api&
...
+          default_expand_within_dist
+              ? default_expand_within_dist->get_value<float>(kDefaultExpansionWithinDist[it->level])
+          : is_bidir ? kDefaultExpansionWithinDistBidir[it->level]
+                     : kDefaultExpansionWithinDist[it->level]);
+    }
+    default_hierarchy_limits.push_back(default_hl);
+  }
+
+  if (!found)
+    LOG_WARN("Incomplete config for hierarchy limits found for " + algorithm +
```
kinkard, line 1417: @chrstnbwnkl This warning a bit confusing. The default values generated by `valhalla_build_config` tool causes this message even if no customisation was made by user.

Is it possible to move default values to `valhalla_build_config` to silence this warning in the default setup with minimal customization?

image

chrstnbwnkl: The default values are produced by `valhalla_build_config`, so if you have updated the config generation script as well, you shouldn't get this warning.

kinkard: Oh, my bad. I forgot to remove the script from the build folder and rerun cmake to get a new version of it...

Btw, thank you very much for this PR!

PR #5016, opened 2024-12-10
Store discretized bounding circles in graph id spare bits

chrstnbwnkl:
Work in progress


chrstnbwnkl:
Yep unfortunately the 18 spare bits in `GraphId` can not be used in a backwards compatible way, since the ids stored in the bins leak into the rest of the application, where they are compared by value to `GraphId`s that don't have the same information stored on those spare bits. Since we don't mask off those bits when comparing two graph ids, the equality operator would fail to identify two graph ids pointing to the same edge.  

In order to not let this effort go to waste, I'll open a new PR that aims to use the same mechanism (i.e. bounding circles) but stores the information in a separate section of the graphtile instead.

Issue #5018, opened 2024-12-15
Isochrone GeoTIFFs are vertically flipped when rendering in non-projected apps


saranrapjs:
When generating isochrones serialized as GeoTIFF using `valhalla_run_isochrone`, the GeoTIFF will display correctly in GeoTIFF-aware software like GDAL, but the north/south axis is flipped when displayed in other software like the Preview app on Macs, or other programs that can decode TIFF's but don't make use of coordinate reference system data stored in GeoTIFF.

### Example isochrone

Here's how I'm generating geotiffs, on my Mac:

```shell
valhalla_run_isochrone \
	-c ./config.json \
	-f ./auto.tif \
	--json '{"format":"geotiff","locations":[{"lat":40.74,"lon":-73.91}],"costing":"auto","contours":[{"time":120,"color":"ff0000"}]}'
```

...which produces a GeoTIFF file that will display fine in qGIS or similar GeoTIFF software. But in the MacOS preview app, or in other software that decodes tiff files but doesn't reproject pixel information using geotiff headers, the GeoTIFF looks like this:

<img width="1053" alt="Screenshot 2024-12-14 at 9 11 38 PM" src="https://github.com/user-attachments/assets/375a4486-fdf6-47da-b555-d88ab950013b" />

(this is an isochrone from New York City, and you can see that the vertical axis is flipped)

PR #5019, opened 2024-12-15
Make isochrone geotiff serialization use "north up" geotransform

saranrapjs:
# Issue

The isochrone GeoTIFF serialization in Valhalla is really cool! I'm building an app that uses GeoTIFF data _without_ using gdal (or similar) libraries to post-process the data — and I noted, when playing around with this, that the vertical orientation for Valhalla isochrones is flipped ([more detail in this issue](https://github.com/valhalla/valhalla/issues/5018)).

To get the output TIFF rendering with the correct axis, on this branch I ended up:
- Making the last element of the Geotransform negative — I looked at the serialization code, and noted in the [gdal docs for "Geotransform"](https://gdal.org/en/stable/tutorials/geotransforms_tut.html) that north-south imagery expects to have a negative value for the last number in the Geotransform, which I think matches the data produced by the Isochrone tiles. 
- Assembling the "y" data passed into the gdal raster method from bottom-to-top, rather than top-to-bottom, which seemed to be required in order to actually render the raster data in the right projected location ([based on this stack exchange post](https://gis.stackexchange.com/questions/114978/raster-created-by-python-gdal-interpreted-upside-down))
- Using the bounding box's "maximum" column and "maximum Y" as the 5th element in the Geotransform, which also seemed to be required in order to project the raster in the right place.

I'm not totally confident that my understanding of the min/max for the Isochrone tiles, or tile extents, is correct here — e.g. it's possible this logic might not be safe across the meridian or the equator? — but I figured I'd get a PR going and make sure I'm on the right track and that this issue is worth solving this way before getting deeper into the coordinate math. I'm also extremely green to C++, and very happy for this code (or the unit test) to work differently!

On this branch, I'm finding that the Isochrone described in the linked issue correctly renders with north-south vertical orientation (and still projects in the correct place within software like qgis);

Screenshot 2024-12-14 at 9 22 47 PM

This fixes https://github.com/valhalla/valhalla/issues/5018

## Tasklist

 - [x] Add tests

### DIFF
--- a/test/isochrone.cc
+++ b/test/isochrone.cc
@@ -511,24 +511,7 @@ TEST(Isochrones, test_geotiff_vertical_orientation) {
   std::copy(geotiff.cbegin(), geotiff.cend(), buffer);
   auto handle = VSIFileFromMemBuffer(name.c_str(), buffer, static_cast<int>(geotiff.size()), 0);
   auto geotiff_dataset = GDALDataset::FromHandle(GDALOpen(name.c_str(), GA_ReadOnly));
-  int x = geotiff_dataset->GetRasterXSize();
   int y = geotiff_dataset->GetRasterYSize();
-  GDALRasterBand* band = geotiff_dataset->GetRasterBand(1);
-  uint16_t data_array[x * y];
-  CPLErr err = band->RasterIO(GF_Read, 0, 0, x, y, data_array, x, y, GDT_UInt16, 0, 0);
-  double min_max[2];
-
-  band->ComputeRasterMinMax(0, min_max);
-
-  ASSERT_EQ(err, CE_None);
-  ASSERT_NE(x, 0);
-  ASSERT_NE(y, 0);
-  ASSERT_EQ(static_cast<int>(min_max[0]), 0);
-  ASSERT_EQ(static_cast<int>(min_max[1]), 1100);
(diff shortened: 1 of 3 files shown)


test/isochrone.cc
```diff
@@ -495,6 +495,48 @@ TEST(Isochrones, test_geotiff_output_time_distance) {
...
+  ASSERT_EQ(static_cast<int>(min_max[1]), 1100);
+  ASSERT_EQ(band->GetNoDataValue(), std::numeric_limits<uint16_t>::max());
+
+  check_raster_edges(x, y, data_array);
+
+  double geoTransform[6];
+  geotiff_dataset->GetGeoTransform(geoTransform);
+  double topY = geoTransform[3] + 0 * geoTransform[4] + 0 * geoTransform[5];
+  double bottomY = geoTransform[3] + 0 * geoTransform[4] + y * geoTransform[5];
+  ASSERT_TRUE(topY > bottomY);
```
kevinkreiser, line 536: would it be possible to shorten the test to really just this assertion?

chrstnbwnkl (approved):
Seems like I skipped that line in the GDAL docs when I was working on this :smile: thanks!

chrstnbwnkl:
One more thing though: would you mind adding a changelog entry?

kevinkreiser (approved):

PR #5023, opened 2024-12-16
Tag smoothness=impassable breaks pedestrian routing

### LINKED ISSUE #4967: Tag smoothness=impassable breaks pedestrian routing
The smoothness tag shouldn't impact pedestrian routing, as this tag is meaningful only for vehicles (https://wiki.openstreetmap.org/wiki/Key:smoothness). During graph.lua processing, ways with smoothness=impassable get kv["pedestrian_forward"] = "false" and kv["pedestrian_backward"] = "false". This issue started after https://github.com/valhalla/valhalla/commit/329c4e62428a249c9b59a674baf1241e6bb8ef8d commit. 

For example: https://www.openstreetmap.org/directions?engine=fossgis_valhalla_foot&route=52.43489%2C14.82347%3B52.43359%2C14.82847

gknisely:
# Issue

Smoothness tag should not delete the edge in general.  The smoothness tag is for the physical usability of a way for wheeled vehicles as defined in the [wiki](https://wiki.openstreetmap.org/wiki/Key:smoothness) and pedestrian access should not be removed.  

## Tasklist

 - [x] Updated tests
 - [x] Add fixes #4967

## Requirements / Relations

### DIFF
--- a/test/gurka/test_exclude_unpaved_smoothness.cc
+++ b/test/gurka/test_exclude_unpaved_smoothness.cc
@@ -205,7 +205,7 @@ TEST(Standalone, SmoothnessAccess) {
     auto node_id = gurka::findNode(*graph_reader, map.nodes, "1");
     const auto* node = graph_reader->nodeinfo(node_id);
     // ped only access due to smoothness = impassable
-    EXPECT_EQ(node->access(),baldr::kPedestrianAccess);
+    EXPECT_EQ(node->access(), baldr::kPedestrianAccess);
   }
 }
 
--- a/test/gurka/test_exclude_unpaved_smoothness.cc
+++ b/test/gurka/test_exclude_unpaved_smoothness.cc
@@ -182,10 +182,10 @@ TEST(Standalone, SmoothnessAccess) {
     // ped only access due to smoothness = impassable.
     // edge_1 = AB
     // edge_2 = BA
-    EXPECT_EQ(edge_1->forwardaccess(), (baldr::kWheelchairAccess | baldr::kPedestrianAccess));
-    EXPECT_EQ(edge_1->reverseaccess(), (baldr::kWheelchairAccess | baldr::kPedestrianAccess));
-    EXPECT_EQ(edge_2->forwardaccess(), (baldr::kWheelchairAccess | baldr::kPedestrianAccess));
-    EXPECT_EQ(edge_2->reverseaccess(), (baldr::kWheelchairAccess | baldr::kPedestrianAccess));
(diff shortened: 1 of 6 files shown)


test/gurka/test_exclude_unpaved_smoothness.cc
```diff
@@ -179,11 +179,16 @@ TEST(Standalone, SmoothnessAccess) {
     const DirectedEdge* edge_2 = nullptr;
 
     std::tie(edge_id_1, edge_1, edge_id_2, edge_2) = findEdge(*graph_reader, map.nodes, "AB", "B");
-    // no access due to smoothness = impassable and are therefore tossed.
+    // ped only access due to smoothness = impassable.
     // edge_1 = AB
     // edge_2 = BA
-    EXPECT_EQ(edge_1, nullptr);
-    EXPECT_EQ(edge_2, nullptr);
+    EXPECT_NE(edge_1->forwardaccess(), baldr::kPedestrianAccess);
```
chrstnbwnkl, line 185: Shouldn't this be `EXPECT_EQ`? What other access does it have?

gknisely: yes.  let me fix that.  Probably has wheelchair and pedestrian

gknisely: @chrstnbwnkl thank you for seeing that issue.  Test was all wrong.  Good to go now.

chrstnbwnkl (approved):

chrstnbwnkl:
Ah wait, it shouldn't have wheelchair access though right? :smile:

gknisely:
> Ah wait, it shouldn't have wheelchair access though right? 😄

Yes.  We use pedestrian access to assign wheelchair when we can do so.  https://github.com/valhalla/valhalla/blob/master/src/mjolnir/directededgebuilder.cc#L164-L165

chrstnbwnkl:
I guess it doesn't matter because wheelchair will still be disallowed due to the surface value? 

https://github.com/valhalla/valhalla/blob/a0c27bca9ea105c83d701ad8cf50b1c9210c8e05/src/sif/pedestriancost.cc#L590-L595

gknisely:
> I guess it doesn't matter because wheelchair will still be disallowed due to the surface value?
> 
> https://github.com/valhalla/valhalla/blob/a0c27bca9ea105c83d701ad8cf50b1c9210c8e05/src/sif/pedestriancost.cc#L590-L595

Nah.  I think it is better to not assign.  Give me a few

gknisely:
@chrstnbwnkl should be all set

PR #5026, opened 2024-12-21
Fix costing in loki benchmark script

chrstnbwnkl:
If `type` is not explicitly set, it will always be `Costing::Type::none`.

### DIFF
--- a/src/valhalla_benchmark_loki.cc
+++ b/src/valhalla_benchmark_loki.cc
@@ -71,13 +71,15 @@ using results_t = std::set<result_t>;
 
 valhalla::sif::cost_ptr_t create_costing() {
   valhalla::Options options;
-  valhalla::Costing::Type costing;
-  if (valhalla::Costing_Enum_Parse(costing_str, &costing)) {
-    options.set_costing_type(costing);
+  valhalla::Costing::Type costing_type;
+  if (valhalla::Costing_Enum_Parse(costing_str, &costing_type)) {
+    options.set_costing_type(costing_type);
   } else {
     options.set_costing_type(valhalla::Costing::none_);
   }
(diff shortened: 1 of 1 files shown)


src/valhalla_benchmark_loki.cc
```diff
@@ -71,13 +71,15 @@ using results_t = std::set<result_t>;
 
 valhalla::sif::cost_ptr_t create_costing() {
   valhalla::Options options;
-  valhalla::Costing::Type costing;
-  if (valhalla::Costing_Enum_Parse(costing_str, &costing)) {
-    options.set_costing_type(costing);
+  valhalla::Costing::Type costing_type;
```
chrstnbwnkl, line 74: just renaming

src/valhalla_benchmark_loki.cc
```diff
@@ -71,13 +71,15 @@ using results_t = std::set<result_t>;
...
+  if (valhalla::Costing_Enum_Parse(costing_str, &costing_type)) {
+    options.set_costing_type(costing_type);
   } else {
     options.set_costing_type(valhalla::Costing::none_);
   }
-  (*options.mutable_costings())[costing];
+  auto& co = (*options.mutable_costings())[costing_type];
+  co.set_type(costing_type);
```
chrstnbwnkl, line 81: apart from inserting costing into the `costings` map, we also need to set its type.

kevinkreiser:
`none` was in purpose if not specified iirc

chrstnbwnkl:
there's already a cli arg `costing` that defaults to auto but does not have any effect without these changes. Maybe change the default to none then?

kevinkreiser:
Yep, default to none makes sense to me. It must have happened when we got rid of boost for this other program opts library

Issue #4148, opened 2023-06-06
Enabling route finding on roads under construction


aez99:
I'm trying to plan a route within a part of the highway which seems to be labeled as road under construction in the map ([https://www.openstreetmap.org/#map=16/41.2321/28.8358](https://www.openstreetmap.org/#map=16/41.2321/28.8358))

The default options seem to completely disregard that part of the road when finding a route as shown in [https://route.getyourmap.com/demos/routing/#loc=17,41.229895,28.844582&locations=[{%22lat%22:41.22846687919765,%22lon%22:28.842911180108786},{%22lat%22:41.23120222113064,%22lon%22:28.843683656305075}]&costing=%22auto%22](https://route.getyourmap.com/demos/routing/#loc=17,41.229895,28.844582&locations=[{%22lat%22:41.22846687919765,%22lon%22:28.842911180108786},{%22lat%22:41.23120222113064,%22lon%22:28.843683656305075}]&costing=%22auto%22)

So I tried to turn off the flags related to closed roads when constructing a request:
{"locations":[{"lat":41.226734,"lon":28.842563,"search_filter":{"exclude_closures":false}},{"lat":41.231201,"lon":28.843670,"search_filter":{"exclude_closures":false}}],"costing":"auto","costing_options":{"auto":{"closure_factor":1.0}},"directions_options":{"units":"miles"}}

However, it didn't make any difference and it still avoids the road labelled as under construction. Is there way to force Valhalla to find a route through roads under construction?

nilsnolde:
Nope, not via a request. That's only possible if you're running Valhalla yourself, see https://github.com/gis-ops/valhalla/blob/731923446bfb32f6edfa4d24e1c428d1ef0067f3/scripts/valhalla_build_config#L50. 

`exclude_closures` is only relevant for live traffic.

aez99:
@nilsnolde Thank you for your response. This link doesn't exist, can you please re-send it? Also, yeah I'm running Valhalla myself using this image gisops/valhalla:latest

nilsnolde:
Oh sorry, that's our private fork. but you get the gist, it's the same location and the same line in this repository

aez99:
Found it thanks! One last thing, to re-build the graph, is it enough to restart the container? Tried that and it seems that the changes didn't take effect

eikes:
As far as I can tell, you currently cannot create routes on ways with a construction tag. You _can_ build the routing graph with the construction ways included but they always get excluded in the costing process.

PR #5030, opened 2025-01-02
Adds 'ignore_construction' flag to allow routing on construction ways

eikes:
# Issue

In the past the ability to "Include roads under construction into the graph" was added: #3455

Currently it is not possible to create routes on ways which have the construction tag though. There is an issue about that, so this PR fixes: https://github.com/valhalla/valhalla/issues/4148

The situation is, that you can include construction ways in the routing graph by enabling the `include_construction` flag in the configuration json before building the graph. You cannot use these ways for routing though as they are explicitly marked as inaccessible in both directions and there is an additional check in the dynamic costing.

This PR adds a flag which can be enabled on a request in the costing options to explicitly allow routing on construction ways. 

This might not be useful when doing ad-hoc routing, but when planning routes the current construction sites aren't of interest.

## Tasklist

 - [x] Add tests
 - [ ] ~If you made changes to the lua files, update the [taginfo](taginfo.json) too.~

## Requirements / Relations

- https://github.com/valhalla/valhalla/pull/3455
- https://github.com/valhalla/valhalla/issues/4148


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
    * ADDED: steps maneuver improvements [#4960](https://github.com/valhalla/valhalla/pull/4960)
    * ADDED: instruction improvements for node-based elevators [#4988](https://github.com/valhalla/valhalla/pull/4988)
    * ADDED: increased precision in route lengths [#5020](https://github.com/valhalla/valhalla/pull/5020)
+   * ADDED: `ignore_constructions` allows routing on ways with construction tag [#5030](https://github.com/valhalla/valhalla/pull/5030)
```
kevinkreiser, line 20: drop the plural everywhere please (sorry i know english is weird)

```suggestion
   * ADDED: `ignore_construction` allows routing on ways with construction tag [#5030](https://github.com/valhalla/valhalla/pull/5030)
```

proto/options.proto
```diff
@@ -322,6 +322,9 @@ message Costing {
     bool exclude_tolls = 89;
     bool exclude_highways = 90;
     bool exclude_ferries = 91;
+    oneof has_ignore_constructions {
+      bool ignore_constructions = 92;
+    }
```
kevinkreiser, line 327: default is false so you dont need `oneof`
```suggestion
    bool ignore_constructions = 92;
```

valhalla/sif/dynamiccost.h
```diff
@@ -345,7 +345,9 @@ class DynamicCost {
         ((disallow_mask & kDisallowEndRestriction) && edge->end_restriction()) ||
         ((disallow_mask & kDisallowSimpleRestriction) && edge->restrictions()) ||
         ((disallow_mask & kDisallowShortcut) && edge->is_shortcut());
-    return accessible && !assumed_restricted && (edge->use() != baldr::Use::kConstruction);
+    return accessible && !assumed_restricted &&
+           (edge->use() != baldr::Use::kConstruction ||
+            (ignore_constructions_ && edge->use() == baldr::Use::kConstruction));
```
kevinkreiser, line 350: i think you can simplify, if the first part is false we already know the use is construction
```suggestion
    return accessible && !assumed_restricted &&
           (edge->use() != baldr::Use::kConstruction || ignore_constructions_);
```

valhalla/sif/dynamiccost.h
```diff
@@ -359,10 +361,11 @@ class DynamicCost {
...
     // you dont care about the direction the mode has access to
-    return ((edge->forwardaccess() & access_mask_) ||
-            (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
-            (ignore_oneways_ && (edge->reverseaccess() & access_mask_))) &&
-           (edge->use() != baldr::Use::kConstruction);
+    return (((edge->forwardaccess() & access_mask_) ||
+             (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
+             (ignore_oneways_ && (edge->reverseaccess() & access_mask_))) &&
+            (edge->use() != baldr::Use::kConstruction)) ||
+           (ignore_constructions_ && edge->use() == baldr::Use::kConstruction);
```
kevinkreiser, line 368: this one is confusing we should update the comment :smile: could it be a bit simpler, something like:
```suggestion
    return (((edge->forwardaccess() & access_mask_) ||
             (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
             (ignore_oneways_ && (edge->reverseaccess() & access_mask_))) &&
           (ignore_constructions_ || edge->use() != baldr::Use::kConstruction);
```

CHANGELOG.md
eikes: Done!

proto/options.proto
eikes: Done!

valhalla/sif/dynamiccost.h
eikes: That is indeed simpler and still correct.

valhalla/sif/dynamiccost.h
eikes: That is not correct. I did try that and noticed that it didn't work anymore. The reason being, that the parentheses are not on the same level as in the IsAccessible function. If a way has the construction tag, then both forward and reverse access is set to false in when building the graph in graph.lua: https://github.com/valhalla/valhalla/blob/master/lua/graph.lua#L1873-L1908

So what I do here is allow access regardless of the access, but only if it is a construction way and we are ignoring construction.

test/gurka/gurka.cc
```diff
@@ -121,7 +121,14 @@ std::string build_valhalla_request(const std::vector<std::string>& location_type
 
   // we do this last so that options are additive/overwrite
   for (const auto& kv : options) {
-    rapidjson::Pointer(kv.first).Set(doc, kv.second);
+    // Check if the value is "true" or "false" and set as bool
+    if (kv.second == "true") {
+      rapidjson::Pointer(kv.first).Set(doc, true);
+    } else if (kv.second == "false") {
+      rapidjson::Pointer(kv.first).Set(doc, false);
```
eikes, line 128: This is a bit of a hack, but currently there is no way to enable boolean flags in the `do_action` function options.

test/gurka/test_construction.cc
```diff
@@ -149,7 +158,7 @@ TEST_P(ConstructionRoutingTest, CheckAvoidRoadsUnderConstruction) {
   gurka::assert::raw::expect_path(result, {"AB", "BE", "EF", "FC", "CD"});
 
   result = gurka::do_action(valhalla::Options::route, map, {"1", "2"}, costing,
-                            {{"/costing_options/" + costing + "/ignore_access", "1"}});
+                            {{"/costing_options/" + costing + "/ignore_access", "true"}});
```
eikes, line 161: This previously simply had no effect.

test/gurka/gurka.cc
chrstnbwnkl: It's a bit ugly but it should work using `"0"` or `"1"`

eikes: You are right! I removed it.

valhalla/sif/dynamiccost.h
```diff
@@ -359,10 +360,13 @@ class DynamicCost {
...
-            (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
-            (ignore_oneways_ && (edge->reverseaccess() & access_mask_))) &&
-           (edge->use() != baldr::Use::kConstruction);
+    // AND it is not a construction way
+    // OR if it is a construction and you ignore construction access restrictions
+    return ((((edge->forwardaccess() & access_mask_) ||
+              (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
+              (ignore_oneways_ && (edge->reverseaccess() & access_mask_))) &&
+             (edge->use() != baldr::Use::kConstruction)) ||
+            (ignore_construction_ && (edge->use() == baldr::Use::kConstruction)));
```
kevinkreiser, line 369: can we logically pair the two construction parts together and reduce some parentheses, this is very hard to read otherwise. here's how i would do it without asinine autoformatting:

```suggestion
// you have access to the edge
return 
(
  (edge->forwardaccess() & access_mask_) ||
  (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
  (ignore_oneways_ && (edge->reverseaccess() & access_mask_))
) &&
// AND you either dont care if its under construction or its not under construction
(
  ignore_construction_ || edge->use() != baldr::Use::kConstruction
);
```

eikes: As I pointed out before, that is not the same though. On a road under construction, this first part of the `AND` clause will always be false:

```
(
  (edge->forwardaccess() & access_mask_) ||
  (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
  (ignore_oneways_ && (edge->reverseaccess() & access_mask_))
)
```

If a way has the construction tag, then both `edge->forwardaccess()`  and `edge->reverseaccess()` are `false`. See here in the graph.lua: https://github.com/valhalla/valhalla/blob/master/lua/graph.lua#L1873-L1908

So what I do here is allow access regardless of  `edge->forwardaccess()`  and `edge->reverseaccess()`, but only if it is a construction way and if `ignore_construction` is set.

kevinkreiser: I mean I get what you're saying but it doesn't make sense when you look at what the code previously was. If construction turns off access then why did we even need to check it before? I guess it's possible the code before was just redundant? Also possible I'm just being an idiot :)

eikes: I didn't think about it that way before. You are right, the check previously was not doing anything as long as the graph.lua stays the same. If we can assume that the graph.lua stays the same then we can skip that check. I believe we can simplify the function to this:

```CPP
  inline virtual bool IsAccessible(const baldr::DirectedEdge* edge) const {
    // you can go on it if:
    // you have forward access for the mode you care about
    // you dont care about what mode has access so long as its forward
    // you dont care about the direction the mode has access to
    return
    // you have access to the edge
    (
      (edge->forwardaccess() & access_mask_) ||
      (ignore_access_ && (edge->forwardaccess() & baldr::kAllAccess)) ||
      (ignore_oneways_ && (edge->reverseaccess() & access_mask_))
    ) ||
    // OR it is under construction but you don't care about that
    (
      ignore_construction_ && edge->use() != baldr::Use::kConstruction
    );
  }
```

What do you think?

kevinkreiser: yeah when i saw the code it made me think that in graph.lua we would turn on access for the appropriate modes of transportation and then this line was saving us from routing on it but if you go over and read graph.lua you find a sound explanation: https://github.com/valhalla/valhalla/blob/b40b23a1b0671fec3648b50a2f3687a9a9701727/lua/graph.lua#L1873-L1877

the problem then is that we are kind of in a broken state because of backwards compatibility. if you want to route on something that is under construction you dont know whether its inteded for cars or pedestrians or bikes or whatever because the access is all turned off. still its the best we can do

eikes: I suppose the question is if in the future the access to construction roads might be turned on in graph.lua and if we want to safeguard against that case here already.

eis-ioki: So what's the verdict? I changed it so it doesn't include the redundant construction check anymore, instead it adds the ignore_construction check.

kevinkreiser: sorry i should have been less terse, this looks good to me. what you have is the best we can do without planning for breaking changes to the data compatibility

test/gurka/test_construction.cc
```diff
@@ -141,6 +141,16 @@ class ConstructionRoutingTest : public ::testing::TestWithParam<std::string> {
 
 gurka::map ConstructionRoutingTest::map = {};
 
+TEST_F(ConstructionRoutingTest, IgnoreConstructions) {
```
kevinkreiser, line 144:
```suggestion
TEST_F(ConstructionRoutingTest, IgnoreConstruction) {
```

kevinkreiser (dismissed):

kevinkreiser (approved):

eikes:
🚧 Hooray! 🚧

PR #5039, opened 2025-01-06
Test `valhalla_service.exe` in one shot mode on win build

chrstnbwnkl:
# Issue

If we don't run any tests on the windows build, we should at least do a smoke test for `valhalla_service.exe` which is more important to most users than the `valhalla_run_*` scripts. I also wonder if there's a specific reason none of the tests are run in the windows build?

### DIFF
--- a/test/win/valhalla.json
+++ b/test/win/valhalla.json
@@ -194,8 +194,8 @@
       "min_resample": 10.0
     },
     "trace": {
-      "max_best_paths": 4,
-      "max_best_paths_shape": 100,
+      "max_alternates": 4,
+      "max_alternates_shape": 100,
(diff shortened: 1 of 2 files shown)


test/win/valhalla.json
```diff
@@ -194,8 +194,8 @@
       "min_resample": 10.0
     },
     "trace": {
-      "max_best_paths": 4,
-      "max_best_paths_shape": 100,
+      "max_alternates": 4,
```
chrstnbwnkl, line 197: this file hasn't been kept up to date for a long while it seems... there are probably more entries missing but these changes were necessary for the smoke test to work

PR #5042, opened 2025-01-07
Exclude executables from coverage

chrstnbwnkl:
# Issue

`fastcov` uses substring matching to exclude files, so passing `src/valhalla_` and `src/mjolnir/valhalla_` should do the trick. Tested locally and worked as expected.


kevinkreiser:
and yet coverage is reduced somehow?

chrstnbwnkl:
What hahah how? Zu früh gefreut

kevinkreiser:
scheint so ja.. when i look at the report it still says there are tracked lines in those files: https://app.codecov.io/gh/valhalla/valhalla/tree/cb-coverage/src so maybe we just have the flag to do the exclusions wrong? i saw you put multiple on the same argument maybe you have to specify the flag multiple times, once for each exclude?

CMakeLists.txt
```diff
@@ -360,7 +360,7 @@ if(ENABLE_COVERAGE)
 
   set(FASTCOV_PATH ${VALHALLA_SOURCE_DIR}/third_party/fastcov/fastcov.py)
   add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/coverage.info
-    COMMAND ${FASTCOV_PATH} -d . --exclude /usr/ third_party/ ${CMAKE_CURRENT_BINARY_DIR}/ --lcov -o coverage.info
+    COMMAND ${FASTCOV_PATH} -d . --exclude /usr/ third_party/ src/valhalla_ src/mjolnir/valhalla_ ${CMAKE_CURRENT_BINARY_DIR}/ --lcov -o coverage.info
```
kevinkreiser, line 363: maybe for things that arent directories you need wildcards or something..

kevinkreiser:
i noticed the coverage still doesnt quite agree so there is still more to update here

Issue #2050, opened 2019-11-12
Unparsed conditional access restriction


purew:
I came across a conditional access restriction in OSM that Valhalla does not parse:

```
<tag k="access:conditional" v="no @ (Mo-Sa 07:00-16:00; PH off)"/>
```
https://www.openstreetmap.org/edit?way=212430475#map=18/41.90549/12.50074

Valhalla parses the more specific access restrictions by transportation mode. But the transportation mode is actually optional: https://wiki.openstreetmap.org/wiki/Conditional_restrictions#Tagging

 For example, tweaking the above key to `motor_vehicle:conditional` parses fine.

PR #5048, opened 2025-01-10
Support `"access:conditional"` conditional restrictions like `"access:conditional"="no @ (Oct-May)"`

kinkard:
# Issue

fixes https://github.com/valhalla/valhalla/issues/2050

There are plenty of roads that are closed during some period of time and that are tagged by `"access:conditional". For example, this road in Alps is closed from October to May - https://www.openstreetmap.org/way/1012815911.

See more using use https://overpass-turbo.eu/

```
[out:json][timeout:25];
(
	way["access:conditional"](45.0, 5.0, 53.0, 15.0);
);
out body;
>;
out skel qt;
```

image

## Tasklist

 - [x] Add tests

## Requirements / Relations


src/mjolnir/timeparsing.cc
```diff
@@ -231,6 +231,9 @@ std::vector<uint64_t> get_time_range(const std::string& str) {
     // rm white space at both ends
     boost::algorithm::trim(condition);
 
+    boost::algorithm::replace_all(condition, "winter", "Dec-Feb");
+    boost::algorithm::replace_all(condition, "summer", "Jun-Aug");
```
kevinkreiser, line 235: and if its the southern hemisphere we need to flip them 😄 i guess we dont have an easy way we could, once we know the latitude, make this choice... we'd have to store some other temp info about how we assumed northern hemisphere and how we need to shift the time by 6 months if it happened to be in the southern hemisphere? sounds like a pita but getting it wrong isnt really an option either right? or am i missing something?

kinkard: 🤦 Now I understand why Valhalla didn't support this in the first place.

Let me remove this change and keep only the `"access:conditional"`.

kevinkreiser: yeah its sad but it could be done, just a bit more involved passing state from at least the way/relation tagging until we get to the node tagging (where we finally know the lat lng). totally doable imho but def more involved (have to make storage for it)

kinkard: Actually, now I'm thinkinking that seasons support is quite orthogonal to the tag support.

And for seasons, AFAIK, Valhalla aways processes ways after nodes, so a conditional substitution ("summer"/"winter" -> month range) somewhere in the pbfgraphparser.cc should do the job.

kevinkreiser: i was pretty sure we did ways, relations, then finally nodes. the reason to do this is because there are lots of ways and relations (with even more nodes) that we dont care about routing on at all and so by first filtering down the ways and relations to what we care about we can skip tons and tons of nodes that we wouldnt need to parse and then just throw away.

but i guess it doesnt matter that i dont understand what you're thinking :smile: if you think you can get the seasons straightened out please by all means do so!

kinkard: Ah, you're right. Nodes get resolved after ways and relations =\
```cpp
enum class BuildStage : int8_t {
  kInvalid = -1,
  kInitialize = 0,
  kParseWays = 1,
  kParseRelations = 2,
  kParseNodes = 3,
  kConstructEdges = 4,
  //...
  };
```

Well, then I have no idea how to do it without introducing additional complexity.

kinkard: I've removed this piece of code in b802645a411c6eba65f4fe0f014e811687c0e1b7

chrstnbwnkl: Couldn't this be handled in the `enhance` stage? There, access restriction units are updated based on the admin they're in. You would have access to the node, and there are spare bits on `AccessRestriction`, so maybe use one to temporarily flip it to signal that it needs to be checked for the hemisphere it's in?

kevinkreiser (approved):

PR #5053, opened 2025-01-16
Fix `valhalla_run_matrix` script

chrstnbwnkl:
The config wasn't passed on to the algos


src/valhalla_run_matrix.cc
```diff
@@ -213,7 +213,12 @@ int main(int argc, char* argv[]) {
   }
 
   // Timing with CostMatrix
-  CostMatrix matrix;
+  CostMatrix matrix(config.get_child("thor"));
+  hierarchy_limits_config_t hl_config =
```
chrstnbwnkl, line 217: While at it, also add reading the hierarchy_limits config

nilsnolde (dismissed):
another low-hanging review fruit:)

PR #5064, opened 2025-01-21
allow multi arch builds in manual docker build

chrstnbwnkl:
Just like we do in the main docker build action when building from latest master

### DIFF
--- a/.github/workflows/docker_build_branch.yml
+++ b/.github/workflows/docker_build_branch.yml
@@ -22,4 +22,10 @@ jobs:
 
       - name: Build latest image for given branch
         run: |
+          sudo apt update --yes
+          sudo apt install --yes --quiet binfmt-support qemu-user-static qemu-system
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_build_branch.yml
```diff
@@ -22,4 +22,10 @@ jobs:
 
       - name: Build latest image for given branch
         run: |
+          sudo apt update --yes
+          sudo apt install --yes --quiet binfmt-support qemu-user-static qemu-system
+          docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+          docker buildx create --name mybuilder
+          docker buildx use mybuilder
+          docker buildx inspect --bootstrap
           docker buildx build --push --platform linux/amd64,linux/arm64 --tag ghcr.io/valhalla/valhalla:branch-${{ github.ref_name }} .
```
kevinkreiser, line 31: i wonder what it did before for the arm target

kevinkreiser (approved):

PR #5066, opened 2025-01-22
Slim down matrix pbf response

chrstnbwnkl:
Just a tiny improvement: like in the JSON serializer, don't put out date time info when `verbose` is set to `false`, so the matrix PBF output is as small as can be.


nilsnolde (approved):
it might be my addition in the JSON serializer, but I'd challenge that today:) if date time is passed, IMO we should return the time info. at least not mangle it with `verbose`. but the "damage" is done, you're just making it consistent which is right.

PR #5078, opened 2025-01-29
Turn lane information for valhalla serializer. Fixes #3435

### LINKED ISSUE #3435: Lane guidance in JSON output format
As I understand it, currently the lane guidance information is only available for the OSRM format. Are there any plans to add this information to the native json format? Is there any reason why this is difficult to do or just lack of time? If the second, I am willing to help. :) Any suggestions on how to store this information in JSON?

molind:
# Issue

Default valhalla json answer lacks turn lane information. I suggest this is minimal solution. What do you think?

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/proto/options.proto
+++ b/proto/options.proto
@@ -524,5 +524,6 @@ message Options {
   bool voice_instructions = 57;                                    // Whether to return voiceInstructions in the OSRM serializer response
   bool dedupe = 58;                                                // Keep track of edges and override their properties during expansion,
                                                                    // ensuring that each edge appears in the output only once. [default = false]
-  bool admin_crossings = 59;                                     // Include administrative boundary crossings
+  bool admin_crossings = 59;                                       // Include administrative boundary crossings
+  bool turn_lane_info = 60;                                        // Include turn lane information into Valhalla serializer response.
 }
--- a/src/tyr/route_serializer_valhalla.cc
+++ b/src/tyr/route_serializer_valhalla.cc
@@ -220,6 +220,38 @@ void locations(const valhalla::Api& api, int route_index, rapidjson::writer_wrap
   writer.end_array();
 }
 
+// Serialize turn lane information
+void turn_lanes(const TripLeg& leg, const DirectionsLeg_Maneuver& maneuver, rapidjson::writer_wrapper_t& writer) {
+  
+  // Read edge from a trip leg
+  if (maneuver.begin_path_index() == 0 || maneuver.begin_path_index() >= leg.node_size())
+    return;
+
+  auto prev_index = maneuver.begin_path_index() - 1;
+  const auto& prev_edge = leg.node(prev_index).edge();
+    
+  if (prev_edge.turn_lanes_size() > 1) {
+    writer.start_array("lanes");
+    
+    for (const auto& turn_lane : prev_edge.turn_lanes()) {
+      writer.start_object();
+            
+      // Directions as a bit mask
+      writer("directions", static_cast<uint64_t>(turn_lane.directions_mask()));
+      
+      if (turn_lane.state() == TurnLane::kActive) {
+        writer("active", static_cast<uint64_t>(turn_lane.active_direction()));
+      } else if (turn_lane.state() == TurnLane::kValid) {
+        writer("valid", static_cast<uint64_t>(turn_lane.active_direction()));
+      }
+      
+      writer.end_object();
+    }
+    
+    writer.end_array();
+  }
+}
+
 void legs(valhalla::Api& api, int route_index, rapidjson::writer_wrapper_t& writer) {
   writer.start_array("legs");
   const auto& directions_legs = api.directions().routes(route_index).legs();
@@ -529,6 +561,11 @@ void legs(valhalla::Api& api, int route_index, rapidjson::writer_wrapper_t& writ
       // “portionsUnpavedNote” : “<portionsUnpavedNote>”,
       // “gateAccessRequiredNote” : “<gateAccessRequiredNote>”,
       // “checkFerryInfoNote” : “<checkFerryInfoNote>”
+
+      // Add Line info if enabled
+      if (api.options().turn_lane_info()) {
+        turn_lanes(*trip_leg_itr, maneuver, writer); 
+      }
 
       writer.end_object(); // maneuver
       maneuver_index++;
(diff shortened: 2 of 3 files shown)


molind:
@kevinkreiser Could you take a look?

kevinkreiser:
yep will do tomorrow or the day after

src/tyr/route_serializer_valhalla.cc
```diff
@@ -220,6 +220,40 @@ void locations(const valhalla::Api& api, int route_index, rapidjson::writer_wrap
...
+  const auto& prev_edge = leg.node(prev_index).edge();
+
+  if (prev_edge.turn_lanes_size() > 1) {
+    writer.start_array("lanes");
+
+    for (const auto& turn_lane : prev_edge.turn_lanes()) {
+      writer.start_object();
+
+      // Directions as a bit mask
+      writer("directions", static_cast<uint64_t>(turn_lane.directions_mask()));
```
kevinkreiser, line 242: just to be pedantic maybe we should call it a mask. also can you add this to the turn by turn documentation? that way we can spell out which bits of the mask mean which direction etc.? or maybe once you have the docs written we dont need to append `_mask` to the item in the json output because the docs will say its a mask?

```suggestion
      // Directions as a bit mask
      writer("directions_mask", static_cast<uint64_t>(turn_lane.directions_mask()));
```

src/tyr/route_serializer_valhalla.cc
```diff
@@ -530,6 +564,11 @@ void legs(valhalla::Api& api, int route_index, rapidjson::writer_wrapper_t& writ
       // “gateAccessRequiredNote” : “<gateAccessRequiredNote>”,
       // “checkFerryInfoNote” : “<checkFerryInfoNote>”
 
+      // Add Line info if enabled
+      if (api.options().turn_lane_info()) {
+        turn_lanes(*trip_leg_itr, maneuver, writer);
```
kevinkreiser, line 569: so this is at the maneuver point. what about including lane information at non-maneuver points? it is sometimes useful to, especially during in car navigation, to show the lane one belongs in even if its not so drastic a situation as to call for giving an active maneuver (like stay right or stay on highway blah blah blah). if we put this here in this way is it precluding us from including these other lane infos? is there a way we can at least plan for that in the future and not shut that out as a possibility? just thinking out loud

kevinkreiser:
this is pretty much ready to go just had some questions and th oughts about the future and maybe we want to make a couple changes just to avoid painting ourselves into a corner

src/tyr/route_serializer_valhalla.cc
molind: When I was thinking about it, it seemed to me that it would be better to keep the JSON response short. If you add a description of the lanes in the documentation, you can just leave 'directions' here.

proto/options.proto
```diff
@@ -524,5 +524,6 @@ message Options {
   bool voice_instructions = 57;                                    // Whether to return voiceInstructions in the OSRM serializer response
   bool dedupe = 58;                                                // Keep track of edges and override their properties during expansion,
                                                                    // ensuring that each edge appears in the output only once. [default = false]
-  bool admin_crossings = 59;                                     // Include administrative boundary crossings
+  bool admin_crossings = 59;                                       // Include administrative boundary crossings
+  bool turn_lane_info = 60;                                        // Include turn lane information into Valhalla serializer response.
```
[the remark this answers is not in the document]
molind: Ok

src/tyr/route_serializer_valhalla.cc
molind: Again, for the sake of brevity, I included lane information only at the maneuver point. In other places, you usually just go straight, and lane information isn’t as important. I was eager to make this change quickly and try out lane information in Guru Maps in practice. (The iOS version already shows maneuver lanes.)

A month of testing has passed. Over 90% of the intersections between maneuvers don’t need additional lane info—you just go straight. There are some large interchanges with two left-turn lanes, two straight, and two right-turn, where you may need to switch lanes. But that’s more of an exception. So the current approach isn’t perfect, but it’s already very good, and I like that it’s concise (unlike the OSRM response).

If we think about improving this simple logic, we’d need to add heuristics to display lanes when a lane changes its allowed direction—say, it used to be straight only, then it becomes right-turn only.

src/tyr/route_serializer_valhalla.cc
kevinkreiser: yes exactly, if we document it then `directions` is sufficient

src/tyr/route_serializer_valhalla.cc
molind: Currently, it's simply `"lanes": [ lane_info ]`.

To make it easier to add information about new intersections in the future, it could be like this:
  
```
"lanes": { shape index: [ lane_info ] }
```

Using the shape index as a key doesn't look very good — we need to convert every index to a string.

Alternatively, later on we can simply introduce `"crossings"`:

```
"crossings": { shape index: [ lane_info ] }
```

and set up another option for this extension.

kevinkreiser:
thank you for bringing this to the top of my inbox this on is on my list to review as well! sorry for the wait

kevinkreiser:
>  It seems to me that if we ever need to include details for more than one intersection at once, it would be a good reason to split that maneuver into two or more.

i have to disagree with this mainly because i know there are production vehicles that do indeed consume lane information even when they arent providing "active" (meaning a command uttered) guidance. they do what is sometimes called "passive" guidance where the show a little visual cue to tell you which lane is the lane you should remain in. its kind of a way of saying "i know you could get in that lane or turn in that lane but you should stay in this lane". it doesnt make sense, at every given choice to tell the person no, stay where you are but it certainly can be useful to show them in a less intrusive way, something to boost their confidence if they need it, that they are in the right place.

if we tie them to maneuvers we preclude that option which i dont think makes sense. i think you're right though that its clunky to have integers as string keys and at the moment valhallas native format doesnt have something like osrms format where they have every intersection or the other larger arrays of attributes. allowing something like this in valhallas native format is probably a larger discussion though so i think it can totally wait for some time in the future

kevinkreiser (approved):

kevinkreiser (approved):

PR #5082, opened 2025-01-30
Use layer caching in Docker builds in GHA 

chrstnbwnkl:
# Issue

Ever since the base image was updated to Ubuntu 24.04, image builds take up to 3 hours in GH actions. This uses Docker's own build/push action that comes with layer caching and can use github action's cache API. Worth a try maybe?


.github/workflows/docker-build.yml
```diff
@@ -29,23 +29,25 @@ jobs:
...
-        run: |
-          sudo apt update --yes
-          sudo apt install --yes --quiet binfmt-support qemu-user-static qemu-system
-          docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
-          docker buildx create --name mybuilder
-          docker buildx use mybuilder
-          docker buildx inspect --bootstrap
-          docker buildx build --push --platform linux/amd64,linux/arm64 --tag ghcr.io/valhalla/valhalla:${{ steps.extract_tag.outputs.tag }} .
-      - name: Build latest image
+        uses: docker/build-push-action@v6
```
kevinkreiser, line 38: but like what is in this image? what operating system is it even?

chrstnbwnkl: I don't fully understand what you're asking. This just runs the docker build inside the current runner, so nothing should change compared to what the process was before, right? It's just a shorthand for running the build steps like we did previously (including the multi-platform configuration) plus now we also get caching, which this action takes care of for us.

kevinkreiser: ok so basically we dont have custom commands these "actions" encapsolate the commands plus hopefully add caching for us? i havent look at github actions much, just a couple of times when trying them out

kevinkreiser:
can you manually run this to see if it works?

.github/workflows/docker-build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v3
```
kinkard, line 18: The dedicated checkout action is probably no longer needed - `docker/build-push-action` uses git context by default https://github.com/marketplace/actions/build-and-push-docker-images#git-context

chrstnbwnkl: Oh yep, good catch. Will take this for a test run shortly.

chrstnbwnkl:
Okay so I found out why the Docker builds were failing, it was related to QEMU. Pinning it to an older version worked but then I found out GitHub recently made linux/arm runners publicly available! Since building the arm image with QEMU was the real bottleneck, I tried out building for each platform natively and using `docker manifest create` to converge the two images into one. That alone made the build time go down to 13 minutes (up from 3 hours!). The layer caching gives a little speed boost too, but since we're not using ccache (and I haven't found an easy way to do so), the caching gain is mostly from not having to to `apt-install` dependencies every time. 

This is an example run with cached dependencies: https://github.com/valhalla/valhalla/actions/runs/13178220740
And this is an example run where the build step was not re-run because it didn't detect any source change (took about a minute, which is great for docs only PRs): https://github.com/valhalla/valhalla/actions/runs/13178897269

For some reason, github doesn't seem to pick up the metadata though of the new converged image: https://github.com/valhalla/valhalla/pkgs/container/valhalla/350568874?tag=debug-action but if I pull and inspect the image manually it looks all good.

.dockerignore
```diff
@@ -60,9 +61,23 @@ CMakeLists.txt.user
...
+run_route_scripts/
+test_requests/
 
-# documentation
-site/
-
-.DS_Store
+# docker
 Dockerfile
+
```
chrstnbwnkl, line 71: a couple more things we don't care about in the docker build

.github/workflows/docker-build.yml
chrstnbwnkl: I tried it without the checkout action but it doesn't work because we sync submodules in the Dockerfile

Dockerfile
```diff
@@ -55,29 +55,30 @@ RUN rm -rf valhalla
 
 ####################################################################
 # copy the important stuff from the build stage to the runner image
-FROM ubuntu:24.04 as runner
-MAINTAINER Kevin Kreiser <kevinkreiser@gmail.com>
+FROM ubuntu:24.04 AS runner
```
chrstnbwnkl, line 58: casing

Dockerfile
```diff
@@ -55,29 +55,30 @@ RUN rm -rf valhalla
...
+LABEL org.opencontainers.image.authors="Kevin Kreiser <kevinkreiser@gmail.com>"
 
 # basic paths
 ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH
 ENV LD_LIBRARY_PATH=/usr/local/lib:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib32:/usr/lib32
 
 # github packaging niceties
-LABEL org.opencontainers.image.description = "Open Source Routing Engine for OpenStreetMap and Other Datasources"
-LABEL org.opencontainers.image.source = "https://github.com/valhalla/valhalla"
+LABEL org.opencontainers.image.description="Open Source Routing Engine for OpenStreetMap and Other Datasources"
```
chrstnbwnkl, line 66: remove whitespace

Dockerfile
```diff
@@ -55,29 +55,30 @@ RUN rm -rf valhalla
...
+LABEL org.opencontainers.image.source="https://github.com/valhalla/valhalla"
+
+# we need to add back some runtime dependencies for binaries and scripts
+# install all the posix locales that we support
+RUN export DEBIAN_FRONTEND=noninteractive && apt update && \
```
chrstnbwnkl, line 71: this can be moved up before copying the build artifacts from the builder so it will always be cached

.github/workflows/docker-build.yml
```diff
@@ -2,50 +2,101 @@ name: Publish Docker image
 on:
   push:
     branches:
-      - 'master'
+      - "master"
```
chrstnbwnkl, line 5: my IDE's yaml linter has some strong opinions here

Issue #5071, opened 2025-01-25
Predicted speeds lower than 5 km/h completely ignored


chrstnbwnkl:
I noticed some odd speeds for a few edges in a Valhalla tile set enriched with predicted traffic sourced from a commercial provider. Visualizing the difference between the source data and what's returned by `GraphTile::GetSpeed()`, I noticed that something happened to values lower than 5 kph:   
![Image](https://github.com/user-attachments/assets/f33a8e30-b777-4434-9be4-6da0bf6caf56)

Turns out that free flow, constrained and live speeds are only taken into account when they are between 5 and 140 kph:

https://github.com/valhalla/valhalla/blob/69a6bfc70aae90a7398315e94c10f1b098c785f0/valhalla/baldr/graphtile.h#L735-L740

https://github.com/valhalla/valhalla/blob/40e898e5025a5be479fc622b10f730673e6ddcc4/valhalla/baldr/graphconstants.h#L114-L116

I would have expected that instead of ignoring these values and using the tagged speed, the speed value would at least be clamped to the valid range, but maybe I'm not seeing the bigger picture?

kevinkreiser:
there are certain low class roads typically going through or near parking lots which will have abnormally low speeds probably because stationary outlier observations. this is our very coarse way of removing those outliersby basically saying no anything outside the range is not valid

radekvermirovsky:
If I’m understanding this correctly, if the live or constrained speed is 5.1, everything is fine, and it’s used. However, if a second later the speed is 4.9, it’s completely ignored and, for example, suddenly jumps to 20.

I understand the need to validate against a range, but I also think clamping toward one of the limit sides makes more sense than totally ignoring this speed. From the picture that @chrstnbwnkl showed, it seems that, for example, the predicted travel time at one moment could be 300 percent (going on that edge 20 kmh instead of 4.9 kmh)  faster (in situation that data says it should be slower)  than it was just one minute earlier.

kevinkreiser:
@radekvermirovsky that's not correct. live speed is not subject to this validity check as far as I remember. it has its own marker for whether it has a speed exists or not. At least that's what I remember. check graphtile:: getspeed

radekvermirovsky:
> [@radekvermirovsky](https://github.com/radekvermirovsky) that's not correct. live speed is not subject to this validity check as far as I remember. it has its own marker for whether a speed exists or not. At least that's what I remember. check graphtile:: getspeed

[...]

PR #5087, opened 2025-02-04
Widen acceptable range of historical speed values 

chrstnbwnkl:
# Issue

See discussion in #5071. This removes the validity check entirely (except for making sure it's larger than zero). I do see @kevinkreiser's point that the curve fitting will lead to some outliers (e.g. in sudden jumps that can occur when predicted speeds are not available during the whole week), but those will be sparse. 

I also understand this change will affect current users bringing their own predicted speed data. On the other hand, I don't think Valhalla should dictate these limits if it is agnostic about the historical traffic source.


chrstnbwnkl:
I added a quick validity check in `valhalla_add_predicted_traffic`. This is least we can do. I considered removing outlier values by default but I think we can only reasonably do this on data that was not previously compressed.

valhalla/baldr/graphconstants.h
```diff
@@ -105,14 +105,15 @@ constexpr uint32_t kMaxAssumedTruckSpeed = 120; // ~75 MPH
...
 constexpr uint32_t kMinSpeedKph = 5; // ~3 MPH
+constexpr uint32_t kMinValidSpeedKph = 1;
 
 // Default Fixed Speed. This is the default fixed speed that is assumed.
 // Unless otherwised specified no fixed speed will be assumed and speed will be
 // calculated from costing algorithm.
 constexpr uint32_t kDisableFixedSpeed = 0; // ~0 MPH
 
-inline bool valid_speed(float speed) {
-  return speed > kMinSpeedKph && speed < kMaxAssumedSpeed;
```
nilsnolde, line 115: hm, I don't think we should get rid of the `kMaxAssumedSpeed` validity check. we might run into trouble on the a* heuristic, which is defaulted to that value, so we could end up having a real cost lower than the a* estimate. also, as you mentioned, there don't seem to be higher values than that at least in the europe tomtom data.

src/mjolnir/valhalla_add_predicted_traffic.cc
```diff
@@ -103,6 +107,9 @@ ParseTrafficFile(const std::vector<std::string>& filenames, stats& stat) {
             case 1: {
               try {
                 traffic->second.free_flow_speed = std::stoi(t);
+                if (is_possible_outlier(static_cast<float>(traffic->second.free_flow_speed)))
+                  LOG_WARN("Detected free flow speed outlier: " +
+                           std::to_string(traffic->second.free_flow_speed) + ".");
```
nilsnolde, line 112:
```suggestion
                stat.lower_bound += speed < kMinSpeedKph
                stat.upper_bound += speed > kMaxAssumedSpeed
```

nilsnolde (changes requested):
apart from aggregating stats instead of logging, I'd also re-instate the upper bound speed check to not introduce (even more) problems with heuristic admissibility.

valhalla/baldr/graphconstants.h
chrstnbwnkl: Good point, but don't we clamp speeds to top_speed, which is what actually informs the heuristic? In that case this check would be redundant.

nilsnolde (approved):
🚢

PR #5024, opened 2024-12-17
Add maneuver bearings in route json response

jthiard:
# Issue

Fixes #5021

## Tasklist

 - [ ] Add tests
 - [x] Add #fixes with the issue number that this PR addresses
 - [x] Update the docs with any new request parameters or changes to behavior described
 - [x] Update the [changelog](CHANGELOG.md)
 - [ ] If you made changes to the lua files, update the [taginfo](taginfo.json) too.

## Requirements / Relations

n/a

PR #5093, opened 2025-02-06
fix maneuver bearing_before in valhalla response

jthiard:
use the correct index to get previous edge end heading

# Issue

Didn't open an issue for that simple fix but let me know if relevant.

When I introduced maneuver bearings into json response format in #5024 I somehow messed up something with the `bearing_before` calculation. Using the maneuver index to get the previous edge was not ok, instead we need to use the node index of the current maneuver.

In some situations the two are the same and everything looks fine, in some other it doesn't.

example:

https://valhalla.openstreetmap.de/directions?profile=pedestrian&wps=2.341643571853638%2C48.84337433623694%2C2.341520190238953%2C48.84313779693215

The third maneuver ("Turn left onto the walkway.") has

- `bearing_before`: 197
- `bearing_after: 197

Which is obviously wrong for a left turn.

The OSRM response format give the good answer which is:

- `bearing_before`: 289
- `bearing_after`: 197

## Tasklist

 - [ ] Add tests

## Requirements / Relations

### DIFF
--- a/src/tyr/route_serializer_valhalla.cc
+++ b/src/tyr/route_serializer_valhalla.cc
@@ -282,7 +282,8 @@ void legs(valhalla::Api& api, int route_index, rapidjson::writer_wrapper_t& writ
       bool depart_maneuver = (maneuver_index == 0);
       bool arrive_maneuver = (maneuver_index == directions_leg.maneuver_size() - 1);
       if (!depart_maneuver) {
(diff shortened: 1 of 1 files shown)


chrstnbwnkl:
I remember thinking about this briefly when reviewing but I tested it on a sample data set and it seemed fine, probably because there both indices happened to be 0 :smile:

PR #5095, opened 2025-02-06
Fix spurious bear left/right maneuvers

### LINKED ISSUE #5094: Spurious "bear left/right" maneuver generation
Under certain conditions, maneuver generation yields distinct maneuvers that don't actually make sense to drivers. That probably isn't very surprising to anyone, but I believe I've narrowed down the cause of at least one annoyance ;)

# Example + analysis

Consider this intersection in Boston.

Image

[OSM link](https://www.openstreetmap.org/#map=19/42.336361/-71.087621)

When traveling northwest on Melena Cass Blvd, Valhalla will generate a "bear right" instruction. Tracing through the major sections of the maneuver generation, it boils down to these factors:

1. OSM models the overall topology, but is rather divorced from the real-world geometry. The "turn" from Melena Cass to the unnamed service road appears to be a 33 degree angle, since the larger road is modeled as two one-ways. A 33 degree transition is classified as a slight right, even though it's really obvious on the ground and even fairly obvious to a human looking at the graph that this isn't really supposed to be a turn. This is a relatively common scenario in the OSM road network.
2. The name changes (to unnamed). If both the "to" and "from" edge were unnamed OR they had the same name, this would not generate a maneuver. But because there is a name change, and this doesn't meet any of the other heuristics in `ManeuversBuilder::CanManeuverIncludePrevEdge`, this edge transition _must_ start a new maneuver (at least initially).
3. In the next phase of maneuver building, where the combinations happen, the transition from Melena Cass to the service road is not eligible for classification as an "obvious" maneuver, because we currently only consider "continue" maneuvers. The angle is too sharp in this case, even though it's pretty obvious to a human driver in this situation.

# Possible solution

At a high level, I think the place to fix this is in `ManeuversBuilder::IsNextManeuverObvious`.

1. Allow slight turn maneuvers (not just continue).
2. Before other "obviousness" checks, ensure that slight turn maneuvers are not possibly ambiguous. `EnhancedTripLeg_Node::HasForwardIntersectingEdge` encapsulates what I think is a reasonable heuristic. Basically it looks for other transitions that are approximately straight (±45 degrees). Crucially, this does **not** check for traversability of the other path, because these cases may not always be obvious.

I'll open a PR with some minimal code changes that seem to improve things.

ianthetechie:
# Issue

Fixes #5094 (or at least substantially mitigates it; can open separate issues for other edge cases as needed)

## Tasklist

 - [ ] Add tests

## Requirements / Relations

N/A

This doesn't appear to break any existing tests, and improves handling of the edge cases that I've tested from real world experience. If maintainers think this is the right approach, I can add some tests and a CHANGELOG.

### DIFF
--- a/src/odin/maneuversbuilder.cc
+++ b/src/odin/maneuversbuilder.cc
@@ -3136,7 +3136,7 @@ bool ManeuversBuilder::IsNextManeuverObvious(const std::list<Maneuver>& maneuver
   if (next_man->type() == DirectionsLeg_Maneuver_Type_kContinue ||
       next_man->type() == DirectionsLeg_Maneuver_Type_kSlightLeft ||
       next_man->type() == DirectionsLeg_Maneuver_Type_kSlightRight) {
-    // Get the node between the the current and next maneuver
+    // Get the node between the current and next maneuver
     auto node = trip_path_->GetEnhancedNode(next_man->begin_node_index());
 
     // Slight turn maneuvers may be obvious ONLY if there are no other forward intersecting edges.
--- a/src/odin/maneuversbuilder.cc
+++ b/src/odin/maneuversbuilder.cc
@@ -3144,8 +3144,7 @@ bool ManeuversBuilder::IsNextManeuverObvious(const std::list<Maneuver>& maneuver
     // multiple forward paths of travel are not obvious
     // (e.g. a slight left to stay on a road,
     // when another "straight" road from the intersection is one way or otherwise non-traversable).
-    if (node &&
-        next_man->type() != DirectionsLeg_Maneuver_Type_kContinue &&
+    if (node && next_man->type() != DirectionsLeg_Maneuver_Type_kContinue &&
         node->HasForwardIntersectingEdge(curr_man->end_heading())) {
       return false;
     }
--- a/src/odin/maneuversbuilder.cc
+++ b/src/odin/maneuversbuilder.cc
@@ -3132,10 +3132,23 @@ bool ManeuversBuilder::AreRampManeuversCombinable(std::list<Maneuver>::iterator 
 bool ManeuversBuilder::IsNextManeuverObvious(const std::list<Maneuver>& maneuvers,
                                              std::list<Maneuver>::const_iterator curr_man,
                                              std::list<Maneuver>::const_iterator next_man) const {
-  // The next maneuver must be a continue maneuver
-  if ((next_man->type() == DirectionsLeg_Maneuver_Type_kContinue)) {
+  // The next maneuver must be a continue maneuver or a slight turn
+  if (next_man->type() == DirectionsLeg_Maneuver_Type_kContinue ||
+      next_man->type() == DirectionsLeg_Maneuver_Type_kSlightLeft ||
+      next_man->type() == DirectionsLeg_Maneuver_Type_kSlightRight) {
     // Get the node between the the current and next maneuver
     auto node = trip_path_->GetEnhancedNode(next_man->begin_node_index());
+
+    // Slight turn maneuvers may be obvious ONLY if there are no other forward intersecting edges.
+    // It does not matter if the forward edge is traversable;
+    // multiple forward paths of travel are not obvious
+    // (e.g. a slight left to stay on a road,
+    // when another "straight" road from the intersection is one way or otherwise non-traversable).
+    if (node &&
+        next_man->type() != DirectionsLeg_Maneuver_Type_kContinue &&
+        node->HasForwardIntersectingEdge(curr_man->end_heading())) {
+      return false;
+    }
 
     // Return true if there are no traversable intersecting edges
     if (node && !node->HasTraversableIntersectingEdge(next_man->travel_mode())) {
--- a/test/astar_bikeshare.cc
+++ b/test/astar_bikeshare.cc
@@ -147,13 +147,13 @@ TEST(AstarBss, test_With_Mode_Changes) {
   std::vector<std::string> expected_route{"Rue Gabriel Vicaire", "Rue Perrée",
                                           "Rue Perrée",          "Rue Caffarelli",
                                           "Rue de Bretagne",     "Rue de Turenne",
-                                          "Rue du Parc Royal",   "Place de Thorigny",
-                                          "Rue de la Perle",     "Rue de la Perle"};
+                                          "Rue du Parc Royal",   "Rue de la Perle",
+                                          "Rue de la Perle"};
 
   const std::map<size_t, BssManeuverType>&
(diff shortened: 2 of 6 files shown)


ianthetechie:
I'm quite surprised to see test failures; everything works locally 🤔

I think I fixed the formatting issues as well.

kevinkreiser:
on the surface the change makes sense to me but there are still test failures. looks like turn lane tests?

ianthetechie:
Sorry it took a bit to get back to the test cleanup, but let's see if we can move it forward :) All tests should pass now.

`turnlanes.cc` - One test was failing; not obvious from the output, but it's the one for `test/pinpoints/turn_lanes/slight_right_active_pinpoint.pbf`. I decoded the test fixture with `cat ../test/pinpoints/turn_lanes/slight_right_active_pinpoint.pbf |  protoc --decode valhalla.Api api.proto` 😂 From there, I grabbed the shape and dumped it in the [demo page](https://valhalla.github.io/demos/polyline/?unescape=true&polyline6=true#yrm%60gApd%7BnhFxQzVlO%7CSpHzKzFrG%60B%7CINpGT%7CJNfDN%60Hj%40lU). This is a textbook example of the sort of improvement that the PR aims to make; perfect! Very similar to the cases that I was working with @haysmike on improving in the Boston area.

image

The bikeshare test changes are similar in nature. Here's a representative sample; the middle maneuver gets combined now, as it should. Most are pretty clear.

image

The pedestrian bikeshare test is worth a quick note. if you take a look at the area in OSM today, you'll probably come to the conclusion that it's a confusing intersection which should be treated otherwise, but inspecting the graph as it was generated for the test fixtures, the behavior is correct.

`test_route_summary.cc` - Already discussed above.

`test_tagged_values.cc` - Tunnel test; the exit maneuver could be improved IMO (there isn't really a left turn per se...), but the test is overall improved. We can improve these sorts of cases in a future pass.

aki1770-del:
Thank you for this fix — and apologies it has been waiting so long for attention.

I came to this issue independently and traced the same root: `IsNextManeuverObvious()` only handles `kContinue`, so a slight-turn maneuver forced by a name change on a dual carriageway slips through and becomes a spurious "bear left/bear right". When I read the code I found your PR already here with exactly the right approach — using `HasForwardIntersectingEdge` to guard the slight-turn case.

Your reasoning on using the non-traversable version of the check (a non-traversable forward edge still creates driver ambiguity) is correct, and I think it is the safer choice.

If it would help move this forward, I am happy to add the CHANGELOG entry, rebase on master, or assist with any remaining CI issues. Let me know what would be most useful.

*AI-assisted — authored with Claude, reviewed by Komada.*

ianthetechie:
Just a quick update on this... I think that some parts of this are actually incorrect.  I will try to revisit soon but it is sorta working for the wrong reasons in some cases due to my lack of understanding the layers of complexity and I haven't had time to revisit.

ianthetechie:
Closing this for now. After some extensive tests with a customer, we found that the fixes 1) were not actually targeting all of the behaviors correctly, and 2) introduced other subtle narrative gen regressions, so it was a wash.

I'm still interested (+ working with a customer on) ways we can improve narrative gen but it's pretty complicated, and understanding Odin's data flow, branches, and assumptions has consumed... a lot of brain cells 😂

aki1770-del:
Thanks for the honest update @ianthetechie — and better a wash caught in customer testing than a subtle regression shipped. The layered complexity in Odin's maneuver logic is real; I hit the same `IsNextManeuverObvious()` edge independently, so I sympathize with the brain cells 😄. If it's ever useful to have another set of eyes on a specific narrative-gen case, I'm around, and I'll keep watching here. Appreciate you pushing on it.

nilsnolde:
@aki1770-del anyone more familiar & _interested_ in odin would be a bliss for the project. The one requirement I had for almost all Valhalla jobs, was to not work on Odin haha partly bcs of the code craziness (which I'm sure most of it is warranted!), but also bcs guidance for some reason doesn't interest me as much.

That's to say, if we can at least have you as reviewer or even contributor of guidance code, huge win for us:)

aki1770-del:
Thanks @nilsnolde — I'd genuinely like that. Happy to be a reviewer on guidance/Odin PRs and to dig into specific narrative-gen cases as they come up; the maneuver-obviousness logic is a fun kind of gnarly to reason about.

Honest about scope so I'm not overpromising: my main work is a snow-driving navigation stack, where guidance matters a lot — a driver in a whiteout needs the *right* "bear left" at the right moment — so this is close to what I care about, but I can't commit full-time Odin hours. Where I can genuinely help is reviewing guidance PRs and taking on concrete maneuver bugs. Tag me on anything guidance-related and I'll take a look.

And yeah — Odin's complexity is real; I get why it's the job everyone dodges 😄. Glad to lend eyes.

nilsnolde:
Oh man, so happy to hear that!! Thanks so much, and even more for cleanly stating expectations!

Otherwise I'd tag you now on 10-20 existing issues haha no worries, now that I know your project, I can maybe better judge if a bug would even be relevant to you as well. Just for the motivational kick:) I know you probably meant general Odin PRs. It's rare that we have them so far, the main problem being a lack of heavily used guidance apps.

@ianthetechie et al are giving it a real shot though over in https://github.com/stadiamaps/ferrostar.

ianthetechie:
Yeah exactly. We have several customers (some of them really caring about guidance) on Ferrostar. The main challenge with Odin is that it's so poorly understood. The architecture is currently such that it doesn't lend itself to thinking in domain concepts. It's very low level and has a dizzying number of branches.

Secondarily. this is a legit hard problem, and Odin actually does a really good job most of the time in spite of that :)

So I'm currently trying to think about it from first principles and what I'd like to see in such a system long term. I'll post aronud here about it once I have something shareable, but I suspect I'll be running most of the experiments in [Valinor](https://github.com/stadiamaps/valinor) as a microservice that (optionally) replaces Odin on a per-request basis as a testbed. It'll all be open source, so if we end up hitting on something that works well + there's interest, we could help port it back over to Odin.

aki1770-del:
This really resonates @ianthetechie. Coming at Odin fresh, the thing that struck me is exactly what you said — it's nearly all mechanism and very little domain vocabulary, so you end up reasoning about turn-degree bands and edge-topology flags instead of "would a driver find this turn obvious." A first-principles pass that lets you think in domain concepts sounds like the right instinct, and a per-request-replaceable Odin testbed is a clever way to iterate without dragging the whole pipeline along. I'd genuinely love to follow the Valinor experiments. And honestly the guidance layer is squarely relevant to what I'm building (a snow-driving nav stack), so if a second set of eyes on the domain-modeling is ever useful as it takes shape, I'm in. Looking forward to whatever you post.

